PR #1087, opened 2015-11-18
Remove extraenous log dir and add catalina.out log rotation.

PaulAngus:
remove unused /var/log/cloudstack-management directory
add cloudstack-catalina to logrotate.d directory so that /var/log/cloudstack/management/catalina.out is rotated.

### DIFF
--- a/packaging/centos63/cloud.spec
+++ b/packaging/centos63/cloud.spec
@@ -581,7 +581,7 @@ fi
 %{_defaultdocdir}/%{name}-management-%{version}/LICENSE
 %{_defaultdocdir}/%{name}-management-%{version}/NOTICE
 %attr(0644,cloud,cloud) %{_localstatedir}/log/%{name}/management/catalina.out
-%attr(0644,root,root) %{_sysconfdir}/logrotate.d/%{name}-catalina
+%config(noreplace) %{_sysconfdir}/logrotate.d/%{name}-catalina
 
 %files agent
 %attr(0755,root,root) %{_bindir}/%{name}-setup-agent
--- a/packaging/centos63/cloud.spec
+++ b/packaging/centos63/cloud.spec
@@ -581,7 +581,7 @@ fi
 %{_defaultdocdir}/%{name}-management-%{version}/LICENSE
 %{_defaultdocdir}/%{name}-management-%{version}/NOTICE
 %attr(0644,cloud,cloud) %{_localstatedir}/log/%{name}/management/catalina.out
-%config(noreplace) %{_sysconfdir}/logrotate.d/%{name}-catalina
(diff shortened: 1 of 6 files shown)


wilderrodrigues:
@PaulAngus, every PR needs an issue on Apache Jira.

In addition, how did you test your changes? Some input will help us to test it as well.

Cheers,
Wilder

PaulAngus:
Apologies.
https://issues.apache.org/jira/browse/CLOUDSTACK-9073  created

removal of /var/log/cloudstack-management
and addition of catalina log rotation in /etc/logrotate.d/
has been running against 4.5.1 for 2 months with no issues.

updated packaging was built against current master and (centos6 and 7) deployed with files/directories as expected.
Left running overnight /var/log/cloudstack/management/catalina.log seen to rotate and compress previous file.

jburwell:
@PaulAngus the change looks reasonable.  Have you checked upgrading the RPM to ensure that the any modified logrotate configuration files are left alone?

bhaisaab:
LGTM, here's a rpm repo built out of this patch: http://sb.bhaisaab.org/cloudstack/paul-rpmpr-1087/centos/paul-rpmpr-1087/

remibergsma:
LGTM, tested on CentOS7:

Logrotate is there:

```
[root@cs1 x86_64]# cat /etc/logrotate.d/cloudstack-catalina 
/var/log/cloudstack/management/catalina.out {
    copytruncate
    daily
    rotate 14
    compress
    missingok
    create 0644 cloud cloud
}
```

And this folder is missing (expected):
```
ls -la /var/log/cloudstack-management/
ls: cannot access /var/log/cloudstack-management/: No such file or directory
```

remibergsma:
@PaulAngus please respond to the comment by @jburwell. After that we can merge.

PaulAngus:
So looking at the rpm spec options using +%config(noreplace) will cause a new version of cloudstack-catalina (cloudstack-catalina.rpmnew) to be created if cloudstack-catalina has been edited.
However as the contents of /etc/logrotate.d/ are enumerated and run this would result in both the new and old behaviour both being applied - which seems the worst of all worlds.
Therefore I suggest the orginal PR (without commit ec698f5377ad98ac2a34488a0102c45c2408d089) which will overwrite if the new rpm has an updated value in it [1]; is the lesser of two evils.

[1] http://www-uxsup.csx.cam.ac.uk/~jw35/docs/rpm_config.html

jburwell:
@PaulAngus sounds like a reasonable solution to preserve custom logrotate configurations on update.  Can you push update the PR to remove commit ec698f5?

bhaisaab:
@PaulAngus please rebase against latest master and force push

rhtyd:
@PaulAngus can you rebase against latest master and share state of your PR, thanks

PR #1088, opened 2015-11-18
CLOUDSTACK-9053 security upgrade as per COLLECTIONS-580

DaanHoogland:
cloustack is not vulnerable but as the classes are in they might
  be used in the future so we upgrade to prevent accidental
  vulnerabilities.

build, unit tested and starting regression tests on it now


DaanHoogland:
should have been against 4.6

PR #1038, opened 2015-11-05
Metrics views for CloudStack UI

bhaisaab:
FS: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Metrics+Views+for+CloudStack+UI
JIRA: https://issues.apache.org/jira/browse/CLOUDSTACK-9020

PR #1090, opened 2015-11-19
[4.6] metrics view

bhaisaab:
From #1038, for 4.6 branch


remibergsma:
Merged to master, cannot merge to 4.6 as it would conflict when forward merging. It will be in 4.7.

PR #1091, opened 2015-11-19
[4.5] Metrics view

bhaisaab:
From #1038, for 4.5 branch


bhaisaab:
Merging on 4.5 based on no objections and popular demands.

karuturi:
Please wait. This lands a feature/fix/enhancement in 4.5 and 4.7 while not in 4.6. This will just confuse the users.

This basically means that votes for only in 4.7 on the PR  #1038  aren't valid while its already merged. We have diverged opinion here with respect to few wanting it in 4.5+ as enhancement and a few on 4.7+ as a feature. Let us not count them independently as +1. They are actually related. +1 for 4.5+ means -1 for 4.7+ and vice-versa.

We decided to fwd-merge stuff and let us please stick to it.

bhaisaab:
@karuturi I can revert the metrics view feature, but would keep the other css/js improvements wrt sorting etc.

bhaisaab:
@remibergsma @karuturi this is confusion, we've been backporting changes for long. These are simply UI changes that don't affect backend/core/mgmt server. I would say, it helps to include them in 4.5/4.6 branches. If there is no strict rule, I want to include them in 4.5 because users are not going to simply upgrade every time we do a major release, it makes sense to backport bugfixes and small changes that can be consumed in minor releases.

karuturi:
I am ok with merging the fix to 4.5+. I am not against that. But, missing one release is not acceptable. Given that its already merged on 4.7, merging in 4.5 now is not an option. I am ok with merging css/js improvements in 4.5+ releases.
Yes, we have been backporting fixes since a long time and thats what we are trying to change. We fix forward now.
If the upgrades are proper with minimal set of features and code change, users will upgrade. Right now they dont because its broken. They dont know if the issues fixed in the current release will be regressed in the next release. There can be hesitation for 5.0 release but 4.7 should be smooth if we fwd merge. Anyone following the release process closely will be able to figure that out (and we are doing quick releases)
Lets not change the semantics of version number (MAJOR.MINOR.PATCH)

karuturi:
For future reference: this PR is reverted in commit https://github.com/apache/cloudstack/commit/ea5aecbe07f360bdf5f461c5ea2bce73c3a0e7df

PR #1097, opened 2015-11-20
L10N : Updated locale files with latest translation.

barunhalder1234:


remibergsma:
Pinging @milamberspace to have a look.

milamberspace:
@barunhalder1234 I thinks you don't use the Transifex website to update the translations?

For example on Transifex (for 4.6/4.7 resources) the Japanese translation for the key label.ha.enabled (HA Enable in US string) is: 
高可用性有効
(translate in Google Translator: Enable high availability)

But in your PR the same key is translated with:
高可用性を有効にする
(translate in Google Translator: We want to enable high availability)

(curiously the current japanese string on transifex seems better than your update string...)

The issue is that the transifex Japanese / Chinese resources files aren't up-to-date, so next time, when we update the resources files from transifex to the git repo, we lost all your updated strings.

The right way to translate the Web UI of CloudStack is to use the Transifex project [1] and use the script inside the cloudstack project to download and prepare the resources files (unicode/native conversion and alphabetical order) [2]

To deal with this issue, you can update the transifex JA and CH resources files with the Transifex client. (See this script [3]) I can do this if you prefer.

[1] https://www.transifex.com/ke4qqq/CloudStack_UI/
[2] https://cwiki.apache.org/confluence/display/CLOUDSTACK/Update+L10N+files+from+Transifex+to+git+repo
[3] https://github.com/apache/cloudstack/blob/master/tools/transifex/sync-transifex-ui.sh

cc @remibergsma

PR #1101, opened 2015-11-21
[4.6] Add select template dropdown when reinstall VM

ustcweizhou:


remibergsma:
@ustcweizhou LGTM, tested it and works as expected.

Minor comment: the first item in the list is empty (it will then keep the same template), rest of the items are the available templates. Should we make that first entry more explicit ("keep same template" or so)?

screen shot 2015-11-28 at 14 28 31 pm

Tried both and they work fine:
screen shot 2015-11-28 at 14 36 32 pm

Thanks!

remibergsma:
@bhaisaab @DaanHoogland Can you guys also review please?

DaanHoogland:
@remibergsma I had already seen the code but missed the PR; LGTM

PR #1103, opened 2015-11-22
update release to 4.7.0 so it use proper packages and systemvm

pdion891:
This update container creation of master branch to 4.7.0 version. 

It doesn't fix the /dev/loop0 issue.

### DIFF
--- a/tools/build/build_asf.sh
+++ b/tools/build/build_asf.sh
@@ -96,11 +96,13 @@ perl -pi -e "s/<cs.xapi.version>6.2.0-1-SNAPSHOT<\/cs.xapi.version>/<cs.xapi.ver
 perl -pi -e "s/-SNAPSHOT//" tools/checkstyle/pom.xml
 perl -pi -e "s/-SNAPSHOT//" deps/XenServerJava/pom.xml
 perl -pi -e "s/-SNAPSHOT//" tools/apidoc/pom.xml
-perl -pi -e "s/-SNAPSHOT//" Dockerfile
 perl -pi -e "s/-SNAPSHOT//" build/replace.properties
 perl -pi -e "s/-SNAPSHOT//" services/console-proxy/plugin/pom.xml
 perl -pi -e "s/-SNAPSHOT//" tools/marvin/setup.py
 perl -pi -e "s/-SNAPSHOT//" tools/marvin/marvin/deployAndRun.py
+perl -pi -e "s/-SNAPSHOT//" tools/docker/Dockerfile
+perl -pi -e "s/-SNAPSHOT//" tools/docker/Dockerfile.marvin
+perl -pi -e "s/-SNAPSHOT//" tools/docker/Dockerfile.centos6
 case "$currentversion" in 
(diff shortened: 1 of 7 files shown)


remibergsma:
Thanks @pdion891! I agree with all of them, except the systemvm template. There is no reason to bump that version because there is no change yet. As soon as we introduce a change that is not compatible with 4.6.0 template, we should bump the version. Until that happens, I'd like to be able to have 4.7.0 to use the same template as 4.6.0.

In the future we should get a separate systemvm version numbering schema (separate repo even?), to break with the "new cloudstack version == new systemvm version". Upgrades are much easier / less impacting if one does not need to recreate all sytemvms (1000+ in our case). Especially when we're releasing more often.

Pinging @DaanHoogland @bhaisaab to see what they think and if I maybe forget something? Can we keep the version of systemvms at 4.6.0 until something changes? If not, can we make it work?

remibergsma:
@pdion891 See comments on https://issues.apache.org/jira/browse/CLOUDSTACK-9077 for possible solution for `/dev/loop0` issue.

DaanHoogland:
@remibergsma you are right, we don't need to up the sysvm template version yet. For other version numbers mentioned, @pdion891 can you add thos location in the build_asf.sh and setnextversion.sh scripts?

bhaisaab:
@remibergsma I think we will need to create new systemvmtemplate for 4.7.0 if we'll get the openswan->strongswan patch merged.

remibergsma:
@bhaisaab Sure, that would be a good reason. But it's not ready so let's bump when we merge that PR.

pdion891:
I have updated release scripts as per @DaanHoogland comments  but didn't change template version, based on @bhaisaab I would suspect we will upgrade sysvm on 4.7, soon ?

Thanks @DaanHoogland  for the location of those scripts, I was wondering how to automate this version changes! :)

remibergsma:
@pdion891 @bhaisaab In about two weeks (Dec 7) we freeze for 4.7.0. That would be a nice time to look back and see if the systemvm template needs to be bumped or not. Right now, there is no PR merged yet that requires a new version, so I don't want to bump it now. It's just too damn annoying upgrading all systemvms if there is no reason. Once there is, we will bump it of course. 

Let me add it to the release procedure, to check at freeze time whether the systemvm version needs to be bumped. We will most likely have releases coming where we don't need to bump the version.

Also @pdion891 could you please squash the commits. Thanks!

remibergsma:
@pdion891 I'm OK with this, but I think you do not need to change the version in `tools/docker/systemtpl.sh` until there is a 4.7 version. Otherwise it will result in a 4.4

bhaisaab:
@pdion891 LGTM, but you'll need to fix the version again for master/4.8 and squash the commits

bhaisaab:
@pdion891 please rebase against latest and squash commits

pdion891:
Thanks for the reminder Rohit. I will update this PR accordingly.

pdion891:
replaced by #1435

PR #1112, opened 2015-11-23
Fix Instance creation with swift as Secondary Storage [CLOUDSTACK-9061]

syed:
Swift is currently broken when used as Secondary storage. This fix does the right thing when creating directories on the NFS staging server. We should think of a better solution as this should have been common code path between S3 and Swift. We should also make the "Object Storage as Secondary Storage" interface generic so that we can add other object stores too. 

-Syed


remibergsma:
Pinging @pdion891 to have a look.

pdion891:
@remibergsma, yep, let me test this, we would also like to have this into 4.6.1 if we can test it on time. 
Thanks.

syed:
Guys, I am closing this and opening this request against 4.6.1 as I have been told that it is the new way now.

PR #1113, opened 2015-11-24
CLOUDSTACK-9064: The users should be able to create multiple Guest Sh…

abhinandanprateek:
Fix for https://issues.apache.org/jira/browse/CLOUDSTACK-9064

The users should be able to create multiple Guest Shared Networks in same Vlan ID, same Physical Network and same network, just with a different IP ranges.


bhaisaab:
LGTM (yet to test it)

remibergsma:
What about 4.6?

PR #1117, opened 2015-11-25
CLOUDSTACK-9064: The users should be able to create multiple Guest Sh…

abhinandanprateek:
https://issues.apache.org/jira/browse/CLOUDSTACK-9064
    
    CLOUDSTACK-9064: The users should be able to create multiple Guest Shared Networks in same Vlan ID, same Physical Network and same network, just with a different IP ranges.


bhaisaab:
LGTM

bhaisaab:
Merging on 4.5, since the fix has been tested and merged on 4.6

PR #1106, opened 2015-11-23
[4.6] CLOUDSTACK-9020: UI enhancements from metrics view

bhaisaab:
- Sortable columns
- Logos with cloudmonkey
- Wider UI
- CSS fixes and new status icons
- Configurable UI list api requests page size
- IP address column on instances page
- Listview widget enhancements

cc @remibergsma @karuturi

PR #1119, opened 2015-11-25
[master] ui: Fix undefined check in instances.js, regression from 459d638

bhaisaab:
(fix is already included in the https://github.com/apache/cloudstack/pull/1106 for 4.6 as well)


remibergsma:
PR #1136 will bring this to master.

PR #1135, opened 2015-11-27
Reducing expunge delay and interval on default dev cfgs

pdube:
Significantly reduces the time it takes to execute the integration tests

### DIFF
--- a/setup/dev/advanced.cfg
+++ b/setup/dev/advanced.cfg
@@ -186,7 +186,7 @@
         },
         {
             "name": "expunge.delay",
-            "value": "60"
+            "value": "5"
         },
         {
             "name": "vm.allocation.algorithm",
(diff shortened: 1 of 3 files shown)


remibergsma:
@pdube Nice one, will give it a try!

remibergsma:
@pdube Had a look and you may also want to tweak these settings:

# Network garbage collection
network.gc.wait
network.gc.interval

# Storage/Account cleanups
storage.cleanup.interval
account.cleanup.interval

Furthermore I'd set default.page.size as the new infra tab in the UI doesn't like 10000 (which is quite large anyway).

Going to test these and will let you know. Thanks!

pdube:
I will check it out!

remibergsma:
FYI: @pdube I set all timers to 5 sec but that didn't result in a faster integration test run for me.

DaanHoogland:
@pdube @remibergsma and i have been noticing extra failures in the integration tests with such short expunge timeouts. maybe it should work and can with some extra work but like this I don't like :(

pdube:
@DaanHoogland np, I will close the PR

PR #1136, opened 2015-11-28
4.6 to master

DaanHoogland:
two conflicts:
  ui/scripts/system.js
  ui/scripts/instances.js

### DIFF
--- a/ui/scripts/instances.js
+++ b/ui/scripts/instances.js
@@ -397,11 +397,13 @@
                     data: data,
                     success: function(json) {
                         var items = json.listvirtualmachinesresponse.virtualmachine;
-                        $.each(items, function(idx, vm) {
-                            if (vm.nic && vm.nic.length > 0 && vm.nic[0].ipaddress) {
-                                items[idx].ipaddress = vm.nic[0].ipaddress;
-                            }
-                        });
+                        if (items) {
+                            $.each(items, function(idx, vm) {
+                                if (vm.nic && vm.nic.length > 0 && vm.nic[0].ipaddress) {
+                                    items[idx].ipaddress = vm.nic[0].ipaddress;
+                                }
+                            });
+                        }
                         args.response.success({
                             data: items
                         });
--- a/ui/scripts/system.js
+++ b/ui/scripts/system.js
@@ -7805,7 +7805,7 @@
                                             return 'label.metrics';
                                         }
                                     }
-                                },
+                                }
                             },
 
                             detailView: {


DaanHoogland:
@bhaisaab please make sure I made the right merge decisions. @karuturi @remibergsma second LGTM?

remibergsma:
LGTM, since the change is in the UI I built that from this branch.

Two checks:
- The metrics feature still works properly
screen shot 2015-11-28 at 20 58 26 pm

- The bug mentioned in #1119 is resolved, as the fix was merged to 4.6 and now lands in master.
The instance tab now shows "No data to show" instead of an endless loading loop.
screen shot 2015-11-28 at 20 55 05 pm

Thanks @DaanHoogland for resolving the merge conflict.

bhaisaab:
@remibergsma @DaanHoogland the merge conflict resolution did not include the fix for master, I've merged #1119 to address that. `git diff master..4.6 ui/scripts` confirms that changes now.

remibergsma:
@bhaisaab The change in PR #1119 is the same as in this PR, except for an extra comma?

Anyway, we should now close this PR @DaanHoogland as the result now also is none.

remibergsma:
@bhaisaab The problem with your approach, to be honest, is that the commit ID changed. As this was a bit of a special case it may not be a problem. General speaking, the forward merging approach is the better one as it allows for tracking back the commit ids to all branches a commit is in.

bhaisaab:
@remibergsma I used the git pr tool so not sure how the commit ID changed, I merged the other PR because 4.6 merge on master chose the other change while merge conflict resolution which did not solve the regression fixed in 4.6.

Fwd-merging is fine, but the way we're merging Github PRs it's difficult to track changes even using --no-merges with git log. IMO the git log history is practically unreadable or followable, along with the fact that we don't squash merge changes for each github PR.

remibergsma:
@bhaisaab The commit ID didn't change from the PR and your merge, that was fine. It was just different in 4.6 branch because the change was done combined with something else there. Never mind, it's fine this way and the end result is also great: 4.6 had an improved UI and 4.7 has that + metrics. Thanks!

I think we should discuss the squashing on the list. I'm fine, as long as we decide something and then live it. Let's not do it here ;-)

DaanHoogland:
@bhaisaab you should not have merged #1119 as it is also in 4.6. This merge commit was wrong because of this isse and should have been fixed before merging. Now the same thing is doen in two different commits. the idea is that you always merge forward so you can always track the branches a change is in. Squashing??? we don't go back to the mess we were in are we?

PR #1140, opened 2015-11-30
CLOUDSTACK-9094: Multiple threads are being used to collect the stats…

harikrishna-patnala:
CLOUDSTACK-9094: Multiple threads are being used to collect the stats from the same VR

Same thread is being intialised by two managers, VirtualNetworkApplianceManager and VpcVirtualNetworkApplianceManager

### DIFF
--- a/server/src/com/cloud/network/router/VpcVirtualNetworkApplianceManagerImpl.java
+++ b/server/src/com/cloud/network/router/VpcVirtualNetworkApplianceManagerImpl.java
@@ -685,6 +685,16 @@ public class VpcVirtualNetworkApplianceManagerImpl extends VirtualNetworkApplian
     }
 
     @Override
+    public boolean start() {
+        return true;
+    }
+
+    @Override
+    public boolean stop() {
+        return true;
+    }
+
+    @Override
     public boolean startRemoteAccessVpn(final RemoteAccessVpn vpn, final VirtualRouter router) throws ResourceUnavailableException {
         if (router.getState() != State.Running) {
             s_logger.warn("Unable to apply remote access VPN configuration, virtual router is not in the right state " + router.getState());


server/src/com/cloud/network/router/VpcVirtualNetworkApplianceManagerImpl.java
```diff
@@ -685,6 +685,16 @@ public void finalizeStop(final VirtualMachineProfile profile, final Answer answe
     }
 
     @Override
+    public boolean start() {
+        return true;
```
bhaisaab: @harikrishna-patnala Hi Hari, can you explain what the changes are doing to fix the issue of multiple thread synchronisation ? Also, what happens across multiple mgmt servers since we don't have any distributed consensus implementation in place, I suppose the statcollector values might be different.

koushik-das:
LGTM based on code changes.

@bhaisaab Since VpcVirtualNetworkApplianceManager is derived from VirtualNetworkApplianceManager, start() call on both managers ended up calling VirtualNetworkApplianceManager.start(). This resulted in the same threads/executors to get initialised/started twice.

bhaisaab:
@koushik-das thanks, in that case LGTM. 
Will merge now.

DaanHoogland:
Guys, this got merged on only code review. No tests are described showing the old and new behaviours. No unit tests are added. No regression tests are added. It is like this fix is not important.

My intuition says that if the VpcVirtualNetworkApplianceManager.start() is called then VirtualNetworkApplianceManager.start() should not have been and the real bug is not in this code at all.

bhaisaab:
@DaanHoogland let's revert, and ask @harikrishna-patnala to add more tests?

koushik-das:
Actually I was looking at the history and this is what I found. This was originally fixed with commit

commit e444a03819ccf72f61cb04e8428d20cc65b145e1
Author: Wei Zhou <w.zhou@leaseweb.com>
Date:   Thu Nov 14 16:51:28 2013 +0100

    remove duplicated scheduled tasks from VpcVirtualNetworkApplianceManagerImpl


Then was removed with commit

commit a5d6f90f666798e7575c6b7d3af5b3024af8768e
Author: wilderrodrigues <wrodrigues@schubergphilis.com>
Date:   Mon Jan 26 17:33:50 2015 +0100

    Implementing redundant router arguments to add redundant_state
    Implementing the arguments on the python side


And then fixed again with this PR. Not sure why it got removed at the first place.

bhaisaab:
@koushik-das good find, that means we also need to fix this for 4.6. We would still need a round of tests just in case.

@harikrishna-patnala can you open a PR against 4.6 branch (and 4.5 branch if needed). Thanks.

DaanHoogland:
Thanks for picking this up, guys.

I still think there is calling code that should not be allowed to call start twice. It should skip the VirtualNetworkApplianceManager.start() in case we are dealing with a VPC.

More important this has been allowed to become regression because it was not put into tests.
/me is looking for an :i told you so: emoticon.

ustcweizhou:
@DaanHoogland the fix is simple. We do not need special /extra process for VPC routers.
@bhaisaab this issue does not exist in 4.5, as @koushik-das mentioned, it is a regression bug in 4.6.

LGTM
this is almost same to commit e444a03, so we can merge it or cherry pick the old commit

ustcweizhou:
sorry guys, just notice it is already merged.

DaanHoogland:
@wilderrodrigues can you have a look? you rmoved the code that is reintroduced with this PR.

wilderrodrigues:
@bhaisaab 

I thought only RMs would merge PR... isn't that so?

Plus, the PR was merged without one single person executing any test against it.

@remibergsma, could you please revert it so someone can actually test the change?

Cheers,
Wilder

DaanHoogland:
@wilderrodrigues it was merged before the freeze so I didn't revert it as RM. But you are right about the not being tested.

DaanHoogland:
@harikrishna-patnala This is reverted because of the lack of test report, not the logic of your code. If you feel it must go in anyway please rebase the code and make a new PR, make sure you add test instruction to up the chance reviewers actually test.

PR #1146, opened 2015-11-30
make sure all files are updates with new version

remibergsma:
During the previous release I had to edit these manually. This will fix that so all versions get updated.

### DIFF
--- a/tools/build/build_asf.sh
+++ b/tools/build/build_asf.sh
@@ -101,6 +101,11 @@ perl -pi -e "s/-SNAPSHOT//" build/replace.properties
 perl -pi -e "s/-SNAPSHOT//" services/console-proxy/plugin/pom.xml
 perl -pi -e "s/-SNAPSHOT//" tools/marvin/setup.py
 perl -pi -e "s/-SNAPSHOT//" tools/marvin/marvin/deployAndRun.py
+perl -pi -e "s/-SNAPSHOT//" debian/changelog
+perl -pi -e "s/-SNAPSHOT//" services/iam/plugin/pom.xml
+perl -pi -e "s/-SNAPSHOT//" services/iam/pom.xm
+perl -pi -e "s/-SNAPSHOT//" services/iam/server/pom.xml
+
 case "$currentversion" in 
   *-SNAPSHOT*)
     perl -pi -e 's/-SNAPSHOT//' debian/rules
--- a/tools/build/build_asf.sh
+++ b/tools/build/build_asf.sh
@@ -127,12 +127,8 @@ mv $tmpfilenm debian/changelog
 git clean -f
 
 #create a RC branch
-RELEASE_BRANCH="RC"`date +%Y%m%dT%H%M`
-if [ "$branch" = "master" ]; then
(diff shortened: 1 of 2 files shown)


tools/build/build_asf.sh
```diff
@@ -101,6 +101,11 @@ perl -pi -e "s/-SNAPSHOT//" build/replace.properties
 perl -pi -e "s/-SNAPSHOT//" services/console-proxy/plugin/pom.xml
 perl -pi -e "s/-SNAPSHOT//" tools/marvin/setup.py
 perl -pi -e "s/-SNAPSHOT//" tools/marvin/marvin/deployAndRun.py
+perl -pi -e "s/-SNAPSHOT//" debian/changelog
+perl -pi -e "s/-SNAPSHOT//" services/iam/plugin/pom.xml
```
DaanHoogland: to bad the poms don't get changed by the mvn versions:set task :(

DaanHoogland:
Looks good to me. The version-branch should have gone in like this from the get go.

tools/build/build_asf.sh
```diff
@@ -101,6 +101,11 @@ perl -pi -e "s/-SNAPSHOT//" build/replace.properties
...
+perl -pi -e "s/-SNAPSHOT//" services/iam/pom.xm
+perl -pi -e "s/-SNAPSHOT//" services/iam/server/pom.xml
```
pdube: Why do you have to add these manually? What about the other folders in /services?

tools/build/build_asf.sh
```diff
@@ -123,11 +128,7 @@ git clean -f
 
 #create a RC branch
 RELEASE_BRANCH="RC"`date +%Y%m%dT%H%M`
```
pdube: Might be clearer to name this variable RC_BRANCH_SUFFIX

remibergsma:
The build script now makes a branch that makes sense: `4.6.1-RC20151130T2206` instead of `4.6-RC20151130T2206`. Also the versions are updated properly.

Output example:
screen shot 2015-11-30 at 22 07 02

tools/build/build_asf.sh
remibergsma: It is supposed to go automatically, but it doesn't pick these ones up. It should be investigated why that is and if these are still used.

tools/build/build_asf.sh
remibergsma: Yep, changed it.

pdube:
Ok then, LGTM

rafaelweingartner:
It LGTM, 
I would just recommend using “sed” in combination with “find” to look up for every file that we want to change the version number, instead of perl, but that is more a matter of taste.

remibergsma:
@rafaelweingartner I agree and I think this needs some TLC indeed. Investigate why it doesn't work with mvn and then do one global search/replace for the remaining. I'll merge this now and proceed with 4.6.1RC, then make the improvement later.

pdion891:
LGTM

rafaelweingartner:
@remibergsma , I think it would have been a good idea to have squashed your commits

PR #1152, opened 2015-12-02
CLOUDSTACK-9099: SecretKey is returned from the APIs - Fixed

kansal:
The current implementation of User and account management API (in general) return the secret key as a user or account response. 
Fix: Added a new API to explicitly return the secretKey and removed it from the user and account response.

### DIFF
--- a/api/src/com/cloud/user/AccountService.java
+++ b/api/src/com/cloud/user/AccountService.java
@@ -16,17 +16,20 @@
 // under the License.
 package com.cloud.user;
 
+import java.util.List;
 import java.util.Map;
 
 import org.apache.cloudstack.acl.ControlledEntity;
 import org.apache.cloudstack.acl.RoleType;
 import org.apache.cloudstack.acl.SecurityChecker.AccessType;
+import org.apache.cloudstack.api.command.admin.user.GetUserKeysCmd;
 import org.apache.cloudstack.api.command.admin.user.RegisterCmd;
 
 import com.cloud.domain.Domain;
 import com.cloud.exception.PermissionDeniedException;
 import com.cloud.offering.DiskOffering;
 import com.cloud.offering.ServiceOffering;
+
 
 public interface AccountService {
 
@@ -124,6 +127,8 @@ public interface AccountService {
 
     void checkAccess(Account account, DiskOffering dof) throws PermissionDeniedException;
 
+    void checkAccess(User user, ControlledEntity entity);
+
     void checkAccess(Account account, AccessType accessType, boolean sameOwner, String apiName,
             ControlledEntity... entities) throws PermissionDeniedException;
 
@@ -136,4 +141,5 @@ public interface AccountService {
      */
     UserAccount getUserAccountById(Long userId);
 
+    public List<String> getKeys(GetUserKeysCmd cmd);
 }
--- a/api/src/org/apache/cloudstack/api/command/admin/user/GetUserKeysCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/admin/user/GetUserKeysCmd.java
@@ -1,0 +1,74 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package org.apache.cloudstack.api.command.admin.user;
+
+
+import com.cloud.user.Account;
+import com.cloud.user.User;
+import org.apache.cloudstack.api.APICommand;
+import org.apache.cloudstack.api.ApiConstants;
+import org.apache.cloudstack.api.BaseCmd;
+import org.apache.cloudstack.api.Parameter;
+import org.apache.cloudstack.api.response.RegisterResponse;
+import org.apache.cloudstack.api.response.UserResponse;
+
+import java.util.List;
+import java.util.logging.Logger;
+
+@APICommand(name = "getUserKeys",
+            description = "This command allows the user to query the seceret and API keys for the account",
+            responseObject = RegisterResponse.class,
+            requestHasSensitiveInfo = false,
+            responseHasSensitiveInfo = true)
+
+public class GetUserKeysCmd extends BaseCmd{
+
+    @Parameter(name= ApiConstants.ID, type = CommandType.UUID, entityType = UserResponse.class, required = true, description = "ID of the user whose keys are required")
+    private Long id;
+
+    public static final Logger s_logger = Logger.getLogger(RegisterCmd.class.getName());
+    public static final String s_name = "getuserkeysresponse";
+
+    public Long getID(){
+        return id;
+    }
+
+    public String getCommandName(){
+        return s_name;
+    }
+
+    public long getEntityOwnerId(){
+        User user = _entityMgr.findById(User.class, getID());
+        if(user != null){
+            return user.getAccountId();
+        }
+        else return Account.ACCOUNT_ID_SYSTEM;
+    }
+    public void execute(){
+        List<String> keys = _accountService.getKeys(this);
+        RegisterResponse response = new RegisterResponse();
+        if(keys != null){
+            response.setApiKey(keys.get(0));
+            response.setSecretKey(keys.get(1));
+        }
+
+        response.setObjectName("listkeys");
+        response.setResponseName(getCommandName());
+        this.setResponseObject(response);
+    }
+}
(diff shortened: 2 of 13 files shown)


api/src/org/apache/cloudstack/api/command/admin/user/ListKeysCmd.java
```diff
@@ -0,0 +1,72 @@
...
+        User user = _entityMgr.findById(User.class, getID());
+        if(user != null){
+            return user.getAccountId();
+        }
+        else return Account.ACCOUNT_ID_SYSTEM;
+    }
+    public void execute(){
+        String[] keys = _accountService.getKeys(this);
+        RegisterResponse response = new RegisterResponse();
+        if(keys != null){
```
jburwell: Add an check before setting the keys to check that ``keys`` has a length = 2 to avoid an ``ArrayIndexOutOfBoundsException``.  If the length is not equal to 2, throw an ``IllegalStateException``.

api/src/com/cloud/user/AccountService.java
```diff
@@ -136,4 +140,6 @@ void checkAccess(Account account, AccessType accessType, boolean sameOwner, Stri
      */
     UserAccount getUserAccountById(Long userId);
 
+    public String[] getKeys(ListKeysCmd cmd);
```
jburwell: Why is the return type defined as an array and not a ``List<String>``?

kansal:
@jburwell Sure!! Will look into these. Adding the test cases and some UI changes for this to work. Will update the PR.

api/src/com/cloud/user/AccountService.java
DaanHoogland: I agree with @jburwell

DaanHoogland:
@kansal looking forward to your update. your intended change makes sense

kansal:
@jburwell @DaanHoogland @kishankavala  Have included some changes related to the UI. Now after generating the keys from UI, after ListUserCmd() api, listKeysCmd() will be called to fill the secret key as I have removed it from the response value of other API's. 
Also added a test in which a normal user tries to call the listKeysCmd()  for the admin account and hence giving a permission denied exception.

api/src/org/apache/cloudstack/api/command/admin/user/ListKeysCmd.java
```diff
@@ -0,0 +1,74 @@
...
+            responseObject = RegisterResponse.class,
+            requestHasSensitiveInfo = false,
+            responseHasSensitiveInfo = true)
+
+public class ListKeysCmd extends BaseCmd{
+
+    @Parameter(name= ApiConstants.ID, type = CommandType.UUID, entityType = UserResponse.class, required = true, description = "ID of the user whose keys are required")
+    private Long id;
+
+    public static final Logger s_logger = Logger.getLogger(RegisterCmd.class.getName());
```
DaanHoogland: how about using LOGGER as a name for this static final? would be more in line with standards

api/src/org/apache/cloudstack/api/command/admin/user/ListKeysCmd.java
```diff
@@ -0,0 +1,74 @@
...
+    public static final String s_name = "listuserkeysresponse";
```
DaanHoogland: same here

DaanHoogland:
@kansal looks good, but for a change like this I would like a marvin test to prove it and guarantee it's continued functioning/functionality Do you see chance to add that?

kansal:
@DaanHoogland Sure will try. Will take some time as I have to go through the documentation first.

kansal:
Have updated this PR. Instead of directly removing the secret key from response, I have deprecated that as many regressions were using the secret key from those APIs for authentication. Maybe from next major release we can remove that. 

@DaanHoogland  marvin test cases on the way!!!

server/src/com/cloud/user/AccountManager.java
```diff
@@ -198,4 +200,11 @@ void buildACLViewSearchCriteria(SearchCriteria<? extends ControlledViewEntity> s
     public static final String MESSAGE_ADD_ACCOUNT_EVENT = "Message.AddAccount.Event";
 
     public static final String MESSAGE_REMOVE_ACCOUNT_EVENT = "Message.RemoveAccount.Event";
+    public static final ConfigKey<Boolean> UseSecretKeyInResponse = new ConfigKey<Boolean>(
+            "Advanced",
+            Boolean.class,
+            "use.secret.key.in.response",
+            "true",
```
DaanHoogland: default should be false! this is a security issue.

DaanHoogland:
@kansal when you say 'I have deprecated that as many regressions were using the secret key from those APIs for authentication', I think we should adjust those regression test to set the setting to true. Let's not do consessions to security for the sake of testing.

kansal:
@DaanHoogland Agreed with that point. But its not only about the testing. I'm sure many people will be using it in their own integration. I think we should not change the response immediately like this without informing or making a noise about it. And it is because of this concern only, I added a flag for enabling/disabling the secret key in response.

DaanHoogland:
@kansal I don't agree that making noise first is the way to go. We should disable the return of the key first and document it. Security demands that we play it that way. We can allow users to enable this insecure bahaviour by setting a flag somewhere but it should not be default and catch the unaware users of guard. It will be work in the integration tests but that just will have to happen.

server/src/com/cloud/user/AccountManager.java
remibergsma: Agree with @DaanHoogland, it is easy enough to enable again should people need it. Is this setting available to the user or does it need to be added to the database as well?

jburwell:
@DaanHoogland I complete agree with you regarding exposing credential information.  The best practice when credentials are lost is to require that they be changed.  This approach makes the access to the sensitive information obvious to all users -- making it impossible for an attacker to hide such a breach.

In the past, we have removed sensitive data from existing API responses.  For example, for CVE-2015-3251, we removed exposure of KVM credentials from the [listHosts call](https://github.com/apache/cloudstack/pull/682).  Therefore, as a project, we have previously determined that security should trump API backwards compatibility.  It should most certainly be prioritized over making the task of integration testing easier.

kansal:
cc @DaanHoogland @jburwell Okay. Agreed with that. So I am setting the default value to false but for running tests and maybe some other existing integration  we will have to make that value to true. Is that fine? Of course we need to fix the existing test cases so that maybe from the next release we can get away with this thing completely?

server/src/com/cloud/user/AccountManager.java
kansal: @remibergsma It is a part of global config and the admin can change it.

DaanHoogland:
@kansal did you get to this yet?

kansal:
@DaanHoogland I am not sure of the amount of work that needs to be done for fixing all the existing test cases. Will revisit this and update. I still personally think that going with the optional parameter presently and assuming( and making sure) that the new test cases are written in compliance with this API will be a good way to go forward. Your views?  
PS: I am still not very versed with the test case suits. Will check and revisit this.

DaanHoogland:
@kansal as the complexity is unknown I would just go ahead and update the pr. We'll see the damage and think of fixes as we go. As for setting the value to true for existing tests, fine. as long as the default is false. the fix (setting it to true for some test cases) is trivial and will probably not be needed very often. Who is going to check a response for a private key unless they really need it.

rhtyd:
@kansal can you rebase against latest master and share state of your PR, thanks

@DaanHoogland @jburwell do we still have outstanding issues on PR; do we want this or not? thanks

DaanHoogland:
@kansal please go ahead and remove the key from the response. We'll test run it and add fixes to tests if needed. (cc @rhtyd my last comment is still valid)

kansal:
@DaanHoogland  sure. Will rebase and keep the default value to false.

api/src/org/apache/cloudstack/api/command/admin/user/GetUserKeysCmd.java
```diff
@@ -0,0 +1,74 @@
...
+import org.apache.cloudstack.api.response.UserResponse;
+
+import java.util.List;
+import java.util.logging.Logger;
+
+@APICommand(name = "getUserKeys",
+            description = "This command allows the user to query the seceret and API keys for the account",
+            responseObject = RegisterResponse.class,
+            requestHasSensitiveInfo = false,
+            responseHasSensitiveInfo = true)
```
jburwell: Please add the version annotation to indicate that this command was added for 4.9.

server/src/com/cloud/user/AccountManager.java
jburwell: @kansal I agree with @DaanHoogland and @remibergsma -- it's about reasonable and secure defaults.  We should not configure a management server insecurely by default.

server/test/com/cloud/user/MockAccountManagerImpl.java
```diff
@@ -401,5 +403,24 @@ public Long finalyzeAccountId(String accountName, Long domainId, Long projectId,
         return null;
     }
 
+    @Override
+    public List<String> getKeys(GetUserKeysCmd cmd) {
+        return null;
```
jburwell: Please return a ``Collections.emptyList()`` to avoid NPEs.

server/test/com/cloud/user/MockAccountManagerImpl.java
```diff
@@ -401,5 +403,24 @@ public Long finalyzeAccountId(String accountName, Long domainId, Long projectId,
...
+    }
+
+    @Override
+    public void checkAccess(User user, ControlledEntity entity)
+        throws PermissionDeniedException {
+
+    }
+    @Override
+    public String getConfigComponentName() {
+        return null;
```
jburwell: Please return a blank string to avoid NPEs.

server/test/com/cloud/user/MockAccountManagerImpl.java
```diff
@@ -401,5 +403,24 @@ public Long finalyzeAccountId(String accountName, Long domainId, Long projectId,
...
+    }
+
+    @Override
+    public ConfigKey<?>[] getConfigKeys() {
+        return null;
```
jburwell: Please return an empty array to avoid NPEs.

karuturi:
closing in favor of #1996

PR #1153, opened 2015-12-03
CLOUDSTACK-9087:adding projectid parameter to create method of class VmSnapshot

nitt10prashant:
Adding projectid parameter to class VmSnapshot of base.py .

currently create method does not support create vm snpashot under a project . modifying create method  for the same

### DIFF
--- a/tools/marvin/marvin/lib/base.py
+++ b/tools/marvin/marvin/lib/base.py
@@ -4511,7 +4511,7 @@ class VmSnapshot:
 
     @classmethod
     def create(cls, apiclient, vmid, snapshotmemory="false",
-               name=None, description=None):
+               name=None, description=None, projectid=None):
         cmd = createVMSnapshot.createVMSnapshotCmd()
         cmd.virtualmachineid = vmid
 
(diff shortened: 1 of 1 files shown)


pdube:
Isn't the project inferred from the instance? The [CreateVmSnapshot](http://cloudstack.apache.org/docs/api/apidocs-4.6/root_admin/createVMSnapshot.html) API doesn't even take a projectId

nitt10prashant:
@pdube you are right projectid is not required , i did check API doc but got confused with UI and took it as doc bug.
when createVMSnapshot is being called from UI it passes projectid which is ignored by API.
http://*.*.*.*:8080/client/api?command=createVMSnapshot&virtualmachineid=cdf1991b-211c-4d18-a2c8-bbeb3dc23110&snapshotmemory=false&quiescevm=false&name=sd&response=json&projectid=a899bd21-7be6-4113-8cc2-e81771e0c0a1&_=1449223237150

thank , i am closing it

pdube:
@nitt10prashant No problemo

PR #1154, opened 2015-12-03
[UI] fix bug: Cannot delete SSH keypairs in projects

ustcweizhou:
Before change:
error message: ProjectId and account/domainId can't be specified together

After change:
SSH keypairs can be removed


DaanHoogland:
code looks good to me

bhaisaab:
LGTM, merging now.

PR #1158, opened 2015-12-03
CLOUDSTACK-8892:If VR enters out-of band context, routers lose their private/link-local IP

maneesha-p:
If router is powered-on out-of-band , we try to reboot it n times with some retry interval where n and the interval are configurable existing parameters whose default values are 5 and 10 min resp.


remibergsma:
-1

I don't think it is a good idea to reboot routers when out-of-band operation is detected. We discussed this before, let's not do it again. Please search the archives on this subject.

We use out-of-band migrations when patching and this way I can migrate 1000s of routers out-of-band without issues. This is on XenServer and KVM. Your code change seems to make them reboot all, which would be a disaster. It's not even needed.

resmo:
big -1 
We tested out of band migrations on VR on VMware and they work without reboots. 

A VM/VR does not "loose" the IP just because it gets migrated. Look for the root cause.

PR #1159, opened 2015-12-03
CLOUDSTACK-9103 : Missing OS Mappings for VMware 6.0

maneesha-p:
Added suitable db entries in tables hypervisor_capabilities and guest_os_hypervisor to support VMware 6.0 by copying from 5.5 and excluding few depricated guest os.


jburwell:
Master is now 4.7.0.  Therefore, these database changes should be in ``schema-460to470.sql`` and ``schema-461to470.sql`` files.

bhaisaab:
@maneesha-p thanks for the PR, please fix as John suggests. We also support vmware 6.0 on 4.5 branch, so do you think we need this in 4.5.2 to 4.5.3 upgrade path as well?

DaanHoogland:
@maneesha-p It doesn't hurt to put this statement every where but let's reduce the number of paths. as 4.6.0 to 4.6.1 contains a new view, I made a pr to re-route the 4.6.0 to 4.7.0 upgrade via 4.6.1!

maneesha-p:
@jburwell @DaanHoogland Raised a new PR(1216) with changes suggested .
@bhaisaab As master is in 4.7 we dont need the change in every upgrade path before 4.6 as we can pull the changes required.
Please review pr-1216.

PR #1160, opened 2015-12-03
CLOUDSTACK-9101: fix some issues in resize volume

ustcweizhou:
(1) fix issue: volume size is not updated even if the operation succeed
(2) Add ui support for root volume resize
(3) resize on qcow2 type ROOT volume of stopped vm does not really work
see https://issues.apache.org/jira/browse/CLOUDSTACK-9101


ustcweizhou:
wrong branch

PR #1165, opened 2015-12-04
CLOUDSTACK-9104: VM naming convention in case vmware is used

priyankparihar:
I have reverted all the changes. Now functionality is same as it was  in earlier version. 
ACS Link ==>
https://issues.apache.org/jira/browse/CLOUDSTACK-9104

### DIFF
--- a/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -1889,11 +1889,20 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
 
     // Pair<internal CS name, vCenter display name>
     private Pair<String, String> composeVmNames(VirtualMachineTO vmSpec) {
-        String vmInternalCSName = vmSpec.getName();
-        String vmNameOnVcenter = vmSpec.getName();
-        if (_instanceNameFlag && vmSpec.getHostName() != null) {
-            vmNameOnVcenter = vmSpec.getHostName();
+
+        String vmInternalCSName = null;
+        String vmNameOnVcenter  = null;
+        if(vmSpec != null)
+        {
+            vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+            if (_instanceNameFlag == true) {
+                String[] tokens = vmInternalCSName.split("-");
+                assert (tokens.length >= 3); // vmInternalCSName has format i-x-y-<instance.name>
+                vmNameOnVcenter = String.format("%s-%s-%s-%s", tokens[0], tokens[1], tokens[2], vmSpec.getHostName());
+            }
+
         }
+
         return new Pair<String, String>(vmInternalCSName, vmNameOnVcenter);
     }
 


DaanHoogland:
@priyankparihar I do not know any vmware users
@bhaisaab Do you have any insight on this change? or do you know any candidate reviewer for this?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1889,11 +1889,20 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+
+        String vmInternalCSName = null;
+        String vmNameOnVcenter  = null;
+        if(vmSpec != null)
+        {
+            vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+            if (_instanceNameFlag == true) {
+                String[] tokens = vmInternalCSName.split("-");
+                assert (tokens.length >= 3); // vmInternalCSName has format i-x-y-<instance.name>
+                vmNameOnVcenter = String.format("%s-%s-%s-%s", tokens[0], tokens[1], tokens[2], vmSpec.getHostName());
```
bhaisaab: @priyankparihar can you comment on the issues and motivation around this? Do you think we have users who would want to keep hostname as VM names on vCenter side if the global setting is enabled (the _instanceNameFlag is true).

If this is not true, by default the name (as returned by the vmSpec) is like: i-[0-9]*-[0-9]*-VM. (The suffix "VM", is taken from another global setting where the default value is "VM").

bhaisaab:
@DaanHoogland yes I can do that, I'm personally dealt with a lot of vmware/cloudstack naming issues. The syntax/pattern has changes several times in the past between 4.0, 4.2, 4.3 causing VM sync issues when users upgrade.

DaanHoogland:
@bhaisaab @priyankparihar what is the verdict on this change?

priyankparihar:
@bhaisaab 
User with different account cannot create VMs with the same name, which was possible earlier (I am not sure in which CCP version). That time naming convention used was like this “I-<user-id>-<Display-Name>”
Currently if vm.instancename.flag is set to true the VM name will be exactly as display name given.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
bhaisaab: @priyankparihar alright, even if we go with this change. If vmSpec.getHostName() was null then we've a problem with the vm name (if _instanceNameFlag, i.e. the global setting was true).

Once you put a null check, it should be good to go.

bhaisaab:
@priyankparihar Fair enough, please see a code review comment. Once you fix the null check, I think your PR is good to go. Also, write a unit test to enforce the changes; i.e. the unit test should break in future if someone tries to change the VM naming convention in future.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
bhaisaab: Do something like;
String hostString = vmSpec.getHostName();
if (hostString == null) {
  hostString = // get value from the vm suffix global setting (by default "VM");
}

priyankparihar:
@bhaisaab I think null check is not required. I have verified( because in case of null if automatically picks  some random value). But if you still think it may break something then please notify.

bhaisaab:
@priyankparihar I won't be able to give a LGTM without the null check, split string related code and a unit test that breaks if someone tries to change the convention in future

PR #1149, opened 2015-12-02
[4.6/4.6.1] IMPORTANT for 4.6.1 release -- debian: allow rules to pick ACS_BUILD_OPTS from env

bhaisaab:
Only now debian builds can be noredist etc.

cc @remibergsma @wido  - I think the apt-get.eu 4.6.0 deb repo was oss and not noredist one. I also checked the jenkins job it does not "export ACS_BUILD_OPTS='-Dnoredist -Dnonoss'" before running dpkg-buildpackage

I discovered this while testing 4.6.1 RC1 today, nothing major just a build profile issue.

PR #1169, opened 2015-12-04
debian: allow rules to pick ACS_BUILD_OPTS from env

bhaisaab:
Only now debian builds can be noredist etc.

Old PR: https://github.com/apache/cloudstack/pull/1149

cc @remibergsma


remibergsma:
Thanks @bhaisaab I will merge this now, as I count the LGTMs in PR #1149 (in this case running the tests again is not needed).

PR #1181, opened 2015-12-07
CLOUDSTACK-9113: skip vm with inconsistent state when getVmStats/getVmDiskStats

ustcweizhou:
on KVM, if there is a vm has inconsistent state between hypervisor and db, the getVmStat will terminate and return nothing, all vm stats will not be updated.
we should skip the vm which has inconsistent state, and continue on others.


ustcweizhou:
wrong branch

PR #1193, opened 2015-12-09
QUOTA: ENSURING THAT THE DATES DISPLAYED ARE AS PER USER EXPECTATIONS

agneya2001:
When querying db we use start of next day to query quota usage for
    today, but while displaying it to user we still need to show it as
    todays date.

    A regression due to feature review fixes is also made.

https://issues.apache.org/jira/browse/CLOUDSTACK-9126


remibergsma:
@agneya2001 Can you please change the title/commit to not only use capitals?

agneya2001:
Closing this to change the case of the title.

PR #1206, opened 2015-12-10
CLOUDSTACK-9132: API createVolume takes empty string for name parameter

nitin-maharana:
Steps to Reproduce:
================
Create a volume using createVolume API where parameter name is empty.
It creates a volume with empty name.
But the name parameter is mandatory.(Issue)
Expected Behaviour:
================
It shouldn't create a volume with an empty name. Error should be returned.

Solution:
=======
Added a condition to check in case of empty string. If the name is an empty string, it generates a random name for the volume. Made the name field optional in UI as well as in API.

### DIFF
--- a/api/src/org/apache/cloudstack/api/command/user/volume/CreateVolumeCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/volume/CreateVolumeCmd.java
@@ -79,7 +79,7 @@ public class CreateVolumeCmd extends BaseAsyncCreateCustomIdCmd {
                description = "the ID of the disk offering. Either diskOfferingId or snapshotId must be passed in.")
     private Long diskOfferingId;
 
-    @Parameter(name = ApiConstants.NAME, type = CommandType.STRING, required = true, description = "the name of the disk volume")
+    @Parameter(name = ApiConstants.NAME, type = CommandType.STRING, description = "the name of the disk volume")
     private String volumeName;
 
     @Parameter(name = ApiConstants.SIZE, type = CommandType.LONG, description = "Arbitrary volume size")
--- a/server/src/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/com/cloud/storage/VolumeApiServiceImpl.java
@@ -672,7 +672,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
         }
 
         String userSpecifiedName = cmd.getVolumeName();
-        if (userSpecifiedName == null) {
+        if (userSpecifiedName == null || userSpecifiedName.isEmpty()) {
             userSpecifiedName = getRandomVolumeName();
         }
 
--- a/ui/scripts/docs.js
+++ b/ui/scripts/docs.js
@@ -1008,7 +1008,7 @@ cloudStack.docs = {
     },
     // Add volume
     helpVolumeName: {
-        desc: 'Give the volume a unique name so you can find it later.',
+        desc: 'Give the volume a unique name. If not, it will be generated randomly.',
         externalLink: ''
     },
     helpVolumeAvailabilityZone: {
--- a/ui/scripts/storage.js
+++ b/ui/scripts/storage.js
@@ -80,10 +80,7 @@
                                 fields: {
                                     name: {
                                         docID: 'helpVolumeName',
-                                        label: 'label.name',
-                                        validation: {
-                                            required: true
-                                        }
+                                        label: 'label.name'
                                     },
                                     availabilityZone: {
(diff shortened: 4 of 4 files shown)


DaanHoogland:
@nitin-maharana I don't agree we should allow the user to get a default name if (s)he doesn't care. I'd say let the name be the uuid-generated one when empty (as well as when null).

nitin-maharana:
@DaanHoogland : Yes, your idea also looks good. Previously, it was generating a random value only in case of NULL. But if we pass an empty string, it was creating a volume with the empty name. I will modify the condition and update the PR. Thanks.

nitin-maharana:
API with non-empty name:
====================
http://10.102.192.122:8080/client/api?command=createVolume&response=json&name=TempVolume&zoneId=18b1cb1a-bb8b-4e3d-aab8-a48475ed3c60&diskOfferingId=dc8a8d64-34a6-481e-99f9-89a49ee94824&size=1&_=1450077422731

ss_withname_nitin


API with empty name:
=================
http://10.102.192.122:8080/client/api?command=createVolume&response=json&name=&zoneId=18b1cb1a-bb8b-4e3d-aab8-a48475ed3c60&diskOfferingId=dc8a8d64-34a6-481e-99f9-89a49ee94824&size=1&_=1450077422731

ss_withoutname_nitin

nitin-maharana:
@DaanHoogland : In UI, we put the field as required field, so user has to give some name to create a volume. It cannot be null or blank. But according the change, the API doesn't support the logic of UI. Either we should make the field as optional in UI or we should throw an error in API when the name is null or empty.

priyankparihar:
@nitin-maharana Your point looks good. Both(UI and API) should be consistent.

DaanHoogland:
I don't agree, for two reasons
1. making the field required constitutes a backwards incompatibility of the API
2. we should use as much reasonable defaults as posible as to not burdon the user with decision they don't strictly have to make.

I do agree with the present implementation as can be seen in the diff at this moment:
```
-        if (userSpecifiedName == null) {
+        if (userSpecifiedName == null || userSpecifiedName.isEmpty()) {
```
Though the screenshots are not real tests I can live with them for this change: LGTM

rafaelweingartner:
Hi @nitin-maharana I agree with @DaanHoogland, we should maintain backward compatibility. I almost liked you code, I would just suggest you extracting it to a method such as getVolumeName or getCreateVolumeNameIfNeeded (sorry for the lack of creativity, today I am out of inspiration).

After you extracted the code to a method, you should be able to write at least two test cases to make sure it is working as expected.

nitin-maharana:
@DaanHoogland : You mean the name should be an optional field? If that is the case, we should make the field as optional in UI as well as in API.

@rafaelweingartner : The volume name is either collected from the UI or we have a method in API called getRandomVolumeName() which sets a random value. I think we don't need to extract further.
The above method sets a value.

Thanks.

rafaelweingartner:
@nitin-maharana I do not agree with you. Take a look at the size of the method “allocVolume”, there are more than 100 lines, that is terrible to write a test case. We should start creating little chunks of code, test them as unit, and then we can write integration tests.

Something like this would do: 
protected String getVolumeNameFromCommand(CreateVolumeCmd cmd){
        String userSpecifiedName = cmd.getVolumeName();
        if (userSpecifiedName == null || userSpecifiedName.isEmpty()) {
            userSpecifiedName = getRandomVolumeName();
        }
	return userSpecifiedName;
}

After that, you can write test cases to check if the method “getVolumeNameFromCommand” is doing its job as expected. That means, one test to check if it calls “getRandomVolumeName” and returns the value of that call when “cmd.getVolumeName()” returns null and other when it returns empty.

I would also add a third test case to check if the method cmd.getVolumeName() returns something that is not empty, if that value is properly returned.

Having said that, I would also ask, is “            ” (string with a lot of space empty for you?) Because, for me empty is “”, this “         ” is blank, and this way we have another test case to write a make sure everything is proper coded.

For Java, empty is "", which means if cmd.getVolumeName() returns "             ", that is the value that will be used.

nitin-maharana:
@rafaelweingartner : Thanks for the suggestion. I will follow the same to write test cases for this.
@DaanHoogland : Should we make the Name field as optional in UI as well as API?

Thanks.

DaanHoogland:
@nitin-maharana that would be my preference. In the UI a pop-up could explain the consequence of not entering a name.

nitin-maharana:
cc @koushik-das @kishankavala

nitin-maharana:
Made a new PR https://github.com/apache/cloudstack/pull/1273 with master.

PR #1207, opened 2015-12-10
CLOUDSTACK-9133: Two volume.delete usage events are getting generated…

priyankparihar:
It was happening because event triggering  was happening in UserVmManagerImpl.java and VolumeStateListener.java respectively. So i have removed it from UserVmManagerImpl.java.

ACS Link ==>
https://issues.apache.org/jira/browse/CLOUDSTACK-9133

### DIFF
--- a/server/src/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/com/cloud/vm/UserVmManagerImpl.java
@@ -3919,13 +3919,6 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         if (status) {
             // Mark the account's volumes as destroyed
             List<VolumeVO> volumes = _volsDao.findByInstance(vmId);
-            for (VolumeVO volume : volumes) {
-                if (volume.getVolumeType().equals(Volume.Type.ROOT)) {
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@priyankparihar please expand on testing that can prove prior and new behaviour.

priyankparihar:
@DaanHoogland 

Prior  Behaviour
| 127 | VOLUME.DELETE           |          2 | 2015-12-08 09:30:18 |       1 |          25 | ROOT-14                                 |        NULL |        NULL |       NULL | NULL           |         0 |         NULL |
| 128 | VOLUME.DELETE           |          2 | 2015-12-08 09:30:20 |       1 |          25 | ROOT-14                                 |        NULL |        NULL |       NULL | NULL           |         0 |         NULL |

New  Behaviour
| 127 | VOLUME.DELETE           |          2 | 2015-12-08 09:30:18 |       1 |          25 | ROOT-14                                 |        NULL |        NULL |       NULL | NULL           |         0 |         NULL |

DaanHoogland:
@priyankparihar  these are nice results but I have no way of verifying the latter is actually the good result. can you write a test that proves proper behaviour?

borisroman:
@priyankparihar Could you point me where in VolumeStateListener it is emitted?

PR #1211, opened 2015-12-10
[UI] bug fix: Delete added ACL lists is not available for Domain Admin and normal users

ustcweizhou:
Only the owner or domain admin/admin can access the page (vpc->Network ACL Lists), they are also able to remove the network ACL lists, which is not allowed on UI.

### DIFF
--- a/ui/scripts/vpc.js
+++ b/ui/scripts/vpc.js
@@ -638,7 +638,7 @@
                             args.response.success({
                                 _custom: {
                                     jobId: jobID,
-                                    getUpdateIdtem: function() {
+                                    getUpdatedItem: function() {
                                         $(window).trigger('cloudStack.fullRefresh');
                                     }
                                 },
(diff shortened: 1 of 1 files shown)


remibergsma:
LGTM based on the below test.

Original on current master:
- create domain, account and login
- create vpc
- create acl
- no option to delete it

screen shot 2015-12-12 at 10 23 15

The same on this PR:
screen shot 2015-12-12 at 10 17 57

Option to delete is now there:
screen shot 2015-12-12 at 10 18 09

And gone:
screen shot 2015-12-12 at 10 18 29

Thanks for the fix @ustcweizhou !

Pinging @DaanHoogland @bhaisaab @wilderrodrigues to do a second review.

DaanHoogland:
LGTM, thanks for testing @remibergsma. I am going to trust on sight and as this as it is a port from the custom leaseweb code in 4.2.1.

PR #1213, opened 2015-12-10
CLOUDSTACK-9137 Allow domain admin to manage private gw

remibergsma:
To create a private gateway you need a root admin account. This does not make sense, as you can do a lot more with such a powerful account. Other network related API calls can be made by a domain admin.

This PR allows domain admins to create their own private gateways.

### DIFF
--- a/client/tomcatconf/commands.properties.in
+++ b/client/tomcatconf/commands.properties.in
@@ -486,9 +486,9 @@ deleteVPCOffering=1
 listVPCOfferings=15
 
 #### Private gateway commands
-createPrivateGateway=1
+createPrivateGateway=7
 listPrivateGateways=15
-deletePrivateGateway=1
+deletePrivateGateway=7
 
 #### Network ACL commands
 createNetworkACL=15


koushik-das:
@remibergsma Can domain admin1 delete gateway created by domain admin2? If the existing code is properly handling these scenarios then fine.

remibergsma:
It seems this is more complex than it seemed. Since we've to put in a 'physical network id' (which is obviously owned by ROOT domain, this does not work yet.

Creating fails:
```
(admin) 🐵 > create privategateway gateway=1.2.3.4 ipaddress=4.3.2.1 netmask=255.255.255.0 vpcid=cc80ae2c-c3ad-4ea3-96f4-dc40970c81e4 vlan="lswitch://uuid"
Error 531: Acct[74caa349-7f41-4e1a-b4b1-d386c0c2a1a2-rbergsma] does not have permission to operate within domain id=9c2baf29-9846-11e5-9afa-525400b8977a
cserrorcode = 4365
errorcode = 531
errortext = Acct[74caa349-7f41-4e1a-b4b1-d386c0c2a1a2-rbergsma] does not have permission to operate within domain id=9c2baf29-9846-11e5-9afa-525400b8977a
uuidList:
```

This is the mentioned domain id:

```
(admin) 🐵 > set profile root
(root) 🐵 > list domains id=9c2baf29-9846-11e5-9afa-525400b8977a filter=name
count = 1
domain:
name = ROOT
```

Most likely due to the physical network id. Hmm..

PR #1223, opened 2015-12-11
Remove template ulimit from createtmplt.sh and createvolume.sh

syed:
Both createvolume.sh and createtmplt.sh  have a 40GB hardcoded limit for the size of the template that gets created. I could not find any justification of that. I am just removing them as they caused us a huge headache when we tried to create bigger templates and they failed without any good error.

### DIFF
--- a/scripts/storage/secondary/createtmplt.sh
+++ b/scripts/storage/secondary/createtmplt.sh
@@ -27,7 +27,6 @@ usage() {
 
 
 #set -x
-ulimit -f 41943040 #40GiB in blocks
 ulimit -c 0
 
 rollback_if_needed() {
--- a/scripts/storage/secondary/createvolume.sh
+++ b/scripts/storage/secondary/createvolume.sh
@@ -27,7 +27,6 @@ usage() {
 
 
 #set -x
-ulimit -f 41943040 #40GiB in blocks
 ulimit -c 0
 
 rollback_if_needed() {


remibergsma:
Hi @syed can you please squash the commits into one?

syed:
@remibergsma Done!

remibergsma:
@syed Thanks!

@bhaisaab @DaanHoogland Do you guys know why this was put in? It was in the original source import years ago.

bhaisaab:
@remibergsma possible someone added this historically and nobody cared but now. I can check this later and get back to you.

DaanHoogland:
40G sounds like an old fs limit. doesn't make sense nowadays.

syed:
@DaanHoogland I agree, someone must have kept this limit way back and forgot about it.

syed:
Hi Guys ( @bhaisaab, @DaanHoogland, @remibergsma ).Did you get a chance to look at this? I think we can remove this safely.

lttmtins:
Actually the createtmplt.sh script starts with the hardcoded ulimit -f, but there is an '-S" flag within the script which would set the ulimit -f value to the Size+10. 
however the services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java code does not implement the -S flag; see line 314.

lttmtins:
However just removing the ulimit as per suggestion also works for me.

remibergsma:
@syed Thanks for the fix, let's get it it. I'd like to merge it to 4.7 branch. Can you please close this PR, make a new one against 4.7 and then refer the new one to this PR? I'll then verify once more and merge it.

We do this, so the fix appears both in 4.7 and master/4.8. Thanks!

remibergsma:
For the record, LGTM from me. I'll merge the other PR (against 4.7) and close this one.

PR #1236, opened 2015-12-13
CLOUDSTACK-9158: Removed SLF4J, it's abundant due to log4j

borisroman:
In the entire project we use log4j, so why also use slf4j? I've removed slf4j and all references to it from the codebase.

Also I've let the top-level pom depend on log4j so not every project has to define it for itself.

Ping @remibergsma @DaanHoogland @miguelaferreira @wilderrodrigues @wido 

Builds ok, running integration tests now!
```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 6:33.329s
[INFO] Finished at: Sun Dec 13 16:53:55 CET 2015
[INFO] Final Memory: 104M/813M
[INFO] ------------------------------------------------------------------------
```

### DIFF
--- a/agent/src/com/cloud/agent/Agent.java
+++ b/agent/src/com/cloud/agent/Agent.java
@@ -37,7 +37,7 @@ import javax.naming.ConfigurationException;
 
 import org.apache.cloudstack.managed.context.ManagedContextTimerTask;
 import org.apache.log4j.Logger;
-import org.slf4j.MDC;
+import org.apache.log4j.MDC;
 
 import com.cloud.agent.api.AgentControlAnswer;
 import com.cloud.agent.api.AgentControlCommand;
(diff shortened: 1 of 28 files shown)


miguelaferreira:
@borisroman slf4j sits in a layer above log4j. I?m not sure about the particular case you are dealing with, however if we were using slf4j with log4j consistently it would be much easier to now move to logback (the successor of log4j). In thta case the code would be depending on the slf4j classes, and the implementation would be configurable (i.g. log4j, logback, other logging libs).

DaanHoogland:
@miguelaferreira good point. It looks like it has been added for that purpose but never actually used. So now we must choose: start using it or chop it out.

borisroman:
Closing because of preference to logback.

PR #1239, opened 2015-12-13
build_asf.sh: fix debian changelog altering

remibergsma:
The script already adds a new entry so we shouldn't replace the
previous one. That is done only in the next version script.

As discussed with @DaanHoogland


DaanHoogland:
LGTM, let's merge on my review only as we don't have a second RM to review

borisroman:
The changelog file will need to be handled more carefully.

LGTM :+1:

remibergsma:
@borisroman we should separate it from the CloudStack repo ;-)

PR #1241, opened 2015-12-14
CLOUDSTACK-9162: Unable to add VPN user via API with Required Paramet…

kansal:
…ers - Fixed

RCA:
When we do not provide account parameter which is an optional field , API picks up the account from which we are logged in. So this works fine with cloudmonkey and ACS UI. But if we fire API directly from the browser(Make sure no cookies are stored and you are not logged in), by default system account is picked and the VPN users addition fails. This behaviour is expected, but no exception is thrown. So users think that VPN user is created.

Fix: Added an exception when VPN users are added in the system account.

### DIFF
--- a/server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
+++ b/server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
@@ -377,6 +377,10 @@ public class RemoteAccessVpnManagerImpl extends ManagerBase implements RemoteAcc
     public VpnUser addVpnUser(final long vpnOwnerId, final String username, final String password) {
         final Account caller = CallContext.current().getCallingAccount();
 
+        if(caller.getId() == Account.ACCOUNT_ID_SYSTEM){
+            throw new InvalidParameterValueException("Not logged in and no account and domain specified. Can't create VPN users on System Account.");
+        }
+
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@kansal I think this is a good fix but can't be sure. A typical case where an integration test showing the behaviour would help greatly. I'm sure marvin can simulate the not-logged-on behaviour somehow. Actually I think it does by default.

jayapalu:
@kansal Can you please rebase it with the master once. Code changes are fine, I will test once rebased branch.

kansal:
@jayapalu  rebased. Please have a look.

rhtyd:
LGTM

tag:easypr

rhtyd:
tag:mergeready

rhtyd:
LGTM

swill:
@kansal can you close and reopen for do a force push to try to get Jenkins green.  I am not sure what to do about this PR because I don't know that I have a way to verify it works.  @rhtyd and @DaanHoogland suggestions?

karuturi:
@blueorangutan help

karuturi:
@rhtyd @borisstoyanov can you run tests on this pr?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

ustcweizhou:
I am wondering if this is a general issue when create other resources (volume,vm,etc) using integration port 8096.

PR #1242, opened 2015-12-14
test: Fix Libvirt test so that it works on Windows

wido:
This test failed on Windows, using the File.separator it should run fine on Windows.

### DIFF
--- a/plugins/hypervisors/kvm/test/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtUtilitiesHelperTest.java
+++ b/plugins/hypervisors/kvm/test/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtUtilitiesHelperTest.java
@@ -16,6 +16,7 @@
 // under the License.
 package com.cloud.hypervisor.kvm.resource.wrapper;
 
+import java.io.File;
 import java.util.UUID;
 
 import com.cloud.utils.script.Script;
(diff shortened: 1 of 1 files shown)


plugins/hypervisors/kvm/test/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtUtilitiesHelperTest.java
```diff
@@ -37,8 +38,8 @@ public void testSSHKeyPaths() {
          * Hardcoded paths are not what we want in the longer run
          */
         assertEquals("/root/.ssh", helper.retrieveSshKeysPath());
-        assertEquals("/root/.ssh/id_rsa.pub.cloud", helper.retrieveSshPubKeyPath());
-        assertEquals("/root/.ssh/id_rsa.cloud", helper.retrieveSshPrvKeyPath());
+        assertEquals("/root/.ssh" + File.separator + "id_rsa.pub.cloud", helper.retrieveSshPubKeyPath());
```
DaanHoogland: only the last file-separator needs to be a File.separator?

borisroman: The root path isn't available at all in Windows right?

wido: @DaanHoogland True, but the test still works :)

@borisroman Yes, but this is just to make the Unit Tests work under Windows.

wido:
Any LGTMs on this one?

DaanHoogland:
didn't test but I have no further remarks: LGTM

remibergsma:
Pinging @bheuvel to have a look.

bheuvel:
Well, short answer, LGTM and seen it compile on Windows (only) with this adjustment.

Longs answer has more to do with the object/code generating the path `/root/.ssh\id_rsa.pub.cloud`; why mixing path separators like this, how is it made to work on Windows, is the respective code supposed to run on windows at all, why hardcode path to `/root`-home and then some :smile:

PR #1243, opened 2015-12-14
CLOUDSTACK-9166:Build failed in Jenkins: cloudstack-rat-master #7038

SudharmaJain:
With my last PR#1196, I missed to place the license disclaimer into CitrixHelperTest.java. Updated the file with the disclaimer.

### DIFF
--- a/plugins/hypervisors/xenserver/test/com/cloud/hypervisor/xenserver/resource/CitrixHelperTest.java
+++ b/plugins/hypervisors/xenserver/test/com/cloud/hypervisor/xenserver/resource/CitrixHelperTest.java
@@ -1,3 +1,22 @@
+//
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
 package com.cloud.hypervisor.xenserver.resource;
 
 import junit.framework.Assert;


DaanHoogland:
argh, thanks @SudharmaJain LGTM

borisroman:
@DaanHoogland Could you please explain why you merged this?

We have a policy stating **2 (TWO)** persons will have to review a PR. One code review and one running tests. **At all times**.

How trivial it is doesn't matter.

I therefore want it to be reverted.

wilderrodrigues:
+1 for reverting it. And I don't even care if the RAT is broken after the rever. Reasons are obvious, but we have to stick to the 2 LGTM otherwise the project will become the hell it was a few months ago.

I have been to hell... not funny! Do not want to go back.

For a such good reason, as stated in this PR, a simply ping someone would make it get the 2 LGTM and the merge would be fine.

Cheers,
Wilder

rafaelweingartner:
@Wilderrodrigues, @borisroman I agree with you that we cannot merge PRs without two(2) LGTMs and integration tests to check if nothing was broken. @DaanHoogland should have not merged this PR without the second LGTM. However, before he had merged it, I reviewed the PR but have not given LGTM because normally these simple PRs a lot of folks appear to give LGTM, my bad for that. Additionally, I got lost when I checked the right way to include the Apache license, I notice in same classes we use /** * * **/ blocks, in others we use //, while thinking about it something else called my attention and I forgot to give the LGTM.

Before, throwing some stones, I would first check why PR #1196 was merged without the Apache license. I noticed LGTMs there; maybe they were giving without proper review? So, here comes a question, should we revert this one that fixes a build that was broken by #1196, or should we revert the #1196 and fix it before merging it again? Or we can let it go and learn the lesson that should not be repeated.

wilderrodrigues:
@rafaelweingartner 

It's not about throwing stones, it's about the process.

Simply put: 2 LGTM, and depending on the PR at least 1 with integration tests executed. In this case, integration tests were not needed. It is a very simple PR which could have been reviewed very quickly and none of this would have happened. The whole storm behind it is that we have to keep the process and avoid that something like that happens with more complex code - as it happened just after 4.6.0.

Let's save some typing and time and not start a discussion about the obvious.

Cheers,
Wilder

borisroman:
@SudharmaJain Could you please open a new PR so it can be merged with reviews.

DaanHoogland:
Don't go here. this is about licensing. We have a duty do do this with
extreme prejudice.

On Mon, Dec 14, 2015 at 5:00 PM, borisroman <notifications@github.com>
wrote:

> @SudharmaJain <https://github.com/SudharmaJain> Could you please open a
> new PR so it can be merged with reviews.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/apache/cloudstack/pull/1243#issuecomment-164475560>.
>



-- 
Daan

wilderrodrigues:
I truly believe we need a meeting to discuss this. I understand the hole licensing thing, but having a PR open to review is already a prove that the issue is being dealt with.

It's your call... I'm out of this.

PR #1247, opened 2015-12-16
[4.7.0/master] DB Paths fixes: 4.5.3->4.7.0, 4.6.2->4.7.0

bhaisaab:
- Adds missing upgrade path objects from 4.5.3 to 4.7.0
- Adds an explicit upgrade path class from 4.6.2 to 4.7.0, that extends over
  the upgrade path from 4.6.1 to 4.7.0

Notes:
- The getUpgradableVersionRange() is not used to validate from/to versions which is why upgrading from 4.6.2 to 4.7.0 works without any errors when using the Upgrade461to470 for 4.6.2 version.

### DIFF
--- a/engine/schema/src/com/cloud/upgrade/DatabaseUpgradeChecker.java
+++ b/engine/schema/src/com/cloud/upgrade/DatabaseUpgradeChecker.java
@@ -60,6 +60,7 @@ import com.cloud.upgrade.dao.Upgrade452to460;
 import com.cloud.upgrade.dao.Upgrade453to460;
 import com.cloud.upgrade.dao.Upgrade460to461;
 import com.cloud.upgrade.dao.Upgrade461to470;
+import com.cloud.upgrade.dao.Upgrade462to470;
 import com.cloud.upgrade.dao.UpgradeSnapshot217to224;
 import com.cloud.upgrade.dao.UpgradeSnapshot223to224;
 import com.cloud.upgrade.dao.VersionDao;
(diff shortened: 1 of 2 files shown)


bhaisaab:
cc @remibergsma @DaanHoogland

remibergsma:
@bhaisaab Thanks for fixing / improving this. Will have a look and test it. Are you OK with handling this in 4.7.1 (due in say 2 weeks)? By that time 4.5.3 is also there.

bhaisaab:
@remibergsma yes, I've already shared this on dev@ that we need to fix this either in 4.7.0 or 4.7.1 :)
Since, it's not a blocker for 4.7.0 per se we don't need to stop the voting but it would be ideal if we could fix the issue in 4.7.0

engine/schema/src/com/cloud/upgrade/DatabaseUpgradeChecker.java
```diff
@@ -242,13 +243,13 @@ public DatabaseUpgradeChecker() {
...
 
-        _upgradeMap.put("4.5.3", new DbUpgrade[] {new Upgrade453to460()});
+        _upgradeMap.put("4.5.3", new DbUpgrade[] {new Upgrade453to460(), new Upgrade460to461(), new Upgrade461to470()});
 
         _upgradeMap.put("4.6.0", new DbUpgrade[] {new Upgrade460to461(), new Upgrade461to470()});
 
         _upgradeMap.put("4.6.1", new DbUpgrade[] {new Upgrade461to470()});
 
-        _upgradeMap.put("4.6.2", new DbUpgrade[] {new Upgrade461to470()});
+        _upgradeMap.put("4.6.2", new DbUpgrade[] {new Upgrade462to470()});
```
DaanHoogland: there is no reason for this

engine/schema/src/com/cloud/upgrade/dao/Upgrade462to470.java
```diff
@@ -0,0 +1,29 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
DaanHoogland: there is no need for this file

engine/schema/src/com/cloud/upgrade/DatabaseUpgradeChecker.java
bhaisaab: I've explained the reason in detail on dev@. The upgrade path used for 4.6.2 version (source) is Upgrade461to470 that assumes from version 4.6.1 to 4.7.0 in getUpgradableVersionRange(). The fix tries to be consistent with the semantics followed in writing/using upgrade paths. Till upgrade process changes and new tools are introduced, I would like to be consistent with the pattern we've here.

DaanHoogland: Please show what fails, it was tested before release. Let's not fix what isn't broken.

bhaisaab: The PR tries to make the usage be consistent with the implementation and usage; see Upgrade453to460 class for example. The fix here is mainly for the 4.5.3 version, 4.6.2 upgrade path is semantically broken while it still works is another thing.

bhaisaab:
Closing as the 4.5.3 issue has been fixed on 4.7/master. Will open a new PR for the 4.6.2 to 4.7.0 semantic upgrade path/class fix.

PR #1264, opened 2015-12-21
CLOUDSTACK-9192: UpdateVpnCustomerGateway is failing

nitin-maharana:
Reproducible Steps:
================
1.Create a customer gateway for a VPC.
2.Edit it using UI(API call is UpdateVpnCustomerGateway).
3.When we try to update the customer vpn gateway with connection state is not in "Error", we see the API error but that won't be reflected to the user in UI.

Actual Behaviour:
==============
The API throws error. But UI doesn't show it to user.

Expected Behaviour:
=================
The UI should show the error to user.

Fix:
===
TypeError: json.updatecustomergatewayresponse is undefined
The response name was wrong so corrected it.
It should be json.updatevpncustomergatewayresponse.

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -6219,7 +6219,7 @@
                                         url: createURL('updateVpnCustomerGateway'),
                                         data: data,
                                         success: function(json) {
-                                            var jobId = json.updatecustomergatewayresponse.jobid;
+                                            var jobId = json.updatevpncustomergatewayresponse.jobid;
                                             args.response.success({
                                                 _custom: {
                                                     jobId: jobId,
@@ -6231,6 +6231,9 @@
                                                     }
                                                 }
                                             });
+                                        },
+                                        error: function(json) {
+                                            args.response.error(parseXMLHttpResponse(json));
                                         }
                                     });
                                 },


nitin-maharana:
cc @JayapalUradi

ustcweizhou:
@nitin-maharana 
I am not sure if it is necessary to add the following change after success, but I added it in my fix, basides your change.

```
                                        error: function(json) {
                                            args.response.error(parseXMLHttpResponse(json));
                                         }
```

nitin-maharana:
Hi @ustcweizhou, The error function is not necessary but it is good to write this in call. Here the request doesn't fail but throws an error. Thanks for the comment, I will update.

nitin-maharana:
cc @kishankavala @bhaisaab @remibergsma

ustcweizhou:
LGTM.
This is same to the fix in our internal branch.

remibergsma:
@nitin-maharana Thanks for the fix! I'd suggest making this PR against 4.7 (it will end up on master as well).

nitin-maharana:
Sure @remibergsma. I will a PR against 4.7. Thanks.

nitin-maharana:
Closing this PR as made a new PR #1300 (Against 4.7 which will be merged in master later).

PR #1274, opened 2015-12-22
CLOUDSTACK-9196: Fixing null pointer exception when vm meta data is synced on upgraded setup

anshul1886:
https://issues.apache.org/jira/browse/CLOUDSTACK-9196

NullPointerException can occur if XenServer reports non-existing VM in cloud DB.

### DIFF
--- a/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -2641,7 +2641,7 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
 
             if(!found) {
                 VMInstanceVO vm = _vmDao.findVMByInstanceName(name);
-                if(vm.getType() == VirtualMachine.Type.User) {
+                if(vm != null && vm.getType() == VirtualMachine.Type.User) {
                     updateVmMetaData(vm.getId(), platform);
                 }
             }


DaanHoogland:
LGTM but please describe (preferable in a unit test) what situation is prevented from happening here.

anshul1886:
@DaanHoogland Updated the descrition

koushik-das:
Code changes LGTM

pdube:
Shouldn't we log this information? Doesn't this mean that some resources could be freed up on the hypervisor?

anshul1886:
@pdube That part is handled elsewhere. Here we are concerned about syncing of VM metadata. So here if the VM is not in cloud DB then we are ignoring it.

pdube:
Null check LGTM

bhaisaab:
LGTM, merging based on comments

runseb:
@bhaisaab @pdube @DaanHoogland @koushik-das lots of LGTM here , so great. but we don't know why you LGTM'd and as far as I know no tests were run...

bhaisaab:
@runseb checked the diff, travis/jenkins passed, diff looked good to me.
I've merged this and few others without integration tests as they *don't* explictly need them; see the diff -- do you think an explicit integration test is necessary for this simple NPE check?

IMO, some changes cannot be tested by integration tests - if they are text related, in non-Java files such as READMEs, reset scripts (not even used by systemvm/systemvm.iso). Then, there are some changes such as this where we can cheat and avoid running integration tests given you know the code, and build/travis passes.

runseb:
@bhaisaab actually I am not even seeing the travis runs here anymore.

Reason I am asking is genuine (again), because in the last few months Travis has shown to be flaky and was Green when actually tests did not run ...

PR #1283, opened 2015-12-23
CLOUDSTACK-4572: findHostsForMigration API does not return correct host list

anshul1886:
https://issues.apache.org/jira/browse/CLOUDSTACK-4572

Fix is missing in recent releases.

The api should return the list of suitable/unsuitable hosts
Added fix that creates a deep copy of the the variable allHosts and prevents faulty host list return.

### DIFF
--- a/server/src/com/cloud/server/ManagementServerImpl.java
+++ b/server/src/com/cloud/server/ManagementServerImpl.java
@@ -1225,7 +1225,10 @@ public class ManagementServerImpl extends ManagerBase implements ManagementServe
             plan = new DataCenterDeployment(srcHost.getDataCenterId(), srcHost.getPodId(), srcHost.getClusterId(), null, null, null);
         }
 
-        final Pair<List<? extends Host>, Integer> otherHosts = new Pair<List<? extends Host>, Integer>(allHosts, new Integer(allHosts.size()));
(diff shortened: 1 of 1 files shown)


koushik-das:
@anshul1886 If the fix was added in 4.2, why it got missed out? Can you check the history and update?

anshul1886:
@koushik-das It was merged to 4.2-forward  but was not merged to master.

DaanHoogland:
@anshul1886 (ignoring my usual remarks in testing) isn't this for 4.6 and to be merged forward then? It must be missing in all releases in between, right?

anshul1886:
Yes, It is missing in all releases. It can be backported to 4.6.

DaanHoogland:
I would like it to be a PR to 4.6 and be forward merged to 4.7 and master, instead of being applied on master.
ping @remibergsma @wilderrodrigues @borisroman @ustcweizhou

anshul1886:
Can we have review on this PR and when everybody is satisfied then create PR for 4.6?

DaanHoogland:
yes, and that would mean that we need to run the regression suite against it twice. I am satisfied with it for now, but want to see testing. better to run that against the version we will merge it in.

anshul1886:
Closing this PR,  Created new PR against 4.6 https://github.com/apache/cloudstack/pull/1286.

PR #1284, opened 2015-12-23
Map ldap group to Cloudstack account

miguelaferreira:
The LDAP plugin authenticates Cloudstack users against users in an LDAP group, and when doing so creates a Cloudstack account with the same name as the LDAP user, and a Cloudstack user also with the same name as the LDAP user.
This makes it difficult to manage the users in a given LDAP group because each user will have it's own account and user in Cloudstack, and cannot therefore collaborate in managing cloud resources.

This PR changes the LDAP plugin to use the LDAP group for the Cloudstack account name, and create the users (belonging to the LDAP group) under the same Cloudstack account.

### DIFF
--- a/plugins/user-authenticators/ldap/pom.xml
+++ b/plugins/user-authenticators/ldap/pom.xml
@@ -71,6 +71,7 @@
         <configuration>
           <includes>
             <include>**/*Spec*</include>
+            <include>**/*Test.java</include>
           </includes>
         </configuration>
       </plugin>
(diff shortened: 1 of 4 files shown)


miguelaferreira:
I've tested this PR manually in our beta environment, however ince there is a marvin test for this functionality I would also like to run that. I haven't yet because I haven't found a way to feed the required test data into the test run.

miguelaferreira:
Closing as this should have been directed at 4.7 branch.

PR #1286, opened 2015-12-24
CLOUDSTACK-4572: findHostsForMigration API does not return correct host list

anshul1886:
https://issues.apache.org/jira/browse/CLOUDSTACK-4572

Fix is missing in recent releases.

The api should return the list of suitable/unsuitable hosts
Added fix that creates a deep copy of the the variable allHosts and prevents faulty host list return.


wido:
What is the best way to test this?

anshul1886:
It seems in latest releases bug is already fixed with commit for bug https://issues.apache.org/jira/browse/CLOUDSTACK-8484 so closing PR.

PR #1288, opened 2015-12-27
trailing commas in javascripts removed

DaanHoogland:
According to SonarQube this is a bug on internet explorer. It is the only 'blocker' level issue in cloudstack. @abhinandanprateek (@agneya2001) @bhaisaab @miguelaferreira should we enforce this or, as alternative, have it disabled in SonarQube?

### DIFF
--- a/ui/plugins/plugins.js
+++ b/ui/plugins/plugins.js
@@ -16,7 +16,7 @@
 // under the License.
 (function($, cloudStack) {
   cloudStack.plugins = [
-    'quota',
-    //'testPlugin'
+    //'testPlugin',
+    'quota'
   ];
 }(jQuery, cloudStack));
--- a/ui/plugins/quota/quota.js
+++ b/ui/plugins/quota/quota.js
@@ -79,9 +79,9 @@
                               indicator: {
                                   'enabled': 'on',
                                   'disabled': 'off',
-                                  'locked': 'off',
+                                  'locked': 'off'
                               }
-                          },
+                          }
                       },
                       dataProvider: function(args) {
                           var data = {
@@ -171,7 +171,7 @@
                                             label: 'label.quota.enforcequota',
                                             isBoolean: true,
                                             isChecked: false
-                                        },
+                                        }
                                     }
 
                                 },
@@ -196,7 +196,7 @@
                                     });
                                     $(window).trigger('cloudStack.fullRefresh');
                                  }
-                            },
+                            }
                           },
                           tabs: {
                              details: {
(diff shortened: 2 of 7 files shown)


bhaisaab:
@DaanHoogland IMO, we don't have a javascript checkstyle or linter; and such things are non-issues, I would say either have a checkstyle like build step (like a linter) for JS code included in our maven build or disable in SonarQube.

DaanHoogland:
@bhaisaab did you check the internet explorer claim that sonarqube makes?

bhaisaab:
@DaanHoogland no

ui/plugins/plugins.js
```diff
@@ -16,7 +16,7 @@
 // under the License.
 (function($, cloudStack) {
   cloudStack.plugins = [
-    'quota',
-    //'testPlugin'
+    //'testPlugin',
```
borisroman: Why not remove it? It's already commented out...

bhaisaab: also, why not remote the testPlugin folder and the test plugin js etc. implementation?

DaanHoogland: fine but out of scope

DaanHoogland:
@bhaisaab we do not have anything better the sonarqube atm. so let's listen to it.

bhaisaab:
sure

remibergsma:
LGTM, UI seems to display OK.

screen shot 2016-01-18 at 15 00 56

bhaisaab:
LGTM, merging based on comments and tests

runseb:
Folks, I did not have time to fully test the new merging process from @DaanHoogland and @remibergsma but let's assume this is good to merge should this really be merged in master first. What about the other release branches @bhaisaab ?

bhaisaab:
@runseb this was sent for master as the author intended, non-issue actually but would satisfy SonarQube from showing up redflags

runseb:
Hi @bhaisaab I am genuinely asking because of the new release principles:
https://cwiki.apache.org/confluence/display/CLOUDSTACK/Release+principles+for+Apache+CloudStack+4.6+and+up

where we agreed to apply to release branches and merge forward.

Now this is not a bug per se, so could go directly to master, but I think it could be applied as a "fix" and merged forward

remibergsma:
@runseb We could see it as a bug, but I agree with @bhaisaab that it is a non-issue that was "fixed" to make SonarQube happy. Fine to do that straight on master IMHO.

DaanHoogland:
@runseb @bhaisaab I think this is worth a small but intensive discussion, preferably over a beer. What we are talking about is neither a bug nor a feature. We never mentioned this kind of improvement in are procedures. My instict goes with the way @bhaisaab handled this but that is worth less then €0,02.

runseb:
@bhaisaab @remibergsma @DaanHoogland ok then, just checking.

PR #1292, opened 2015-12-28
Add Health Check Command to NSX plugin

miguelaferreira:
The NSX plugin does not support the HeathCheckCommand. Instead it fakes a PingCommand as a call tot he control cluster status API.
However, we have seen in production that the management server will sometimes find the NSX controller to be behind on ping and that will trigger a HealthCheckCommand which will return with an unsupported command answer.
Once this happens the controller is put into Alert state and will not recover until the management sever is restarted.

In addition, during the investigation, there will be a null pointer exception due tot he fact that the NSX controllers do not live in a pod.

This PR tries to address those two issues.


miguelaferreira:
closing since this should not target master

PR #1293, opened 2015-12-28
Add Health Check Command to NSX plugin

miguelaferreira:
The NSX plugin does not support the HeathCheckCommand. Instead it fakes a PingCommand as a call tot he control cluster status API.
However, we have seen in production that the management server will sometimes find the NSX controller to be behind on ping and that will trigger a HealthCheckCommand which will return with an unsupported command answer.
Once this happens the controller is put into Alert state and will not recover until the management sever is restarted.

In addition, during the investigation, there will be a null pointer exception due tot he fact that the NSX controllers do not live in a pod.

This PR tries to address those two issues.

### DIFF
--- a/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
@@ -843,7 +843,7 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
                 }
 
                 final Status currentStatus = host.getStatus();
-                s_logger.info("The agent " + hostId + " state determined is " + determinedState);
+                s_logger.info("The agent from host " + hostId + " state determined is " + determinedState);
 
                 if (determinedState == Status.Down) {
                     final String message = "Host is down: " + host.getId() + "-" + host.getName() + ". Starting HA on the VMs";
--- a/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
@@ -1536,7 +1536,7 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
                             /*
                              * Host is in non-operation state, so no investigation and direct put agent to Disconnected
                              */
-                            status_logger.debug("Ping timeout but host " + agentId + " is in resource state of " + resourceState + ", so no investigation");
+                            status_logger.debug("Ping timeout but agent " + agentId + " is in resource state of " + resourceState + ", so no investigation");
                             disconnectWithoutInvestigation(agentId, Event.ShutdownRequested);
                         } else {
                             final HostVO host = _hostDao.findById(agentId);
@@ -1546,7 +1546,7 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
                                 s_logger.warn("Disconnect agent for CPVM/SSVM due to physical connection close. host: " + host.getId());
                                 disconnectWithoutInvestigation(agentId, Event.ShutdownRequested);
                             } else {
-                                status_logger.debug("Ping timeout for host " + agentId + ", do invstigation");
+                                status_logger.debug("Ping timeout for agent " + agentId + ", do invstigation");
                                 disconnectWithInvestigation(agentId, Event.PingTimeout);
                             }
                         }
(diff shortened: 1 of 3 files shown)


miguelaferreira:
I've built the entire project running the existing and additional unit tests. I've also deployed this jar into our beta environment to check that it wound't break it. I don't really know how to test this further, since I can't trigger ACS to consider that the NSX controller is behind on ping (because the ping map is held in memory, not in the DB).

I will run this change in our beta environment (which has production like load) for a few days and will monitor the logs.

plugins/network-elements/nicira-nvp/src/main/java/com/cloud/network/resource/wrapper/NiciraCheckHealthCommandWrapper.java
```diff
@@ -0,0 +1,63 @@
...
+import com.cloud.agent.api.Answer;
+import com.cloud.agent.api.CheckHealthAnswer;
+import com.cloud.agent.api.CheckHealthCommand;
+import com.cloud.network.nicira.ControlClusterStatus;
+import com.cloud.network.nicira.NiciraNvpApi;
+import com.cloud.network.nicira.NiciraNvpApiException;
+import com.cloud.network.resource.NiciraNvpResource;
+import com.cloud.resource.CommandWrapper;
+import com.cloud.resource.ResourceWrapper;
+
```
GabrielBrascher: @miguelaferreira I will appreciate if you add Javadoc for this class (NiciraCheckHealthCommandWrapper) and its methods. After all, there is no better time to document a code than at its creation ;)

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -821,9 +819,9 @@ protected boolean handleDisconnectWithInvestigation(final AgentAttache attache,
             try {
                 nextStatus = host.getStatus().getNextStatus(event);
             } catch (final NoTransitionException ne) {
-                /* Agent may be currently in status of Down, Alert, Removed, namely there is no next status for some events.
-                 * Why this can happen? Ask God not me. I hate there was no piece of comment for code handling race condition.
-                 * God knew what race condition the code dealt with!
+                /*
```
GabrielBrascher: @miguelaferreira Could you please remove this comment (or, at least, this part : "Why this can happen? Ask God not me..." )? I don't think that this comment is necessary. Thanks

rafaelweingartner: @miguelaferreira  If you extract the code from lines 819-827 to a method, you could add that comment block to it as a Java doc, a method such as “getHostNextStatus” would do. I would only suggest you to remove the sarcastic bit of it.

plugins/network-elements/nicira-nvp/src/main/java/com/cloud/network/resource/wrapper/NiciraCheckHealthCommandWrapper.java
miguelaferreira: @GabrielBrascher thank you for taking the time to look at this. Unfortunately, or not, I have to disagree with you. You know I believe there is no better documentation than maintainable code. I would be very happy to hear from you on how I can make my code more maintainable. Thank you

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
miguelaferreira: @GabrielBrascher and @rafaelweingartner Although I appreciate your remarks (especially the one about refactoring a method out of a bigger method), I'm not sure if you realise I've just formatted this class, and fixed a few logging messages. Therefore I won't really be maintaining the rest of the functionality. I do encourage you two to make a PR with your improvements in it. This is after all open-source and cooperation based.

rafaelweingartner: I understand your point. However, you have not just formatted the class you added some extra comments.

plugins/network-elements/nicira-nvp/src/main/java/com/cloud/network/resource/wrapper/NiciraCheckHealthCommandWrapper.java
GabrielBrascher: @miguelaferreira When I am debugging, extending or just reading some code it is always good to find some documentation. I think that if you add a Javadoc block explaining the code and its purpose, it will help on our maintenance process.

For example, the “NiciraCheckHealthCommandWrapper” name doesn't tell much for me about the code in a glance. What is a CheckHealth? What is this Nicira component? 

The idea of a documentation is to help people that never saw a class or method to understand what it does, why it is needed, what can be added or not to its context, exceptions that it may throw and other useful information. Thus, someone can understand the main idea of a class or method without diving deep inside of the code.

I believe that the use of Javadoc can make the code friendlier, specially to new developers.

remibergsma:
Tested it and works well in our environment. LGTM.

borisroman:
LGTM :+1: 

Been running in beta for a couple of weeks now, no problems!

PR #1294, opened 2015-12-28
Implement a NSX API request execution counter per thread

miguelaferreira:
The NSX plugin has a execution counter to prevent infinite recursion (and as a result a stack overflow exception). However, the thread safeness of this counter are not as desired. The counter was implemented with an AtomicInteger which make it safe for multiple threads to update and read it. The desired property would be to have a counter per thread.

This PR addresses that issue.

### DIFF
--- a/plugins/network-elements/nicira-nvp/src/main/java/com/cloud/network/nicira/ExecutionCounter.java
+++ b/plugins/network-elements/nicira-nvp/src/main/java/com/cloud/network/nicira/ExecutionCounter.java
@@ -19,15 +19,19 @@
 
 package com.cloud.network.nicira;
 
-import java.util.concurrent.atomic.AtomicInteger;
-
 public class ExecutionCounter {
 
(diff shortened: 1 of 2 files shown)


remibergsma:
Haven't seen the issue in our production cloud since we run with this fix. Thanks! LGTM.

remibergsma:
@borisroman said he had an improvement to the unit test. Let's discuss and add it to the PR so we can merge it.

DaanHoogland:
makes sense @miguelaferreira . I have no use to this but I think it needs to go in
LGTM @remibergsma

miguelaferreira:
@remibergsma @DaanHoogland I've fixed a concurency issue in a unit tests (pointed out by @borisroman) and in the process rebased this branch on latest 4.7.

borisroman:
LGTM :+1:

Did code review and ran the PR in beta

PR #1302, opened 2016-01-04
CLOUDSTACK-9104: VM naming convention in case vmware is used

priyankparihar:
I have reverted all the changes. Now functionality is same as it was in earlier version. 
ACS  Bug Link is as follow ==>
https://issues.apache.org/jira/browse/CLOUDSTACK-9104

### DIFF
--- a/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -43,6 +43,7 @@ import java.util.UUID;
 
 import javax.naming.ConfigurationException;
 
+import com.cloud.exception.InvalidParameterValueException;
 import org.apache.log4j.Logger;
 import org.apache.log4j.NDC;
 
@@ -2226,13 +2227,30 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
         return new String[] {datastoreDiskPath};
     }
 
-    // Pair<internal CS name, vCenter display name>
-    private Pair<String, String> composeVmNames(VirtualMachineTO vmSpec) {
-        String vmInternalCSName = vmSpec.getName();
-        String vmNameOnVcenter = vmSpec.getName();
-        if (_instanceNameFlag && vmSpec.getHostName() != null) {
-            vmNameOnVcenter = vmSpec.getHostName();
+    /**
+     * This method generates VM name for Vcenter and Cloudstack( when Hypervisor is VMware).
+     * It generates VM name according to  _instanceNameFlag setting.
+     *
+     * @param VirtualMachineTO
+     *        vmSpec
+     * @return Pair<internal CS name, vCenter display name>. A pair which contain 'internal CS name'  and
+     *         'vCenter display name'(vCenter knows VM by this name).
+     **/
+    protected Pair<String, String> composeVmNames(VirtualMachineTO vmSpec) throws CloudRuntimeException {
+
+        String vmInternalCSName = "";
+        String vmNameOnVcenter  = "";
+        if(vmSpec == null) {
+            throw new CloudRuntimeException("vmSpec is null");
         }
+
+        vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+        if (_instanceNameFlag && vmSpec.getType() == VirtualMachine.Type.User) {
+           String[] tokens = vmInternalCSName.split("-");
+           if(tokens.length<3) throw new InvalidParameterValueException("Invalid vmInternalCSName format: "+vmInternalCSName+" expected format : *-*-* (e.g -> i-x-y)"); // vmInternalCSName has format i-x-y-<instance.name>
+              vmNameOnVcenter = String.format("%s-%s-%s-%s", tokens[0], tokens[1], tokens[2], vmSpec.getHostName());
+        }
+
         return new Pair<String, String>(vmInternalCSName, vmNameOnVcenter);
     }
 
--- a/plugins/hypervisors/vmware/test/com/cloud/hypervisor/vmware/resource/VmwareResourceTest.java
+++ b/plugins/hypervisors/vmware/test/com/cloud/hypervisor/vmware/resource/VmwareResourceTest.java
@@ -16,6 +16,8 @@
 // under the License.
 package com.cloud.hypervisor.vmware.resource;
 
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
@@ -35,8 +37,12 @@ import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.Map;
 
+import com.cloud.exception.InvalidParameterValueException;
+import com.cloud.utils.Pair;
+import com.cloud.vm.VirtualMachine;
 import org.apache.cloudstack.storage.command.CopyCommand;
 import org.apache.cloudstack.storage.to.PrimaryDataStoreTO;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -404,4 +410,46 @@ public class VmwareResourceTest {
         verify(vmMo, never()).getRunningHost();
     }
 
+    @Test
+    public  void  testComposeVmNames() throws Exception{
+
+        Pair<String, String> op= null;
+
+        //case 1: vmSpec == null
+        try {
+            _resource.composeVmNames(null);
+            Assert.fail("Function should throw CloudRuntimeException "+ "vmSpec is null");
+        }catch(Exception e) {
+            assertThat(e, instanceOf(CloudRuntimeException.class));
+        }
+
+        //case 2: vmSpec != null
+        //case 2.1:  _instanceNameFlag == false
+        when(vmSpec.getName()).thenReturn("vmname");
+        _resource._instanceNameFlag = false;
+        op = _resource.composeVmNames(vmSpec);
+        assert(op.first().compareTo("vmname") == 0 && op.second().compareTo("vmname") == 0);
+
+        //case 2.2.1:  _instanceNameFlag == true  && vmSpec.getType() == VirtualMachine.Type.User
+        when(vmSpec.getName()).thenReturn("vmname-a-1");
+        when(vmSpec.getHostName()).thenReturn("host");
+        when(vmSpec.getType()).thenReturn(VirtualMachine.Type.User);
+        _resource._instanceNameFlag = true;
+        op = _resource.composeVmNames(vmSpec);
+        assert(op.first().compareTo("vmname-a-1") == 0 && op.second().compareTo("vmname-a-1-host") == 0);
+
+        //case 2.2.2:  _instanceNameFlag == true  && vmSpec.getType() == VirtualMachine.Type.User && tokens.length<3
+        when(vmSpec.getName()).thenReturn("vmname+a-1");
+        try {
+            _resource.composeVmNames(vmSpec);
+            Assert.fail("Function should throw Invalid Parameter Value Exception");
+        }catch(Exception e) {
+            assertThat(e, instanceOf(InvalidParameterValueException.class));
+        }
+
+        //case 2.2.3:  _instanceNameFlag == true  && vmSpec.getType() != VirtualMachine.Type.User
+        when(vmSpec.getType()).thenReturn(VirtualMachine.Type.ConsoleProxy);
+        op = _resource.composeVmNames(vmSpec);
+        assert(op.first().compareTo("vmname+a-1") == 0 && op.second().compareTo("vmname+a-1") == 0);
+    }
 }
--- a/server/src/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/com/cloud/vm/UserVmManagerImpl.java
@@ -3383,12 +3383,6 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                     hostName = generateHostName(uuidName);
                 }
             }
-            // If global config vm.instancename.flag is set to true, then CS will set guest VM's name as it appears on the hypervisor, to its hostname.
-            // In case of VMware since VM name must be unique within a DC, check if VM with the same hostname already exists in the zone.
-            VMInstanceVO vmByHostName = _vmInstanceDao.findVMByHostNameInZone(hostName, zone.getId());
-            if (vmByHostName != null && vmByHostName.getState() != VirtualMachine.State.Expunging) {
-                 throw new InvalidParameterValueException("There already exists a VM by the name: " + hostName + ".");
-            }
         } else {
             if (hostName == null) {
                 //Generate name using uuid and instance.name global config


plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2032,11 +2032,20 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
     // Pair<internal CS name, vCenter display name>
     private Pair<String, String> composeVmNames(VirtualMachineTO vmSpec) {
-        String vmInternalCSName = vmSpec.getName();
-        String vmNameOnVcenter = vmSpec.getName();
-        if (_instanceNameFlag && vmSpec.getHostName() != null) {
-            vmNameOnVcenter = vmSpec.getHostName();
+
+        String vmInternalCSName = null;
+        String vmNameOnVcenter  = null;
+        if(vmSpec != null)
```
rafaelweingartner: Is there a case in which "vmSpec" can be null here?

That is nice that you are trying to avoid a null pointer exception, but the way it is, you just change the null pointer to somewhere else

priyankparihar: @rafaelweingartner Currently. it is not possible. Because user does not generate it. Cloud stack generates it only if it has all required info. otherwise it does not call this function.

rafaelweingartner: If today there is no way a null value is received in that method, I see no reason to have that check in 2038.

alexandrelimassantana:
@priyankparihar, since you are modifying this method, you could also generate it's javadoc. It would be a nice upgrade from that single line comment placed just before the method signature ( line 2033 ). You could even state your parameter null check in the @parameter field in it.

priyankparihar:
Hi, 
@alexandrelimassantana thanks for notifying. I have made modification,  according to your suggestion.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2030,12 +2030,29 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
 
-    // Pair<internal CS name, vCenter display name>
-    private Pair<String, String> composeVmNames(VirtualMachineTO vmSpec) {
-        String vmInternalCSName = vmSpec.getName();
-        String vmNameOnVcenter = vmSpec.getName();
-        if (_instanceNameFlag && vmSpec.getHostName() != null) {
-            vmNameOnVcenter = vmSpec.getHostName();
+
+    /**
+     * This method gemerate VM name for Vcenter and Cloudstack( when Hypervisor is VMware).
```
alexandrelimassantana: Just a little typo in line 2035, "gemerate".

Thanks for taking the time to write the javadoc.

priyankparihar: Hi  @alexandrelimassantana,
Now, i think there is no typo.

--
Thanks for your careful observation.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2030,12 +2030,29 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+    private Pair<String, String> composeVmNatmes(VirtualMachineTO vmSpec) {
+
+        String vmInternalCSName = null;
+        String vmNameOnVcenter  = null;
+        if(vmSpec != null)
+        {
+            vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+            if (_instanceNameFlag == true && vmSpec.getType()==VirtualMachine.Type.User) {
+                String[] tokens = vmInternalCSName.split("-");
+                assert (tokens.length >= 3); // vmInternalCSName has format i-x-y-<instance.name>
```
sateesh-chodapuneedi: Because assertions may be disabled, programs must not assume that the boolean expression contained in an assertion will be evaluated. Try to check it as well.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2030,12 +2030,29 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+     * It generates VM name according to  _instanceNameFlag setting.
+     *
+     * @param VirtualMachineTO
+     *        vmSpec
+     * @return Pair<internal CS name, vCenter display name>. A pair which contain 'internal CS name'  and
+     *         'vCenter display name'(vCenter knows VM by this name).
+     **/
+    private Pair<String, String> composeVmNatmes(VirtualMachineTO vmSpec) {
+
+        String vmInternalCSName = null;
```
sateesh-chodapuneedi: Try to use emtpy/blank rather than null here, because these are being returned. Good practice to avoid returning nulls.
Of course, empty vm name is obviously invalid both in ACS database as well as hypervisor (VMware here, this code is not being invoked for other hypervisor any way)

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2030,12 +2030,29 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+        String vmNameOnVcenter  = null;
+        if(vmSpec != null)
+        {
+            vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+            if (_instanceNameFlag == true && vmSpec.getType()==VirtualMachine.Type.User) {
```
sateesh-chodapuneedi: add a whitespace before and after '=='

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2030,12 +2030,29 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+    /**
+     * This method generates VM name for Vcenter and Cloudstack( when Hypervisor is VMware).
+     * It generates VM name according to  _instanceNameFlag setting.
+     *
+     * @param VirtualMachineTO
+     *        vmSpec
+     * @return Pair<internal CS name, vCenter display name>. A pair which contain 'internal CS name'  and
+     *         'vCenter display name'(vCenter knows VM by this name).
+     **/
+    private Pair<String, String> composeVmNatmes(VirtualMachineTO vmSpec) {
```
sateesh-chodapuneedi: @priyankparihar 
Can you please add a unit test for composeVmNatmes?

sateesh-chodapuneedi: @priyankparihar 
Can you rename composeVmNatmes to composeVmNames?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
jburwell: Guava's ``Preconditions.checkState`` is terse way to implement such checks.

bhaisaab:
@priyankparihar we'll need a unit test that shows that this change is backward compatible. I like that we're considering both default instance name with host name here (i.e. the names should be globally unique) but we'll need few more units tests to make sure the contract is not broken again in future.

We've changed how internal/external vmware vm names are composed 3-4 times since 4.0, can you explain why we want to do this again.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2030,12 +2030,29 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+
+        String vmInternalCSName = null;
+        String vmNameOnVcenter  = null;
+        if(vmSpec != null)
+        {
+            vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+            if (_instanceNameFlag == true && vmSpec.getType()==VirtualMachine.Type.User) {
+                String[] tokens = vmInternalCSName.split("-");
+                assert (tokens.length >= 3); // vmInternalCSName has format i-x-y-<instance.name>
+                vmNameOnVcenter = String.format("%s-%s-%s-%s", tokens[0], tokens[1], tokens[2], vmSpec.getHostName());
```
bhaisaab: what if one of the tokens are null/empty? why not simply add the hostname (if only it's not null/empty) at the end of vmInternalCSName; simply do a `vmInternalCSName + "-" + vmSpace.getHostName()` (iff !Strings.isNullOrEmpty(getHostName()))

Using the same internal CS name on vmware should be alright and ideal, solve the issue of inconsistent and unique naming.

priyankparihar:
Hi @rhtyd and @sateesh-chodapuneedi,
I have made modification, according to your suggestion. Please take a look.

priyankparihar:
Hi @borisstoyanov Please check.

karuturi:
@blueorangutan @rhtyd @borisroman vmware tests please

rhtyd:
@blueorangutan package

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2226,17 +2227,33 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+     * This method generates VM name for Vcenter and Cloudstack( when Hypervisor is VMware).
+     * It generates VM name according to  _instanceNameFlag setting.
+     *
+     * @param VirtualMachineTO
+     *        vmSpec
+     * @return Pair<internal CS name, vCenter display name>. A pair which contain 'internal CS name'  and
+     *         'vCenter display name'(vCenter knows VM by this name).
+     **/
+    protected Pair<String, String> composeVmNames(VirtualMachineTO vmSpec) {
+
```
rhtyd: - Why is the method attribute changed from private to protected?
- If vmspec is null, this method should throw an exception, to avoid a situation where empty vmnames are returned. Discuss?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2226,17 +2227,33 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+        String vmInternalCSName = "";
+        String vmNameOnVcenter  = "";
+        if(vmSpec != null)
+        {
+            vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+            if (_instanceNameFlag == true && vmSpec.getType() == VirtualMachine.Type.User) {
```
rhtyd: If _instanceNameFlag is boolean, so there is no need of an explicit `== true`.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2226,17 +2227,33 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+                String[] tokens = vmInternalCSName.split("-");
+                if(tokens.length<3) throw new InvalidParameterValueException("Invalid vmInternalCSName format: "+vmInternalCSName+" expected format : *-*-* (e.g -> i-x-y)"); // vmInternalCSName has format i-x-y-<instance.name>
+                vmNameOnVcenter = String.format("%s-%s-%s-%s", tokens[0], tokens[1], tokens[2], vmSpec.getHostName());
```
rhtyd: The token length comparison may break backward compatibility for existing VMs that may not conform to this scheme. Also, assuming the token length is 3 the line can be written as:
`vmNameOnVcenter = String.format("%s-%s", vmInternalCSName, vmSpec.getHostName());`

As per previous code, `vmNameOnVcenter = vmSpec.getHostName()`, which will introduce backward compatibility issues for older 4.3 environments. Is there any code that handles for such old VMs. If a 4.3 user decides to upgrade to 4.10+, their VMs might not start the next time someone restarts. Please comment on how the change addresses backward compatibility issue.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2226,17 +2227,33 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+            }
+
         }
         return new Pair<String, String>(vmInternalCSName, vmNameOnVcenter);
     }
 
-    protected void configNestedHVSupport(VirtualMachineMO vmMo, VirtualMachineTO vmSpec, VirtualMachineConfigSpec vmConfigSpec) throws Exception {
+    protected static void configNestedHVSupport(VirtualMachineMO vmMo, VirtualMachineTO vmSpec, VirtualMachineConfigSpec vmConfigSpec) throws Exception {
```
rhtyd: Please explain why `static` was added, thanks.

rhtyd:
I've left some outstanding comments, I'm -1 on the PR.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
priyankparihar: >Why is the method attribute changed from private to protected?

Otherwise Unit testing will be difficult. 

>If vmspec is null, this method should throw an exception, to avoid a situation where empty vmnames are returned.

In CS vmspec can not be null. But still code will be  modified according your suggestion.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
priyankparihar: @rhtyd Good point. Code will be modified.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
priyankparihar: >As per previous code, vmNameOnVcenter = vmSpec.getHostName(), which will introduce backward compatibility issues for older 4.3 environments. Is there any code that handles for such old VMs.

Yes, vm.instancename.flag will handle backward compatibility issue. 

Case (vm.instancename.flag = True):
User can create the VM with same name form different account. 
Case (vm.instancename.flag = False):
User can not create the VM with same name form different account.
Note - But in both the cases, VM with same name are not allowed on a particular network. And only for the User VM's name format will be like "I-<user-id>-<Display-Name>" if "vm.instancename.flag = True". It will not cause any impact on other VM names( like Router VM).

priyankparihar:
@rhtyd I think now all of your questions are answered. Please provide your LGTM comment.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
priyankparihar: Removed.

karuturi:
@rhtyd can you review again?

plugins/hypervisors/vmware/test/com/cloud/hypervisor/vmware/resource/VmwareResourceTest.java
```diff
@@ -404,4 +410,46 @@ public void testConfigNestedHVSupportFlagFalse() throws Exception{
         verify(vmMo, never()).getRunningHost();
     }
 
+    @Test
+    public  void  testComposeVmNames() throws Exception{
```
rhtyd: @priyankparihar please split the tests into multiple case methods, it's easier to debug failures.

server/src/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -3383,12 +3383,6 @@ protected UserVm createVirtualMachine(DataCenter zone, ServiceOffering serviceOf
                     hostName = generateHostName(uuidName);
                 }
             }
-            // If global config vm.instancename.flag is set to true, then CS will set guest VM's name as it appears on the hypervisor, to its hostname.
-            // In case of VMware since VM name must be unique within a DC, check if VM with the same hostname already exists in the zone.
```
rhtyd: Instead of removing the check, does it make sense to amend it to verify that with the updated naming convention -- there are no duplicate VMs by name?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2226,13 +2227,30 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
...
+        String vmNameOnVcenter  = "";
+        if(vmSpec == null) {
+            throw new CloudRuntimeException("vmSpec is null");
         }
+
+        vmInternalCSName = vmNameOnVcenter = vmSpec.getName();
+        if (_instanceNameFlag && vmSpec.getType() == VirtualMachine.Type.User) {
+           String[] tokens = vmInternalCSName.split("-");
+           if(tokens.length<3) throw new InvalidParameterValueException("Invalid vmInternalCSName format: "+vmInternalCSName+" expected format : *-*-* (e.g -> i-x-y)"); // vmInternalCSName has format i-x-y-<instance.name>
+              vmNameOnVcenter = String.format("%s-%s-%s-%s", tokens[0], tokens[1], tokens[2], vmSpec.getHostName());
```
rhtyd: Should we validate here that tokens items are not null or empty; or we might get a vm name that has null in its name?

rhtyd:
@abhinandanprateek @DaanHoogland -- ping for review?
@PaulAngus would you like to share your thoughts in changing VM naming convention for vmware?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
DaanHoogland: @rhtyd I don't mind "null" in a name as long as the name is not null or duplicate. (see your comment further on)

plugins/hypervisors/vmware/test/com/cloud/hypervisor/vmware/resource/VmwareResourceTest.java
DaanHoogland: i agree, let's test as little as possible per test

server/src/com/cloud/vm/UserVmManagerImpl.java
DaanHoogland: are those not allowed (in different domains/accounts)?

DaanHoogland:
ping @PaulAngus @rhtyd is this useful to keep open?

@priyankparihar this needs rebasing (or closing)

PR #1309, opened 2016-01-05
CLOUDSTACK-9210: Pass secondary IPs to default_network_rules() function

wido:
This is a mandatory argument but it was NOT passed which caused the
re-programming of security groups to fail.

Simple fix to just add the argument since the variable is available
there.

### DIFF
--- a/scripts/vm/network/security_group.py
+++ b/scripts/vm/network/security_group.py
@@ -818,7 +818,7 @@ def add_network_rules(vm_name, vm_id, vm_ip, signature, seqno, vmMac, rules, vif
       execute("iptables -F " + egress_vmchain)
     except:
       logging.debug("Error flushing iptables rules for " + vmchain + ". Presuming firewall rules deleted, re-initializing." )
-      default_network_rules(vm_name, vm_id, vm_ip, vmMac, vif, brname)
+      default_network_rules(vm_name, vm_id, vm_ip, vmMac, vif, brname, sec_ips)
     egressrule = 0
     for line in lines:
         tokens = line.split(':')


wido:
@DaanHoogland @remibergsma @resmo Could you take a look?

DaanHoogland:
simple fix but a test on the contents in the called method makes sense as well. Not sure how to test.

wido:
@DaanHoogland We do not have any unit testing in place for Python. If you look just 10 lines up you see the method being called exactly that way.

Without this line the script simply errors out with a Python stacktrace.

wido:
Forgot to add, pylint still works fine. It scores 3.54/10. Not great, but a lint check still succeeds.

DaanHoogland:
@wido My point exactly. 10 lines up a conditional should guard the split() call

DaanHoogland:
btw LGTM if you don't want to improve further.

wido:
@DaanHoogland I agree, this script isn't the best in the world.

In my opinion we should offload this work to libvirt:
- http://libvirt.org/formatnwfilter.html
- https://issues.apache.org/jira/browse/CLOUDSTACK-1164

DaanHoogland:
@wido will your proposal from CLOUDSTACK-1164 work for other hypervisors as well?

wido:
@DaanHoogland No idea. Might work with Xen if we manage those Instances through libvirt. But in case of KVM we can probably do this with libvirt. Might need to patches in libvirt though.

ustcweizhou:
LGTM, just code review. simple but really make sense.

wido:
@remibergsma Can we merge this one? It fixes a bug.

To clarify, this is how the method has been defined:

<pre>def default_network_rules(vm_name, vm_id, vm_ip, vm_mac, vif, brname, sec_ips):</pre>

remibergsma:
@wido I have no way to test this myself. If you want this in a point release, please point the PR to 4.7 branch and reference the new PR to this one (so we can count the reviews). It will then end up in both 4.7.1 and 4.8.0.

wido:
@remibergsma I'm find with 4.8 since that should be here this month.

I really don't know how to test this. My python foo-skills tell me this patch is right. The function invocation was wrong and is now done right.

wido:
@remibergsma We have this one running in production right now and it is working as expected.

I spent some time in trying to get a test-case up and running, but I can't. I can just tell you that by looking at the code this was a stupid thing which wasn't caught during the addition of secondary IPs.

Our logs were full with Python stack traces and they are now gone :)

ustcweizhou:
I agree we need to fix some issues like this , which is difficult to be reproduced but it is obviously an issue from code view.

PR #1313, opened 2016-01-06
fix mariadb related listCapacity bug (CLOUDSTACK-8966)  a bigint(20) …

lttmtins:
…with varchar type does not work well on MariaDB
so forcing it to type decimal
Tested on ACS4.7 with MariaDB10.0.22 (galera)

To test (this is the output with the pull request code underneath):
(nl2) > list capacity sortby=usage type=0
count = 1
capacity:
capacitytotal = 0
capacityused = 3501740523520
percentused = 0
type = 0
zoneid = 76251030-aca1-44c6-b47d-8010ee17e0ad
zonename = NL2
(nl2) >


lttmtins:
On advice of release master this is now a pull request against 4.7

PR #1335, opened 2016-01-13
CLOUDSTACK-9230: Remove unnecessary return statement from cloudStack.js

nitin-maharana:
Removed the unnecessary return statement.
The statement is never reached.

### DIFF
--- a/ui/scripts/cloudStack.js
+++ b/ui/scripts/cloudStack.js
@@ -213,8 +213,6 @@
                         domainid: g_domainid
                     }
                 } : false;
-
-                return testAddUser;
             },
 
             // Actual login process, via form


ui/scripts/cloudStack.js
```diff
@@ -213,8 +213,6 @@
                         domainid: g_domainid
                     }
                 } : false;
-
```
DaanHoogland: statement is unreachable, indeed

DaanHoogland:
LGTM, this return statement is immediately following another return statement and is thus unreachable.

remibergsma:
LGTM, makes sense. Checked syntax and that is also still OK.

screen shot 2016-01-17 at 20 36 55

PR #1336, opened 2016-01-13
CLOUDSTACK-9231: Root volume migration from one primary to another primary storage within the same cluster is failing

nitin-maharana:
EXPECTED BEHAVIOUR:
====================
Root Volume migration within cluster should work.

ACTUAL BEHAVIOUR:
==================
Root volume migration within cluster failed.

This situation arises when there are two management server accessing the same database.
When the migration request comes the command is forwarded from one management server to another because the host is owned by the second management server. So, serialisation of map from one to another fails.

Fix:
===
This is fixed by converting the maps to lists.

### DIFF
--- a/core/src/com/cloud/agent/api/MigrateWithStorageCommand.java
+++ b/core/src/com/cloud/agent/api/MigrateWithStorageCommand.java
@@ -40,6 +40,13 @@ public class MigrateWithStorageCommand extends Command {
         this.tgtHost = null;
     }
 
+    public MigrateWithStorageCommand(VirtualMachineTO vm, List<Pair<VolumeTO, StorageFilerTO>> volumeToFilerAsList) {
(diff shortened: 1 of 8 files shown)


cristofolini:
Fix looks simple enough. Code LGTM. Although I would usually suggest adding a javadoc the new `MigrateWithStorageCommand` method you've created, this just changes the existing method (which looks clear enough to me already) to use Lists instead of maps, so I'll leave that decision up to you @nitin-maharana.

DaanHoogland:
[1336.vpc.results.txt](https://github.com/apache/cloudstack/files/93940/1336.vpc.results.txt)
[1336.network.results.txt](https://github.com/apache/cloudstack/files/93941/1336.network.results.txt)

haven't tested the fix itself yet

PR #1337, opened 2016-01-13
CLOUDSTACK-9235: Autoscale button is missing in VPC

nitin-maharana:
Autoscale button is missing. This should not be the case since we are able to add NS as the external LB provider in VPC.

Steps:
=====
1. Create a VPC offering with NS as the external LB provider
2. Create a VPC and configure the public tier with the above offering
3. Acquire an IP address and try to configure Load Balancing rule

Result:
======
Autoscale option is not visible at all for the LB.

Expected Result:
=============
Autoscale option should be available and should work exactly like the way it works for a normal isolated network with NS. If we choose NS as service provider for LB, autoscale should also be visible. If VR is chosen, only then we should not display autoscale.

Fix:
===
In case of VPC, it checks the services available.
If LB is there, It checks the provider is Netscaler then it shows the button or hides it.

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -3643,22 +3643,7 @@
                                                             success: function(json) {
                                                                 var networkOffering = json.listnetworkofferingsresponse.networkoffering[0];
                                                                 var services = networkOffering.service;
-                                                                if (services != null) {
-                                                                    for (var i = 0; i < services.length; i++) {
-                                                                        if (services[i].name == 'Lb') {
-                                                                            var providers = services[i].provider;
-                                                                            if (providers != null) {
-                                                                                for (var k = 0; k < providers.length; k++) {
-                                                                                    if (providers[k].name == 'Netscaler') {
-                                                                                        lbProviderIsNetscaler = true;
-                                                                                        break;
-                                                                                    }
-                                                                                }
-                                                                            }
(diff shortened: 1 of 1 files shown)


rafaelweingartner:
Hi @nitin-maharana, 
What about a little change in your code?

The complexity of that if/else structure that you are working on is too big.

I believe that you can extract the code from lines 3673-3686 to a method that returns a Boolean indicating if there is a provider named “NetScaler”. Then you could do something like : 
var lbProviderIsNetscaler = checkIfNetscalerProviderIsEnabled (services);

ui/scripts/network.js
```diff
@@ -3667,8 +3652,16 @@
...
                                                         }
                                                     } else { //from VPC section
-                                                        //VPC doesn't support autoscale
-                                                        return 2;
+                                                        var lbProviderIsNetscaler;
+                                                        var services = args.context.vpc[0].service;
+
+                                                        lbProviderIsNetscaler = checkIfNetScalerProviderIsEnabled(services);
+
+                                                        if (lbProviderIsNetscaler == true) { //AutoScale is only supported on Netscaler (but not on any other provider like VirtualRouter)
```
pdube: You could remove the condition here, it is unnecessary

pdube:
Code LGTM

PR #1338, opened 2016-01-13
CLOUDSTACK-9236: Load Balancing Health Check button displayed when non-NetScaler offering is used

nitin-maharana:
Load balancing health check option / button should only be displayed when a NetScaler based networking offering is being used. 

If you try to use the health check option when NetScaler is not being used as the load balancing technology the UI throws an error stating that the health check option is not supported. 

This button / option should be removed from the UI if NetScaler is not included in the network offering as customers will be confused. 

Fix:
===
This button will be shown only when the load balancer is NetScaler.
Otherwise it is hidden.

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -3620,7 +3620,37 @@
                                                     requireValidation: true,
                                                     buttonLabel: 'Configure',
                                                     action: cloudStack.uiCustom.healthCheck()
-
+                                                },
+                                                isHidden: function(args) {
+                                                    if (!('vpc' in args.context)) {  //From Guest Network section
+                                                        var lbProviderIsNetscaler = false;
+                                                        $.ajax({
+                                                            url: createURL('listNetworkOfferings'),
+                                                            data: {
+                                                                id: args.context.networks[0].networkofferingid
+                                                            },
+                                                            async: false,
+                                                            success: function(json) {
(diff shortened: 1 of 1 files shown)


rafaelweingartner:
@nitin-maharana, if you create the function I suggested you to create in PR #1337, you could re-use it here. The code block 3637-3650 is the same as the one in the other PR and you are working in the same file.

nitin-maharana:
@rafaelweingartner : Super I will make the modification on both PRs. Thanks.

remibergsma:
@nitin-maharana I like it when you put screenshots, as this speeds up my testing because I can see what you mean. I will test it soon, can you help me by putting a screenshot so that I look for the right thing? Thanks!

remibergsma:
LGTM based on these tests:

Situation before when not using Netscaler:

screen shot 2016-01-18 at 15 37 25

screen shot 2016-01-18 at 15 34 47

screen shot 2016-01-18 at 15 34 58

Basically, the button is displayed but throws an ugly message.

After the fix in this PR the button is hidden:

screen shot 2016-01-18 at 15 38 46

Didn't test if it shows when using Netscaler, as I don't have it in my test environment.

DaanHoogland:
I have some remarks about style (made one) but the logic looks fine. LGTM

nitin-maharana:
Thanks @remibergsma. Sorry I saw your comments bit late. You have already figured it out. Thanks @DaanHoogland

PR #1341, opened 2016-01-14
CLOUDSTACK-9238: Increase URL fields to 2048 charachters from 255

wido:
255 characters is to small for various URLs like S3 pre-signed URLs.

This causes one or more characters to be chopped of the end of the URL
and this renders them useless.

Internally in the code all URLs are passed as Strings and they are not
sized limited. This was purely in the database.

Other URL fields in the database were already 2048 characters.

This limit was introduced in the 4.1 to 4.2 upgrade when Object storage
like S3 and Swift was introduced in CloudStack for Secondary Storage.

### DIFF
--- a/setup/db/db/schema-471to480.sql
+++ b/setup/db/db/schema-471to480.sql
@@ -20,3 +20,10 @@
 --;
 
 ALTER TABLE `cloud`.`nicira_nvp_router_map` DROP INDEX `logicalrouter_uuid` ;
+
+ALTER TABLE `cloud`.`volume_host_ref` MODIFY COLUMN `url` varchar(2048);
+ALTER TABLE `cloud`.`object_datastore_ref` MODIFY COLUMN `url` varchar(2048);
+ALTER TABLE `cloud`.`image_store` MODIFY COLUMN `url` varchar(2048);
+ALTER TABLE `cloud`.`template_store_ref` MODIFY COLUMN `url` varchar(2048);
+ALTER TABLE `cloud`.`volume_store_ref` MODIFY COLUMN `url` varchar(2048);
+ALTER TABLE `cloud`.`volume_store_ref` MODIFY COLUMN `download_url` varchar(2048);


wido:
I tested this by manually applying all the SQL patches and verifying that all URLs are still in the database like they should be.

Also added a new template and saw that it registered as it should.

kevindierkx:
Should resolve problems with template extracts where the generated URL exceeds the 255 character limit.

BTW: 2048 is somewhere in the middle of what is defined as standard (website url length) and what is supported by most browsers.

borisroman:
@wido @kevindierkx Why not use TEXT? Then we'll never have to worry again :-)

kevindierkx:
@borisroman Because VARCHAR can handle 'most common' supported URL lengths perfectly fine.
Most modern browsers have a limit somewhere around 2000 characters.

More information about the matter: http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

wido:
Like @kevindierkx indeed says. A VARCHAR can be up to 64k characters long. It seems that 2k is the maximum length of a URL used in most systems.

That's why we sticked with the VARCHAR and didn't go for text. A VARCHAR is also a bit more optimal inside MySQL than a TEXT field.

bhaisaab:
LGTM

remibergsma:
This page has some interesting background:
http://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers

It seems URLs can be much longer. The question is whether we need those in CloudStack. For now I'm OK with 2K instead of 256. We can always bump them later when needed.

LGTM

wido:
Thanks @remibergsma !

That is the same reasoning as @kevindierkx and me had while looking into this. We went for 2k for now as it seems to match almost 100% of the use cases.

marcaurele:
Hi! Talking with @footplus on the mailing list, I realized that this patch is not correct. It's missing modifications on the VO objects to let the String holds more than 255 characters because in the `GenericDaoBase`, `String` fields are by default of 255 chars, see:
https://github.com/apache/cloudstack/blob/master/framework/db/src/com/cloud/utils/db/GenericDaoBase.java#L1510

This PR should be updated, or should there be a new one to add those changes?

marcaurele:
I guess a PR is easier, so here you go https://github.com/apache/cloudstack/pull/1567

PR #1342, opened 2016-01-15
CLOUDSTACK-6181: Add 'root disk size' field in instance wizard for KVM templates

ustcweizhou:
This commit includes three changes:
(1) Revert commit 13bf1ec5da9891eca98ef5dab61bf74b6db9530b to show the 'root disk size' field
(2) Set the default value from '1' to empty. This also changes the default value of customized service offerings from '1' to empty.
(3) show 'root disk size' field only when users choose a KVM template, as cloudstack supports deployvm with rootdisksize on KVM only.


DaanHoogland:
the jenkins failure is due to an old and unrelated licensing issue.

DaanHoogland:
LGTM based on code and function- manual test (see screen print)
screen shot 2016-01-16 at 12 02 44

@wilderrodrigues @remibergsma You guys have any use for this too? (hint intended)

remibergsma:
@ustcweizhou @DaanHoogland Will test this soon. Can you please add that the size that one needs to specify is in GB? Didn't look so I had to look that up ;-)

remibergsma:
LGTM, the UI shows fine and the value is passed to the API as shown in screenshot. See comment above, please add "GB" after the field.

cloudstack_pr1342_1

ui/scripts/ui-custom/instanceWizard.js
```diff
@@ -429,12 +427,52 @@
                                     $step.find('.custom-size-label').remove();
 
                                     if (!custom) {
```
pdube: Should become if(custom){//logic} to avoid empty condition

bhaisaab:
LGTM

bhaisaab:
@ustcweizhou please add the "GB" label after the string, then we can merge this. Thanks.

ustcweizhou:
add (GB) to the label. Please test it.

bhaisaab:
cc @remibergsma requires merging on 4.6 and consequent fwd merges

PR #1343, opened 2016-01-17
CLOUDSTACK-9240 remove 40GB filesize limit from SSVM scripts

remibergsma:
Both createvolume.sh and createtmplt.sh have a 40GB hardcoded limit for the size of the template that gets created. I could not find any justification of that. I am just removing them as they caused us a huge headache when we tried to create bigger templates and they failed without any good error.

This closes #1223 (This PR was against master, made a new one against 4.7)

Thanks Syed <syed1.mushtaq@gmail.com>

### DIFF
--- a/scripts/storage/secondary/createtmplt.sh
+++ b/scripts/storage/secondary/createtmplt.sh
@@ -27,7 +27,6 @@ usage() {
 
 
 #set -x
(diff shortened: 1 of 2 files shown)


remibergsma:
Will merge this based on the LGTMs in #1223 as the only difference is the branch.

PR #1350, opened 2016-01-19
Quota: consolidated lockable account check to a method. 

agneya2001:
Added unit tests to check lockablity of various accounts.

Currently normal user and domain admin accounts are eligible for locking.

### DIFF
--- a/framework/quota/test/org/apache/cloudstack/quota/QuotaManagerImplTest.java
+++ b/framework/quota/test/org/apache/cloudstack/quota/QuotaManagerImplTest.java
@@ -197,4 +197,42 @@ public class QuotaManagerImplTest extends TestCase {
         Mockito.verify(quotaAcc, Mockito.times(1)).persistQuotaAccount(Mockito.any(QuotaAccountVO.class));
     }
 
+    private AccountVO accountVO = new AccountVO();
+
+    @Test
+    public void testAdminLockableAccount() {
+        accountVO.setType(Account.ACCOUNT_TYPE_ADMIN);
+        assertFalse(quotaManager.isLockable(accountVO));
+    }
+
+    @Test
+    public void testNormalLockableAccount() {
+        accountVO.setType(Account.ACCOUNT_TYPE_NORMAL);
+        assertTrue(quotaManager.isLockable(accountVO));
+    }
+
+    @Test
+    public void tesDomainAdmingLockableAccount() {
+        accountVO.setType(Account.ACCOUNT_TYPE_DOMAIN_ADMIN);
+        assertTrue(quotaManager.isLockable(accountVO));
+    }
+
+    @Test
+    public void testReadOnlyAdminLockableAccount() {
+        accountVO.setType(Account.ACCOUNT_TYPE_READ_ONLY_ADMIN);
+        assertFalse(quotaManager.isLockable(accountVO));
+    }
+
+    @Test
+    public void testResourceDomainAdminLockableAccount() {
+        accountVO.setType(Account.ACCOUNT_TYPE_RESOURCE_DOMAIN_ADMIN);
+        assertFalse(quotaManager.isLockable(accountVO));
+    }
+
+    @Test
+    public void testProjectLockableAccount() {
+        accountVO.setType(Account.ACCOUNT_TYPE_PROJECT);
+        assertFalse(quotaManager.isLockable(accountVO));
+    }
+
 }
--- a/framework/quota/src/org/apache/cloudstack/quota/QuotaAlertManagerImpl.java
+++ b/framework/quota/src/org/apache/cloudstack/quota/QuotaAlertManagerImpl.java
@@ -27,7 +27,6 @@ import com.cloud.user.dao.UserDao;
 import com.cloud.utils.NumbersUtil;
 import com.cloud.utils.component.ManagerBase;
 import com.cloud.utils.db.TransactionLegacy;
-import com.cloud.utils.exception.CloudRuntimeException;
 import com.google.common.base.Strings;
 import com.sun.mail.smtp.SMTPMessage;
 import com.sun.mail.smtp.SMTPSSLTransport;
@@ -83,6 +82,8 @@ public class QuotaAlertManagerImpl extends ManagerBase implements QuotaAlertMana
     private ConfigurationDao _configDao;
     @Inject
     private QuotaUsageDao _quotaUsage;
+    @Inject
+    private QuotaManager _quotaManager;
 
     private EmailQuotaAlert _emailQuotaAlert;
     private boolean _lockAccountEnforcement = false;
@@ -158,7 +159,7 @@ public class QuotaAlertManagerImpl extends ManagerBase implements QuotaAlertMana
                 }
                 if (accountBalance.compareTo(zeroBalance) < 0) {
                     if (_lockAccountEnforcement && (lockable == 1)) {
-                        if (account.getType() == Account.ACCOUNT_TYPE_NORMAL) {
+                        if (_quotaManager.isLockable(account)) {
                             s_logger.info("Locking account " + account.getAccountName() + " due to quota < 0.");
                             lockAccount(account.getId());
                         }
@@ -382,7 +383,8 @@ public class QuotaAlertManagerImpl extends ManagerBase implements QuotaAlertMana
 
         public void sendQuotaAlert(List<String> emails, String subject, String body) throws MessagingException, UnsupportedEncodingException {
             if (_smtpSession == null) {
-                throw new CloudRuntimeException("Unable to create smtp session.");
+                s_logger.error("Unable to create smtp session.");
+                return;
             }
             SMTPMessage msg = new SMTPMessage(_smtpSession);
             msg.setSender(new InternetAddress(_emailSender, _emailSender));
--- a/framework/quota/src/org/apache/cloudstack/quota/QuotaManager.java
+++ b/framework/quota/src/org/apache/cloudstack/quota/QuotaManager.java
@@ -16,10 +16,13 @@
 //under the License.
 package org.apache.cloudstack.quota;
 
+import com.cloud.user.AccountVO;
 import com.cloud.utils.component.Manager;
 
 public interface QuotaManager extends Manager {
 
     boolean calculateQuotaUsage();
+
+    boolean isLockable(AccountVO account);
 
 }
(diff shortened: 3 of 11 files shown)


bhaisaab:
LGTM

cristofolini:
I see you've put a bunch of different test cases into a single `testLockableAccount` method. I'd strongly suggest you put each of those different cases into their own methods, so that if one of them fails in a build, we can easily determine what is failing. As it stands, that single test would fail if any of those cases goes wrong, leaving us to guess what failed.

agneya2001:
@cristofolini @bhaisaab @remibergsma test code split into simpler methods.

bhaisaab:
@agneya2001 LGTM again, need one more review to get this merged along with some testing (though in this case, we've some unit tests for the change)

rafaelweingartner:
Hi @agneya2001, there are a lot of duplicated lines at “QuotaManagerImplTest”, would you mind removing them?

Additionally, the lines you use to set the domainId and the id at the AccountVO object, you do not need to execute them, the method that you want to test does not need them.

agneya2001:
@rafaelweingartner Can you comment on the duplicated lines. There is a way you can get to the code and comment there.

rafaelweingartner:
I know that. I just did not want to go one by one and tell you that they are duplicated, I believe you can easily see that. Those duplicated lines I mentioned are in the test methods you created.

agneya2001:
@rafaelweingartner if you mean initialisation of AccountVO to be moved into a method, and replacing it with a method call. I guess I would prefer it this way as this keeps things simple. 
The initialisation for domainId and id is done as the method will receive similar object, so no harm in passing a object that is close to the one passed at runtime, though as you said the function tested is not using it.

rafaelweingartner:
@agneya2001  that is exactly what I meant. 
I disagree with you, letting this as they are, it is not simple. Why duplicate code, when we can do better? Especially if we use a good and descriptive method name.

About the initialization, I also disagree with you, at that point those values are not used, so they are needed and you do not need to code those lines.
If those attributes are needed somewhere else, they should be used only at the point in which they are needed.

Please, keep in mind the less the better. We need quality in our code, not quantity.

agneya2001:
@rafaelweingartner quality is what we are discussing and I simply do not think that putting 3 lines of object initialisation in a method will make it high quality. Are you referring to some guidelines that I can see too ?

rafaelweingartner:
It is cultural; we have to stop duplicating lines; no matter where and no matter how simple those blocks may seem. 

We also have to stop coding things that are not needed as you did in the test. If a code is not needed, why bother writing it down? I am referring to the “accountVO.setId” and accountVO.setDomainId” at the tests you wrote. 

I do not know if ACS has a guideline for that, for me, it looks like common sense. That was my opinion, but if others here are ok with the duplication of lines, feel free to merge the PR.

agneya2001:
@rafaelweingartner If it is a black box test then the person testing will obviously try to pass a realistic data set, that is what I have done. It is possible that tomorrow there are some changes to the tested method and the data that looks superficial may come into play and expose some erratic behaviour. The dataset passed mocks the dataset that will go thru the method in real runtime condition. Now offcourse we can see that the data is not being used in this method but does that mean I should change the dataset that mocks the real value ?

rafaelweingartner:
If in a remotely distant future someone that we do not even know yet changes that method and make use of some extra data (forgetting to check the test cases), the test cases should catch that and then the person that is doing that change should fix the tests. That is why we write test cases, for the code of today, not for the code of tomorrow.

IMO we should prepare the architecture of software to the future, but the code itself should not be developed waiting for some futuristic hypothetical condition that may not ever happen.

agneya2001:
@rafaelweingartner  It is only about realistic test data that is being passed to the method tested. You should differentiate between the tested code and the code that is testing. The logic/code that is being tested should have all the characteristics that you are talking about ie any data that is not being used should not be there.
But when it comes to writing tests and creating mocks you do it in a more realistic manner without deducing what could be happening inside the method as you only test what the method is supposed to be doing. That makes it error proof in future and that was what i was talking about.

rafaelweingartner:
I do not agree.
If people are comfortable with it, so let it be.

bhaisaab:
@rafaelweingartner fair points, but it would make sense if you can comment on the diff in the context of the change.

Specific to the PR I see several testXXXLockableAccount which tries to test account types separately. Are you suggesting to get them all refactored, say put in a single testLockableAccount() and use a single Account/User object to test the lockability instead of explicitly separate test methods?

rafaelweingartner:
@bhaisaab I think the test cases in different methods, each one for a single and unit test is the right way to go.

What I disagree is with the creation of “AccountVO” object. Anyone that looks at lines 202 – 204, can see that they are the same as 211- 214. Additionally, those "Id" and "DomainId" parameters that are being set are not used by the “isLockable” method. 

On top of everything, can you please look at lines 221-222, 231-232 and so forth; that is a clear example of code duplication without even thinking.
```
AccountVO accountVO = new AccountVO(); 
accountVO = new AccountVO();
```

bhaisaab:
@agneya2001 how about you have a look at the test tomorrow and see if you can incorporate some of the suggestions like pull out the account from the method as a class variable?

agneya2001:
@rafaelweingartner @bhaisaab I have taken out the duplicate account initialisation.

agneya2001:
@cristofolini your changes are incorporated can you review these.

framework/quota/test/org/apache/cloudstack/quota/QuotaManagerImplTest.java
```diff
@@ -197,4 +197,58 @@ public void testProcessQuotaBalanceForAccount() {
...
+        assertFalse(quotaManager.isLockable(accountVO));
+    }
+
+    @Test
+    public void testProjectLockableAccount() {
+        AccountVO accountVO = new AccountVO();
+        accountVO.setId(7L);
+        accountVO.setDomainId(1L);
+        accountVO.setType(Account.ACCOUNT_TYPE_PROJECT);
+        assertFalse(quotaManager.isLockable(accountVO));
```
bhaisaab: @abhinandanprateek We can actually remove the setId and setDomainId in all the testxxxLockableAccounts; create like a single class leve AccountVO and then in each of the testxxxLockableAccounts methods simply call the setType method and then call assert on them.

agneya2001: @bhaisaab I don't agree with the change as it impacts simplicity and quality. All the three lines that you are asking me have a data element that will become a function parameter, I am not sure how it will simplify or result in enhanced quality.

rafaelweingartner:
I disagree that it would make the code more complex.
That is what I meant:
```
private AccountVO createAccountVoForType(short type) {
        AccountVO accountVO = new AccountVO();
        accountVO.setType(type);
        return accountVO;
  }
    
   @Test
    public void testAdminLockableAccount() {
            AccountVO accountVO = createAccountVoForType(Account.ACCOUNT_TYPE_ADMIN);
            assertFalse(quotaManager.isLockable(accountVO));
        }
...
```

agneya2001:
This code that was suggested:

private AccountVO createAccountVoForType(short type) {
        AccountVO accountVO = new AccountVO();
        accountVO.setType(type);
        return accountVO;
  }

   @Test
    public void testAdminLockableAccount() {
            AccountVO accountVO = createAccountVoForType(Account.ACCOUNT_TYPE_ADMIN);
            assertFalse(quotaManager.isLockable(accountVO));
        }
        
        @Test
    public void testAdminLockableAccount() {
            AccountVO accountVO = createAccountVoForType(Account.ACCOUNT_TYPE_ADMIN);
            assertFalse(quotaManager.isLockable(accountVO));
        }

and this code that will otherwise be:


   @Test
   public void testResourceDomainAdminLockableAccount() {
       AccountVO accountVO = new AccountVO();
            accountVO.setType(Account.ACCOUNT_TYPE_RESOURCE_DOMAIN_ADMIN);
	     assertFalse(quotaManager.isLockable(accountVO));
   }
   
   @Test
    public void testProjectLockableAccount() {
        AccountVO accountVO = new AccountVO();
        accountVO.setType(Account.ACCOUNT_TYPE_PROJECT);
        assertFalse(quotaManager.isLockable(accountVO));
   }
   

@rafaelweingartner  To me the code below looks simple and more readable. (In general if this was not the test code some of this code will be in constructor. Coming to constructors, there is much more repeatable code there in many cloudstack classes.) Now this might not look good to you and some others too. I suggest if that is the case take some effort and put up a code style guideline or at least adopt a guideline that already exists so that we donot have to go thru this. This will also be food for further cleanup as findbug and code style check already are CC: @bhaisaab @remibergsma @DaanHoogland @cristofolini. Till then this stands as it is.

DaanHoogland:
@agneya2001 Sounds like you want to set the style guide. you and @rafaelweingartner should take this discussion to a higher level and take it to a discuss thread.

agneya2001:
@DaanHoogland I agree there should be some guideline on code style. If these are there I think most of us will be willing to follow these. I guess it will not be difficult to adapt something that is already out there.

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 121
 Hypervisor xenserver
 NetworkType Advanced
 Passed=104
 Failed=14
 Skipped=4

**Failed tests:**
* <nose.suite

 * ContextSuite context=TestVpcRemoteAccessVpn>:setup Failed

* integration.smoke.test_privategw_acl.TestPrivateGwACL

 * test_01_vpc_privategw_acl Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_deploy_vgpu_enabled_vm
test_06_copy_template
test_06_copy_iso

**Passed test suits:**

rhtyd:
LGTM, there are some comments on cosmetics but in general we should get this CI tested and merged cc @swill 

tag:easypr

swill:
@rafaelweingartner This one seems to be in OK shape.  I am missing one code review, given that you have actively reviewed this code, can I get your final status?

rafaelweingartner:
@swill, 
I did a review on the comments and discussions to get me up to date (we had a discussion about some code duplications). Those discussions were pretty good, I believe that there was some block of code being repeated, and also some blocks (or lines I do not recall) that were never used or not needed.

I did a code walkthrough and I cannot say anything about the code. It is very good now. I can for sure give an LGMT here ;)

Sorry guys to say this, but, those discussions were not merely about cosmetics!

swill:
@rafaelweingartner thanks for the review.  :)  I think what @rhtyd meant is anything outstanding is only cosmetic, I don't think he meant the entire discussion.

I think this one is ready to merge...

rhtyd:
@swill yes that's what I meant, therefore commented that it's ready for merge. Thanks.

PR #1359, opened 2016-01-22
Refactor system VM default network creation

ProjectMoon:
Two small commits which moves the retrieval of the default network for the console proxy and the SSVM into a separate protected method. It's a small change that makes the code more readable/maintainable and also makes the class more suitable for overriding should one want to do this. It's forward-ported from our 4.2 branch.

No new tests since this should not change any functionality, and thus should be covered by the existing unit tests.

### DIFF
--- a/server/src/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
+++ b/server/src/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
@@ -668,18 +668,10 @@ public class ConsoleProxyManagerImpl extends ManagerBase implements ConsoleProxy
         return null;
     }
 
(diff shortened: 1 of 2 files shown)


wilderrodrigues:
Hi @ProjectMoon 

Could you please create you PR against Master. If that's a refactor does not really make sense to create the PR against 4.6 because we are already releasing 4.7.1 and 4.8.0.

Cheers,
Wilder

ProjectMoon:
OK. So just to be sure: it's oldest supported maintenance release for bug fixes and then master for any new features/refactoring?

remibergsma:
@ProjectMoon Indeed! :-)

PR #1363, opened 2016-01-25
CLOUDSTACK-9251: Fix issue in scale VM to dynamic service offering

ustcweizhou:
This reverts commit 9c4162ac7f451fc3e2155418dcfff224c8c08a4a and 16baa1289b7de383e98d0070717b3f1873fa2db3

Before change: exception when change compute offering (to dynamic service offering) on UI
After change: succeed

### DIFF
--- a/api/src/org/apache/cloudstack/api/command/user/vm/UpgradeVMCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/vm/UpgradeVMCmd.java
@@ -16,6 +16,9 @@
 // under the License.
 package org.apache.cloudstack.api.command.user.vm;
 
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 
 import org.apache.log4j.Logger;
@@ -77,7 +80,18 @@ public class UpgradeVMCmd extends BaseCmd {
     }
 
     public Map<String, String> getDetails() {
-        return details;
+        Map<String, String> customparameterMap = new HashMap<String, String>();
+        if (details != null && details.size() != 0) {
+            Collection parameterCollection = details.values();
+            Iterator iter = parameterCollection.iterator();
+            while (iter.hasNext()) {
+                HashMap<String, String> value = (HashMap<String, String>)iter.next();
+                for (String key : value.keySet()) {
+                    customparameterMap.put(key, value.get(key));
+                }
+            }
+        }
+        return customparameterMap;
     }
 
     /////////////////////////////////////////////////////
--- a/api/src/org/apache/cloudstack/api/command/user/vm/ScaleVMCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/vm/ScaleVMCmd.java
@@ -16,6 +16,9 @@
 // under the License.
 package org.apache.cloudstack.api.command.user.vm;
 
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
@@ -78,8 +81,22 @@ public class ScaleVMCmd extends BaseAsyncCmd {
         return serviceOfferingId;
     }
 
+    //instead of reading a map directly we are using collections.
+    //it is because details.values() cannot be cast to a map.
+    //it gives a exception
     public Map<String, String> getDetails() {
-        return details;
+        Map<String, String> customparameterMap = new HashMap<String, String>();
+        if (details != null && details.size() != 0) {
+            Collection parameterCollection = details.values();
+            Iterator iter = parameterCollection.iterator();
+            while (iter.hasNext()) {
+                HashMap<String, String> value = (HashMap<String, String>)iter.next();
+                for (String key : value.keySet()) {
+                    customparameterMap.put(key, value.get(key));
+                }
+            }
+        }
+        return customparameterMap;
     }
 
     /////////////////////////////////////////////////////


glennwagner:
This looks good , Will test it in the LAB today

pdube:
Code LGTM

alexandrelimassantana:
@ustcweizhou there are some modifications you could do to make this code a little more clean and readable.

**1)** You are duplicating code, since two classes have the same code written. You should reallocate this method to a common superclass.

**2)** You could change lines **84-86** (ScaleVMCmd.java file) from line comments to the Javadoc's pattern, which can be done much like a block comment, but with double asterisks /** Javadoc **/

```Java
/**
* I am a Javadoc and this is general information about the method
* @param This is a param documentation.
* @return This is a method's return documentation.
**/
```

**3)** You should use **!details.isEmpty()** instead of **details.size() != 0**. It better states what you are looking for and it's performance is slightly better than the comparison you are using.

**4)** It's a good practice to declare the variables as the first statements of the method, that will help when someone is reading inside conditionals, the same method written with these changes would look like:

```Java
public Map<String, String> getDetails(){
    Map<String, String> customparameterMap = new HashMap<String, String>();
    Collection parameterCollection;
    Iterator iter;
    HashMap<String, String> value;

    if (details == null || details.isEmpty())
        return customparameterMap;

    parameterCollection = details.values();
    iter = parameterCollection.iterator();

    while (iter.hasNext()) {
        value = (HashMap<String, String>)iter.next();
        for (String key : value.keySet())
            customparameterMap.put(key, value.get(key));
    }

    return customparameterMap;
}
```

ustcweizhou:
@alexandrelimassantana I agree with you about the code changes.
However, this commit only reverts the commit 9c4162a and 16baa12 to fix the issue.

The issue you mentioned does exist in our codebase,  not only in these two commands, but more than 10 other commands (some are implemented in different way), you can change them in a single PR if you want.

```
[root@cloudstack(4.6)]# grep -r details.values api/src/
api/src/org/apache/cloudstack/api/BaseUpdateTemplateOrIsoCmd.java:        Collection paramsCollection = this.details.values();
api/src/org/apache/cloudstack/api/command/admin/storage/UpdateCloudToUseObjectStoreCmd.java:            Collection<?> props = details.values();
api/src/org/apache/cloudstack/api/command/admin/storage/CreateSecondaryStagingStoreCmd.java:            for (Map<String, String> detail : details.values()) {
api/src/org/apache/cloudstack/api/command/admin/storage/AddImageStoreCmd.java:            Collection<?> props = details.values();
api/src/org/apache/cloudstack/api/command/admin/offering/CreateServiceOfferingCmd.java:            Collection<?> props = details.values();
api/src/org/apache/cloudstack/api/command/admin/network/CreateNetworkOfferingCmd.java:        Collection paramsCollection = details.values();
api/src/org/apache/cloudstack/api/command/admin/account/UpdateAccountCmd.java:        Collection paramsCollection = details.values();
api/src/org/apache/cloudstack/api/command/admin/account/CreateAccountCmd.java:        Collection<String> paramsCollection = details.values();
api/src/org/apache/cloudstack/api/command/user/template/RegisterTemplateCmd.java:        Collection paramsCollection = details.values();
api/src/org/apache/cloudstack/api/command/user/template/CreateTemplateCmd.java:        Collection paramsCollection = details.values();
api/src/org/apache/cloudstack/api/command/user/template/GetUploadParamsForTemplateCmd.java:        Collection paramsCollection = details.values();
api/src/org/apache/cloudstack/api/command/user/volume/AddResourceDetailCmd.java:            Collection<?> servicesCollection = details.values();
api/src/org/apache/cloudstack/api/command/user/vm/ScaleVMCmd.java:    //it is because details.values() cannot be cast to a map.
api/src/org/apache/cloudstack/api/command/user/vm/ScaleVMCmd.java:            Collection parameterCollection = details.values();
api/src/org/apache/cloudstack/api/command/user/vm/UpdateVMCmd.java:        Collection<String> paramsCollection = this.details.values();
api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java:            Collection parameterCollection = details.values();
api/src/org/apache/cloudstack/api/command/user/vm/UpgradeVMCmd.java:            Collection parameterCollection = details.values();
```

remibergsma:
The commits that get reverted here were supposed to fix a findbugs issue. Will this reintroduce them? Can we both fix it and resolve the findbugs issue? Ping @DaanHoogland

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 115
 Hypervisor xenserver
 NetworkType Advanced
 Passed=105
 Failed=11
 Skipped=4

test_vpc_remote_access_vpn
test_vpc_site2site_vpn
test_04_change_offering_small
test_01_primary_storage_iscsi
test_04_extract_template
test_04_extract_Iso
test_07_list_default_iso
test_dedicateGuestVlanRange


**Failed tests:**

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_deploy_vgpu_enabled_vm
test_06_copy_template
test_06_copy_iso

**Passed test suits:**

koushik-das:
+1 based on the code review (reverts 2 commits) and based on the test results posted by @bvbharatk (no failed tests).
Since existing functionality got broken by findbug fixes, I would prefer to have them reverted and later relook at them. @DaanHoogland What do you think?

swill:
@koushik-das Can you specify which functionality tests were broken by which PR?  Thanks...

koushik-das:
@swill As mentioned by @ustcweizhou in the first comment. I am not sure if the current scale VM tests cover the dynamic offering scenario.

"This reverts commit 9c4162a and 16baa12
Before change: exception when change compute offering (to dynamic service offering) on UI"

swill:
Ok, I understand what is going on here better.  @DaanHoogland would you mind reviewing this for us?  Thanks...

api/src/org/apache/cloudstack/api/command/user/vm/UpgradeVMCmd.java
```diff
@@ -77,7 +80,18 @@ public Long getServiceOfferingId() {
     }
 
     public Map<String, String> getDetails() {
-        return details;
+        Map<String, String> customparameterMap = new HashMap<String, String>();
+        if (details != null && details.size() != 0) {
+            Collection parameterCollection = details.values();
+            Iterator iter = parameterCollection.iterator();
+            while (iter.hasNext()) {
+                HashMap<String, String> value = (HashMap<String, String>)iter.next();
```
DaanHoogland: the findbugs issue is here, a string is returned as a HashMap<String, String>. The fact that this works is very ugly. It means that details is really a Map<String,Map<String,String>>. If we want to both revert the issue and solve the findsbugs issue, the solution might be something like writing a parser and parse the string iter.next() into a map. Not sure  though

swill: Thanks @DaanHoogland.  @ustcweizhou can you look into improving this?

ustcweizhou:
@swill sorry I will not, that is another issue.

DaanHoogland:
@swill @ustcweizhou has a point. Id say merge. I'm not going to take the time to do this now either. One point to consider is that it will up the number of findbugs issues again and will let jenkins bug us with loads of mails again untill we find the time to fix this issue. You could cheat and set the new baseline for findbugs to the merge build.

swill:
Hmmm, not sure how I feel about this.  My inbox can't handle a whole bunch of new findbugs issues.  :)

@remibergsma would you mind giving me some guidance on this one?

koushik-das:
How about suppressing the findbug issue (using something like @SuppressWarnings) till someone comes up with a proper fix? In that case there won't be any mails and the regression is fixed as well. Currently there are lot of places in code where suppress is used.

jburwell:
@koushik-das I am a hard -1 to a ``@SupressWarnings`` because it will never get fixed.  Was the FindBugs issue present before this patch?  If so, it would be unfair to block it because developers could be forced to fix a pile of preexisting and nothing would more forward.  However, if this FindBugs issue was introduced by this patch, then it needs to be fixed before it can be merged.  The whole point of the static analysis checks is prevent new issues from entering master.

koushik-das:
@jburwell I am not saying FindBugs issues should be suppressed in general. As part of fixing FindBugs issues some functionality got regressed (refer PR heading). This PR simply reverts some of the commits that caused the regression. Now fixing the original FindBugs issue in correct way involves some amount of work which is out of scope of this PR. That's the reason I mentioned to suppress the FindBugs issue which will resurface once the regression is fixed.

swill:
So my understand is there was a FindBugs issue originally.  In the process of fixing the issue, some logic was changed to introduce an issue.  This commit is to revert to the original code that had the FindBugs issue.  I guess it is a lesser of two evils.  If this goes in, I think we should do what @DaanHoogland suggested and up the number of findbugs to account for the known findbug issues in this commit.  This way we are not spammed with findbugs issues and if other commits have findbugs issues, we are correctly notified.  Does that make sense?

jburwell:
I am confused as to the issue at this point.  Building the PR using ``mvn clean install findbugs:check`` passes.  It appears that commit reversions have addressed the issues.  Assuming I ran the appropriate, I don't see why static analysis should be blocking the progress of this PR.

@koushik-das ``@SuppressWarnings`` will hide the issue forever because, in reality, we will never come back to it.  Therefore, it should never be used if we intend for static analysis to be useful/valuable practice

@swill Speaking generally, my opinion is that the suggestion does not work.  The point of static analysis is to protect the integrity of the codebase.  If a patch introduces a new issue, it shouldn't be accepted until the issue is fixed.  The only exception to this rule is that we determine the rule is no longer valuable and stop checking it.  We only gain value from static analysis if we are disciplined about addressing the issues found rather than taking the most expedient path to get a pass.

swill:
So do we have 2 LGTM code reviews for this then?  @jburwell do you give this a +1?

jburwell:
@swill LGTM -- passes static analysis and unit tests + code looks good

swill:
Ok, thanks.  I will add to merge list.

PR #1364, opened 2016-01-25
CLOUDSTACK-9256 add unique key for static routes in json

remibergsma:
Static routes that are being set do not show up in the static_routes.json file. The reason for this is that the index that is used, is the gateway address, which is not unique. Hence stuff is overwritten and lost.

Ping @borisroman @wilderrodrigues @DaanHoogland

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs_staticroutes.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs_staticroutes.py
@@ -20,7 +20,7 @@ from pprint import pprint
 
 def merge(dbag, staticroutes):
     for route in staticroutes['routes']:
-        key = route['ip_address']
+        key = route['network']
         revoke = route['revoke']
         if revoke:
             try:


borisroman:
LGTM :+1: 

Already running in production at SBP!

remibergsma:
`cat /etc/cloudstack/staticroutes.json` before and after change:

Before always had one entry:


```
{
    "10.135.255.10": {
        "gateway": "10.135.255.129",
        "ip_address": "10.135.255.10",
        "network": "10.0.0.0/9",
        "revoke": false
    },
 }
```

After this PR:

```
{
    "10.0.0.0/9": {
        "gateway": "10.135.255.129",
        "ip_address": "10.135.255.10",
        "network": "10.0.0.0/9",
        "revoke": false
    },
    "10.128.0.0/13": {
        "gateway": "10.135.255.129",
        "ip_address": "10.135.255.10",
        "network": "10.128.0.0/13",
        "revoke": false
    },
}
```

DaanHoogland:
Routes are supposed to be to networks when going out, so this makes sense. production at sbp is good enough a test for me: LGTM

PR #1365, opened 2016-01-25
[4.7] vmware: improve support for disks

rhtyd:
- Improve disk chain usage while attaching, migrating disks
- Gets root disk controller based diskDeviceBusName from volume's chain info

### DIFF
--- a/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -236,7 +236,7 @@ import com.cloud.hypervisor.vmware.mo.HypervisorHostHelper;
 import com.cloud.hypervisor.vmware.mo.NetworkDetails;
 import com.cloud.hypervisor.vmware.mo.TaskMO;
 import com.cloud.hypervisor.vmware.mo.VirtualEthernetCardType;
-import com.cloud.hypervisor.vmware.mo.VirtualMachineDiskInfo;
+import org.apache.cloudstack.utils.volume.VirtualMachineDiskInfo;
 import com.cloud.hypervisor.vmware.mo.VirtualMachineDiskInfoBuilder;
 import com.cloud.hypervisor.vmware.mo.VirtualMachineMO;
 import com.cloud.hypervisor.vmware.mo.VirtualSwitchType;
@@ -1412,7 +1412,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
         String vmInternalCSName = names.first();
         String vmNameOnVcenter = names.second();
         String dataDiskController = vmSpec.getDetails().get(VmDetailConstants.DATA_DISK_CONTROLLER);
-        String rootDiskController = vmSpec.getDetails().get(VmDetailConstants.ROOK_DISK_CONTROLLER);
+        String rootDiskController = vmSpec.getDetails().get(VmDetailConstants.ROOT_DISK_CONTROLLER);
 
         // If root disk controller is scsi, then data disk controller would also be scsi instead of using 'osdefault'
         // This helps avoid mix of different scsi subtype controllers in instance.
@@ -1451,7 +1451,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                 s_logger.error(msg);
                 throw new Exception(msg);
             }
-
+            String guestOsId = translateGuestOsIdentifier(vmSpec.getArch(), vmSpec.getOs(), vmSpec.getPlatformEmulator()).value();
             DiskTO[] disks = validateDisks(vmSpec.getDisks());
             assert (disks.length > 0);
             NicTO[] nics = vmSpec.getNics();
@@ -1564,7 +1564,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                         tearDownVm(vmMo);
                     }else if (!hyperHost.createBlankVm(vmNameOnVcenter, vmInternalCSName, vmSpec.getCpus(), vmSpec.getMaxSpeed().intValue(),
                             getReservedCpuMHZ(vmSpec), vmSpec.getLimitCpuUse(), (int)(vmSpec.getMaxRam() / (1024 * 1024)), getReservedMemoryMb(vmSpec),
-                            translateGuestOsIdentifier(vmSpec.getArch(), vmSpec.getOs(), vmSpec.getPlatformEmulator()).value(), rootDiskDataStoreDetails.first(), false, controllerInfo, systemVm)) {
+                            guestOsId, rootDiskDataStoreDetails.first(), false, controllerInfo, systemVm)) {
                         throw new Exception("Failed to create VM. vmName: " + vmInternalCSName);
                     }
                 }
@@ -1588,7 +1588,6 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
             }
 
             VirtualMachineConfigSpec vmConfigSpec = new VirtualMachineConfigSpec();
-            String guestOsId = translateGuestOsIdentifier(vmSpec.getArch(), vmSpec.getOs(), vmSpec.getPlatformEmulator()).value();
 
             VmwareHelper.setBasicVmConfig(vmConfigSpec, vmSpec.getCpus(), vmSpec.getMaxSpeed(),
                     getReservedCpuMHZ(vmSpec), (int)(vmSpec.getMaxRam() / (1024 * 1024)), getReservedMemoryMb(vmSpec),
@@ -2322,14 +2321,14 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
 
         if (vol.getType() == Volume.Type.ROOT) {
             Map<String, String> vmDetails = vmSpec.getDetails();
-            if (vmDetails != null && vmDetails.get(VmDetailConstants.ROOK_DISK_CONTROLLER) != null) {
-                if (vmDetails.get(VmDetailConstants.ROOK_DISK_CONTROLLER).equalsIgnoreCase("scsi")) {
+            if (vmDetails != null && vmDetails.get(VmDetailConstants.ROOT_DISK_CONTROLLER) != null) {
+                if (vmDetails.get(VmDetailConstants.ROOT_DISK_CONTROLLER).equalsIgnoreCase("scsi")) {
                     s_logger.info("Chose disk controller for vol " + vol.getType() + " -> scsi, based on root disk controller settings: " +
-                            vmDetails.get(VmDetailConstants.ROOK_DISK_CONTROLLER));
+                            vmDetails.get(VmDetailConstants.ROOT_DISK_CONTROLLER));
                     controllerKey = scsiControllerKey;
                 } else {
                     s_logger.info("Chose disk controller for vol " + vol.getType() + " -> ide, based on root disk controller settings: " +
-                            vmDetails.get(VmDetailConstants.ROOK_DISK_CONTROLLER));
+                            vmDetails.get(VmDetailConstants.ROOT_DISK_CONTROLLER));
                     controllerKey = ideControllerKey;
                 }
             } else {
--- a/plugins/hypervisors/vmware/src/com/cloud/storage/resource/VmwareStorageProcessor.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/storage/resource/VmwareStorageProcessor.java
@@ -33,6 +33,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
 
+import com.google.common.base.Strings;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 
@@ -88,7 +89,7 @@ import com.cloud.hypervisor.vmware.mo.HostMO;
 import com.cloud.hypervisor.vmware.mo.HostStorageSystemMO;
 import com.cloud.hypervisor.vmware.mo.HypervisorHostHelper;
 import com.cloud.hypervisor.vmware.mo.NetworkDetails;
-import com.cloud.hypervisor.vmware.mo.VirtualMachineDiskInfo;
+import org.apache.cloudstack.utils.volume.VirtualMachineDiskInfo;
 import com.cloud.hypervisor.vmware.mo.VirtualMachineMO;
 import com.cloud.hypervisor.vmware.mo.VmwareHypervisorHost;
 import com.cloud.hypervisor.vmware.resource.VmwareResource;
@@ -1363,24 +1364,15 @@ public class VmwareStorageProcessor implements StorageProcessor {
             AttachAnswer answer = new AttachAnswer(disk);
 
             if (isAttach) {
-                String dataDiskController = controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER);
-                String rootDiskController = controllerInfo.get(VmDetailConstants.ROOK_DISK_CONTROLLER);
-                DiskControllerType rootDiskControllerType = DiskControllerType.getType(rootDiskController);
-
-                if (dataDiskController == null) {
-                    dataDiskController = getLegacyVmDataDiskController();
-                } else if ((rootDiskControllerType == DiskControllerType.lsilogic) ||
-                           (rootDiskControllerType == DiskControllerType.lsisas1068) ||
-                           (rootDiskControllerType == DiskControllerType.pvscsi) ||
-                           (rootDiskControllerType == DiskControllerType.buslogic)) {
-                    //TODO: Support mix of SCSI controller types for single VM. If root disk is already over
-                    //a SCSI controller then use the same for data volume as well. This limitation will go once mix
-                    //of SCSI controller types for single VM.
-                    dataDiskController = rootDiskController;
-                } else if (DiskControllerType.getType(dataDiskController) == DiskControllerType.osdefault) {
-                    dataDiskController = vmMo.getRecommendedDiskController(null);
+                String diskController = getLegacyVmDataDiskController();
+                if (controllerInfo != null &&
+                        !Strings.isNullOrEmpty(controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER))) {
+                    diskController = controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER);
                 }
-                vmMo.attachDisk(new String[] {datastoreVolumePath}, morDs, dataDiskController);
+                if (DiskControllerType.getType(diskController) == DiskControllerType.osdefault) {
+                    diskController = vmMo.getRecommendedDiskController(null);
+                }
+                vmMo.attachDisk(new String[] {datastoreVolumePath}, morDs, diskController);
             } else {
                 vmMo.removeAllSnapshots();
                 vmMo.detachDisk(datastoreVolumePath, false);
--- a/pom.xml
+++ b/pom.xml
@@ -92,7 +92,7 @@
     <cs.servlet.version>2.5</cs.servlet.version>
     <cs.jstl.version>1.2</cs.jstl.version>
     <cs.selenium.server.version>1.0-20081010.060147</cs.selenium.server.version>
-    <cs.vmware.api.version>5.5</cs.vmware.api.version>
+    <cs.vmware.api.version>6.0</cs.vmware.api.version>
     <org.springframework.version>3.2.12.RELEASE</org.springframework.version>
     <cs.mockito.version>1.9.5</cs.mockito.version>
     <cs.powermock.version>1.5.3</cs.powermock.version>
--- a/server/src/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/com/cloud/storage/VolumeApiServiceImpl.java
@@ -30,8 +30,10 @@ import javax.inject.Inject;
 
 import com.cloud.utils.EncryptionUtil;
 import com.cloud.utils.db.TransactionCallbackWithException;
+import com.google.common.base.Strings;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonParseException;
 
 import org.apache.cloudstack.api.command.user.volume.GetUploadParamsForVolumeCmd;
 import org.apache.cloudstack.api.response.GetUploadParamsResponse;
@@ -39,6 +41,7 @@ import org.apache.cloudstack.engine.subsystem.api.storage.DataObject;
 import org.apache.cloudstack.engine.subsystem.api.storage.EndPoint;
 import org.apache.cloudstack.storage.command.TemplateOrVolumePostUploadCommand;
 import org.apache.cloudstack.utils.imagestore.ImageStoreUtil;
+import org.apache.cloudstack.utils.volume.VirtualMachineDiskInfo;
 import org.apache.log4j.Logger;
 import org.apache.cloudstack.api.command.user.volume.AttachVolumeCmd;
 import org.apache.cloudstack.api.command.user.volume.CreateVolumeCmd;
@@ -112,12 +115,14 @@ import com.cloud.hypervisor.Hypervisor.HypervisorType;
 import com.cloud.hypervisor.HypervisorCapabilitiesVO;
 import com.cloud.hypervisor.dao.HypervisorCapabilitiesDao;
 import com.cloud.org.Grouping;
+import com.cloud.serializer.GsonHelper;
 import com.cloud.service.dao.ServiceOfferingDetailsDao;
 import com.cloud.storage.Storage.ImageFormat;
 import com.cloud.storage.dao.DiskOfferingDao;
 import com.cloud.storage.dao.SnapshotDao;
 import com.cloud.storage.dao.VMTemplateDao;
 import com.cloud.storage.dao.VolumeDao;
+import com.cloud.storage.dao.VolumeDetailsDao;
 import com.cloud.storage.snapshot.SnapshotApiService;
 import com.cloud.storage.snapshot.SnapshotManager;
 import com.cloud.template.TemplateManager;
@@ -146,6 +151,7 @@ import com.cloud.utils.db.UUIDManager;
 import com.cloud.utils.exception.CloudRuntimeException;
 import com.cloud.utils.fsm.NoTransitionException;
 import com.cloud.utils.fsm.StateMachine2;
+import com.cloud.vm.UserVmManager;
 import com.cloud.vm.UserVmVO;
 import com.cloud.vm.VMInstanceVO;
 import com.cloud.vm.VirtualMachine;
@@ -190,6 +196,8 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
     ConfigurationManager _configMgr;
     @Inject
     VolumeDao _volsDao;
+    @Inject
+    VolumeDetailsDao _volDetailDao;
     @Inject
     HostDao _hostDao;
     @Inject
@@ -240,6 +248,9 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
     VmWorkJobDao _workJobDao;
     @Inject
     ClusterDetailsDao _clusterDetailsDao;
+    @Inject
+    UserVmManager _userVmMgr;
+    protected Gson _gson;
 
     private List<StoragePoolAllocator> _storagePoolAllocators;
 
@@ -253,6 +264,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
 
     protected VolumeApiServiceImpl() {
         _volStateMachine = Volume.State.getStateMachine();
+        _gson = GsonHelper.getGsonLogger();
     }
 
     /*
@@ -1835,6 +1847,26 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
         }
     }
 
+    public void updateMissingRootDiskController(final VMInstanceVO vm, final String rootVolChainInfo) {
+        if (vm == null || !VirtualMachine.Type.User.equals(vm.getType()) || Strings.isNullOrEmpty(rootVolChainInfo)) {
+            return;
+        }
+        String rootDiskController = null;
+        try {
+            final VirtualMachineDiskInfo infoInChain = _gson.fromJson(rootVolChainInfo, VirtualMachineDiskInfo.class);
+            if (infoInChain != null) {
+                rootDiskController = infoInChain.getControllerFromDeviceBusName();
+            }
+            final UserVmVO userVmVo = _userVmDao.findById(vm.getId());
+            if ((rootDiskController != null) && (!rootDiskController.isEmpty())) {
+                _userVmDao.loadDetails(userVmVo);
+                _userVmMgr.persistDeviceBusInfo(userVmVo, rootDiskController);
+            }
+        } catch (JsonParseException e) {
+            s_logger.debug("Error parsing chain info json: " + e.getMessage());
+        }
+    }
+
     @DB
     @Override
     @ActionEvent(eventType = EventTypes.EVENT_VOLUME_MIGRATE, eventDescription = "migrating volume", async = true)
@@ -1924,6 +1956,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
                                 throw new InvalidParameterValueException("Cannot migrate ROOT volume of a stopped VM to a storage pool in a different VMware datacenter");
                             }
                         }
+                        updateMissingRootDiskController(vm, vol.getChainInfo());
                     }
                 }
             }
@@ -2472,9 +2505,10 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
             }
             _userVmDao.loadDetails(vm);
             Map<String, String> controllerInfo = new HashMap<String, String>();
-            controllerInfo.put(VmDetailConstants.ROOK_DISK_CONTROLLER, vm.getDetail(VmDetailConstants.ROOK_DISK_CONTROLLER));
+            controllerInfo.put(VmDetailConstants.ROOT_DISK_CONTROLLER, vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER));
             controllerInfo.put(VmDetailConstants.DATA_DISK_CONTROLLER, vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER));
             cmd.setControllerInfo(controllerInfo);
+            s_logger.debug("Attach volume id:" + volumeToAttach.getId() +  " on VM id:" + vm.getId() + " has controller info:" + controllerInfo);
 
             try {
                 answer = (AttachAnswer)_agentMgr.send(hostId, cmd);
--- a/server/src/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/com/cloud/vm/UserVmManagerImpl.java
@@ -89,6 +89,7 @@ import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
 import org.apache.cloudstack.storage.datastore.db.TemplateDataStoreDao;
 import org.apache.cloudstack.storage.datastore.db.TemplateDataStoreVO;
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 
 import com.cloud.agent.AgentManager;
@@ -3509,15 +3510,15 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                 if (hypervisorType.equals(HypervisorType.VMware)) {
                     if (guestOS.getDisplayName().toLowerCase().contains("apple mac os")) {
                         vm.setDetail("smc.present", "TRUE");
-                        vm.setDetail(VmDetailConstants.ROOK_DISK_CONTROLLER, "scsi");
+                        vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, "scsi");
                         vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "scsi");
                         vm.setDetail("firmware", "efi");
                         s_logger.info("guestOS is OSX : overwrite root disk controller to scsi, use smc and efi");
                     } else {
                         String controllerSetting = _configDao.getValue("vmware.root.disk.controller");
                         // Don't override if VM already has root/data disk controller detail
-                        if (vm.getDetail(VmDetailConstants.ROOK_DISK_CONTROLLER) == null) {
-                            vm.setDetail(VmDetailConstants.ROOK_DISK_CONTROLLER, controllerSetting);
+                        if (vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER) == null) {
+                            vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, controllerSetting);
                         }
                         if (vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER) == null) {
                             if (controllerSetting.equalsIgnoreCase("scsi")) {
@@ -5452,6 +5453,17 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
 
             vm.setDetail("Encrypted.Password", encryptedPasswd);
             _vmDao.saveDetails(vm);
+        }
+    }
+
+    public void persistDeviceBusInfo(UserVmVO vm, String rootDiskController) {
+        String existingVmRootDiskController = vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER);
+        if (StringUtils.isEmpty(existingVmRootDiskController) && !StringUtils.isEmpty(rootDiskController)) {
+            vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, rootDiskController);
+            _vmDao.saveDetails(vm);
+            if (s_logger.isDebugEnabled()) {
+                s_logger.debug("Persisted device bus information rootDiskController=" + rootDiskController + " for vm: " + vm.getDisplayName());
+            }
         }
     }
 
--- a/server/test/com/cloud/vm/UserVmManagerTest.java
+++ b/server/test/com/cloud/vm/UserVmManagerTest.java
@@ -928,4 +928,25 @@ public class UserVmManagerTest {
             CallContext.unregister();
         }
     }
+
+    @Test
+    public void testPersistDeviceBusInfoWithNullController() {
+        when(_vmMock.getDetail(any(String.class))).thenReturn(null);
+        _userVmMgr.persistDeviceBusInfo(_vmMock, null);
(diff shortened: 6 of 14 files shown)


GabrielBrascher:
@bhaisaab Could you please do the following changes?

At **VirtualMachineVolumeChainInfo** class:
**1 -** create Javadoc blocks (would be nice to document all the class, but I think that at least the class and the "getControllerFromDeviceBusName" method deserves a documentation);
**2 -** simplify the **if** conditional at line **43**. Test with "( (StringUtils.isEmpty) || (!this.diskDeviceBusName.contains(":")) )" condition. The StringUtils.isEmpty method already checks if the String is empty("") or null (https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html).
**3 -** create test for the getControllerFromDeviceBusName method.

At **UserVmManagerImpl** class:
**1 -** document the "persistDeviceBusInfo(UserVmVO vm, String rootDiskController)" method (lines **5458-5466**).
**2 -** use the "StringUtils.isEmpty()" with the condition at line **5460**, instead of "(existingVmRootDiskController==null) || (existingVmRootDiskController.isEmpty())".
**3 -** create test for "persistDeviceBusInfo(UserVmVO vm, String rootDiskController)" method.

**VolumeApiServiceImpl**:
**1 -** Remove “**_**” from variables names (lines **200**, **252**, **267**): private variables with “**_**” at the beginning is common in C++ but not in Java.

Thanks.

bhaisaab:
@GabrielBrascher Thanks for reviewing. I'm avoiding comments in the code, methods are pretty small, self-explanatory. Tests have been added, along with the StringUtils fix. I've deduped a class defition as well. I've deduped a class definition as well. Regarding the "_" issue, it's a codebase wide issue. For now I'm sticking to how rest of the private vars are named.

GabrielBrascher:
@bhaisaab Thanks for the test and StringUtils fix.

**-** Could you please change the Access Levels of diskDeviceBusName and diskChain from public to private (at class VirtualMachineDiskInfo)?

I asked for a Javadoc at the getControllerFromDeviceBusName() method because at first I thought that it was a simple get of a private String, then I saw the logic inside it.

I agree that the method is clean and easy to understand, so it is your call. Maybe it would be interesting to change its name then.

As a javadoc I was expecting something simple as:
`/**Returns the diskDeviceBusName substring, with the index starting at ":"*/`

Thanks.

bhaisaab:
@GabrielBrascher thanks, fixed the accessor

core/src/org/apache/cloudstack/storage/command/AttachAnswer.java
```diff
@@ -34,6 +37,12 @@ public AttachAnswer(DiskTO disk) {
         setDisk(disk);
     }
 
+    public AttachAnswer(DiskTO disk, Map<String, String> diskDetails) {
```
GabrielBrascher: @bhaisaab Thanks for fixing the access level. The code seems ok, just one more question.

Sorry to ask you now, I should have seen this before. Is this constructor used (**AttachAnswer(DiskTO disk, Map<String, String> diskDetails)**)?

Thanks.

bhaisaab: @GabrielBrascher Thanks. I started some small feature to be able to save/retrieve disk details, it is likely I dropped it but forgot to remove remove this -- I'll check and remove this.

bhaisaab: @GabrielBrascher thanks, fixed.

vmware-base/src/com/cloud/hypervisor/vmware/mo/GuestOsDescriptorType.java
```diff
@@ -0,0 +1,46 @@
...
+    if (value == null) {
+      return null;
+    }
+    GuestOsDescriptorType guestOsType = null;
+    try {
+      guestOsType = valueOf(value);
+    } catch (IllegalArgumentException e) {
+    }
+    return guestOsType;
+  }
```
GabrielBrascher: @bhaisaab Seems that the **com.cloud.utils.EnumUtils** class can do what you want (https://github.com/apache/cloudstack/blob/3ded3e90007d08fa98465f2b8c68b7fb075557c0/utils/src/main/java/com/cloud/utils/EnumUtils.java)

If you like of this approach, it would be something like `return EnumUtils.fromString(GuestOsDescriptorType.class, value);`

bhaisaab: Thanks @GabrielBrascher I checked that with latest VMware SDK I could remove this wrapper. I've pushed another PR on master that bumps up the vmware sdk version to 6.0. I'll remove this soon, and resend for a review.

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 114
 Hypervisor xenserver
 NetworkType Advanced
 Passed=105
 Failed=13
 Skipped=4

test_vpc_remote_access_vpn
test_vpc_site2site_vpn
test_04_change_offering_small
test_01_primary_storage_iscsi
test_04_extract_template
test_04_extract_Iso
test_07_list_default_iso
test_dedicateGuestVlanRange 


**Failed tests:**
* integration.smoke.test_guest_vlan_range.TestDedicateGuestVlanRange

 * test_dedicateGuestVlanRange Failing since 4 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_deploy_vgpu_enabled_vm
test_06_copy_template
test_06_copy_iso

**Passed test suits:**

bhaisaab:
@swill fyi, this was a production issue that was delivered as a patch and this was heavily tested by ShapeBlue internally against vmware environment with several rounds of upgrade tests. This is now running in one of the large prod. environments for weeks now. With this, I'm going to merge this fix on 4.5 (https://github.com/apache/cloudstack/pull/1366), please help merge on 4.7+.

swill:
I would like to get at least one code review with a LGTM before I merge.  I am pretty confident with the tests run.  As long as we get 105 tests passing in that CI environment things should be good.  This run was from a while ago before he cleaned up the tests that always fail in that env, so this is pretty consistent with the latest passing runs...

@jburwell would you mind doing a quick once through of this for me for the LGTM?  You are going to regret offering.  :P

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -4729,6 +4729,17 @@ private VirtualMachineGuestOsIdentifier translateGuestOsIdentifier(String cpuArc
         return VirtualMachineGuestOsIdentifier.OTHER_GUEST;
     }
 
+    private String translateGuestOsIdentifierEx(String cpuArchitecture, String guestOs, String cloudGuestOs) {
+        if (cloudGuestOs != null) {
```
rafaelweingartner: What about inverting the logic here? So, we can reduce one if-inception.
If(cloudGuestOs == null){
return translateGuestOsIdentifier(cpuArchitecture, guestOs, cloudGuestOs).value();
}
...the rest of the code that was initially inside the if.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1924,6 +1935,23 @@ public Volume migrateVolume(MigrateVolumeCmd cmd) {
...
+                                VirtualMachineDiskInfo infoInChain = _gson.fromJson(rootVolChainInfo, VirtualMachineDiskInfo.class);
+                                if (infoInChain != null) {
+                                    rootDiskController = infoInChain.getControllerFromDeviceBusName();
+                                }
+                                UserVmVO userVmVo = _userVmDao.findById(vm.getId());
+                                if ((rootDiskController != null) && (!rootDiskController.isEmpty())) {
+                                    _userVmDao.loadDetails(userVmVo);
+                                    _userVmMgr.persistDeviceBusInfo(userVmVo, rootDiskController);
+                                }
+                            } catch (JsonParseException ignored) {
```
rafaelweingartner: At least a warning/debug message should be printed. This approach of silencing exception is never a good idea.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1924,6 +1935,23 @@ public Volume migrateVolume(MigrateVolumeCmd cmd) {
                                 throw new InvalidParameterValueException("Cannot migrate ROOT volume of a stopped VM to a storage pool in a different VMware datacenter");
                             }
                         }
+
```
rafaelweingartner: What about extracting lines 1939-1954 to a method? This can really help us to work on that code and the unit tests it deserves.

server/src/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -5455,6 +5456,15 @@ private void encryptAndStorePassword(UserVmVO vm, String password) {
         }
     }
 
+    public void persistDeviceBusInfo(UserVmVO vm, String rootDiskController) {
+        String existingVmRootDiskController = vm.getDetail(VmDetailConstants.ROOK_DISK_CONTROLLER);
```
rafaelweingartner: what about fixing the typo "ROOK_DISK_CONTROLLER"?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -4729,6 +4729,17 @@ private VirtualMachineGuestOsIdentifier translateGuestOsIdentifier(String cpuArc
         return VirtualMachineGuestOsIdentifier.OTHER_GUEST;
     }
 
+    private String translateGuestOsIdentifierEx(String cpuArchitecture, String guestOs, String cloudGuestOs) {
```
rafaelweingartner: What about two or three unit tests for this method?

server/src/com/cloud/vm/UserVmManagerImpl.java
bhaisaab: Fixed

server/src/com/cloud/storage/VolumeApiServiceImpl.java
bhaisaab: Fixed

server/src/com/cloud/storage/VolumeApiServiceImpl.java
bhaisaab: Not necessary unless this method needs to be used somewhere else.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
bhaisaab: You've proposed a cosmetics fix that does not work either. If `guestOsId` is not null always (on line 4737) only then your suggestion makes sense; the return method is used as a catchall case.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
bhaisaab: Curious, why 2 or 3; not 10 or 100?

This method can be removed if PR https://github.com/apache/cloudstack/pull/1397 is accepted, which bumps vmware dependency version to sdk 6.0 which has the additional guest os identifier (i.e. no need to this extra translation).

rafaelweingartner: 2 or 3 was a matter of speaking. It would be needed tests to cover all of the expected behavior, which I do not know.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
rafaelweingartner: @bhaisaab, I did not understand what you said.

The way it is coded right now when “cloudGuestOs == null ”, it will be executed the line 4740. Do you agree with that?

My suggestion is not purely cosmetics; it is to reduce code complexity. You can maintain that behavior while removing cyclomatic complexity and that I think it is worth.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
rafaelweingartner: I disagree with you here. Those huge (really huge) methods that ACS has are the rock in our shoes when writing unit and integration tests (not talking about functional tests here). It is almost impossible to write and maintain a test case for a method with 50, 100, 500+ lines. 

Even though a method is not used in more than one place, I believe it is a good practice to break them into smaller ones whenever possible. That facilitates the writing of unit and integration tests cases.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
jburwell: agreed with @rafaelweingartner -- short circuiting out of the method greatly reduces complexity by reducing nesting.

Also, it seems like the check should be ``Strings.isBlankorNull`` to check for empty and, as well as, null values.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -4729,6 +4729,17 @@ private VirtualMachineGuestOsIdentifier translateGuestOsIdentifier(String cpuArc
...
+        if (cloudGuestOs != null) {
+            GuestOsDescriptorType guestOsId = GuestOsDescriptorType.fromValue(cloudGuestOs);
+            if (guestOsId != null) {
+                s_logger.debug("Found guest OS mapping name for guest os: " + guestOs);
```
jburwell: This ``debug`` call should be wrapped in a ``if (s_logger.isDebugEnabled())`` block.  Also, please include the value of ``cloudGuestOs`` in the log message to explain from which value it was mapped.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -4729,6 +4729,17 @@ private VirtualMachineGuestOsIdentifier translateGuestOsIdentifier(String cpuArc
         return VirtualMachineGuestOsIdentifier.OTHER_GUEST;
     }
 
+    private String translateGuestOsIdentifierEx(String cpuArchitecture, String guestOs, String cloudGuestOs) {
+        if (cloudGuestOs != null) {
+            GuestOsDescriptorType guestOsId = GuestOsDescriptorType.fromValue(cloudGuestOs);
+            if (guestOsId != null) {
```
bhaisaab: @jburwell @rafaelweingartner if I do what you are proposing and if `guestOsId` here is null, I would need to return translateGuestOsIdentifier again

rafaelweingartner: @bhaisaab , you are right! Sorry for my stubbornness, I missed that, my eyes are tired, I must be getting old :)
@jburwell, I agree with the use of StringUtils though.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
rafaelweingartner: I agree with the log message. However, I believe that adding the “isDebugEnabled()” would not be needed. The Logger.debug will check that, of course, it might be a little costly, but for legibility, I would let it as it is.

plugins/hypervisors/vmware/src/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -1363,24 +1363,17 @@ private Answer attachVolume(Command cmd, DiskTO disk, boolean isAttach, boolean
...
-                    dataDiskController = rootDiskController;
-                } else if (DiskControllerType.getType(dataDiskController) == DiskControllerType.osdefault) {
-                    dataDiskController = vmMo.getRecommendedDiskController(null);
+                String diskController = null;
+                if (controllerInfo != null) {
+                    diskController = controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER);
                 }
-                vmMo.attachDisk(new String[] {datastoreVolumePath}, morDs, dataDiskController);
+                if (diskController == null) {
+                    diskController = getLegacyVmDataDiskController();
```
jburwell: Why not set ``diskController`` to ``getLegacyVmDataDiskController`` when initializing the variable and avoid this additional conditional?

plugins/hypervisors/vmware/src/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -1363,24 +1363,17 @@ private Answer attachVolume(Command cmd, DiskTO disk, boolean isAttach, boolean
...
+                }
+                if (DiskControllerType.getType(diskController) == DiskControllerType.osdefault) {
+                    diskController = vmMo.getRecommendedDiskController(null);
+                }
```
jburwell: Lines 1366-1375 seem ripe to extraction to a method with package default visibility and a set of unit tests.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
GabrielBrascher: @bhaisaab I have to agree with @rafaelweingartner. Create methods is also a great way to improve the code (no just a way to get a reusable code). It helps on keep it readable, ensure its correct implementation by implementing tests, and allows to create Javadoc. I think that it could improve this code too.
Thanks.

jburwell: agree with @rafaelweingartner.  All methods should have a single responsibility -- it is critical to maintainability and testability.  While it is not possible to fix this problem all at once, we certainly shouldn't be adding to it.

server/src/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -5455,6 +5456,15 @@ private void encryptAndStorePassword(UserVmVO vm, String password) {
         }
     }
 
+    public void persistDeviceBusInfo(UserVmVO vm, String rootDiskController) {
+        String existingVmRootDiskController = vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER);
+        if (StringUtils.isEmpty(existingVmRootDiskController) && !StringUtils.isEmpty(rootDiskController)) {
+            vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, rootDiskController);
+            _vmDao.saveDetails(vm);
+            s_logger.debug("Persisted device bus information rootDiskController=" + rootDiskController + " for vm: " + vm.getDisplayName());
```
jburwell: Wrap this ``debug`` call in a ``if (isDebugEnabled()`` block.  Also, please add the vm id to the log message to assist with debugging.

utils/src/main/java/org/apache/cloudstack/utils/volume/VirtualMachineDiskInfo.java
```diff
@@ -39,4 +38,11 @@ public void setDiskDeviceBusName(String diskDeviceBusName) {
     public void setDiskChain(String[] diskChain) {
         this.diskChain = diskChain;
     }
+
+    public String getControllerFromDeviceBusName() {
+        if (StringUtils.isEmpty(diskDeviceBusName) || !diskDeviceBusName.contains(":")) {
+            return null;
```
jburwell: Return an empty string rather than null to avoid potential NPEs.

utils/src/main/java/org/apache/cloudstack/utils/volume/VirtualMachineDiskInfo.java
```diff
@@ -39,4 +38,11 @@ public void setDiskDeviceBusName(String diskDeviceBusName) {
...
+        }
+        return diskDeviceBusName.substring(0, diskDeviceBusName.indexOf(":") - 1);
+    }
```
jburwell: Please add a unit test for this method.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
jburwell: @rafaelweingartner the issue is the string concatenation that is performed before the method is invoked.  It is best practice to wrap all debug and trace log calls in ``isDebugEnabled`` or ``isTraceEnabled`` to avoid the pressure on the constant pool and garbage collector from such string concatenation across the codebase.

utils/src/test/java/org/apache/cloudstack/utils/volume/VirtualMachineDiskInfoTest.java
```diff
@@ -0,0 +1,46 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class VirtualMachineDiskInfoTest {
+
+    @Test
+    public void testGetControllerFromDeviceBusName() {
+        VirtualMachineDiskInfo vmDiskInfo = new VirtualMachineDiskInfo();
+        vmDiskInfo.setDiskDeviceBusName("scsi0:0");
+        vmDiskInfo.setDiskChain(new String[]{"[somedatastore] i-3-VM-somePath/ROOT-1.vmdk"});
+        Assert.assertEquals(vmDiskInfo.getControllerFromDeviceBusName(), "scsi");
+    }
```
jburwell: Please add test cases to pass null and blank strings, as well as, a strings without a ":" into the method.

vmware-base/src/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
```diff
@@ -2118,8 +2118,12 @@ public int getScsiDiskControllerKey(String diskController) throws Exception {
         List<VirtualDevice> devices = (List<VirtualDevice>)_context.getVimClient().
                 getDynamicProperty(_mor, "config.hardware.device");
 
+        String deviceList = "";
```
jburwell: Why not represent as an actual ``List`` and use ``Joiner`` to render it to a ``String``?

vmware-base/src/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
```diff
@@ -2137,7 +2141,8 @@ public int getScsiDiskControllerKey(String diskController) throws Exception {
         }
 
         assert (false);
-        throw new Exception(diskController + " Controller Not Found");
+        throw new Exception("Scsi disk controller of type " + diskController + " not found among configured devices:" + deviceList +
```
jburwell: Throwing ``Exception`` is an anti-pattern.  Use a specific checked or unchecked exception type to better express the nature of the problem (e.g. ``IllegalStateException``).  Also, since an exception is being thrown, ``assert(false)`` is not necessary and can be removed.

vmware-base/src/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
```diff
@@ -2137,7 +2141,8 @@ public int getScsiDiskControllerKey(String diskController) throws Exception {
...
+                ". Unable to find and return scsi disk controller key.");
```
jburwell: Throwing ``Exception`` is an anti-pattern.  Use a specific checked or unchecked exception type to better express the nature of the problem (e.g. ``IllegalStateException``).  Also, since an exception is being thrown, ``assert(false)`` is not necessary and can be removed.

server/test/com/cloud/vm/UserVmManagerTest.java
```diff
@@ -928,4 +928,13 @@ public void testUpdateVmNicIpFailure3() throws Exception {
...
         }
     }
+
+    @Test
+    public void testPersistDeviceBusInfo() {
+        when(_vmMock.getDetail(any(String.class))).thenReturn(null);
+        _userVmMgr.persistDeviceBusInfo(_vmMock, null);
+        verify(_vmDao, times(0)).saveDetails(any(UserVmVO.class));
+        _userVmMgr.persistDeviceBusInfo(_vmMock, "lsilogic");
+        verify(_vmDao, times(1)).saveDetails(any(UserVmVO.class));
```
jburwell: Why not split these two ``verify`` into two separate test methods to better express the scenario being tested?  It would help with diagnosing the specific failure and make the code more maintainable.

Also, please add test cases for when the ``existingVmRootDiskController`` is empty and ``rootDiskController`` is empty.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
rafaelweingartner: Ok, got it. I agree with you

vmware-base/src/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
bhaisaab: Fixed

vmware-base/src/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
bhaisaab: Removed, was not necessary

vmware-base/src/com/cloud/hypervisor/vmware/mo/GuestOsDescriptorType.java
bhaisaab: Removed this, bumping the sdk version to 6.0 we won't need this extra layer of translation

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
bhaisaab: Removed this, bumping the sdk version to 6.0 we won't need this extra layer of translation

plugins/hypervisors/vmware/src/com/cloud/storage/resource/VmwareStorageProcessor.java
bhaisaab: Fixed

server/src/com/cloud/vm/UserVmManagerImpl.java
bhaisaab: Fixed

server/test/com/cloud/vm/UserVmManagerTest.java
bhaisaab: Fixed

utils/src/main/java/org/apache/cloudstack/utils/volume/VirtualMachineDiskInfo.java
bhaisaab: In either case, we'll need to do a comparison to check if the string is null or empty. Returning an empty string might cover up potential functional issues, while NPEs are bad but can expose the issue without hiding it.

utils/src/test/java/org/apache/cloudstack/utils/volume/VirtualMachineDiskInfoTest.java
bhaisaab: Fixed

bhaisaab:
@swill @rafaelweingartner @jburwell fixed outstanding issues and added more unit tests

swill:
I can't specifically test the VMware functionality, but I can validate that it does not break anything else.  I have added this to my queue for CI.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1835,6 +1847,25 @@ private Volume orchestrateDetachVolumeFromVM(long vmId, long volumeId) {
         }
     }
 
+    public void updateMissingRootDiskController(final VMInstanceVO vm, final String rootVolChainInfo) {
+        if (vm != null && vm.getType().equals(VirtualMachine.Type.User) && !Strings.isNullOrEmpty(rootVolChainInfo)) {
```
rafaelweingartner: very nice the extraction you did here.
What about the short-circuit for the condition at line 1851?

```
if (vm == null || !VirtualMachine.Type.User.equals(vm.getType()) || Strings.isNullOrEmpty(rootVolChainInfo)) {
	return;
}
...
the rest of the code (lines 1852-1865)
...

```

bhaisaab: @rafaelweingartner yeah why not, fixed

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1835,6 +1847,26 @@ private Volume orchestrateDetachVolumeFromVM(long vmId, long volumeId) {
         }
     }
 
+    public void updateMissingRootDiskController(final VMInstanceVO vm, final String rootVolChainInfo) {
+        if (vm == null || !vm.getType().equals(VirtualMachine.Type.User) || Strings.isNullOrEmpty(rootVolChainInfo)) {
```
rafaelweingartner: We can avoid a NEP here by changing 
`vm.getType().equals(VirtualMachine.Type.User)`
to 
`VirtualMachine.Type.User.equals(vm.getType())`

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1835,6 +1847,26 @@ private Volume orchestrateDetachVolumeFromVM(long vmId, long volumeId) {
...
+            return;
+        }
+        String rootDiskController = null;
+        try {
+            final VirtualMachineDiskInfo infoInChain = _gson.fromJson(rootVolChainInfo, VirtualMachineDiskInfo.class);
+            if (infoInChain != null) {
```
rafaelweingartner: Do we need this check ?
is it possible that `gson.fromJson` returns a null ?
I believe it does not return null values. If it cannot execute the unmarshall process, an exception happens, right?

jburwell: Since it is a third-party library, it never safe to make such an assumption.  Better to be defensive in the event that the behavior of the library varies or changes in the future.

bhaisaab:
@swill can we get this merged, while ShapeBlue has tested this internally if you test with your non-vmware CI that should confirm it does not break anything else; ideally we would need a vmware-based CI to confirm this but if you can consider our internal testing results. Thanks.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
rafaelweingartner: @jburwell I understand the idea of being defensive against NPE, but we should be pragmatic too. Otherwise, we will end up adding tons of null checks that are not needed.

rafaelweingartner:
@swill , @bhaisaab 
There are still questions to be addressed here. One example is the question at “VolumeApiServiceImpl”  line1857. Another one is at “VolumeApiServiceImpl” line 1851.

There maybe more, some message were hidden due to code changes. This is a huge PR, which makes it pretty hard to review and then to check the changes that are being made (the ones that are consequences of code review). I think we need more work here.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
bhaisaab: @rafaelweingartner we'll get NPE anyway, we'll have to check vm != null as in what you've suggested we're doing equals against vm.getType() (therefore vm!=null check needed here).

rafaelweingartner: @bhaisaab I was no talking about removing the "vm != null", that check I know we need to execute. 
I was only refering to the "vm.getType().equals(VirtualMachine.Type.User)", the "vm.getType" can return a null value.

Did you understand what I mean? Just inverting the comparison as I said before.
from
`vm.getType().equals(VirtualMachine.Type.User)`
to 
`VirtualMachine.Type.User.equals(vm.getType())`

bhaisaab:
@rafaelweingartner both your comments are opinionated and not technical; we've to agree to disagree here ask you to be practical and pragmatic here. I've left comments above and I disagree on both mentioned places. Finally, what you would consider may be a huge change but for me a +190/-42 lines is not huge. If you go on this path, you might even find issues with quarks and neutrinos, and there is no end to over-engineering.

rafaelweingartner:
if you say so.
Be my guest.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
bhaisaab: the order of evalution will be left to right -- the first case is vm == null; if this qualifies the if statement will branch into executing code; otherwise vm.getType() code will be executed (i.e. vm is not null); now here I think you're suggesting that perhaps vm.getType() may return null in which case we may have a NPE (unlikely based on db contraints (the type column should not be null), but I'll modify as suggested.

swill:
I think this one is pretty close.  I want to run CI against it since there have been changes since the last CI run.  I also need some LGTM code reviews.  I don't see any yet even though there has been some pretty extensive review of this PR so far.  Can @jburwell @GabrielBrascher and @rafaelweingartner revisit this PR and give me your final review of the current state.  Thx...

rafaelweingartner:
@swill my final state is LGTM, will this be merged forward?
I believe @bhaisaab has mentioned that he has another PR that bumps VMware dependencies and that makes this PR obsolete, right? Or am my mistaking this with other PR?

swill:
@rafaelweingartner I am not sure.  @bhaisaab would you mind clarifying for us.  I am trying to stay on top of this stuff, but I am very much playing catchup.  :)

pom.xml
```diff
@@ -92,7 +92,7 @@
     <cs.servlet.version>2.5</cs.servlet.version>
     <cs.jstl.version>1.2</cs.jstl.version>
     <cs.selenium.server.version>1.0-20081010.060147</cs.selenium.server.version>
-    <cs.vmware.api.version>5.5</cs.vmware.api.version>
+    <cs.vmware.api.version>6.0</cs.vmware.api.version>
```
bhaisaab: @rafaelweingartner @swill I'll moved version change from the other PR here ^^

bhaisaab:
@swill we're good with this PR and it is not dependent on any other PR. For building noredist rpms/deb, one can use this repo with the dep-jars -- https://github.com/bhaisaab/cloudstack-nonoss

bhaisaab:
@GabrielBrascher @swill @jburwell @rafaelweingartner please share your LGTM or comment what needs fixing, thanks.

swill:
### CI RESULTS

```
Tests Run: 85
  Skipped: 0
   Failed: 0
   Errors: 0
```



**Associated Uploads**

**`/tmp/MarvinLogs/DeployDataCenter__Apr_28_2016_06_35_56_T0QPTY:`**
* [dc_entries.obj](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/DeployDataCenter__Apr_28_2016_06_35_56_T0QPTY/dc_entries.obj)
* [failed_plus_exceptions.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/DeployDataCenter__Apr_28_2016_06_35_56_T0QPTY/failed_plus_exceptions.txt)
* [runinfo.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/DeployDataCenter__Apr_28_2016_06_35_56_T0QPTY/runinfo.txt)

**`/tmp/MarvinLogs/test_network_IQO9KO:`**
* [failed_plus_exceptions.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/test_network_IQO9KO/failed_plus_exceptions.txt)
* [results.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/test_network_IQO9KO/results.txt)
* [runinfo.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/test_network_IQO9KO/runinfo.txt)

**`/tmp/MarvinLogs/test_vpc_routers_WQUMTV:`**
* [failed_plus_exceptions.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/test_vpc_routers_WQUMTV/failed_plus_exceptions.txt)
* [results.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/test_vpc_routers_WQUMTV/results.txt)
* [runinfo.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1365/tmp/MarvinLogs/test_vpc_routers_WQUMTV/runinfo.txt)


Uploads will be available until `2016-06-28 02:00:00 +0200 CEST`

*Comment created by [`upr comment`](https://github.com/cloudops/upr).*

swill:
These CI results basically show that the standard VM workflow in KVM is not affected.  I don't have the ability to test this on VMware, but I know you guys are already using this, yes?

I think I am pending one LGTM and this is ready...

rhtyd:
tag:easypr

rhtyd:
@agneya2001 @jburwell @DaanHoogland @sureshanaparti LGTM/review please, thanks

DaanHoogland:
did a code walk through and saw no extravaganzas and some useful extractions and renamings. As for the functionality, I have no clue. I am not a vmware user, let alone expert but for what it's worth: LGTM

rhtyd:
Thanks @DaanHoogland -- though there is a way to run ESXi6 with vCenter 6 on KVM (http://rohit.yadav.xyz/logs/vmware-esxi-vcenter-on-kvm-cloudstack)

/cc @swill this is ready for merge now, thanks

plugins/hypervisors/vmware/src/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -1363,24 +1364,15 @@ private Answer attachVolume(Command cmd, DiskTO disk, boolean isAttach, boolean
...
-                           (rootDiskControllerType == DiskControllerType.lsisas1068) ||
-                           (rootDiskControllerType == DiskControllerType.pvscsi) ||
-                           (rootDiskControllerType == DiskControllerType.buslogic)) {
-                    //TODO: Support mix of SCSI controller types for single VM. If root disk is already over
-                    //a SCSI controller then use the same for data volume as well. This limitation will go once mix
-                    //of SCSI controller types for single VM.
-                    dataDiskController = rootDiskController;
-                } else if (DiskControllerType.getType(dataDiskController) == DiskControllerType.osdefault) {
-                    dataDiskController = vmMo.getRecommendedDiskController(null);
+                String diskController = getLegacyVmDataDiskController();
```
swill: This has been merged into `4.7` and have forward merged into `4.8` without a problem.

I have merge conflicts trying to merge this from `4.8` into `master` though.

Here is the merge conflict:
```
            AttachAnswer answer = new AttachAnswer(disk);

            if (isAttach) {
<<<<<<< HEAD
                String dataDiskController = controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER);
                String rootDiskController = controllerInfo.get(VmDetailConstants.ROOT_DISK_CONTROLLER);
                DiskControllerType rootDiskControllerType = DiskControllerType.getType(rootDiskController);

                if (dataDiskController == null) {
                    dataDiskController = getLegacyVmDataDiskController();
                } else if ((rootDiskControllerType == DiskControllerType.lsilogic) ||
                           (rootDiskControllerType == DiskControllerType.lsisas1068) ||
                           (rootDiskControllerType == DiskControllerType.pvscsi) ||
                           (rootDiskControllerType == DiskControllerType.buslogic)) {
                    //TODO: Support mix of SCSI controller types for single VM. If root disk is already over
                    //a SCSI controller then use the same for data volume as well. This limitation will go once mix
                    //of SCSI controller types for single VM.
                    dataDiskController = rootDiskController;
                } else if (DiskControllerType.getType(dataDiskController) == DiskControllerType.osdefault) {
                    dataDiskController = vmMo.getRecommendedDiskController(null);
=======
                String diskController = getLegacyVmDataDiskController();
                if (controllerInfo != null &&
                        !Strings.isNullOrEmpty(controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER))) {
                    diskController = controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER);
>>>>>>> 4.8
                }
                if (DiskControllerType.getType(diskController) == DiskControllerType.osdefault) {
                    diskController = vmMo.getRecommendedDiskController(null);
                }
                vmMo.attachDisk(new String[] {datastoreVolumePath}, morDs, diskController);
            } else {
```

Please advise...

rhtyd: @swill as you can see the diff here, we need the latter block:

```
                String diskController = getLegacyVmDataDiskController();
                if (controllerInfo != null &&
                        !Strings.isNullOrEmpty(controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER))) {
                    diskController = controllerInfo.get(VmDetailConstants.DATA_DISK_CONTROLLER);
```

rhtyd: I'll fix the merge conflict

swill:
Obviously this merge conflict exists because something has been refactored in master which was not merged into previous versions.  We need to understand what was done and how to make sure both functionality can exist together...  I had to run to catch a train when I ran into this, so I have not gotten too deep into which PR was merged into master which is conflicting with this.

swill:
@rhtyd thanks for fixing the merge conflicts.  👍 

@GabrielBrascher and @mike-tutkowski, I know you guys have spent some time with this file, would you guys mind reviewing the approach that @rhtyd has taken to resolve the merge conflict and make sure you guys are happy with it?  I will also do a CI run against master before I do any additional merges to verify we are all good before we move on.

Thanks for the support on this guys...

mike-tutkowski:
@swill It seems reasonable to me. It doesn't look like equivalent functionality to what was previously there, but presumably that was why it was changed?

swill:
@mike-tutkowski, yes I think the change is expected.  I just want some more eyes on this since it did not go through the usual CI and code review channels because we were fixing a merge conflict, so I figured it was prudent to do a bit of an 'after the fact' code review to make sure everyone is comfortable with the current state.

I meant to CC @nvazquez as well since I know he has spent time with this file as well.

GabrielBrascher:
@swill I haven't worked much on that file (I had fixed a typo in the name of the "ROOT_DISK_CONTROLLER" variable).

At the first look I would agree with @mike-tutkowski, but I have a limited background on this file. I need to look with more care to be helpful.

swill:
No worries @GabrielBrascher, I realized after I CCed you that you really just fixed a typo.  :)  Thanks for the feedback...

swill:
Just as an update for everyone.  CI ran against master with the merge conflict fix and everything went smoothly, so we are all set.  Thanks for the help on this @rhtyd.

rhtyd:
Thanks @swill for verifying

PR #1370, opened 2016-01-27
[4.9+] WIP: Support Java 8, Maven 3.3, Upgraded dependencies

bhaisaab:
Was learning about JDK8, lambdas, streams etc today. Thought it would be a good idea to get ACS codebase to latest mvn and jdk8. This is a WIP branch pushed on ACS remote so feel free to push changes if you're a committer or send PR.

Major changes:
- Moved to latest JDK8
- Moved to latest Maven 3.3.9 (-T threaded compilations appears faster)
- Upgraded several dependencies to latest minor versions (or major in few cases)

What fails/unsupported or TODOs:
- Unit tests in cloud-server and few other places
- F5 plugin fails, tried to resolve the issue but I was unable to grok it
- Major spring framework version (that would be a lot of work on its own)
- Major servlet version changes
- Fix packaging, init.d/system.d JAVA_HOME paths
- Check ACS agents with JDK/JRE8 (kvm, cpvm, ssvm)
- Check and configure GC related changes wrt JRE7/JRE8 (in any configs, say tomcat configs)
- Check against compatibility guide: http://www.oracle.com/technetwork/java/javase/8-compatibility-guide-2156366.html

What works:
- Compiling in -Dnoredist with -DskipTests to test if ACS codebase (minus unit tests) worked (non-issue once we fix the failing unit tests)
- Deploy db worked
- Running the management server locally with Java8

### DIFF
--- a/engine/api/pom.xml
+++ b/engine/api/pom.xml
@@ -38,7 +38,7 @@
     <dependency>
       <groupId>org.apache.cxf</groupId>
       <artifactId>cxf-bundle-jaxrs</artifactId>
-      <version>2.7.13</version>
+      <version>2.7.18</version>
       <exclusions>
         <exclusion>
           <groupId>org.eclipse.jetty</groupId>
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ cache:
   directories:
   - $HOME/.cache
 jdk:
-- oraclejdk7
+- oraclejdk8
 notifications:
   email: false
 env:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,12 +14,12 @@ was tested against a CentOS 6.5 x86_64 setup.
 
 Install tools and dependencies used for development:
 
-    $ yum install git ant ant-devel java-1.6.0-openjdk java-1.6.0-openjdk-devel
-    mysql mysql-server tomcat6 mkisofs gcc python MySQL-python openssh-clients wget
+    $ yum install git java-1.8.0-openjdk java-1.8.0-openjdk-devel
+      mysql mysql-server tomcat6 mkisofs gcc python MySQL-python openssh-clients wget
 
     # yum -y update
-    # yum -y install java-1.7.0-openjdk
-    # yum -y install java-1.7.0-openjdk-devel
+    # yum -y install java-1.8.0-openjdk
+    # yum -y install java-1.8.0-openjdk-devel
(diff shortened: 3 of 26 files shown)


bhaisaab:
cc @DaanHoogland @jburwell @wido @wilderrodrigues @remibergsma  @borisroman and others

wido:
I agree with switching to Java 8, but shouldn't this be a small PR first where we ONLY switch to Java 8?

Other PRs can follow up quickly.

bhaisaab:
@wido I was getting some compilation issue specifically to F5, to which someone previously on ML had advised an issue related to dependency version so I've included upgrading dependency version (to stable/latest minor revision but no major version changes causing compatibility issues). Unfortunately, I did that but F5 was still not happy while getting compiled.

jburwell:
To ease transition, it seems like the first step should be to support running ACS built to 1.7 on a 1.8 runtime.  This approach would allow customers unprepared to transition to continue using Java7.  We would then notify users that Java7 support has been deprecated and will be removed in two releases (two months).

bhaisaab:
Alright, let's keep this PR hanging then. We should then open another PR which fixes the packaging first to switch from JRE7 to JRE8 along with any JAVA_OPT or tomcat config changes

jburwell:
@bhaisaab if we make the packaging require Java 1.8, then users will be automatically upgraded -- defeating the purpose of the deprecation.  It seems like an appropriate first step is to audit all dependencies for Java8 support, and upgrade all affected dependencies.  For example, we need to upgrade to Spring 4 because (Spring 3 does not support Java 8)[http://spring.io/blog/2013/05/21/spring-framework-4-0-m1-3-2-3-available/].  In my opinion, we should perform the dependency upgrades as part of 4.9.

engine/api/pom.xml
```diff
@@ -38,7 +38,7 @@
     <dependency>
       <groupId>org.apache.cxf</groupId>
       <artifactId>cxf-bundle-jaxrs</artifactId>
-      <version>2.7.13</version>
+      <version>2.7.18</version>
```
DaanHoogland: can the version become a var in the root pom?

bhaisaab: Sure.

Wondering if we can remove many of the engine/{projects}; not sure where it is used or who uses that. Looks like a WIP attempt to creating a RESTful api server.

PR #1381, opened 2016-01-28
Add missing license header to ActionEventUtilsTest.

ProjectMoon:
The test class was merged without the license header. This commit fixes that problem.

Also note that the license header exists on the master branch only as a result of commit 8a5fc16f. The commit seems to be on the master branch and the 4.7 branch only. So there may be some conflicts when forward merging.


ProjectMoon:
My god I swear GitHub's PR interface can die in a fire. Wrong base branch again.

PR #1384, opened 2016-01-29
Display hostname the VPC router runs on

remibergsma:
The hostname a router is running on is only displayed on Infra tab and not on the VPC page (the link local is mentioned though). That is now corrected, so you have the correct details to login to the router straight away.

Before:
screen shot 2016-01-29 at 20 58 58

After:
screen shot 2016-01-29 at 21 18 21

So the option 'Host' was added and displays the hypervisor the VPC runs on.

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -5859,6 +5859,9 @@
                                     linklocalip: {
                                         label: 'label.linklocal.ip'
                                     },
+                                    hostname: {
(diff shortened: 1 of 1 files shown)


borisroman:
LGTM :+1: Based on screenshots and code-review

Already running in PROD at SBP.

DaanHoogland:
convenient cos(mat)ic change: LGTM

PR #1387, opened 2016-01-30
CLOUDSTACK-8300: Set indexes on event table

remibergsma:
When there are many events in the cloud.event table, the UI throws an SQL exception and the management server spikes at 100% CPU for minutes. That causes other API calls to fail with 530 errors.

It seems an index on the 'archived' field is missing (since it is used in the WHERE clause).

We have 1.4M events in the table:
`select count(*) from event;`
1497838

This PR adds the index to the `archived` field and the `state` field.

### DIFF
--- a/setup/db/db/schema-481to490.sql
+++ b/setup/db/db/schema-481to490.sql
@@ -18,3 +18,6 @@
 --;
 -- Schema upgrade from 4.8.1 to 4.9.0;
 --;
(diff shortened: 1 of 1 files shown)


bhaisaab:
LGTM

wido:
Yes, simple but sane fix indeed. LGTM

PR #1388, opened 2016-01-30
CLOUDSTACK-7836: Make s2s VPN CIDR list longer

remibergsma:
We found the 200 char limit not sufficient. This should provide more room for longer CIDR lists.

123.123.123.123/24 ==> 18 chars + comma = 19 chars per cidr max

Was: 200/19 = ~10
Now: 4096/19 = ~215


DaanHoogland:
@remibergsma a better solution in the make?

PR #1391, opened 2016-02-02
Restore iptables at once using iptables-restore instead of calling iptables numerous times

borisroman:


borisroman:
New PR with additional fixes.

PR #1394, opened 2016-02-02
CLOUDSTACK-9268: Display VM in Load balancing rule in UI

nitin-maharana:
Steps of Repro:
=============
1:Create VMs 
2:Make LoadBalancing rule in GUI 
　Name:WWW 
　PrivatePort:80 
　PublicPort:80 
　Add VMs:some VMs

Expected Result:
==============
The VMs which has been already assigned is should not be listed when you add the VM to an existing rule.

Actual Result:
===========
The VMs which has been already assigned is still being listed when you add the VM to an existing rule.

Fix:
===
Added jsonObj to newly created row in multiedit.js to stop listing the same VM again.

### DIFF
--- a/ui/scripts/ui/widgets/multiEdit.js
+++ b/ui/scripts/ui/widgets/multiEdit.js
@@ -104,7 +104,7 @@
                         $(data).each(function() {
                             var item = this;
                             var $itemRow = _medit.multiItem.itemRow(item, options.itemActions, multiRule, $tbody);
-
+                            $itemRow.data('json-obj', item);
                             $itemRow.appendTo($tbody);
                             newItemRows.push($itemRow);
 


DaanHoogland:
@nitin-maharana can you add a screenshot of the intended change?

nitin-maharana:
Before Fix:
=========
before fix - nitin

After Fix:
=======
after fix - nitin

glennwagner:
LGTM - Tested and working correctly

DaanHoogland:
LGTM, code is trivial and functionality intuitive

koushik-das:
Merging as UI change and 2 LGTM.

swill:
@koushik-das I am not seeing the commit in master.  Did you merge into 4.7 and then forward merge into 4.8 and then forward merge into master???

koushik-das:
@swill I didn't fwd merge into 4.8 and master. Are we fwd merging for every individual commit in order to avoid conflicts?

swill:
Every commit that goes into 4.7 gets fwd-merged into 4.8 and then gets fwd-merged into master.  We have tools in `cloudstack/tools/git/` which make this process MUCH easier.  Would you like me to do it or do you want to try the tool and make sure you are comfortable with it?

koushik-das:
@swill I wanted to understand if it is ok to do these fwd merges once every week/fortnight (based on number of commits that went into the release branch) or do it after every commit.
If the merges can wait till tomorrow I can do it or else you can do it.

swill:
@koushik-das oh, i understand your question better now.  i would like to do it on every commit.  It makes the commit history clearer and it mitigates conflict issues.  when we ask people to rebase their PR we know for sure they are getting everything that is committed.  i would like for us to forward merge every time we do a commit to previous branches.

In this case I will do the forward merge because I have other commits I will merge today and if the forward merge is done right after the commit it makes it easier to verify the forward merge was done when scanning the commit history.

PR #1395, opened 2016-02-02
[4.7] VMware boot order fix

bhaisaab:
In case of VMware, since root disk needs to be first in the hdd boot order, we
collect list of hdds and create a hdd boot order with the root disk as the first
disk. This fixes a rare edge case where VMs would stop booting due to changed
boot order in hdds.

For sanity, we want cdroms to boot first, then hdds, floppies etc.

### DIFF
--- a/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -41,6 +41,7 @@ import java.util.UUID;
 
 import javax.naming.ConfigurationException;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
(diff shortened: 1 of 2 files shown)


DaanHoogland:
looks good, I cannot test this. Can you create a marvin test (or point one out) for this?

bhaisaab:
@DaanHoogland thanks Daan, this particular fix needs to be checked on both cloudstack and vcenter side so I'm not sure if a marvin tests exists or would be easy to write. It might be the case that this fix is not necessary (had sent an email on dev@ seeking advise on this), I'm waiting for other vmware devs who can advise or review -- otherwise we can close this.

PaulAngus:
I believe Abhi is going to be looking at Marvin soon.   We could maybe consider using the vmware python libraries (https://github.com/vmware/pyvmomi) when using Marvin to test vSphere specific code.

bhaisaab:
Figured it is not needed, closing for now

PR #1399, opened 2016-02-03
CLOUDSTACK-9272: No option in UI to add GSLB with service type "HTTP"

nitin-maharana:
Steps to Repro:
============
Go to Regions -> Local -> View GSLB -> Add GSLB
Click on the service type dropdown
Observe http is missing. Please see the attached snapshot.

Expected Behaviour:
================
As it supports http also, So http should be in the list.

Actual Behaviour:
==============
http is missing from the list.

Fix:
===
It supports http also. Added http to the list.

Snapshot:
========
gslb-http-missing-nitin

### DIFF
--- a/ui/scripts/regions.js
+++ b/ui/scripts/regions.js
@@ -334,6 +334,9 @@
                                             }, {
                                                 id: 'udp',
                                                 description: 'udp'
+                                            }, {
+                                                id: 'http',
+                                                description: 'http'
                                             }];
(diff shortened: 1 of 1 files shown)


glennwagner:
LGTM - HTTP added to GSLB

koushik-das:
LGTM - Verified on 4.7 that the dropdown shows http option.

PR #1405, opened 2016-02-08
CLOUDSTACK-9280: Allow system VM volumes to be expunged when there are no system VMs remaining

ProjectMoon:
This pull request is our proposed fix for https://issues.apache.org/jira/browse/CLOUDSTACK-9280. I added a new special SSVM endpoint that happily accepts any command given to it. This endpoint is used in only a very specific scenario:

* The volume's VM is in state destroyed or expunging, but the volume still lingers.
* The volume's VM is a system VM (SSVM or console proxy).
* There are no secondary storage machines existing in the volume's zone.

This necessitated a small change to VolumeObject which allows it to find removed VMs (`findByIdIncludingRemoved`). The main part of the work is in the DefaultEndpointSelector.

We would like some thorough review of this PR as well as what tests to create/run. I'm not sure if the scope of this fix will lead to unintentional behavior changes in other scenarios.


ProjectMoon:
Wrong base branch.

PR #1407, opened 2016-02-09
4.8.0 fix i18n es

ntavares:
Hi all,

I checked this issue yesterday with Daan, the problem is double UNICODE encoding, and it's the source in Transifex that is broken. There are ~250 broken characters, so fixing one by one in Transifex is a bit of a waste of time. I could fix this with a script, but I'm still wondering how the file received patches and remain in this state (as in: is there any other scenario where this double encoding shows up correct?)

I've created a PR based on the tag 4.8.0.

-NT


ntavares:
I also spotted what could be an error in the messages DB:

label.snapshot.s=Snapshots
label.snapshots=Snapshots

Looks like a typo to me, maybe you want to merge those two.

milamberspace:
Tested on 4.8 snapshot branch. The Web UI with Spanish language works well.

LGTM

DaanHoogland:
skimmed the code to check that indeed double encodings are gone. Not being spanish I can not judge the capitalisations. These would look silly in dutch. LGTM

ntavares:
@Daan, yes the capitalizations are wrong in several places.... I've fixed
the most visible (home page), but that's another matter.
As for the double encoding, there were only two strings missing which
include a upside-down question mark. Spanish use that a lot :) I've fixed
these also on the PR.

-NT

-- 
--
Nuno Tavares

rhtyd:
@ntavares please rebase against latest master
we'll need blessing from @milamberspace on this one

jburwell:
@ntavares am I correct in understanding that this PR fixes a bug in 4.8?  If so, could you please rebase and retarget this PR to the 4.8 branch?  The fix will be forward merged to 4.9 and master.

rhtyd:
@ntavares  rebase against latest master, change base branch please. This may not be applicable anymore.

wido:
This Pull Requests just got the label 'closable' for not seeing any activity for a long period.

The CloudStack project is in a transition from the Apache Foundation's Git infrastructure to Github. This has created a very large backlog of PRs. A lot of them don't merge anymore and didn't get attention.

To filter these PRs we are adding the tag closable to them to make a distinction between PRs which need our attention and which are a candidate to be closed.

It's not our intention to say that we don't value the PR, but it's a way to get a overview of which PRs need our resources and attention.

If you think closing this PR is a mistake, please add a comment and let us know! If you do that, could you please make sure that the PR merges against the branch it was submitted against?

Thank you very much for your understanding and cooperation!

rhtyd:
Thanks @ntavares for your PR. The latest master has moved to json/js based translation and I checked to find several fixes/changes already in master. On this note, I would thank you again and close this. You may re-submit this for current master changing the js i18n files as messages.properties files don't exist anymore.

PR #1412, opened 2016-02-12
Add lsb-release dependency to agent and mgmt server on Debian/Ubuntu.

ProjectMoon:
The cloudstack-setup-management script needs the lsb_release command. This command is provided by the `lsb-release` package on Debian and Ubuntu, and this is not always installed by default. So, it should be added as a dependency to cloudstack-management.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,14 +15,14 @@ Description: A common package which contains files which are shared by several C
 
 Package: cloudstack-management
 Architecture: all
-Depends: ${misc:Depends}, ${python:Depends}, cloudstack-common (= ${source:Version}), tomcat6, sudo, jsvc, python-mysqldb, libmysql-java, augeas-tools, mysql-client, adduser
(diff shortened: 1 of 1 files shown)


bhaisaab:
LGTM, simple change

ProjectMoon:
Apparently the build of this pull request sucked Jenkins into a black hole. The build returns 404.

ProjectMoon:
I've updated this pull request @bhaisaab. The dependency needs to be added to cloudstack-agent as well due to a similar problem with the cloudstack-setup-agent script (they both depend on the same Python libraries).

swill:
What do we do about this since it is against the 4.6 branch?  Do you want to close this and open it against the 4.7 branch?

ProjectMoon:
Since it's a bug fix, according to the current CS workflow it should be put on the oldest supported release branch and then forward merged. I can resubmit it against 4.7 if need be though.

kiwiflyer:
@ProjectMoon Can you open a new PR against 4.7 with this?  We're really only supporting 4.7.x onward.
Trivial packing change. LTGM

rhtyd:
@ProjectMoon please close this and open against 4.7+, thanks

tag:closeme

ProjectMoon:
Closing to open new PR.

PR #1418, opened 2016-02-19
CLOUDSTACK-9261: Query to traffic sentinel requesting for usage stats…

kansal:
… is too long resulting in traffic sentinel sending HTTP 414 error response - Fixed

Root Cause Analysis:
In the current implementation, we are using GET method to send the request tom the traffic sentinel instance. Thus, if there are many IP's that need to be queried, the request becomes too long and the HTTP Error 414 is thrown. 

Fix: Changed the GET request to the POST request.

### DIFF
--- a/core/src/com/cloud/network/resource/TrafficSentinelResource.java
+++ b/core/src/com/cloud/network/resource/TrafficSentinelResource.java
@@ -22,6 +22,7 @@ package com.cloud.network.resource;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.io.OutputStream;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLEncoder;
@@ -53,6 +54,7 @@ import com.cloud.agent.api.StartupTrafficMonitorCommand;
 import com.cloud.host.Host;
 import com.cloud.resource.ServerResource;
 import com.cloud.utils.exception.ExecutionException;
+import java.net.HttpURLConnection;
 
 public class TrafficSentinelResource implements ServerResource {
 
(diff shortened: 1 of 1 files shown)


alexandrelimassantana:
Hi @kansal, can you make a test with this change? Try issuing an request with more than 2048 characters to see if everything still works fine then.

kansal:
@alexandrelimassantana , I am not sure of how to simulate the actual traffic sentinel resource. I can create a request with 20148 or more, but the actual error is encountered when the request reaches the resource.

GabrielBrascher:
@kansal Could you please consider the following adjustments?
- create a method for the "Part 1" (lines 213-219). With the command as parameter and returning the connection to be used for "Part 2" (lines 221-223);
- create a method for "Part 2", returning the output stream and receiving as parameter the connection and post data.

I think that with these two methods, the code would be more readable.

Thanks.

kansal:
@GabrielBrascher These adjustments can be done, but I don't see any point of introducing the function calls. Anyhow this code is not used anywhere else in the whole project. As far as readability is concerned, I think with the proper comments it should be just fine.

kishankavala:
@kansal 
LGTM. See if you can add an integration test.  
Actual response from traffic sentinel resource in the logs will also help.

GabrielBrascher:
@kansal create methods would also help to test the behavior of this code, using unit test and integration test (to verify if the methods that have unit test are called in the correct order by a more generic method).

For example, you could create a method `queryTrafficSentinel`, this method calls the part 1, 2 and 3 in this exact sequence and one time each (you can use [InOrder](https://mockito.googlecode.com/hg-history/1.6/javadoc/org/mockito/InOrder.html) to ensure the order and times that a method was called). Methods that would do the part 1, 2 and 3 could have unit tests.

At the end, you could create a Javadoc for those methods using all four comment blocks that you created.

Thanks.

rhtyd:
@kansal please rebase against latest master

tag:needlove

kansal:
@rhtyd Rebased.

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 148
 Hypervisor xenserver
 NetworkType Advanced
 Passed=70
 Failed=1
 Skipped=3

**Failed tests:**
* test_service_offerings.py

 * ContextSuite context=TestServiceOfferings>:setup Failing since 8 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vpc_vpn.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

PR #1419, opened 2016-02-19
systemvm: preserve file permissions, set default umask

bhaisaab:
- In injectkeys.sh which is used to inject new public keys everytime cloudstack
  starts; while copying files preserve the mode/ownership. This ensures the
  scripts have same mode bits as originally configured in the iso file
- The default umask of 0022 is set in Ubuntu and other packages. Set the same
  in case of CentOS startup scripts

cc @abhinandanprateek @wido @remibergsma @jburwell @DaanHoogland

### DIFF
--- a/packaging/centos63/cloud-management.rc
+++ b/packaging/centos63/cloud-management.rc
@@ -87,6 +87,7 @@ handle_pid_file() {
 }
 
 start() {
+    umask 0022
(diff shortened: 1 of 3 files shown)


bhaisaab:
This needs to be on all 4.7+ closing & reopening

PR #1422, opened 2016-02-21
Improve ordering of fields of VPC router detail tab

remibergsma:
The field we use most are now on the top:
- name
- state
- hypervisor
- link local ip
- redundant state

The other fields are nice, but not needed most of the time.

Result:
screen shot 2016-02-21 at 20 11 49

Before:
vpc_router_details_before

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -5838,35 +5838,14 @@
                                         label: 'label.name'
                                     }
                                 }, {
-                                    id: {
-                                        label: 'label.id'
-                                    },
-                                    zonename: {
-                                        label: 'label.zone'
-                                    },
-                                    dns1: {
-                                        label: 'label.dns'
-                                    },
-                                    gateway: {
-                                        label: 'label.gateway'
-                                    },
-                                    publicip: {
-                                        label: 'label.public.ip'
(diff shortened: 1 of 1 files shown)


DaanHoogland:
I agree with this at first sight. so LGTM but is this debatable? that would inspire a requirement to have it conigurable or so, no? maybe this is not the place for that discussion but that doesn't keep me from starting it.

remibergsma:
@DaanHoogland I debugged so many routers recently that I think it's fair to say this info on the top is most valuable :-) But feel free to disagree.

DaanHoogland:
@remibergsma sounds like you got the tone of comment just right.

glennwagner:
Made Simple - LGTM - Can we look at making these Fields configurable?

kiwiflyer:
I agree that it would be nice for this to be configurable (in the future). But I will say that those 5 fields are also the ones we use religiously when in VR Hell ;-)

LGTM.

DaanHoogland:
@swill this is a well tested UI change with several LGTM, can you(we) merge?

swill:
Yes, I think this is good.  I will add it to the list of merges I will do later today.

PR #1426, opened 2016-03-02
ADD be explicit about the underlying limitation - OpenSwan

ntavares:
According to:
http://linux.die.net/man/5/ipsec.secrets

A preshared secret is most conveniently represented as a sequence of characters, delimited by the double-quote character ("). The sequence cannot contain a newline or double-quote.


GabrielBrascher:
It seems a fair update.
LGTM due to the lack of code.

DaanHoogland:
LGTM

kiwiflyer:
LGTM

rafaelweingartner:
merging giving that it is a change only in a description of an enum and that we already have 2+ LGTM

PR #1427, opened 2016-03-02
findbugs: new found old instance of DM_DEFAULT_ENCODING fixed

DaanHoogland:
this quick fix to silence the findbugs
   which seems to think this is a new issue

### DIFF
--- a/services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
+++ b/services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
@@ -16,8 +16,9 @@
 // under the License.
 package org.apache.cloudstack.storage.resource;
 
-import static com.cloud.utils.storage.S3.S3Utils.putFile;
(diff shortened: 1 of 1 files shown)


swill:
It looks like your IDE had a field day with this class.  So other than the reordering of imports, adding `final` to a ton of variables and removing some unneeded brackets, this is really the only change?  Right?

`
-        String dir = UUID.nameUUIDFromBytes(nfsPath.getBytes()).toString();
+        final String dir = UUID.nameUUIDFromBytes(nfsPath.getBytes(getPreferredCharset())).toString();
`

DaanHoogland:
:) @swill it is the only part I typed

swill:
Haha.  Fair enough.  I will let Raja test, but otherwise it LGTM.

pvr9711:
LGTM
Ran BVT  with XS 6.5  - all the test passed with the exception of the following tests due to env issues:
test_01_create_volume - ssh issue.  
test_07_list_default_iso - iso didn't exist.  
test_01_test_vm_volume_snapshot - test case issue - operation is not allowed?

DaanHoogland:
#1438 covers this

PR #1438, opened 2016-03-13
Fix new error found in findbugs slow build #3455

rafaelweingartner:
Fix new find bug error that was introduced in PR #1361 
Report: http://jenkins.buildacloud.org/job/build-master-slowbuild/3455/findbugsResult/new/

It is the same fix as the one proposed in #1427; the difference is that this PR tried to change only the code that was strictly needed. However, I took the liberty to remove a dead code and few lines of code (annotations) that were not needed

### DIFF
--- a/services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
+++ b/services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
@@ -16,8 +16,8 @@
 // under the License.
 package org.apache.cloudstack.storage.resource;
 
-import static com.cloud.utils.storage.S3.S3Utils.putFile;
 import static com.cloud.utils.StringUtils.join;
+import static com.cloud.utils.storage.S3.S3Utils.putFile;
 import static java.lang.String.format;
 import static java.util.Arrays.asList;
 import static org.apache.commons.lang.StringUtils.substringAfterLast;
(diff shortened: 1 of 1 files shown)


services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
```diff
@@ -395,16 +391,9 @@ protected Answer copyFromS3ToNfs(CopyCommand cmd, DataTO srcData, S3TO s3, DataT
                     return new CopyCmdAnswer(errMsg);
                 }
             }
-
             File destFile = new File(downloadDirectory, substringAfterLast(srcData.getPath(), S3Utils.SEPARATOR));
-
             S3Utils.getFile(s3, s3.getBucketName(), srcData.getPath(), destFile).waitForCompletion();
 
-
-            if (destFile == null) {
```
rafaelweingartner: I removed this because there was no way this variable could be null

pvr9711:
@swill @DaanHoogland  tested this PR - LGTM.  
All BVTs have passed with the exception of the following tests
- test_07_list_default_iso - iso didn't exist.
- test_01_test_vm_volume_snapshot - test case issue - operation is not allowed?

Raja

services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
swill: Is this because `new File(...)` will always throw an exception if the value would be `null`?

swill:
I don't think these tests failing are relevant to this change.  We agree this is ready to be merged?  Only against master?

services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
rafaelweingartner: I removed that because the new File (...) will always create an object that is referenced by destFile; therefore, it will never be == null.

nvazquez:
Code LGTM

rafaelweingartner:
Tests were performed and reviews executed; so, should we merge this or the #1427?

swill:
I think we should probably merge this one and not #1427.  Do you agree?

rafaelweingartner:
I agree, but I was the one that opened the PR, so in my opinion, my opinion should no count much here :)

DaanHoogland:
I opened #1427 and I agree, @rafaelweingartner and @swill so please go ahead. I will close #1427

PR #1439, opened 2016-03-13
Changed Void Methods to void

rodrigo93:
I have changed many "Void" methods to "void" since they have to return
null at the end and it has the same effect of "void" methods.
The method _HandleVolumeCreateAsyncCallback_ cannot be void, so I let it as Void.

### DIFF
--- a/engine/storage/image/src/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
+++ b/engine/storage/image/src/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
@@ -589,7 +589,7 @@ public class TemplateServiceImpl implements TemplateService {
         return null;
     }
 
-    protected Void createTemplateCallback(AsyncCallbackDispatcher<TemplateServiceImpl, CreateCmdResult> callback, TemplateOpContext<TemplateApiResult> context) {
+    protected void createTemplateCallback(AsyncCallbackDispatcher<TemplateServiceImpl, CreateCmdResult> callback, TemplateOpContext<TemplateApiResult> context) {
         TemplateObject template = context.getTemplate();
         AsyncCompletionCallback<TemplateApiResult> parentCallback = context.getParentCallback();
         TemplateApiResult result = new TemplateApiResult(template);
@@ -600,9 +600,7 @@ public class TemplateServiceImpl implements TemplateService {
             if (parentCallback != null) {
                 parentCallback.complete(result);
             }
-            return null;
         }
-
         try {
             template.processEvent(ObjectInDataStoreStateMachine.Event.OperationSuccessed);
         } catch (Exception e) {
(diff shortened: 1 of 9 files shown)


rodrigo93:
Squashed last 2 commits.

rafaelweingartner:
@rodrigo93 you removed the return type Object (Void) to void; therefore, you have to fix those methods usage, such as the one at line 202 ( caller.setCallback(caller.getTarget().createTemplateAsyncCallBack(null, null));) of  classHypervisorTemplateAdapter.

There is no return of method "createTemplateAsyncCallBack", so you cannot use its return to set somewhere else.

syed:
@rodrigo93 Do you know why we have `Void` return types? I've seen them and not sure why they were added.

rodrigo93:
Sorry for the delay to answer.
Hi @rafaelweingartner , thanks for the suggestion. I will revise my PR and do the proper changes.

@syed I am not sure why we Void types as well. I know that Void is used for reflections and when generic types is needed. Someone correct me if I am wrong.

rafaelweingartner:
@rodrigo93 have you built this PR successfully?

wido:
This Pull Requests just got the label 'closable' for not seeing any activity for a long period.

The CloudStack project is in a transition from the Apache Foundation's Git infrastructure to Github. This has created a very large backlog of PRs. A lot of them don't merge anymore and didn't get attention.

To filter these PRs we are adding the tag closable to them to make a distinction between PRs which need our attention and which are a candidate to be closed.

It's not our intention to say that we don't value the PR, but it's a way to get a overview of which PRs need our resources and attention.

If you think closing this PR is a mistake, please add a comment and let us know! If you do that, could you please make sure that the PR merges against the branch it was submitted against?

Thank you very much for your understanding and cooperation!

engine/storage/image/src/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
```diff
@@ -687,7 +681,7 @@ public Void deleteTemplateCallback(AsyncCallbackDispatcher<TemplateServiceImpl,
         return future;
     }
 
-    protected Void syncTemplateCallBack(AsyncCallbackDispatcher<TemplateServiceImpl, CopyCommandResult> callback, TemplateOpContext<TemplateApiResult> context) {
+    protected void syncTemplateCallBack(AsyncCallbackDispatcher<TemplateServiceImpl, CopyCommandResult> callback, TemplateOpContext<TemplateApiResult> context) {
```
fmaximus: Changing this creates compilation errors in caller.setCallback(...) in deleteTemplateAsync.
So the reason that Void is used, is that the response is passed to a method expecting an Object as parameter.

rhtyd:
Thanks for your contribution. I'm closing this on the remark that there is valid usage for the refactored/removed code.

PR #1441, opened 2016-03-16
CLOUDSTACK-9297 - Reworked logic in StorageSystemSnapshotStrategy and XenserverSnapshotStrategy

mike-tutkowski:
The ticket this PR fixes was opened because KVM-specific code had been added to the StorageSystemSnapshotStrategy class and that class' canHandle method was only prepared to handle managed storage being used with XenServer (and a case was hit for KVM that triggered a CloudRuntimeException to be thrown).

To solve the problem, I moved the KVM logic to the default snapshot strategy class, which is (unfortunately) named XenserverSnapshotStrategy.

I plan to rename XenserverSnapshotStrategy to something like DefaultSnapshotStrategy in 4.9.

My guess is that when XenserverSnapshotStrategy was originally written, it was written only for XenServer, but has since that time had its usage increased to support other hypervisors (with non-managed storage).

### DIFF
--- a/engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
@@ -35,12 +35,10 @@ import org.apache.cloudstack.engine.subsystem.api.storage.SnapshotResult;
 import org.apache.cloudstack.engine.subsystem.api.storage.StrategyPriority;
 import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
 import org.apache.cloudstack.engine.subsystem.api.storage.VolumeService;
-import org.apache.cloudstack.engine.subsystem.api.storage.ObjectInDataStoreStateMachine.Event;
 import org.apache.cloudstack.storage.command.SnapshotAndCopyAnswer;
 import org.apache.cloudstack.storage.command.SnapshotAndCopyCommand;
 import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
 import org.apache.cloudstack.storage.datastore.db.SnapshotDataStoreDao;
-import org.apache.cloudstack.storage.datastore.db.SnapshotDataStoreVO;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
 
 import com.cloud.agent.AgentManager;
@@ -57,8 +55,6 @@ import com.cloud.storage.DataStoreRole;
 import com.cloud.storage.Snapshot;
 import com.cloud.storage.SnapshotVO;
 import com.cloud.storage.Storage.ImageFormat;
-import com.cloud.storage.StoragePool;
-import com.cloud.storage.StoragePoolStatus;
 import com.cloud.storage.Volume;
 import com.cloud.storage.VolumeVO;
 import com.cloud.storage.dao.SnapshotDao;
@@ -157,43 +153,7 @@ public class StorageSystemSnapshotStrategy extends SnapshotStrategyBase {
 
     @Override
     public boolean revertSnapshot(SnapshotInfo snapshot) {
-        if (canHandle(snapshot,SnapshotOperation.REVERT) == StrategyPriority.CANT_HANDLE) {
-            throw new UnsupportedOperationException("Reverting not supported. Create a template or volume based on the snapshot instead.");
-        }
-
-        SnapshotVO snapshotVO = _snapshotDao.acquireInLockTable(snapshot.getId());
-        if (snapshotVO == null) {
-            throw new CloudRuntimeException("Failed to get lock on snapshot:" + snapshot.getId());
-        }
-
-        try {
-            VolumeInfo volumeInfo = snapshot.getBaseVolume();
-            StoragePool store = (StoragePool)volumeInfo.getDataStore();
-            if (store != null && store.getStatus() != StoragePoolStatus.Up) {
-                snapshot.processEvent(Event.OperationFailed);
-                throw new CloudRuntimeException("store is not in up state");
-            }
-            volumeInfo.stateTransit(Volume.Event.RevertSnapshotRequested);
-            boolean result = false;
-            try {
-                result =  snapshotSvr.revertSnapshot(snapshot);
-                if (! result) {
-                    s_logger.debug("Failed to revert snapshot: " + snapshot.getId());
-                    throw new CloudRuntimeException("Failed to revert snapshot: " + snapshot.getId());
-                }
-            } finally {
-                if (result) {
-                    volumeInfo.stateTransit(Volume.Event.OperationSucceeded);
-                } else {
-                    volumeInfo.stateTransit(Volume.Event.OperationFailed);
-                }
-            }
-            return result;
-        } finally {
-            if (snapshotVO != null) {
-                _snapshotDao.releaseFromLockTable(snapshot.getId());
-            }
-        }
+        throw new UnsupportedOperationException("Reverting not supported. Create a template or volume based on the snapshot instead.");
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


ustcweizhou:
tested on KVM, it is good.
so XenserverSnapshotStrategy LGTM.

mike-tutkowski:
Thanks, @ustcweizhou!

mike-tutkowski:
Ideally, at some point in the future, we should rename XenserverSnapshotStrategy to something more generic (maybe DefaultSnapshotStrategy or something).

mike-tutkowski:
Perhaps I'll rename it in 4.9 (probably not a good idea to do so in 4.7 and then move the name change up to 4.8 and 4.9).

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/XenserverSnapshotStrategy.java
```diff
@@ -279,7 +282,52 @@ public boolean deleteSnapshot(Long snapshotId) {
 
     @Override
     public boolean revertSnapshot(SnapshotInfo snapshot) {
-        throw new CloudRuntimeException("revert Snapshot is not supported");
+        if (canHandle(snapshot,SnapshotOperation.REVERT) == StrategyPriority.CANT_HANDLE) {
```
rafaelweingartner: What about changing this "canHandle" method to return a boolean if the snapshot can be handled, instead of a enum value?

rafaelweingartner:
@mike-tutkowski  could you fix the PR title and improve its description (a brief description of the problem and solution would do)?

Additionally, what about some unit tests to "canHandle" method? I also recommend to change that method return, a boolean return is more intuitive in my opinion; I also believe that a Java doc to that method would be appreciated.

mike-tutkowski:
Hi @rafaelweingartner Thanks for the comments.

Just wanted to point out that if the revertSnapshot method is invoked, then that is actually a failure of the canHandle method to inform the calling code to not call the revertSnapshot method, so that's why the decision was made to throw an exception.

I see returning true as your revert was successful and returning false as it failed (but it was OK for you to ask us to revert because canHandle properly told you that you could do so).

rafaelweingartner:
@mike-tutkowski thanks for the improvements on PR title and description. 
I understand the point of throwing an exception and I really like that approach.

About the suggested change on “canHandle” method, I still did not see the need to use the enum. If you were using more than 2 values as its returns, it would be something I would understand; but you only use two values that could be represented as a “true” or “false”.

Could you elaborate a little bit more?
I also did not understand this sentence:
“I see returning true as your revert was successful and returning false as it failed (but it was OK for you to ask us to revert because canHandle properly told you that you could do so)”

mike-tutkowski:
The canHandle method is used by CloudStack's snapshot infrastructure to determine which snapshot strategy to use.

For example, StorageSystemSnapshotStrategy returns either StrategyPriority.HIGHEST or StrategyPriority.CANT_HANDLE whereas XenserverSnapshotStrategy returns StrategyPriority.DEFAULT or StrategyPriority.CANT_HANDLE.

That being the case, there are at least three options in play (and other people can write custom snapshot strategies that they don't even check into the CloudStack repo, so we can't easily just change the interface where canHandle is declared).

As for this sentence:

“I see returning true as your revert was successful and returning false as it failed (but it was OK for you to ask us to revert because canHandle properly told you that you could do so)”

That was in regards to the revertSnapshot method. It sounds like you agree with the exception being thrown there.

rafaelweingartner:
I understand that your point that some other classes can overwrite that method and may return other elements of the enum. However, I do not see anything else in the code that used the “canHandle” method return that uses the returned value. I mean, it is used in the “revertSnapshot” method at line 285, but it only checks if the return is equal to “StrategyPriority.CANT_HANDLE” and then throws an exception if it is. To me, that is the same as if the method were returning a boolean value.

mike-tutkowski:
Maybe we're looking at different code?

I see the canHandle method being used from StorageStrategyFactoryImpl.getSnapshotStrategy.

It has a method, bestMatch, that it makes use of to pick a snapshot strategy with the ordinal value.

rafaelweingartner:
Ok @mike-tutkowski, now I believe I understand a little bit more of the code.

You are reusing a method that already exist and returns the StrategyPriority (you did not create that method to suffice your code). I was misled by the name of the method, when I first saw it, I thought that the method was used solely to check if it can or cannot proceed with the snapshot when in fact it is used to do much more.

I believe "getSnapshotStrategy" would be a more suitable name.

Now that I understand what we have there, I am ok with that conditional as it
 is.

mike-tutkowski:
CloudStack iterates over all known snapshot strategies and asks each if it can handle a particular situation (take, delete, revert) for a volume or volume snapshot.

One with a priority higher than the others is selected.

If you can't handle the particular use case, you return CANT_HANDLE.

On Mar 17, 2016, at 12:49 PM, Rafael Weing?rtner <notifications@github.com<mailto:notifications@github.com>> wrote:


Ok @mike-tutkowski<https://github.com/mike-tutkowski>, now I believe I understand a little bit more of the code.

You are reusing a method that already exist and returns the StrategyPriority (you did not create that method to suffice your code). I was misled by the name of the method, when I first saw it, I thought that the method was used solely to check if it can or cannot proceed with the snapshot when in fact it is used to do much more.

I believe "getSnapshotStrategy" would be a more suitable name.

Now that I understand what we have there, I am ok with that conditional as it
is.

-

rafaelweingartner:
got it @mike-tutkowski

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
```diff
@@ -440,41 +400,28 @@ private void markAsBackedUp(SnapshotObject snapshotObj) {
...
     public StrategyPriority canHandle(Snapshot snapshot, SnapshotOperation op) {
-        long volumeId = snapshot.getVolumeId();
-        VolumeVO volumeVO = _volumeDao.findById(volumeId);
         if (SnapshotOperation.REVERT.equals(op)) {
-            if (volumeVO != null && ImageFormat.QCOW2.equals(volumeVO.getFormat()))
-                return StrategyPriority.DEFAULT;
-            else
-                return StrategyPriority.CANT_HANDLE;
+            return StrategyPriority.CANT_HANDLE;
         }
```
alexandrelimassantana: Good work on rewritting the logic to a slimmer state.

Wouldn't be it more readable if you turn lines 407-422 into a boolean function which would return true if "supportsStorageSystemSnapshots". Snap this chunk of code could prove useful on creating unit tests and documentation as this segment clearly aims to check this condition solely.

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/XenserverSnapshotStrategy.java
DaanHoogland: @rafaelweingartner if you consider changing a canHandle() method take care. These are called over all types of resources and should not break contract as to not confuse any Orchestrators in the system.

rafaelweingartner: Hi @DaanHoogland, this was only my first doubt (question). After that, @mike-tutkowski  and I exchanged some other messages and then I understood the method. I see why a boolean return would not suffice all of its needs.

I was misled by the name of the method. Every time I see a method name that starts with "can/has/is", I think that it will return a boolean value. 

The code of this PR is OK for me; I would only suggest a name change on that "canHandle" method to a more descriptive one.

swill:
I will get CI run against this PR as soon as I have my CI reinstalled.  Thanks @mike-tutkowski

mike-tutkowski:
Thanks, Will!

On Apr 5, 2016, at 12:35 PM, Will Stevens <notifications@github.com<mailto:notifications@github.com>> wrote:


I will get CI run against this PR as soon as I have my CI reinstalled. Thanks @mike-tutkowski<https://github.com/mike-tutkowski>

-

swill:
## CI RESULTS

### 84/85 TESTS PASSED

The only failed test is a test that often fails in my environment and is unrelated to this change.


**Associated Uploads**

**`test_network_2NUDW1`:**

* [results.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1441/test_network_2NUDW1/results.txt)

**`test_vpc_routers_CHF744`:**

* [results.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1441/test_vpc_routers_CHF744/results.txt)


Uploads will be available until `2016-06-07 02:00:00 +0200 CEST`

*Comment created by [`upr comment`](https://github.com/swill/upr).*

swill:
I think this is ready pending a code review.  Can someone who has spent some time reviewing this PR with Mike please give me the go ahead on this PR.  @rafaelweingartner, you have been the most active reviewing this PR, are you happy with the state of this PR?

DaanHoogland:
@swill you have a failure, why do you want an LGTM, seems to me it didn't pass the tests

swill:
@DaanHoogland I have some issues with networking when using redundant VRs in my environment.   I consistently have this type of issue, and it is unrelated to the actual code.  If it is not clear if the issue is the environment or the PR, I rerun the CI to validate again.  In this case I am confident it is my environment.  I was experimenting with running 3 stacks all in parallel and have found that is not realistic.

rafaelweingartner:
@swill, all of my enquiries here were answered. My only complaint was about the method name "canHandle"; it misled me during my first and second look at this PR.I would prefer something like "getStorageStrategyPriority".

The code looks good to me though. LGTM ;)

mike-tutkowski:
Thanks for the comments @rafaelweingartner and @ustcweizhou.

Since canHandle is part of the infrastructure, we decided this particular PR isn't the place to change such a thing. Perhaps a future PR.

rafaelweingartner:
@mike-tutkowski, @swill I am ok with that

swill:
It looks like I have two LGTM on this PR and the tests are passing.  This one is ready?  Any final comments?

rafaelweingartner:
@swill no final comments ;)

mike-tutkowski:
It's all set to go. :-)

On Apr 8, 2016, at 7:09 AM, Rafael Weing?rtner <notifications@github.com<mailto:notifications@github.com>> wrote:


@swill<https://github.com/swill> no final comments ;)

-

PR #1451, opened 2016-03-23
CLOUDSTACK-9319: Use timeout when applying config to virtual router

insom:
From the [JIRA issue](https://issues.apache.org/jira/browse/CLOUDSTACK-9319):

> The timeout parameter is not passed down to `applyConfigToVR` inside `VirtualRoutingResource` in all cases.
> 
> This timeout is worked out as 3 seconds per command or 120 seconds (whichever is larger), but because it's not passed to the first invocation, the default (120 seconds, DEFAULT_EXECUTEINVR_TIMEOUT) is used.
> 
> In a recent upgrade of our Virtual Routers, the timeout was being hit and increasing `router.aggregation.command.each.timeout` had no effect. I built a custom 4.8 agent with the timeout increased to allow the upgrade to continue.

### DIFF
--- a/core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
+++ b/core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
@@ -152,11 +152,10 @@ public class VirtualRoutingResource {
         }
     }
 
-    private ExecutionResult applyConfigToVR(String routerAccessIp, ConfigItem c) {
-        return applyConfigToVR(routerAccessIp, c, VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT);
-    }
-
     private ExecutionResult applyConfigToVR(String routerAccessIp, ConfigItem c, int timeout) {
+        if (timeout < VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT) {
+            timeout = VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT;
+        }
         if (c instanceof FileConfigItem) {
             FileConfigItem configItem = (FileConfigItem)c;
             return _vrDeployer.createFileInVR(routerAccessIp, configItem.getFilePath(), configItem.getFileName(), configItem.getFileContents());
@@ -180,7 +179,7 @@ public class VirtualRoutingResource {
         boolean finalResult = false;
         for (ConfigItem configItem : cfg) {
             long startTimestamp = System.currentTimeMillis();
-            ExecutionResult result = applyConfigToVR(cmd.getRouterAccessIp(), configItem);
+            ExecutionResult result = applyConfigToVR(cmd.getRouterAccessIp(), configItem, VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT);
             if (s_logger.isDebugEnabled()) {
                 long elapsed = System.currentTimeMillis() - startTimestamp;
                 s_logger.debug("Processing " + configItem + " took " + elapsed + "ms");
@@ -375,9 +374,6 @@ public class VirtualRoutingResource {
                 ScriptConfigItem scriptConfigItem = new ScriptConfigItem(VRScripts.VR_CFG, "-c " + VRScripts.CONFIG_CACHE_LOCATION + cfgFileName);
                 // 120s is the minimal timeout
                 int timeout = answerCounts * _eachTimeout;
-                if (timeout < 120) {
-                    timeout = 120;
-                }
 
                 ExecutionResult result = applyConfigToVR(cmd.getRouterAccessIp(), fileConfigItem, timeout);
                 if (!result.isSuccess()) {
--- a/core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
+++ b/core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
@@ -379,7 +379,7 @@ public class VirtualRoutingResource {
                     timeout = 120;
                 }
 
-                ExecutionResult result = applyConfigToVR(cmd.getRouterAccessIp(), fileConfigItem);
+                ExecutionResult result = applyConfigToVR(cmd.getRouterAccessIp(), fileConfigItem, timeout);
                 if (!result.isSuccess()) {
                     return new Answer(cmd, false, result.getDetails());
                 }


DaanHoogland:
@insom any chance you can supply some test-mech?

alexandrelimassantana:
@insom with this changes there is still one place where the applyConfigToVR is not provided with this timeout parameter (line 183, same class). This function is actually called only in those 2 places. If it comes to a state where you will always supply the timeout, remove the one which doesn't require you to.

Also, if you plan to always supply the timeout, a change into lines 378 and 379 would provide more coersion:
```Java
if (timeout < 120) {
    timeout = 120;
}
```
into:

```Java
if (timeout < VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT) {
    timeout = VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT;
}
```

insom:
@alexandrelimassantana Agreed, thanks for the review. I've added a commit removing the prototype with the default. That has allowed me to remove that if altogether and place it inside the `applyConfigToVR` method.

@DaanHoogland Because the original commit is about the wrong signature for the method being used I'm not quite sure how to add tests for it. I hope that this second commit will help as it means the compiler can enforce things. Thanks for the review!

(Also it's Friday and I don't have a Maven install on this laptop - should pass the CI tests, if not, I may need to come back to it on Tuesday).

core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
```diff
@@ -180,7 +179,7 @@ private Answer applyConfig(NetworkElementCommand cmd, List<ConfigItem> cfg) {
         boolean finalResult = false;
         for (ConfigItem configItem : cfg) {
             long startTimestamp = System.currentTimeMillis();
-            ExecutionResult result = applyConfigToVR(cmd.getRouterAccessIp(), configItem);
+            ExecutionResult result = applyConfigToVR(cmd.getRouterAccessIp(), configItem, VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT);
```
cristofolini: @insom Since you're now checking the timeout's validity when calling `applyConfigToVR` shouldn't the last parameter passed through this call be `timeout` instead of `VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT`?

alexandrelimassantana:
@cristofolini there is no _timeout_ variable in the scope you commented. His change is valid because  _VRScripts.DEFAULT_EXECUTEINVR_TIMEOUT_ is mapped to the previous default _timeout_ value. This is equivalent to the previous  applyConfigToVR() call without passing the timout value.

cristofolini:
@alexandrelimassantana Ah, I see that now. Thanks for the clarification! :)

core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
```diff
@@ -180,7 +179,7 @@ private Answer applyConfig(NetworkElementCommand cmd, List<ConfigItem> cfg) {
...
             if (s_logger.isDebugEnabled()) {
                 long elapsed = System.currentTimeMillis() - startTimestamp;
                 s_logger.debug("Processing " + configItem + " took " + elapsed + "ms");
```
pedro-martins: Hi @insom .

I know that this isn't your code but could you use String.format to create the string that is used by logger?
It turns the code more readable than using multiple strings concatenation.

Ty.

insom:
@pedro-martins Thanks for the review but I don't want to make this change because (a) it'll mix two intents into on PR - fixing the bug *and* improving the logging - and (b) it would be *less* consistent with the rest of the CloudStack code base which, for better or worse, largely uses string concatenation for debug logging.

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 152
 Hypervisor xenserver
 NetworkType Advanced
 Passed=68
 Failed=5
 Skipped=3

**Failed tests:**
* test_vpc_vpn.py

 * ContextSuite context=TestRVPCSite2SiteVpn>:setup Failing since 13 runs

 * ContextSuite context=TestVpcRemoteAccessVpn>:setup Failing since 13 runs

 * ContextSuite context=TestVpcSite2SiteVpn>:setup Failing since 13 runs

* test_volumes.py

* test_vm_life_cycle.py

 * test_10_attachAndDetach_iso Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_disk_offerings.py

insom:
@bvbharatk Hi, am I supposed to do something with this CI output? It looks like all of the errors are related to other parts of CloudStack that aren't touched by my commits. I came back to check on this bug because I got bitten by this again today (timeout when adding rules to a virtual router, but the timeout is actually not getting applied properly).

insom:
I got contacted directly by someone else hit by this bug and looking for a fix. They saw I had commits relating to their problem. What needs to happen to get it merged?

ustcweizhou:
many people thought that changing router.aggregation.command.each.timeout in global setting will increase the timeout. Actually it is not, at least for KVM.

for KVM hypervisors, we need to put router.aggregation.command.each.timeout=??? in agent.properties and restart cloudstack-agent.

This PR increases the minimal timeout to 120 seconds which is helpful but does not really fix the issue.

I think it is better to add host-level setting for each host in the future, like we did for domain/account, zone/cluster.

insom:
@ustcweizhou A _side effect_ of this PR is setting the minimal timeout to 120, but that's what the original code intended - except that because of not passing through the timeout to `applyConfigToVR` it wasn't taking effect.

What this PR really does is make the code work as intended. I'm up for improvements but if this is blocking upgrades (and at least iWeb and one other CloudStack user were hit by it) seems like merging this and _then_ moving it to a per-host setting would be a good idea?

ustcweizhou:
@insom do not worry, actually I am +1 with this PR. I agree with what you said.

ozhanrk:
Hi, last week we experience some problems because of the timeouts which mentioned on this PR. Could we get this PR to upcoming versions asap please. Without this timeout configuration functionality working, its very hard to make upgrades on VR, timeout issue generates real headaches.

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 120
 Hypervisor xenserver
 NetworkType Advanced
 Passed=100
 Failed=2
 Skipped=5

**Failed tests:**
* test_deploy_vm_iso.py

 * test_deploy_vm_from_iso Failing since 5 runs

* test_vm_life_cycle.py

 * test_10_attachAndDetach_iso Failing since 6 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_ssvm.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_routers_network_ops.py
test_disk_offerings.py

karuturi:
Code LGTM.

BVTs are good(iso failures are URL access issues and not related to this PR).

merging this now

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 121
 Hypervisor xenserver
 NetworkType Advanced
 Passed=103
 Failed=2
 Skipped=5

**Failed tests:**
* test_deploy_vm_iso.py

 * test_deploy_vm_from_iso Failing since 6 runs

* test_vm_life_cycle.py

 * test_10_attachAndDetach_iso Failing since 7 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_ssvm.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_routers_network_ops.py
test_disk_offerings.py

PR #1457, opened 2016-03-31
CLOUDSTACK-9333: Exclude clusters from OVF operations

nvazquez:
JIRA TICKET: https://issues.apache.org/jira/browse/CLOUDSTACK-9333 

## Introduction
In some environments there is a need to exclude certain VMware clusters from performing OVF operations. This operations are part of: 
* create template
* create volume snaphsot
* copy template, volume, images from primary storage to secondary storage
* migrate volume 
* participate when a template gets cached over to primary storage.

In ESX/ESXi, OVF operations are low priority and bound to a single CPU and most likely get throttled to certain IOPS and network limits. 
If the hypervisor chosen for OVF operations is weak or overloaded this results in significantly longer execution of such OVF command and therefore degraded performance of underlying CloudStack API call.

### Proposed solution
It is proposed to add a way to exclude hosts from selected clusters for OVF operations.
To exclude a cluster, would be necessary to insert a record in <code>cluster_details</code> specifying property **vmware.exclude_from_ovf** in this way: (supposing we want to exclude cluster X)

| cluster_id | name| value |
|:-------------:|:-------------:|:-------------:|
|X|vmware.exclude_from_ovf|true|

### DIFF
--- a/engine/components-api/src/com/cloud/capacity/CapacityManager.java
+++ b/engine/components-api/src/com/cloud/capacity/CapacityManager.java
@@ -53,6 +53,16 @@ public interface CapacityManager {
             "0.85",
             "Percentage (as a value between 0 and 1) of allocated storage utilization above which allocators will disable using the pool for low allocated storage available.",
             true, ConfigKey.Scope.Zone);
+    static final ConfigKey<Boolean> StorageOperationsExcludeCluster =
+            new ConfigKey<Boolean>(
+                    Boolean.class,
+                    "cluster.storage.operations.exclude",
+                    "Advanced",
+                    "false",
+                    "Exclude cluster from storage operations",
+                    true,
+                    ConfigKey.Scope.Cluster,
+                    null);
 
     public boolean releaseVmCapacity(VirtualMachine vm, boolean moveFromReserved, boolean moveToReservered, Long hostId);
 
--- a/engine/storage/src/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
+++ b/engine/storage/src/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
@@ -44,6 +44,7 @@ import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
 import org.apache.cloudstack.storage.LocalHostEndpoint;
 import org.apache.cloudstack.storage.RemoteHostEndPoint;
 
+import com.cloud.capacity.CapacityManager;
 import com.cloud.host.Host;
 import com.cloud.host.HostVO;
 import com.cloud.host.Status;
@@ -64,9 +65,13 @@ public class DefaultEndPointSelector implements EndPointSelector {
     private static final Logger s_logger = Logger.getLogger(DefaultEndPointSelector.class);
     @Inject
     HostDao hostDao;
-    private final String findOneHostOnPrimaryStorage =
-        "select h.id from host h, storage_pool_host_ref s  where h.status = 'Up' and h.type = 'Routing' and h.resource_state = 'Enabled' and"
-            + " h.id = s.host_id and s.pool_id = ? ";
+    private final String findOneHostOnPrimaryStorage = "select t.id from "
+                            + "(select h.id, cd.value "
+                            + "from host h join storage_pool_host_ref s on h.id = s.host_id  "
+                            + "join cluster c on c.id=h.cluster_id "
+                            + "left join cluster_details cd on c.id=cd.cluster_id and cd.name='" + CapacityManager.StorageOperationsExcludeCluster.key() + "' "
+                            + "where h.status = 'Up' and h.type = 'Routing' and h.resource_state = 'Enabled' and s.pool_id = ? ";
+
     private String findOneHypervisorHostInScope = "select h.id from host h where h.status = 'Up' and h.hypervisor_type is not null ";
 
     protected boolean moveBetweenPrimaryImage(DataStore srcStore, DataStore destStore) {
@@ -115,6 +120,7 @@ public class DefaultEndPointSelector implements EndPointSelector {
             sbuilder.append(scope.getScopeId());
         }
         // TODO: order by rand() is slow if there are lot of hosts
+        sbuilder.append(") t where t.value<>'true' or t.value is null");    //Added for exclude cluster's subquery
         sbuilder.append(" ORDER by rand() limit 1");
         String sql = sbuilder.toString();
         HostVO host = null;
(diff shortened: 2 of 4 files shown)


rafaelweingartner:
@nvazquez after reading your explanation at the PR's body I understood the change. 

That change can be used to any kind of hypervisor, right?
What about a different parameter name? Giving that it is not exclusive to VMware environments

serg38:
@rafaelweingartner How about 'cluster.exclude' or 'cluster.storage.operations.exclude'. I think the latter is the better choice.

rafaelweingartner:
@serg38, I agree with you.
The name 'cluster.storage.operations.exclude' seems to suit better the parameter.

rafaelweingartner:
@nvazquez I was thinking, isn't there any documentation/enumeration for all of the possible parameters to be used at the "cluster_details" table?

Shouldn’t that property be visible at the cluster configuration view in ACS?

For that, I believe you should write an SQL script inserting an entry (with the default value) at the cluster_details for every cluster that the environment has configured. I also believe that there should be somewhere else to be configured with the default values to be used when a new cluster is created.

nvazquez:
@rafaelweingartner thanks for reviewing! I couldn't find any document with all cluster_details, however we could find out that this property could be defined like in <code>cloudstack/engine/components-api/src/com/cloud/alert/AlertManager.java</code>, those properties are inserted in <code>configuration</code> table, specifying the class which defines it in <code>component</code> field. 
I think this new property <code>cluster.storage.operations.exclude</code> could be done that way, do you agree? In that case, in which component could it be defined?

rafaelweingartner:
@nvazquez,
I find the “AlertManager” class terrible for that. Its name gives the idea of something that manages alerts and not clusters parameters/configurations. 

To add that property, I was thinking about the “com.cloud.resource.ResourceManagerImpl.discoverCluster(AddClusterCmd)” method (lines 491 and 492 are already doing that for over-commitment factors), and then you could create a static entry at “CapacityManager” that seems to be the best option to hold such configurations.

If you change the “discoverCluster”, you would still need a script to add that parameter to those clusters that the environment already has. You have to do that to enable the users to change those parameters and use that functionality through the web interface.

What do you think?

nvazquez:
@rafaelweingartner 
Ok, nice. I like that option. The only question I have about this is that if property could be changed from UI by users after inserted in "cluster_details", as I have observed most properties set in "cluster_details" table are not listed in the UI (Infrastructure -> Clusters -> Cluster X -> Settings).
How can I do to enable changing it from the UI instead of updating DB table?

rafaelweingartner:
I have just tested; the cluster_details table does not work as the configuration table in which you can add an entry there and then that parameter is displayed at the UI.

I really do not know how to enable that for cluster_details, I think we need to dive a little deeper at the code. I do not have much spare time now, but could you try to check how we can enable that? Latter at the end of the week, I can give you a hand.

rafaelweingartner:
@nvazquez just an update.
I was going over "org.apache.cloudstack.api.command.admin.config.ListCfgsByCmd.execute()" that is used to list the cluster configuration at the UI. It seems that it uses the configuration table too. The method used is 
"com.cloud.server.ManagementService.searchForConfigurations(ListCfgsByCmd)". I did not have much time to analyse it, but I believe that a careful inspection can help us to understand the way those configurations work.

nvazquez:
@serg38 and I found out a way to change this property from UI, inserting the property in "configuration" table with scope="Cluster", insert sql:

INSERT INTO `cloud`.`configuration` 
(`category`, `instance`, `component`, `name`, `description`, `default_value`, `updated`, `scope`, `is_dynamic`) VALUES 
('Advanced', 'DEFAULT', ‘CapacityManager', 'cluster.storage.operations.exclude', 'Exclude cluster from storage operations', 'false', now(), 'Cluster', '1');

This way, all clusters will be inheriting the property in its default value, which is false. So if you want to set it true, you can do it from UI, and will be inserted in "cluster_details" table with value = true.

nvazquez:
I pushed a solution with this changes, without including insert in sql schema, do you think this way could be ok?

rafaelweingartner:
I think this is the best way to tackle the problem.
You just need to add that insert into the version’s upgrade script.

nvazquez:
@rafaelweingartner cool! I added it.

rafaelweingartner:
For me everything (with the code) is ok now, LGTM for the PR.
Could you please run functional tests?

swill:
@rafaelweingartner I can run tests via KVM, but I don't have the ability to run tests against VMware hardware.  Are tests against KVM good enough in this case?

rafaelweingartner:
Good question, those changes affect every single hypervisor. 
I have reviewed the code and after that I can say that if those changes work for KVM, I see no reason not to work for VMware, XenServer and others. In my opinion that would be enough.

swill:
Perfect.  Thank you both for verifying this.  I probably won't be able to get to testing this till tomorrow evening because I am about to reinstall my CI to get my tests running on SSD drives.

nvazquez:
thanks @rafaelweingartner and @swill !

swill:
## CI RESULTS

**84/85 TESTS PASSED**

The only failed test is a test that often fails in my environment and is unrelated to this change.


**Associated Uploads**

**`test_network_R8I4SB`:**

* [results.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1457/test_network_R8I4SB/results.txt)

**`test_vpc_routers_RRQPFX`:**

* [results.txt](https://objects-east.cloud.ca/v1/e465abe2f9ae4478b9fff416eab61bd9/PR1457/test_vpc_routers_RRQPFX/results.txt)


Uploads will be available until `2016-06-07 02:00:00 +0200 CEST`

*Comment created by [`upr comment`](https://github.com/swill/upr).*

nvazquez:
Looks nice! Thanks @swill for testing it in your environment!

swill:
Based on @rafaelweingartner's code review and the CI results, I think this is ready to merge.

rafaelweingartner:
@swill, I was going to say the same.
I also believe that this PR is ready to be merged. But, I would like to have heard some other committer feedback.

swill:
@rafaelweingartner Ya, ideally I would like to try to get two code reviews for each PR.  I try to do a quick code review on each, but with the number for PRs I am trying to get through, it is not realistic that I do a thorough review of each.

@serg38 I know you have reviewed this code previously, can you have a look and give us your feedback?

rafaelweingartner:
@swill you are doing a marvelous job; I totally understand your workload.
I hope we can get more people to help us reviewing PRs; that number has been reduced to a handful of people lately.

serg38:
@swill , @rafaelweingartner I work very closely with @nvazquez. His PRs pass very rigorous testing in our Lab on physical Vmware hypervisors. I reviewed his code and give my OK.

serg38:
@rafaelweingartner  On a general note we got a verbal commitment from Accelerite who bought CloudPlatform from Citrix last month to contribute more to the community and to certain extent to spearhead some of the overdue improvements in ACS.

rafaelweingartner:
@serg38 thats is great.
So, Do I or you (@Swill) do the honors to execute the merge.

swill:
@serg38 awesome, thank you.  I will get this merged...

Yes, good new about Accelerite.  I am very hopeful they will be able to add value to the ACS community going forward.

PR #1476, opened 2016-04-07
Fix deleting template from UI

remibergsma:
It failed because zoneid was zero. Added a check for it and only send it when it makes sense.

With zoneid:
screen shot 2016-03-29 at 21 40 11

Without:
screen shot 2016-03-29 at 21 43 16

Both result in:
screen shot 2016-03-29 at 21 43 27

### DIFF
--- a/ui/scripts/templates.js
+++ b/ui/scripts/templates.js
@@ -1451,8 +1451,18 @@
                                                      }
                                                  },
                                                  action: function(args) {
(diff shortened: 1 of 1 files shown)


pdube:
Hey @remibergsma  just noticed that you had a PR open for this. I fixed it as well with #1505 so we might be able to close this one

PR #1477, opened 2016-04-07
When no zone name is available display a default

remibergsma:
When a zone name is available, the previous behaviour is still there:
screen shot 2016-03-30 at 21 00 44

When there is no zone name, it used to display an empty name (where you had to click on to see details):
screen shot 2016-03-30 at 21 03 06

With this change, a default name `All` is displayed (because this happens when S3 storage is used that is region wide aka all zones):
screen shot 2016-03-30 at 20 53 50

Region wide S3:
screen shot 2016-03-30 at 21 04 55

FYI: Screenshot shows 'All', later renamed to 'All Zones', see code.

### DIFF
--- a/ui/scripts/templates.js
+++ b/ui/scripts/templates.js
@@ -1427,6 +1427,9 @@
                                                                 jsonObj.xenserverToolsVersion61plus = false;
                                                         }
                                                     }
+                                                    if (!'zonename' in jsonObj) {
+                                                        jsonObj.zonename = 'All Zones';
+                                                    }
                                                 }
 
                                                 args.response.success({


pdube:
LGTM. Do you think we should translate this?

rhtyd:
LGTM

A manual UI testing is needed

tag:mergeready

swill:
I think this is ready to merge.  @remibergsma has shown screenshots to verify the UI change does what it is supposed to and we have the LGTMs required.

rhtyd:
@swill apart from the PR author, it would be great if at least one more reviewer can share a screenshot; please proceed as per your discretion

swill:
@rhtyd in general, I agree with you.  The main blocker with that is the fact that very few people have the ability to test regional templates.  The code is pretty straight forward too, so that helps our confidence in the change...

@pdube do you have the ability to test this PR with our swift setup to verify it works as expected?

swill:
Unless someone is able to verify this, I think we have to take the screenshots from @remibergsma as verification.  Not very many people have the ability to test regional templates, so we have a very limited group of users who can give us verification on this.  If we don't have any verification by Monday, I will likely merge this and count on @remibergsma's verification as proof that it works...

swill:
I am going to merge this one.  I am confident with the change and I have verification from @remibergsma, so I think we are good to go on this one.

PR #1466, opened 2016-04-07
CLOUDSTACK-9340: General DB Optimization

nvazquez:
## Description
In some production environments there were being experimented delays in most of the jobs. A search for DB optimization was taken and some deficiencies were discovered, we can group them in 4 groups:
* Incorrect PRIMARY key
* Duplicate PRIMARY KEY
* Missing indexes (Add indexes to avoid full table scans)
* Some views query (Change view to improve account retrieval speed)

PR #1481, opened 2016-04-08
travis: increase build verbosity

bhaisaab:
Outputs for modules that fail or succeed with unit tests results.

Based on the issue raised in https://github.com/apache/cloudstack/pull/1466 this PR aims at increasing Travis build output so we can know which unit test fail.

/cc @swill @nvazquez -- let me know if we just allow outputting everything will help? I had restricted the output as including all of them would disallow viewing the tabular final integration/marvin tests at the end (only possible by viewing raw output of each travis job).

### DIFF
--- a/tools/travis/install.sh
+++ b/tools/travis/install.sh
@@ -37,9 +37,9 @@ fi
 export MAVEN_OPTS="-Xmx2048m -XX:MaxPermSize=500m -Djava.security.egd=file:/dev/./urandom"
 
 if [ $TEST_SEQUENCE_NUMBER -eq 1 ]; then
-   mvn -Pdeveloper,systemvm -Dsimulator clean install -T4 | egrep "Building|SUCCESS"
+   mvn -Pdeveloper,systemvm -Dsimulator clean install -T4 | egrep "Building|Tests|SUCCESS|FAILURE"
 else
-   mvn -Pdeveloper -Dsimulator clean install -DskipTests=true -T4 | egrep "Building|SUCCESS"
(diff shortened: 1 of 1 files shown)


swill:
@bhaisaab thanks for doing this.  I suspect the travis run will fail so I will have to break my own rule and force merge.  :)  I will wait for it to finish regardless and I will get this in so we can better understand what is going on.  Thanks again for the quick response to this, this is very helpful.

bhaisaab:
@swill I would say let's wait for travis to go green on this and then merge. Then in future if anything break, we can have a look at the logs to see which unit test or maven module failed.

swill:
Yes, that is what I plan to do.  I just saw a different squash commit complete successfully, so it seems that travis is not 'always' failing, even though it has failed with that same message on more than one PR.  One of the ones that failed was only pushing a comment in order to test if travis was broken and it still failed, so that means that travis is periodically failing against master right now.  That concerns me, so hopefully this goes green.  Fingers crossed.  :)

bhaisaab:
@swill sure :) one more thing I want to discuss, should we enable -Dnoredist for Travis builds? I maintain this noredist repo which we can use to git clone as the time of running these tests and install deps (without including them in our main repo) https://github.com/bhaisaab/cloudstack-nonoss

bhaisaab:
on a second thought I'll probably ask this on dev@

swill:
I will have to think about that a bit.  I think it would be valuable, but I
have to get my head around if there are implications in doing that.  Are
there any implications that you can think of?  Also, does this add value?
Does it enable us to do more complete testing?  Sorry if I am being naive,
but I like to go into this sort of stuff with measured expectations.

swill:
Ya, probably a good discussion to take to dev@.

PR #1485, opened 2016-04-10
Set default networkDomain to empty instead of username

remibergsma:
The 10th field of `createUserAccount` is `networkDomain` (See `AccountService.java`) and it is set to a var named `admin`, which is the user name.
So, the first user that is created in a domain that links to LDAP, creates the account within the domain, and sets the `networkDomain` field to the username. All next users are created in the same account.

Then we have the situation that in domain SBP we have a user `rbergsma` that logs in first, gets an account created and then (unless you override) all VMs started in the SBP domain will have network domain `rbergsma`. That is highly confusing and not what is should be.

The `linkDomainToLdap` api call has no `networkDomain` field, so I propose to make this field empty (set it to null). It's a sting and null / empty is allowed.

One can also specify the networkDomain when creating a VPC and also there it is allowed to be null.

When te networkDomain is needed (and is not set in the domain and not in the VPC) it is constructed by using `guest.domain.suffix` so there always is a networkDomain to be used.

It makes more sense to manually set it on a domain level, or specify it on the VPC and in the final case end up with something that is clearly generated (like cs342cloud.local) rather than the username of someone else.

### DIFF
--- a/plugins/user-authenticators/ldap/src/org/apache/cloudstack/api/command/LinkDomainToLdapCmd.java
+++ b/plugins/user-authenticators/ldap/src/org/apache/cloudstack/api/command/LinkDomainToLdapCmd.java
@@ -82,7 +82,7 @@ public class LinkDomainToLdapCmd extends BaseCmd {
                     if (account == null) {
                         try {
                             UserAccount userAccount = _accountService.createUserAccount(admin, "", ldapUser.getFirstname(), ldapUser.getLastname(), ldapUser.getEmail(), null,
-                                    admin, Account.ACCOUNT_TYPE_DOMAIN_ADMIN, domainId, admin, null, UUID.randomUUID().toString(), UUID.randomUUID().toString(), User.Source.LDAP);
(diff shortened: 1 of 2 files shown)


rhtyd:
@karuturi @imduffy15 comments on the change, thanks

tag:needlove

imduffy15:
I'm not familiar with that block of code, remi's changes and justification seem fine though.....
Some very oddly/badly worded variable names.

rhtyd:
LGTM

rhtyd:
tag:easypr

nlivens:
According to the explanation of @remibergsma, and the code changes : LGTM

kiwiflyer:
@swill This has 2 x LGTM. Simple change and probably doesn't require any additional CI IMO.
This is Ready to Merge

swill:
My CI does not have the capability of testing this, so even if I ran CI it would not cover this.  I tend to agree that his is fine as is.

PR #1490, opened 2016-04-13
Installing bzip2 since it is required for extracting templates

sverrirab:
If you do not install bzip2, then installing templates that are bzip2 compressed will result in Cloudstack not being able to extract the contents and ending up copying the template in compressed form.  This will result in VMs not being able to start.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,7 +15,7 @@ Description: A common package which contains files which are shared by several C
 
 Package: cloudstack-management
 Architecture: all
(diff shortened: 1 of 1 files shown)


rhtyd:
LGTM

tag:easypr

swill:
We need more more code review on this one...

rafaelweingartner:
That change makes sense to me.
I do now if I have installed the “bzip2” program without noticing while installing some other tool, but in my MSs Servers it was already installed.
Anyways I can give LGTM here.

swill:
I think this one is ready to merge given the scope of the change.  Thanks...

PR #1495, opened 2016-04-14
Add perl-modules as install dependency for cloudstack-agent

sverrirab:
Required to run perl scripts that configure networking for VMs.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -22,7 +22,7 @@ Description: CloudStack server library
 
 Package: cloudstack-agent
 Architecture: all
-Depends: ${misc:Depends}, ${python:Depends}, openjdk-8-jre-headless | openjdk-7-jre-headless, cloudstack-common (= ${source:Version}), lsb-base (>= 4.0), libcommons-daemon-java, openssh-client, qemu-kvm (>= 1.0), libvirt-bin (>= 0.9.8), uuid-runtime, iproute, ebtables, vlan, jsvc, ipset, python-libvirt, ethtool, iptables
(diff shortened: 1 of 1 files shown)


eriweb:
Quite horribly that we need those modules for a simple perl script of 58 lines (including license header and empty lines). Not related to your PR though, but a good candidate to be rewritten in bash or python

jburwell:
@sverrirab how much effort would it be to port to Python?  That seems far preferable for two reasons.  First, as @eriweb rightly points out, it is a large dependency to take on for such a small script.  Second, Python and bash are our scripting languages of choice.  It does not seem sustainable to add runtimes/libraries for every scripting language people might want to use.  Such sprawl will not only lead to bloat, but a system that fewer and fewer people can maintain.

sverrirab:
I agree that this is not a good approach but it is the current one and it is broken.  How about we commit this fix and I submit a new PR to master with a new python script?

wido:
If this is broken right now and this fixes it, LGTM. But we should really ditch Perl and rewrite in Python or Bash.

Can you afterwards submit a PR against master?

sverrirab:
do you want this change as a new PR against master or just the rewritten script in python on master?

ProjectMoon:
As this is a bug fix, I think it would be forward-merged into master. The next PR should be to rewrite the script in Python instead of Perl, and that would be on master only.

rhtyd:
tag:easypr

sverrirab:
Created new PR on master: https://github.com/apache/cloudstack/pull/1526

rhtyd:
LGTM

wido:
LGTM

swill:
I think we should merge this one at this point.  @sverrirab my understanding is that once #1533 is merged, this will be required.  If that is the case, when this one is merged, maybe you can work into that PR the reverting of this PR.  Does that make sense?

PR #1498, opened 2016-04-16
CLOUDSTACK-9352: Test fails in Widows as the file separator "/" is different from "\"

GabrielBrascher:
**Problem:**
File separator in windows ("\") is different from the expected in the test ("/"); thus, the test *com.cloud.utils.SwiftUtilTest.testSplitSwiftPath()* will fail in Windows systems.

The problem is that the input of the test is "*container/object*" but the tested method uses the *File.separator* (that depends from the OS), in windows systems the tested method (*com.cloud.utils.SwiftUtil.splitSwiftPath(String)*) looks for a "\", as the string does not contain "\" it returns an empty string and consequently results in a test failure.

**Solution:**
Create a string `String input = "container" + File.separator + "object";`, with that the test will validate the tested method verifying if the method splits the string around matches of the given regular expression (in this case *File.separator*).

*JIRA link: https://issues.apache.org/jira/browse/CLOUDSTACK-9352*

### DIFF
--- a/utils/src/test/java/com/cloud/utils/SwiftUtilTest.java
+++ b/utils/src/test/java/com/cloud/utils/SwiftUtilTest.java
@@ -19,9 +19,11 @@
 
 package com.cloud.utils;
 
+
 import org.junit.Test;
 import org.mockito.Mockito;
 
+import java.io.File;
 import java.net.URL;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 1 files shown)


cloudsadhu:
reviewed your code change and looks good. Please post  the test result.
LGTM based on code review !!

swill:
I need one more LGTM and if you can post some proof this is working as expected, I will feel better about merging it.  Thanks...  :)

GabrielBrascher:
@swill it succesfully builds in windows: [build-result.txt](https://github.com/apache/cloudstack/files/230237/build-result.txt);

Not sure it if helps, but there is also a print showing the test failing, after the adjustment the test is ok: testrun

Let me know if you need more information :)

swill:
That is what I was looking for.  Thank you sir...  :)

PR #1504, opened 2016-04-20
CLOUDSTACK-9172 Added cross zones check to delete template and iso

pdube:
Added a check to ignore the zoneid, in the delete template UI, if the template is cross zones.

reference : CLOUDSTACK-9172


pdube:
Pushing towards 4.7

PR #1508, opened 2016-04-21
CLOUDSTACK-9362: Skip VXLANs when rewriting the bridge name for migrations

insom:
From the [JIRA issue](https://issues.apache.org/jira/browse/CLOUDSTACK-9362):

> https://github.com/apache/cloudstack/commit/bb8f7c652e42caacff5adce1ce60342603677605
> 
> The above commit introduces rewriting of bridge device names when migrating a virtual machine from one host to another. However, it also matches bridges called "brvx-1234" and rewrites them to (in my case) "brem1-1234" - this doesn't match the bridge name on the destination and causes the migration to fail with the error:
> 
> error : virNetDevGetMTU:397 : Cannot get interface MTU on 'brem1-1234': No such device
> 
> I have flagged this as major because it's not possible to migrate VMs using VXLANs for maintenance, which seems important (it's certainly important to me!).

### DIFF
--- a/agent/bindir/libvirtqemuhook.in
+++ b/agent/bindir/libvirtqemuhook.in
@@ -26,6 +26,8 @@ def isOldStyleBridge(brName):
     else:
        return False
 def isNewStyleBridge(brName):
+    if brName.startswith('brvx-'):
+        return False
     if re.match(r"br(\w+)-(\d+)", brName) == None:
        return False
     else:


kiwiflyer:
@insom Thanks for the PR. We'll pull this in for testing as we use vxlan.

As it looks like this bug was introduced in 4.8, I'd suggest you may want to issue your PR against the 4.8 branch. It can then be forward committed into master as well.

//cc @dmabry  @swill

swill:
@kiwiflyer if you wouldn't mind posting the results of your tests that would be appreciated.  :)

@insom if this was actually introduced in 4.8, I agree that we should open this against 4.8 and then I will forward merge into master.  Thank you for the effort.  Cheers...

insom:
@swill Thanks! Should I rebase this against the latest 4.8 branch then? I don't think I can change where a PR is based against, so I think I'd need to open a new PR - is that right call?

kiwiflyer:
@insom Yes, rebase it against the 4.8 branch and yes you'll need to issue a new PR and close this one.

swill:
@insom, yes @kiwiflyer is correct.  You will need to base your change off the 4.8 branch and close this PR and open a new PR based off the 4.8 branch.  Thanks for the effort.  :)

insom:
#1512 is the new version of this.

PR #1509, opened 2016-04-22
CLOUDSTACK-6975: Prevent dnsmasq from starting on backup RvR

dsclose:
CLOUDSTACK-6975 refers to service monitoring bringing up dnsmasq but this is no-longer accurate, as service monitoring is not active on the post-4.6 routers. These routers still suffer an essentially identical issue, however, because "dnsmasq needs to be restarted each time configure.py is called in order to avoid lease problems." As such, dnsmasq is still running on backup RvRs, causing the issues described in CLOUDSTACK-6975.

This PR is based on a patch submitted by @ustcweizhou. The code now checks the redundant state of the router before restarting dnsmasq.

RvR networks without this patch have dnsmasq running on both master and backup routers. RvR networks with this patch have dnsmasq running on only the master router.


ustcweizhou:
LGTM, tested

swill:
@dsclose this PR is currently opened against `apache:4.7.1-RC20160120T2318`.  Should this be merged into 4.7 and then we will forward merge it to 4.8 and master?

Need one more LGTM code review and I would like to verify this PR via CI.

dsclose:
@swill I raised the PR against `apache:4.7.1-RC20160120T2318` because it seemed sensible to include it in the latest 4.7 branch. I'd assumed it could be forward merged but as this is only my second PR for Cloudstack I don't think I should be making any recommendations.

It's not clear to me how to determine which branch I should raise PRs against - but that's probably a question for the mailing list rather than here. Let me know if you want it based off something else and I'll sort that out.

swill:
Lets close this PR and open it based off the 4.7 branch.  @ustcweizhou's LGTM vote will still count on the new PR.  Lets just reference this PR in the new PR so we remember to count his vote.  :)

Basically how this works is as follows:
- The current release is 4.8.  
- We support fixes in one release prior the the current release, and those fixes are put against the main branch, so in this case 4.7.  I am not sure of the release cycle on this branch, but at some point a newer minor release may be cut from this branch (something like 4.7.2 for example).
- Whenever a fix is made in a previous release, the fix is then forward merged into the later releases.  This is so the commit hash is consistent between all branches making it easier to understand if a specific fix is in a specific branch.  So if a commit is added to the 4.7 branch, it will be forward merged to the 4.8 branch, and then forward merged again to master.
- We treat master as the next release and add new features there and make sure everything is stable for the next release (in this case 4.9).

Does that all make sense?  Don't be shy if you have questions.  I will do my best to support you.  :)

dsclose:
Rebased off of 4.7 branch and opened PR #1514 - closing this PR.

dsclose:
@swill and thank you for you're support - much appreciated :-)

PR #1517, opened 2016-04-25
engine/schema: fix upgrade path to work with MySQL 5.7

rhtyd:
Found this issue when using MySQL 5.7 with Ubuntu 16.04. The upgrade path fix removes an invalid `IGNORE` param that is deprecated now, in the upgrade path we run the alter statement to add an index only if it does not exist so we're good.

For MySQL 5.7, we'll also need to update the docs at some point to include `server-id` along with other parameters. Some of the SQL statements used throughout engine/schema don't adhere to SQL 99 standard which is enforced by default in MySQL 5.7, therefore the following sql-mode (for backward compatibility with mysql 5.6 modes) will be necessary for anyone willing to use MySQL 5.7 (until we fix codebase wide raw and generated sql statements to be SQL99 compliant):

sql-mode="STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION,ERROR_FOR_DIVISION_BY_ZERO,NO_ZERO_DATE,NO_ZERO_IN_DATE,NO_ENGINE_SUBSTITUTION"
server-id   = 1
innodb_rollback_on_timeout=1
innodb_lock_wait_timeout=600
max_connections=350
log-bin=mysql-bin
binlog-format = 'ROW'

/cc @swill @jburwell @agneya2001 @wido @DaanHoogland  and others

### DIFF
--- a/engine/schema/src/com/cloud/upgrade/dao/Upgrade410to420.java
+++ b/engine/schema/src/com/cloud/upgrade/dao/Upgrade410to420.java
@@ -1297,7 +1297,7 @@ public class Upgrade410to420 implements DbUpgrade {
                     s_logger.debug("Index already exists on host_details - not adding new one");
                 } else {
                     // add the index
-                    try(PreparedStatement pstmtUpdate = conn.prepareStatement("ALTER IGNORE TABLE `cloud`.`host_details` ADD INDEX `fk_host_details__host_id` (`host_id`)");) {
+                    try(PreparedStatement pstmtUpdate = conn.prepareStatement("ALTER TABLE `cloud`.`host_details` ADD INDEX `fk_host_details__host_id` (`host_id`)");) {
                         pstmtUpdate.executeUpdate();
                         s_logger.debug("Index did not exist on host_details -  added new one");
                     }catch (SQLException e) {


rhtyd:
I found this article by Oracle MySQL team useful: https://www.digitalocean.com/community/tutorials/how-to-prepare-for-your-mysql-5-7-upgrade

jburwell:
Do these instructions apply to MariaDB as well?

rhtyd:
@jburwell I've not tested, but if it is using/enforcing SQL99 it should fail too (or it could be mysql 5.7.4+ issue);

"As of MySQL 5.7.4, the IGNORE clause for ALTER TABLE is removed and its use produces an error."
from http://dev.mysql.com/doc/refman/5.7/en/alter-table.html

wido:
So what happens if the Index already exists?

engine/schema/src/com/cloud/upgrade/dao/Upgrade410to420.java
```diff
@@ -1297,7 +1297,7 @@ private void addHostDetailsIndex(Connection conn) {
                     s_logger.debug("Index already exists on host_details - not adding new one");
```
rhtyd: @wido the index is not added, if you look at the code above ^^ the if statement to check if index already exists with given key name and it only adds the index if it does not exist otherwise skips.

wido:
Ah, yes, I see @rhtyd !

LGTM

koushik-das:
Code changes LGTM.
@swill The change is in upgrade code and CI won't be able to test it. So I think this can be merged.

rhtyd:
Thanks @koushik-das 
CI will be able to test it though as cloudstack-setup-database deploys 4.0 schema and ACS mgmt server upgrades it or DatabaseUpgrader runs the upgrade path as well. Travis deployed database and is green confirms upgrade path fix does not break with existing MySQL 5.6 or less version.

koushik-das:
@rhtyd Thats right, setup will anyway do the upgrades.

swill:
Thanks for the validation discussion.  I think this is ready...

PR #1348, opened 2016-01-18
CLOUDSTACK-9142 Migrate VM changes xmlDesc in a safe way

DaanHoogland:
The problem arises when the origin hypervisor has an ip addres that ends with 1, like '10.10.10.1' and the qemu VM description is containing an address that has that as part of its address, '10.10.10.100' for instance.
now migrating to '10.10.10.10' will change both addresses in the xml description file for qemu. It is fixed and unit tests are added. I am not sure yet how to integration test this. Regression will probably work so creating a PR now.

PR #1521, opened 2016-04-27
Kvm plugin restructure

DaanHoogland:
test and project restructure without functional implications


DaanHoogland:
@swil @rhtyd this is the follow up of #1348. The initial rat failure was genuine but next files where found that are .gitignored. not sure what happened

rhtyd:
In my experience when you are doing a lot of backporting or forward-porting work, if a plugin/module has got its filesystem structure (for example, due to moving to maven standard layout) changed it creates a lot of issues and the process becomes painful.

One option is that we decide to move the entire codebase to maven standards before LTS (do it for once and for all), and bear the pain for next some months only. Though, lately at some point what if we decide to migrate to gradle or something else (just speculating future).

@jburwell @DaanHoogland @wido @swill what do you guys think?

DaanHoogland:
@rhtyd I agree we should. At the moment we are already in an intermediate state. Some projects were moved to maven standard (I only know utils by head and the ovm3 plugin was created as such)

The reason to move this project was that I was going to create test only resources and didn't want to put them in the resources dir for production code.

jburwell:
@rhtyd +1 for Gradle though it would be a large undertaking.  Since Gradle's Java plugin natively supports Maven's directory structure, such a transition would actually help a migration Gradle.

DaanHoogland:
+ 1 for moving away from maven. How about sbt?

rhtyd:
@DaanHoogland I'm not sure about others, but gradle has become most popular and widely-support in modern java development
@jburwell agree, the discussion here is to evaluate how useful it would be to move to maven standard layout if the build chain gets migrated (to say gradle, sbt or something else) when doing this adds pain to the backporting/forward-porting process at the cost of a cosmetic change

wido:
I am not against this restructure perse. But shall we not try to change to many things at once?

We have enough on our plates, so talking about leaving Maven seems like a too-early discussion imho.

jburwell:
I would love to dump Maven for Gradle.  However, I agree with @wido -- we currently have much higher priorities to address.  For me, it is fun to banter a bit about it in a PR thread. ;)

Changing the Maven structure has other benefits.  Not only could we rid ourselves of the additional code to override the structure, conforming wider Java idioms and standards it easier for new contributors to understand the codebase.  @wido I agree with your suggestion to make one bulk change.  It would require coordination across the community as it would require a lot of rebasing once it is merged, but it seems like we would get an economy of scale on the effort.

DaanHoogland:
Enjoying the discussion I would like to -1 for the entire refactor all at once. I changed this project because it was feasible and convenient. I can see myself doing more then just this one but why refactorring unsupported projects that are laying around waiting for a quiet or banging death out there. Let's do only those things that are actively being maintained.

swill:
@DaanHoogland please rebase as master has caused merge conflicts...

DaanHoogland:
@rhtyd @jburwell @wido @swill Are we not doing this? I see no :-1: but also no LGTM

swill:
We have merge conflicts on this one.  As soon as the merge conflicts are sorted out, I will run CI against it and hopefully we can get some code reviews quickly so we can keep it moving...

wido:
I am not against it, but we have more KVM Pull Requests open like #1545

At what point do we ask other PRs to be rebased after this one merges in?

DaanHoogland:
I will rebase again, if we do want this.

wido:
At this moment I don't see the benefit of doing a restructure as it will kill any current work for somebody working on something in KVM.

I personally thing we have other things to worry about then the filestructure of the KVM plugin.

PR #1524, opened 2016-04-30
KVM: Fix plugging ip addresses to the wrong interface

remibergsma:
Fix for KVM, where the public ip would be added to the guest network. Reason is because it used 'broadcastUrl' as an index, which turned out not to be unique in all cases and was overwritten. This resulted in the public ip address being added to the wrong interface (see screenshots).

Problem:
kvm-wrong-interface-1

After patch:
kvm-wrong-interface-2

The `192.168.23.0/24` range is used as 'public' in this lab.

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -1600,24 +1600,33 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             conn = LibvirtConnection.getConnectionByVmName(routerName);
             final IpAddressTO[] ips = cmd.getIpAddresses();
             Integer devNum = 0;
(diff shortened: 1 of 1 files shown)


kiwiflyer:
Strangely enough, I just ran into this problem yesterday.I'll pull it in. Thanks Remi!

remibergsma:
Jenkins error says: `Unable to instrument project`, seems like a workspace issue or such.

remibergsma:
This time around Jenkins is happy and Travis times out :-s
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

remibergsma:
Force pushed a final time, else I give up on this.

remibergsma:
@kiwiflyer Sorry but this seems to have issues. We run this in prod with Cosmic but apparently some other patch is missing here in order for this to work properly. Unfortunately I have no more time to work on this now.

ustcweizhou:
this does not work for me as well.
If there are multiple public interfaces in the VR, only the last one works.

PR #1526, opened 2016-05-02
Add perl-modules as install dependency for cloudstack-agent

sverrirab:
Required to run perl scripts that configure networking for VMs.
That script fails silently if this is not installed.

As requested here: https://github.com/apache/cloudstack/pull/1495

Another PR will be submitted with the module written in python to remove this dependency...

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -22,7 +22,7 @@ Description: CloudStack server library
 
 Package: cloudstack-agent
 Architecture: all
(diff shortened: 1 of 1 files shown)


swill:
You can close this PR.  Once #1495 is merged, it will be forward merged into 4.8 and master, so this PR is not needed.  Thanks...  :)

PR #1374, opened 2016-01-27
Update L10N resource files with 4.7 strings from Transifex (20160127)

milamberspace:
cc @remibergsma @bhaisaab @DaanHoogland

PR #1527, opened 2016-05-02
Update L10N resource files with 4.7 strings from Transifex (20160502)

milamberspace:
Force "translator" mode with the transifex client.

cc @swill the new PR.


swill:
Previous PR was #1374, for book keeping...

PR #1535, opened 2016-05-07
Honour GS use_ext_dns and redundant VR VIP

ntavares:
This patch addresses two issues:

- On redundant VR setups, the primary resolver being handed out to instances is the guest_ip (primary IP for the VR). This might lead to problems upon failover, at least while the DHCP lease doesn't update (because the primary resolver will be checked first until times out, however it'll be gone upon failover).

- If Global Setting use_ext_dns is true, we don't want the VR to be the primary resolver at all.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsConfig.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsConfig.py
@@ -68,8 +68,11 @@ class CsConfig(object):
 
     def get_dns(self):
         dns = []
-        # Check what happens with use_ext_dns
-        dns.append(self.address().get_guest_ip())
+        if not self.cl.get_use_ext_dns():
+            if not self.is_vpc() and self.cl.is_redundant():
+                dns.append(self.cl.get_guest_gw())
+            else:
+                dns.append(self.address().get_guest_ip())
         names = ["dns1", "dns2"]
         for name in names:
             if name in self.cmdline().idata():
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsDatabag.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsDatabag.py
@@ -148,3 +148,9 @@ class CsCmdLine(CsDataBag):
         if "gateway" in self.idata():
             return self.idata()['gateway']
         return False
+
+    def get_use_ext_dns(self):
+        if "useextdns" in self.idata():
+            return self.idata()['useextdns']
+        return False
+


systemvm/patches/debian/config/opt/cloud/bin/cs/CsConfig.py
```diff
@@ -68,8 +68,11 @@ def get_domain(self):
 
     def get_dns(self):
         dns = []
-        # Check what happens with use_ext_dns
-        dns.append(self.address().get_guest_ip())
+        if not self.cl.get_use_ext_dns():
+            if not self.is_vpc() and self.cl.is_redundant():
+                dns.append(self.cl.get_guest_gw())
+            elif not self.cl.get_use_ext_dns():
```
DaanHoogland: the exression (not self.cl.get_use_ext_dns()) already known to be true due to line 71, so this check is redundant.

DaanHoogland:
logic makes sense, one remark about the code made. starting integration tests on it.

DaanHoogland:
should this not go on 4.7?

DaanHoogland:
one other concern; should the behaviour be more unified between regular vr and vpc somehow? it diverting here, can we prevent that?

systemvm/patches/debian/config/opt/cloud/bin/cs/CsConfig.py
ntavares: @DaanHoogland you're right... cleaned it up.. thanks for spotting it.

ntavares:
@DaanHoogland The patch is ported from 4.7 yes. 

The whole VR and VPC should be unified anyway :-)) I noticed there is CsGuestNetwork.get_dns() but we're not there yet, so not sure yet if we can apply same logic.

DaanHoogland:
@ntavares are you saying this is already on 4.7 or on your own 4.7 branch? If it is on 4.7 it should get to master automatically.

ntavares:
Ooops sorry about the confusion. It's on my own 4.7 branch. 4.7 is also missing this, but I wasn't sure how to create the PR for it.

DaanHoogland:
Ah, ok. Plese do make it to 4.7. It's the same way as to master except you need to make some extra choices during creation. Create it on your fork of 4.7 and when creating your PR you need to make sure 4.7 is selected on the apache/cloudstack side as well.

ntavares:
I understand that, my question is which one of the 1000 branches is 4.7? :)
-NT

A Sáb, 7/05/2016, 18:38, Daan Hoogland <notifications@github.com> escreveu:

DaanHoogland:
:) the one with the name '4.7'.

ustcweizhou:
LGTM, tested on 4.7

DaanHoogland:
would like to see it on 4.7 first otherwise: LGTM code reviewed and tested

CI result:

only the ping outside wolrd failure, which is not due to any code in this PR

[1535.results.network.txt](https://github.com/apache/cloudstack/files/253957/1535.results.network.txt)

ntavares:
Here: https://github.com/apache/cloudstack/pull/1536

DaanHoogland:
@ntavares  can you close this one?

ntavares:
@DaanHoogland closed.

PR #1536, opened 2016-05-08
Honour GS use_ext_dns and redundant VR VIP

ntavares:
This patch addresses two issues:

On redundant VR setups, the primary resolver being handed out to instances is the guest_ip (primary IP for the VR). This might lead to problems upon failover, at least while the DHCP lease doesn't update (because the primary resolver will be checked first until times out, however it'll be gone upon failover).

If Global Setting use_ext_dns is true, we don't want the VR to be the primary resolver at all.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsConfig.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsConfig.py
@@ -68,8 +68,11 @@ class CsConfig(object):
 
     def get_dns(self):
         dns = []
-        # Check what happens with use_ext_dns
-        dns.append(self.address().get_guest_ip())
+        if not self.cl.get_use_ext_dns():
+            if not self.is_vpc() and self.cl.is_redundant():
+                dns.append(self.cl.get_guest_gw())
+            else:
+                dns.append(self.address().get_guest_ip())
         names = ["dns1", "dns2"]
         for name in names:
             if name in self.cmdline().idata():
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsDatabag.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsDatabag.py
@@ -148,3 +148,9 @@ class CsCmdLine(CsDataBag):
         if "gateway" in self.idata():
             return self.idata()['gateway']
         return False
+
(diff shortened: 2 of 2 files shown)


ntavares:
Same as: https://github.com/apache/cloudstack/pull/1535 but for 4.7

DaanHoogland:
@swill can we get this in. It's got 2 LGTM and was integration tested on master and monkey tested on 4.7

DaanHoogland:
LGTM

DaanHoogland:
@ustcweizhou, I suppose it still LGTM to you as well

DaanHoogland:
CI RESULTS:

initial fails due to environmental issues all tests passed after rerun:

[1536.results.internal_lb.txt](https://github.com/apache/cloudstack/files/262244/1536.results.internal_lb.txt)
[1536.results.network.txt](https://github.com/apache/cloudstack/files/262245/1536.results.network.txt)
[1536.results.routers_network_ops.txt](https://github.com/apache/cloudstack/files/262246/1536.results.routers_network_ops.txt)
[1536.results.ssvm.txt](https://github.com/apache/cloudstack/files/262247/1536.results.ssvm.txt)
[1536.results.vm_life_cycle.txt](https://github.com/apache/cloudstack/files/262248/1536.results.vm_life_cycle.txt)
[1536.results.vpc_routers.txt](https://github.com/apache/cloudstack/files/262249/1536.results.vpc_routers.txt)

DaanHoogland:
@swill, given that @ustcweizhou gave his ok on #1535, this one can be merged.

swill:
I just realized this one changes the system vm. Did you deploy a new system VM when you tested this one @DaanHoogland?

DaanHoogland:
It only changes the iso not the template. So what do you mean @swill it is part of the run to create a new iso and use it, right?

swill:
So does this change require that we push new system VM templates?

DaanHoogland:
no installing the new iso should do it, but wasn't there already another change that requires a template update?

swill:
So far I don't think any of the other changes that have gotten in have required system template changes.  Which is why I am asking.  There is one that I think is probably ready to get added, but it is a one liner.  StrongSwan and OSPF were the big ones that would have required a new system vm template...

PR #1538, opened 2016-05-11
Disable NioTest

rhtyd:
Historically NioTest has caused issue in CI environment and several developer machines due to network requirements which could be disabled by firewall or security enforcers such as selinux. This disables the test once again using a historic commit 881a6e1
Signed-off-by: Rohit Yadav <rohit.yada

To build and just run this test: mvn clean install -pl utils -Dtest=NioTest

### DIFF
--- a/utils/pom.xml
+++ b/utils/pom.xml
@@ -208,6 +208,7 @@
           <excludes>
             <exclude>com/cloud/utils/testcase/*TestCase*</exclude>
             <exclude>com/cloud/utils/db/*Test*</exclude>
+            <exclude>com/cloud/utils/testcase/NioTest.java</exclude>
           </excludes>
         </configuration>
       </plugin>


rhtyd:
@swill @nvazquez @mike-tutkowski appreciate if we can discuss the issue here instead of ML, we'll lose discussion otherwise and I'm unable to keep up discussions at two places. Thanks.

mike-tutkowski:
I should be able to try your branch out later today. Thanks!

On May 12, 2016, at 12:18 AM, Rohit Yadav <notifications@github.com<mailto:notifications@github.com>> wrote:


@swill<https://github.com/swill> @nvazquez<https://github.com/nvazquez> @mike-tutkowski<https://github.com/mike-tutkowski> appreciate if we can discuss the issue here instead of ML, we'll lose discussion otherwise and I'm unable to keep up discussions at two places. Thanks.

-

DaanHoogland:
LGTM

mike-tutkowski:
@rhtyd I was able to successfully run mvn -P developer,systemvm clean install -D noredist now on Mac OS X El Capitan 10.11.4, so LGTM.

rafaelweingartner:
Hi guys, 
Please do not take me in a bad way, but why do we keep a test code that does not work? To me it is clearly that the test and the code being tested need a complete refactoring to enable true unit and integration tests (not functional ones).  I was looking at the history of “utils\pom.xml” and it feels like those tests were being ignored since ever.

swill:
@rhtyd Can you re-push now that we have fixed the issue that is failing in travis?  Thx...

swill:
I have forced pushed this because it is blocking other PRs that are only pending Jenkins and Travis and since we are so close to freeze date, I don't want stuff to not get in because of stupid stuff like an unrelated test failing...

PR #1497, opened 2016-04-15
CLOUDSTACK-9351: Add ids parameter to resource listing API calls

nvazquez:
## General behaviour
A new parameter is added in each method, its type a list of IDs of the entity, it will be mutually exclusive with id. (Similar to <code>id</code> and <code>ids</code> parameters in <code>listVirtualMachines</code> method)

### API Methods affected
* <code>listTemplates</code>: new parameter **<code>ids</code>**, mutually exclusive with <code>id</code>
* <code>listVolumes</code>: new parameter **<code>ids</code>**, mutually exclusive with <code>id</code>
* <code>listSnapshots</code>: new parameter **<code>ids</code>**, mutually exclusive with <code>id</code>
* <code>listVMSnapshots</code>: new parameter **<code>vmsnapshotids</code>**, mutually exclusive with <code>vmsnapshotid</code>

PR #1539, opened 2016-05-11
CLOUDSTACK-9378: Fix for #1497

nvazquez:
JIRA TICKET: https://issues.apache.org/jira/browse/CLOUDSTACK-9378

After #1497 was merged, test_list_ids_parameter.py failed on Travis. It was noticed that snapshots could be removed from cleaup

### DIFF
--- a/test/integration/smoke/test_list_ids_parameter.py
+++ b/test/integration/smoke/test_list_ids_parameter.py
@@ -194,10 +194,7 @@ class TestListIdsParams(cloudstackTestCase):
         cls._cleanup = [
                         cls.disk_offering,
                         cls.account,
-                        cls.service_offering,
-                        cls.snapshot_1,
-                        cls.snapshot_2,
-                        cls.snapshot_3
+                        cls.service_offering
                         ]
 
     @classmethod
--- a/test/integration/smoke/test_list_ids_parameter.py
+++ b/test/integration/smoke/test_list_ids_parameter.py
@@ -192,8 +192,11 @@ class TestListIdsParams(cloudstackTestCase):
                                          )
         
         cls._cleanup = [
-                        cls.disk_offering,
+                        cls.snapshot_1,
+                        cls.snapshot_2,
+                        cls.snapshot_3,
                         cls.account,
+                        cls.disk_offering,
                         cls.service_offering
                         ]
 


rhtyd:
LGTM, cc @swill let's merge this first to avoid other PRs failing

swill:
Sorry for the delay getting to this.  I think this probably works in all cases, but I suspect it will not clean up correctly always.  Because this was working sometimes and failing sometimes, I believe that means that the cleanup of `account` does not always clean up the snapshots (otherwise it would have never worked).

I think the code should be like this to make sure it cleans up everything up and does not fail.

```
cls._cleanup = [
	cls.snapshot_1,
	cls.snapshot_2,
	cls.snapshot_3,
	cls.disk_offering,
	cls.account,
	cls.service_offering
	]
```

Does this make sense to you guys?  This way it will remove the snapshots first and then it will clean up everything else.  I could be wrong, but since the current code ever worked, I think this guarantees that the cleanup of everything will happen and the `account` won't error when it tries to clean up.  I guess if the account keeps a reference to the snapshot after it is deleted, that could cause the account deletion to fail.  Feedback please...

rhtyd:
I think one way to cleanup would be to do it case by case basis for each test case manually instead of relying on the class level cleanup. Also, it's best to avoid using `cls` instead use cleanup for instance (use self.cleanup). I personally prefer to avoid class level setups, if we use instance `self.stuff` -- we know things are going to be created and cleaned-up for each test case.

test/integration/smoke/test_list_ids_parameter.py
```diff
@@ -194,10 +194,7 @@ def setUpClass(cls):
         cls._cleanup = [
                         cls.disk_offering,
                         cls.account,
-                        cls.service_offering,
-                        cls.snapshot_1,
-                        cls.snapshot_2,
-                        cls.snapshot_3
+                        cls.service_offering
```
rhtyd: I've tried to fix this, by reordering the items by ownership here: 
https://github.com/apache/cloudstack/pull/1502/commits/6135f6d98fa6b363b0468c17a8e713aeb5054437

DaanHoogland: I would say that cleanup should happen in reverse order of creation, irrespective of ownership. Leaving these items after an integration test while other tests may follow seems like a bad idea.

rhtyd: agree

DaanHoogland: Ah, I now looked at your link and, yes that is what I mean.

swill: I agree with your order.  @nvazquez can you update the order to what rohit posted in his commit.  @rhtyd is there a reason you added this to your PR?  Will you just remove it from that PR once this is updated and merged?

rhtyd: @swill I wanted the out-of-band management PR to go green so it could be merged soon without missing the freeze date, that's why I fixed it in my PR.

swill: Haha.  Ok.

nvazquez:
@swill @rhtyd @DaanHoogland I reordered cleanup as in [@rhtyd's commit](https://github.com/apache/cloudstack/pull/1502/commits/6135f6d98fa6b363b0468c17a8e713aeb5054437)

test/integration/smoke/test_list_ids_parameter.py
nvazquez: Done

swill:
@nvazquez, @rhtyd's commit has been merged as part of his larger PR at #1502, so we can close this PR since the result of this PR has already been merged.  Thanks everyone...

nvazquez:
Thanks, closed

PR #1542, opened 2016-05-12
CLOUDSTACK-9379: Support nested virtualization at VM level on VMware Hypervisor

nvazquez:
## Introduction

[JIRA TICKET](https://issues.apache.org/jira/browse/CLOUDSTACK-9379)

It is desired to support nested virtualization at VM level for VMware hypervisor. Current behaviour supports enabling/desabling global nested virtualization by modifying global config `'vmware.nested.virtualization'`. It is wished to improve this feature, having control at VM level instead of a global control only.
## Proposal

A new global configuration is added, to enable/disable VM nested virtualization control: `'vmware.nested.virtualization.perVM'`. Default value=false
## Behaviour

After a vm deployment or start command, vm params include `'nestedVirtualizationFlag'` key and its value is:
- true -> nested virtualization enabled
- false -> nested virtualization disabled

**We will determinate nested virtualization enabled/disabled by examining this 3 values:**
- **(1)** global configuration `'vmware.nested.virtualization'` value
- **(2)** global configuration `'vmware.nested.virtualization.perVM'` value
- **(3)** `'nestedVirtualizationFlag'` value in `user_vm_details` if present, `null` if not.

Using this 3 values, there are different use cases:
- **(1)** = TRUE, **(2)** = TRUE, **(3)** is null -> _ENABLED_
- **(1)** = TRUE, **(2)** = TRUE, **(3)** = TRUE -> _ENABLED_
- **(1)** = TRUE, **(2)** = TRUE, **(3)** = FALSE -> _DISABLED_
- **(1)** = TRUE, **(2)** = FALSE, **(3)** indifferent  -> _ENABLED_
- **(1)** = FALSE, **(2)** = TRUE, **(3)** is null -> _DISABLED_
- **(1)** = FALSE, **(2)** = TRUE, **(3)** = TRUE -> _ENABLED_
- **(1)** = FALSE, **(2)** = TRUE, **(3)** = FALSE -> _DISABLED_
- **(1)** = FALSE, **(2)** = FALSE, **(3)** indifferent -> _DISABLED_

### DIFF
--- a/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -2208,7 +2208,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
         return new Pair<String, String>(vmInternalCSName, vmNameOnVcenter);
     }
 
-    private static void configNestedHVSupport(VirtualMachineMO vmMo, VirtualMachineTO vmSpec, VirtualMachineConfigSpec vmConfigSpec) throws Exception {
+    protected void configNestedHVSupport(VirtualMachineMO vmMo, VirtualMachineTO vmSpec, VirtualMachineConfigSpec vmConfigSpec) throws Exception {
 
         VmwareContext context = vmMo.getContext();
         if ("true".equals(vmSpec.getDetails().get(VmDetailConstants.NESTED_VIRTUALIZATION_FLAG))) {
--- a/test/integration/smoke/test_nested_virtualization.py
+++ b/test/integration/smoke/test_nested_virtualization.py
@@ -1,0 +1,152 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+""" Tests for Nested Virtualization
+"""
+#Import Local Modules
+from marvin.codes import FAILED
+from marvin.cloudstackTestCase import cloudstackTestCase
+from marvin.lib.utils import (cleanup_resources,
+                              get_hypervisor_type,
+                              get_process_status)
+from marvin.lib.base import (Account,
+                             ServiceOffering,
+                             NetworkOffering,
+                             Configurations,
+                             VirtualMachine,
+                             Network)
+from marvin.lib.common import (get_zone,
+                               get_domain,
+                               get_template)
+from nose.plugins.attrib import attr
+from marvin.sshClient import SshClient
+import logging
+
+class TestNestedVirtualization(cloudstackTestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        testClient = super(TestNestedVirtualization, cls).getClsTestClient()
+        cls.apiclient = testClient.getApiClient()
+        cls.services = testClient.getParsedTestDataConfig()
+        
+        cls.logger = logging.getLogger('TestNestedVirtualization')
+        cls.stream_handler = logging.StreamHandler()
+        cls.logger.setLevel(logging.DEBUG)
+        cls.logger.addHandler(cls.stream_handler)
+
+        cls.zone = get_zone(cls.apiclient, testClient.getZoneForTests())
+        cls.services['mode'] = cls.zone.networktype
+        cls.services["isolated_network"]["zoneid"] = cls.zone.id
+        cls.domain = get_domain(cls.apiclient)
+        cls.service_offering = ServiceOffering.create(
+            cls.apiclient,
+            cls.services["service_offerings"]["tiny"]
+        )
+        cls.account = Account.create(cls.apiclient, services=cls.services["account"])
+        cls.template = get_template(
+            cls.apiclient,
+            cls.zone.id,
+            cls.services["ostype"]
+        )
+        cls.hypervisor = get_hypervisor_type(cls.apiclient)
+        
+        cls.isolated_network_offering = NetworkOffering.create(
+                                                cls.apiclient,
+                                                cls.services["isolated_network_offering"])
+        # Enable Isolated Network offering
+        cls.isolated_network_offering.update(cls.apiclient, state='Enabled')
+        
+        if cls.template == FAILED:
+            assert False, "get_template() failed to return template with description %s" % cls.services["ostype"]
+            
+        cls.services["small"]["zoneid"] = cls.zone.id
+        cls.services["small"]["template"] = cls.template.id
+
+        cls.cleanup = [cls.account]
+
+    @attr(tags=["advanced"], required_hardware="true")
+    def test_nested_virtualization_vmware(self):
+        """Test nested virtualization on Vmware hypervisor"""
+        if self.hypervisor.lower() not in ["vmware"]:
+             self.skipTest("Skipping test because suitable hypervisor/host not present")
+             
+        # 1) Update nested virtualization configurations, if needed
+        configs = Configurations.list(self.apiclient, name="vmware.nested.virtualization")
+        rollback_nv = False
+        rollback_nv_per_vm = False
+        for conf in configs:
+            if (conf.name == "vmware.nested.virtualization" and conf.value == "false"):
+                config_update = Configurations.update(self.apiclient, "vmware.nested.virtualization", "true")
+                self.logger.debug("Updated global setting vmware.nested.virtualization to true")
+                rollback_nv = True
+            elif (conf.name == "vmware.nested.virtualization.perVM" and conf.value == "false"):
+                config_update = Configurations.update(self.apiclient, "vmware.nested.virtualization.perVM", "true")
+                self.logger.debug("Updated global setting vmware.nested.virtualization.perVM to true")
+                rollback_nv_per_vm = True
+                
+        # 2) Deploy a vm
+        virtual_machine = VirtualMachine.create(
+            self.apiclient,
+            self.services["small"],
+            accountid=self.account.name,
+            domainid=self.account.domainid,
+            serviceofferingid=self.service_offering.id,
+            mode=self.services['mode']
+        )
+        self.assert_(virtual_machine is not None, "VM failed to deploy")
+        self.assert_(virtual_machine.state == 'Running', "VM is not running")
+        self.logger.debug("Deployed vm: %s" % virtual_machine.id)
+        
+        isolated_network = Network.create(
+            self.apiclient,
+            self.services["isolated_network"],
+            self.account.name,
+            self.account.domainid,
+            networkofferingid=self.isolated_network_offering.id)
+
+        virtual_machine.add_nic(self.apiclient, isolated_network.id)
+        
+        # 3) SSH into vm
+        ssh_client = virtual_machine.get_ssh_client()
+
+        if ssh_client:
+            # run ping test
+            result = ssh_client.execute("cat /proc/cpuinfo | grep flags")
+            self.logger.debug(result)
+        else:
+            self.fail("Failed to setup ssh connection to %s" % virtual_machine.public_ip)
+            
+        # 4) Revert configurations, if needed
+        if rollback_nv:
+            config_update = Configurations.update(self.apiclient, "vmware.nested.virtualization", "false")
+            self.logger.debug("Reverted global setting vmware.nested.virtualization back to false")
+        if rollback_nv_per_vm:
+            config_update = Configurations.update(self.apiclient, "vmware.nested.virtualization", "false")
+            self.logger.debug("Reverted global setting vmware.nested.virtualization.perVM back to false")
+            
+        #5) Check for CPU flags: vmx for Intel and svm for AMD indicates nested virtualization is enabled
+        self.assert_(result is not None, "Empty result for CPU flags")
+        res = str(result)
+        self.assertTrue('vmx' in res or 'svm' in res)
+
+    @classmethod
+    def tearDownClass(cls):
+        try:
+            cleanup_resources(cls.apiclient, cls.cleanup)
+        except Exception, e:
+            raise Exception("Cleanup failed with %s" % e)
+    
--- a/plugins/hypervisors/vmware/src/com/cloud/hypervisor/guru/VMwareGuru.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/hypervisor/guru/VMwareGuru.java
@@ -31,13 +31,13 @@ import org.apache.cloudstack.engine.subsystem.api.storage.VolumeDataFactory;
 import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
 import org.apache.cloudstack.framework.config.ConfigKey;
 import org.apache.cloudstack.framework.config.Configurable;
-import org.apache.cloudstack.framework.config.dao.ConfigurationDao;
 import org.apache.cloudstack.storage.command.CopyCommand;
 import org.apache.cloudstack.storage.command.DeleteCommand;
 import org.apache.cloudstack.storage.command.StorageSubSystemCommand;
 import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
 import org.apache.cloudstack.storage.to.VolumeObjectTO;
+import org.apache.commons.lang.BooleanUtils;
 import org.apache.log4j.Logger;
 
 import com.cloud.agent.api.BackupSnapshotCommand;
@@ -57,7 +57,6 @@ import com.cloud.agent.api.to.DiskTO;
 import com.cloud.agent.api.to.NicTO;
 import com.cloud.agent.api.to.VirtualMachineTO;
 import com.cloud.cluster.ClusterManager;
-import com.cloud.configuration.Config;
 import com.cloud.exception.InsufficientAddressCapacityException;
 import com.cloud.host.Host;
 import com.cloud.host.HostVO;
@@ -127,8 +126,6 @@ public class VMwareGuru extends HypervisorGuruBase implements HypervisorGuru, Co
     @Inject
     private NetworkModel _networkMgr;
     @Inject
-    private ConfigurationDao _configDao;
-    @Inject
     private NicDao _nicDao;
     @Inject
     private PhysicalNetworkTrafficTypeDao _physicalNetworkTrafficTypeDao;
@@ -154,6 +151,12 @@ public class VMwareGuru extends HypervisorGuruBase implements HypervisorGuru, Co
     public static final ConfigKey<Boolean> VmwareReserveMemory = new ConfigKey<Boolean>(Boolean.class, "vmware.reserve.mem", "Advanced", "false",
         "Specify whether or not to reserve memory when not overprovisioning, In case of memory overprovisioning we will always reserve memory.", true,
         ConfigKey.Scope.Cluster, null);
+
+    protected ConfigKey<Boolean> VmwareEnableNestedVirtualization = new ConfigKey<Boolean>(Boolean.class, "vmware.nested.virtualization", "Advanced", "false",
+            "When set to true this will enable nested virtualization when this is supported by the hypervisor", true, ConfigKey.Scope.Global, null);
+
+    protected ConfigKey<Boolean> VmwareEnableNestedVirtualizationPerVM = new ConfigKey<Boolean>(Boolean.class, "vmware.nested.virtualization.perVM", "Advanced", "false",
+            "When set to true this will enable nested virtualization per vm", true, ConfigKey.Scope.Global, null);
 
     @Override
     public HypervisorType getHypervisorType() {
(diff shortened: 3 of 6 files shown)


DaanHoogland:
@nvazquez I like the functional description and will probably run through the code over the weekend but one thing struck me. A lot of this should apply equally for other hypervisors, shouldn't it? Did you compare design or have a reason to implement it this VMWare specific?

nvazquez:
Hi @DaanHoogland, unfortunately I really didn't examine it for other hypervisors. In our environment we only use VMware ESXi hypervisor, that's the main reason for implementing it only in VMware code

DaanHoogland:
:) no worries, just expect some review comments to this extend ;)

remibergsma:
KVM also has this feature and there you can enable/disable it in the agent.properties file (ant not per VM). We should try to make the code as generic as possible, instead of implementing it like this IMHO.

Why do you want to control this per VM by the way?

ustcweizhou:
@nvazquez Can you use ConfigKey instead of changes on Config.java and schema-481to490.sql ?

serg38:
@remibergsma In some private cloud environments nested virtualization is not a good fit so it is generally disabled. That's fine but we found that it would be very nice to have a way to support exceptions e.g. to support users to run Android emulators that as it turned out needs nested virtualization.

serg38:
@remibergsma, @DaanHoogland Current global support for nested virtualization is only vmware specific and done in VmwareGuru.  
Are you proposing to move the logic up to the HypervisorGuru and change vmware.nested.virtualization to just nested.virtualization ? Then the implementation can be done in respective Resources.

nvazquez:
Hi @ustcweizhou @DaanHoogland @remibergsma,

I moved global config definitions to `ConfigKey.java` instead of `Config.java` and `schema-481to490.sql`.

I also created a JIRA ticket https://issues.apache.org/jira/browse/CLOUDSTACK-9391 to extend this functionality to any hypervisor, I can work on it and create a separate pull request, what do you think?

DaanHoogland:
@nvazquez sounds great, keep us posted

nvazquez:
Could this be reviewed for VMware please? Then I can work on extending it to other HVs

Thanks!

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 198
 Hypervisor xenserver
 NetworkType Advanced
 Passed=73
 Failed=0
 Skipped=3

**Failed tests:**

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vpc_vpn.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

nvazquez:
Hi @DaanHoogland @remibergsma @ustcweizhou,

I rebased master branch for this pull request as it's been opened for a while, could it be reviewed please?

DaanHoogland:
@nvazquez unfortunately I no longer am involved in cloud computing. I'd really need a deep dive to give value, sorry.

nvazquez:
Oh that's sad, no problem, thanks anyways @DaanHoogland

rafaelweingartner:
@nvazquez long time we don’t do this ;)

First of all, your PR explanation is great. The code is very well documented and explained and with test cases (very good ones). Congratulations. I really like your work. 

I have some suggestions for you, though.

I think that the method “enableNestedVirtualization” could return a Boolean. I see no need to transform the Boolean in a String there; It seems better to use Boolean.toString(bool) at line 376. Moreover, when reading the method “enableNestedVirtualization”, I felt like it was going to enable something on the VM, which is not the case. This looks more a “should”, “can”, “has” type of method. I mean, it is a method that checks if something has to/should/can be done; in this case, the enabling of nested virtualization. Therefore, I think names such as “canUseNestedVirtualization”, “should enableNestedVirtualization”, “hasToEnableNestedVirtualization” seem more appropriate.

I also think that the code has room for improvements. First, to reduce the cyclomatic complexity, you can invert the first if, which become something like this:
```
if (globalNestedV == null || globalNestedVPerVM == null) {
    return false;
}
```
Then, the other conditional can be further improved. It is a bit complicated. Something like this would have the same result:
```
if (globalNVPVM) {
            return (localNestedV == null && globalNV) || BooleanUtils.toBoolean(localNestedV);
        }
  return globalNV;
```

On method “configNestedVirtualization”, I would just suggest using the word “configure” instead of “config”. At least for me, when I read config, I think configuration and not configure (this is a very personal opinion, so if you are ok with config, be my guest). A method, for me, means an action that is executed, so it seems a better fit the word “configure” (verb).

The method “testConfigNestedVirtualization”, I think it should check if the “VmDetailConstants.NESTED_VIRTUALIZATION_FLAG” flag (parameter) is being loaded properly from the “vmDetails”. I also suggest you using the “inOrder” to verify the calls in order. If the order of the calls changes, the behavior of the method changes too, right?

About the method “enableNestedVirtualizationBaseTest”, I think it could be a little bit more self-explaining, such as: executeAndVerify<nameOfTheMethod>Test.

And finally, about the others test methods, I think instead of TFT, TFN, and others at the end, I think if you were a bit more literal, and self-explaining, it would be better. For instance, the test method “testEnableNestedVirtualizationCaseTFF”, in a more detailed version, could be read as “testEnableNestedVirtualizationCaseGlobalNvTrueGlobalNvPvFalseLocalVmfalse”. I know it is a huge method name, but I think it facilitates for newcomers and also for the @nvazquez of the future ;)

nvazquez:
@rafaelweingartner thanks a lot for your help as always!

I refactored based on your comments. I really liked your review, especially your refactor on `shouldEnableNestedVirtualization`, it really simplifies method's understanding. I also included all your suggestions on method's names and in-order verification.

I had to make some changes on `VmwareGuruTest` using reflection to include mocks in configuration keys variables on `VmwareGuru`. What do you think about this solution?

Thanks again for taking the time to review this PR

rafaelweingartner:
@nvazquez  the code is great now.

I have just a concern about that hack to change the field declared as final.
In my opinion, the final there does not bring benefits. So, I believe it can be removed. 
I have seen people using final a lot, but I think that final should be used only when we have 100% certainty that it is needed; otherwise, it complicates our lives when writing test cases or while extending some method/class.

What do you think?

nvazquez:
@rafaelweingartner I agree it wasn't the best solution but I meant to include it for the sake of testing `configureNestedVirtualization` method. Which approach can I take to test this method, as it uses configuration final variables to determinate enabling/desabling nested virtualization?

rafaelweingartner:
I understand the need to create a test. And I agree that it deserves a test. What I meant was the following.

IMHO, you could simply remove the final keyword of the variable declaration. Sometimes in the ACS code, I have seen people using some design/ways of coding without having a good and well-thought explanation for them (the use of final is one good example). For instance, variables “VmwareEnableNestedVirtualization” and “VmwareEnableNestedVirtualizationPerVM” are only used in the “VMwareGuru” class. They do not need to be either static or public. Moreover, there is no need for those variables to be final. 

If those variables were protected (only), it would be much easier to write test cases.

nvazquez:
Thanks @rafaelweingartner, I agree with you, I refactored it!

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2184,7 +2184,7 @@ int getReservedCpuMHZ(VirtualMachineTO vmSpec) {
         return new Pair<String, String>(vmInternalCSName, vmNameOnVcenter);
     }
 
-    private static void configNestedHVSupport(VirtualMachineMO vmMo, VirtualMachineTO vmSpec, VirtualMachineConfigSpec vmConfigSpec) throws Exception {
+    protected static void configNestedHVSupport(VirtualMachineMO vmMo, VirtualMachineTO vmSpec, VirtualMachineConfigSpec vmConfigSpec) throws Exception {
```
rafaelweingartner: This method does not need to be static.

rafaelweingartner:
@nvazquez, it seems that there was a problem with the Jenkins Job. Can you force another build?

nvazquez:
Sure, done

rafaelweingartner:
There is still one point I would like to highlight. The method VmwareResource.configNestedHVSupport does not need to be static.

nvazquez:
I had removed it, I saw your comment in that file and I included it in last refactor, I forgot to mention it in my last comment :)

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
nvazquez: Thanks, done

rafaelweingartner:
Great.I am ok with the code. I give my LGTM here.
I have one extra question, though.  How do you intend to change that parameter (NESTED_VIRTUALIZATION_FLAG) to tell which VM should have the nested virtualization enabled? I am assuming that there is no option through the ACS API, is that right? If so, you are going to change the parameter directly in the database, right?

serg38:
@rafaelweingartner That's correct. At this point advanced VM details and advanced template details have no UI exposure. At some point we can bring in PR to show this in UI and allow API based changes

rafaelweingartner:
Great. You have done a lot of work already; but, just to make sure things do not get lost. What about opening a Jira ticket to expose that customization through the ACS API?

nvazquez:
Thanks @rafaelweingartner 
This is the link to JIRA ticket: https://issues.apache.org/jira/browse/CLOUDSTACK-9457

rafaelweingartner:
I am the one thas has to thank you @nvazquez, great work!

serg38:
@jburwell @rhtyd @koushik-das Can you review this PR?

rhtyd:
@nvazquez @serg38 how about we remove vmware.nested.virtualization.perVM and instead if the user vm detail exists and is true, it overrides the global setting?

serg38:
@rhtyd @jburwell @rafaelweingartner @koushik-das The rationale to have vmware.nested.virtualization.perVM was that advanced vm_details and template details are user controlled. With https://issues.apache.org/jira/browse/CLOUDSTACK-9457
‘Allow retrieval and modification of VM and template details via API and UI’ users will be able to change any advanced details in UI and API.
If admin wants to preserve that nested virtualization is controlled in centralized fashion they can do so by setting vmware.nested.virtualization.perVM to ‘false’. Idea was to allow organizations that currently have nested virtualization disabled to have it so after upgrade. Does it make sense or everyone prefers that setting on VM level overrides global one?

serg38:
@rhtyd @jburwell @rafaelweingartner @koushik-das Can we agree if this PR is OK as it is ?

rafaelweingartner:
@serg38 with your explanations, I am ok with the current state of this PR. No further changes are needed.
There is only one problem now. There are some merge conflicts that have to be resolved.

nvazquez:
Solved, thanks @rafaelweingartner

serg38:
@rafaelweingartner @jburwell @karuturi Looks like this PR has everything to get merged. Can we do it by the release freeze?

jburwell:
@serg38 @nvazquez I apologize that I am late to the review.  I would like to see a Marvin test added that verifies that nested virtualization works when the global setting is enabled (i.e. successfully create a VM within a VM).  Would that be possible?

serg38:
@jburwell Creating Marvin test for this PR will be a huge undertaking. It would require building and configuring a hypervisor in a deployed VM and then creating VM in VM. Considering storage, environmental and network implication the test will be complicated and error prone. An alternative approach would require integrating vSphere SDK with Marvin python libraries and query vSphere for the actual nested virtualization settings from a deployed VM. The effort doesn't seem to be warranted for such a minor change. As long as standard integration suite is passing this shouldn't affect any existing functionality. What do you think?

jburwell:
@serg38 the goal is not only verify that functionality works now, but in the future to protect against regressions.  My thinking is that we could deploy a CentOS VM, install KVM, and verify that we can create an ephemeral VM within it.  A specially packaged template could have the KVM packages pre-installed and a small Debian or Alpine template on-board.  The test would pass when it is able to SSH into the nested VM over the link-local interface.  Such a test would be self-contained and not require any particular storage or network settings.  Does such a test seem reasonable and valuable?

serg38:
@jburwell The only BUILTIN Centos template for Vmware is 5.3 that doesn't have KVM support to my knowledge. Are you proposing to introduce another BUILTIN template so that it is available for anyone to use during the integration testing? We don't have any KVM expertise at this point so this will be next to mission impossible. The only option for us would be to package vSphere python SDK and try to talk to vCenter directly.

jburwell:
@rhytd can you help @serg38 craft a template with KVM installed and the virsh commands to create a nested VM?

serg38:
@jburwell @rhtyd Thanks in advance

jburwell:
@serg38 no problem

rhtyd:
@jburwell nested VM is something tricky to test, we can support this as a subtle feature but not a generally supported one. In our test/trillian environment, we're already running nested virtualization. I'm not sure investing time on inception of a test environment is useful. The feature is specific to vmware, I don't understand why KVM needs to be involved.

A quick test could be to run a guest VM on VMware with specific cpu flags (say vmx, smx, ht) and verify that the VM launched with these flags have the cpu features for example, the test could do: (1)ssh into the vm, (2) cat /proc/cpuinfo and (3) grep `flags` for specific cpu flags such as vmx, ht etc. I think such a test should work with the default centos template or with macchanina.

The feature could also be written and tested in such a way to add `any` key/value details on a VM and be interpreted by the hypervisor specific resource.

serg38:
Great tip, @rhtyd This way it is actually very straightforward. We should be able to come up with integration test for this PR shortly

jburwell:
@rhtyd good idea to check the CPU flags.  There is a part of me that would like to see a VM actually spin up inside a VM, but there are also many things that could go wrong that are not related to CloudStack.  Plus, it would be slow.  Therefore, checking the CPU flags is an excellent compromise.  

Another thing that I see in this PR is that nested virtualization seems like a capability that should be more generally expressed and managed in CloudStack's core abstractions.  @rhtyd could you open a ticket to extract the aspects of this enhancement that are common across all hypervisors that support nested virtualization?

serg38:
@jburwell This sounds like a good idea. In this PR all abstraction work is done so the things to do would be to implement support in respective hypervisor resources as well as strip "vmware." prefix from conjuration items. @nvazquez is preparing intragrtion test based on @rhtyd idea

nvazquez:
@jburwell @rhtyd I included new marvin test for nested virtualization, thanks @rhtyd for your idea!

test/integration/smoke/test_nested_virtualization.py
```diff
@@ -0,0 +1,152 @@
...
+            config_update = Configurations.update(self.apiclient, "vmware.nested.virtualization", "false")
+            self.logger.debug("Reverted global setting vmware.nested.virtualization back to false")
+        if rollback_nv_per_vm:
+            config_update = Configurations.update(self.apiclient, "vmware.nested.virtualization", "false")
+            self.logger.debug("Reverted global setting vmware.nested.virtualization.perVM back to false")
+            
+        #5) Check for CPU flags: vmx for Intel and svm for AMD indicates nested virtualization is enabled
+        self.assert_(result is not None, "Empty result for CPU flags")
+        res = str(result)
+        self.assertTrue(res.find('vmx') != -1 or res.find('svm') != -1)
```
rhtyd: How about we make it idiomatic with: self.assertTrue('vmx' in res or 'svm' in res)

rhtyd:
@nvazquez you're welcome. If you've some bandwidth, can you run your test against this template which will also help validate the new vmware test appliance: https://github.com/rhtyd/tinylinux

nvazquez:
@rhtyd sure, will do and post results

test/integration/smoke/test_nested_virtualization.py
nvazquez: Done, thanks

nvazquez:
Hi @rhtyd, we ran `test_internal_lb.py` and `test_vpc_vpn.py` using `tl-vmware.ova` template and got this results:

`test_vpc_vpn` is ok:
```
Test Site 2 Site VPN Across redundant VPCs ... === TestName: test_01_redundant_vpc_site2site_vpn | Status : SUCCESS ===
ok
Test Remote Access VPN in VPC ... === TestName: test_01_vpc_remote_access_vpn | Status : SUCCESS ===
ok
Test Site 2 Site VPN Across VPCs ... === TestName: test_01_vpc_site2site_vpn | Status : SUCCESS ===
ok
 
Ran 3 tests in 1949.605s
 
OK
```

However, `test_internal_lb` fails, we could find some issues:
- /sbin/ip is needed but not present in template {Cmd: /sbin/ip addr show eth0 |grep 'inet '| cut -f6 -d' ' > /tmp/index.html via Host: 10.140.12.107} {returns: [u'sh: /sbin/ip: not found']}
- 2 more tests fail due to a difference in wget on new template: {Cmd: /usr/bin/wget -T3 -qO- --user=XX --password=XX http://XXXX:8081/admin?stats via Host: 10.140.12.106} {returns: [u"/usr/bin/wget: unrecognized option '..
- We have some strange erros in this functions which require more arguments than given but couldn't find out why
```
2016-10-02 22:13:20,881 - CRITICAL - EXCEPTION: execute_internallb_haproxy_tests: ['Traceback (most recent call last):\n', '  File "/usr/local/lib/python2.7/unittest/case.py", line 331, in run\n    testMethod()\n', 'TypeError: execute_internallb_haproxy_tests() takes exactly 2 arguments (1 given)\n']
2016-10-02 22:13:20,883 - CRITICAL - EXCEPTION: execute_internallb_roundrobin_tests: ['Traceback (most recent call last):\n', '  File "/usr/local/lib/python2.7/unittest/case.py", line 331, in run\n    testMethod()\n', 'TypeError: execute_internallb_roundrobin_tests() takes exactly 2 arguments (1 given)\n']
2016-10-02 22:13:20,885 - CRITICAL - EXCEPTION: run_ssh_test_accross_hosts: ['Traceback (most recent call last):\n', '  File "/usr/local/lib/python2.7/unittest/case.py", line 331, in run\n    testMethod()\n', 'TypeError: run_ssh_test_accross_hosts() takes at least 3 arguments (1 given)\n']
```

rhtyd:
Thanks for testing @nvazquez I think since @NuxRo built new vmware templates, we can consume macchinina/vmware: http://dl.openvm.eu/cloudstack/macchinina/x86_64/testing

The TinyCore based template I had created lacks 'ip/iproute' package and httpd, I'll try to fix it though as this point since Lucian has already fixed macchinina it might be not necessary.

jburwell:
@nvazquez I would like to get this PR merged for 4.10.0.0.  Could you please rebase this PR to pull the latest Marvin enhancements/fixes?  Once that is done, I will kick off a regression test in preparation to merge it.

rhtyd:
@blueorangutan package

rhtyd:
The debian pkg failed due to an intermittent unit test failure (ignorable, env issue).
@blueorangutan test

nvazquez:
Thanks @jburwell! Rebased

jburwell:
@nvazquez @rhtyd is working on an enhancement to blueoragnutan that will allow to specify component tests to run in addition to the smoke tests.  When that is ready, we will re-execute the smoke tests with the new component test introduced by the PR.

nvazquez:
That's great, thanks!

serg38:
@karuturi @jburwell As per @rhtyd we can run additional tests with blueorangutan . Can we execute them and merge this PR since it  passes the  standard test suite just fine?

jburwell:
@blueorangutan test centos7 vmware-55u3 test_nested_virtualization.py

serg38:
@jburwell @rhtyd Looks like blueorangutain vmware side has issues on hypervisor. Can you take a look and then re-kick the tests? Some errors were due to that.

530, errortext : u'Unable to add NIC to VM[User|i-39-54-VM]: com.cloud.exception.AgentUnavailableException: Resource [Host:1] is unreachable: Host 1: Unable to start instance due to Unable to start  VM:f22d93f6-e197-46e4-8e0c-0a926fece518 due to error in finalizeStart, not retrying'},

jburwell:
@serg38 these errors look similar to those we have seen on other PRs.  We are currently investigating them with the assumption that they are most likely environmental.  Once we have them resolved, we will re-kick the tests.

/cc @PaulAngus @murali-reddy @rhtyd

rhtyd:
@blueorangutan package

rhtyd:

serg38:
@rhytd @jburwell  @karuturi          test_nested_virtualization_vmware	Success	305.03

The other tests fail due to environmental issues e.g. below. I think it is ready for merging now.

2016-11-20 16:32:00,324 - CRITICAL - EXCEPTION: test_3d_gpu_support: ['Traceback (most recent call last):\n', '  File "/usr/lib64/python2.7/unittest/case.py", line 369, in run\n    testMethod()\n', '  File "/marvin/tests/smoke/test_deploy_vgpu_enabled_vm.py", line 316, in test_3d_gpu_support\n    self.virtual_machine.add_nic(self.apiclient, self.isolated_network.id)\n', '  File "/usr/lib/python2.7/site-packages/marvin/lib/base.py", line 784, in add_nic\n    return apiclient.addNicToVirtualMachine(cmd)\n', '  File "/usr/lib/python2.7/site-packages/marvin/cloudstackAPI/cloudstackAPIClient.py", line 637, in addNicToVirtualMachine\n    response = self.connection.marvinRequest(command, response_type=response, method=method)\n', '  File "/usr/lib/python2.7/site-packages/marvin/cloudstackConnection.py", line 379, in marvinRequest\n    raise e\n', "Exception: Job failed: {jobprocstatus : 0, created : u'2016-11-20T16:29:43+0000', cmd : u'org.apache.cloudstack.api.command.admin.vm.AddNicToVMCmdByAdmin', userid : u'1575bde5-af3a-11e6-b333-06aa7801071a', jobstatus : 2, jobid : u'78b21bfa-bc82-4df6-ad54-67ef0899ceeb', jobresultcode : 530, jobresulttype : u'object', jobresult : {errorcode : 530, errortext : u'Unable to add NIC to VM[User|i-7-9-VM]: com.cloud.exception.AgentUnavailableException: **Resource [Host:1] is unreachable:** Host 1: Unable to start instance due to Unable to start  VM:a42c5549-c3c4-47b3-8e49-0785d1c2ef44 due to error in finalizeStart, not retrying'}, accountid : u'1575a6c7-af3a-11e6-b333-06aa7801071a'}\n"]

rhtyd:
Based on test results LGTM. /cc @jburwell 
Can we have one more LGTM on this, @serg38 @jburwell ?

serg38:
LGTM

rhtyd:
LGTM, merging based on code reviews and tests, ignoring known intermittent errors in vpc/rvr.

PR #1546, opened 2016-05-13
[WIP][4.10+/long-term] Moving towards Java8 and Spring4

rhtyd:
This is a `long term` change initiative.

Goals:
- [x] Java8 for building and running CloudStack
- [x] Migrate to Spring 4.x
- [x] Use latest Jetty during development
- [ ] Deprecate Tomcat and migrate to embedded Jetty for production
- [ ] Fix any Spring 4.x migration issue
- [ ] Fix rpm/debian packaging wrt Java8 and systemd
- [ ] Build new systemvm template with Java8 runtime
- [ ] Migrate to KVM/packer for building systemvmtemplate
- [ ] Migrate to Debian8 based systemvm template

Outstanding issues:
- [ ] F5 fails to compile with Java8. Contact F5 authors/maintainers for resolution.
- [ ] Fix failing unit test com/cloud/vm/DeploymentPlanningManagerImplTest.java

Testing:
- [x] TravisCI can build/run/test management server with Java8
- [ ] End to end regression testing
- [ ] Functional testing of all the existing plugins
- [ ] Testing CloudStack long-running services under Java8 runtime
- [ ] Test SystemVM agent with JRE8


rhtyd:
Request for comments - @jburwell @DaanHoogland @wido @DaanHoogland @abhinandanprateek @imduffy15 @kiwiflyer @terbolous and others

wido:
So, this means we will drop Ubuntu 14.04 support. There is no Java 8 available in those repositories

rhtyd:
@wido it's true that the package is not in the default, but Java8 is available through few popular ppa(s). We need to migrate to Java8 since Java7 has EOL-ed for so long, if that means dropping support for distro that don't have it -- then yes.

jburwell:
@rhtyd I think it would be a good idea to this PR up a bit.  I see three distinct pieces:

1. **Spring 4 Upgrade**:  Spring 4 supports both Java7 and Java8.  Therefore, this upgrade can be done without involving Java8.
2. **Jetty Embedding**:  Again, Jetty 9 supports both Java7 and Java8.  Therefore, this work can be done without involving Java8.
3. **Java8**: Ideally, this piece should be fairly small -- disabling the F5 plugin, changing the POMs, and a sundry of minor code changes to achieve compilation.

The harder part is that all of these changes will require a full end-to-end test of the system.

Finally, for Java8, I suggest that we follow the convention of many of Java-based projects and remove the package dependencies on the JDK.  Some users choose to use OpenJDK and others choose Oracle's JDK.  Therefore, the current dependency makes an invalid assumption about the JDK users want to use -- making it more difficult to use the implementation of their choice.

@wido as @rhtyd points out, Java8 is available via PPAs.  IMHO, Ubuntu's decision to keep their distribution out-of-date is not a sufficient reason to continue requiring the use of an unsupported Java runtime.  Given the number of significant security issues that have been discovered in the JRE over the past few years, the longer we wait to upgrade to Java8, the greater the risk that a significant security issue arises that forces CloudStack users to either turn off their instances or run and hope no one hacks them.

rhtyd:
@jburwell Thanks, I'll split the changes into separate PR if everyone demands that. Otherwise each distinct pieces are implemented as separate commits that can be reviewed per commit basis as well. Since, this is long term I'm hoping no testing/merge activity until at least 4.9.0 releases.

While some of the tasks don't require Java8 upgrade, I started working on this for the sole purpose of moving to Java8 because as a developer I want to start using Java8 specific features in CloudStack and that Java7 has reached EOL. I've also listed among these pieces, certain other long term goals around systemvmtemplates, systemd/packaging and reducing distro specific dependencies.

resmo:
We use, when ever possible, openjdk as we get support from the distribution. Support for openjdk be it 7 or 8, manually patching oraclejdk is time consuming. 

RHEL support both openjdk 7 (June 2018) and 8 (October 2020) (see https://access.redhat.com/articles/1299013), to me it would make sense to keep openjdk 8 support in cloudstack.

It seems, travis does not support openjdk 8 yet.

rhtyd:
@resmo have you found if Travis specific reliance on oraclejdk7/8 causes any issue with building/testing CloudStack?

resmo:
@rhtyd I am not aware of any no. Just wanted to point out that oracle "sucks" and openjdk should be the jdk that should be tested and supported in the first place as it became the default in all distributions. Same for mariadb btw.

rashmidixit:
While we are on the topic of updating the tech stack, have we considered moving to using either LESS or Sass for CSS. CSS pre-processors bring more modularity and clarity into the CSS code.

abhinandanprateek:
With java 8 we will also get enhanced testing frameworks, specially junit-quickcheck, that will help us in writing more expressive and compact test cases.

rhtyd:
I'll split this PR.

jburwell:
@rhtyd Can we close this PR in favor of #1638?

rhtyd:
Spring4+jdk8 migration successful now with master, closing this.

PR #1552, opened 2016-05-18
add DHCP lease folders for Ubuntu

romain-dartigues:
Add `/var/lib/dhcp/*` to the search path for Ubuntu; tested under Ubuntu Precise, Trusty and Xenial.

### DIFF
--- a/setup/bindir/cloud-set-guest-sshkey.in
+++ b/setup/bindir/cloud-set-guest-sshkey.in
@@ -27,7 +27,7 @@
 user=root
 
 # Add your DHCP lease folders here
-DHCP_FOLDERS="/var/lib/dhclient/* /var/lib/dhcp3/*"
+DHCP_FOLDERS="/var/lib/dhclient/* /var/lib/dhcp3/* /var/lib/dhcp/*"
 keys_received=0
 file_count=0
 


wido:
Good catch! I can confirm that under Ubuntu 16.04 (Xenial) that is where the DHCP leases are stored.

<pre>root@example:~# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04 LTS
Release:	16.04
Codename:	xenial
root@example:~# ls /var/lib/dhcp/
dhclient.ens3.leases  dhclient6.ens3.leases
root@example:~#</pre>

LGTM for me

syed:
Can cofirm the dhcp folder on Ubuntu Trusty LTS (14.04) LGTM.

```
 $ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 14.04.1 LTS
Release:	14.04
Codename:	trusty

$ls  /var/lib/dhcp/
dhclient.leases

```

swill:
Does this need CI?

wido:
@swill I don't think so, CI doesn't test this feature. It is a script for INSIDE Guest Instances.

swill:
Ok, so I have two LGTM both with validation that the code behaves as expected.  I think that is enough for this to get in.  Thanks...

PR #1557, opened 2016-05-21
Removed unused methods from XenServerConnectionPool

rodrigo93:
Removed the following methods that are unused from the class
_com.cloud.hypervisor.xenserver.resource.XenServerConnectionPool_:
- static void **forceSleep**(long sec)  
- protected Session **slaveLocalLoginWithPassword**(Connection conn, String
  username, Queue<String> password)  
- static public Pool.Record **getPoolRecord**(Connection conn) throws
  XmlRpcException, XenAPIException

From XenServerConnectionPool.XenServerConnection:
- public String **getUsername**()
- public Queue<String> **getPassword**()  
- protected Map **dispatch**(String methodcall, Object[] methodparams)

Removed unused attributes:
- protected static long **s_sleepOnError**

I also removed the lines from 58 to 74 since they were never used (just logging, as @rafaelweingartner  mentioned).

**I would like to make a note about the "TrustAllManager" static class
inside the XenServerConnectionPool class. I noticed methods from TrustAllManager class returns true or "nothing". This can be a security problem, since it returns true without any
verification.**

### DIFF
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/XenServerConnectionPool.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/XenServerConnectionPool.java
@@ -16,8 +16,21 @@
 // under the License.
 package com.cloud.hypervisor.xenserver.resource;
 
-import com.cloud.utils.NumbersUtil;
-import com.cloud.utils.PropertiesUtil;
+import java.net.URL;
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
+import java.util.HashMap;
+import java.util.Queue;
+
+import javax.net.ssl.HostnameVerifier;
+import javax.net.ssl.HttpsURLConnection;
+import javax.net.ssl.SSLSession;
+
+import org.apache.cloudstack.utils.security.SSLUtils;
+import org.apache.cloudstack.utils.security.SecureSSLSocketFactory;
+import org.apache.log4j.Logger;
+import org.apache.xmlrpc.XmlRpcException;
+
 import com.cloud.utils.exception.CloudRuntimeException;
 import com.xensource.xenapi.APIVersion;
 import com.xensource.xenapi.Connection;
@@ -27,25 +40,6 @@ import com.xensource.xenapi.Session;
 import com.xensource.xenapi.Types;
 import com.xensource.xenapi.Types.BadServerResponse;
 import com.xensource.xenapi.Types.XenAPIException;
-import org.apache.cloudstack.utils.security.SSLUtils;
-import org.apache.cloudstack.utils.security.SecureSSLSocketFactory;
-import org.apache.log4j.Logger;
-import org.apache.xmlrpc.XmlRpcException;
-import org.apache.xmlrpc.client.XmlRpcClientException;
-
-import javax.net.ssl.HostnameVerifier;
-import javax.net.ssl.HttpsURLConnection;
-import javax.net.ssl.SSLSession;
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.net.URL;
-import java.security.KeyManagementException;
-import java.security.NoSuchAlgorithmException;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Queue;
 
 public class XenServerConnectionPool {
     private static final Logger s_logger = Logger.getLogger(XenServerConnectionPool.class);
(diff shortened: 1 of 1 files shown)


rafaelweingartner:
@rodrigo93, 
I was reviewing your PR, nice work; but, I think you can remove some other lines too.

I checked and the variable “s_sleepOnError ” is never used. I mean, it is used at lines 68 and 70; but, that is all (just logging). Also, that property “sleep.interval.on.error” from which a new value is assigned to “s_sleepOnError ” is never referenced anywhere else. Therefore, the “s_sleepOnError ” variable can be removed, and the code from lines 60 – 76 can also be removed.

rodrigo93:
Hi @rafaelweingartner ,
Thanks for the advice. I will take a look on that and do the proper changes as soon as I can.

rodrigo93:
Hello again @rafaelweingartner,
I did as you suggested. I also checked that the variable **file** was not used anywhere else besides lines 60 - 76, so I also removed it.

Thanks for the suggestion.

rafaelweingartner:
@rodrigo93 thanks.
@DaanHoogland, @swill, @rhtyd can someone help us here with the note @rodrigo93 has highlighted?

I have sent an email to the security@cloudstack.apache.org that presents our perception of that, but we did not receive any feedback.

DaanHoogland:
@rodrigo93 @rafaelweingartner Do you have a scenario that proves these methods to be a problem? returning true or null (or void) does not imply a problem unless the result can be abused.

rafaelweingartner:
@DaanHoogland are you asking about the code at lines 53-70?

DaanHoogland:
No I am talking about the problems with TrustAllManager. I am not convinced there is a problem by the not by @rodrigo93 .

rafaelweingartner:
Hi @Daan,
We are also not sure, that is why we have sent an email to the security maligning list to start a discussion about that. The question regarding the “com.cloud.hypervisor.xenserver.resource.XenServerConnectionPool. "TrustAllManager" class was the following. 

> The code at line 77-96 is setting the default SSL socket factory for HTTPS connection to an implementation (TrustAllManager class created at lines 483- 506) that always considers the certificate of the server valid (without any checks). Also, at line 90, it was overridden the host name verifier (the check that matches hostname and certificates CN) to always returns true. 

We also mentioned that:

> We might be misinterpreting the java doc of methods “javax.net.ssl.HttpsURLConnection.setDefaultSSLSocketFactory(SSLSocketFactory)” and “javax.net.ssl.HttpsURLConnection.setDefaultHostnameVerifier(HostnameVerifier)”, but those sets seem to affect every single HTTPS request that is performed by ACS (JVM runtime in which those methods were executed). We do not know how many HTTPS requests, and to what extent they are essential to ACS workings, but that code seems to make ACS vulnerable to man in the middle attacks (if it uses HTTPS requests to execute some of its tasks).

Speculating a bit, we had an idea why that code was added:

> We believe that was done in order to allow the use of the XAPI via rest API through HTTPS, probably because the host is using self-signed certificates. If that is the case, I think we have better options to work around that, such as just disabling the check for those specific requests and or using a keystore with the hosts’ public keys that could be downloaded at the host addition process.

In conclusion, if ACS uses an HTTPS connection to either download a file (template or something else) or to send a command to a remote region over the internet that connection/request may be vulnerable to MID attacks. Even though we will have an SSL connection, we will not know with whom, because the certificate checks were all disabled.

Again I would like to say that we do not know how many (if) ACS uses much HTTPS request to execute some of its tasks; that is why we raised the question.

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 188
 Hypervisor xenserver
 NetworkType Advanced
 Passed=73
 Failed=0
 Skipped=3

**Failed tests:**

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vpc_vpn.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

DaanHoogland:
@rodrigo93 please rebase and re-open if still relevant

PR #1561, opened 2016-05-25
CLOUDSTACK-9388: Remove string conversion in assertion statement

sanju1010:
Remove string convertion in Assertion statement, since the start port parameter in listFirewallAPI response is of type integer

Test Result:
=========
"Checking firewall rules deletion after static NAT disable ... === TestName: test_01_firewall_rules_port_fw | Status : SUCCESS ===
ok

Ran 1 test in 153.974s

OK

### DIFF
--- a/test/integration/component/test_blocker_bugs.py
+++ b/test/integration/component/test_blocker_bugs.py
@@ -409,7 +409,7 @@ class TestNATRules(cloudstackTestCase):
                         )
         self.assertEqual(
                             firewall_response[0].startport,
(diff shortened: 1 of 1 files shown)


pavanb018:
Very small change , conversion of firewall port to string was not necessary and that seems to be removed. Looks Good to me !!

koushik-das:
Simple change. LGTM

swill:
@sanju1010 thanks for posting the test results with this change.  This one should be good now.  Thanks everyone...

PR #1569, opened 2016-05-28
fix noredist build because of missing maven dependency of vmware 6.0 lib

pdion891:
fix noredist build for new vmware lib

### DIFF
--- a/deps/install-non-oss.sh
+++ b/deps/install-non-oss.sh
@@ -35,3 +35,6 @@ mvn install:install-file -Dfile=vim25_51.jar        -DgroupId=com.cloud.com.vmwa
 
 # From https://my.vmware.com/group/vmware/get-download?downloadGroup=WEBSDK550
 mvn install:install-file -Dfile=vim25_55.jar        -DgroupId=com.cloud.com.vmware -DartifactId=vmware-vim25    -Dversion=5.5   -Dpackaging=jar
+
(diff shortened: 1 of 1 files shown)


swill:
This is to fix novelist on master. Correct?

swill:
noredist* phone loves to change things.

swill:
@rhtyd can you review why travis is failing?  It seems to be failing to find something at shapeblue.  We should probably review what is being hosted at shapeblue which is in our critical path and get them moved.  Thx...

rhtyd:
@swill there is a ipmitool pkg that needs to be installed before each run. I'll send a PR for a workaround.
The packages.shapeblue.com repository is down, which we'll chase with BT who is hosting the server on Monday.

rhtyd:
@pdion891 usually, I clone this repo: https://github.com/rhtyd/cloudstack-nonoss and execute the install-non-oss.sh script in that repo: https://github.com/rhtyd/cloudstack-nonoss/blob/master/install-non-oss.sh

rhtyd:
LGTM

pdion891:
@rhtyd  why not update the main "deps/install-non-oss.sh" instead of duplicating it in another repo? Also, I think we are due to update the documentation on build ACS with noredist. It's an oudated documentation :-S

swill:
Fixed...  You can resume testing `master` @pdion891.  Cheers...

PR #1572, opened 2016-05-30
CLOUDSTACK-9395: Add Virtio RNG device to Instances when configured

wido:
By adding a Random Number Generator device to Instances we can prevent
entropy starvation inside guest.

The default source is /dev/random on the host, but this can be configured
to another source when present, for example a hardware RNG.

When enabled it will add the following to the Instance's XML definition:
    
      <rng model='virtio'>
        <backend model='random'>/dev/random</backend>
      </rng>
    
If the Instance has the proper support, which most modern distributions have,
it will have a /dev/hwrng device which it can use for gathering entropy.

### DIFF
--- a/agent/conf/agent.properties
+++ b/agent/conf/agent.properties
@@ -152,3 +152,22 @@ hypervisor.type=kvm
 # kvmclock.disable=false
 # Some newer linux kernels are incapable of reliably migrating vms with kvmclock
 # This is a workaround for the bug, admin can set this to true per-host
+#
+# vm.rng.enable=false
+# This enabled the VirtIO Random Number Generator device for guests.
+#
+# vm.rng.model=random
+# The model of VirtIO Random Number Generator (RNG) to present to the Guest.
+# Currently only 'random' is supported.
+#
+# vm.rng.path=/dev/random
+# Local Random Number Device Generator to use for VirtIO RNG for Guests.
+# This is usually /dev/random, but per platform this might be different
+#
+# vm.rng.rate.bytes=2048
+# The amount of bytes the Guest may request/obtain from the RNG in the period
+# specified below.
+#
+# vm.rng.rate.period=1000
+# The number of milliseconds in which the guest is allowed to obtain the bytes
+# specified above.
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -44,6 +44,7 @@ import java.util.regex.Pattern;
 import javax.ejb.Local;
 import javax.naming.ConfigurationException;
 
+import com.google.common.base.Strings;
 import org.apache.cloudstack.storage.to.PrimaryDataStoreTO;
 import org.apache.cloudstack.storage.to.VolumeObjectTO;
 import org.apache.cloudstack.utils.hypervisor.HypervisorUtils;
@@ -115,6 +116,8 @@ import com.cloud.hypervisor.kvm.resource.LibvirtVMDef.SerialDef;
 import com.cloud.hypervisor.kvm.resource.LibvirtVMDef.TermPolicy;
 import com.cloud.hypervisor.kvm.resource.LibvirtVMDef.VideoDef;
 import com.cloud.hypervisor.kvm.resource.LibvirtVMDef.VirtioSerialDef;
+import com.cloud.hypervisor.kvm.resource.LibvirtVMDef.RngDef;
+import com.cloud.hypervisor.kvm.resource.LibvirtVMDef.RngDef.RngBackendModel;
 import com.cloud.hypervisor.kvm.resource.wrapper.LibvirtRequestWrapper;
 import com.cloud.hypervisor.kvm.resource.wrapper.LibvirtUtilitiesHelper;
 import com.cloud.hypervisor.kvm.storage.KVMPhysicalDisk;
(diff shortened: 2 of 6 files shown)


ustcweizhou:
@wido what is the minimum of qemu/libvirt version this require? maybe better to add some version check ?

wido:
@ustcweizhou: You need at least libvirt 1.0.3. Ubuntu 14.04 contains Libvirt 1.2.2, so with that hypervisor you are fine.

remibergsma:
@wido Thanks for the feature, I like it! Although I'm a bit worried that many VMs could potentially exhaust the hypervisor's `/dev/random`? How do you handle that?

wido:
Yes, valid point @remibergsma , see: https://libvirt.org/formatdomain.html#elementsRng

I'll add support so that you can configure:

``
<rate period="2000" bytes="1234"/>
``

That way you can prevent the guest from exhausting the HV.

ustcweizhou:
@wido , thanks for your reply. As I know, there are still many people using Ubuntu 12.04. Especially for one starting using CloudStack/Ubuntu 2-3 years ago ( before Ubuntu 14.04 release), it is difficult for them to (live) upgrade their production to CloudStack 4.9/Ubuntu 14.04 or 16.04. I think it is better to be compatible with Ubuntu 12.04, by disabling the feature.

wido:
@ustcweizhou Using 'vm.rng.enable=true' you can enable or disable the RNG. This controls if the XML is generated or not.

This is set to false by default, so it will work on Ubuntu 12.04 without any problems as long as you don't enable it.

Ubuntu 12.04 is however still way to old and is using a ancient Java version. So Ubuntu 12.04 will become unsupported anyway in the near future.

wido:
@remibergsma: Could you check again? You can now set the rate with period and bytes to prevent exhaustion.

ustcweizhou:
@wido  thanks. it sound fine to me.

wido:
Ok, thanks @ustcweizhou . Could you take a technical look and see if you LGTM?

remibergsma:
Thanks @wido will have a look

wido:
Did you get a chance to take a look @remibergsma ?

bvbharatk:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 203
 Hypervisor xenserver
 NetworkType Advanced
 Passed=71
 Failed=2
 Skipped=3

**Failed tests:**
* test_vpc_vpn.py

 * ContextSuite context=TestRVPCSite2SiteVpn>:setup Failing since 23 runs

* test_volumes.py

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

wido:
No LGTM on this one yet. Are we good @ustcweizhou and @remibergsma ?

wido:
Can I get a review on this one so it can be merged?

karuturi:
code LGTM
👍  from @ustcweizhou and there are test runs
I am merging this now.

jburwell:
@karuturi am I missing the test results on this PR for the enhancements?  All I see are tests against XenServer which have no bearing on this change.   Why was this PR merged without a full run against KVM and while we are frozen to fix the smoke test suite?

karuturi:
1. the current set of integration tests doesnt test the new functionality added in this PR. Its merely a check to see nothing else is broken.
2. when did we freeze? I skimmed through the mails didnt see anything about master frozen. I do not see any blocker defects for 4.10.0.0. The latest understanding I have is anybody can merge with required code LGTMs and a BVT run.
3. Which smoke tests are broken? Is it you environment issue you are talking about? If not, are these new or did we release 4.9.0 with these broken tests/features?

rhtyd:
@karuturi I think it beats the purpose of smoke tests if we don't run them against the hypervisor(s) the PR/change is most likely related to. Without valid test results it is hard to validate a change/PR, the guideline was to have two LGTMs and at least one test result/run though I agree any committer should be able to merge a PR. This is the email thread you're looking for:
http://markmail.org/message/xp7ckhxhip2rbnr7

For the past two weeks several of us are trying to fix tests/issues and get most of the smoke tests pass on all three major hypervisors. You may follow that effort on PR #1692 and #1703 that uses Trillian (https://github.com/shapeblue/Trillian).

karuturi:
@rhtyd my point is that, even if its run on KVM, this functionality wont be covered with existing tests. I agree to your point. It was just that I had xen tests available and another run on KVM would take huge amount of effort from my side and at the end wont test the new functionality.

Thanks for the email. will go through.

I went through the commits on #1692. It has two bug fixes(CLOUDSTACK-9515, CLOUDSTACK-9533) and some test fixes. I dont see why its a blocker for commits to master. (probably the mail has some context. will go through)

mlsorensen:
I realize I'm late to the party on this, but I'd like to suggest that we change from the agent.properties "vm.rng.enable=true" to simply using the libvirt version detection like we do for other things, and perhaps allow a "vm.rng.disable". It really seems like this is a useful setting and should probably be default wherever possible.

wido:
@mlsorensen We can indeed do that, but the good thing for now is that the RNG is disabled by default.

You also need a very ancient libvirt for it not to support RNG and that will get less and less over time.

PR #1574, opened 2016-05-31
Make sure that the DB drivers are loaded before creating connections

nlivens:
I've digged deeper, and found out that Tomcat is really specific in how it loads the JDBC drivers apparently.
If we would be using the standard JDBC connection pooling of Tomcat (tomcat-jdbc) instead of commons-dbcp, we would have the option to specify a "driverClassName" when creating our connection.
This is not the case for commons-dbcp, which we are using within ACS.

If you check an official example of Tomcat : 
https://tomcat.apache.org/tomcat-7.0-doc/jdbc-pool.html#Plain_Ol'_Java 
or
https://tomcat.apache.org/tomcat-7.0-doc/jdbc-pool.html#As_a_Resource

As you can see in the above examples, both of them specify the driverClassName.
In the underlying implementation of Tomcat, Tomcat will do ```Class.forName(driverClassName)``` which will trigger the auto-registration of the Driver.

Tomcat code :
```java
            if (driver==null) {
                if (log.isDebugEnabled()) {
                    log.debug("Instantiating driver using class: "+poolProperties.getDriverClassName()+" [url="+poolProperties.getUrl()+"]");
                }
                driver = (java.sql.Driver) Class.forName(poolProperties.getDriverClassName(),
                                                         true, PooledConnection.class.getClassLoader()
                                                         ).newInstance();
            }
```

### DIFF
--- a/framework/db/src/com/cloud/utils/db/DriverLoader.java
+++ b/framework/db/src/com/cloud/utils/db/DriverLoader.java
@@ -1,0 +1,69 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// the License.  You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+package com.cloud.utils.db;
+
+import com.cloud.utils.exception.CloudRuntimeException;
+import org.apache.log4j.Logger;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DriverLoader {
+
+    private static final Logger LOGGER = Logger.getLogger(DriverLoader.class.getName());
+    private static final List<String> LOADED_DRIVERS;
+    private static final Map<String, String> DRIVERS;
+
+    static {
+        DRIVERS = new HashMap<String, String>();
+        DRIVERS.put("jdbc:mysql", "com.mysql.jdbc.Driver");
+        DRIVERS.put("jdbc:postgresql", "org.postgresql.Driver");
+        DRIVERS.put("jdbc:h2", "org.h2.Driver");
+
+        LOADED_DRIVERS = new ArrayList<String>();
+    }
+
+
+    public static void loadDriver(String dbDriver) {
+        String driverClass = DRIVERS.get(dbDriver);
+        if (driverClass == null) {
+            LOGGER.error("DB driver type " + dbDriver + " is not supported!");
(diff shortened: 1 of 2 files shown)


nlivens:
@rhtyd, @DaanHoogland, @swill, pinging you guys for an update regarding the Tomcat start up issue. This is related to #1553. I've also updated this in the mailing thread.

framework/db/src/com/cloud/utils/db/DriverLoader.java
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.log4j.Logger;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DriverLoader {
+
+    private static final Logger s_logger = Logger.getLogger(DriverLoader.class.getName());
```
rhtyd: avoid using s_, just use LOGGER, or LOG.

framework/db/src/com/cloud/utils/db/DriverLoader.java
```diff
@@ -0,0 +1,49 @@
...
+    private static List<String> s_loadedDrivers;
+    private static Map<String, String> s_drivers;
```
rhtyd: Remove s_, moving forward we should follow modern naming conventions.

framework/db/src/com/cloud/utils/db/DriverLoader.java
```diff
@@ -0,0 +1,49 @@
...
+
+        if (s_loadedDrivers.contains(dbDriver)) {
+            s_logger.debug("DB driver " + driverClass + " was already loaded.");
+            return;
+        }
+
+        try {
+            Class.forName(driverClass).newInstance();
+            s_loadedDrivers.add(dbDriver);
+            s_logger.debug("Successfully loaded DB driver " + driverClass);
```
rhtyd: add a check to call debug if it's enabled.

rhtyd:
In general, LGTM though it's up to general community acceptance.

nlivens:
@rhtyd, thanks for your quick review, handled your comments.

swill:
@nlivens `master` is frozen for the 4.9 release.  I have not been having issues with this in my environment, so I don't think this is required for 4.9.  Are you not able to use master because of this?

swill:
Please close and reopen or amend and force push to kick off travis again...

nlivens:
@swill, that's what I was afraid of. Are you running ACS as a service, or just running it like a development setup (through jetty)? Since that makes a huge difference. And yes, we are unable to start up ACS 4.9.0 because of this issue.

swill:
I am running it with `mvn -pl :cloud-client-ui jetty:run > jetty.log 2>&1 &`.

You are are starting it with `service start cloudstack-management` or something like that?

swill:
I will do some testing with installing from the packages and running as a service...

nlivens:
Yes, that's what I was thinking, in my development setup (```mvn -pl :cloud-client-ui jetty:run```) it's working fine as well. But on our real test setups (```service cloudstack-management start```), it's failing with the issue described. There are other ways we can fix this, but then we have to rethink the whole DB connection pooling :)

swill:
Ok, thanks for being on top of this.  I will get this tested.

nlivens:
@swill, great, thanks!

swill:
I gave one of my guys RPMs for master without this PR and he was able to reproduce the issue you have described.  I gave him new RPMs for master including your PR and he got up and running without issues.  I will merge this as it fixes master.  Thanks...

nlivens:
@swill, thanks, I appreciate the effort!

PR #1595, opened 2016-06-23
ui: show resize volume button to all users

rhtyd:
screenshot from 2016-06-23 12-29-56

The resize volume is support on all major hypervisors (Xen, VMware, KVM).
The hypervisor key is returned by the list volumes response only for admins
but not for users or domain admin users. This removes the check, as the operation
is supported on all major hypervisors that CloudStack supports.

With this bug fix all users would see resize volume button in the UI.

/cc @swill

### DIFF
--- a/ui/scripts/storage.js
+++ b/ui/scripts/storage.js
@@ -2365,10 +2365,8 @@
             }
         }
 
-        if (jsonObj.hypervisor == "KVM" || jsonObj.hypervisor == "XenServer" || jsonObj.hypervisor == "VMware") {
-            if (jsonObj.state == "Ready" || jsonObj.state == "Allocated") {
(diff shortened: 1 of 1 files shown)


remibergsma:
Thanks @rhtyd nice fix. Tested it and LGTM.

Before:
screen shot 2016-06-23 at 21 45 41 pm

After:
screen shot 2016-06-23 at 21 47 08 pm

rhtyd:
Thanks @remibergsma /cc @swill ?

syed:
LGTM :+1:

wido:
LGTM

rhtyd:
@swill this is a simple UI bugfix, please consider merging for 4.9.0 rc2 as well.

PR #1600, opened 2016-06-30
Support Backup of Snapshots for Managed Storage

syed:
```
This PR adds an ability to Pass a new parameter, locationType,
to the “createSnapshot” API command. Depending on the locationType,
we decide where the snapshot should go in case of managed storage.

There are two possible values for the locationType param

1) `Primary`: The standard operation for managed storage is to
keep the snapshot on the device (primary). For non-managed storage, this will
give an error as this option is only supported for managed storage

2) `Secondary`: Applicable only to managed storage. This will
keep the snapshot on the secondary storage. For non-managed
storage, this will result in an error.

The reason for implementing this feature is to avoid a single
point of failure for primary storage. Right now in case of managed
storage, if the primary storage goes down, there is no easy way
to recover data as all snapshots are also stored on the primary.
This features allows us to mitigate that risk.
```


syed:
This was done in two parts: 

The work to add the API parameter was done by @mike-tutkowski and the work to enable this on the backend was done by me. The rebase destroyed history but I would like to correctly credit the contributors :)

mike-tutkowski:
My first test run is successful - TestVolumes.py without resigning:

test_00_check_template_cache (TestVolumes.TestVolumes) ... === TestName: test_00_check_template_cache | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_01_attach_new_volume_to_stopped_VM | Status : SUCCESS ===
ok
Attach, detach, and attach volume to a running VM ... === TestName: test_02_attach_detach_attach_volume | Status : SUCCESS ===
ok
Attach volume to running VM, then reboot. ... === TestName: test_03_attached_volume_reboot_VM | Status : SUCCESS ===
ok
Detach volume from a running VM, then reboot. ... === TestName: test_04_detach_volume_reboot | Status : SUCCESS ===
ok
Detach volume from a stopped VM, then start. ... === TestName: test_05_detach_vol_stopped_VM_start | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_06_attach_volume_to_stopped_VM | Status : SUCCESS ===
ok
Destroy and expunge VM with attached volume ... === TestName: test_07_destroy_expunge_VM_with_volume | Status : SUCCESS ===
ok
Delete volume that was attached to a VM and is detached now ... === TestName: test_08_delete_volume_was_attached | Status : SUCCESS ===
ok
Attach a data disk to a VM in one account and attach another data disk to a VM in another account ... === TestName: test_09_attach_volumes_multiple_accounts | Status : SUCCESS ===
ok
Attach more than one disk to a VM ... === TestName: test_10_attach_more_than_one_disk_to_VM | Status : SUCCESS ===
ok

Ran 11 tests in 2000.082s

OK

mike-tutkowski:
My second test run is successful - TestVolumes with resigning:

test_00_check_template_cache (TestVolumes.TestVolumes) ... === TestName: test_00_check_template_cache | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_01_attach_new_volume_to_stopped_VM | Status : SUCCESS ===
ok
Attach, detach, and attach volume to a running VM ... === TestName: test_02_attach_detach_attach_volume | Status : SUCCESS ===
ok
Attach volume to running VM, then reboot. ... === TestName: test_03_attached_volume_reboot_VM | Status : SUCCESS ===
ok
Detach volume from a running VM, then reboot. ... === TestName: test_04_detach_volume_reboot | Status : SUCCESS ===
ok
Detach volume from a stopped VM, then start. ... === TestName: test_05_detach_vol_stopped_VM_start | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_06_attach_volume_to_stopped_VM | Status : SUCCESS ===
ok
Destroy and expunge VM with attached volume ... === TestName: test_07_destroy_expunge_VM_with_volume | Status : SUCCESS ===
ok
Delete volume that was attached to a VM and is detached now ... === TestName: test_08_delete_volume_was_attached | Status : SUCCESS ===
ok
Attach a data disk to a VM in one account and attach another data disk to a VM in another account ... === TestName: test_09_attach_volumes_multiple_accounts | Status : SUCCESS ===
ok
Attach more than one disk to a VM ... === TestName: test_10_attach_more_than_one_disk_to_VM | Status : SUCCESS ===
ok

Ran 11 tests in 2313.359s

OK

mike-tutkowski:
My third test run is successful - TestVMSnapshots.py:

test_01_take_VM_snapshot (TestVMSnapshots.TestVMSnapshots) ... === TestName: test_01_take_VM_snapshot | Status : SUCCESS ===
ok
test_02_take_VM_snapshot_with_data_disk (TestVMSnapshots.TestVMSnapshots) ... === TestName: test_02_take_VM_snapshot_with_data_disk | Status : SUCCESS ===
ok

Ran 2 tests in 765.252s

OK

mike-tutkowski:
My fourth test run is successful - TestSnapshots.py:

test_01_create_volume_snapshot_using_sf_snapshot (TestSnapshots.TestSnapshots) ... === TestName: test_01_create_volume_snapshot_using_sf_snapshot | Status : SUCCESS ===
ok
test_02_create_volume_snapshot_using_sf_volume (TestSnapshots.TestSnapshots) ... === TestName: test_02_create_volume_snapshot_using_sf_volume | Status : SUCCESS ===
ok
test_03_create_volume_snapshot_using_sf_volume_and_sf_snapshot (TestSnapshots.TestSnapshots) ... === TestName: test_03_create_volume_snapshot_using_sf_volume_and_sf_snapshot | Status : SUCCESS ===
ok

Ran 3 tests in 7114.793s

OK

mike-tutkowski:
My fifth and final test run is successful - TestAddRemoveHosts.py:

test_add_remove_host_with_solidfire_plugin_1 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_1 | Status : SUCCESS ===
ok
test_add_remove_host_with_solidfire_plugin_2 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_2 | Status : SUCCESS ===
ok
test_add_remove_host_with_solidfire_plugin_3 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_3 | Status : SUCCESS ===
ok
test_add_remove_host_with_solidfire_plugin_4 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_4 | Status : SUCCESS ===
ok

Ran 4 tests in 3665.055s

OK

mike-tutkowski:
The code LGTM.

In addition, my automated regression tests have all come back successful (documented above).

I plan to write new automated tests for this PR.

mike-tutkowski:
In manual testing, I've noticed that the SR that contains the VDI that we copy from primary to secondary storage is not removed from XenServer.

I see that the SolidFire volume that contains the SR is no longer in a volume access group (good) and it is deleted (good), but the SR is still attached to by XenServer.

syed:
Good catch @mike-tutkowski. I will fix it.

mike-tutkowski:
Can you return the new locationtype parameter in the listSnapshots API call?

syed:
Sure @mike-tutkowski

syed:
A bit of background on how this feature works. When locationType is specified as `Archve`, a snaphsot is taken at the storage, but the copy to seconday storage is done via the hypervisor as it is able to correctly attach the LUN and extract the underlying virtual disk. This was done as an alternative to taking the snapshot on the hypervisor and upload it to secondary storage directly becuse, we've seen cases where hypervisor snapshots cause a significant delay in read/write performance and snapshot on storage device allows us to take snaspots without sacrificing performance.

syed:
@mike-tutkowski , the list snapshot already returns the `loacationType`. I think you've already added that.

mike-tutkowski:
My mistake. I forgot that I added that a while ago. :)

On Jul 4, 2016, at 12:40 PM, Syed Mushtaq Ahmed <notifications@github.com<mailto:notifications@github.com>> wrote:


@mike-tutkowski<https://github.com/mike-tutkowski> , the list snapshot already returns the loacationType. I think you've already added that.

-

syed:
@mike-tutkowski I've fixed the problem where the volume was not being detached after the copy.

mike-tutkowski:
A comment:


I have observed that the SR that was previously not being detached from XenServer is now being detached with your newer version of the code.


One thing I noticed in the DB:


In the snapshot_store_ref table, we still have a row for the snapshot that indicates there is a version of the snapshot on primary storage. We also - correctly - have a row indicating that there is a version of the snapshot on secondary storage.


I think we should remove the row from the snapshot_store_ref table that links the snapshot to a version on primary storage.


________________________________

syed:
@mike-tutkowski I think the `Destroyed` state should suffice.

api/src/org/apache/cloudstack/api/command/user/snapshot/CreateSnapshotCmd.java
```diff
@@ -195,24 +196,37 @@ public void create() throws ResourceAllocationException {
...
 
+    private Snapshot.LocationType getLocationType() {
+        if (Snapshot.LocationType.values() == null || Snapshot.LocationType.values().length == 0) {
+            return null;
+        }
+
+        Snapshot.LocationType locationTypeToReturn = Snapshot.LocationType.values()[0];
+
+        if (locationType != null && locationType >= 1 && locationType <= Snapshot.LocationType.values().length) {
+            locationTypeToReturn = Snapshot.LocationType.values()[locationType-1];
```
karuturi: I would suggest keeping the enum values all upper case, make locationType string in cmd and use LocationType.valueOf(cmd.locationType.toUpperCase())
https://books.google.co.in/books?id=ka2VUBqHiWkC&lpg=PA158&dq=Never%20derive%20a%20value%20associated%20with%20an%20enum%20from%20its%20ordinal&pg=PA158#v=onepage&q&f=false

api/src/org/apache/cloudstack/api/response/SnapshotResponse.java
```diff
@@ -82,6 +82,10 @@
     @Param(description = "valid types are hourly, daily, weekly, monthy, template, and none.")
     private String intervalType;
 
+    @SerializedName(ApiConstants.LOCATION_TYPE)
+    @Param(description = "valid location types are primary and archive.")
+    private String locationType;
```
karuturi: LocationType is a String in response. It makes sense for it be String in request as well.

karuturi:
`locationType` Standard and Archive refer to either secondary or primary based on the storage type and is confusing(atleast for me). Can we make this simple and use the locationType as primary or secondary?

karuturi:
BTw, does this work only with xenserver? I dont see any vmware related changes

syed:
Thanks @karuturi for your comments, I will get to them as soon as I have some free cycles. Right now this is only supported for XenServer as this requires the backend to be able to do native snapshots. If there is need, we can add VMWare/KVM support as well.

api/src/org/apache/cloudstack/api/response/SnapshotResponse.java
jburwell: +1 @karuturi -- types should be symmetric between request and response.

api/test/org/apache/cloudstack/api/command/test/CreateSnapshotCmdTest.java
```diff
@@ -0,0 +1,130 @@
...
+        responseGenerator = Mockito.mock(ResponseGenerator.class);
+        SnapshotResponse snapshotResponse = Mockito.mock(SnapshotResponse.class);
+        Mockito.when(responseGenerator.createSnapshotResponse(snapshot)).thenReturn(snapshotResponse);
+        Mockito.doNothing().when(snapshotResponse).setAccountName(anyString());
+
+        createSnapshotCmd._accountService = accountService;
+        createSnapshotCmd._responseGenerator = responseGenerator;
+        createSnapshotCmd._volumeService = volumeApiService;
+
+        createSnapshotCmd.execute();
```
jburwell: Why aren't there any assertions in this test method?  Seems like it should at least validate the response returned by ``createSnapshotCmd.execute()``.

api/test/org/apache/cloudstack/api/command/test/CreateSnapshotCmdTest.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    @Test
+    @Ignore
```
jburwell: Why are we adding a new unit test that is being ignored?

core/src/org/apache/cloudstack/storage/to/PrimaryDataStoreTO.java
```diff
@@ -51,6 +50,7 @@
     private final String url;
     private Map<String, String> details;
     private static final String pathSeparator = "/";
+    private boolean isManaged;
```
jburwell: Since no mutator is exposed and the value gets set in the constructor, considering making it ``final``.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -328,15 +412,64 @@ private void handleCreateTemplateFromSnapshot(SnapshotInfo snapshotInfo, Templat
...
+                cmd.setManaged(true);
+                cmd.setStorageHost(storagePool.getHostAddress());
+                cmd.setStoragePort(storagePool.getPort());
+                cmd.set_iScsiName(getProperty(snapshotInfo.getId(), DiskTO.IQN));
+
+                try {
+                    DettachAnswer dettachAnswer = (DettachAnswer) _agentMgr.send(hostVO.getId(), cmd);
+
+                    if (!dettachAnswer.getResult()) {
+                        throw new CloudRuntimeException("Error detaching volume:" + dettachAnswer.getDetails());
```
jburwell: Throwing an exception here will end processing of this ``finally`` block, and skipping the ``_volumeService.revokeAccess`` call at the end of the method.  What would be the impact of skipping that call in the event an exception is thrown here?

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -380,6 +513,122 @@ private void handleCreateTemplateFromSnapshot(SnapshotInfo snapshotInfo, Templat
...
+
+        HostVO hostVO = null;
+        try {
+
+            //create a volume on the storage
+            AsyncCallFuture<VolumeApiResult> future = _volumeService.createVolumeAsync(volumeInfo, volumeInfo.getDataStore());
+            VolumeApiResult result = future.get();
+
+            if (result.isFailed()) {
+                LOGGER.warn("Failed to create a volume: " + result.getResult());
```
jburwell: Why is this message being logged to ``WARN`` and not ``ERROR``?

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -380,6 +513,122 @@ private void handleCreateTemplateFromSnapshot(SnapshotInfo snapshotInfo, Templat
...
+
+        CopyCmdAnswer copyCmdAnswer = null;
+        String errMsg = null;
+
+        HostVO hostVO = null;
+        try {
+
+            //create a volume on the storage
+            AsyncCallFuture<VolumeApiResult> future = _volumeService.createVolumeAsync(volumeInfo, volumeInfo.getDataStore());
+            VolumeApiResult result = future.get();
```
jburwell: Should this call be bounded by a timeout?

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -380,6 +513,122 @@ private void handleCreateTemplateFromSnapshot(SnapshotInfo snapshotInfo, Templat
...
+
+                if (!dettachAnswer.getResult()) {
+                    throw new CloudRuntimeException("Error detaching volume:" + dettachAnswer.getDetails());
+                }
+
+            } catch (Exception e) {
+                LOGGER.warn("Error detaching volume " + volumeInfo.getId() + " Error: " + e.getMessage());
+            }
+
+            _volumeService.revokeAccess(volumeInfo, hostVO, destDataStore);
```
jburwell: Lines 571-593 appear to duplicate Line 463-475 in ``StorageSystemDataMotionStrategy``.  Please consider consolidation of the commonalities.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -380,6 +513,122 @@ private void handleCreateTemplateFromSnapshot(SnapshotInfo snapshotInfo, Templat
...
+                Answer deleteAnswer = ep.sendMessage(deleteCommand);
+                if (deleteAnswer != null && deleteAnswer.getResult()) {
+                    LOGGER.warn("Unable to cleanup staging NFS " + deleteAnswer.getDetails() +
+                    "Object: " + destOnStore.getType() + " ID: " + destOnStore.getId());
+                    destOnStore.processEvent(Event.OperationFailed);
+                }
+            }
+
+            destOnStore.processEvent(Event.OperationSuccessed);
+        } catch (Exception e) {
```
jburwell: Why are all checked and unchecked exceptions being caught?

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
```diff
@@ -92,7 +90,34 @@
...
+
+        HostVO host = getHost(snapshotInfo.getVolumeId());
+        boolean canStorageSystemCreateVolumeFromSnapshot = canStorageSystemCreateVolumeFromSnapshot(snapshotInfo.getBaseVolume().getPoolId());
+        boolean computeClusterSupportsResign = clusterDao.getSupportsResigning(host.getClusterId());
+
+        if (!canStorageSystemCreateVolumeFromSnapshot || !computeClusterSupportsResign) {
+            String mesg = "Cannot archive snapshot: Either canStorageSystemCreateVolumeFromSnapshot or " +
+                    "computeClusterSupportsResign were false.  ";
+
+            s_logger.warn(mesg);
```
jburwell: Why is the message being logged to ``WARN`` and not ``ERROR``?

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
```diff
@@ -92,7 +90,34 @@
 
     @Override
     public SnapshotInfo backupSnapshot(SnapshotInfo snapshotInfo) {
-        return snapshotInfo;
+
```
jburwell: Please add a ``Preconditions.checkArgument(snapshotInfo != null, "<explaination of the precondition>")`` to protect against potential NPEs.

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
```diff
@@ -92,7 +90,34 @@
...
+        if (!snapshotInfo.getLocationType().equals(Snapshot.LocationType.SECONDARY)) {
```
jburwell: Since Enum values are guaranteed to be singletons, the `==` is that same as `equals`.  By changing this check to be ``snapshotInfo.getLocationType() != Snapshot.LocationType.SECONDARY``, a potential NPE is avoided and the code is more straightforward.

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
```diff
@@ -92,7 +90,34 @@
...
+            markAsBackedUp((SnapshotObject)snapshotInfo);
+            return snapshotInfo;
+        }
+
+        // At this point the snapshot is either taken as a native
+        // snapshot on the storage or exists as a volume on the storage (clone).
+        // If archive flag is passed, we will copy this snapshot to secondary
+        // storage and delete it from the primary storage.
+
+        Preconditions.checkState(Snapshot.LocationType.SECONDARY.equals(snapshotInfo.getLocationType()));
```
jburwell: Why is the check necessary given the check on Line 94?

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
```diff
@@ -212,20 +251,47 @@ public SnapshotInfo takeSnapshot(SnapshotInfo snapshotInfo) {
...
+            updateLocationTypeInDb(backedUpSnapshot);
+
         }
         finally {
             if (result != null && result.isSuccess()) {
                 volumeInfo.stateTransit(Volume.Event.OperationSucceeded);
-            }
-            else {
+
+                Preconditions.checkState(snapshotOnPrimary != null);
```
jburwell: Should an exception be thrown in this ``finally`` block or just skip processing?

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
syed: There is a `catch` statement which catches this excepiton. The `revokeAccess()` call will always happen.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
syed: The storage drivers have a timeout logic in them already.

jburwell: @syed it is not safe to assume that current/future underlying storage driver will properly implement timeouts.  The consequences of a buggy implementation are to starve the Management Server of resources causing it to become unresponsive or crash.  Please bound this call in a timeout -- ideally specified as a global setting.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
jburwell: I lost track of the catch block below.  Why are we throwing an exception to catch a few lines later?  It appears to be flow of control using exceptions is a code anti-pattern.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
syed: We don't want to error out when we fail to clean up cache storage. I have added a statemachine transition using the event `OperationFailed` in the catch block

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
syed: Done

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
syed: Removed the check

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
syed: Removed the precondition, added an if check which better handles the situation

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
syed: Added precondition

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/StorageSystemSnapshotStrategy.java
syed: Changed to ERROR

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
syed: moved to a common function

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
syed: Added timout

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
syed: changed to ERROR

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
syed: Logging, instead of throwing exception

core/src/org/apache/cloudstack/storage/to/PrimaryDataStoreTO.java
syed: Made the variable `final`

api/test/org/apache/cloudstack/api/command/test/CreateSnapshotCmdTest.java
syed: Added assertion

syed:
@karuturi @jburwell  Can we merge this in? Anthing left from my side?

jburwell:
@syed I want to hold this PR until we have a chance to review #1642.  On the surface, there appears to be some overlap, and I like to understand it before proceeding.

mike-tutkowski:
@jburwell @syed There is no real overlap. Syed and I are working on items from a shared to-do list.

The PR here is something he and I built so that volume snapshots can be taken using a SAN (like SolidFire) and then moved to secondary storage later (the benefits are outlined in his description section above).

The other PR is related to running system VMs using managed storage. At a high level, that means having a virtual disk of a system VM on its own backend volume (for QoS reasons).

mike-tutkowski:
@syed @karuturi @jburwell Just wondering how we're doing on this one. Is there anything else that Syed needs to do here? Thanks!

jburwell:
@mike-tutkowski @syed needs to resolve the conflicts before we can proceed with package and functional testing.

syed:
@jburwell I've rebased against the master. The merge should be good now. I've addressed your comments as well. @mike-tutkowski has ran his tests on this functionality. 

https://github.com/apache/cloudstack/pull/1600#issuecomment-229981990
https://github.com/apache/cloudstack/pull/1600#issuecomment-229965967

What would be the next steps here 

cc: @karuturi

jburwell:
@syed could you please squash your commits?  Once that is done, the following items will be need to be completed:

* [ ] Re-execute functional test since the code has changed since they were last run (@mike-tutowski)
* [ ] Execute regression tests on KVM, VMware, and Xen with unmanaged storage (@jburwell using blueorangutan)

Are there any specific component tests that should be run in addition to the smoke tests?

syed:
@jburwell Squashed the commits. You can run the smoke tests on this. This change is mostly on the managed storage side so running any volume related tests would be helpful apart from the tests @mike-tutkowski has added.

mike-tutkowski:
@syed @jburwell I should be able to run the tests toward the end of the week or during the weekend. At the moment, my resources in the lab are devoted to other CloudStack dev activities.

jburwell:
@blueorangutan package

jburwell:
@syed #1692 is testing now.  It addresses ping test related failures in the smoke test suite.  I am hoping the tests pass, and we can merge it first thing tomorrow.  Once it has, we can rebase this branch and kick of the smoke test run.

jburwell:
@syed it appears that conflicts were introduced when #1689 was merged.  Could you please resolve?

mike-tutkowski:
@syed You can actually just remove the integration tests from this PR as they were added (along with updates to use the new SolidFire SDK for Python) in #1689.

syed:
@jburwell Rebased with master and resolved conflicts

mike-tutkowski:
@syed It actually looks like the rebase lead to the integration tests being put back in their pre-official SolidFire SDK for Python state.

I recommend you just remove all integration tests from this PR as all managed-storage integration tests are now current as of the merge of #1689.

syed:
@mike-tutkowski I've added your updated tests and rebased.

mike-tutkowski:
Sounds good, @syed Since what you added is the same as what's currently in master, it looks like your list of changed files for this PR is down to 35 (only one of which is a Python file, base, py). That looks good.

jburwell:
@syed @mike-tutkowski should the tests be re-executed to make sure everything is super green?

mike-tutkowski:
@jburwell Yes, the managed-storage tests should be re-executed. I can do this over the weekend as my CloudStack systems in the lab here are being fully utilized for the time being with another project I'm working on at the moment. I should be able to provide the results by Monday.

jburwell:
@mike-tutkowski sounds good.  I am putting together a test results email which should be out shortly. It outlines the issues we have found thus far.  As such, I expect the testing period for the releases to be extended while we address those issues and then rebase and retest a number of open PRs.

mike-tutkowski:
I have run all of the relevant managed-storage tests and all passed (details below).

I did encounter one issue:

When the fourth test of TestSnapshots ran, it failed because I did not have the base.py file from this PR in my Marvin install. I went ahead and rebuilt Marvin and re-installed it.

The same test then encountered another issue:

As it turns out, the new (optional) parameter, locationtype, was still annotated as a CommandType.SHORT (the original design), even though the variable it annotates was changed to be a String during the PR review. I went ahead and made changes to CreateSnapshot.java (now the variable is annotated as a CommandType.STRING) and to the Marvin test, TestSnapshot.py.

The only code that would have noticed this parameter issue is this one test (none of the other tests in our system pass in this new (optional) parameter).

Here are the details of the test results:

TestSnapshots:

Note the error on the fourth test (and then that it passes after the noted code changes):

test_01_create_volume_snapshot_using_sf_snapshot (TestSnapshots.TestSnapshots) ... === TestName: test_01_create_volume_snapshot_using_sf_snapshot | Status : SUCCESS ===
ok
test_02_create_volume_snapshot_using_sf_volume (TestSnapshots.TestSnapshots) ... === TestName: test_02_create_volume_snapshot_using_sf_volume | Status : SUCCESS ===
ok
test_03_create_volume_snapshot_using_sf_volume_and_sf_snapshot (TestSnapshots.TestSnapshots) ... === TestName: test_03_create_volume_snapshot_using_sf_volume_and_sf_snapshot | Status : SUCCESS ===
ok
test_04_create_volume_snapshot_using_sf_snapshot_and_archiving (TestSnapshots.TestSnapshots) ... === TestName: test_04_create_volume_snapshot_using_sf_snapshot_and_archiving | Status : EXCEPTION ===
ERROR

After the noted code changes:

test_04_create_volume_snapshot_using_sf_snapshot_and_archiving (TestSnapshots.TestSnapshots) ... === TestName: test_04_create_volume_snapshot_using_sf_snapshot_and_archiving | Status : SUCCESS ===
ok

Ran 1 test in 3107.229s

OK

TestAddRemoveHosts:

test_add_remove_host_with_solidfire_plugin_1 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_1 | Status : SUCCESS ===
ok
test_add_remove_host_with_solidfire_plugin_2 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_2 | Status : SUCCESS ===
ok
test_add_remove_host_with_solidfire_plugin_3 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_3 | Status : SUCCESS ===
ok
test_add_remove_host_with_solidfire_plugin_4 (TestAddRemoveHosts.TestAddRemoveHosts) ... === TestName: test_add_remove_host_with_solidfire_plugin_4 | Status : SUCCESS ===
ok

Ran 4 tests in 4357.199s

OK

TestVMSnapshots:

test_01_take_VM_snapshot (TestVMSnapshots.TestVMSnapshots) ... === TestName: test_01_take_VM_snapshot | Status : SUCCESS ===
ok
test_02_take_VM_snapshot_with_data_disk (TestVMSnapshots.TestVMSnapshots) ... === TestName: test_02_take_VM_snapshot_with_data_disk | Status : SUCCESS ===
ok

Ran 2 tests in 750.531s

OK

TestVolumes (no XenServer SR resigning):

test_00_check_template_cache (TestVolumes.TestVolumes) ... === TestName: test_00_check_template_cache | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_01_attach_new_volume_to_stopped_VM | Status : SUCCESS ===
ok
Attach, detach, and attach volume to a running VM ... === TestName: test_02_attach_detach_attach_volume | Status : SUCCESS ===
ok
Attach volume to running VM, then reboot. ... === TestName: test_03_attached_volume_reboot_VM | Status : SUCCESS ===
ok
Detach volume from a running VM, then reboot. ... === TestName: test_04_detach_volume_reboot | Status : SUCCESS ===
ok
Detach volume from a stopped VM, then start. ... === TestName: test_05_detach_vol_stopped_VM_start | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_06_attach_volume_to_stopped_VM | Status : SUCCESS ===
ok
Destroy and expunge VM with attached volume ... === TestName: test_07_destroy_expunge_VM_with_volume | Status : SUCCESS ===
ok
Delete volume that was attached to a VM and is detached now ... === TestName: test_08_delete_volume_was_attached | Status : SUCCESS ===
ok
Attach a data disk to a VM in one account and attach another data disk to a VM in another account ... === TestName: test_09_attach_volumes_multiple_accounts | Status : SUCCESS ===
ok
Attach more than one disk to a VM ... === TestName: test_10_attach_more_than_one_disk_to_VM | Status : SUCCESS ===
ok

Ran 11 tests in 1955.862s

OK

TestVolumes (XenServer SR resigning):

test_00_check_template_cache (TestVolumes.TestVolumes) ... === TestName: test_00_check_template_cache | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_01_attach_new_volume_to_stopped_VM | Status : SUCCESS ===
ok
Attach, detach, and attach volume to a running VM ... === TestName: test_02_attach_detach_attach_volume | Status : SUCCESS ===
ok
Attach volume to running VM, then reboot. ... === TestName: test_03_attached_volume_reboot_VM | Status : SUCCESS ===
ok
Detach volume from a running VM, then reboot. ... === TestName: test_04_detach_volume_reboot | Status : SUCCESS ===
ok
Detach volume from a stopped VM, then start. ... === TestName: test_05_detach_vol_stopped_VM_start | Status : SUCCESS ===
ok
Attach a volume to a stopped virtual machine, then start VM ... === TestName: test_06_attach_volume_to_stopped_VM | Status : SUCCESS ===
ok
Destroy and expunge VM with attached volume ... === TestName: test_07_destroy_expunge_VM_with_volume | Status : SUCCESS ===
ok
Delete volume that was attached to a VM and is detached now ... === TestName: test_08_delete_volume_was_attached | Status : SUCCESS ===
ok
Attach a data disk to a VM in one account and attach another data disk to a VM in another account ... === TestName: test_09_attach_volumes_multiple_accounts | Status : SUCCESS ===
ok
Attach more than one disk to a VM ... === TestName: test_10_attach_more_than_one_disk_to_VM | Status : SUCCESS ===
ok

Ran 11 tests in 2362.066s

OK

TestVMMigrationWithStorage:

test_01_storage_migrate_root_and_data_disks (TestVMMigrationWithStorage.TestVMMigrationWithStorage) ... === TestName: test_01_storage_migrate_root_and_data_disks | Status : SUCCESS ===
ok
test_02_storage_migrate_root_and_data_disks (TestVMMigrationWithStorage.TestVMMigrationWithStorage) ... === TestName: test_02_storage_migrate_root_and_data_disks | Status : SUCCESS ===
ok
test_03_storage_migrate_root_and_data_disks_fail (TestVMMigrationWithStorage.TestVMMigrationWithStorage) ... === TestName: test_03_storage_migrate_root_and_data_disks_fail | Status : SUCCESS ===
ok
test_04_storage_migrate_root_disk_fails (TestVMMigrationWithStorage.TestVMMigrationWithStorage) ... === TestName: test_04_storage_migrate_root_disk_fails | Status : SUCCESS ===
ok
test_05_storage_migrate_data_disk_fails (TestVMMigrationWithStorage.TestVMMigrationWithStorage) ... === TestName: test_05_storage_migrate_data_disk_fails | Status : SUCCESS ===
ok

Ran 5 tests in 3589.890s

OK

mike-tutkowski:
It looks like we're having an issue with Travis?

syed:
@mike-tutkowski @jburwell I've re-opened the PR to trigger jenkins again.

syed:
@mike-tutkowski @jburwell looks like we are green now.

mike-tutkowski:
@syed Thanks! I had a similar issue with #1642 and fixed it the same way.

mike-tutkowski:
Just following up on this one. Is there any pending work item that remains or are we able to merge? Thanks!

jburwell:
@mike-tutkowski we are holding merges until we get all smoke tests passing on #1692 (and the subsequent PRs for the 4.9 and master branches).  When those fixes are merged, we have you rebase, run the full smoke test suite, and and merge to master.  This PR is the list to make 4.10.0.0.

mike-tutkowski:
Thanks for the clarification, @jburwell. I had the same question for #1642, but your answer here works for both PRs.

mike-tutkowski:
@karuturi Anything else @syed needs to do for this one? Thanks!

mike-tutkowski:
@syed Just an FYI that this branch had two conflicts with master (after recent changes to master). I went ahead and fixed the conflicts and re-pushed the commit.

jburwell:
@mike-tutkowski I will kick off regression tests now.

jburwell:
@blueorangutan package

syed:
Thanks @mike-tutkowski

jburwell:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

mike-tutkowski:
@rhtyd I don't think either of those two errors are related to this PR (probably environment issues). I think we are good to go here. What do you think?

rhtyd:
@mike-tutkowski yes, those failing tests are known to cause intermittent failures. LGTM on tests. If all the reported issues have been fixed, this can be merged. /cc @jburwell

karuturi:
Looks like all the issues are resolved and the tests are good. I am merging this now.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -538,7 +562,7 @@ public Answer backupSnapshot(final CopyCommand cmd) {
                     final String[] tmp = result.split("#");
                     snapshotBackupUuid = tmp[0];
                     physicalSize = Long.parseLong(tmp[1]);
-                    finalPath = folder + File.separator + snapshotBackupUuid;
+                    finalPath = folder + File.separator + snapshotBackupUuid + ".vhd";
```
abhinandanprateek: This change will result in a ".vhd" extension being written to snapshot_store_ref as this is being returned in the CopyCmdAnswer. Although I prefer this, the implication of this change is that there will be inconsistency in snapshot names in a setup upgraded to this version. I am wondering and please confirm that this is not going to cause the UI to not show the older snapshots, only thing will be that the older snapshots' install_path will be without extension. For consistency can be modify the upgrade script to append ".vhd" to all the old Xen based snapshots ?

cc @rhtyd @jburwell

jburwell: I am more inclined to reverse this change rather than incur the complexity and testing of an upgrade script.  @syed what is the value of adding the extension?  It is correct, and the way it should have been originally implemented.  However, as @abhinandanprateek points out, this could snapshots created with previous versions of ACS to be inaccessible.

rhtyd: While I would want to add the extension, but I think we should revert this specific change, as otherwise it would add upgrade complexities. Given the short time for next RCs, such a change can be properly addressed in future with tooling around upgrading/renaming the files and paths.

syed: @abhinandanprateek @jburwell @rhtyd So I've looked at why I added that change. This was required for object store as secondary storage to work.

Let's say you are creating a template from a snapshot in a standard scenario where you have NFS as secondary storage. Now the request for this will go to the SSVM. Without the extension, SSVM will try to guess the extension based on the hypervisor and invoke a different function for XenServer which is not ideal but works  see: `NfsSecondaryStorageResource.java:copySnapshotToTemplateFromNfsToNfsXenserver`  However, if you want to backup to an object storage, it doesn't have this logic. It basically takes the install path as the source of truth see: `NfsSecondaryStorageResource.java:copyFromNfsToImage` Same thing when we create a volume from snapshot but this time the request goes to the hypervisor and it will add the `.vhd` extension if it doesn't find one see `XenServer625StorageProcessor:createVolumeFromSnapshot`  where it invokes the `getSnapshotUuid` to extract the UUID from the installpath. 

So I've basically fixed the case where we use Object storage and want to backup the snapshot. Now, in the normal case where we have NFS as secondary storage, are the snapshot operations failing? My impression was that they were working correctly.

syed: cc @mike-tutkowski

abhinandanprateek: @syed This is the way it should have been, but now that in a upgraded setup there will be mixed names with ".vhd" and without, do you think this will not cause issues ? If it does we need to come up with a plan to deal with them like a upgrade step ? I think the cleanup threads may not be able to handle the old naming convention for one.  @rhtyd

PR #1604, opened 2016-07-05
improvement: systemvm slow operations

romain-dartigues:
On our setup we had issues with vrouters when a large number of VM were created at once.

The vrouters try to arping the gateway (to refresh the ARP cache I suppose), but in some case it take too much time and the vrouters got killed for being unresponsive.

The following patches attempt to solve this by:
- a96ab18f, 8908371b defensive fix, maybe a bug?
- fce10c55 do not block on arping command

This patch has been running for a month on our test environment and seems to solve the initial problem.

An operation (starting 220 VM on XenServer, shared storage) which took 46 mn, now take 4 mn (without errors).

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
@@ -593,9 +593,11 @@ class CsIP:
         return ip in self.address.values()
 
     def arpPing(self):
+        if not self.address['gateway'] or self.address['gateway'] == 'None':
+            return
         cmd = "arping -c 1 -I %s -A -U -s %s %s" % (
             self.dev, self.address['public_ip'], self.address['gateway'])
-        CsHelper.execute(cmd)
+        CsHelper.execute_background(cmd)
 
     # Delete any ips that are configured but not in the bag
     def compare(self, bag):
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsHelper.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsHelper.py
@@ -24,8 +24,16 @@ import logging
 import os.path
 import re
 import shutil
+
+try:
+    from subprocess import DEVNULL # py3k
+except ImportError:
+    import os
+    DEVNULL = open(os.devnull, 'wb')
+
 from netaddr import *
 from pprint import pprint
+
 
 PUBLIC_INTERFACES = {"router" : "eth2", "vpcrouter" : "eth1"}
 
@@ -180,11 +188,35 @@ def get_hostname():
 
 
 def execute(command):
-    """ Execute command """
+    """Execute a command in a new process and return stdout
+
+    The command is executed in a shell (which allow use of pipes and such),
+    wait for the command to terminate, and return stdout.
+
+    :param command: command with arguments to be executed,
+                    see :py:class:`subprocess.Popen`
+    :type command: str or list or tuple
+    :return: command stdout
+    :rtype: list
+    """
     logging.debug("Executing: %s" % command)
-    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-    result = p.communicate()[0]
-    return result.splitlines()
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=DEVNULL, shell=True)
+    return p.communicate()[0].splitlines()
+
+
+def execute_background(command):
+    """Execute a command in a new process
+
+    The command is executed in a shell (which allow use of pipes and such).
+    The function does not wait command completion to return.
+
+    :param command: command with arguments to be executed,
+                    see :py:class:`subprocess.Popen`
+    :type command: str or list or tuple
+    :return: None
+    """
+    logging.debug("Executing: %s" % command)
+    subprocess.Popen(command, stdout=DEVNULL, stderr=DEVNULL, shell=True)
 
 
 def save_iptables(command, iptables_file):
--- a/systemvm/patches/debian/config/opt/cloud/bin/merge.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/merge.py
@@ -230,12 +230,9 @@ class updateDataBag:
             dp['add'] = True
             dp['one_to_one_nat'] = False
             if nw_type == "public":
-                dp['gateway'] = self.qFile.data['cmd_line']['gateway']
+                dp['gateway'] = self.qFile.data['cmd_line'].get('gateway', 'None')
             else:
-                if('localgw' in self.qFile.data['cmd_line']):
-                    dp['gateway'] = self.qFile.data['cmd_line']['localgw']
-                else:
-                    dp['gateway'] = 'None'
+                dp['gateway'] = self.qFile.data['cmd_line'].get('localgw', 'None')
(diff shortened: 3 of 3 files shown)


systemvm/patches/debian/config/opt/cloud/bin/merge.py
```diff
@@ -228,7 +228,7 @@ def processCLItem(self, num, nw_type):
             if('localgw' in self.qFile.data['cmd_line']):
                 dp['gateway'] = self.qFile.data['cmd_line']['localgw']
             else:
-                dp['gateway'] = 'None'
+                dp['gateway'] = self.qFile.data['cmd_line'].get('gateway', 'None')
```
syed: there could be a possibility of having an ipv6 gateway, in that case, we would be looking at `ip6gateway`.

syed:
@romain-dartigues  do we know why the `arping` is failing? Ideally, if the vif is configured correctly, the arping should go through properly. If the `arping` fails, do we still know that the connectivity is correct/establised?

romain-dartigues:
@syed

> there could be a possibility of having an ipv6 gateway

Right, but IPv6 does not use ARP but [NDP](https://en.wikipedia.org/wiki/Neighbor_Discovery_Protocol) ([RFC 4861](https://tools.ietf.org/html/rfc4861)).

> do we know why the arping is failing?

Apparently because the file "/etc/cloudstack/ips.json" is generated with a gateway at "None" even when the gateway is present in the input "cmd_line.json" (but my memory could fail me at this time of the day).

One month ago we found the "problem" and wanted a fix; trying to understand *why* it did happen made me start #1575 (now #1603).

syed:
Thanks @romain-dartigues based on your comments the code LGTM

syed:
we do have to remember that arping is ipv4 only so we need to fix this for ipv6 too

rhtyd:
@blueorangutan package

systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
```diff
@@ -584,9 +584,11 @@ def hasIP(self, ip):
         return ip in self.address.values()
 
     def arpPing(self):
+        if not self.address['gateway'] or self.address['gateway'] == 'None':
+            return
         cmd = "arping -c 1 -I %s -A -U -s %s %s" % (
             self.dev, self.address['public_ip'], self.address['gateway'])
-        CsHelper.execute(cmd)
+        CsHelper.execute(cmd, wait=False)
```
jburwell: Flag arguments are an anti-pattern.  Please consider introducing and ``executeAndWait`` function instead.

systemvm/patches/debian/config/opt/cloud/bin/cs/CsHelper.py
```diff
@@ -179,12 +179,12 @@ def get_hostname():
...
 
-def execute(command):
+def execute(command, wait=True):
     """ Execute command """
     logging.debug("Executing: %s" % command)
     p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-    result = p.communicate()[0]
-    return result.splitlines()
+    if wait:
+        return p.communicate()[0].splitlines()
```
jburwell: What happens if the command does not return?  It seems like there should be a wait period to timeout operations that run too long.

rhtyd:
@romain-dartigues can you fix the merge conflict, squash the changes, open a JIRA issue referencing that in the commit message.
Pl. rebase against latest 4.9, also edit the PR's base branch against 4.9?

romain-dartigues:
@rhtyd < JIRA issue [#9568](https://issues.apache.org/jira/browse/CLOUDSTACK-9568)

I'll rebase as soon as possible.

rhtyd:
@murali-reddy can you help review this one?
@romain-dartigues thanks, can you also squash your commits to a single commit, and amend the git commit message to reflect the JIRA ID in the format: `CLOUDSTACK-9569: <one line summary>`.

rhtyd:
@romain-dartigues @murali-reddy ping

@blueorangutan package

romain-dartigues:
@rhtyd pong (is it OK like that? please advise)

systemvm/patches/debian/config/opt/cloud/bin/cs/CsHelper.py
rhtyd: @romain-dartigues can you make it return a `return []` after line 187. This would handle the case when wait if `False`, and still return empty array which may be consumed/necessary the the caller of the method. Please consider any other change necessary to handle the case that John has mentioned.

rhtyd:
Thanks @romain-dartigues that looks good. I'll kick a test round soon.
@blueorangutan package

rhtyd:
@blueorangutan test

romain-dartigues:
I'll need help to understand what gone wrong.

I just peeked at `pr1604-t379-kvm-centos7/test_vpc_router_nics_CC22OE/results.txt` and:
`TestVPCNics: DEBUG: SSH result: [u'PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.', u'64 bytes from 8.8.8.8: icmp_seq=1 ttl=52 time=15.6 ms', u'64 bytes from 8.8.8.8: icmp_seq=2 ttl=52 time=15.8 ms', u'64 bytes from 8.8.8.8: icmp_seq=3 ttl=52 time=15.7 ms', u'', u'--- 8.8.8.8 ping statistics ---', u'3 packets transmitted, 3 received, 0% packet loss, time 1999ms', u'rtt min/avg/max/mdev = 15.617/15.723/15.804/0.164 ms']; COUNT is ==> 0`

But `" 0% packet loss"` is in the string.

rhtyd:
@romain-dartigues I'll have a look at the failure, meanwhile can you see my comments against the changes https://github.com/apache/cloudstack/pull/1604/files

systemvm/patches/debian/config/opt/cloud/bin/cs/CsHelper.py
romain-dartigues: @jburwell < I think your request, while valid, goes beyond the scope of this patch which was made to fix a problem we have in production.

To wait for an amount of time, it would be required to spawn a separate thread, as the `timeout` argument was added in Python 3 (3.2 I think).

romain-dartigues:
@rhtyd < separate function done.

rhtyd:
okay @romain-dartigues I'll kick some tests
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@murali-reddy @abhinandanprateek can you help review this, thanks
@blueorangutan test

murali-reddy:
@romain-dartigues we fixed issue where ips.json has wrong gateway due to error in parsing the cmd_line.json. its committed in 4.8 and above [1]. Do you still see (or can you please verify)  performance hit due to arpping to wrong gateway on 4.8/4.9/master?

[1] https://github.com/apache/cloudstack/pull/1692/commits/93ac134f1e5c4c4c22029eca241f75362303837a

romain-dartigues:
@murali-reddy I'm not in position to run tests anymore
@dpassante could you take the lead for it?

rhtyd:
@romain-dartigues @dpassante any update on this?

romain-dartigues:
No comment, since it does not seems required anymore I close this request.

PR #1611, opened 2016-07-15
marvin: deploy clusters in separate threads

rhtyd:
This speeds up cluster deployment by Marvin, each cluster deployment can be
executed in separate threads. This for example, allows for setting up a Simulator
based environment with 1000s of hosts across 10s clusters in significantly
less time.

### DIFF
--- a/tools/marvin/marvin/deployDataCenter.py
+++ b/tools/marvin/marvin/deployDataCenter.py
@@ -34,10 +34,13 @@ from marvin.codes import (FAILED, SUCCESS)
 from marvin.lib.utils import (random_gen)
 from marvin.config.test_data import test_data
 from sys import exit
-import os
-import pickle
 from time import sleep, strftime, localtime
 from optparse import OptionParser
+
+import os
+import pickle
+import threading
+import Queue
 
 
 class DeployDataCenters(object):
@@ -197,7 +200,7 @@ class DeployDataCenters(object):
                 vmwareDc.zoneid = zoneId
                 self.addVmWareDataCenter(vmwareDc)
 
-            for cluster in clusters:
+            def createCluster(cluster):
                 clustercmd = addCluster.addClusterCmd()
                 clustercmd.clustername = cluster.clustername
                 clustercmd.clustertype = cluster.clustertype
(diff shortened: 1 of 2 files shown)


rhtyd:
One can test deployment of a large simulator environment using this cfg file with marvin+Simulator:
[1000.cfg.txt](https://github.com/apache/cloudstack/files/365853/1000.cfg.txt)

Without this fix, it takes more than 30 mins to get it to deploy. With this PR/fix, it takes less than 3 minutes.

tools/marvin/marvin/deployDataCenter.py
```diff
@@ -224,6 +225,16 @@ def createClusters(self, clusters, zoneId, podId, vmwareDc=None):
                                                podId,
                                                clusterId)
 
+            for cluster in clusters:
+            threads = []
+            for cluster in clusters:
```
pdube: Line 228 can be removed

pdube:
Nice. LGTM, code reviewed

tools/marvin/marvin/deployDataCenter.py
rhtyd: That's a typo, will fix it.

rhtyd:
Pinging for review - @wido @karuturi @koushik-das @abhinandanprateek @murali-reddy @jburwell 

This is marvin specific change, Travis validation is enough to show that deployDatacenter works.

rhtyd:
@blueorangutan package

tools/marvin/marvin/deployDataCenter.py
```diff
@@ -224,6 +225,15 @@ def createClusters(self, clusters, zoneId, podId, vmwareDc=None):
...
                                                clusterId)
 
+            threads = []
+            for cluster in clusters:
+                t = threading.Thread(target=createCluster, args=(cluster,))
+                t.start()
+                threads.append(t)
+
+            for t in threads:
+                t.join(1800)
```
koushik-das: Instead of hardcoding, better to set it based on host count in the cluster. Also consider that waitForHost() can wait upto 60 secs.

If one t.join() throws what happens to the other threads. Also will __cleanAndExit() behave consistently if there are exceptions in multiple threads.

rhtyd: Thanks I'll fix this later this week.

tools/marvin/marvin/deployDataCenter.py
```diff
@@ -224,6 +225,15 @@ def createClusters(self, clusters, zoneId, podId, vmwareDc=None):
                                                podId,
                                                clusterId)
 
+            threads = []
+            for cluster in clusters:
+                t = threading.Thread(target=createCluster, args=(cluster,))
```
jburwell: Would it make sense to create a fix sized thread pool and use a producer-consumer queue to enqueue work?

jburwell:
@rhtyd this enhancement/fix seems like it would be useful for on-going testing of LTS.  Therefore, could you please re-target this PR to the 4.9 release branch?  Also, could you please investigate the Jenkins failures?

tools/marvin/marvin/deployDataCenter.py
rhtyd: Sure, that makes sense. I'll fix that.

rhtyd:
I've re-implemented the fix with a bounded queue/thread size (limit to 50 threads max). On my local machine I was able to deploy a zone with 1000 hosts across 50 clusters and it took less than 3 minutes which previously took close to one hour.

There is no change other than marvin itself, a Travis test run would validate the change. I've also added a sample 1k zone deployment cfg.
@blueorangutan package

screenshot from 2017-08-01 22-06-38

rhtyd:
@blueorangutan test

PR #1618, opened 2016-07-29
[CLOUDSTACK-9643] Now returning os info with the list snapshot response

ProjectMoon:
This commit adds the ID and display name of the OS on the volume.


ProjectMoon:
Rebased to latest master.

rhtyd:
@blueorangutan package
@ProjectMoon can you push -f to rekick Travis?

server/src/com/cloud/api/ApiResponseHelper.java
```diff
@@ -157,6 +156,7 @@
 import org.apache.cloudstack.usage.Usage;
 import org.apache.cloudstack.usage.UsageService;
 import org.apache.cloudstack.usage.UsageTypes;
+import org.apache.commons.collections.CollectionUtils;
```
jburwell: Why was this import added?  I don't see it used anywhere in the patch.

server/src/com/cloud/api/ApiResponseHelper.java
```diff
@@ -479,6 +479,17 @@ public SnapshotResponse createSnapshotResponse(Snapshot snapshot) {
             if (zone != null) {
                 snapshotResponse.setZoneId(zone.getUuid());
             }
+
+            if (volume.getVolumeType().equals(Volume.Type.ROOT)) {
```
jburwell: Enumerated values are guaranteed to be singletons.   Therefore, the ``==`` operator deterministically performs both reference and value equality checks.  It is also preferred because it avoid an NPE in the event that ``volume.getVolumeType()`` returns null.

server/src/com/cloud/api/ApiResponseHelper.java
```diff
@@ -479,6 +479,17 @@ public SnapshotResponse createSnapshotResponse(Snapshot snapshot) {
...
+                VMInstanceVO instance = ApiDBUtils.findVMInstanceById(volume.getInstanceId());
+                if (instance != null) {
+                    GuestOS guestOs = ApiDBUtils.findGuestOSById(instance.getGuestOSId());
```
jburwell: It feels like lines 484 & 486 could be combined into a single join exposed by a method on ``VolumeDao``.  Do you agree?

api/src/org/apache/cloudstack/api/response/SnapshotResponse.java
```diff
@@ -199,4 +207,20 @@ public boolean isRevertable() {
     public void setRevertable(boolean revertable) {
         this.revertable = revertable;
     }
-}
+
+    public String getOsTypeId() {
```
koushik-das: The get* methods are not used, please remove.

api/src/org/apache/cloudstack/api/response/SnapshotResponse.java
```diff
@@ -102,6 +102,14 @@
     @Param(description = "indicates whether the underlying storage supports reverting the volume to this snapshot")
     private boolean revertable;
 
+    @SerializedName(ApiConstants.OS_TYPE_ID)
+    @Param(description = "id of the os on volume")
```
koushik-das: Please add "since" along with the description.

koushik-das:
Please add/modify Marvin tests based on these scenarios.

rhtyd:
@blueorangutan package

ProjectMoon:
Rebased to latest master. The comments will also be addressed soon, hopefully.

server/src/com/cloud/api/ApiResponseHelper.java
ProjectMoon: I was going to do this, but then I wondered... what exactly would it return? A pair of `Long`s perhaps? For now I will leave it as-is, pending more discussion.

server/src/com/cloud/api/ApiResponseHelper.java
ProjectMoon: It is used in ApiResponseHelper.

server/src/com/cloud/api/ApiResponseHelper.java
ProjectMoon: My mistake. I guess it would only return the ID of the Guest OS... so from the interface level it'd be from volume ID -> Guest OS ID?

ProjectMoon:
Code review issues have been addressed, with the exception of combining the Guest OS ID/VM Instance ID into a join method on VolumeDao. There was some compilation failure in the F5 plugin when I tried building locally. Will see if that happens here too.

rhtyd:
@ProjectMoon thanks, can you add a JIRA id for this and also use that in your commit's PR.
@blueorangutan package

ProjectMoon:
Updated to latest master, and now with a JIRA ticket.

https://issues.apache.org/jira/browse/CLOUDSTACK-9643

rhtyd:
Thanks @ProjectMoon 
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
LGTM, the failure in Travis is caused by env issue and not the PR itself. Merging this based on review, and test results.

PR #1629, opened 2016-08-06
CLOUDSTACK-9449: Fix typo in user role description

rhtyd:
Fixes description in default user role entry.


rhtyd:
@blueorangutan kick

rhtyd:
@blueorangutan help

pdion891:
Hum, This is a change to a file for an upgrade SQL to a release that is already GA. 
I think it would be preferable to get it into the 490_to_4100 file to avoid divergence between the voted release and the repo.

rhtyd:
@pdion891 thanks I'll fix that.

PR #1533, opened 2016-05-04
Convert patchviasocket to python (removes perl dependency for KVM agent)

sverrirab:
As requested here: https://github.com/apache/cloudstack/pull/1495

No scripts are using perl so that install requirement can be removed.
The new scripts are using standard python packages only.
Includes extensive unit test.
Note: perl-modules requirement is missing (fixed in mentioned PR) so do not merge that onto master.

PR #1633, opened 2016-08-10
[blocker] CLOUDSTACK-9452: use optparse instead of argparse in patchviasocket

rhtyd:
Use optparse that is supported by python 2.3+, instead of argparse that is
supported by python 2.7+. On KVM hosts which don't have python-argparse pkg
installed, the patchviasocket script will break. This fixes so that patchviasocket
would work on CentOS6 KVM hosts, or KVM hosts without python 2.7/argparse in
general.

The patchviasocket script was rewritten as a Python script from PR #1533

@blueorangutan package

### DIFF
--- a/scripts/vm/hypervisor/kvm/patchviasocket.py
+++ b/scripts/vm/hypervisor/kvm/patchviasocket.py
@@ -22,9 +22,10 @@
 # reads it from /dev/vport0p1 in cloud_early_config
 #
 
-import argparse
 import os
 import socket
+
+from optparse import OptionParser
 
 SOCK_FILE = "/var/lib/libvirt/qemu/{name}.agent"
 PUB_KEY_FILE = "/root/.ssh/id_rsa.pub.cloud"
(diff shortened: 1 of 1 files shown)


rhtyd:
@karuturi @jburwell @wido @sverrirab please review, it's a blocker for kvm centos6 hosts without argparse/python2.7 installed.

I could run an existing test: (gave same results before and after the change)
$ python ./scripts/vm/hypervisor/kvm/test_patchviasocket.py
ERROR: unable to connect to /tmp/tmpDuudOL.sck - Permission denied
.ERROR: unable to open /tmp/tmpoEzOGN.sck - Permission denied
.ERROR: ssh public key not found on host at must-not-exist
.ERROR: unable to connect to /tmp/tmpm2rGlP.sck - Connection refused
.ERROR: must-not-exist socket not found
..
Ran 6 tests in 1.633s

OK

wido:
Do we seriously want to write code that still works on CentOS 6 or systems without Python 2.7?

Looking at the Python website: https://docs.python.org/2/library/optparse.html

<pre>Deprecated since version 2.7: The optparse module is deprecated and will not be developed further; development will continue with the argparse module.</pre>

I wouldn't call this a blocker at all. CentOS 6 is pre-historic and we can't go back to using a old module like OptParse.

Sorry, -1 for me on this one.

rhtyd:
@wido I'm sorry we will need to support CentOS6 kvm hosts, just because you don't use it does not mean we should not support for people. It is a valid `blocker` as the KVM CentOS6 hosts have stopped working from deploying VMs.

`optparse` is deprecated which does not mean that it will fail or won't be available. We can have another way of fixing this: (1) revert #1533, (2) add a package dependency `python-argparse` on (at least centos6) rpms to ensure that argparse will be available on centos6/7 hosts. Which way should we go, if not the way PR fixes it?

sverrirab:
The updated script that now depends on python 2.7 (or python 2.6 + argparse I assume) is replacing the perl installation requirement that was there before.  Would it not be the correct solution to require python 2.7 to be installed?  If the hosts are dedicated KVM hosts that should be relatively low risk?

sverrirab:
The reason for my suggestion is that I don't think anyone should be running python 2.6 in production.   The last update for it was in 2008 so I would consider it risky at this point.

rhtyd:
CentOS6 default/epel repos don't have python 2.7 :) There are large installations running old code, old hardware, old kernels/OSs; they will get upgraded over time, but they need to be supported for the time being. Lastly, I could make the same argument for our usage of Java6/7, why aren't we running Java8. Java7 has eoled, but most of us I know are still using it in production.

rhtyd:
@wido @sverrirab will adding python-argparse as a dependency be an acceptable solution to the problem? this package is available on el6, but python 2.7 is not available (without using one of the 3rd party repositories).

sverrirab:
adding the python-argparse dependency is fine by me.  I have not tested that combination though.

wido:
@rhtyd I'm fine with adding it as a dependency.

Perception is different here I think. My personal opinion is that nobody should be running CentOS 6 with CloudStack in production just as much as you shouldn't use Ubuntu 12.04 anymore for CloudStack.

I'm against writing code which brings in legacy instead of removing it.

PR #1634, opened 2016-08-10
[blocker] CLOUDSTACK-9452: add python-argparse dependency on el6,7 rpms

rhtyd:
The patchviasocket script was rewritten in Python from PR #1533 and made
assumptions that Python 2.7 would be available. In case of CentOS, python 2.7
may not be available or installed. This change ensures that python-argparse
is installed which is used by this script.

/cc @wido @sverrirab @karuturi @jburwell 

@blueorangutan package

### DIFF
--- a/packaging/centos63/cloud.spec
+++ b/packaging/centos63/cloud.spec
@@ -96,6 +96,7 @@ management, and intelligence in CloudStack.
 %package common
 Summary: Apache CloudStack common files and scripts
 Requires: python
+Requires: python-argparse
 Obsoletes: cloud-test < 4.1.0 
 Obsoletes: cloud-scripts < 4.1.0
 Obsoletes: cloud-utils < 4.1.0
--- a/packaging/centos7/cloud.spec
+++ b/packaging/centos7/cloud.spec
@@ -91,6 +91,7 @@ management, and intelligence in CloudStack.
 %package common
 Summary: Apache CloudStack common files and scripts
 Requires: python
+Requires: python-argparse
 Group:   System Environment/Libraries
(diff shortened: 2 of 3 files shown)


sverrirab:
+1 from me (for what it's worth)

jburwell:
LGTM for code

rhtyd:
@karuturi @jburwell can we merge this, this is only packaging related fix. Using above packages, installation confirmed that argparse was installed on el6/el7 kvm hosts.

rhtyd:
@blueorangutan test

jburwell:
@rhtyd the test results from blueorangutan are looking pretty good.  A couple of observations:

1. The markdown for the table seems to be invalid as it is not rendering to a table
2. Would it be possible to sort the rows in Failure, Skip, Success order?  We are most interested in failures and skips.  Placing them at the top will speed visual scanning up a bit.
3.  Include a summary of the environment used to run the tests

rhtyd:
@jburwell  thanks

- I've fixed the markdown issues -- see this for example -- https://github.com/apache/cloudstack/pull/1638#issuecomment-241644475
- I'll fix the sorting issues, to sort status column
- I've added summary at the bottom of the table, noting the hypervisor, mgmt server os, zone type and link to the marvin logs for anyone to inspect further

rhtyd:
We've enough LGTMs and test result (with vmware), I'll fire another Trillian job with KVM.

rhtyd:
Merging based on lgtms and test results (the failures are due to env issues, and failing tests not related to the pkging fix).

PR #1632, opened 2016-08-09
Honor the forced parameter for stopVirtualMachine api call

nathanejohnson:
CLOUDSTACK-9451

https://issues.apache.org/jira/browse/CLOUDSTACK-9451

PR #1635, opened 2016-08-11
CLOUDSTACK-9451

nathanejohnson:
https://issues.apache.org/jira/browse/CLOUDSTACK-9451

Re-doing against 4.8 since this is a bug, not a feature.

### DIFF
--- a/engine/api/src/org/apache/cloudstack/engine/cloud/entity/api/VirtualMachineEntity.java
+++ b/engine/api/src/org/apache/cloudstack/engine/cloud/entity/api/VirtualMachineEntity.java
@@ -115,6 +115,12 @@ public interface VirtualMachineEntity extends CloudStackEntity {
     boolean stop(String caller) throws ResourceUnavailableException, CloudException;
 
     /**
+     * Stop the virtual machine, by force if necessary
+     *
+     */
+    boolean stopForced(String caller) throws ResourceUnavailableException, CloudException;
+
+    /**
      * Cleans up after any botched starts.  CloudStack Orchestration Platform
      * will attempt a best effort to actually shutdown any resource but
      * even if it cannot, it releases the resource from its database.
--- a/engine/api/src/com/cloud/vm/VirtualMachineManager.java
+++ b/engine/api/src/com/cloud/vm/VirtualMachineManager.java
@@ -90,6 +90,8 @@ public interface VirtualMachineManager extends Manager {
 
     void stop(String vmUuid) throws ResourceUnavailableException;
 
+    void stopForced(String vmUuid) throws ResourceUnavailableException;
+
     void expunge(String vmUuid) throws ResourceUnavailableException;
 
     void registerGuru(VirtualMachine.Type type, VirtualMachineGuru guru);
--- a/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -1239,6 +1239,18 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         } catch (final ConcurrentOperationException e) {
             throw new CloudRuntimeException("Unable to stop vm because of a concurrent operation", e);
         }
+
+    }
+
+    @Override
+    public void stopForced(String vmUuid) throws ResourceUnavailableException {
+        try {
+            advanceStop(vmUuid, true);
+        } catch (final OperationTimedoutException e) {
+            throw new AgentUnavailableException("Unable to stop vm because the operation to stop timed out", e.getAgentId(), e);
+        } catch (final ConcurrentOperationException e) {
+            throw new CloudRuntimeException("Unable to stop vm because of a concurrent operation", e);
+        }
     }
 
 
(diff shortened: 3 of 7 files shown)


kiwiflyer:
References: https://github.com/apache/cloudstack/pull/1632

ustcweizhou:
LGTM

jburwell:
@nathanejohnson please amend your commit message to explain the change(s) made and their motivation?

kiwiflyer:
@ProjectMoon Jeff, could you help review this, as I believe you were the first to point it out on the list.

From talking to Nathan, it sounds like we're not sure KVM is actually affected by this bug. He was in the process of doing some digging. Regardless of that, it probably affects other hypervisors.

ProjectMoon:
All hypervisors are affected by this, since the parameter is just not passed into the VM Manager. That means its effects don't work at all anywhere. :) I will try to review it to tomorrow.

kiwiflyer:
I'll let him comment on this, but If I recall, it had more to do with implementation of force stopping with KVM rather than the lack of the parameter being passing into the VM Manager.

engine/api/src/org/apache/cloudstack/engine/cloud/entity/api/VirtualMachineEntity.java
```diff
@@ -115,6 +115,12 @@ String reserve(DeploymentPlanner plannerToUse, @BeanParam DeploymentPlan plan, E
     boolean stop(String caller) throws ResourceUnavailableException, CloudException;
 
     /**
+     * Stop the virtual machine, optionally force
+     *
+     */
+    boolean stop(String caller, boolean forced) throws ResourceUnavailableException, CloudException;
```
jburwell: [Flag arguments](http://martinfowler.com/bliki/FlagArgument.html) are a nasty code smell/anti-pattern.  Please consider changing method to something more descriptive with a flag argument such as ``forceStop(String caller)``.  The contrasting version of the method already exists on the interface so it would be a compatible change.

nathanejohnson: For better or worse, I was just trying to follow the lead of the actual method that does the work in question, namely advanceStop, and some of the methods called from advanceStop.  I have no issue changing the name to have force in it and skip the bool flag (lest Fowler think I'm a bad programmer), but I am not thinking that I'd like to change the signature of advanceStop or any other the other methods that eventually get called in the chain where there is a bool flag in it.  I think the much bigger issue with this bit of code is that the refactor that lead to this issue in the first place abstracted the call to advanceStop under so many layers that it made it really hard to notice that functionality was reduced in the first place, to the point where the author nor anyone reviewing caught it.  If I was to place serious energy into refactoring, I'd probably start with undoing a lot of that abstraction, but that would require me being familiar with a lot more of the codebase than I currently am.  Also, more immediately, I want to come up with a good test for this, but so far haven't been able to come up with something reliable / reproducible under kvm.

jburwell: @nathanejohnson I wasn't thinking to change any of the existing code.  I am just interested in avoiding incurring any more flag argument technical debt.

In terms of testing, this change is not hypervisor specific.  Therefore, a test case that utilizes the simulator might be a more fruitful path.  @rhtyd may have some thoughts as to how you could accomplish it.

rhtyd: @jburwell the `forced` flag has always existed, the `stopVirtualMachine` API accepts this and the query layer too has a method with this signature (https://github.com/apache/cloudstack/pull/1635/files#diff-0a1cd9df984252594918eec5acfed08cR3850) but it does not honour or pass this flag to the layers below, so effectively stopVM API with/without the `forced` flag has the same effect.

jburwell: @rhtyd since this PR adds new methods, let's not carry it forward.  The implementation of ``forceStop`` would end up passing ``true`` for the ``forced`` flag down the stack, but at least we are not expanding an anti-pattern.

jburwell:
@nathanejohnson I want to get this bug fix into 4.8.2.0.  In order to make it, we need to complete the following:

  * [] Resolve the question about the name of the method (``stop`` vs ``forcedStop``)
  * [] Amend the commit message to summarize the change and explain the reason/motivation for it
  * [] Code review LGTM
  * [] Test LGTM

@rhtyd or I will provide the code review LGTM once we resolve the naming question.  Do you have someone who can run the test?  Also, is this change something we can check via a Marvin test case?

nathanejohnson:
@jburwell sorry I've been silent, I've been on vacation for a bit and I will be juggling plans this weekend as well.  I plan to update this PR soon, and as far as a Marvin test I am not aware of any current tests that would cover this.  I haven't forgotten about this but it might be a few days yet before I can dig in fully.

nathanejohnson:
I did find one marvin test that does pass forced flag.  Here are the results.

https://gist.github.com/nathanejohnson/c4d1ffb8b61b19a8105dcb5326794b15

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@nathanejohnson can you reply to @jburwell 's comments

nathanejohnson:
@rhtyd @jburwell  I believe the last commit should have addressed his concerns.

rhtyd:
@blueorangutan package

rhtyd:
Checked centos6 failed due to an intermittent unit test failure, we can ignore that for now.
@blueorangutan test

rhtyd:
Test LGTM, the two failures are intermittent and env related.

PR #1541, opened 2016-05-12
Systemd packaging for Ubuntu 16.04

wido:
This PR is for introducing packaging for Ubuntu 16.04

Ubuntu 14.04 still uses sysvinit and Java 7. 16.04 uses systemd and Java 8.

The 'build-deb.sh' script can be used to generate packages. The comments in the file show how to use Docker to build for the different versions of Ubuntu.

I've just build:
* cloudstack-agent_4.9.0-SNAPSHOT~xenial_all.deb
* cloudstack-agent_4.9.0-SNAPSHOT~trusty_all.deb

PR #1647, opened 2016-08-19
[lts] CLOUDSTACK-9462: Systemd support for Ubuntu 16.04

rhtyd:
Created this based on @wido 's origin PR #1541 .

Requesting for review and testing -- @jburwell @karuturi @wido @vincentbernat 
@wido I think this change only brings systemd support to agent and usage packages, or does cloudstack-management pkg has systemd support too?

@blueorangutan package

- systemd: Add a /etc/sysconfig/cloudstack-* file

  This allows users to easily override variables passed to Java when
  starting up.

  It also creates a foundation for sharing the systemd service profile
  between CentOS and Ubuntu since it only requires the environment file
  to be changed.

- deb: Add Ubuntu 16.04 support

  Ubuntu 16.04 differs from Ubuntu 14.04 in a few ways:
  - systemd instead of sysvinit / upstart
  - Java 8 support

  The packaging now detects on which distribution it is being
  build and based on that it installs different files in the
  packages, but it also changes the Dependencies.


vincentbernat:
LGTM for the Ubuntu part. You could share `/etc/sysconfig/*` and `/etc/default/*` files between RHEL/Ubuntu (same files, just installed at different locations). Same for the service files:

```
[service]
EnvironmentFile=-/etc/default/cloudstack-agent
EnvironmentFile=-/etc/sysconfig/cloudstack-agent
```

wido:
@rhtyd No, it doesn't do the management server.

Let's just merge this PR and make a new one for that. I don't want to delay merging this any longer.

Thanks @vincentbernat for the comment, that's something for a next PR.

rhtyd:
@wido given it's a pure packaging fix, and as per RMs we've enough LGTMs but we lack a test LGTM -- I can help merge this after some tests wrt usage and kvm agent (on ubuntu 16.04) over the weekend or next week.

wido:
I did the test myself, but since it's my code I can't LGTM this.

rhtyd:
@wido with this change will we have to host two versions of deb repository -- one built on ubuntu without systemd (such as 14.04) and one built on ubuntu with systemd (16.04+)?

jburwell:
@wido While I understand that it is quite old and presents maintenance challenges, it remains supported until [April 2017](https://wiki.ubuntu.com/Releases).  Therefore, it would be reasonable for users to expect us to continue to support it until EOL.  I have opened conversation on dev@ and user@ in an effort to determine the impact (if any) to users post 4.9.  

In the meantime, would it be possible to split this PR into Ubuntu 16.04 support and removal of Ubuntu 12.04 support?  If we decide to remove it, I propose that we introduce it in the 4.10.x release.  Not only is 4.9 an LTS branch, it doesn't seem reasonable to drop support for a distribution in a patch release (i.e. 4.9.0.0 supports Ubuntu 12.04 and 4.9.1.0 does not).  Conversely, Ubuntu 16.04 support seems like a worthy addition to LTS.  Splitting these two issues will provide the flexibility to implement such removal of Ubuntu 12.04 in 4.10.x and the addition of Ubuntu 16.04 in 4.9.x+.  Does this approach seem reasonable?

rhtyd:
@wido based on the ML discussion, I've removed the specific change that removed 12.04 support. Can you send another PR for master for removing support for 12.04.

debian/rules
```diff
@@ -5,8 +5,17 @@ PACKAGE = $(shell dh_listpackages|head -n 1|cut -d '-' -f 1)
 SYSCONFDIR = "/etc"
 DESTDIR = "debian/tmp"
 
+ifeq ($(shell lsb_release -sr), 14.04)
+        SUBSTVARS = -Vjre:Depends="openjdk-7-jre-headless" -Vjdk:Depends="openjdk-7-jdk" -Vinit:"Depends:jsvc"
+else
+        SUBSTVARS = -Vjre:Depends="openjdk-8-jre-headless" -Vjdk:Depends="openjdk-8-jdk" -Vinit:"Depends:systemd"
```
rhtyd: @wido should this be written like:   `-Vinit:Depends="systemd"`. Would it be possible to support both systemd/initd environments by adding the dependency like `systemd | jsvc` -- this way it would try to install both? On (systemd based) Ubuntu 16.04 systemd will be already installed, we can install jsvc for all systems -- so why not just depend on jsvc alone? On 14.04, the pkgs can continue to use jsvc/initd script and on 16.04 it can uses systemd, both using the same packages.

I want us to avoid hosting two separate deb repositories (one for non-systemd based and one for systemd based) if possible. Thoughts?

wido: That might be possible, I am not sure. I think we could do it. Haven't tried that.

I'm however very, very short on time these weeks, so I won't be able to work on it.

wido:
@rhtyd Why another PR for against master? I thought we were merging forward?

This stuff is really starting to confuse me.

wido:
@rhtyd When this one merges and is merged forward I will send a PR for master to remove 12.04 support.

debian/rules
rhtyd: Thanks @wido I'll try to build experimental packages to test if it just might work.

rhtyd:
Thanks @wido that makes sense.

vincentbernat:
it is not possible to have systemd support and build on 12.04 at the same time. dh-systemd doesn't exist on 12.04 and its actions are quite important. If you remove it and just install the units manually, they won't start on boot or on upgrade.

rhtyd:
@vincentbernat I made some changes, built packages and was able to install and run agent and usage server on both Ubuntu 12.04, 14.04 using the same packages. 
@wido @vincentbernat can you review my change and if it's okay, based on installation tests we can merge this.

Additional notes: on fresh installation the /etc/default/cloudstack-{agent, usage} files use JAVA_HOME path for 1.8 jre which does not exist and needed manually fixing. This was the only thing I had to do and can be documented in install/upgrade and release notes.

wido:
@rhtyd Seems good, but it would still not force the install of Java 8 on Ubuntu 16.04

However, I don't think that's a blocker.

LGTM for me

rhtyd:
@wido actually on a fresh system it would first attempt to find and install openjdk8, on existing ones if openjdk7 is installed then it won't force openjdk8 by default. Thanks.

vincentbernat:
Otherwise, I think the change to be compatible with 12.04 is OK (no Upstart script should ever be provided in this case).

debian/rules
```diff
@@ -5,18 +5,9 @@ PACKAGE = $(shell dh_listpackages|head -n 1|cut -d '-' -f 1)
 SYSCONFDIR = "/etc"
 DESTDIR = "debian/tmp"
 
-ifeq ($(shell lsb_release -sr), 14.04)
-        SUBSTVARS = -Vjre:Depends="openjdk-7-jre-headless" -Vjdk:Depends="openjdk-7-jdk" -Vinit:"Depends:jsvc"
-else
-        SUBSTVARS = -Vjre:Depends="openjdk-8-jre-headless" -Vjdk:Depends="openjdk-8-jdk" -Vinit:"Depends:systemd"
-endif
-
```
rhtyd: @vincentbernat won't this force some release of Java8, what if the openjdk/oracle java8 pkg is not available on that system, say on 12.04 etc.?

vincentbernat: That's why I think that the use of `$SUBSTVARS` and `dh_gencontrol` override was a good idea.

rhtyd: @vincentbernat that's right, but if we use that then we'll need to build/host/publish two version of the deb repository, depending on whether the deb pkgs were built of Ubuntu 14.04 or 16.04+. I think the current is still a good compromise, I can add new dependency options to allow users to use oracle jre/jdk.

rhtyd:
Thanks @vincentbernat I'll update the dependencies, and in the post-installation step we can modify the default JAVA_HOME path of the host in the default/{cloudstack-usage, cloudstack-agent} files if available.

debian/rules
vincentbernat: Oh, OK, that's why you removed it. OK.

rhtyd:
@wido @vincentbernat I've incorporate all the changes now, including refactoring of systemd file and now we've both centos and ubuntu/debian packaging consuming from the the same service/default systemd files. I tried to see if we can include support for cloudstack-management and systemd but that's challenging and different. Please review and let's merge this.

packaging/centos7/cloud.spec
```diff
@@ -482,7 +488,7 @@ pip install --upgrade /usr/share/cloudstack-marvin/Marvin-*.tar.gz
 %dir %attr(0770,root,cloud) %{_localstatedir}/cache/%{name}/management/work
 %dir %attr(0770,root,cloud) %{_localstatedir}/cache/%{name}/management/temp
 %dir %attr(0770,root,cloud) %{_localstatedir}/log/%{name}/management
-%config(noreplace) %{_sysconfdir}/sysconfig/%{name}-management
```
vincentbernat: RHEL users won't expect this file to be in /etc/default. In the systemd unit:

```
EnvironmentFile=-/etc/default/cloudstack-agent
EnvironmentFile=-/etc/sysconfig/cloudstack-agent
```

debian/cloudstack-agent.postinst
```diff
@@ -41,6 +41,12 @@ case "$1" in
             mkdir /etc/libvirt/hooks
         fi
         cp -a /usr/share/cloudstack-agent/lib/libvirtqemuhook /etc/libvirt/hooks/qemu
+
+        # Update JAVA_HOME in /etc/default/cloudstack-agent to default JRE
+        JAVA_HOME=$(readlink -f /usr/bin/java | sed 's/\/jre.*java//g')
+        if [ -d "$JAVA_HOME" ]; then
+            sed -i "s:^JAVA_HOME=.*:JAVA_HOME=${JAVA_HOME}:" /etc/default/cloudstack-agent
+        fi
```
vincentbernat: This would overwrite a change made by a user during upgrades. I would suggest to move the first line directly in /etc/default/cloudstack-agent (if it's compatible with RHEL). A user that didn't change anything will get the new file and it will work as expected. A user that did alter the file will get a prompt to solve the situation.

However, this won't work with systemd without further ugly hacks.

On my system, I notice that `/usr/lib/jvm/default-java` is a symlink. It also seems to be the case on Precise. Maybe you could patch `/etc/default/cloudstack-agent` after installing it in `debian/rules`.

packaging/centos7/cloud.spec
rhtyd: @vincentbernat thanks, I was not sure but found that /etc/default exists and is used by several packages on CentOS and it would keep paths uniform across distributions. Should I revert back to sysconfig for CentOS packages then?

vincentbernat: I am a Debian guy, so I can't give you an authoritative answer. I didn't know that `/etc/default` already existed on CentOS too. I suppose that's up to you. :)

rhtyd: Cool, I see several packages hosting default config (such as what we've here) at `/etc/default` so I'll keep 'em in `/etc/default` to make things consistent.

debian/cloudstack-agent.postinst
rhtyd: @vincentbernat  thanks, it makes sense. I'll see if we can move this to the `default/files`. An alternative is to document this, we can keep some default path and in our installation/upgrade documentation suggest the admin to fix the JAVA_HOME here if it's something non-default. The code I've used here simply replaces JAVA_HOME to whatever is set globally on that host.

wido:
LGTM for me right now.

One side note, I hate changes being made in postinst files. But I wouldn't know a different route though.

rhtyd:
Thanks @wido I'll see if I can fix the issue without including the changes in postinst files.

rhtyd:
@wido @vincentbernat  I could not get rid of the postinst step as shell is not executed when I move the code in the EnvironmentFile itself. I've reduced its impact to only run when the default /etc/default/cloudstack-{usage, agent} are installed.

Since we're re-using the common systemd files, the JAVA_HOME declared in the files are set to the default we see on CentOS and only in that case the postinst step will replace it with the system default. On future upgrades, the postinst code won't run. Sorry, but this hack was needed.

Please see one last time, if we're in agreement as soon as my packaging testing succeed I'll merge this. /cc @jburwell @karuturi

rhtyd:
Packaging tests confirmed, PR is OK. Debian packages tested on Ubuntu 12.04/14.04/16.04, rpms tests on CentOS 7.2. Both usage, agent came up on installation.

rhtyd:
Update: postinstall code has been removed now @wido @vincentbernat

packaging/systemd/cloudstack-usage.default
```diff
@@ -0,0 +1,25 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+JAVA_HOME=/usr/lib/jvm/jre
+if [ ! -f "$JAVA_HOME"/bin/java ] ; then
+    JAVA_HOME=$(update-alternatives --display java | grep 'currently points to' | sed 's:.*currently points to ::' | sed 's:/bin/java::')
+fi
```
vincentbernat: How is that supposed to work with systemd? Moreover, the `readlink -f` solution was far more robust than parsing `update-alternatives` output. Let me propose something simpler inside `debian/rules` instead.

debian/rules
```diff
@@ -35,13 +37,20 @@ override_dh_auto_install:
...
 	install -D plugins/hypervisors/kvm/target/dependencies/* $(DESTDIR)/usr/share/$(PACKAGE)-agent/lib/
-	install -D packaging/debian/init/cloud-agent $(DESTDIR)/$(SYSCONFDIR)/init.d/$(PACKAGE)-agent
+
+	install -m0755 packaging/debian/$(PACKAGE)-agent.init $(DESTDIR)/$(SYSCONFDIR)/init.d/$(PACKAGE)-agent
+	install -d -m0755 debian/$(PACKAGE)-agent/lib/systemd/system
+	# Fix libvirt service name for Debian/Ubuntu
+	sed -i 's/Requires=libvirtd.service/Requires=libvirt-bin.service/g' packaging/systemd/$(PACKAGE)-agent.service
+	sed -i 's/After=libvirtd.service/After=libvirt-bin.service/g' packaging/systemd/$(PACKAGE)-agent.service
+	install -m0644 packaging/systemd/$(PACKAGE)-agent.service debian/$(PACKAGE)-agent/lib/systemd/system/$(PACKAGE)-agent.service
+	install -m0644 packaging/systemd/$(PACKAGE)-agent.default $(DESTDIR)/$(SYSCONFDIR)/default/$(PACKAGE)-agent
```
vincentbernat: Here:

```
sed 's+JAVA_HOME=.*+JAVA_HOME=/usr/lib/jvm/default-java+' $(DESTDIR)/$(SYSCONFDIR)/default/$(PACKAGE)-agent
```

If people don't want to use the default JRE, they are free to change `/etc/default/cloudstack-agent`.

packaging/systemd/cloudstack-usage.default
rhtyd: @vincentbernat okay I'll revert it back to the readlink solution.

rhtyd:
@vincentbernat @wido thank you for your suggestion and review, I've pushed the final version from my end. Packages works on both CentOS7 and Ubuntu 12.04/14.04/16.04.

debian/rules
rhtyd: This has one issue if there is no /usr/lib/jvm/default-java installed on either Ubuntu 14.04/16.04 using the default repositories. I've reverted using the readlink solution now. Thanks.

vincentbernat: OK, the symlink comes with default-jre-headless package. If you don't have it, you don't have the symlink.

packaging/systemd/cloudstack-agent.default
```diff
@@ -0,0 +1,24 @@
...
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+JAVA_HOME=/usr/lib/jvm/jre
+if [ ! -f "$JAVA_HOME"/bin/java ] ; then
+    JAVA_HOME=$(readlink -f /usr/bin/java | sed 's:/bin/java::')
+fi
```
vincentbernat: Are you sure this will work with systemd? On Ubuntu, the best it could do is to keep `JAVA_HOME=/usr/lib/jvm/jre` which is not valid on Ubuntu. Maybe, it would be easier to push that inside the ExecStart (and equivalent in init.d):

```
if [ -n "$JAVA_HOME" ]; then
  export JAVA_HOME=$(readlink -f /usr/bin/java | sed 's:/bin/java::')
fi
```

Or, keeping it for RHEL in the default file but using in ExecStart:

```
[ -d "$JAVA_HOME" ] || export JAVA_HOME=$(readlink -f /usr/bin/java | sed 's:/bin/java::')
```

rhtyd: @vincentbernat yes this worked for me both on CentOS 7.2, Ubuntu 16.04 (systemd). Can you build packages and help with a secondary test. When the service starts it executes the environment/default file which gives us the $(readlink...) to replace $JAVA_HOME usage. I see this in my CentOS 7.2 environment (and similar on Ubuntu 16.04):

```
 root     27929  0.0  0.0 115340  1616 ?        Ss   20:57   0:00 /bin/sh -ec      export UCP=`ls /usr/share/cloudstack-usage/cloud-usage-*.jar /usr/share/cloudstack-usage/lib/*.jar /usr/share/cloudstack-mysql-ha/lib/*.jar | tr "\n" ":"`;      export CLASSPATH="$UCP:/etc/cloudstack/usage:/usr/share/java/mysql-connector-java.jar";      $(readlink -f /usr/bin/java | sed 's:/bin/java::')/bin/java -Dpid=$$ -Xms256m -Xmx2048m -cp "$CLASSPATH" $JAVA_CLASS
 root     27936  7.4 13.3 3616388 251860 ?      Sl   20:57   0:07 java -Dpid=27929 -Xms256m -Xmx2048m -cp /usr/share/cloudstack-usage/cloud-usage-4.9.1.0-SNAPSHOT.jar:/usr/share/cloudstack-usage/lib/activation-1.1.jar:/usr/share/cloudstack-usage/lib/annotations-2.0.1.jar:/usr/share/cloudstack-usage/lib/antisamy-1.5.3.jar:/usr/share/cloudstack-usage/lib/aopalliance-1.0.jar
```

rhtyd:
@wido @vincentbernat and others -- any objections, this is good to merge?

packaging/systemd/cloudstack-agent.default
vincentbernat: I think the result is quite fragile. systemd just reads the first JAVA_HOME, then the second one, ignoring the condition. You could just have (without the condition):

```
JAVA_HOME=$(readlink -f /usr/bin/java | sed 's:/bin/java::')
```

The only use of JAVA_HOME seems to be to choose a Java vesion. Why not just, in `/etc/default/...`:

```
JAVA=/usr/bin/java
```

And in the service file:

```
ExecStart=/bin/sh -ec '\
     export ACP=`ls /usr/share/cloudstack-agent/lib/*.jar /usr/share/cloudstack-agent/plugins/*.jar 2>/dev/null|tr "\\n" ":"`; \
     export CLASSPATH="$ACP:/etc/cloudstack/agent:/usr/share/cloudstack-common/scripts"; \
${JAVA} -Xms${JAVA_HEAP_INITIAL} -Xmx${JAVA_HEAP_MAX} -cp "$CLASSPATH" $JAVA_CLASS'
```

People could still overload the version of Java used by changing the JAVA variable.

If you are uncomfortable with this, at least, just remove the condition and simplify the default file to be:

```
JAVA_HOME=$(readlink -f /usr/bin/java | sed 's:/bin/java::')
JAVA_HEAP_INITIAL=256m
JAVA_HEAP_MAX=2048m
JAVA_CLASS=com.cloud.agent.AgentShell
```

From your tests, it works both on Ubuntu and CentOS.

rhtyd: Thanks @vincentbernat that's a nice idea. I've removed usage of JAVA_HOME, I checked and indeed all we want was path to java which I checked by default is at /usr/bin/java on most distros including the ones I tested - Ubuntu 12.04, 14.04, 16.04 and on CentOS 7(.2). Are we good to merge with the new change?

rhtyd:
Alright, given the reviews and that all changes have been incorporated with exhaustive testing with Ubuntu and CentOS I'll proceed with merging this without blocking 4.9.1.0 RC much further.

Any more improvements and changes should be sent in future PRs for example support for cloudstack-management/systemd on Ubuntu 16.04 is yet to be done. Thanks @vincentbernat and @wido for your review and feedback.

PR #1650, opened 2016-08-19
fix macchinina template specificied for vmware in Marvin tests

PaulAngus:
fix macchinina template specificied for vmware - .vmdk.bz2 is not valid… for CloudStack. edit to use .ova

### DIFF
--- a/test/integration/smoke/test_internal_lb.py
+++ b/test/integration/smoke/test_internal_lb.py
@@ -254,7 +254,7 @@ class Services:
                     "format": "ova",
                     "hypervisor": "vmware",
                     "ostype": "Other PV (64-bit)",
-                    "url": "http://dl.openvm.eu/cloudstack/macchinina/x86_64/macchinina-vmware.vmdk.bz2",
+                    "url": "http://dl.openvm.eu/cloudstack/macchinina/x86_64/macchinina-vmware.ova",
                     "requireshvm": "True",
                 }
             }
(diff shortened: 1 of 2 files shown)


rhtyd:
LGTM.

jburwell:
@abhinandanprateek are these two test suites where we should be using ``vm.instance`` config variable instead specifying the template directly in the test suite?

serg38:
We tried macchinina template during test cases on ESX with E1000 adapter and VMs doesn't come with eth0 causing test_internal_lb.py and test_vpc_vpn.py to fail on a real hypervisor test. Has anyone actually tried it running on Vmware?

jburwell:
@PaulAngus @abhinandanprateek could you investigate the VMware issue described by @serg38?

jburwell:
@PaulAngus this issue seems to affect 4.8+.  Since we have one more 4.8 (4.8.2.0) release, it would be nice to have this fix for those test runs.  Therefore, could you re-point it to the 4.8 branch?

rhtyd:
LGTM. @abhinandanprateek is the changes okay, or we should refactor move them to the default test_data file?

rhtyd:
I've moved the changes to PR #1692  this can be closed now.

rhtyd:
This has been merged/fixed with #1692 and fwd-merges, please close this @PaulAngus

PR #1648, opened 2016-08-19
test/integration: fix tearDown order in list_acl_ tests

rhtyd:
test/integration: fix tearDown order in list_acl_ tests
    
    In several of the list_acl_tests, the tests run for simulator only where
    in the (class) setup domains and accounts are created for the test. When the
    tests end the (class) teardown methods would delete and remove these resources.
    Due to dependence of one of the resources on the other, domain2 on domain1,
    domain2 needs to be removed/cleaned up before domain1. Due to this issue,
    several Travis test runs have failed in the past such as:
    
    https://travis-ci.org/apache/cloudstack/jobs/152610967
    https://travis-ci.org/apache/cloudstack/jobs/152610968
    
    Changing the order of cleanup fixes the tests.

/cc @jburwell @karuturi 

The fix is specific to tests that run 'only' on simulator with Travis. A passing Travis run should be enough to validate the changes.

PR #1652, opened 2016-08-19
Marvin Tests: Fix VPC network offering selection .

PaulAngus:
It appears that original code did not expect there to be an alternate default VPC offering (Nuage) .
VPC selection uses isdefault=True  which selects the Nuage Network offering, stopping the test from being run.
Fix was to look for name="Default VPC Offering" instead.

### DIFF
--- a/test/integration/smoke/test_network_acl.py
+++ b/test/integration/smoke/test_network_acl.py
@@ -64,7 +64,7 @@ class TestNetworkACL(cloudstackTestCase):
         self.assert_(networkOffering is not None and len(networkOffering) > 0, "No VPC based network offering")
 
         # 1) Create VPC
-        vpcOffering = VpcOffering.list(self.apiclient,isdefault=True)
+        vpcOffering = VpcOffering.list(self.apiclient, name="Default VPC offering")
         self.assert_(vpcOffering is not None and len(vpcOffering)>0, "No VPC offerings found")
         self.services["vpc"] = {}
         self.services["vpc"]["name"] = "vpc-networkacl"


rhtyd:
LGTM, the Travis test failures are not related to this PR but with tearDown issues fixed here: https://github.com/apache/cloudstack/pull/1648

koushik-das:
@PaulAngus Shouldn't the product code be fixed instead of the test? If the isDefault property on VPC offerings in not working then that needs to be fixed. I checked the vpc_offerings table in the DB and there is a 'default' field.

PaulAngus:
Without full understanding of the Nuage plugin, fixing the test seemed the safest path, one hopes that the nuage offering was configured that way for a good reason.

fmaximus:
As Nuage plugin developer, I can't come up with any good reason why it should be default.
Also I see that Redundant VPC Offering is set to be a default.
Both are most likely just copied from defaultVPCOfferingName creation.

PaulAngus:
Thanks frank, would you make that change to the Nuage Network offering or shall I?

koushik-das:
There should be only one default VPC offering, unless I am missing out something obvious :)

fmaximus:
I will make the change. I also plan to only create the vpc offering if there are Nuage VSP devices configured, which makes more sense for the users of CloudStack that do not use our plugin.

jburwell:
@PaulAngus does this defect occur in 4.9 as well?  If so, please re-target the PR to the 4.9 branch.

jburwell:
@PaulAngus following up on PR feedback.  @fmaximus did you apply this patch in another PR?  If so, could you please provide a reference?

If this patch has not been addressed elsewhere, I would like to get this fix into the upcoming 4.9.10 and 4.10.0.0 releases for which RCs will be cut on 25 Sept 2016.  Assuming this issue is also present in 4.9, could you please re-target the PR and investigate the Jenkins failure.

jburwell:
@PaulAngus is their PR superceded by #1688?

PR #1673, opened 2016-09-15
CLOUDSTACK-9071: Properly parse stats.output.uri in StatsCollector

wido:
Both host and path could have been NULL which causes the StatsCollector
no to start properly.

By checking if the Strings are not Empty or Null we make sure the StatsCollector
always runs and does not prevent the Management Server from starting.

### DIFF
--- a/server/src/com/cloud/server/StatsCollector.java
+++ b/server/src/com/cloud/server/StatsCollector.java
@@ -34,6 +34,7 @@ import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 
 import org.apache.cloudstack.utils.usage.UsageUtils;
+import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.springframework.stereotype.Component;
 
@@ -271,12 +272,18 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
                     s_logger.info(scheme + " is not a valid protocol for external statistics. No statistics will be send.");
                 }
 
-                externalStatsHost = uri.getHost();
+                if (!StringUtils.isEmpty(uri.getHost())) {
+                    externalStatsHost = uri.getHost();
+                }
+
                 externalStatsPort = uri.getPort();
-                externalStatsPrefix = uri.getPath().substring(1);
+
+                if (!StringUtils.isEmpty(uri.getPath())) {
+                    externalStatsPrefix = uri.getPath().substring(1);
+                }
 
                 /* Append a dot (.) to the prefix if it is set */
-                if (externalStatsPrefix != null && !externalStatsPrefix.equals("")) {
+                if (!StringUtils.isEmpty(externalStatsPrefix)) {
                     externalStatsPrefix += ".";
                 } else {
                     externalStatsPrefix = "";


server/src/com/cloud/server/StatsCollector.java
```diff
@@ -285,9 +286,15 @@ private void init(Map<String, String> configs) {
...
+
                 externalStatsPort = uri.getPort();
-                externalStatsPrefix = uri.getPath().substring(1);
+
+                if (!StringUtils.isEmpty(uri.getPath())) {
+                    externalStatsPrefix = uri.getPath().substring(1);
+                }
 
                 /* Append a dot (.) to the prefix if it is set */
                 if (externalStatsPrefix != null && !externalStatsPrefix.equals("")) {
```
jburwell: I realize it's not in the scope of this PR, but would you consider changing this `if` block to use `StringUtils.isEmpty`?

jburwell:
@wido this fix seems like it would be good for 4.8 and 4.9 as well.  Would you mind re-pointing this PR to the 4.8 branch?

Also, could you investigate the cause of the Jenkins failure?

server/src/com/cloud/server/StatsCollector.java
```diff
@@ -285,9 +286,15 @@ private void init(Map<String, String> configs) {
                     s_logger.info(scheme + " is not a valid protocol for external statistics. No statistics will be send.");
```
jburwell: I realize it is out of scope for this PR, but would mind changing the log level to ``WARN`` for this message?  It is clearly a user intent that is not going to be carried through due to bad configuration data.

wido:
@jburwell I've done that, that if now also uses StringUtils.

What is the easiest way to point this against 4.8? Changing the base on Github breaks everything and rebasing locally against 4.8 it does the same.

My git-foo isn't that good!

jburwell:
@wido sorry for the delayed response -- I lost track of this browser tab.  The quickest way may be to create a new branch from 4.8, cherry pick this commit into it, push the new branch, and open a new PR.

wido:
Done @jburwell, it's against the 4.8 branch now

jburwell:
@wido the Travis and Jenkins are failing due a compilation error.  It appears to be that `StringUtils` has not been imported into `StatsCollector`.

wido:
@jburwell Ah, that went wrong with the rebase. Fixed!

jburwell:
@wido thank you for re-targeting the PR.  Do you think running the smoke tests against one of the hypervisors is sufficient to verify this defect?  Should we update a Marvin test to cover it?

wido:
@jburwell You need to configure the management server with Graphite enabled with a wrongly entered URI to trigger this bug.

This is a manual error when somebody gives in a wrongly formed URL. The Graphite output was not broken by default.

rhtyd:
LGTM. @blueorangutan package

jburwell:
@blueorangutan test

rhtyd:
The test results have some intermittent failures around VPC VR related tests. The PR does not break anything, LGTM on both tests and code review.

wido:
This one has two LGTMs, I think we can merge it in?

jburwell:
@wido there are two errors related to template download in the blueorganutan run that we need to explain before merging this PR.  We are seeing a similar issue on #1587.  @abhinandanprateek and @murali-reddy are looking into the failures.  Hopefully, we will have an explanation/fix shortly.

murali-reddy:
@jburwell failures are redundant VPC related being tracked CLOUDSTACK-9541

wido:
Ok @jburwell , understood. I highly doubt they are related, but OK.

Just want to make sure this small fix makes it into our next point releases.

jburwell:
@wido we are seeing these errors on other PRs.  Based on the fact that they are occurring on other PRs and change does not impact these tests, I am LGTM for tests.

PR #1684, opened 2016-09-23
CLOUDSTACK-9489: the new config vars that are added do not goto DB if…

abhinandanprateek:
… values are set to NULL, removing this check so the entries in DB are made with NULL values

### DIFF
--- a/framework/config/src/org/apache/cloudstack/framework/config/dao/ConfigurationDaoImpl.java
+++ b/framework/config/src/org/apache/cloudstack/framework/config/dao/ConfigurationDaoImpl.java
@@ -90,7 +90,6 @@ public class ConfigurationDaoImpl extends GenericDaoBase<ConfigurationVO, String
                 configurations = listIncludingRemovedBy(sc);
 
                 for (ConfigurationVO config : configurations) {
-                    if (config.getValue() != null)
                         _configs.put(config.getName(), config.getValue());
                 }
             }


jburwell:
LGTM for code.

@pdion891 could you re-execute your upgrade scenario to verify that CLOUDSTACK-9489 is fixed?

jburwell:
@blueorangutan package

jburwell:
@abhinandanprateek this issue also affects 4.8.  Would it be possible to [change the base branch](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request) of this PR to 4.8?

jburwell:
@blueorganutan test matrix

jburwell:
@blueorangutan test matrix

pdion891:
I will test this PR, allow me few days..
This will be ported into master as well right? 

Thanks

jburwell:
@pdion891 yes, it will be merged to 4.8 and forward merged to 4.9 and master.

pdion891:
Unfortunately, Authentication is not working after an upgrade.  the configration table as been updated  but login is not working anymore.

Upgrade test from 4.4.4 to the PR (4.9), I can see that configuration table changes, origin confi count = 413, after starting 4.9 = 488

Here is the error display from jetty when trying to login into the UI or via cloudmonkey:

```
WARN  [o.a.c.s.a.PBKDF2UserAuthenticator] (qtp368929124-24:ctx-97af69ff) (logid:0d6ca986) The stored password for admin isn't in the right format for this authenticator
```

pdion891:
I've applied this fix to 4.7.0 and I got the upgrade from 4.4.4 working.  but when I upgrade 4.4.4 to this PR  I end with inability to logon into the UI.

pdion891:
upgrade from 4.7.0 to 4.9.0.1 not working with same error:
```
WARN  [o.a.c.s.a.PBKDF2UserAuthenticator] (qtp204585401-22:ctx-9e4c5615) (logid:1ca9f551) The stored password for admin isn't in the right format for this authenticator
```

jburwell:
@DagSonsteboSB @rhtyd this issue looks similar one we had with another upgrade.  Do you agree?  If so, do you recall what we did to fix it?

abhinandanprateek:
@pdion891 You need to look at user.authenticators.order and user.authenticators.exclude and see if they will provide same authenticator in chain as the one configured pre-upgrade.
Also check values of user.password.encoders.exclude and user.password.encoders.order for consistency. cc @jburwell @rhtyd

jburwell:
@abhinandanprateek do we need to add a note to the upgrade documentation and/or refine the upgrade script to protect against a change in `user.password.encoders.exclude` and `user.password.encoders.order`?

abhinandanprateek:
@jburwell yes we might have to further investigate this. I am waiting for info about the global params from @pdion891 to further this.
I think it will require a fix in upgrade scripts as the original authenticator should be preserved.

pdion891:
@abhinandanprateek  I did replace the following 4 settings by the original values. But after restarting jetty I'm unable to login, but this time the behavior is a bit different, it look like it doesn't try to login, no log or message in the UI.

```
update configuration set value=null where name='user.password.encoders.exclude';
update configuration set value='SHA256SALT,MD5,LDAP,PLAINTEXT' where name='user.authenticators.order';
update configuration set value=null where name='user.authenticators.exclude';
update configuration set value='SHA256SALT,MD5,LDAP,PLAINTEXT' where name='user.password.encoders.order';
```

Could it be possible the  authentication order is not followed at the login? if the password does not meet ``PBKDF2`` requirement, which is the first in the order list of 4.9.0, shouldn't try with the next encoder in the list and so on?

dagsonstebo:
@abhinandanprateek @jburwell @pdion891 - my twopence worth - I've come across issues with this on two uprade projects in the last 8-10 months. Both cases were CCP to ACS upgrades:

1) In the first case we hit an issue where the admin account could log in - but no users. We carried out a preliminary check of this, and the suspicion was on changes in the MD5 authenticator - but the client eventually chose to just reset all users passwords, hence no accurate RCA carried out.

2) In the second, more recent case the client wanted to use either PLAINTEXT or MD5 authenticators, hence we updated the authenticators order similar to @pdion891 did above. During troubleshooting of this trace level logging showed the authenticator order in ACS4.9 was respected, no issues found. RCA of the underlying issue concluded with it being a user configuration and usage issue rather than a bug.

pdion891:
@dagsonstebo, did you upgrade to 4.9 ?  So far, I had no authentication issue upgrading to 4.7.

dagsonstebo:
@pdion891 in my list above:

1) Was an upgrade CCP4.5.1 to ACS4.5.2 - and this was where we potentially saw a bug, but never proven.
2) Was an upgrade CCP3.0.7 to ACS4.9.0 - no issues found.

pdion891:
@dagsonstebo for your scenario 2 can you get the count of global settings ? 

via cloudmonkey as admin:
```
list configurations
```
You should have a count of about 468.

Also this current PR did apply some changes from the Config.java that were not applied in your upgrades. I have test upgrade path without this fix and I had no issues, but we end up with missing configurations entrees.

dagsonstebo:
@pdion891 - client system so limited access, not got cloudmonkey access but here's what's in the DB:

mysql> select name from configuration;
| name                                                        |
| account.cleanup.interval                                    |
| affinity.processors.exclude                                 |
| affinity.processors.order                                   |
| agent.lb.enabled                                            |
| agent.load.threshold                                        |
| alert.email.addresses                                       |
| alert.email.sender                                          |
| alert.purge.delay                                           |
| alert.purge.interval                                        |
| alert.smtp.connectiontimeout                                |
| alert.smtp.host                                             |
| alert.smtp.password                                         |
| alert.smtp.port                                             |
| alert.smtp.timeout                                          |
| alert.smtp.useAuth                                          |
| alert.smtp.username                                         |
| alert.wait                                                  |
| allow.public.user.templates                                 |
| allow.subdomain.network.access                              |
| allow.user.create.projects                                  |
| allow.user.expunge.recover.vm                               |
| allow.user.view.destroyed.vm                                |
| api.checkers.exclude                                        |
| api.throttling.cachesize                                    |
| api.throttling.enabled                                      |
| api.throttling.interval                                     |
| api.throttling.max                                          |
| apply.allocation.algorithm.to.pods                          |
| backup.snapshot.wait                                        |
| baremetal.ipmi.fail.retry                                   |
| baremetal.ipmi.lan.interface                                |
| blacklisted.routes                                          |
| capacity.check.period                                       |
| capacity.skipcounting.hours                                 |
| check.pod.cidrs                                             |
| check.txn.before.sending.agent.commands                     |
| cloud.dns.name                                              |
| cloud.identifier                                            |
| cloudbyte.management.apikey                                 |
| cloudbyte.management.ip                                     |
| cluster.cpu.allocated.capacity.disablethreshold             |
| cluster.cpu.allocated.capacity.notificationthreshold        |
| cluster.heartbeat.interval                                  |
| cluster.heartbeat.threshold                                 |
| cluster.localStorage.capacity.notificationthreshold         |
| cluster.memory.allocated.capacity.disablethreshold          |
| cluster.memory.allocated.capacity.notificationthreshold     |
| cluster.message.timeout.seconds                             |
| cluster.storage.allocated.capacity.notificationthreshold    |
| cluster.storage.capacity.notificationthreshold              |
| cluster.storage.operations.exclude                          |
| concurrent.snapshots.threshold.perhost                      |
| console.proxy.allocator.exclude                             |
| consoleproxy.capacity.standby                               |
| consoleproxy.capacityscan.interval                          |
| consoleproxy.cmd.port                                       |
| consoleproxy.disable.rpfilter                               |
| consoleproxy.launch.max                                     |
| consoleproxy.loadscan.interval                              |
| consoleproxy.management.state                               |
| consoleproxy.management.state.last                          |
| consoleproxy.restart                                        |
| consoleproxy.service.offering                               |
| consoleproxy.session.max                                    |
| consoleproxy.session.timeout                                |
| consoleproxy.url.domain                                     |
| control.cidr                                                |
| control.gateway                                             |
| copy.volume.wait                                            |
| cpu.overprovisioning.factor                                 |
| create.private.template.from.snapshot.wait                  |
| create.private.template.from.volume.wait                    |
| create.volume.from.snapshot.wait                            |
| custom.diskoffering.size.max                                |
| custom.diskoffering.size.min                                |
| data.motion.strategies.exclude                              |
| default.page.size                                           |
| default.ui.page.size                                        |
| deployment.planners.exclude                                 |
| deployment.planners.order                                   |
| detail.batch.query.size                                     |
| dhcp.providers.exclude                                      |
| direct.agent.load.size                                      |
| direct.agent.pool.size                                      |
| direct.agent.scan.interval                                  |
| direct.agent.thread.cap                                     |
| direct.attach.network.externalIpAllocator.enabled           |
| direct.attach.network.externalIpAllocator.url               |
| direct.network.no.default.route                             |
| direct.network.stats.interval                               |
| disable.extraction                                          |
| dynamic.apichecker.enabled                                  |
| eip.use.multiple.netscalers                                 |
| enable.baremetal.securitygroup.agent.echo                   |
| enable.dynamic.scale.vm                                     |
| enable.ec2.api                                              |
| enable.ha.storage.migration                                 |
| enable.s3.api                                               |
| enable.storage.migration                                    |
| enable.usage.server                                         |
| encode.api.response                                         |
| endpointe.url                                               |
| event.purge.delay                                           |
| event.purge.interval                                        |
| execute.in.sequence.hypervisor.commands                     |
| execute.in.sequence.network.element.commands                |
| expunge.delay                                               |
| expunge.interval                                            |
| expunge.workers                                             |
| external.baremetal.resource.classname                       |
| external.baremetal.system.url                               |
| external.firewall.default.capacity                          |
| external.lb.default.capacity                                |
| external.network.stats.interval                             |
| externaldhcp.vmip.max.retry                                 |
| externaldhcp.vmip.retrieval.interval                        |
| externaldhcp.vmipFetch.threadPool.max                       |
| extract.url.cleanup.interval                                |
| extract.url.expiration.interval                             |
| firewall.service.provider.exclude                           |
| globodns.domain.templateid                                  |
| globodns.override.entries                                   |
| gslb.service.provider.exclude                               |
| guest.domain.suffix                                         |
| guest.vlan.bits                                             |
| ha.fence.builders.exclude                                   |
| ha.investigators.exclude                                    |
| ha.investigators.order                                      |
| ha.planners.exclude                                         |
| ha.tag                                                      |
| ha.workers                                                  |
| healthcheck.update.interval                                 |
| host                                                        |
| host.allocators.exclude                                     |
| host.allocators.order                                       |
| host.capacityType.to.order.clusters                         |
| host.reservation.release.period                             |
| host.retry                                                  |
| host.stats.interval                                         |
| hyperv.guest.network.device                                 |
| hyperv.private.network.device                               |
| hyperv.public.network.device                                |
| hypervisor.gurus.exclude                                    |
| hypervisor.list                                             |
| implicit.host.tags                                          |
| incorrect.login.attempts.allowed                            |
| init                                                        |
| instance.name                                               |
| integration.api.port                                        |
| internallbvm.service.offering                               |
| interval.baremetal.securitygroup.agent.echo                 |
| investigate.retry.interval                                  |
| ip.deployers.exclude                                        |
| job.cancel.threshold.minutes                                |
| job.expire.minutes                                          |
| json.content.type                                           |
| kvm.guest.network.device                                    |
| kvm.private.network.device                                  |
| kvm.public.network.device                                   |
| kvm.snapshot.enabled                                        |
| kvm.ssh.to.agent                                            |
| ldap.basedn                                                 |
| ldap.bind.password                                          |
| ldap.bind.principal                                         |
| ldap.email.attribute                                        |
| ldap.firstname.attribute                                    |
| ldap.group.object                                           |
| ldap.group.user.uniquemember                                |
| ldap.lastname.attribute                                     |
| ldap.nested.groups.enable                                   |
| ldap.provider                                               |
| ldap.read.timeout                                           |
| ldap.request.page.size                                      |
| ldap.search.group.principle                                 |
| ldap.truststore                                             |
| ldap.truststore.password                                    |
| ldap.user.object                                            |
| ldap.username.attribute                                     |
| linkLocalIp.nums                                            |
| load.balancing.service.provider.exclude                     |
| management.network.cidr                                     |
| max.account.cpus                                            |
| max.account.memory                                          |
| max.account.networks                                        |
| max.account.primary.storage                                 |
| max.account.public.ips                                      |
| max.account.secondary.storage                               |
| max.account.snapshots                                       |
| max.account.templates                                       |
| max.account.user.vms                                        |
| max.account.volumes                                         |
| max.account.vpcs                                            |
| max.domain.cpus                                             |
| max.domain.memory                                           |
| max.domain.networks                                         |
| max.domain.primary.storage                                  |
| max.domain.public.ips                                       |
| max.domain.secondary.storage                                |
| max.domain.snapshots                                        |
| max.domain.templates                                        |
| max.domain.user.vms                                         |
| max.domain.volumes                                          |
| max.domain.vpcs                                             |
| max.project.cpus                                            |
| max.project.memory                                          |
| max.project.networks                                        |
| max.project.primary.storage                                 |
| max.project.public.ips                                      |
| max.project.secondary.storage                               |
| max.project.snapshots                                       |
| max.project.templates                                       |
| max.project.user.vms                                        |
| max.project.volumes                                         |
| max.project.vpcs                                            |
| max.template.iso.size                                       |
| mem.overprovisioning.factor                                 |
| mgt.server.vendor                                           |
| midonet.apiserver.address                                   |
| midonet.providerrouter.id                                   |
| migrate.retry.interval                                      |
| migratewait                                                 |
| minreq.sysvmtemplate.version                                |
| mount.parent                                                |
| network.acl.service.provider.exclude                        |
| network.dhcp.nondefaultnetwork.setgateway.guestos           |
| network.disable.rpfilter                                    |
| network.dns.basiczone.updates                               |
| network.elements.registry.exclude                           |
| network.gc.interval                                         |
| network.gc.wait                                             |
| network.guest.cidr.limit                                    |
| network.gurus.exclude                                       |
| network.ipv6.search.retry.max                               |
| network.loadbalancer.basiczone.elb.enabled                  |
| network.loadbalancer.basiczone.elb.gc.interval.minutes      |
| network.loadbalancer.basiczone.elb.network                  |
| network.loadbalancer.basiczone.elb.vm.cpu.mhz               |
| network.loadbalancer.basiczone.elb.vm.ram.size              |
| network.loadbalancer.basiczone.elb.vm.vcpu.num              |
| network.loadbalancer.haproxy.max.conn                       |
| network.loadbalancer.haproxy.stats.auth                     |
| network.loadbalancer.haproxy.stats.port                     |
| network.loadbalancer.haproxy.stats.uri                      |
| network.loadbalancer.haproxy.stats.visibility               |
| network.lock.timeout                                        |
| network.router.EnableServiceMonitoring                      |
| network.securitygroups.defaultadding                        |
| network.securitygroups.work.cleanup.interval                |
| network.securitygroups.work.lock.timeout                    |
| network.securitygroups.work.per.agent.queue.size            |
| network.securitygroups.workers.pool.size                    |
| network.throttling.rate                                     |
| nuagevsp.configure.dns                                      |
| nuagevsp.configure.gateway.systemid                         |
| nuagevsp.dns.external                                       |
| nuagevsp.isolatedntwk.domaintemplate.name                   |
| nuagevsp.sharedntwk.domaintemplate.name                     |
| nuagevsp.vpc.domaintemplate.name                            |
| outofbandmanagement.action.timeout                          |
| outofbandmanagement.ipmitool.interface                      |
| outofbandmanagement.ipmitool.path                           |
| outofbandmanagement.ipmitool.retries                        |
| outofbandmanagement.sync.interval                           |
| outofbandmanagement.sync.poolsize                           |
| ovm.guest.network.device                                    |
| ovm.private.network.device                                  |
| ovm.public.network.device                                   |
| ovm3.heartbeat.interval                                     |
| ovm3.heartbeat.timeout                                      |
| ping.interval                                               |
| ping.timeout                                                |
| pluggableApi.authenticators.exclude                         |
| pluggableApi.authenticators.order                           |
| pod.allocators.exclude                                      |
| pod.privateip.capacity.notificationthreshold                |
| pool.storage.allocated.capacity.disablethreshold            |
| pool.storage.capacity.disablethreshold                      |
| port                                                        |
| port.forwarding.service.provider.exclude                    |
| primary.storage.download.wait                               |
| project.email.sender                                        |
| project.invite.required                                     |
| project.invite.timeout                                      |
| project.smtp.host                                           |
| project.smtp.password                                       |
| project.smtp.port                                           |
| project.smtp.useAuth                                        |
| project.smtp.username                                       |
| query.selectors.exclude                                     |
| quota.currency.symbol                                       |
| quota.enable.enforcement                                    |
| quota.enable.service                                        |
| quota.statement.period                                      |
| quota.usage.smtp.connection.timeout                         |
| quota.usage.smtp.host                                       |
| quota.usage.smtp.password                                   |
| quota.usage.smtp.port                                       |
| quota.usage.smtp.sender                                     |
| quota.usage.smtp.useAuth                                    |
| quota.usage.smtp.user                                       |
| recreate.systemvm.enabled                                   |
| remote.access.vpn.client.iprange                            |
| remote.access.vpn.psk.length                                |
| remote.access.vpn.service.provider.exclude                  |
| remote.access.vpn.user.limit                                |
| resource.discoverers.exclude                                |
| resourcecount.check.interval                                |
| restart.retry.interval                                      |
| router.aggregation.command.each.timeout                     |
| router.alerts.check.interval                                |
| router.check.interval                                       |
| router.check.poolsize                                       |
| router.cpu.mhz                                              |
| router.extra.public.nics                                    |
| router.ram.size                                             |
| router.redundant.vrrp.interval                              |
| router.stats.interval                                       |
| router.template.hyperv                                      |
| router.template.kvm                                         |
| router.template.lxc                                         |
| router.template.ovm3                                        |
| router.template.vmware                                      |
| router.template.xenserver                                   |
| router.version.check                                        |
| s3.rrs.enabled                                              |
| s3.singleupload.max.size                                    |
| saml2.append.idpdomain                                      |
| saml2.default.idpid                                         |
| saml2.enabled                                               |
| saml2.idp.metadata.url                                      |
| saml2.redirect.url                                          |
| saml2.sigalg                                                |
| saml2.sp.contact.email                                      |
| saml2.sp.contact.person                                     |
| saml2.sp.id                                                 |
| saml2.sp.org.name                                           |
| saml2.sp.org.url                                            |
| saml2.sp.slo.url                                            |
| saml2.sp.sso.url                                            |
| saml2.timeout                                               |
| saml2.user.attribute                                        |
| scale.retry                                                 |
| sdn.ovs.controller                                          |
| sdn.ovs.controller.default.label                            |
| secondary.storage.vm                                        |
| secondary.storage.vm.allocators.exclude                     |
| secstorage.allowed.internal.sites                           |
| secstorage.capacity.standby                                 |
| secstorage.cmd.execution.time.max                           |
| secstorage.copy.password                                    |
| secstorage.encrypt.copy                                     |
| secstorage.proxy                                            |
| secstorage.service.offering                                 |
| secstorage.session.max                                      |
| secstorage.ssl.cert.domain                                  |
| secstorage.vm.mtu.size                                      |
| security.checkers.exclude                                   |
| security.checkers.order                                     |
| security.encryption.iv                                      |
| security.encryption.key                                     |
| security.hash.key                                           |
| security.singlesignon.key                                   |
| security.singlesignon.tolerance.millis                      |
| site2site.vpn.customergateway.subnets.limit                 |
| site2site.vpn.service.provider.exclude                      |
| site2site.vpn.vpngateway.connection.limit                   |
| snapshot.backup.rightafter                                  |
| snapshot.delta.max                                          |
| snapshot.max.daily                                          |
| snapshot.max.hourly                                         |
| snapshot.max.monthly                                        |
| snapshot.max.weekly                                         |
| snapshot.poll.interval                                      |
| snapshot.strategies.exclude                                 |
| sortkey.algorithm                                           |
| ssh.privatekey                                              |
| ssh.publickey                                               |
| ssl.keystore                                                |
| start.retry                                                 |
| static.nat.service.provider.exclude                         |
| stats.output.uri                                            |
| stop.retry.interval                                         |
| storage.cache.replacement.enabled                           |
| storage.cache.replacement.interval                          |
| storage.cache.replacement.lru.interval                      |
| storage.cleanup.delay                                       |
| storage.cleanup.enabled                                     |
| storage.cleanup.interval                                    |
| storage.max.volume.size                                     |
| storage.max.volume.upload.size                              |
| storage.overprovisioning.factor                             |
| storage.pool.allocators.exclude                             |
| storage.pool.allocators.order                               |
| storage.pool.max.waitseconds                                |
| storage.stats.interval                                      |
| storage.template.cleanup.enabled                            |
| sync.interval                                               |
| system.vm.auto.reserve.capacity                             |
| system.vm.default.hypervisor                                |
| system.vm.random.password                                   |
| system.vm.use.local.storage                                 |
| task.cleanup.retry.interval                                 |
| template.adapter.exclude                                    |
| template.preloader.pool.size                                |
| timeout.baremetal.securitygroup.agent.echo                  |
| total.retries                                               |
| traffic.sentinel.exclude.zones                              |
| traffic.sentinel.include.zones                              |
| ucs.sync.blade.interval                                     |
| update.wait                                                 |
| upload.monitoring.interval                                  |
| upload.operation.timeout                                    |
| upload.post.secret.key                                      |
| usage.aggregation.timezone                                  |
| usage.execution.timezone                                    |
| usage.sanity.check.interval                                 |
| usage.stats.job.aggregation.range                           |
| usage.stats.job.exec.time                                   |
| use.external.dns                                            |
| use.system.guest.vlans                                      |
| use.system.public.ips                                       |
| user.authenticators.exclude                                 |
| user.authenticators.order                                   |
| user.password.encoders.exclude                              |
| user.password.encoders.order                                |
| vm.allocation.algorithm                                     |
| vm.configdrive.label                                        |
| vm.deployment.planner                                       |
| vm.destroy.forcestop                                        |
| vm.disk.stats.interval                                      |
| vm.disk.throttling.bytes_read_rate                          |
| vm.disk.throttling.bytes_write_rate                         |
| vm.disk.throttling.iops_read_rate                           |
| vm.disk.throttling.iops_write_rate                          |
| vm.instancename.flag                                        |
| vm.job.check.interval                                       |
| vm.job.lock.timeout                                         |
| vm.job.report.interval                                      |
| vm.job.timeout                                              |
| vm.network.nic.max.secondary.ipaddresses                    |
| vm.network.throttling.rate                                  |
| vm.op.cancel.interval                                       |
| vm.op.cleanup.interval                                      |
| vm.op.cleanup.wait                                          |
| vm.op.lock.state.retry                                      |
| vm.op.wait.interval                                         |
| vm.password.length                                          |
| vm.stats.interval                                           |
| vm.tranisition.wait.interval                                |
| vm.user.dispersion.weight                                   |
| vmsnapshot.create.wait                                      |
| vmsnapshot.max                                              |
| vmSnapshot.strategies.exclude                               |
| vmware.additional.vnc.portrange.size                        |
| vmware.additional.vnc.portrange.start                       |
| vmware.create.full.clone                                    |
| vmware.hung.wokervm.timeout                                 |
| vmware.management.portgroup                                 |
| vmware.nested.virtualization                                |
| vmware.ports.per.dvportgroup                                |
| vmware.recycle.hung.wokervm                                 |
| vmware.reserve.cpu                                          |
| vmware.reserve.mem                                          |
| vmware.root.disk.controller                                 |
| vmware.service.console                                      |
| vmware.systemvm.nic.device.type                             |
| vmware.use.dvswitch                                         |
| vmware.use.nexus.vswitch                                    |
| vmware.vcenter.session.timeout                              |
| vpc.cleanup.interval                                        |
| vpc.max.networks                                            |
| vpc.providers.exclude                                       |
| wait                                                        |
| workers                                                     |
| xapiwait                                                    |
| xen.bond.storage.nics                                       |
| xen.create.pools.in.pod                                     |
| xen.guest.network.device                                    |
| xen.heartbeat.interval                                      |
| xen.heartbeat.timeout                                       |
| xen.private.network.device                                  |
| xen.public.network.device                                   |
| xen.pvdriver.version                                        |
| xen.setup.multipath                                         |
| xen.storage.network.device1                                 |
| xen.storage.network.device2                                 |
| xen.vm.vcpu.max                                             |
| xenserver.nics.max                                          |
| xenserver.pvdriver.version                                  |
| zone.directnetwork.publicip.capacity.notificationthreshold  |
| zone.secstorage.capacity.notificationthreshold              |
| zone.virtualnetwork.publicip.capacity.notificationthreshold |
| zone.vlan.capacity.notificationthreshold                    |
493 rows in set (0.00 sec)

abhinandanprateek:
@pdion891 I am unable to reproduce the issue. All the authenticators configured are tried.
I find it strange that there are no logs. Will it be possible to take a second look. I need either logs or some leads to move further on this.

pdion891:
@abhinandanprateek did you ran a fresh install of 4.9.x  or did an upgrade from example 4.4.4 ?

The way I'm testing is as follow, in case I do something wrong:

1) ``git checkout tags/4.4.4 -b 4.4.4``
2) Deploy and start management-server
```
mvn -DskipTests=true clean install -T 2
mvn -Pdeveloper -pl developer -Ddeploydb
mvn -pl client jetty:run
```
4) Auth test on the ui,  then stop jetty 
5) Change version to upgrade: ``git fetch origin pull/1684/head:pr-1684``
6) Build the new version but not deploy the DB to force the upgrade path:
```
mvn -DskipTests=true clean install -T 2
mvn -pl client jetty:run
```
7) test auth as admin in the UI and it's not working.

abhinandanprateek:
@pdion891 The root of this issue is that 4.9 uses dynamic api checker, which by default is not enabled. So even when the login succeed it does not go anywhere as the access to any api is denied.
After upgrade run this on cloud db:  update cloud.configuration set value='true' where name='dynamic.apichecker.enabled';

This will fix the login issue.
I am sure this is there in upgrade documentation, @rhtyd will be the best person to comment if it can be included in upgrade scripts. cc @jburwell

rhtyd:
@abhinandanprateek 4.9+ does not contain commands.properties so which is why dynamic roles is enabled for them (by default). When upgrading, if you or @pdion891 upgraded using maven/git code and not packages you won't have commands.properties file and login won't work if dynamic roles is not enabled, but for existing deployments login should work who will have commands.properties in their classpath.

jburwell:
@rhtyd this issue feels like it should no occur.  For example, could the role permission tables be populated with default values when they are created?  If/when values are imported from an existing `command.properties`, the import tool could clear the tables as the first step in the process.

pdion891:
Thanks @rhtyd, @jburwell @abhinandanprateek for the clarifications.

Look like doing the ``update cloud.configuration set value='true' where name='dynamic.apichecker.enabled';`` fixe the authentication issue.

Should we expect another issue post upgrade because of this db change? should we roll it back later? I haven't found the file command.properties when running 4.9.1.0-SNAPSHOT via jetty in the repo.

also, in the context a user did customized some values, will they be reset?

Thanks!

abhinandanprateek:
@pdion891 The issue regarding dynamic apichecker will happen even without this patch. In 4.9 there is no command.properties and it has been moved to apichecker schema.

 All this patch does is write the configuration that have 'null' values to db. As such it is not going to reset any of the existing values during an upgrade. cc @rhtyd @jburwell

pdion891:
Thanks @abhinandanprateek.  So I guest it's an LGTM for this PR but it's not solving the whole issue for CLOUDSTACK-9489 which is configuration inconsistency between fresh install and upgrade. So to me this PR cannot close CLOUDSTACK-9489 unless we create a new jira issue for the auth issue after an upgrade where we would need to manually change ``dynamic.apichecker.enabled``.

jburwell:
@pdion891 based on the behavior reported, I believe we can make this experience much smoother if not completely consistent.  Do you have time to grab a quick GTM to discuss tomorrow (29 Sept 2016)?  I will report the results of the call back to this PR (and by extension dev@).

rhtyd:
@pdion891 @jburwell as I've already explained, there should be no need to enable dynamic apichecker if you're upgrading from an older system that has a commands.properties file. I think the upgrade tests was performed not using packages but maven/jetty. The current 4.9+ branch don't have commands.properties so when you rebuild mgmt server you don't get the commands.properties in the classpath. This has been documented in the admin docs as well. If you've installed 4.9+ based mgmt server on a new server, the old commands.properties file need to be copied to the new server to continue to use the static checker, otherwise the dynamic checker need to be enabled and migrated using a migration script as documented here: http://docs.cloudstack.apache.org/projects/cloudstack-administration/en/4.9/accounts.html#using-dynamic-roles

jburwell:
@rhtyd discussing this issue with @paul, there is a scenario where this wedged condition can occur.  If you perform an upgrade by creating a new management server VM then no commands.properties file will be present and the dynamic API checker will be set to off.  This upgrade approach is taken as a best practice to provide a simple, reliable rollback strategy and an opportunity to perform OS upgrades.  I think the solution in this situation is combination of improved upgrade documentation (i.e. copy over your commands.properties or run the import script and turn on the dynamic role API checker) and some better error messages that will help admins understand how to remedy the problem.

rhtyd:
@jburwell when a fresh mgmt server is deployed for upgrading an existing CloudStack environment, several files from `/etc/cloudstack/management` are copied to the new server, especially the `db.properties` file. The `commands.properties` is no different, in the past all such deployment (where in-place upgrades were avoided) copied all these config files in /etc/cloudstack/management from old server to new server.

Official docs only talk about in-place upgrades, our packaging system (post-install/upgrade scripts) assume in-place upgrade as well; if someone's doing fresh mgmt server for upgrade, they know what they are doing and its out-of-scope of what the project can support. At the same time, we may add documentation for such kind of upgrades as well. The solution for this problem is documentation and release/upgrade notes.

PaulAngus:
That isnt correct @bhaisaab.  commands.properties can be different on every mgmt server.  Cloudstack would not know which version to use.



________________________________

rhtyd:
@PaulAngus but it's up to the admin and how they want to upgrade/setup a new mgmt server. Assuming you want to create a new mgmt server for each old mgmt server, one strategy could be to copy the commands.properties from each old mgmt server to each such new mgmt server.

jburwell:
@rhtyd we have a case where two experienced users (and likely more to come) have gotten tripped up by this case.  Therefore, we need to address it.  As I mentioned previously, the solution to this issue may be improved documentation.  In discussing it further, we may also find a technical solution as well.

pdion891:
@rhtyd I agree with you for the documentation part.

For sure in-place upgrade must be transparent to user and not required manual steps.

pdion891:
I've re-perform my upgrade test this morning with this PR, I just add to change ``dynamic.apichecker.enabled`` value have the authentication working, so nothing else have to change.

As @PaulAngus raised, in an upgrade scenario where the management-server is build for scratch to perform a cloud upgrade, ``dynamic.apichecker.enabled`` will cause issues and headache.

So to me it would make more sense if during and upgrade the value of ``dynamic.apichecker.enabled`` is automatically set to "true" which is already the value of a new deployment, then we should update the release notes so if users have customized the ``commands.properties`` file, it's up to them to rollback ``dynamic.apichecker.enabled`` to false and make sure they have a copy of the commands.properties file. what do you think @PaulAngus ?

Maybe the ideal scenario would be that if the ``commands.properties`` file exist, cloudstack during is upgrade phase when it's started would update is dynamic apichecker values to those defined in the commands.properties 

Does this should go in this PR or we consider this PR is ok?

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 133
 Hypervisor xenserver
 NetworkType Advanced
 Passed=98
 Failed=5
 Skipped=6

**Failed tests:**
* test_non_contigiousvlan.py

 * test_extendPhysicalNetworkVlan Failed

* test_deploy_vm_iso.py

 * test_deploy_vm_from_iso Failing since 17 runs

* test_list_ids_parameter.py

 * ContextSuite context=TestListIdsParams>:setup Failing since 6 runs

* test_vm_life_cycle.py

 * test_10_attachAndDetach_iso Failing since 18 runs

* test_routers_network_ops.py

 * test_01_isolate_network_FW_PF_default_routes_egress_true Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_network.py
test_router_dns.py
test_login.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_disk_offerings.py

jburwell:
@pdion891 I don't care for the idea of potentially ignoring a custom security configuration as it may unexpectedly reduce a user's security.   By default, we should take the most conservative path when setting security defaults.

/cc @PaulAngus

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
LGTM. Failures in tests around RVR/VPC are known intermittent failures.

pdion891:
I haven't retested recently but the upgrade path will need documentation update I think because of the ``commands.properties``.

Otherwise it's LGTM for me too.

rhtyd:
Thanks @pdion891 we'll open a doc PR shortly. Based on the discussion, review and tests, I'll proceed with the merge now.

pdion891:
btw, I don't think this PR should be backport since it would change behavior for someone that would upgrade from 4.x.x to 4.9.0 or 4.9.x(that would include this PR)

PR #1686, opened 2016-09-23
CLOUDSTACK-9507: ListVM response's guest_os id should be a UUID

rhtyd:
The listVirtualMachines API response contains a guest os_type field which
is an integer, and therefore not useful or easily consumable. This fixes and
migrates the returned type from an id (integer) to a uuid (string). It makes
the response consistent with other API responses such as those of listTemplates
and allow end users to consume this uuid.

/cc @jburwell @karuturi -- this is a minor change. It may be debatable if this break api response compatibility, please advise.

@blueorangutan package

### DIFF
--- a/api/src/org/apache/cloudstack/api/response/UserVmResponse.java
+++ b/api/src/org/apache/cloudstack/api/response/UserVmResponse.java
@@ -284,7 +284,7 @@ public class UserVmResponse extends BaseResponseWithTagInformation implements Co
 
     @SerializedName(ApiConstants.OS_TYPE_ID)
     @Param(description = "OS type id of the vm", since = "4.4")
-    private Long osTypeId;
+    private String osTypeId;
 
     public UserVmResponse() {
         securityGroupList = new LinkedHashSet<SecurityGroupResponse>();
@@ -810,7 +810,7 @@ public class UserVmResponse extends BaseResponseWithTagInformation implements Co
         this.details = details;
     }
 
-    public void setOsTypeId(Long osTypeId) {
+    public void setOsTypeId(String osTypeId) {
         this.osTypeId = osTypeId;
     }
 }
--- a/server/src/com/cloud/api/query/dao/UserVmJoinDaoImpl.java
+++ b/server/src/com/cloud/api/query/dao/UserVmJoinDaoImpl.java
@@ -191,7 +191,7 @@ public class UserVmJoinDaoImpl extends GenericDaoBaseWithTagInformation<UserVmJo
         userVmResponse.setPublicIpId(userVm.getPublicIpUuid());
         userVmResponse.setPublicIp(userVm.getPublicIpAddress());
         userVmResponse.setKeyPairName(userVm.getKeypairName());
-        userVmResponse.setOsTypeId(userVm.getGuestOsId());
+        userVmResponse.setOsTypeId(userVm.getGuestOsUuid());
 
         if (details.contains(VMDetails.all) || details.contains(VMDetails.stats)) {
             // stats calculation


serg38:
@rhtyd Somehow it works fine without this patch
curl "http://localhost:8096//client/api?command=listVirtualMachines&listall=true&response=json"
{"listvirtualmachinesresponse":{"count":1,"virtualmachine":[{"id":"beb85d73-17cd-4de5-8c35-61f70d88e5e1","name":"TEST-3DGPU-NICO","displayname":"TEST-3DGPU-NICO","account":"admin","userid":"d557133a-3fcc-11e5-9c45-005056ad45b7","username":"admin","domainid":"5a7ffa07-3fca-11e5-9c45-005056ad45b7","domain":"ROOT","created":"2016-08-30T16:21:45-0700","state":"Stopped","haenable":false,"zoneid":"0d074f25-ed31-482f-8bc5-44c9314fc417","zonename":"ECS-NSX","templateid":"d6584e60-d500-40cd-9a33-1700a59c464c","templatename":"VMXNET3 template-287NCX","templatedisplaytext":"VMXNET3 Template","passwordenabled":false,"serviceofferingid":"0b9dab64-60ec-4c19-8a36-8725660d37d1","serviceofferingname":"Tiny Instance","cpunumber":1,"cpuspeed":100,"memory":128,"**## guestosid":"5aadaacc-3fca-11e5-9c45-005056ad45b7",**"rootdeviceid":0,"rootdevicetype":"ROOT","securitygroup":[],"nic":[{"id":"08e0b3fe-219f-46bb-a766-70717a15a7fd","networkid":"c05a5c09-eb29-49bc-9cfd-705becc7a909","networkname":"VLAN13","netmask":"255.255.255.0","gateway":"10.140.13.1","ipaddress":"10.140.13.81","isolationuri":"vlan://13","broadcasturi":"vlan://13","traffictype":"Guest","type":"Shared","isdefault":true,"macaddress":"06:0a:d0:00:36:d9","secondaryip":[]}],"hypervisor":"VMware","instancename":"i-2-9253-VM","affinitygroup":[],"displayvm":true,"isdynamicallyscalable":false,"ostypeid":12,"tags":[]}]}}

serg38:
@rhtyd I think you meant guest_os_type not guest_os_id. In this case the issue is not only that it returns ID instead of UUID but also that it doesn't return ID of os type. In the sample output ostypeid":12 is not correct OS type which is supposed to be 1 ( CentOs). The right response seems to be
userVmResponse.setOsTypeId(userVm.getGuestType().toString());
vs
userVmResponse.setOsTypeId(userVm.getGuestOsUuid());

rhtyd:
@serg38 I mean the os_type_id, which was previously returned as an integer but should be returned as a uuid. The uuid would be consumeable by the listGuestOsMapping API.

serg38:
@rhytd Current implementation has a bug as well that returns GuestOsId in place of GuestOsTypeId. I am for not only switching to UUID but also for fixing it and returning correct output.

jburwell:
@rhtyd @serg38 while I agree that this change makes logical sense, it breaks the API compatibility which we [guarantee](https://cwiki.apache.org/confluence/display/CLOUDSTACK/Releases#Releases-Versioning) for major releases.   5.0.0 is [scheduled](https://cwiki.apache.org/confluence/display/CLOUDSTACK/%5BPROPOSAL%5D+2016-2017+Release+Cycle+and+Calendar) to open on 5 December 2016.  I suggest that we hold this PR until that time when we will be accepting changes that break API backwards compatibility.

rhtyd:
Instead of changing the response type, another way is additive -- to introduce a new key (guestosuuid) in the response that returns the uuid for the expected key.

serg38:
@rhtyd This clearly a bug. os_type_id was always supposed to be a OS type as per: 

@SerializedName(ApiConstants.OS_TYPE_ID)
    @Param(description = "OS type id of the vm", since = "4.4")
    private String osTypeId;

Currently it returns an OS Id. Guest OS UUID is already part of response as guestosid . 

Basically there are 3 options:
1. Fix current bug and correctly return OS type ID as os_type_id
2. Fix current bug but return OS type UUID as os_type_id
3. Fix current bug  and correctly return OS type ID as os_type_id plus add a new response os_type_uuid  that would be returning UUID of the guest Os category. 

I think #3 is overkill. I am  for #2.

rhtyd:
@blueorangutan package

serg38:
I am -1  in the current form

rhtyd:
This PR will be on hold until next set of 4.8.x, 4.9.x, 4.10.x releases.

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 442
 Hypervisor xenserver
 NetworkType Advanced
 Passed=104
 Failed=1
 Skipped=7

**Failed tests:**
* test_routers_network_ops.py

 * test_01_RVR_Network_FW_PF_SSH_default_routes_egress_true Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

rhtyd:
Break api response, closing for now

PR #1687, opened 2016-09-23
Marvintests vpcofferingfix

PaulAngus:
It appears that original code did not expect there to be an alternate default VPC offering (Nuage) .
VPC selection uses isdefault=True which selects the Nuage Network offering, stopping the test from being run.
Fix was to look for name="Default VPC Offering" instead.

Replaces PR #1652  (same content, differenet branch)


jburwell:
@PaulAngus it looks like a bunch of commits got picked up inadvertently triggering a large number of unrelated changes.  Ping me, and I will help reduce down to the one expected change.

PR #1694, opened 2016-09-27
CLOUDSTACK-9509: Host Connects Without Storage

rhtyd:
KVM hosts on shared storage failure was accepted by mgmt server with the
host state as Up, even though there was no primary/shared storage available on
it. This patch offers a quick fix by throwing an exception in the storage monitor
which connects storage pool on host. The failure is trapped by agent manager
that disconnects the agent without any investigation.

Based on Lab tests, KVM agent may take upto 2 minutes to attempt NFS mount when
the storage is inaccessible (firewalled, or shutdown) before returning back with
an error. It is safe to assume that this won't add pressure on mgmt server due to
several reconnection attempts, and KVM agent would retry reconnection every 2
minutes.

For such KVM hosts, where failure happens due to storage issues; they will be
briefly put in Alert state but will be mostly be in Connecting state during which
the KVM host attempts to mount/reconfigure NFS storage pool.

/cc @jburwell @karuturi 
@blueorangutan package

### DIFF
--- a/server/src/com/cloud/storage/listener/StoragePoolMonitor.java
+++ b/server/src/com/cloud/storage/listener/StoragePoolMonitor.java
@@ -99,12 +99,14 @@ public class StoragePoolMonitor implements Listener {
                     }
 
                     Long hostId = host.getId();
-                    s_logger.debug("Host " + hostId + " connected, sending down storage pool information ...");
+                    if (s_logger.isDebugEnabled()) {
+                        s_logger.debug("Host " + hostId + " connected, connecting host to shared pool id " + pool.getId() + " and sending storage pool information ...");
+                    }
                     try {
                         _storageManager.connectHostToSharedPool(hostId, pool.getId());
                         _storageManager.createCapacityEntry(pool.getId());
                     } catch (Exception e) {
-                        s_logger.warn("Unable to connect host " + hostId + " to pool " + pool + " due to " + e.toString(), e);
+                        throw new ConnectionException(true, "Unable to connect host " + hostId + " to storage pool id " + pool.getId() + " due to " + e.toString(), e);
                     }
                 }
             }
--- a/server/test/com/cloud/storage/listener/StoragePoolMonitorTest.java
+++ b/server/test/com/cloud/storage/listener/StoragePoolMonitorTest.java
@@ -1,0 +1,80 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+package com.cloud.storage.listener;
+
+import com.cloud.agent.api.StartupRoutingCommand;
+import com.cloud.exception.ConnectionException;
+import com.cloud.exception.StorageUnavailableException;
+import com.cloud.host.HostVO;
+import com.cloud.hypervisor.Hypervisor;
+import com.cloud.storage.ScopeType;
+import com.cloud.storage.StorageManagerImpl;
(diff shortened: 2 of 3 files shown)


rhtyd:
@blueorangutan test centos6 kvm-centos7

server/test/com/cloud/storage/listener/StoragePoolMonitorTest.java
```diff
@@ -0,0 +1,80 @@
...
+    private StoragePoolVO pool;
+    private StartupRoutingCommand cmd;
+
+    @Before
+    public void setUp() throws Exception {
+        storageManager = Mockito.mock(StorageManagerImpl.class);
+        poolDao = Mockito.mock(PrimaryDataStoreDao.class);
+
+        storagePoolMonitor = new StoragePoolMonitor(storageManager, poolDao);
+        host = new HostVO("some-uuid");
```
jburwell: Why not pass in `UUID.randomUUID().toString()`?

server/src/com/cloud/storage/listener/StoragePoolMonitor.java
```diff
@@ -99,12 +99,12 @@ public void processConnect(Host host, StartupCommand cmd, boolean forRebalance)
                     }
 
                     Long hostId = host.getId();
-                    s_logger.debug("Host " + hostId + " connected, sending down storage pool information ...");
+                    s_logger.debug("Host " + hostId + " connected, connecting host to shared pool id " + pool.getId() + " and sending storage pool information ...");
```
jburwell: Please wrap in a `if (s_logger.isDebugEnabled)` to check to avoid unnecessary string concatenation when the `DEBUG` level is disabled.

server/test/com/cloud/storage/listener/StoragePoolMonitorTest.java
rhtyd: This a unit test, uniqueness of the VO object is not necessary. We just want to create a hostVO object for testing that could have any uuid.

server/src/com/cloud/storage/listener/StoragePoolMonitor.java
rhtyd: Fixed.

rhtyd:
The previous test run was incorrect, several tests were run from master the features that don't exist on 4.8.
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

abhinandanprateek:
@rhtyd LGTM on code review

rhtyd:
@murali-reddy @jburwell can we have lgtm/review on this, thanks.

rhtyd:
Merging this based on tests results and review, the failures caused in VPC/VR related tests are known intermittent issues.

nathanejohnson:
@rhtyd It appears that rolling this forward to 4.9 et al has broken builds.  The signature of the CreateStoragePool constructor changed between 4.8 and 4.9.

https://github.com/apache/cloudstack/blob/4.8/server/src/com/cloud/storage/listener/StoragePoolMonitor.java#L52

That's the 4.8 version

https://github.com/apache/cloudstack/blob/4.9/server/src/com/cloud/storage/listener/StoragePoolMonitor.java#L57

That's the 4.9 version, which takes 3 parameters.  When building I get the following error:

[ERROR] <snip> /server/test/com/cloud/storage/listener/StoragePoolMonitorTest.java:[49,29] error: constructor StoragePoolMonitor in class StoragePoolMonitor cannot be applied to given types;

rhtyd:
Thanks for sharing @nathanejohnson in case you've already fixed it, can you send a PR? Otherwise, it's late for me, I'll have a look first in the morning.

nathanejohnson:
@rhtyd I have an issue with circular dependencies trying to make the test work with the new signature.  In order to initialize the StoragePoolMonitor, I need a DataStoreManagerImpl type, and this is defined in cloud-engine-storage .  I cannot add cloud-engine-storage as a maven dependency of cloud-server because this in turn depends on cloud-secondary-storage , which in turn depends on cloud-server.  I'm not sure what the best path forward is, but unfortunately I don't think I can get this resolved today (no pun intended).

rhtyd:
@nathanejohnson I'm not sure what you are doing, the build failure was due to a unit test introduced in this PR. In 4.9+ branches, the unit test did not consider the additional parameter added to the constructor signature in 4.9+ branches. The simplest fix was to simply pass null, as the unit test does not depend/use that additional parameter. Thanks for bringing this to my attention, the build is fixed on 4.9, master branches now.

nathanejohnson:
@rthyd ahh, it didn't occur to me to just pass null.  I was trying to initialize that third parameter with a mock (like the other two).

Thanks again!

PR #1701, opened 2016-10-05
CLOUDSTACK-9534: Allow VR to be destroy in UI when in running state

rhtyd:
This allows us to destroy a VR that is in running state from the UI. This action is possible via cloudmonkey/APIs, and would be useful for anyone who wants to simply destroy a VR from the UI without having to stop it first.

As this is a pure UI fix, manual testing would be needed but no regression integration testing is necessary.
/cc @jburwell @karuturi 

@blueorangutan package

Screenshot:

screenshot from 2016-10-05 18-25-49

### DIFF
--- a/ui/scripts/system.js
+++ b/ui/scripts/system.js
@@ -2585,6 +2585,11 @@
                                                 label: 'label.destroy.router',
                                                 messages: {
                                                     confirm: function (args) {
+                                                        if (args && args.context && args.context.routers[0]) {
+                                                            if (args.context.routers[0].state == 'Running') {
+                                                                return dictionary['message.action.stop.router'] + ' ' + dictionary['message.confirm.destroy.router'];
+                                                            }
+                                                        }
                                                         return 'message.confirm.destroy.router';
                                                     },
                                                     notification: function (args) {
@@ -3752,6 +3757,11 @@
                                                 label: 'label.destroy.router',
                                                 messages: {
                                                     confirm: function (args) {
+                                                        if (args && args.context && args.context.routers[0]) {
+                                                            if (args.context.routers[0].state == 'Running') {
+                                                                return dictionary['message.action.stop.router'] + ' ' + dictionary['message.confirm.destroy.router'];
+                                                            }
+                                                        }
                                                         return 'message.confirm.destroy.router';
                                                     },
                                                     notification: function (args) {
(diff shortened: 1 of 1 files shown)


rhtyd:
Ignore centos6 failure, it succeeded but something went wrong at the backend.

koushik-das:
@rhtyd Will destroy router show a similar message as stop router in UI?
All services provided by this virtual router will be interrupted. Please confirm that you want to stop this router.

rhtyd:
@koushik-das it'll display whatever message is currently displayed for destroy router action, which is not same as stop router action:

screenshot from 2016-10-07 14-12-10

koushik-das:
Since services are disrupted when VR is destroyed, I think it would be useful to provide a message similar to stop VR. Earlier only destroying a stopped VR was possible from UI so caller was already aware.

rhtyd:
@koushik-das this is available/shown only to the admin, can we assume admin would know that it may cause network disruption for guest vms? The translation can be fixed, though in case the VR is stopped -- the message won't make sense when destroy router action is performed.

koushik-das:
@rhtyd You are right, the message only makes sense for destroying running VR. So some conditional check needs to be there for displaying the proper message.

Based on the discussion it appears that may be destroy option was intentionally not shown earlier to avoid the effort of showing conditional messages :)

rhtyd:
@koushik-das I'll see if I can add a conditional in the pop-up :)

rhtyd:
@koushik-das alright I've fixed the message. It shows as follows:

1. When router is in running state:
screenshot from 2016-10-07 15-11-40

2. When router is in stopped state:
screenshot from 2016-10-07 15-17-25

koushik-das:
LGTM. Small comment on the message for running VR. "Please confirm... " is appearing twice, only one would suffice.

rhtyd:
@koushik-das I actually have appended the two translated messages in the conditional, as destroy router would stop the VR first.

rhtyd:
Is this good to merge, purely UI change? /cc @murali-reddy @abhinandanprateek @karuturi

abhinandanprateek:
@rhtyd LGTM on code review...add more such admin UI stuff that admin end up using manually

rhtyd:
Since this is a pure UI change, merging based on manual test LGTMs and code review.

PR #1705, opened 2016-10-11
Made the changes to improve logging.

nemo9cby:
CLOUSTACK-9465 Several log refactoring/improvement suggestions.

There are two scenarios of logging which needs refactoring/improvement: 

Method invocation replaced by variable

This means that in the logging code, the method invocation is pre-defined as a variable. for simplicity,          the method invocation should be replaced by the variable.

Delete variable which must be null

The variable in the logging code is null, there is no need to put the variable there.


rhtyd:
LGTM.
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
All changes are in logging, regression tests won't catch anything, but I'll fire a round.
@blueorangutan test

rhtyd:
Merging based on code review and test LGTM. The one failing test is intermittent and caused by the environment.

PR #1737, opened 2016-10-27
CLOUDSTACK-9561 After domain/account deletion, snapshot taken by the domain/account remains undeleted

SudharmaJain:
While deleting the UserAccount Cleanup for the removed VMs/volumes are not happening. For the removed VMs, snapshots doesn't get cleaned. Only for running VMs(volumes in ready state) the cleanup happens. 

When the VM is desroyed, the volume is marked as destroyed and later storage garbage collector perform the cleanup. But if we try delete domain/account before storage garbage collector runs, then it fails.


rhtyd:
@blueorangutan package

jburwell:
@SudharmaJain this fix seems like it would be good for LTS users as well.  Could you please change the base branch to 4.9?

rhtyd:
@blueorangutan package

rhtyd:
LGTM. @blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Tests LGTM, some failures are known intermittent issues.

PR #1738, opened 2016-10-27
CLOUDSTACK-9566 instance-id metadata for baremetal VM returns ID

SudharmaJain:
There is difference in instance-id metadata across baremetal and other hypervisors.  

On Baremetal
[root@ip-172-17-0-144 ~]# curl http://8.37.203.221/latest/meta-data/instance-id
6021

on Xen
[root@ip-172-17-2-103 ~]# curl http://172.17.0.252/latest/meta-data/instance-id
cbeb517a-e833-4a0c-b1e8-9ed70200fbbf

In both cases it should be vm's uuid.


rhtyd:
@blueorangutan package

rhtyd:
The changes are in baremetal subsystem, which cannot be tested by current test environments.
The changes LGTM, @karuturi @koushik-das can you review/lgtm on this and confirm any test results from your env?

koushik-das:
Code changes LGTM.

serg38:
LGTM as well

rhtyd:
Thanks all, I'll proceed with the merging as build, travis and packaging works, without an explicit Trillian test or BVT as both don't run tests against baremetal and the change is a single line change in the baremetal hypervisor plugin.

PR #1711, opened 2016-10-17
XenServer 7 Support

syed:
This PR adds support for XenServer 7. I have manually done the following tests
- Create a new cluster with XenServer7
- Add Primary storage: Should create an SR on XS7
- Add another XS7 host to the Pool 
- Add host2 to Cloudstack
- Create VM1 from template 
- Create VM2 from template 
- Ping/SSH VM1 to VM2 and vice-versa
- Stop/Delete/Expunge VM2
- Create Data disk 
- Attach it to VM1
- Create VM snaphsot of VM1
- Restore VM snapshot of VM1
- Delete VM snapshot of VM1
- Create Volume snapshot of Datadisk
- Create volume snapshot of Root disk
- Create new template from snapshot of root disk
- Create volume from snapshot of datadisk
- Detach datadisk volume
- Delete datadisk volume
- Aquire a public IP
- Create a static nat to VM1
- Live migrate VM1 while traffic on VM
- Delete VM1

PR #1739, opened 2016-10-27
Fix xapi pluginlib for XenServer 7

ArmedGuy:
This makes cloudstack_pluginlib.py compatible with XenServer 7 dom0, as (from what I gathered) simplejson was included by default as "json" in the version that XenServer 7 dom0 uses. (python --version returns 2.7.5).
This is the only issue I have found so far with this setup.


karuturi:
PR #1711 has this and bunch of other changes needed for XS7 support.

ArmedGuy:
Ah, my bad then. Thought I had searched for it.

PR #1740, opened 2016-10-27
CLOUDSTACK-9572 Snapshot on primary storage not cleaned up after Stor…

yvsubhash:
Snapshot on primary storage not cleaned up after Storage migration. This happens in the following scenario
## Steps To Reproduce
1. Create an instance on the local storage on any host
2. Create a scheduled snapshot of the volume:
3. Wait until ACS created the snapshot. ACS is creating a snapshot on local storage and is transferring this snapshot to secondary storage. But the latest snapshot on local storage will stay there. This is as expected.
4. Migrate the instance to another XenServer host with ACS UI and Storage Live Migration
5. The Snapshot on the old host on local storage will not be cleaned up and is staying on local storage. So local storage will fill up with unneeded snapshots.

### DIFF
--- a/engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/SnapshotDataFactoryImpl.java
+++ b/engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/SnapshotDataFactoryImpl.java
@@ -65,6 +65,24 @@ public class SnapshotDataFactoryImpl implements SnapshotDataFactory {
     }
 
     @Override
+    public List<SnapshotInfo> getSnapshots(long volumeId, DataStoreRole role) {
+
+        SnapshotDataStoreVO snapshotStore = snapshotStoreDao.findByVolume(volumeId, role);
+        if (snapshotStore == null) {
+            return new ArrayList<>();
+        }
+        DataStore store = storeMgr.getDataStore(snapshotStore.getDataStoreId(), role);
+        List<SnapshotVO> volSnapShots = snapshotDao.listByVolumeId(volumeId);
+        List<SnapshotInfo> infos = new ArrayList<>();
+        for(SnapshotVO snapshot: volSnapShots) {
+            SnapshotObject info = SnapshotObject.getSnapshotObject(snapshot, store);
+            infos.add(info);
+        }
+        return infos;
+    }
+
+
+    @Override
     public SnapshotInfo getSnapshot(long snapshotId, DataStoreRole role) {
         SnapshotVO snapshot = snapshotDao.findById(snapshotId);
         SnapshotDataStoreVO snapshotStore = snapshotStoreDao.findBySnapshot(snapshotId, role);
--- a/server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
+++ b/server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
@@ -1332,6 +1332,21 @@ public class SnapshotManagerImpl extends MutualExclusiveIdsManagerBase implement
     }
 
     @Override
+    public void cleanupSnapshotsByVolume(Long volumeId) {
+        List<SnapshotInfo> infos = snapshotFactory.getSnapshots(volumeId, DataStoreRole.Primary);
+        for(SnapshotInfo info: infos) {
+            try {
+               if(info != null) {
+                   snapshotSrv.deleteSnapshot(info);
+               }
+            } catch(CloudRuntimeException e) {
+                String msg = "Cleanup of Snapshot with uuid " + info.getUuid() + " in primary storage is failed. Ignoring";
+                s_logger.warn(msg);
+            }
+        }
+    }
+
+    @Override
     public Snapshot allocSnapshot(Long volumeId, Long policyId, String snapshotName, Snapshot.LocationType locationType) throws ResourceAllocationException {
         Account caller = CallContext.current().getCallingAccount();
         VolumeInfo volume = volFactory.getVolume(volumeId);
--- a/engine/api/src/org/apache/cloudstack/engine/subsystem/api/storage/SnapshotDataFactory.java
+++ b/engine/api/src/org/apache/cloudstack/engine/subsystem/api/storage/SnapshotDataFactory.java
@@ -29,6 +29,8 @@ public interface SnapshotDataFactory {
 
     SnapshotInfo getSnapshot(long snapshotId, DataStoreRole role);
 
+    List<SnapshotInfo> getSnapshots(long volumeId, DataStoreRole store);
+
     List<SnapshotInfo> listSnapshotOnCache(long snapshotId);
 
     SnapshotInfo getReadySnapshotOnCache(long snapshotId);
--- a/engine/storage/volume/src/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
+++ b/engine/storage/volume/src/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
@@ -1543,6 +1543,7 @@ public class VolumeServiceImpl implements VolumeService {
                 future.complete(res);
             } else {
                 srcVolume.processEvent(Event.OperationSuccessed);
+                snapshotMgr.cleanupSnapshotsByVolume(srcVolume.getId());
                 future.complete(res);
             }
         } catch (Exception e) {
@@ -1624,6 +1625,7 @@ public class VolumeServiceImpl implements VolumeService {
             } else {
                 for (Map.Entry<VolumeInfo, DataStore> entry : volumeToPool.entrySet()) {
                     VolumeInfo volume = entry.getKey();
+                    snapshotMgr.cleanupSnapshotsByVolume(volume.getId());
                     volume.processEvent(Event.OperationSuccessed);
                 }
                 future.complete(res);
--- a/server/src/com/cloud/storage/snapshot/SnapshotManager.java
+++ b/server/src/com/cloud/storage/snapshot/SnapshotManager.java
@@ -74,6 +74,8 @@ public interface SnapshotManager extends Configurable {
 
     boolean canOperateOnVolume(Volume volume);
 
+    void cleanupSnapshotsByVolume(Long volumeId);
+
     Answer sendToPool(Volume vol, Command cmd);
 
     SnapshotVO getParentSnapshot(VolumeInfo volume);
--- a/server/test/com/cloud/storage/snapshot/SnapshotManagerTest.java
+++ b/server/test/com/cloud/storage/snapshot/SnapshotManagerTest.java
@@ -16,6 +16,37 @@
 // under the License.
 package com.cloud.storage.snapshot;
 
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.anyLong;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+import java.util.List;
+import java.util.UUID;
+
+import org.apache.cloudstack.acl.ControlledEntity;
+import org.apache.cloudstack.acl.SecurityChecker.AccessType;
+import org.apache.cloudstack.context.CallContext;
+import org.apache.cloudstack.engine.subsystem.api.storage.DataStore;
+import org.apache.cloudstack.engine.subsystem.api.storage.SnapshotDataFactory;
+import org.apache.cloudstack.engine.subsystem.api.storage.SnapshotInfo;
+import org.apache.cloudstack.engine.subsystem.api.storage.SnapshotStrategy;
+import org.apache.cloudstack.engine.subsystem.api.storage.StorageStrategyFactory;
+import org.apache.cloudstack.engine.subsystem.api.storage.VolumeDataFactory;
+import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
+import org.apache.cloudstack.engine.subsystem.api.storage.SnapshotStrategy.SnapshotOperation;
+import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
+import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+import org.mockito.Spy;
+
 import com.cloud.configuration.Resource.ResourceType;
 import com.cloud.exception.InvalidParameterValueException;
 import com.cloud.exception.ResourceAllocationException;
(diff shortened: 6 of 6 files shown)


syed:
> Wait until ACS created the snapshot. ACS is creating a snapshot on local storage and is transferring this snapshot to secondary storage. But the latest snapshot on local storage will stay there. This is as expected.

Why is it expected that latest snapshot be present on the primary storage after it has been copied? Shouldn't it be deleted?

rhtyd:
@yvsubhash can you fix the conflicts and rebase against latest 4.8/4.9. Thanks.

server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
```diff
@@ -1111,6 +1111,20 @@ public boolean canOperateOnVolume(Volume volume) {
...
 
     @Override
+    public void cleanupSnapshotsByVolume(Long volumeId) {
+        List<SnapshotVO> volSnapShots = _snapshotDao.listByVolumeId(volumeId);
+        for(SnapshotVO snapshot: volSnapShots) {
+            SnapshotInfo info = snapshotFactory.getSnapshot(snapshot.getId(), DataStoreRole.Primary);
+            try {
+                snapshotSrv.deleteSnapshot(info);
+            } catch(CloudRuntimeException e) {
+                String msg = "Cleanup of Snapshot with uuid " + snapshot.getUuid() + " in primary storage is failed. Ignoring";
```
jburwell: This local variable is only used once.  Please consider collapsing into lint 1122.  Also, please add the message from the exception to the message to provide greater detail for debugging efforts.

server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
```diff
@@ -1111,6 +1111,20 @@ public boolean canOperateOnVolume(Volume volume) {
     }
 
     @Override
+    public void cleanupSnapshotsByVolume(Long volumeId) {
+        List<SnapshotVO> volSnapShots = _snapshotDao.listByVolumeId(volumeId);
+        for(SnapshotVO snapshot: volSnapShots) {
+            SnapshotInfo info = snapshotFactory.getSnapshot(snapshot.getId(), DataStoreRole.Primary);
```
jburwell: This appears to be an application side join.  Please consider creating a new query to retrieve all snapshot info instances associated with `volumeId` to reduce load on the database and simplify this method.

rhtyd:
@blueorangutan package

rhtyd:
@yvsubhash can you reply on raised review comments?

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 461
 Hypervisor xenserver
 NetworkType Advanced
 Passed=105
 Failed=0
 Skipped=7

**Failed tests:**

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_routers_network_ops.py
test_disk_offerings.py

yvsubhash:
@syed 
On the xenserver we always keep the latest snapshot in primary as xenserver needs it to create delta snapshots. If the snapshot is not present in xenserver, next snapshot would create full snapshot rather than delta snapshot. Please refer to the bug related to that
https://issues.apache.org/jira/browse/CLOUDSTACK-2630

syed:
Thanks @yvsubhash  Did the code review and it LGTM

mrunalinikankariya:
In ACS, all the snapshots are deleted from primary storage as soon as it is successfully backed up and copied on secondary storage, hence this issue is not valid for ACS. 

However a ticket should be raised to analysis and understand how incremental snapshots work if not a single copy kept on primary.

yvsubhash:
Closing this PR as it is no longer needed

yvsubhash:
The following ticket is opened for fixing this issue
https://issues.apache.org/jira/browse/CLOUDSTACK-9870

jahnaviboddu:
LGTM for Testing
 
1. Create a zone set-up with 2 host in one cluster.
2. set system.vm.use.local.storage parameter to "true" in Global Settings.
3. Create Compute Offerings with Storage Type option selected as local.
4. Create Disk Offerings with Storage Type option selected as local.
4. In Zone enable local storage for user vms option selected. 
5. Take recurring snapshot for Root-Disk.
6. Migrate the instance to another host.
7. After Migration snapshots from source host local storage are cleaned  up.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rafaelweingartner:
I have the same doubt as @syed here.

> Why is it expected that latest snapshot be present on the primary storage after it has been copied? Shouldn't it be deleted?

I am having a problem that might be cause by this one:
`create snapshot operation Failed for snapshotId: 3711, reason: SR_BACKEND_FAILURE_109. The snapshot chain is too long SR_BACKEND_FAILURE_109The snapshot chain is too long`

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/SnapshotDataFactoryImpl.java
```diff
@@ -64,6 +64,24 @@ public SnapshotInfo getSnapshot(DataObject obj, DataStore store) {
         return so;
     }
 
+    @Override
+    public List<SnapshotInfo> getSnapshots(long volumeId, DataStoreRole role) {
+
+        SnapshotDataStoreVO snapshotStore = snapshotStoreDao.findByVolume(volumeId, role);
+        if (snapshotStore == null) {
+            return null;
```
rafaelweingartner: This will cause a NPE. Consider returning an empty list.

rhtyd:
@yvsubhash please discuss and attend to outstanding issues

engine/storage/snapshot/src/org/apache/cloudstack/storage/snapshot/SnapshotDataFactoryImpl.java
yvsubhash: @rafaelweingartner  Deleting snapshots from primary after copy is complete taken care in xenserver case in postSnapshotCreation method in XenserverSnapshotStrategy.java for all snapshots except the lost one. So this one only takes care of the last snapshot that we keep in primary for delta snapshots.  null return is replaced with empty set

yvsubhash: @rafaelweingartner  the problem related to SR_BACKEND_FAILURE_109 does not seem to be related to this

rafaelweingartner:
@yvsubhash you are right... My problem is not related to this one here. The snapshots in XenServer is working as you said. However, sometimes if exceptions happen during copy of the snapshot to the secondary storage, snapshots in the primary storage are not cleaned. I am still investigating and trying to find a way to solve this problem.

BTW thanks for the reply!

yvsubhash:
@rafaelweingartner  Please confirm if there are any pending queries on this

server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
```diff
@@ -1331,6 +1331,21 @@ public boolean canOperateOnVolume(Volume volume) {
         return true;
     }
 
+    @Override
+    public void cleanupSnapshotsByVolume(Long volumeId) {
+        List<SnapshotInfo> infos = snapshotFactory.getSnapshots(volumeId, DataStoreRole.Primary);
+        for(SnapshotInfo info: infos) {
+            try {
+               if(info != null) {
```
rafaelweingartner: Do you need this null check?
This only happen if one uses the `set` method of List object to replace a position with a `Null` value

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:

rafaelweingartner:
@rhtyd I tested this fix in my dev environment and it solved the described problem.
I think this PR is in a good state to get merged. It could benefit from some unit tests though.

rhtyd:

rhtyd:
Merging this based on test results and code reviews.

mike-tutkowski:
I have noticed that not only is the snapshot on primary storage deleted when cleanupSnapshotsByVolume is invoked, but so is the copy of it that's on secondary storage. Can anyone else reproduce this?

mike-tutkowski:
In fact, one of my tests fails unless I comment out the cleanupSnapshotsByVolume method. The tests checks to make sure the volume that is migrated from one primary storage to another is still associated with the snapshot on secondary storage. As it turns out, when cleanupSnapshotByVolume is invoked, the snapshot is completed deleted.

mike-tutkowski:
OK, I have been able to reproduce this in a simple environment using only NFS.

I create a VM whose root disk resides on an NFS primary storage. I take a snapshot of this root disk. I then migrate the root disk from one NFS primary storage to another. The volume is no longer associated with the snapshot and the removed cell for the snapshot in the cloud.snapshots table has a date in it.

mike-tutkowski:
The code to retrieve only primary-storage snapshots is OK. The problem is when the snapshot is given to the Snapshot Service (snapshotSrv.deleteSnapshot(info);), at some point not only does the snapshot on the primary storage get deleted, but so does the snapshot on secondary storage.

When I comment out the code in snapshotMgr.cleanupSnapshotsByVolume, everything works the way it used to (meaning the snapshot on the original primary storage is stranded, but the volume is still associated with the correct snapshot on secondary storage).

rafaelweingartner:
You are right. That code seems to be the other way around. I mean, we should clean when there is a failure, right? At least that is what I understood by the author description.

mike-tutkowski:
The intention of the code is OK. A volume is being moved from one primary storage to another. If it had any snapshots on the source primary storage, we do not bring them along to the target primary primary storage. Those (primary-side) snapshots only helped in understanding the delta of changes that needed to be copied to secondary storage in the event that another snapshot was taken. Once we move the volume to its new primary storage, we lose knowledge of this delta information (which is OK). If the user asks for a new snapshot of the volume in question, we must take a full snapshot of that volume (but we should leave a new delta snapshot on this primary storage in case yet another snapshot is requested later). The intent of the code is to remove the unused (delta) snapshot that resided on the source primary storage because it is no longer needed. Its mistake is that it not only removes that delta snapshot, but it also removes the entirety of the snapshot (both what might have been on primary storage and what is on secondary storage). This leads to a data loss for the customer.

mike-tutkowski:
But I should also point out that you are correct in saying that in the event of a failed attempt to create a new snapshot, we should remove the (delta) snapshot from primary storage (as it might act as a false delta if we need to take a future snapshot).

rafaelweingartner:
Thanks for your details @mike-tutkowski. Can you help me understand where the problem is?

Is the problem in `snapshotSrv.deleteSnapshot(info)` that is deleting the snapshot from the secondary storage as well, and not only in the primary storage? Or is the code that the PR is introducing that is using the method `snapshotSrv.deleteSnapshot `  in an incorrect manner?

mike-tutkowski:
I am not in front of the code right now, but I think the root problem is that we do not currently have a mechanism which allows us to delete the (delta) snapshot on primary storage without removing what it corresponds to on secondary storage.

On Aug 11, 2018, at 9:49 AM, Rafael Weingärtner <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks for your details @mike-tutkowski<https://github.com/mike-tutkowski>. Can you help me understand where the problem is?

Is the problem in snapshotSrv.deleteSnapshot(info) that is deleting the snapshot from the secondary storage as well, and not only in the primary storage? Or is the code that the PR is introducing that is using the method snapshotSrv.deleteSnapshot in an incorrect manner?

rafaelweingartner:
Ok, so what this means is that the author when creating the PR used the `snapshotSrv.deleteSnapshot` thinking that it would only remove the snapshot in the hypervisor (primary storage). However, that method destroys the snapshot in ACS, which means removing in the DB and then from both storage systems (primary and secondary).

Having said that, what we need is to create a method to remove only snapshots from the primary storage only, right? Then, instead of using `snapshotSrv.deleteSnapshot`, we use this new method.

mike-tutkowski:
Yeah, we basically need to send the DeleteSnapshot command to the hypervisor and then remove the corresponding row from cloud.snapshot_store_ref. We then want to skip the rest of the stuff around deleting the snapshot from secondary storage and those corresponding DB updates.

On Aug 11, 2018, at 10:11 AM, Rafael Weingärtner <notifications@github.com<mailto:notifications@github.com>> wrote:


Ok, so what this means is that the author when creating the PR used the snapshotSrv.deleteSnapshot thinking that it would only remove the snapshot in the hypervisor (primary storage). However, that method destroys the snapshot in ACS, which means removing in the DB and then from both storage systems (primary and secondary).

Having said that, what we need is to create a method to remove only snapshots from the primary storage only, right? Then, instead of using snapshotSrv.deleteSnapshot, we use this new method.

rafaelweingartner:
Hello Mike, I am not able to reproduce the problem you described. I do think though that the code can be improved (I will explain it).

Let’s see an example. If I create a VM (vm-1); then I create two snapshots of this VM's root volume. I will have the following entries in `snapshot_store_ref` table.
-----------------------------------------------------------------------------------------------------------------------------------------------
| id  | store_id | snapshot_id | store_role | install_path                                              | state | physical_size | size        |
| --- | -------- | ----------- | ---------- | --------------------------------------------------------- | ----- | ------------- | ----------- |
| 422 | 2        | 253         | Image      | snapshots/2/7345/134a2524-72b6-4cfd-a802-89933fe63407.vhd | Ready | 1758786048    | 21474836480 |
| 423 | 5        | 254         | Primary    | 3037a7ed-5c93-45da-bc73-454ad21df793                      | Ready | 21474836480   | 21474836480 |
| 424 | 2        | 254         | Image      | snapshots/2/7345/5ddf9f65-16e7-4edd-9878-4f18ad26137b.vhd | Ready | 113508864     | 21474836480 |
-----------------------------------------------------------------------------------------------------------------------------------------------

Then, when I migrate the volume from shared storage to local storage, the code we are discussing is triggered.  The code is quite odd because it returns a list of `SnapshotObject`. Each snapshot will have a reference there. That object is then created with `SnapshotObject.getSnapshotObject(snapshot, store)`. This means, there will be some `SnapshotObject` that will reference a snapshot that is not in the primary storage. Then, when the delete method is called `snapshotSrv.deleteSnapshot(info)`, I get the following entries in the `snapshot_store_ref` table.
-----------------------------------------------------------------------------------------------------------------------------------------------
| id  | store_id | snapshot_id | store_role | install_path                                              | state | physical_size | size        |
| --- | -------- | ----------- | ---------- | --------------------------------------------------------- | ----- | ------------- | ----------- |
| 422 | 2        | 253         | Image      | snapshots/2/7345/134a2524-72b6-4cfd-a802-89933fe63407.vhd | Ready | 1758786048    | 21474836480 |
| 424 | 2        | 254         | Image      | snapshots/2/7345/5ddf9f65-16e7-4edd-9878-4f18ad26137b.vhd | Ready | 113508864     | 21474836480 |
-----------------------------------------------------------------------------------------------------------------------------------------------

This is the expected behavior. I am not sure if this behavior can be different for other hypervisors. I am using XenServer and testing with the latest commit in master.

There is one thing we can do though. We can change the method `snapshotFactory.getSnapshots` to return only `SnapshotObject` objects that reference the snapshots in the current primary storage from which the volume is being migrated out of. What do you think?

mike-tutkowski:
I should have placed these comments here (in GitHub) and let them propagate to JIRA. I'm going to copy/paste the relevant comments I made in JIRA here:

Comment:

After updating master on my local repo, I can no longer reproduce the issue I described above. With 46c56ea, the workflow behaves as expected. It would seem there was an issue in a previous version of master, but that it no longer exists.

Comment:

With regards to your snapshot-delete question:

As far as I know, if you migrate a volume from one primary storage to another, there should never exist any primary storage-based snapshots of this volume after the migration has taken place. That being the case, I don't think you should get back primary storage-based snapshots from snapshotFactory.getSnapshots that are from different primary storages.

mike-tutkowski:
This PR has been merged, but the corresponding ticket in JIRA is still marked as Open/Unresolved. I'm going to change that ticket to Closed/Fixed.

rafaelweingartner:
> As far as I know, if you migrate a volume from one primary storage to another, there should never exist any primary storage-based snapshots of this volume after the migration has taken place. That being the case, I don't think you should get back primary storage-based snapshots from snapshotFactory.getSnapshots that are from different primary storages

I do agree with you, but that is not how it is working right now. Should we fix that method then? I can open a PR tomorrow then.

mike-tutkowski:
Sure, open a PR and feel free to tag me when you would like a review. Thanks!

On Aug 13, 2018, at 6:41 PM, Rafael Weingärtner <notifications@github.com<mailto:notifications@github.com>> wrote:


As far as I know, if you migrate a volume from one primary storage to another, there should never exist any primary storage-based snapshots of this volume after the migration has taken place. That being the case, I don't think you should get back primary storage-based snapshots from snapshotFactory.getSnapshots that are from different primary storages

I do agree with you, but that is not how it is working right now. Should we fix that method then? I can open a PR tomorrow then.

PR #1743, opened 2016-10-31
CLOUDSTACK-8326: Always fill UDP checksums in DHCP replies in VR

wido:
In some cases the UDP checksums in packets from DHCP servers are
incorrect. This is a problem for some DHCP clients that ignore
packets with bad checksums. This patch inserts an iptables rule
to ensure DHCP servers always send packets with correct checksums.

Due to this bug DHCP offers are sometimes not accepted by Instances.

The end-result without this fix is no connectivity for the Instance
due to the lack of a IPv4 address.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
@@ -359,6 +359,10 @@ class CsIP:
                         "-m state --state RELATED,ESTABLISHED " +
                         "-j CONNMARK --restore-mark --nfmask 0xffffffff --ctmask 0xffffffff"])
 
+        self.fw.append(["mangle", "front",
+                        "-A POSTROUTING " +
+                        "-p udp -m udp --dport 68 -j CHECKSUM --checksum-fill"])
+
         if self.get_type() in ["public"]:
             self.fw.append(["mangle", "front",
                             "-A PREROUTING " +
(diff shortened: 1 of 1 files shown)


rhtyd:
LGTM, not sure if this can lead to any side-effects.
@blueorangutan package

ustcweizhou:
in 4.9, ALL iptables rules will be saved in /tmp/rules.save
I think it is better to add the rules in systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py

```
+        self.fw.append(["mangle", "front",
+                        "-A POSTROUTING " +
+                        "-p udp -m udp --dport 68 -j CHECKSUM --checksum-fill"])
```

in fw_router and fw_vpcrouter

wido:
Thanks @ustcweizhou for the pointer. It was already fixed in fw_router, but not in fw_vpcrouter there.

How does this look?

ustcweizhou:
@wido maybe the guy who made the change forgot to add it in fw_vpcrouter.
In fw_router, it should be moved out from "if self.get_type() in ["public"]", because the network type is 'guest' not 'public' for shared networks (in Basic zone or Advanced zone with SG)

The-Loeki:
See https://bugzilla.redhat.com/show_bug.cgi?id=910619 for what's actually going on (and what actually should be going on)

wido:
Good feedback @ustcweizhou, addressed that.

Thanks @The-Loeki for the report, this is indeed a upstream issue somewhere else, but if we can mitigate in the meantime that would be great.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

ustcweizhou:
@The-Loeki as I remember, that is a bug in dhcp3-client. The issue does exist in Ubuntu 12.04 (guest vm), but fixed in Ubuntu 12.04.3. (I can not find the related link now). Anyway, the iptables rules fix the issue permanently.

rhtyd:
Previous Trillian test job failed due to backend issue, I've rekicked them now.

wido:
Is this one ready to merge? It fixes a rather nasty bug in the VR which causes big problems when no DHCP offer is received by the Instance(s).

rhtyd:
@ustcweizhou @The-Loeki do we have LGTM from you guys on this? Thanks.

LGTM.

ustcweizhou:
@rhtyd @wido I would suggest one more change: move line 378-380 after line 360.

wido:
Good suggestion @ustcweizhou, done that!

ustcweizhou:
LGTM !

rhtyd:
LGTM on code review, merging this based on review and test lgtms. Few errors around VPC/VR are known intermittent issues.

PR #1607, opened 2016-07-08
SSVM downloader now handles redirects properly.

ProjectMoon:
Previously it was using the HttpClient to make an initial request to an ISO. This would follow redirects. Then it would make another request using built-in Java URL and InputStream, which doesn't follow redirects. This results in the ISO getting stuck at 0% forever and also causing DOS effects.

PR #1746, opened 2016-11-01
SSVM downloader now handles redirects properly.

ProjectMoon:
New version of #1607, opened against 4.9 so it can be forward merged.

**Original Description**
Previously it was using the HttpClient to make an initial request to an ISO. This would follow redirects. Then it would make another request using built-in Java URL and InputStream, which doesn't follow redirects. This results in the ISO getting stuck at 0% forever and also causing DOS effects.

### DIFF
--- a/core/src/com/cloud/storage/template/HttpTemplateDownloader.java
+++ b/core/src/com/cloud/storage/template/HttpTemplateDownloader.java
@@ -25,7 +25,6 @@ import java.io.InputStream;
 import java.io.RandomAccessFile;
 import java.net.URI;
 import java.net.URISyntaxException;
-import java.net.URL;
 import java.util.Date;
 
 import org.apache.cloudstack.utils.imagestore.ImageStoreUtil;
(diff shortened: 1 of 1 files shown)


rhtyd:
LGTM.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

core/src/com/cloud/storage/template/HttpTemplateDownloader.java
```diff
@@ -117,7 +116,7 @@ public boolean retryMethod(final HttpMethod method, final IOException exception,
             request = new GetMethod(downloadUrl);
             request.getParams().setParameter(HttpMethodParams.RETRY_HANDLER, myretryhandler);
             completionCallback = callback;
-            //this.request.setFollowRedirects(false);
```
syed: Nevermind. Looks like this is the port to 4.9

ProjectMoon: The change was originally done on our 4.7 branch before being ported to master and then here. Even back in 4.7 this line was commented.

ProjectMoon: In fact, this line has been commented since the very beginning! https://github.com/apache/cloudstack/commit/05c020e1f6679edb74ef9ee513fc6240db981139

rhtyd:
LGTM on tests and code. Pinging for review -- @abhinandanprateek @murali-reddy @karuturi and others.
@syed do you have any outstanding comments on this PR, or you're LGTM on this.

PR #1762, opened 2016-11-11
CLOUDSTACK-9595 Transactions are not getting retried in case of datab…

yvsubhash:
CLOUDSTACK-9595 Transactions are not getting retried in case of database deadlock errors
Problem Statement
--------------------------
MySQLTransactionRollbackException is seen frequently in logs
Root Cause
----------------
Attempts to lock rows in the core data access layer of database fails if there is a possibility of deadlock. However Operations are not getting retried in case of deadlock. So introducing retries here
Solution
-----------
Operations would be retried after some wait time in case of dead lock exception

### DIFF
--- a/server/src/com/cloud/network/IpAddressManagerImpl.java
+++ b/server/src/com/cloud/network/IpAddressManagerImpl.java
@@ -291,6 +291,8 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
 
     SearchBuilder<IPAddressVO> AssignIpAddressSearch;
     SearchBuilder<IPAddressVO> AssignIpAddressFromPodVlanSearch;
+    private final Object _allocatedLock = new Object();
+    private final Object _allocatingLock = new Object();
 
     static Boolean rulesContinueOnErrFlag = true;
 
@@ -759,7 +761,7 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
 
                 Filter filter = new Filter(IPAddressVO.class, "vlanId", true, 0l, 1l);
 
-                List<IPAddressVO> addrs = _ipAddressDao.lockRows(sc, filter, true);
+                List<IPAddressVO> addrs = _ipAddressDao.search(sc, filter, false);
 
                 // If all the dedicated IPs of the owner are in use fetch an IP from the system pool
                 if (addrs.size() == 0 && fetchFromDedicatedRange) {
@@ -769,7 +771,7 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
                         fetchFromDedicatedRange = false;
                         sc.setParameters("vlanId", nonDedicatedVlanDbIds.toArray());
                         errorMessage.append(", vlanId id=" + Arrays.toString(nonDedicatedVlanDbIds.toArray()));
-                        addrs = _ipAddressDao.lockRows(sc, filter, true);
+                        addrs = _ipAddressDao.search(sc, filter, false);
                     }
                 }
 
@@ -804,24 +806,21 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
                 addr.setAllocatedInDomainId(owner.getDomainId());
                 addr.setAllocatedToAccountId(owner.getId());
                 addr.setSystem(isSystem);
+
                 if (displayIp != null) {
                     addr.setDisplay(displayIp);
                 }
-
-                if (assign) {
-                    markPublicIpAsAllocated(addr);
-                } else {
-                    addr.setState(IpAddress.State.Allocating);
-                }
-                addr.setState(assign ? IpAddress.State.Allocated : IpAddress.State.Allocating);
 
                 if (vlanUse != VlanType.DirectAttached) {
                     addr.setAssociatedWithNetworkId(guestNetworkId);
                     addr.setVpcId(vpcId);
                 }
 
-                _ipAddressDao.update(addr.getId(), addr);
-
+                if (assign) {
+                    markPublicIpAsAllocated(addr);
+                } else {
+                    markPublicIpAsAllocating(addr);
+                }
                 return addr;
             }
         });
@@ -836,35 +835,51 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
     @DB
     @Override
     public void markPublicIpAsAllocated(final IPAddressVO addr) {
-
-        Transaction.execute(new TransactionCallbackNoReturn() {
-            @Override
-            public void doInTransactionWithoutResult(TransactionStatus status) {
-                Account owner = _accountMgr.getAccount(addr.getAllocatedToAccountId());
-                synchronized (this) {
+        synchronized (_allocatedLock) {
+            Transaction.execute(new TransactionCallbackNoReturn() {
+                @Override
+                public void doInTransactionWithoutResult(TransactionStatus status) {
+                    Account owner = _accountMgr.getAccount(addr.getAllocatedToAccountId());
                     if (_ipAddressDao.lockRow(addr.getId(), true) != null) {
                         IPAddressVO userIp = _ipAddressDao.findById(addr.getId());
                         if (userIp.getState() == IpAddress.State.Allocating || addr.getState() == IpAddress.State.Free) {
                             addr.setState(IpAddress.State.Allocated);
-                            _ipAddressDao.update(addr.getId(), addr);
-                            // Save usage event
-                            if (owner.getAccountId() != Account.ACCOUNT_ID_SYSTEM) {
-                                VlanVO vlan = _vlanDao.findById(addr.getVlanId());
-                                String guestType = vlan.getVlanType().toString();
-                                if (!isIpDedicated(addr)) {
-                                    UsageEventUtils.publishUsageEvent(EventTypes.EVENT_NET_IP_ASSIGN, owner.getId(), addr.getDataCenterId(), addr.getId(),
-                                            addr.getAddress().toString(),
-                                            addr.isSourceNat(), guestType, addr.getSystem(), addr.getClass().getName(), addr.getUuid());
-                                }
-                                if (updateIpResourceCount(addr)) {
-                                    _resourceLimitMgr.incrementResourceCount(owner.getId(), ResourceType.public_ip);
+                            if (_ipAddressDao.update(addr.getId(), addr)) {
+                                // Save usage event
+                                if (owner.getAccountId() != Account.ACCOUNT_ID_SYSTEM) {
+                                    VlanVO vlan = _vlanDao.findById(addr.getVlanId());
+                                    String guestType = vlan.getVlanType().toString();
+                                    if (!isIpDedicated(addr)) {
+                                        UsageEventUtils.publishUsageEvent(EventTypes.EVENT_NET_IP_ASSIGN, owner.getId(), addr.getDataCenterId(), addr.getId(),
+                                                addr.getAddress().toString(),
+                                                addr.isSourceNat(), guestType, addr.getSystem(), addr.getClass().getName(), addr.getUuid());
+                                    }
+                                    if (updateIpResourceCount(addr)) {
+                                        _resourceLimitMgr.incrementResourceCount(owner.getId(), ResourceType.public_ip);
+                                    }
                                 }
                             }
                         }
                     }
                 }
-            }
-        });
+            });
+        }
+    }
+
+    @DB
+    private void markPublicIpAsAllocating(final IPAddressVO addr) {
+        synchronized (_allocatingLock) {
+            Transaction.execute(new TransactionCallbackNoReturn() {
+                @Override
+                public void doInTransactionWithoutResult(TransactionStatus status) {
+
+                    if (_ipAddressDao.lockRow(addr.getId(), true) != null) {
+                        addr.setState(IpAddress.State.Allocating);
+                        _ipAddressDao.update(addr.getId(), addr);
+                    }
+                }
+            });
+        }
     }
 
     private boolean isIpDedicated(IPAddressVO addr) {
@@ -901,40 +916,28 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
 
         PublicIp ip = null;
         try {
-            ip = Transaction.execute(new TransactionCallbackWithException<PublicIp, InsufficientAddressCapacityException>() {
-                @Override
-                public PublicIp doInTransaction(TransactionStatus status) throws InsufficientAddressCapacityException {
-                    Account owner = _accountDao.acquireInLockTable(ownerId);
+            Account ownerAccount = _accountDao.acquireInLockTable(ownerId);
 
-                    if (owner == null) {
-                        // this ownerId comes from owner or type Account. See the class "AccountVO" and the annotations in that class
-                        // to get the table name and field name that is queried to fill this ownerid.
-                        ConcurrentOperationException ex = new ConcurrentOperationException("Unable to lock account");
-                        throw ex;
-                    }
-                    if (s_logger.isDebugEnabled()) {
-                        s_logger.debug("lock account " + ownerId + " is acquired");
-                    }
-                    boolean displayIp = true;
-                    if (guestNtwkId != null) {
-                        Network ntwk = _networksDao.findById(guestNtwkId);
-                        displayIp = ntwk.getDisplayNetwork();
-                    } else if (vpcId != null) {
-                        VpcVO vpc = _vpcDao.findById(vpcId);
-                        displayIp = vpc.isDisplay();
-                    }
+            if (ownerAccount == null) {
+                // this ownerId comes from owner or type Account. See the class "AccountVO" and the annotations in that class
+                // to get the table name and field name that is queried to fill this ownerid.
+                ConcurrentOperationException ex = new ConcurrentOperationException("Unable to lock account");
+                throw ex;
+            }
+            if (s_logger.isDebugEnabled()) {
+                s_logger.debug("lock account " + ownerId + " is acquired");
+            }
+            boolean displayIp = true;
+            if (guestNtwkId != null) {
+                Network ntwk = _networksDao.findById(guestNtwkId);
+                displayIp = ntwk.getDisplayNetwork();
+            } else if (vpcId != null) {
+                VpcVO vpc = _vpcDao.findById(vpcId);
+                displayIp = vpc.isDisplay();
+            }
 
-                    PublicIp ip = fetchNewPublicIp(dcId, null, null, owner, VlanType.VirtualNetwork, guestNtwkId, isSourceNat, false, null, false, vpcId, displayIp);
-                    IPAddressVO publicIp = ip.ip();
+            return fetchNewPublicIp(dcId, null, null, owner, VlanType.VirtualNetwork, guestNtwkId, isSourceNat, false, null, false, vpcId, displayIp);
 
-                    markPublicIpAsAllocated(publicIp);
-                    _ipAddressDao.update(publicIp.getId(), publicIp);
-
-                    return ip;
-                }
-            });
-
-            return ip;
         } finally {
             if (owner != null) {
                 if (s_logger.isDebugEnabled()) {
--- a/engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
+++ b/engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
@@ -539,7 +539,6 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
     public List<HostVO> findAndUpdateDirectAgentToLoad(long lastPingSecondsAfter, Long limit, long managementServerId) {
         TransactionLegacy txn = TransactionLegacy.currentTxn();
 
-        txn.start();
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Resetting hosts suitable for reconnect");
         }
@@ -555,6 +554,7 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
             s_logger.debug("Acquiring hosts for clusters already owned by this management server");
         }
         List<Long> clusters = findClustersOwnedByManagementServer(managementServerId);
+        txn.start();
         if (clusters.size() > 0) {
             // handle clusters already owned by @managementServerId
             SearchCriteria<HostVO> sc = UnmanagedDirectConnectSearch.create();


serg38:
LGTM. Finally !!! We have been seeing occasional deadlocks in environments with high level transaction rate. @rhtyd @jburwell This could be a good add to 4.8/4.9.

jburwell:
@yvsubhash according to the (MySQL deadlock documenation)[http://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks.html],  a `MYSQL_DEADLOCK_ERROR_CODE` error indicates the enclosing transaction has been rolled back.  The proper handling for this error is to re-execute all statements executed in the aborted transaction.  From a best practices perspective, all base data should be re-retrieved and changed to ensure logical consistency with changes made by the transaction that won deadlock resolution.

As I understand this patch, only the most recently executed DML is retried.  Therefore, any previously executed changes will be discarded and the DML will be re-executed either in a new transaction or in auto-commit (I didn't look up how the client handles the transaction context in this scenario).  If my understanding is correct, this patch could lead to issues ranging from unexpected foreign key integrity errors to data corruption.

Rather attempting to implement a generic retry, I think the best approach to addressing deadlocks is to treat them bugs.  This patch could be modified to provide detailed logging information about the conditions under which a deadlock occurs providing the information necessary to refactor the system to avoid lock contention.

serg38:
@jburwell @yvsubhash My understanding that all roll back statements will receive MYSQL_DEADLOCK_ERROR_CODE  and will be retired as a part of this patch.

jburwell:
@serg38 my reading of the code is that only the most recently attempted DML will be re-executed.  Furthermore, retrying without refreshing the base data can also lead to data corruption.  The best thing to do in a case of a dead lock is to fail and rollback due to the risk of data corruption.

serg38:
@jburwell I thought that most if not all of ACS interaction through DAO is rather atomic transactions. Do we have cases of multiple DML statements as a part of the same transaction? We have been seeing quite a few deadlock in a high transaction volume environments where multiple management servers are employed. This causes quite a pain for users due to the randomness and no good recourse/explanation. I would argue that proper retry is a better choice should we cover all the cases including all cases with complex transactions. We have been successful leveraging this approach in systems built on the top of ACS.

jburwell:
@serg38 that is not a safe assumption.  Transactions often span multiple statements and methods across DAOs.  `TransactionLegacy` has a transaction stacking/nested model that further occludes when a transaction actually completely.

Deadlocks are a severe problem that need to be fixed.  Unfortunately, this patch would do more harm than good as it would eventually corrupt the database.   In, and of themselves, retries are also a very expensive solution to the problem both in terms of the engineering effort required to do it properly and the extra stress placed on the database to perform additional work that will likely fail.  Furthermore, a generic **and** correct retry mechanism is a very difficult thing to write.  Given the way transaction boundaries are managed in ACS, I think such an effort would be nearly impossible.

In a properly written application, deadlocks should very rarely, if ever, occur.  Their presence is a symptom of improper transaction handling and/or poor lock management problems.   Therefore, my suggestion is that we change this patch to log details about the context in which deadlocks occur.  We can then use this information to identify the areas in ACS where these contention problems are location and fix the root cause.

serg38:
@jburwell @yvsubhash  I might be wrong but this PR will retry on deadlock for only 2  DAO methods searchIncludingRemoved and customSearchIncludingRemoved. No update methods are set with this retry mechanism. If that's the case there is no risk of corrupting DB.

jburwell:
@serg38 there remains a risk when those methods are executed in the context of an open transaction where DMLs have already been executed and subsequent DMLs will be executed.  In this scenario, the first set of the changes would be lost due to the rollback triggered by the query deadlock with the second set proceeding successfully.

serg38:
@jburwell I concur but if @yvsubhash verified that those methods don't participate in complex DML transactions this might be still a good start. If so this approach might be expanded later to multi DML transaction so that each piece can be retired individually. I myself traced few deadlocks in ACS using  native mysql deadlock logging and it doesn't seem there would be a viable alternative to retires due to well known complexity of ACS DB operations.

jburwell:
@serg38 with custom plugins, there is no way to reliably perform such tracing.  I can think of batch cleanup operations in the storage layer that follow the pattern I described.  Even if there were, we would have planted a landline for future changes to the system.  Deadlocks are significant technical debt that are clearly causing significant operational issues.  Unfortunately, there is no way to address them generically

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 135
 Hypervisor xenserver
 NetworkType Advanced
 Passed=102
 Failed=3
 Skipped=6

**Failed tests:**
* test_non_contigiousvlan.py

 * test_extendPhysicalNetworkVlan Failed

* test_deploy_vm_iso.py

 * test_deploy_vm_from_iso Failing since 19 runs

* test_vm_life_cycle.py

 * test_10_attachAndDetach_iso Failing since 20 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_snapshots.py
test_network.py
test_router_dns.py
test_login.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_routers_network_ops.py
test_disk_offerings.py

jburwell:
Due to the previous discussion, I am -1 on merging this PR.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@abhinandanprateek can you help reviewing this one, thanks.

jburwell:
@rhtyd I am -1 on this PR

serg38:
@jburwell We've been running this fix as a part of proprietary CS for several weeks now. We are observing elimination of deadlocks and no DB corruption. Retry seems to be the only realistic way of dealing with deadlocks in complex environment like ACS. Can we come up with a limited scope/conditions of this PR to move forward ?

jburwell:
@serg38 corruption could happen at any point -- it's a ticking time bomb.  From a ACID perspective, this patch fails from a consistency perspective.  All data being updated must be re-queried and validated in order to ensure the consistency guarantee is not violated.  In a high volume system, it's not a matter of if, but when a sequence of events will occur and corrupt the database.   Bear in mind, these corruptions be in the content of the data and would not yield a MySQL error.  They will be phenomenon such as phantom rows or inconsistent data updates

As I said previously, the only real solution to deadlocks is to fix the way the system manages transactions and locks.  This patch is merely hiding an error while creating the potential for far larger problems.

For these reasons, I remain -1 on merging this patch.

serg38:
What about if the author can figure out a way to identify all part of transaction being cancelled and retry all parts? Or retry the whole transaction? It would  be nice  to open a path for the author to implement this

abhinandanprateek:
Even trying the full transaction again could be problematic as there might be checks done before firing the transaction that may not be valid now.
The thing is it may mostly work, but it is not fool proof.

serg38:
@rafaelweingartner @swill @wido @koushik-das @karuturi @rhtyd @jburwell  Let's ask a different question. If we are to try to implement a general way of dealing with deadlocks in ACS how could it be done to ensure DB consistency and correct transaction retry?

rafaelweingartner:
@serg38 I have just now started reading this PR (excuse me if I overlooked some information).

> If we are to try to implement a general way of dealing with deadlocks in ACS how could it be done to ensure DB consistency and correct transaction retry?

Answering your question; in my opinion, we should not “try” to implement a general way of managing transactions. We are only having this type of problem because instead of using a framework to manage access and transactions in databases, it was developed a module to do that and incorporated to ACS; this means we have to maintain and live with this code. 

Now, the problem is that it would be a Dantesque task to change the way ACS manages transactions today.

I am with John on this one, retrying is not a good idea; it can hide problems, cause overheads and cause even more headaches.  I think that the best approach is to deal with this type of problem on the fly; this means, as John said, addressing them as bugs when they are reported.

Having said that, I have not helped a bit to solve the problem… Let’s see if I can be of any help. 

I was reading the ticket #CLOUDSTACK-9595. It seems that the problem (reported there) happened when a VM was being removed from a table “instance_group_vm_map”. I just do not understand because the method called is “UserVmManagerImpl.addInstanceToGroup”. I am hoping that this makes sense. Anyways…

The MYSQL docs have the following on deadlocks:
> A deadlock is a situation where different transactions are unable to proceed because each holds a lock that the other needs

This means, there was something else being executed when that VM was deleted/added, and this caused the deadlock and the exception. Probably something else is using the table “instance_group_vm_map”.

I think we should track these two tasks/processes that can cause the problem and work them out, instead of looking for a generic way to deal with this situation. Maybe these processes that are causing deadlock are locking tables that are not needed or executing some processing that could be avoided or modified.

Do we use case that can reproduce the problem?

serg38:
Here it is few samples of deadlocks we observe in high transaction volume environment with multiple management servers. As you can see most of them are concurrent operations from different management servers and either select or select for update statements. The following 4 types account for the majority of deadlock s we saw so far ( 80-90% of all deadlocks). Deadlock 1-3 happens much more often than deadlock 4.  It is next to impossible to reproduce since they occur one in few days with 4 management servers and average VM deployment volume of 3000 a day.

Deadlock type 1:

InnoDB: transactions deadlock detected, dumping detailed information.
151217  3:08:20
*** (1) TRANSACTION:
TRANSACTION BB4D4C91D, ACTIVE 0 sec fetching rows
mysql tables in use 1, locked 1
LOCK WAIT 11 lock struct(s), heap size 3112, 5 row lock(s)
MySQL thread id 47654, OS thread handle 0x7f0475bdd700, query id 3821358107 ussclpdcsmgt012.autodesk.com 10.41.13.14 cloud Sending data
SELECT host.id, host.disconnected, host.name, host.status, host.type, host.private_ip_address, host.private_mac_address, host.private_netmask, host.public_netmask, host.public_ip_address, host.public_mac_address, host.storage_ip_address, host.cluster_id, host.storage_netmask, host.storage_mac_address, host.storage_ip_address_2, host.storage_netmask_2, host.storage_mac_address_2, host.hypervisor_type, host.proxy_port, host.resource, host.fs_type, host.available, host.setup, host.resource_state, host.hypervisor_version, host.update_count, host.uuid, host.data_center_id, host.pod_id, host.cpu_sockets, host.cpus, host.url, host.speed, host.ram, host.parent, host.guid, host.capabilities, host.total_size, host.last_ping, host.mgmt_server_id, host.dom0_memory, host.version, host.created, host.removed FROM host WHERE host.resource IS NOT NULL  AND host.mgmt_server_id = 345048964870 
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
*** (2) TRANSACTION:
TRANSACTION BB4D4C915, ACTIVE 1 sec fetching rows, thread declared inside InnoDB 449
mysql tables in use 3, locked 3
29 lock struct(s), heap size 6960, 15 row lock(s), undo log entries 1
MySQL thread id 47623, OS thread handle 0x7f0a47074700, query id 3821724056 ussclpdcsmgt013.autodesk.com 10.41.13.15 cloud Copying to tmp table
SELECT host.id, host.disconnected, host.name, host.status, host.type, host.private_ip_address, host.private_mac_address, host.private_netmask, host.public_netmask, host.public_ip_address, host.public_mac_address, host.storage_ip_address, host.cluster_id, host.storage_netmask, host.storage_mac_address, host.storage_ip_address_2, host.storage_netmask_2, host.storage_mac_address_2, host.hypervisor_type, host.proxy_port, host.resource, host.fs_type, host.available, host.setup, host.resource_state, host.hypervisor_version, host.update_count, host.uuid, host.data_center_id, host.pod_id, host.cpu_sockets, host.cpus, host.url, host.speed, host.ram, host.parent, host.guid, host.capabilities, host.total_size, host.last_ping, host.mgmt_server_id, host.dom0_memory, host.version, host.created, host.removed FROM host  LEFT OUTER JOIN op_host_transfer ON host.id=op_host_transfer.id
*** (2) HOLDS THE LOCK(S):
RECORD LOCKS space id 0 page no 147488 n bits 840 index `i_host__removed` of table `cloud`.`host` trx id BB4D4C915 lock_mode X locks rec but not gap


Deadlock 2:

InnoDB: transactions deadlock detected, dumping detailed information.
151218 11:03:00
*** (1) TRANSACTION:
TRANSACTION BBB232C81, ACTIVE 51 sec starting index read
mysql tables in use 1, locked 1
LOCK WAIT 3 lock struct(s), heap size 1248, 2 row lock(s)
MySQL thread id 57308, OS thread handle 0x7f0a45c24700, query id 5217973695 ussclpdcsmgt013.autodesk.com 10.41.13.15 cloud Sending data
SELECT resource_count.id, resource_count.type, resource_count.account_id, resource_count.domain_id, resource_count.count FROM resource_count WHERE resource_count.id IN (5083,4867,5079,33652,5077)  FOR UPDATE
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
*** (2) TRANSACTION:
TRANSACTION BBB2254AC, ACTIVE 116 sec starting index read, thread declared inside InnoDB 500
mysql tables in use 1, locked 1
207 lock struct(s), heap size 31160, 1650 row lock(s), undo log entries 2
MySQL thread id 56926, OS thread handle 0x7f04756c9700, query id 5218549710 ussclpdcsmgt014.autodesk.com 10.41.13.16 cloud Sending data
SELECT resource_count.id, resource_count.type, resource_count.account_id, resource_count.domain_id, resource_count.count FROM resource_count WHERE resource_count.id IN (5083,4867,5079,33652,5077)  FOR UPDATE

Deadlock 3:

** (1) TRANSACTION:
TRANSACTION BBB232C81, ACTIVE 51 sec starting index read
mysql tables in use 1, locked 1
LOCK WAIT 3 lock struct(s), heap size 1248, 2 row lock(s)
MySQL thread id 57308, OS thread handle 0x7f0a45c24700, query id 5217973695 ussclpdcsmgt013.autodesk.com 10.41.13.15 cloud Sending data
SELECT resource_count.id, resource_count.type, resource_count.account_id, resource_count.domain_id, resource_count.count FROM resource_count WHERE resource_count.id IN (5083,4867,5079,33652,5077)  FOR UPDATE
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
*** (2) TRANSACTION:
TRANSACTION BBB2254AC, ACTIVE 116 sec starting index read, thread declared inside InnoDB 500
mysql tables in use 1, locked 1
207 lock struct(s), heap size 31160, 1650 row lock(s), undo log entries 2
MySQL thread id 56926, OS thread handle 0x7f04756c9700, query id 5218549710 ussclpdcsmgt014.autodesk.com 10.41.13.16 cloud Sending data
SELECT resource_count.id, resource_count.type, resource_count.account_id, resource_count.domain_id, resource_count.count FROM resource_count WHERE resource_count.id IN (5083,4867,5079,33652,5077)  FOR UPDATE
*** (2) HOLDS THE LOCK(S):

Deadlock 4:
*** (1) TRANSACTION:
TRANSACTION C3BDD81EF, ACTIVE 0 sec fetching rows
mysql tables in use 1, locked 1
LOCK WAIT 55 lock struct(s), heap size 6960, 3 row lock(s), undo log entries 1
MySQL thread id 250487, OS thread handle 0x7f0a460b6700, query id 32833273614 ussclpdcsmgt013.autodesk.com 10.41.13.15 cloud updating
DELETE FROM vm_reservation WHERE vm_reservation.vm_id = 869089
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
*** (2) TRANSACTION:
TRANSACTION C3BDD81FC, ACTIVE 0 sec inserting, thread declared inside InnoDB 500
mysql tables in use 1, locked 1
4 lock struct(s), heap size 1248, 2 row lock(s), undo log entries 2
MySQL thread id 250553, OS thread handle 0x7f0a3720c700, query id 32833273762 ussclpdcsmgt013.autodesk.com 10.41.13.15 cloud update
INSERT INTO volume_reservation (volume_reservation.vm_reservation_id, volume_reservation.vm_id, volume_reservation.volume_id, volume_reservation.pool_id) VALUES (997419, 1009484, 918449, 316)
*** (2) HOLDS THE LOCK(S):

rafaelweingartner:
Thanks, @serg38.
Looking at the SQLs you posted. We could start to discuss whether or not some SQLs statements need locking transactions.

Ignoring Deadlocks 3 and 4 for now, I think we could start with the ones the look the simplest (Deadlocks 1 and 2). 

These SQLS have probably being generated, so tracking them on ACS may not be that easy, but at first glance, I feel that we could execute them without needing lock in the database. 

I tried to find the first SQL, without success. Would you mind helping me pin point where in the code the SQL from transaction 1 at deadlock 1 is generated? Then, we can evaluate if it is or not needed a lock there. 

Are the SQLs you showed complete? I found a place that could generate SQLs similar to the one at transaction 1 and deadlock 1, but this code adds one extra where clause.

The method I am talking about is:
com.cloud.cluster.agentlb.ClusterBasedAgentLoadBalancerPlanner.getHostsToRebalance(long, int)

serg38:
@rafaelweingartner Looks  like the deadlocks 2 and 3 are the same. I scanned our production  log  and since last December we had 6400 deadlocks. Out of them close to 6000 were Deadlock 1 
20 were Deadlock 2 and 700 of a different Deadlock 5. The other deadlocks were in negligible numbers. I think if we figure out Deadlock 1 and Deadlock 5 this will be good start. I will try to find the source of transactions for them. In production we run a commercial distribution based in most part on  4.7 branch of ACS. 

Deadlock 5

*** (1) TRANSACTION:
TRANSACTION D518886F8, ACTIVE 2 sec fetching rows
mysql tables in use 4, locked 4
LOCK WAIT 24 lock struct(s), heap size 3112, 8 row lock(s), undo log entries 17
MySQL thread id 29781, OS thread handle 0x7f9df36db700, query id 3625404021 ussclpdcsmgt012.autodesk.com 10.41.13.14 cloud Sorting result
SELECT user_ip_address.id, user_ip_address.account_id, user_ip_address.domain_id, user_ip_address.public_ip_address, user_ip_address.data_center_id, user_ip_address.source_n
at, user_ip_address.allocated, user_ip_address.vlan_db_id, user_ip_address.one_to_one_nat, user_ip_address.vm_id, user_ip_address.state, user_ip_address.mac_address, user_ip
_address.source_network_id, user_ip_address.network_id, user_ip_address.uuid, user_ip_address.physical_network_id, user_ip_address.is_system, user_ip_address.vpc_id, user_ip
_address.dnat_vmip, user_ip_address.is_portable, user_ip_address.display, user_ip_address.removed, user_ip_address.created FROM user_ip_address  INNER JOIN vlan ON user_ip_a
ddress.vlan_db_id=vlan.id WHERE user_ip_address.data_center_id = 6  AND user_ip_address.allocated IS NULL  AND user_ip_address.vlan_db_id IN (32,33,36,37,41,61,62,91,92,93,9
4,106,107,108,109,11
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
*** (2) TRANSACTION:
TRANSACTION D5188582B, ACTIVE 17 sec updating or deleting, thread declared inside InnoDB 499
mysql tables in use 1, locked 1
25 lock struct(s), heap size 3112, 13 row lock(s), undo log entries 18
MySQL thread id 29820, OS thread handle 0x7fa35a868700, query id 3625417999 ussclpdcsmgt013.autodesk.com 10.41.13.15 cloud Updating
UPDATE user_ip_address SET user_ip_address.source_nat=0, user_ip_address.is_system=0, user_ip_address.account_id=3309, user_ip_address.allocated='2016-03-25 15:36:39', user_ip_address.state='Allocated', user_ip_address.domain_id=335 WHERE user_ip_address.id = 3284
*** (2) HOLDS THE LOCK(S):

serg38:
@rafaelweingartner Tried tracing where  deadlock 5 originated. It seems both transactions are part of the same method fetchNewPublicIp in IpAddressManagerImpl  . Transactions are executed on different management servers. 
Update is triggered through markPublicIpAsAllocated  method 

Select seems to come from there as well  fetchNewPublicIp in IpAddressManagerImpl

        AssignIpAddressFromPodVlanSearch = _ipAddressDao.createSearchBuilder();
        AssignIpAddressFromPodVlanSearch.and("dc", AssignIpAddressFromPodVlanSearch.entity().getDataCenterId(), Op.EQ);
        AssignIpAddressFromPodVlanSearch.and("allocated", AssignIpAddressFromPodVlanSearch.entity().getAllocatedTime(), Op.NULL);
        SearchBuilder<VlanVO> podVlanSearch = _vlanDao.createSearchBuilder();
        podVlanSearch.and("type", podVlanSearch.entity().getVlanType(), Op.EQ);
        podVlanSearch.and("networkId", podVlanSearch.entity().getNetworkId(), Op.EQ);
        SearchBuilder<PodVlanMapVO> podVlanMapSB = _podVlanMapDao.createSearchBuilder();
        podVlanMapSB.and("podId", podVlanMapSB.entity().getPodId(), Op.EQ);
        AssignIpAddressFromPodVlanSearch.join("podVlanMapSB", podVlanMapSB, podVlanMapSB.entity().getVlanDbId(), AssignIpAddressFromPodVlanSearch.entity().getVlanId(),
            JoinType.INNER);
        AssignIpAddressFromPodVlanSearch.join("vlan", podVlanSearch, podVlanSearch.entity().getId(), AssignIpAddressFromPodVlanSearch.entity().getVlanId(), JoinType.INNER);
        AssignIpAddressFromPodVlanSearch.done();

public IPAddressVO doInTransaction(TransactionStatus status) throws InsufficientAddressCapacityException {
                StringBuilder errorMessage = new StringBuilder("Unable to get ip adress in ");
                boolean fetchFromDedicatedRange = false;
                List<Long> dedicatedVlanDbIds = new ArrayList<Long>();
                List<Long> nonDedicatedVlanDbIds = new ArrayList<Long>();

                SearchCriteria<IPAddressVO> sc = null;
                if (podId != null) {
                    sc = **AssignIpAddressFromPodVlanSearch**.create();
                    sc.setJoinParameters("podVlanMapSB", "podId", podId);
                    errorMessage.append(" pod id=" + podId);
                } else {
                    sc = AssignIpAddressSearch.create();
                    errorMessage.append(" zone id=" + dcId);
                }

rafaelweingartner:
@serg38 if that "AssignIpAddressFromPodVlanSearch" object was being used to generate the SQL; should not we see a join with "pod_vlan_map" too? For me this, this SC is very confusing.

Following the same idea of what I would do if using Spring to manage transactions, the method "fetchNewPublicIp" does not need the "@DB" annotation (assuming this is the annotation that opens a transaction and locks tables in ACS). The method “fetchNewPublicIp” is a simple "retrieve/get" method. Whenever we have to lock the table that is being used by this method, we could use the "fetchNewPublicIp" in a method that has the "@DB" annotation (assuming it has transaction propagation). This is something that already seems to happen. Methods "allocateIp" and "assignDedicateIpAddress" use “fetchNewPublicIp” and they have their own “@DB” annotation.

Methods “assignPublicIpAddressFromVlans” and “assignPublicIpAddress” seem not to do anything that requires a transaction; despite misleading (at least for me) with names indicating that something will be assigned to someone, they just call and return the response of  “fetchNewPublicIp” method. Therefore, I do not think they require a locking transaction.

serg38:
@rafaelweingartner You might be right that pod_vlan_map should be in the join. May be I didn't find the correct methods after all. @jburwell @rhtyd What do you think?

I was able to find management serve log for Deadlock 1. Looks like one of transaction came from findAndUpdateDirectAgentToLoad  method in HostDaoImpl which creates rather complex transaction:

2016-11-24 15:04:39,284 DEBUG [host.dao.HostDaoImpl] (ClusteredAgentManager Timer:ctx-a8e9449c) Resetting hosts suitable for reconnect
2016-11-24 15:04:39,320 DEBUG [db.Transaction.Transaction] (ClusteredAgentManager Timer:ctx-a8e9449c) Rolling back the transaction: Time = 36 Name =  ClusteredAgentManager Timer; called by -TransactionLegacy.rollback:879-TransactionLegacy.removeUpTo:822-TransactionLegacy.close:646-TransactionContextInterceptor.invoke:36-ReflectiveMethodInvocation.proceed:161-ExposeInvocationInterceptor.invoke:91-ReflectiveMethodInvocation.proceed:172-JdkDynamicAopProxy.invoke:204-$Proxy48.findAndUpdateDirectAgentToLoad:-1-ClusteredAgentManagerImpl.scanDirectAgentToLoad:195-ClusteredAgentManagerImpl.runDirectAgentScanTimerTask:185-ClusteredAgentManagerImpl.access$100:99
2016-11-24 15:04:39,322 ERROR [agent.manager.ClusteredAgentManagerImpl] (ClusteredAgentManager Timer:ctx-a8e9449c) Unexpected exception DB Exception on: com.mysql.jdbc.JDBC4PreparedStatement@1e58727c: SELECT host.id, host.disconnected, host.name, host.status, host.type, host.private_ip_address, host.private_mac_address, host.private_netmask, host.public_netmask, host.public_ip_address, host.public_mac_address, host.storage_ip_address, host.cluster_id, host.storage_netmask, host.storage_mac_address, host.storage_ip_address_2, host.storage_netmask_2, host.storage_mac_address_2, host.hypervisor_type, host.proxy_port, host.resource, host.fs_type, host.available, host.setup, host.resource_state, host.hypervisor_version, host.update_count, host.uuid, host.data_center_id, host.pod_id, host.cpu_sockets, host.cpus, host.url, host.speed, host.ram, host.parent, host.guid, host.capabilities, host.total_size, host.last_ping, host.mgmt_server_id, host.dom0_memory, host.version, host.created, host.removed FROM host WHERE host.resource IS NOT NULL  AND host.mgmt_server_id = 345048964870  AND host.last_ping <= 1445339907  AND host.cluster_id IS NOT NULL  AND host.status IN ('Disconnected','Down','Alert')  AND host.removed IS NULL  FOR UPDATE 
Caused by: com.mysql.jdbc.exceptions.jdbc4.MySQLTransactionRollbackException: Deadlock found when trying to get lock; try restarting transaction

Beginning of second transaction was 
SELECT host.id, host.disconnected, host.name, host.status, host.type, host.private_ip_address, host.private_mac_address, host.private_netmask, host.public_netmask, host.public_ip_address, host.public_mac_address, host.storage_ip_address, host.cluster_id, host.storage_netmask, host.storage_mac_address, host.storage_ip_address_2, host.storage_netmask_2, host.storage_mac_address_2, host.hypervisor_type, host.proxy_port, host.resource, host.fs_type, host.available, host.setup, host.resource_state, host.hypervisor_version, host.update_count, host.uuid, host.data_center_id, host.pod_id, host.cpu_sockets, host.cpus, host.url, host.speed, host.ram, host.parent, host.guid, host.capabilities, host.total_size, host.last_ping, host.mgmt_server_id, host.dom0_memory, host.version, host.created, host.removed FROM host  LEFT OUTER JOIN op_host_transfer ON host.id=op_host_transfer.id  IN

I will try to trace it to the ACS method.

serg38:
@rafaelweingartner I might be wrong but 2d  came from findAndUpdateDirectAgentToLoad in HostDaoImpl  which also creates a large transaction.

rafaelweingartner:
@serg38, it is great that you found one of the methods that cause the deadlock problem “com.cloud.host.dao.HostDaoImpl.findAndUpdateDirectAgentToLoad(long, Long, long)”.

This method surely is problematic. I would first start asking, (i) does it need to manually open a transaction (at line 512)? Isn’t that the goal of “@DB” annotation? (ii) what is the objective of the method (“findAndUpdateDirectAgentToLoad”)? It is looking too complicated, with too many accesses to the DB.

The method “resetHosts” at line 517 looks for hosts that are “managed” by the current MS and are “Disconnected” to mark them as unmanaged by any MS. That means, it updates the “managementServerId = null” of hosts marked as “Disconnect”.

Would not it be better to have a specific method/transaction only for the aforementioned process?  If we extract that chunk of code to an isolated method, could not we have an atomic access to the DB without locking? “update set managementServerId = null from hosts where ……”; If the method is isolated I do not see reasons for locks here.

A little further, there is another method which could be isolated, lines 527 – 546. This block of code looks for clusters being managed by the current MS. Then, it searches for hosts of clusters that are managed by the current MS, which are not being managed by the current MS (or not managed at all?)? I did not understand that because I have seen in some other piece of code that we have a balancing approach; meaning that, we try to balance the number of hosts managed by an MS.  This piece of code seems to remove the balancing process.

Then, at line 551 and forward (if the number of hosts is less than the limit), it tries to look for hosts of clusters not being managed by any MS. This block could also be an isolated one. And again, we might be able to do this process without using locks.

My final comment, even if we choose not to refactor and improve this piece of code, there is one thing that is very strange for me. The method “findAndUpdateDirectAgentToLoad”  is annotated with “@DB”, and also opens and tries to manage a transaction manually. Then, we have all of the pieces of code I mentioned, all of them call other methods that also are annotated with “@DB”. Can this cause a problem?

For instance, when I use Spring, methods from a service layer (the place where I configure my pattern of transactions) call one another, they will all use/share the same transaction opened when the first method of the service layer was called, unless specified otherwise. How will it work here in ACS?

serg38:
@rafaelweingartner Thanks a lot. I totally agree that resetting hosts doesn't really need to be a part of transaction and should be extracted to a new method. The same is for lines 527-546, and then another one after 551
My understanding of agent LB is that is handled separately from reconnect part. I might be wrong but it is done in ClusteredAgentManagerImpl by scheduling rebalancing task every 60 sec
getAgentRebalanceScanTask which takes care of transferring of connected agents.
@rhtyd @jburwell @koushik-das @karuturi Do you agree that we can split a transaction in findAndUpdateDirectAgentToLoad into 3 non transactional methods and thus eliminate a one side of a repeated deadlock? This is a very core of agent management that is very hard if ever possible to write smoke test. If so @nvazquez might be able to work on refactoring this method later this month

rafaelweingartner:
@serg38 I have the same understanding about the agent LB. And this is one of the problems I think we have found here. It seems that this method is removing the balance created with agent LB. And, of course, this method is also causing deadlocks.

Let’s hear the feedback from others and discuss what we can do forward.

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 464
 Hypervisor xenserver
 NetworkType Advanced
 Passed=104
 Failed=1
 Skipped=7

**Failed tests:**
* test_routers_network_ops.py

 * test_01_RVR_Network_FW_PF_SSH_default_routes_egress_true Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

yvsubhash:
@serg38 Is the refactoring suggested by rafael taken care by  @nvazquez, else I would take it up

serg38:
@yvsubhash Please, take this up. So far this PR hasn't moved forward.

yvsubhash:
@serg38 I am working on this. Will update the code once I am done

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 1130
 Hypervisor xenserver
 NetworkType Advanced
 Passed=107
 Failed=7
 Skipped=12

**Failed tests:**
* test_router_dnsservice.py

 * test_router_dns_guestipquery Failed

* test_loadbalance.py

 * test_01_create_lb_rule_src_nat Failed

* test_volumes.py

* test_routers_network_ops.py

 * test_01_isolate_network_FW_PF_default_routes_egress_true Failing since 125 runs

 * test_02_isolate_network_FW_PF_default_routes_egress_false Failing since 125 runs

 * test_01_RVR_Network_FW_PF_SSH_default_routes_egress_true Failing since 121 runs

 * test_02_RVR_Network_FW_PF_SSH_default_routes_egress_false Failing since 121 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_01_verify_libvirt
test_04_verify_guest_lspci
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vm_snapshots.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_metrics_api.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

yvsubhash:
@rhtyd  Can you please review

yvsubhash:
@serg38 @rafaelweingartner @rhtyd  @karuturi  @koushik-das  please review

rhtyd:
@yvsubhash can you comment what was changed since your initial change which got received objections. I see you're not retrying in the GenericDao layer now, can you perhaps change the PR subject/description describing how to reproduce the exceptions and how to evaluate/review your fix?

rhtyd:
@yvsubhash sure I'll include this in next round of testing. Included in 4.11 milestone for considering.

rafaelweingartner:
@yvsubhash and @rhtyd, now these changes seem to solve the problem. These synchronization and transaction manually managed are a real problem.  We should plan to move to a real JPA implementation together with a framework like Spring-data. Then, these problems are solved by the framework.

@yvsubhash do you have these changes running in production already?

yvsubhash:
@rafaelweingartner these changes are already part of our latest release and hotfixes. Already in production with some customers. In general my observation is at many places transaction are scoped with more lines of code than it is required. That in turn increases the possiblity of deadlocks, We need a way to avoid these.

rafaelweingartner:
@yvsubhash thanks for the feedback. 
I have the same feeling, too much code to handle transactions. One way to avoid this is to use proper JPA implementations, which have transaction management out of the box. It would even be better to use JPA implementation with a framework like Spring data.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

yvsubhash:
@rhtyd All the failures are due to the following error 
Private Gateway ==> Execute cmd: createprivategateway failed, due to: errorCode: 431, errorText:The VLAN tag vlan://846 is already being used for dynamic vlan allocation for the guest network in zone pr1762-t1738-kvm-centos7\n']
This is likely due to issues in cleanup in these test cases, So not related to this PR

rhtyd:
Tests LGTM, I'll merge this based on review for new/updated code and test results.

server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -836,35 +835,51 @@ public IPAddressVO doInTransaction(TransactionStatus status) throws Insufficient
...
-                                VlanVO vlan = _vlanDao.findById(addr.getVlanId());
-                                String guestType = vlan.getVlanType().toString();
-                                if (!isIpDedicated(addr)) {
-                                    UsageEventUtils.publishUsageEvent(EventTypes.EVENT_NET_IP_ASSIGN, owner.getId(), addr.getDataCenterId(), addr.getId(),
-                                            addr.getAddress().toString(),
-                                            addr.isSourceNat(), guestType, addr.getSystem(), addr.getClass().getName(), addr.getUuid());
-                                }
-                                if (updateIpResourceCount(addr)) {
-                                    _resourceLimitMgr.incrementResourceCount(owner.getId(), ResourceType.public_ip);
+                            if (_ipAddressDao.update(addr.getId(), addr)) {
```
rhtyd: @yvsubhash I've found regressions in some tests, specifically around private gw tests where failures look like this:
```
2017-12-19 22:53:00,817 DEBUG [c.c.a.t.Request] (AgentManager-Handler-20:null) (logid:) Seq 1-4815473901565903002: Processing:  { Ans: , MgmtId: 2485222984626, via: 1, Ver: v1, Flags: 10, [{"com.cloud.agent.api.routing.GroupAnswer":{"results":["null - success: Creating file in VR, with ip: 169.254.1.135, file: ip_associations.json.82a129d8-a849-4a63-af8b-3b885abd3b32","null - failed: [INFO] update_config.py :: Processing incoming file => ip_associations.json.82a129d8-a849-4a63-af8b-3b885abd3b32[INFO] Processing JSON file ip_associations.json.82a129d8-a849-4a63-af8b-3b885abd3b32Traceback (most recent call last):  File \"/opt/cloud/bin/update_config.py\", line 143, in <module>    process_file()  File \"/opt/cloud/bin/update_config.py\", line 57, in process_file    finish_config()  File \"/opt/cloud/bin/update_config.py\", line 45, in finish_config    returncode = configure.main(sys.argv)  File \"/opt/cloud/bin/configure.py\", line 1006, in main    config.address().process()  File \"/opt/cloud/bin/cs/CsAddress.py\", line 104, in process    ip = CsIP(dev, self.config)  File \"/opt/cloud/bin/cs/CsAddress.py\", line 261, in __init__    self.dnum = hex(int(dev[3:]))ValueError: invalid literal for int() with base 10: 'None'"],"result":false,"wait":0}}] }      2017-12-19 22:53:00,818 DEBUG [c.c.a.t.Request] (API-Job-Executor-47:ctx-52e4a845 job-76 ctx-b4b917c1) (logid:a5289d87) Seq 1-4815473901565903002: Received:  { Ans: , MgmtId: 2485222984626, via: 1(centos7-kvm1), Ver: v1, Flags: 10, { GroupAnswer } }
```
On deeper analysis, I found that `nic_dev_id` was not passed to a VPC router with no VMs, one network and a private way via a ip_associations.json and it caused interface to have ethNone defined in ips.json instead of eth2 etc. (this is similar to https://issues.apache.org/jira/browse/CLOUDSTACK-9759). Since, the only change in master is your PR I suspect there is some regression and that patch you've submitted here may be different that you private fork/branch. /cc @rafaelweingartner 

For example, in this code if update fails for some reason it should throw an exception or log?

server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -836,35 +835,51 @@ public IPAddressVO doInTransaction(TransactionStatus status) throws Insufficient
...
+    }
+
+    @DB
+    private void markPublicIpAsAllocating(final IPAddressVO addr) {
+        synchronized (_allocatingLock) {
+            Transaction.execute(new TransactionCallbackNoReturn() {
+                @Override
+                public void doInTransactionWithoutResult(TransactionStatus status) {
+
+                    if (_ipAddressDao.lockRow(addr.getId(), true) != null) {
```
rhtyd: Here why is an explicit locking of Row called?

server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -901,40 +916,28 @@ public PublicIp assignDedicateIpAddress(Account owner, final Long guestNtwkId, f
...
+                displayIp = ntwk.getDisplayNetwork();
+            } else if (vpcId != null) {
+                VpcVO vpc = _vpcDao.findById(vpcId);
+                displayIp = vpc.isDisplay();
+            }
 
-                    return ip;
-                }
-            });
+            return fetchNewPublicIp(dcId, null, null, owner, VlanType.VirtualNetwork, guestNtwkId, isSourceNat, false, null, false, vpcId, displayIp);
```
rhtyd: After this, we're not doing the following:
```
-                    IPAddressVO publicIp = ip.ip();
 -
 -                    markPublicIpAsAllocated(publicIp);
 -                    _ipAddressDao.update(publicIp.getId(), publicIp);
```

server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -901,40 +916,28 @@ public PublicIp assignDedicateIpAddress(Account owner, final Long guestNtwkId, f
 
         PublicIp ip = null;
         try {
-            ip = Transaction.execute(new TransactionCallbackWithException<PublicIp, InsufficientAddressCapacityException>() {
```
rhtyd: @yvsubhash this method is no longer bound in transaction and may cause failures.

rhtyd:
@yvsubhash I've left some additional remarks, please see them. After commit triaging, looks like I made an error, and it seems the issue was not caused by this commit/PR. I found the https://github.com/apache/cloudstack/pull/2304 to be an issue on KVM (in my branch this commit/pr was also included causing incorrect nic dev ids to be assigned). I'll continue testing, but do see my remarks and send another PR if you find further scope for improvements, thanks. /cc @rafaelweingartner

server/src/com/cloud/network/IpAddressManagerImpl.java
rhtyd: @yvsubhash I've found a regression with changes to this method. It causes SNAT IPs to not get allocated and we see something like this:
```
2017-12-20 14:32:40,402 ERROR [c.c.n.IpAddressManagerImpl] (API-Job-Executor-16:ctx-d4229b63 job-418 ctx-da9f0f88) (logid:8f31a2f6) Unable to get source nat ip address for account 2 
```
This causes VPC/VPN tests to fail.

rhtyd:
@yvsubhash given this has introduced a regressions in snat IP allocation I'm consider this for a revert unless you can help with a quick fix to unblock other PRs being merged.

The failure found was:
```
test_01_vpc_remote_access_vpn (integration.smoke.test_vpc_vpn.TestVpcRemoteAccessVpn): DEBUG: ========Sending GET Cmd : createRemoteAccessVpn=======
urllib3.connectionpool: DEBUG: Starting new HTTP connection (1): localhost          
urllib3.connectionpool: DEBUG: http://localhost:8080 "GET /client/api?account=test-TestRVPCSite2SiteVpn-CRVT35&domainid=838a592d-e4dc-11e7-9c0d-a434d91cd37e&fordisplay=true&publicipid=2df812b1-57a6-470a-a4db-472afed287df&response=json&apiKey=nG25osQQTwT630F9niTxS9j5qX4cVRv8PKQ80jXskQ2BmMPvTulGySL0H2sh4OjNX0SOZWcEVyMN2vkUw5lorQ&command=createRemoteAccessVpn&signature=S6MoFQwy4UKI6CbYTBffqENjVHg%3D&iprange=10.3.2.1-10.3.2.10 HTTP/1.1" 431 155
test_01_vpc_remote_access_vpn (integration.smoke.test_vpc_vpn.TestVpcRemoteAccessVpn): ERROR: Exception:['Traceback (most recent call last):\n', '  File "/usr/local/lib/python2.7/dist-packages/marvin/cloudstackConnection.py", line 308, in __parseAndGetResponse\n    response_cls)\n', '  File "/usr/local/lib/python2.7/dist-packages/marvin/jsonHelper.py", line 150, in getResultObj\n    raise cloudstackException.CloudstackAPIException(respname, errMsg)\n', 'CloudstackAPIException: Execute cmd: createremoteaccessvpn failed, due to: errorCode: 431, errorText:The Ip address is not ready to be used yet: 172.20.20.16\n']

```

The issue can be reproduced by creating a VPN and deploying a VM on a tier. You'll still see the SNAT ip marked as Allocating and this causes creation of VPN to fail.

server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -901,40 +916,28 @@ public PublicIp assignDedicateIpAddress(Account owner, final Long guestNtwkId, f
...
-                    if (guestNtwkId != null) {
-                        Network ntwk = _networksDao.findById(guestNtwkId);
-                        displayIp = ntwk.getDisplayNetwork();
-                    } else if (vpcId != null) {
-                        VpcVO vpc = _vpcDao.findById(vpcId);
-                        displayIp = vpc.isDisplay();
-                    }
+            Account ownerAccount = _accountDao.acquireInLockTable(ownerId);
 
-                    PublicIp ip = fetchNewPublicIp(dcId, null, null, owner, VlanType.VirtualNetwork, guestNtwkId, isSourceNat, false, null, false, vpcId, displayIp);
```
rhtyd: @yvsubhash The old code used to mark the new public IP as allocated, but new code does not do that here. I found a fix is to pass the assign to true. I'll fix it in my PR #2211.

rhtyd:
@yvsubhash please go through my comments and also see the regression fix. If this is not resolved earlier, we may, unfortunately, have to revert this.

yvsubhash:
@rhtyd I am going through this ll respond  in few hours

server/src/com/cloud/network/IpAddressManagerImpl.java
yvsubhash: @rhtyd  this is because bulk locking got removed earlier in the above line. so avoid two threads picking same ip
 -  addrs = _ipAddressDao.lockRows(sc, filter, true);
 + addrs = _ipAddressDao.search(sc, filter, false);

rhtyd:
@yvsubhash I read through your comment but this is not acceptable as it causes a regression. I'm not sure about the fix I did as that's more of a band-aid fix I did. Irrespective of what you suggested, the issue remains that snat IP for new vpcs stay in allocating state. Please review my fix if that is acceptable or propose a new one, since it's failing tests for others the original commit would need to be either reverted or a proper regression fix need to be done.

server/src/com/cloud/network/IpAddressManagerImpl.java
yvsubhash: @rhtyd  Passing true to fetchNewPublicIp went in as a separate commit internally and got missed out while moving to ACS.

rhtyd:
Thanks @yvsubhash for confirming about passing the assign=true.

server/src/com/cloud/network/IpAddressManagerImpl.java
rhtyd: @yvsubhash maybe re-introduce a transaction? Maybe without a locking/transaction multiple parties may receive the same public IP.

server/src/com/cloud/network/IpAddressManagerImpl.java
rhtyd: @yvsubhash as explained I could reproduce in an environment a case where both cpvm and ssvm got the same public IP. Should we add a check to make sure that you're only marking ip as allocating when they are Free?

server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -759,7 +761,7 @@ public IPAddressVO doInTransaction(TransactionStatus status) throws Insufficient
 
                 Filter filter = new Filter(IPAddressVO.class, "vlanId", true, 0l, 1l);
 
-                List<IPAddressVO> addrs = _ipAddressDao.lockRows(sc, filter, true);
+                List<IPAddressVO> addrs = _ipAddressDao.search(sc, filter, false);
```
rhtyd: The issue seems to be here where search happens without no locks and other threads may end up getting the same list of non-free ip addresses to allocate address from.

server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -769,7 +771,7 @@ public IPAddressVO doInTransaction(TransactionStatus status) throws Insufficient
                         fetchFromDedicatedRange = false;
                         sc.setParameters("vlanId", nonDedicatedVlanDbIds.toArray());
                         errorMessage.append(", vlanId id=" + Arrays.toString(nonDedicatedVlanDbIds.toArray()));
-                        addrs = _ipAddressDao.lockRows(sc, filter, true);
+                        addrs = _ipAddressDao.search(sc, filter, false);
```
rhtyd: Without locking, from the list only first ip address is consumed by `IPAddressVO addr = addrs.get(0)` which is the source of problem.

server/src/com/cloud/network/IpAddressManagerImpl.java
rhtyd: @yvsubhash can you check the changes in your PR against others in your internal/private fork, it seems there are still issues. Alternatively, we can revert this fix and wait for your complete proper PR.

rhtyd:
@yvsubhash I hit another corner case today wrt testing #2211 when both SSVM and CPVM got the same public IP, since it's likely related to your fix please do an code/investigation if you've missed any case? /cc @DagSonstebo @borisstoyanov  I've another regression patch/fix on #2211 please review.

PR #1765, opened 2016-11-16
Cloudstack 9586: When using local storage with Xenserver prepareTemplate does not work with multiple primary store

abhinandanprateek:
The race condition will happen whenever there are multiple primary storages and the CS tries to mount the secondary store to xenserver host simultaneously.

Due to synchronised block one mount will be successful and other thread will get the already mounted SR. Without the fix the two thread will try to mount it parallely and one will fail on Xenserver.

### DIFF
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
@@ -101,35 +101,24 @@ public class Xenserver625StorageProcessor extends XenServerStorageProcessor {
 
         try {
             final String srname = hypervisorResource.getHost().getUuid() + path.trim();
-
-            final Set<SR> srs = SR.getByNameLabel(conn, srname);
-
-            if (srs != null && !srs.isEmpty()) {
-                return srs.iterator().next();
+            synchronized (srname.intern()) {
+                final Set<SR> srs = SR.getByNameLabel(conn, srname);
+                if (srs != null && !srs.isEmpty()) {
+                    return srs.iterator().next();
+                }
+                final Map<String, String> smConfig = new HashMap<String, String>();
+                final Host host = Host.getByUuid(conn, hypervisorResource.getHost().getUuid());
+                final String uuid = UUID.randomUUID().toString();
+                sr = SR.introduce(conn, uuid, srname, srname, "file", "file", false, smConfig);
+                final PBD.Record record = new PBD.Record();
+                record.host = host;
+                record.SR = sr;
+                smConfig.put("location", path);
+                record.deviceConfig = smConfig;
+                pbd = PBD.create(conn, record);
+                pbd.plug(conn);
+                sr.scan(conn);
             }
-
-            final Map<String, String> smConfig = new HashMap<String, String>();
-
-            final Host host = Host.getByUuid(conn, hypervisorResource.getHost().getUuid());
-            final String uuid = UUID.randomUUID().toString();
-
-            sr = SR.introduce(conn, uuid, srname, srname, "file", "file", false, smConfig);
-
-            final PBD.Record record = new PBD.Record();
-
-            record.host = host;
-            record.SR = sr;
-
-            smConfig.put("location", path);
-
-            record.deviceConfig = smConfig;
-
-            pbd = PBD.create(conn, record);
-
-            pbd.plug(conn);
-
-            sr.scan(conn);
-
             return sr;
         } catch (final Exception ex) {
             try {


jburwell:
@abhinandanprateek could you please squash the commits for this PR?

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -100,36 +100,25 @@ protected SR createFileSR(final Connection conn, final String path) {
...
 
         try {
-            final String srname = hypervisorResource.getHost().getUuid() + path.trim();
-
-            final Set<SR> srs = SR.getByNameLabel(conn, srname);
-
-            if (srs != null && !srs.isEmpty()) {
-                return srs.iterator().next();
+            final String srname = path.trim();
+            synchronized (srname.intern()) {
```
jburwell: Should we consider using a `ReentrantLock` instead of `synchronized`?  `ReentrantLock.tryLock(long, TimeUnit)` allows the specification of a timeout to acquire the lock.  Since a timeout cannot be specified to `synchronized`, multiple threads may deadlock attempting to acquire the lock in the case of a stall.  The downside of this approach is that we would need to track locks in a `ConcurrentMap` -- ideally with WeakReferences to allow them to be garbage collected when no threads are operating on them.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -100,36 +100,25 @@ protected SR createFileSR(final Connection conn, final String path) {
...
+                final String uuid = UUID.randomUUID().toString();
+                sr = SR.introduce(conn, uuid, srname, srname, "file", "file", false, smConfig);
+                final PBD.Record record = new PBD.Record();
+                record.host = host;
+                record.SR = sr;
+                smConfig.put("location", path);
+                record.deviceConfig = smConfig;
+                pbd = PBD.create(conn, record);
+                pbd.plug(conn);
+                sr.scan(conn);
```
jburwell: Some of the work in this `synchronized` block does not appear to need a lock to be successful.  Please consider refactoring to move that work out of this block to reduce the size of the critical section.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
abhinandanprateek: All the calls touching external resource in the synchronized block are protected by a timeout. I think it will be redundant to add another layer of timeout on top of it. Even in case of this additional timeout there is no guarantee that the acquiring thread will be successful but will fail in all probability.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
abhinandanprateek: The block as a whole needs synchronization as at the start presence of a SR is checked and the last call in the block creates one on the xen pool.

rhtyd:
@blueorangutan package

rhtyd:

rhtyd:
LGTM, last test run looks okay. Failed tests around vpc/rvr are known intermittent issues.
@abhinandanprateek can you squash your commits, thanks.

abhinandanprateek:
@rhtyd squashed

rhtyd:
@blueorangutan package

rhtyd:
LGTM.
@abhinandanprateek @murali-reddy @jburwell are we lgtm on this PR? 

rhtyd:
@blueorangutan test

rhtyd:

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -100,36 +100,25 @@ protected SR createFileSR(final Connection conn, final String path) {
         PBD pbd = null;
 
         try {
-            final String srname = hypervisorResource.getHost().getUuid() + path.trim();
-
-            final Set<SR> srs = SR.getByNameLabel(conn, srname);
-
-            if (srs != null && !srs.isEmpty()) {
-                return srs.iterator().next();
+            final String srname = path.trim();
```
rhtyd: @abhinandanprateek `srname` is drops `hypervisorResource.getHost().getUuid()`, see line 103.
Should it be:
final String srname = hypervisorResource.getHost().getUuid() + path.trim() ?

rhtyd:
@murali-reddy @abhinandanprateek @jburwell pinging for review continuation.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
abhinandanprateek: @rhtyd, since sr is unique on a pool, adding hostname to its path does not give any advantage. Usually MS will go thru the pool master so effectively it will end up the pool master uuid to sr name.
This could have avoided the race condition by probably introducing redundant SRs, but now that the race condition is dealt with we really donot need the host uuid qualification of SR name.

rhtyd: @abhinandanprateek could there be a case that after upgrade SR.getByNameLabel will fail as the srname has changed? I would say let's not break the naming convention if it's used in/by external resources.

rhtyd:
@blueorangutan package

rhtyd:

jburwell:
@abhinandanprateek could you please rebase this PR to pick up the fixes to the broken tests?  I want to make sure that this PR doesn't introduce any side-effects that break the test runs.

abhinandanprateek:
@jburwell rebased.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
rhtyd: @abhinandanprateek ping, do you think changing the SR name can cause upgrade issues?

rhtyd:
@blueorangutan package

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
abhinandanprateek: The SR for a template that is to be copied to primary is temporarily mounted on Xen, after the copy is complete, it is unmounted and SR forgotten on Xen. The SR is temporary and for the current operation only. Having said that it does not really make a difference. @rhtyd

rhtyd:
Test lgtm, based on the test results I'm seeing no new failures.

abhinandanprateek:
@rhtyd I have reverted to the sr name as it was.

rhtyd:
@blueorangutan package

rhtyd:

rhtyd:
LGTM, tests confirm no regressions. Test LGTM.
Merging this now.

PR #1767, opened 2016-11-18
CLOUDSTACK-9457: Allow retrieval and modification of VM and template details via API and UI

nvazquez:
JIRA TICKET: https://issues.apache.org/jira/browse/CLOUDSTACK-9457

### Goal
This PR proposes list/add/update/delete user vm and vm template details via API and UI.

### VM UI Screenshots
Setting tab is added on Instances page. Actions allowed are: Add/Edit/Remove

Settings tab is only shown if instance is Stopped:

### Templates UI Screenshots
Setting tab is added on Templates page. Actions allowed are: Add/Edit/Remove:

### DIFF
--- a/api/src/org/apache/cloudstack/api/ApiConstants.java
+++ b/api/src/org/apache/cloudstack/api/ApiConstants.java
@@ -648,6 +648,7 @@ public class ApiConstants {
     public static final String OVM3_POOL = "ovm3pool";
     public static final String OVM3_CLUSTER = "ovm3cluster";
     public static final String OVM3_VIP = "ovm3vip";
+    public static final String CLEAN_UP_DETAILS = "cleanupdetails";
 
     public static final String ADMIN = "admin";
 
--- a/api/src/org/apache/cloudstack/api/BaseUpdateTemplateOrIsoCmd.java
+++ b/api/src/org/apache/cloudstack/api/BaseUpdateTemplateOrIsoCmd.java
@@ -17,7 +17,6 @@
 package org.apache.cloudstack.api;
 
 import org.apache.log4j.Logger;
-
 import org.apache.cloudstack.api.command.user.iso.UpdateIsoCmd;
 import org.apache.cloudstack.api.response.GuestOSResponse;
 import org.apache.cloudstack.api.response.TemplateResponse;
@@ -72,6 +71,11 @@ public abstract class BaseUpdateTemplateOrIsoCmd extends BaseCmd {
 
     @Parameter(name = ApiConstants.DETAILS, type = CommandType.MAP, description = "Details in key/value pairs using format details[i].keyname=keyvalue. Example: details[0].hypervisortoolsversion=xenserver61")
     protected Map details;
+
+    @Parameter(name = ApiConstants.CLEAN_UP_DETAILS,
+            type = CommandType.BOOLEAN,
+            description = "optional boolean field, which indicates if details should be cleaned up or not (if set to true, details removed for this resource, details field ignored; if false or not set, no action)")
+    private Boolean cleanupDetails;
 
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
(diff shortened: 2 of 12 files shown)


rhtyd:
@blueorangutan package

koushik-das:
@nvazquez Can you add use case details that will be addressed by these new methods? Also what kind of validation will be performed on the inputs and the state of the entity (VM or template)? Will it be metadata for book keeping only or will impact the entity during runtime? I am assuming that data will be stored in the corresponding *details table in DB. How will these APIs impact already existing data that are created using some other APIs and stored in the details table?

ustcweizhou:
I have the same concerns as @koushik-das .

nvazquez:
Hi @koushik-das @ustcweizhou,
For this new methods as they require the entity id (VM or template) I added a basic validation to check if the entity is found on DB, but I realize that I should also validate that entity is not destroyed. This data is to be stored in *details tables.
About existing API methods, I have checked `updateTemplate` and `updateVirtualMachines` methods before adding this new methods, and they provide a way to add/update details (although updateTemplate overrides existing details). Also, details can be listed on `listTemplates` and `listVirtualMachines` but thought the best approach was introducing this new methods to reduce overhead and treat entity details separately instead of updating/listing the entity every time. What do you think of this approach?

serg38:
@koushik-das @ustcweizhou Just want to add on the use case. There is growing need to change template/VM details for after the deployment e.g. to switch root disk controller or change ratio of cores/socket with cpuid.coresPerSocket or with adding vGPU startup parameters or switch boot firmware from bios to efi as required by some OS. Currently users are unable to easily do it with API and/or UI. As implemented now, all settings with display=1 in user_vm_details and vm_template_details will be returned by API and allowed to changed. If desired we can open another PR to switch details created by other API in automatic fashion to have display=0 so they will not show up in the response of these APIs and users wont be able to edit/delete them. We thought this to be out of scope for now.

koushik-das:
@serg38 @nvazquez Thanks for the update. As I understand form some of the detail parameter examples, these are tied to the lifecycle of the entity (create/destroy or start/stop of the VM). If thats the case then it make sense to pass on these details along with the corresponding lifecycle APIs rather than creating new APIs. Do you see any use case where the details are not tied to entity lifecycle?

serg38:
@koushik-das No I don't. I tend to agree with you but obviously the implementation will be messier since configs are random. @nvazquez Do you think it can be changes so that modification  is done via detail tag of updateVirtuaMachine and updateTemplate APIs.

nvazquez:
@serg38 @koushik-das sure, I can work on it, the actual way is passing details as key value pairs on the update methods but I think it would require to define a way to delete details (for example the actual behaviour of `updateTemplates` overrides any existing detail for the ones provided under `details` field while on `updateVirtualMachines` they are added to the existing ones, not allowing to delete any detail)

serg38:
@nvazquez @koushik-das If we go this route we might want to change update API  to behave consistently. It would make sense to allow overriding any details.

nvazquez:
Hi @rhtyd @koushik-das @ustcweizhou,

I refactored this PR removing previously created API commands, only adding an optional parameter (`cleanupdetails`) on `updateVirtualMachine` and `updateTemplate` commands to allow cleaning up details. Besides that change, add/edit/remove are now performed by `updateVirtualMachine` or `updateTemplate` as you suggested. I'll post screenshots on Description to show changes on the UI.

Happy new year!

serg38:
@rhtyd Can you kick off tests for this PR?

rhtyd:
@serg38 sure
@blueorangutan package

rhtyd:
@blueorangutan test

serg38:
@karuturi Can we  merge this one ?

serg38:
LGTM

rafaelweingartner:
I have just reviewed the PR. There are only small suggestions I made.

BTW: from the Jira ticket https://issues.apache.org/jira/browse/CLOUDSTACK-9457, I ended up at https://issues.apache.org/jira/browse/CLOUDSTACK-9379, which is already done. I do not have permission to mark it as done. I believe @nvazquez might have.

nvazquez:
@karuturi @rafaelweingartner thanks! I closed Jira ticket 9379

PR #1771, opened 2016-11-23
CLOUDSTACK-9611: Dedicating a Guest VLAN range to Project does not work.

nitin-maharana:
Description:
=========

Trying to dedicate a guest VLAN range to an account fails. Either API documentation is wrong or code path needs to be corrected. If we pass both account and projectId parameters to the dedicateGuestVlanRange (which are not mentioned as mutually exclusive in API description) the API layer throws error saying both are mutually exclusive.

Steps to Reproduce:
===============
Create an account. Create a project in that account.

Go to Admin account and change the view to the above project.

Navigate to Infrastructure -> Zone -> Physical Network -> Guest -> Dedicate Guest VLAN range.

Try to dedicate the guest VLAN range from the project view for the account associated with the project.

It fails with error saying "accountName and projectId are mutually exclusive".

image

image

Expected:
=======
The VLAN range should get dedicated to the project account.

image

image

Resolution:
========
Changed the account field of API to optional.
Even if the account name is passed while adding from a particular project view, then it is ignored in API.

Notes:
=====
If we do the dedication from default view then it works fine as no projectId is associated over there.

### DIFF
--- a/ui/scripts/system.js
+++ b/ui/scripts/system.js
@@ -2090,27 +2090,57 @@
                                                 fields: {
                                                     vlanrange: {
                                                         label: 'label.vlan.vni.range',
-                                                        /*  select: function(args) {
-                                                        var items = [];
-                                                        if(args.context.physicalNetworks[0].vlan != null && args.context.physicalNetworks[0].vlan.length > 0) {
-                                                        var vlanranges = args.context.physicalNetworks[0].vlan.split(";");
-                                                        for(var i = 0; i < vlanranges.length ; i++) {
-                                                        items.push({id: vlanranges[i], description: vlanranges[i]});
-                                                        }
-                                                        }
-                                                        args.response.success({data: items});
-                                                        },*/
                                                         validation: {
                                                             required: true
                                                         }
                                                     },
-                                                    account: {
-                                                        label: 'label.account',
-                                                        validation: {
-                                                            required: true
+                                                    scope: {
+                                                        label: 'label.scope',
+                                                        docID: 'helpGuestNetworkZoneScope',
+                                                        select: function(args) {
+                                                            var array1 = [];
+
+                                                            array1.push({
+                                                                id: 'account-specific',
+                                                                description: 'label.account'
+                                                            });
+                                                            array1.push({
+                                                                id: 'project-specific',
+                                                                description: 'label.project'
+                                                            });
+
+                                                            args.response.success({
+                                                                data: array1
+                                                            });
+
+                                                            args.$select.change(function() {
+                                                                var $form = $(this).closest('form');
+
+                                                                if ($(this).val() == "account-specific") {
+                                                                    $form.find('.form-item[rel=domainId]').css('display', 'inline-block');
+                                                                    $form.find('.form-item[rel=account]').css('display', 'inline-block');
+                                                                    $form.find('.form-item[rel=projectId]').hide();
+                                                                } else if ($(this).val() == "project-specific") {
+                                                                    $form.find('.form-item[rel=domainId]').css('display', 'inline-block');
+                                                                    $form.find('.form-item[rel=account]').hide();
+                                                                    $form.find('.form-item[rel=projectId]').css('display', 'inline-block');
+                                                                }
+
+                                                                if (args.context.projects != null && args.context.projects.length > 0) {
+                                                                    $form.find('.form-item[rel=domainId]').hide();
+                                                                    $form.find('.form-item[rel=account]').hide();
+                                                                    $form.find('.form-item[rel=projectId]').hide();
+                                                                }
+                                                            });
+                                                        },
+                                                        isHidden: function(args) {
+                                                            if(args.context.projects != null && args.context.projects.length > 0)
+                                                                return true;
+                                                            else
+                                                                return false;
                                                         }
                                                     },
-                                                    domainid: {
+                                                    domainId: {
                                                         label: 'label.domain',
                                                         validation: {
                                                             required: true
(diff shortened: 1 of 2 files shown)


rajesh-battala:
Gone through the patch. LGTM.

rhtyd:
@nitin-maharana do you think this will be useful for 4.9? Can you change PR's base branch to 4.9?

nitin-maharana:
@rhtyd : Changed PR's base branch to 4.9

rhtyd:
@blueorangutan package

nitin-maharana:
ping @karuturi @sateesh-chodapuneedi

server/src/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -3085,9 +3085,10 @@ public GuestVlan dedicateGuestVlanRange(DedicateGuestVlanRangeCmd cmd) {
         // Verify account is valid
         Account vlanOwner = null;
         if (projectId != null) {
-            if (accountName != null) {
-                throw new InvalidParameterValueException("accountName and projectId are mutually exclusive");
-            }
+            //accountName and projectId are mutually exclusive
```
koushik-das: @nitin-maharana If account and project are mutually exclusive why have you removed the validation?

nitin-maharana: @koushik-das : As I mentioned in resolution part, Even if the account name is passed while adding from a particular project view, then it is ignored(Removal part) in API. If we don't remove that validation check, then the same issue will happen again.

koushik-das: @nitin-maharana So as I understand in order to avoid making a breaking API change you are simply ignoring account and giving priority to project if both are specified.

nitin-maharana: @koushik-das : Yes, We give priority to the project, if not specified then checking the account.

ustcweizhou: @nitin-maharana I disagree this change. We'd better keep like it.
I suggest to make ui change to add a new field 'scope' which can be set to 'domain', 'account', 'project'.
It is similar to the scope field in the dialog to create a shared network in advanced zone.

scope='Domain', domain field is shown.
scope='Account, domain and account are shown
scope='Project', domain and project are shown.

nitin-maharana: @koushik-das @rajesh-battala : Do you agree with @ustcweizhou suggestion?

koushik-das: @nitin-maharana @ustcweizhou The scope field in UI should be fine. But since there is a breaking change (accountName is made optional from required), there may be scenarios which will break. As long as it is documented should be ok.

nitin-maharana: @koushik-das : If we go with the scope field, we have to revert that change and have to make required depending on the scope we select. Like shown in the below snapshot.

screen shot 2017-02-22 at 6 14 19 pm

nitin-maharana:
@ustcweizhou @koushik-das : I implemented the way @ustcweizhou's suggested. It was purely an UI change. But I found an issue while testing.

On API side, the entire logic uses VLAN owner account details, they find it from the account given by the user. So, they made account field as mandatory. But that contradicts when it is accessed from a project view(This issue). 

The current logic works in a default view because we don't add projectid details while accessing the API. But when it is accessed from a particular project view, by default the projectid parameter is added with the API call, So this issue arises. 

Now, if we go with the change suggested by @ustcweizhou, once we select the scope as domain, we don't pass any account details, so the logic can't able to find the owner hence gives an error.

According to my suggestion, On API side, we should make the account field as optional and on UI front, we should make it mandatory and the account field will appear only in case of default view, which will fix the issue. Please give your input on this. Accordingly, I will proceed.

Please see the current change(I pushed it here)

ustcweizhou:
@nitin-maharana you may add "ignoreProject: true" in createURL('dedicateGuestVlanRange') if scope is set to Domain

nitin-maharana:
@ustcweizhou : Passing project with domain won't create any issue but not passing any account details when we call the API from default view creates an issue. Because the internal logic expects account owner of VLAN when dedicated from the default view. In the case of project view, it gets the account detail from selected project.

ustcweizhou:
@nitin-maharana I thought dedicating guest vlan range is already implemented in CLOUDSTACK-8958.
Actually they are two stories.
In this case, we have to remove the Domain scope from UI as it is not supported, unless you make more changes to support it (similar to 37301ed4540450c29be4f975d58b38dbeec6c296 for CLOUDSTACK-8958).

nitin-maharana:
@ustcweizhou : You are correct. We have to remove the domain scope. If we support the domain scope, we need more changes on the logic. In that case we can create a new ticket for that.

nitin-maharana:
@ustcweizhou : I have modified the code, can you review it one more time. Please see the below snapshot.

screen shot 2017-03-01 at 3 48 10 pm

ustcweizhou:
@nitin-maharana should projectId dependsOn domainId ? If the domain changes, the project list should change according to the new domainid.

nitin-maharana:
@ustcweizhou : Yes, if domain changes, the project list should change. 
As per my understanding, a project is similar to an account. When we create a project, by default it is always created in ROOT domain. But an account can be created in a specific domain. I am still not clear what is the need of project.

As all the projects are under ROOT domain, if we select any domain other than ROOT, the project list would be empty. So I think we needn't use the domain list in project scope and instead we should list all projects.

nitin-maharana:
Hi @ustcweizhou, Does it look good to you?

ustcweizhou:
@nitin-maharana I think it is better to list projects by domain.
Just an example, if there are 3 projects with same name created in 3 domains, it will be difficult to choose.

ustcweizhou:
@nitin-maharana cannot we dedicate guest vlan range to projects in other domains ?

nitin-maharana:
@ustcweizhou : The createProject API allows to created a project under a domain but no idea if there is any support from UI. Whenever I create a project, it always goes to ROOT domain.

nitin-maharana:
@ustcweizhou : Please verify, I have changed according to your suggestion.

nitin-maharana:
ping @rhtyd @karuturi

ustcweizhou:
@nitin-maharana 
if this is performed in project view, should projectid be passed ?
projectId: args.context.projects[0].id

nitin-maharana:
@ustcweizhou : In case of project view, the projectId is always passed. Also we consider it as project scope but we won't have any option to choose projects from a list because we are already under a project.

ui/scripts/system.js
```diff
@@ -2090,27 +2090,57 @@
...
+                                                                    $form.find('.form-item[rel=domainId]').css('display', 'inline-block');
+                                                                    $form.find('.form-item[rel=account]').hide();
+                                                                    $form.find('.form-item[rel=projectId]').css('display', 'inline-block');
+                                                                }
+
+                                                                if (args.context.projects != null && args.context.projects.length > 0) {
+                                                                    $form.find('.form-item[rel=domainId]').hide();
+                                                                    $form.find('.form-item[rel=account]').hide();
+                                                                    $form.find('.form-item[rel=projectId]').hide();
+                                                                }
```
nitin-maharana: @ustcweizhou : The above code handles my previous explanation.

ustcweizhou:
@nitin-maharana yes, you are right. I forgot it. 
LGTM

nitin-maharana:
tag:mergeready

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 468
 Hypervisor xenserver
 NetworkType Advanced
 Passed=105
 Failed=0
 Skipped=7

**Failed tests:**

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_routers_network_ops.py
test_disk_offerings.py

PR #1772, opened 2016-11-24
CLOUDSTACK-9627 Fix template sync for region store.

syed:
When using a region store like Swift or S3 as secondary storage,
the `zoneId` can be null. This causes an exception when we try
to convert it to a `long`. This fix guards against that.

Before this fix, if you restart the management server, all the templates
would change to "NOT READY" because the code which syncs the NFS cache
and the object store crashes due to the above mentioned issue.
This PR fixes that.

### DIFF
--- a/engine/storage/image/src/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
+++ b/engine/storage/image/src/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
@@ -18,64 +18,17 @@
  */
 package org.apache.cloudstack.storage.image;
 
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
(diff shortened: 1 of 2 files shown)


rhtyd:
@syed can you add a JIRA id for this and use that in the git commit summary.
LGTM.

syed:
Thanks @rhtyd . I've created an issue in JIRA and updated the summary.

rhtyd:
@syed can you amend the git commit message with this format to include the JIRA id: `CLOUDSTACK-xxx: one line summary`?

syed:
@rhtyd Done!

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
LGTM, tests and review LGTM (one failing VPC/rvr is known intermittent failure and can be ignored). Merging this now. Thanks @syed

PR #1773, opened 2016-11-24
CLOUDSTACK-9607: Preventing template deletion when template is in use.

priyankparihar:
For complete description please refer -> [CLOUDSTACK-9607](https://issues.apache.org/jira/browse/CLOUDSTACK-9607)

### DIFF
--- a/api/src/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
@@ -52,6 +52,9 @@ public class DeleteTemplateCmd extends BaseAsyncCmd {
     @Parameter(name = ApiConstants.ZONE_ID, type = CommandType.UUID, entityType = ZoneResponse.class, description = "the ID of zone of the template")
     private Long zoneId;
 
+    @Parameter(name = ApiConstants.FORCED, type = CommandType.BOOLEAN, required = false, description = "Force delete a template.", since = "4.9+")
+    private Boolean forced;
+
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
     /////////////////////////////////////////////////////
@@ -64,6 +67,9 @@ public class DeleteTemplateCmd extends BaseAsyncCmd {
         return zoneId;
     }
 
+    public boolean isForced() {
+        return (forced != null) ? forced : true;
+    }
     /////////////////////////////////////////////////////
     /////////////// API Implementation///////////////////
     /////////////////////////////////////////////////////
--- a/server/src/com/cloud/template/TemplateManagerImpl.java
+++ b/server/src/com/cloud/template/TemplateManagerImpl.java
@@ -38,6 +38,7 @@ import com.cloud.utils.EncryptionUtil;
 import com.cloud.utils.DateUtil;
 import com.cloud.utils.Pair;
 import com.cloud.utils.EnumUtils;
+import com.google.common.base.Joiner;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 
@@ -1233,6 +1234,19 @@ public class TemplateManagerImpl extends ManagerBase implements TemplateManager,
         VMTemplateVO template = _tmpltDao.findById(templateId);
         if (template == null) {
             throw new InvalidParameterValueException("unable to find template with id " + templateId);
+        }
+
+        List<VMInstanceVO> vmInstanceVOList;
+        if(cmd.getZoneId() != null) {
+            vmInstanceVOList = _vmInstanceDao.listNonExpungedByZoneAndTemplate(cmd.getZoneId(), templateId);
+        }
+        else {
+            vmInstanceVOList = _vmInstanceDao.listNonExpungedByTemplate(templateId);
+        }
+        if(!cmd.isForced() && CollectionUtils.isNotEmpty(vmInstanceVOList)) {
+            final String message = String.format("Unable to delete template with id: %1$s because VM instances: [%2$s] are using it.",  templateId, Joiner.on(",").join(vmInstanceVOList));
+            s_logger.warn(message);
+            throw new InvalidParameterValueException(message);
         }
 
         _accountMgr.checkAccess(caller, AccessType.OperateEntry, true, template);
--- a/engine/schema/src/com/cloud/vm/dao/VMInstanceDao.java
+++ b/engine/schema/src/com/cloud/vm/dao/VMInstanceDao.java
@@ -54,6 +54,14 @@ public interface VMInstanceDao extends GenericDao<VMInstanceVO, Long>, StateDao<
     List<VMInstanceVO> listByPodId(long podId);
 
     /**
+     * Lists non-expunged VMs by  templateId
+     * @param templateId
+     * @return list of VMInstanceVO deployed from the specified template, that are not expunged
+     */
+    public List<VMInstanceVO> listNonExpungedByTemplate(long templateId);
+
+
+    /**
      * Lists non-expunged VMs by zone ID and templateId
      * @param zoneId
      * @return list of VMInstanceVO in the specified zone, deployed from the specified template, that are not expunged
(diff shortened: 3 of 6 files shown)


server/src/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -1176,6 +1176,23 @@ public boolean deleteTemplate(DeleteTemplateCmd cmd) {
...
+            vmInstanceVOList = _vmInstanceDao.listNonExpungedByTemplate(templateId);
+        }
+        if(!cmd.isForced() && CollectionUtils.isNotEmpty(vmInstanceVOList)) {
+            StringBuilder s  = new StringBuilder("Unable to delete template with id: " + templateId + " because some VM instances are using it. ");
+            for (VMInstanceVO elm : vmInstanceVOList) {
+                s.append(elm.getInstanceName() + ", ");
+            }
+
+            s_logger.warn(s.substring(0,s.length()-2));
+            throw new InvalidParameterValueException(s.substring(0,s.length()-2));
```
jburwell: Lines 1187-1193 should replaced with the following to be DRY and improve clarity:
```java
final String message = String.format("Unable to delete template with id: %1$s because VM instances: [%2$s] are using it.",  templateId, Joiner.on(",").join(vmInstanceVOList));
s_logger.warn(message);
throw new InvalidParameterValueException(message);
```

api/src/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
```diff
@@ -52,6 +52,9 @@
     @Parameter(name = ApiConstants.ZONE_ID, type = CommandType.UUID, entityType = ZoneResponse.class, description = "the ID of zone of the template")
     private Long zoneId;
 
+    @Parameter(name = ApiConstants.FORCED, type = CommandType.BOOLEAN, required = false, description = "Force delete a template.")
```
jburwell: Please add `since` to this annotation to indicate that this parameter is available in 4.9+.

priyankparihar:
Hi @jburwell ,
Thanks for teaching me something new. Test is also added.

ustcweizhou:
@priyankparihar 
1. the file change from 10644 to 10755 is not necessary
2. the default value of forced is false, might cause issue on backwards compatibility
3. In the vm installation, as the template is still in use , so we can find it in the storage pool where the root disk is located. We need to improve it by using the existing template image on storage pool, or copying the template from storage pool where the root disk is located to another new storage pool (pool of allocated new root disk).

priyankparihar:
Hi @jburwell , 
Would like to add something on @ustcweizhou comment ?

serg38:
@priyankparihar I agree with @ustcweizhou . Default behavior should remain forced. And in this case Web UI we should give a warning with "yes and no" if template has deployed VM so users can cancel the operation if they need to.

serg38:
@priyankparihar @ustcweizhou There seems to be just one case where it will affect resetVM operations. If template is deleted and root disk is migrated to another PS where there is no template copy at destination and the reset VM wont work. It is probably not easy to implement copying a base template from on PS to another as a part of migrate volume command. I think the right approach would be just to have a check in resetVM and if no local template copy is available then find another PS with template copy and reset root disk over there or if at that point there is no PS with template copy  gracefully error out resetVM command.

ustcweizhou:
@serg38 yes, your approach is the best solution for now.

By the way, I've implemented the copy of template between storage pools on kvm/nfs in our branch, not implemented for other situations yet. The missing template also impact storage migration (live/offline).

borisstoyanov:
I agree with @serg38 and @ustcweizhou that upon deletion user should be displayed with an yes/no dialogue if there are deployed VMs, but a lot of CS user aren't using UI, but just the APIs. If the force option is enabled by default I guess the API call will go and delete the all selected templates without any warning. I think the `force=true` should only stay with the UI call.

priyankparihar:
Hi @ustcweizhou @serg38 and @borisroman,
Thanks for your valuable suggestions. 

> @ustcweizhou  the default value of forced is false, might cause issue on backwards compatibility.

If the force option is enabled by default API call will go and delete the all selected templates without any warning so default value of forced is kept 'false'.

> @ustcweizhou In the vm installation, as the template is still in use , so we can find it in the storage pool where the root disk is located. We need to improve it by using the existing template image on storage pool, or copying the template from storage pool where the root disk is located to another new storage pool (pool of allocated new root disk).

I think it will be good to take it in separate PR.

ustcweizhou:
@priyankparihar Then we need to notify hundreds of users to modify their application to add forced=true in deleteTemplate API (Otherwise their application might fail), which is not feasible I think.
You also need to modify your API description to add the default value, and modify UI as well.

priyankparihar:
Hi @ustcweizhou,
By default enabling force option(force=true) will be very risky, same is mentioned by @borisstoyanov.

ustcweizhou:
@priyankparihar the default behaviour is forced=true from few years ago. If the template is removed, the user vm is not impacted which is very good. There is indeed an issue when reinstall the vm. For you it could be a very risky issue. For us at least it is not a big issue as we provide some other templates for vm reinstallation.

serg38:
@ustcweizhou @priyankparihar @borisstoyanov The default behavior was always equivalent of force=true. We seem to agree that the UI behavior should remain forced=true but with extra warning. We can either change default  confirmation "Please confirm that you want to delete this template." e.g. to "Please confirm that you want to delete this template.  Features that depends on the template presence might no longer work". But I think the better option will be to have response tag in listTemplate call to indicate if active VMs exists or not. Then warning can be given only to such templates.
As for API side we might still change default behavior but we need to explicitly document it in Release Notes so users can get prepared and change instances of deleteTemplate execution to use forced=true or not as needed.

borisstoyanov:
@blueorangutan package

priyankparihar:
Hi @koushik-das  @rajesh-battala ,

>The default value of forced is false, might cause issue on backwards compatibility.

Should i make changes according to @serg38 and @ustcweizhou suggestion ? ( or current code LGT you, Please notify. )

jburwell:
@priyankparihar I agree with @ustcweizhou regarding the default value of `forced` in terms of backwards compatibility.

Also, why we permit deletion of a template when it is associated with one or more active volumes?  It seems like we are giving the user the means to corrupt their system.

serg38:
@jburwell That was default behavior for few years to allow deletion of the template even if active VMs exist. Deletion of the template on secondary doesn't remove the template copy on primary storage so all existing VM function work just fine. 
From my prospective if we allow forced deletion from the UI I am fine with switching default to forced=no and documenting it in Release Notes.

koushik-das:
I think changing the default API behaviour for fixing a bug should be ok as long as documented properly. Also in this case the deletion criteria is made more strict, so there is no unwanted data loss.

priyankparihar:
Hi @serg38 @ustcweizhou @koushik-das @jburwell @borisstoyanov ,
Thanks for valuable insight. Code is modified according to your suggestion. :)

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

serg38:
@priyankparihar A part is still missing  that will use forced option in UI but give a user option to cancel deletion if active VMs exist

priyankparihar:
@serg38  
 
>A part is still missing that will use forced option in UI but give a user option to cancel deletion if active VMs exist.

I think it will be good to take all additional changes in separate PR.

serg38:
@priyankparihar Well. API use cases can be recoded but not UI. Presently we can delete templates in UI even with active VMs. With your changes there will be no recourse.  An easy option will be to   add a checkbox 'Forced' into Delete template confirmation dialog. I am -1 until UI supports deletion of templates with active VMs.

priyankparihar:
@borisroman and @serg38 Sure.

priyankparihar:
Hi @koushik-das, @borisroman,  @serg38,  @jburwell and @ustcweizhou
Things are modified and added according to your suggestions. I think, Now LGTM comment should be provided.

serg38:
Thanks @priyankparihar . We verified in our environment that all is working fine  including UI part.
LGTM
I think this one is 
tag:mergeready

ustcweizhou:
LGTM

priyankparihar:
@karuturi This seems to be ready for merge (tag:mergeready).
It has 3 LGTMs and tests are passed.

SudharmaJain:
@priyankparihar This PR will break the reset VM feature. I think we should also prevent reset vm if the template is deleted.

priyankparihar:
Hi @SudharmaJain, 

>This PR will break the reset VM feature. I think we should also prevent reset vm if the template is deleted.

'Force' option is introduced to solve above mentioned problem.

rhtyd:
@blueorangutan package

rhtyd:
@priyankparihar fix conflicts please.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Merging this based on code review lgtms and test results.

PR #1776, opened 2016-11-24
CLOUDSTACK-9603: 'concurrent.snapshots.threshold.perhost' parameter should not accept String.

priyankparihar:
'concurrent.snapshots.threshold.perhost' is a  global and cluster level setting. It should accept only integer values but without this fix it is accepting string  as well.  For complete description please refer -> [CLOUDSTACK-9603](https://issues.apache.org/jira/browse/CLOUDSTACK-9603)

### DIFF
--- a/server/src/com/cloud/configuration/ConfigurationManagerImpl.java
+++ b/server/src/com/cloud/configuration/ConfigurationManagerImpl.java
@@ -749,6 +749,9 @@ public class ConfigurationManagerImpl extends ManagerBase implements Configurati
             } else if (type.equals(Float.class)) {
                 errMsg = "There was error in trying to parse value: " + value + ". Please enter a valid float value for parameter " + name;
                 Float.parseFloat(value);
+            } else if (type.equals(Long.class)) {
+                errMsg = "There was error in trying to parse value: " + value + ". Please enter a valid long value for parameter " + name;
+                Long.parseLong(value);
             }
         } catch (final Exception e) {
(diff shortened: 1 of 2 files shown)


jburwell:
@priyankparihar could you please provide further explanation as to how this fix addresses the issue of `concurrent.snapshots.threshold.perhost` not being validated?

Also, is there an existing Marvin test case and/or unit test that verifies this fix?  If not, could you please add one?

rajesh-battala:
@priyankparihar  I have gone through the bug and the bug description is causing ambiguity that's the  @jburwell had made above comment. 
You can modify the description and which can reflect the fix or the check you are doing in the code.

priyankparihar:
Thanks @rajesh-battala and @jburwell  
>You can modify the description and which can reflect the fix or the check you are doing in the code.
Modified.

sowmyakrishn:
Reviewed test code. LGTM.

sateesh-chodapuneedi:
@borisstoyanov Can you please trigger tests for this PR? Tests could be over any hypervisor/simulator as the fix is pretty generic.

borisstoyanov:
@blueorangutan package

borisstoyanov:
Thanks for this fix @priyankparihar , if it addresses only the input validation should be pretty easy to test.

borisstoyanov:
@blueorangutan test

sateesh-chodapuneedi:
tag:mergeready

PR #1777, opened 2016-11-24
clearCLOUDSTACK-9133: Two volume.delete usage events are getting generated…

priyankparihar:
Two volume.delete usage events are getting generated for destroy vm.


syed:
Where is the other usage event generated? Does this only apply to ROOT volumes?

PR #1778, opened 2016-11-24
CLOUDSTACK-9613: Unable to set NAT rules on any interfaces except first interface of VM.

nitin-maharana:
Description:
=========
VMs have multiple network interfaces.
It is not possible to set a NAT rule on any of the interfaces except the first interface of the VM from UI. Because the UI only lists the IPs of the first interface.

The rules can only be set on the "default" interface but even after setting the secondary as the default interface the outcome is still the same. Only the first interface is listed in the UI as available for setting NAT rules.

Steps to Reproduce:
===============
1. Create two networks.
2. Create a VM.
3. Attach both the networks to the VM.
4. Create NAT rules using both the networks.
5. Observe the add VM section while creating the rule.
6. The same IPs appear for both the case.

Please see the below snaps to observe the issue clearly.

VM NIC1
image

VM NIC2
image

Creating NAT rule from Network-1
image

Add VMs section
image

Creating NAT rule from Network-2
image

Add VMs section
image

Expected:
========

Add VMs section from Network-1
image

Add VMs section from Network-2
image

Resolution:
========
Instead of always passing the id of the first nic, passing the id of
the nic which is on the same network as the IP address of which
we are enabling port-forwarding/load-balancing.

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -78,18 +78,13 @@
             });
 
         } else { //non-portable IP which has only one NIC
-            /*
-            var nic = $.grep(instance.nic, function(nic) {
-                return nic.networkid == network.id;
-            })[0];
-            */
 
             // Get NIC IPs
             $.ajax({
                 url: createURL('listNics'),
                 data: {
                     virtualmachineid: instance.id,
-                    nicId: instance.nic[0].id
+                    networkid: network.id
                 },
                 success: function(json) {
                     var nic = json.listnicsresponse.nic[0];


ustcweizhou:
@nitin-maharana Hi Nitin, can we use "networkId: network.id" instead nicId ?

nitin-maharana:
@ustcweizhou : Yes, we can directly pass networkid instead nicId. They handled the same way in case of multiple vm.

ustcweizhou:
@nitin-maharana so I suggest to use "networkId" instead of nicId, it is simpler change.

nitin-maharana:
@ustcweizhou : I made the changes. Thanks.

rhtyd:
Thanks @nitin-maharana a similar fix was merged with #1785 since it was for 4.9. Please close this PR.

PR #1780, opened 2016-11-24
Fix the KVM hypervisor's handling of not-really-SCSI devices.

ppenchev-storpool:
Hi,

Thanks for developing and maintaining CloudStack!

What do you think about the attached trivial patch that lets the KVM storage pool use more than one type of storage adaptor?  If a disk is handled by another adaptor and a request to disconnect it comes in, the KVM storage pool might hand it off to the iSCSI storage adaptor first.  With the current iSCSI adaptor code, it will recognize that the disk is not one of its own and return "true", signalling to the storage pool that the disk has been successfully disconnected.  The patch makes it return "false" so that the storage pool will pass the request to the *next* storage adaptor in the pool which might very well actually handle it.

Thanks in advance for your time and attention, and keep up the great work!

Best regards,
Peter

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/IscsiAdmStorageAdaptor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/IscsiAdmStorageAdaptor.java
@@ -349,8 +349,8 @@ public class IscsiAdmStorageAdaptor implements StorageAdaptor {
         String search4 = "-lun-";
 
         if (!localPath.contains(search3)) {
-            // this volume doesn't below to this adaptor, so just return true
-            return true;
+            // this volume doesn't below to this adaptor, so just return false
+            return false;
         }
 
         int index = localPath.indexOf(search2);


wido:
These small fixes *can* be very dangerous. So wondering how we can test this easily.

ppenchev-storpool:
Hi,

Thanks for looking at this!

Testing this could simply be a matter of adding more than one storage adaptor to a KVM storage pool, then connecting and disconnecting physical disks.  The problem will only manifest itself when disconnecting a disk that is not handled by the iSCSI provider if the loop in KVMStoragePoolManager.disconnectPhysicalDiskByPath() (file plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/storage/KVMStoragePoolManager.java around line 161) finds the iSCSI adaptor first.  Well, yeah, making sure the loop finds the iSCSI adaptor first may not be easy if Java orders the map keys almost randomly, but with some luck and/or some changes to the loop itself it could be arranged (change only the order in which the entries are walked, not the processing of the entries themselves).

BTW, this is actually the *only* place where the adaptor's disconnectPhysicalDiskByPath() method is invoked, so just analyzing the code there should show that this change ought to be safe.

Best regards,
Peter

syed:
@ppenchev-storpool I haven't worked with a non-iscsi adaptor for KVM but wouldn't it change the default behaviour now for non-iscsi disks? What if the other adaptors also ignore this? Can we add a default? Also is there a way of adding another adopter to test this from an integration test?

borisstoyanov:
@blueorangutan package

rhtyd:
Please fix the conflict.

ppenchev-storpool:
This, among other issues, seems to have been addressed in https://github.com/apache/cloudstack/commit/bf209405e7d60b6a5abf87677d368c429359d98a; thanks!

PR #1592, opened 2016-06-15
CLOUDSTACK-9416 (Bug in ACS master GUI) : Enabling Static NAT on an associated Public IP to one of the NICs (networks) of a multi-NIC VM fails due to a wrong (default) Guest VM IP being selected in the GUI

prashanthvarma:
Bug in ACS master GUI: Enabling Static NAT on an associated Public IP to one of the NICs (networks) of a multi-NIC VM fails due to a wrong (default) Guest VM IP being selected in the GUI.

ERROR:
INFO  [c.c.a.ApiServer](catalina-exec-7:ctx-83926837 ctx-fc7aa5ed) VM ip 10.10.2.163 address not belongs to the vm

ACS GUI before the bug fix:

network1
network2

As you can see in the above attached ACS GUI screen shots, same (default) guest VM IP is being selected in the ACS GUI for both the NICs (networks) of the multi-NIC VM while enabling Static NAT on the corresponding associated Public IPs. Thus, we hit this issue.

ACS GUI after this bug fix:

fixed_network1
fixed_network2

As you can see in the above attached ACS GUI screen shots, this bug fix resolves the above mentioned issue.

PR #1785, opened 2016-11-25
CLOUDSTACK-9416 : Enabling Static NAT on an associated Public IP to one of the NICs (networks) of a multi-NIC VM fails due to a wrong (default) Guest VM IP being selected in the GUI

prashanthvarma:
Bug in ACS master GUI: Enabling Static NAT on an associated Public IP to one of the NICs (networks) of a multi-NIC VM fails due to a wrong (default) Guest VM IP being selected in the GUI.

ERROR:
INFO  [c.c.a.ApiServer](catalina-exec-7:ctx-83926837 ctx-fc7aa5ed) VM ip 10.10.2.163 address not belongs to the vm

ACS GUI before the bug fix:

network1
network2

As you can see in the above attached ACS GUI screen shots, same (default) guest VM IP is being selected in the ACS GUI for both the NICs (networks) of the multi-NIC VM while enabling Static NAT on the corresponding associated Public IPs. Thus, we hit this issue.

ACS GUI after this bug fix:

fixed_network1
fixed_network2

As you can see in the above attached ACS GUI screen shots, this bug fix resolves the above mentioned issue.

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -89,7 +89,7 @@
                 url: createURL('listNics'),
                 data: {
                     virtualmachineid: instance.id,
-                    nicId: instance.nic[0].id
+                    networkId: network.id
                 },
                 success: function(json) {
                     var nic = json.listnicsresponse.nic[0];


prashanthvarma:
@rhtyd  This is a shadow of PR #1592 , rebased against 4.9 as requested.

Note: The original PR #1592 has enough LGTMs to merge this PR.

Once you merge this PR, I will close the PR #1592 .

Sorry for the inconvenience !!

ustcweizhou:
LGTM
this PR solves the same issue as #1778

rhtyd:
LGTM, this fixes the same issue as #1778. Since this is a pure UI change, manual testing with screenshot confirm the fix. I'll move with the merge now. The failure at Travis is not due to this change but intermittent/env issue.

PR #1786, opened 2016-11-25
CLOUDSTACK-9618: Load Balancer configuration page does not have "Source" method in the drop down list.

nitin-maharana:
If we create an isolated network with NetScaler published service offering for Load balancing service, then the load balancing configuration UI does not show "Source" as one of the supported LB methods in the drop down list. It only shows "Round-Robin" and "LeastConnection" methods in the list. However, It successfully creates LB rule with "Source" as the LB method using API.

### DIFF
--- a/plugins/network-elements/netscaler/src/com/cloud/network/element/NetscalerElement.java
+++ b/plugins/network-elements/netscaler/src/com/cloud/network/element/NetscalerElement.java
@@ -232,6 +232,10 @@ public class NetscalerElement extends ExternalLoadBalancerDeviceManagerImpl impl
 
     @Override
     public boolean validateLBRule(Network network, LoadBalancingRule rule) {
+        if (canHandle(network, Service.Lb)) {
+            String algo = rule.getAlgorithm();
+            return (algo.equals("roundrobin") || algo.equals("leastconn") || algo.equals("source"));
+        }
         return true;
     }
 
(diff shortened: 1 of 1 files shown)


plugins/network-elements/netscaler/src/com/cloud/network/element/NetscalerElement.java
```diff
@@ -260,7 +264,7 @@ public boolean applyLBRules(Network config, List<LoadBalancingRule> rules) throw
         Map<Capability, String> lbCapabilities = new HashMap<Capability, String>();
 
         // Specifies that the RoundRobin and Leastconn algorithms are supported for load balancing rules
-        lbCapabilities.put(Capability.SupportedLBAlgorithms, "roundrobin,leastconn");
+        lbCapabilities.put(Capability.SupportedLBAlgorithms, "roundrobin,leastconn,source");
```
jburwell: The list of strings declared on this line must be sync with the list used to validate the algorithm on line 237.  Therefore, please consider extracting this list to a constant `ImmutableSet<String>`, `SUPPORTED_ALGORITHMS` in order to ensure cohesion between declaration and validation.  With this change, the validateLBRule can be re-implemented as follows:

```java
public boolean validateLBRule(Network network, LoadBalancingRule rule) {
    Preconditions.checkArgument(network != null, "validateLBRule requires a non-null network");
    Preconditions.checkArgument(rule != null, "validateLBRule requires a non-null rule");

    return (canHandle(network, Service.Lb)) : SUPPORTED_ALGORITHMS.contains(rule.getAlgorithm()) : true;
}
```

This line can be re-implemented as follows:

```java
lbCapabilities.put(Capability.SupportedLBAlgorithms, Joiner.on(",").join(SUPPORTED_ALGORITHMS));
```

rajesh-battala: add space after "," 
should be like lbCapabilities.put(Capability.SupportedLBAlgorithms, "roundrobin,leastconn, source")

nitin-maharana:
Changed the base branch to 4.9.

nitin-maharana:
@jburwell : As this is Netscaler specific. Yes, you need a Netscaler to verify.

nitin-maharana:
ping @sateesh-chodapuneedi @rajesh-battala @karuturi

koushik-das:
LGTM based on code review
@karuturi Since this is a network plugin related fix and requires Netscaler, I don't think there is a need to run the BVTs.

PR #1804, opened 2016-12-01
CLOUDSTACK-9639: Unable to create shared network with vLan isolation

nitin-maharana:
Description:
=========
Create shared network fails with Error.
While creating a shared network it fails to create with Error "The new IP range you have specified has overlapped with the guest network in the zone: XYZ. Please specify a different gateway/netmask"

Steps to Reproduce:
===============
1. Create an isolated network with a subnet eg: 10.1.1.0/24
2. Create a shared network with the same subnet but different VLAN, we should observe this issue.

Expected Behaviour:
===============
It shouldn't restrict the creation of the guest network with the same subnet as long as they are segmented by VLAN.

ACTUAL BEHAVIOR:
===============
It doesn't allow the creation of shared guest networks if there is any isolated guest network using the same subnet although it allows using the same subnet in multiple shared networks.

Cause:
=====
The issue is happening because, when Shared network is getting created we are checking if there is any guest network already implemented with the same CIDR and throwing the error without checking if they are having same VLAN also. Creating the same CIDR shared network with different VLAN should be allowed.

Fix:
===
When creating a shared network, if there is any existing guest network with the same CIDR, we check if they are having the same VLAN, if they are in same VLAN, then we don't allow creating it. If they are in the same CIDR with different VLAN then we allowing to create the network.

### DIFF
--- a/server/src/com/cloud/configuration/ConfigurationManagerImpl.java
+++ b/server/src/com/cloud/configuration/ConfigurationManagerImpl.java
@@ -3052,12 +3052,12 @@ public class ConfigurationManagerImpl extends ManagerBase implements Configurati
             // Check if the new VLAN's subnet conflicts with the guest network
             // in
             // the specified zone (guestCidr is null for basic zone)
+            // when adding shared network with same cidr of zone guest cidr,
+            // if the specified vlan is not present in zone, physical network, allow to create the network as the isolation is based on VLAN.
             final String guestNetworkCidr = zone.getGuestNetworkCidr();
-            if (guestNetworkCidr != null) {
-                if (NetUtils.isNetworksOverlap(newCidr, guestNetworkCidr)) {
-                    throw new InvalidParameterValueException("The new IP range you have specified has  overlapped with the guest network in zone: " + zone.getName()
-                            + ". Please specify a different gateway/netmask.");
-                }
+            if (guestNetworkCidr != null && NetUtils.isNetworksOverlap(newCidr, guestNetworkCidr) && _zoneDao.findVnet(zoneId, physicalNetworkId, vlanId).isEmpty() != true) {
+                throw new InvalidParameterValueException("The new IP range you have specified has  overlapped with the guest network in zone: " + zone.getName()
+                        + "along with existing Vlan also. Please specify a different gateway/netmask");
             }
 
             // Check if there are any errors with the IP range


server/src/com/cloud/configuration/ConfigurationManagerImpl.java
```diff
@@ -3092,8 +3092,12 @@ public Vlan createVlanAndPublicIpRange(final long zoneId, final long networkId,
             final String guestNetworkCidr = zone.getGuestNetworkCidr();
             if (guestNetworkCidr != null) {
                 if (NetUtils.isNetworksOverlap(newCidr, guestNetworkCidr)) {
-                    throw new InvalidParameterValueException("The new IP range you have specified has  overlapped with the guest network in zone: " + zone.getName()
-                            + ". Please specify a different gateway/netmask.");
+                    // when adding shared network with same cidr of zone guest cidr,
+                    // if the specified vlan is not present in zone, physical network, allow to create the network as the isolation is based on VLAN.
+                    if (_zoneDao.findVnet(zoneId, physicalNetworkId, vlanId).size() > 0) {
```
jburwell: Please consider using `isEmpty` rather than a size check to determine whether or not a list contains elements.  It is more idiomatic/clear, but less error prone.  Also, why isn't this `if` condition combined with the previous `if` block and an `&&` operator.

nitin-maharana: @jburwell : Thanks for your suggestion. I will consider this change. The effect would be same if we put all the conditions with && operator.

rhtyd:
@nitin-maharana  this looks useful, can you change the base branch for the PR to 4.9, rebase your PR branch against 4.9? Can you add a marvin test for this?

rhtyd:
@nitin-maharana please push -f on the PR to re-kick travis, or close/open the PR

nitin-maharana:
@rhtyd : Changed the base branch to 4.9

jburwell:
@rhtyd looks like the Travis build is failing due to a timeout.  Do you have any ideas what could be causing this timeout?

@nitin-maharana are their existing Marvin tests that verify this fix?

rhtyd:
@jburwell I'll have a look. I've fixed on issue with Travis component test.

murali-reddy:
LGTM.

a small side note. User can create a guest VM with nic's in two networks with same CIDR but different VLAN isolations and mess-up routing.

PR #1806, opened 2016-12-02
travis: cleanup apt before installing packages

rhtyd:
This aims at fixing recent Travis failures.


rhtyd:
Based on recent Travis runs, merging this on discretion. The changes are only in Travis specific script so additional regression tests are not needed.

PR #1754, opened 2016-11-09
[4.10] Latest health test run

rhtyd:
Dummy PR to run trillian tests on latest master (4.10-snapshot).

@blueorangutan package

PR #1807, opened 2016-12-02
CLOUDSTACK-9633: Revert addition of `vhd` extention to snapshots

rhtyd:
This reverts commit f1fd325c085cd61336ac616ba76e2c1f3f916cd1 and changes
introduced in commit f46651e6721106941deeb6b5e6bf51d7e9efc61c that changed the
snapshot file name to include a `vhd` extension.

With this change, CloudStack users should not hit upgrade issues.

Reference: https://github.com/apache/cloudstack/pull/1600#pullrequestreview-10955963

### DIFF
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
@@ -529,7 +529,7 @@ public class Xenserver625StorageProcessor extends XenServerStorageProcessor {
                         // finalPath = folder + File.separator +
                         // snapshotBackupUuid;
                     } else {
-                        finalPath = folder + File.separator + snapshotBackupUuid + ".vhd";
+                        finalPath = folder + File.separator + snapshotBackupUuid;
                     }
 
                 } finally {
@@ -562,7 +562,7 @@ public class Xenserver625StorageProcessor extends XenServerStorageProcessor {
                     final String[] tmp = result.split("#");
                     snapshotBackupUuid = tmp[0];
                     physicalSize = Long.parseLong(tmp[1]);
-                    finalPath = folder + File.separator + snapshotBackupUuid + ".vhd";
+                    finalPath = folder + File.separator + snapshotBackupUuid;
                 }
             }
             final String volumeUuid = snapshotTO.getVolume().getPath();
--- a/tools/marvin/marvin/lib/utils.py
+++ b/tools/marvin/marvin/lib/utils.py
@@ -292,7 +292,7 @@ def is_snapshot_on_nfs(apiclient, dbconn, config, zoneid, snapshotid):
     # snapshot extension to be appended to the snapshot path obtained from db
     snapshot_extensions = {"vmware": ".ovf",
                             "kvm": "",
-                            "xenserver": "",
+                            "xenserver": ".vhd",
                             "simulator":""}
 
     qresultset = dbconn.execute(


rhtyd:
Pinging for review -- @jburwell @abhinandanprateek @borisstoyanov 
@blueorangutan package

rhtyd:

rhtyd:

serg38:
LGTM

abhinandanprateek:
@syed can you look at this 'revert' i guess this is going to be an issue with the managed storage change that has been put in. cc @rhtyd

abhinandanprateek:
@syed yes I assume that PR in itself is good. You need to consider the broken test and broken upgrade ,where upgrading users will have path without the extension.

rhtyd:
@abhinandanprateek @syed I'm seeing known intermittent failures with vpc/vr related tests, can I get LGTM on this? thanks.

syed:
@rhtyd I understand that we want to avoid breaking upgrades. But I have upgraded my setup and I did not find any problems. I am trying to understand what will break if upgrades happen, given the fact that this change will definitely break the newer functionality that we have added.

syed:
@abhinandanprateek the snapshot path is something which is internal to Cloudstack, as an end user, I would not be able to tell the difference before and after the upgrade, unless I missed something.

serg38:
@syed @rhtyd @abhinandanprateek Why cant we simply rename all install paths in DB for existing snapshots instead. It can be as simple as adding this query to schema upgrade path
UPDATE cloud.snapshots s
        JOIN
    cloud.snapshot_store_ref ssr ON ssr.snapshot_id = s.id 
SET 
    ssr.install_path = CONCAT(ssr.install_path, '.vhd')
WHERE
    s.hypervisor_type = 'XenServer'
        AND RIGHT(ssr.install_path, 4) != '.vhd';

rhtyd:
@serg38 I think we can do that, but will be also need some tooling to rename the actual files on the SRs? /cc @abhinandanprateek

rhtyd:
@syed Thanks, so just to confirm... with this PR, is your work with #1600 works and is not broken?

swill:
@rhtyd If you merge this you will break #1600.  We may want to do some more testing on the upgrade path, but I believe @syed has done some testing of the upgrade path.  Maybe we can put together a list of items to test for the upgrade path so we can confirm that this revert is not needed.

abhinandanprateek:
@syed @swill the change is good. What needs to be considered is what happens with old snapshots, from a older version upgraded to this version, that are without the .vhd extension. If it can be checked that in a upgraded version where the older snapshots are without .vhd extension and new ones with .vhd extension, that cloudstack operations will work fine. Otherwise during upgrade we need to add .vhd to older snapshots for Xen.
Another thing is that the marvin tests still assume that there is no .vhd extension and adds it. This needs to be fixed. We were seeing some marvin test fail on Xenserver due to a additonal .vhd extension. Look at the test_01_snapshot_root_disk  failure for Xen in this PR for example: https://github.com/apache/cloudstack/pull/1754 and the cause that @serg38  pointed out.

abhinandanprateek:
@syed @swill @rhtyd @serg38 I think i missed some updates. This PR will only break the original PR. The patch suggested by @serg38 looks good. 
@syed in a upgraded setup, I think removal of a older snapshot will fail for one, as these donot have a .vhd extension.

syed:
@abhinandanprateek I will check this. I will start with a non-upgraded cloudstack, create snapshots, upgrade cloudstack to have #1600 and then try to delete older snapshots. From the code it looks like it should delete irrespective of weather the snapshot has the `vhd` extension or not (see [here] (https://github.com/apache/cloudstack/blob/master/services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java#L1546) ) because it does a wildcard delete of the directory before.

Are you sure that delete will break? Can you point out where this might happen?

abhinandanprateek:
@syed please do a check of various snapshot functions on a upgraded setup. 

@rhtyd this PR can be closed.

rhtyd:
Thanks @abhinandanprateek @syed

PR #1816, opened 2016-12-05
CLOUDSTACK-9564: Fix NPE due to intermittent test assertion

rhtyd:
The test assertion on a pool object may return a null object, as objects
can be randomly expired/tombstoned. This will fix a NPE sometimes seen due
to recently merge for the fix for CLOUDSTACK-9564.

(we can merge this if Travis passes)

/cc @abhinandanprateek @murali-reddy


rhtyd:
LGTM, merging this on last TravisCI's first job result. Only the first job runs/builds with unit tests.

PR #672, opened 2015-08-10
Allow VM snapshots and volume snapshots to exist together

anshul1886:
This branch fixes issues involved in allowing VM snapshots and volume snapshots to exist together.

https://issues.apache.org/jira/browse/CLOUDSTACK-8663

PR #1829, opened 2016-12-14
CLOUDSTACK-9363: Fix HVM VM restart bug in XenServer

syed:
Here is the longer description of the problem:

By default XenServer limits HVM guests to only 4 disks. Two of those are reserved for the ROOT disk (deviceId=0) and CD ROM (device ID=3) which means that we can only attach 2 data disks. This limit however is removed when Xentools is installed on the guest. The information that a guest has Xentools installed and can handle more than 4 disks is stored in the VM metadata on XenServer. When a VM is shut down, Cloudstack removes the VM and all the metadata associated with the VM from XenServer. Now, when you start the VM again, even if it has Xentools installed, it will default to only 4 attachable disks.

Now this problem manifests itself when you have a HVM VM and you stop and start it with more than 2 data disks attached. The VM fails to start and the only way to start the VM is to detach the extra disks and then reattach them after the VM start.

In this fix, I am removing the check which is done before creating a `VBD` which enforces this limit. This will not affect current workflow and will fix the HVM issue.

@koushik-das this is related to the "autodetect" feature that you introduced a while back (https://issues.apache.org/jira/browse/CLOUDSTACK-8826). I would love your review on this fix.

### DIFF
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
@@ -1149,8 +1149,8 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
             vbdr.unpluggable = (volume.getType() == Volume.Type.ROOT) ? false : true;
             vbdr.userdevice = "autodetect";
             final Long deviceId = volume.getDiskSeq();
-            if (deviceId != null && !isDeviceUsed(conn, vm, deviceId)) {
-                vbdr.userdevice = deviceId.toString();
+            if (deviceId != null && (!isDeviceUsed(conn, vm, deviceId) || deviceId > 3)) {
+                    vbdr.userdevice = deviceId.toString();
             }
         }
         final VBD vbd = VBD.create(conn, vbdr);


rhtyd:
@syed can you rebase against 4.9 and change PR's branch to 4.9? As soon as 4.9.1.0 is out, we'll work towards 4.9.2.0 and work towards the next minor release to include several such bugfixes.

koushik-das:
@syed Please check #672. There was a discussion sometimes back on dev@ and this PR was mentioned. I feel that fix is slightly better than removing the check altogether.

syed:
@koushik-das I like that fix. I've modified my fix to do a better check. @rhtyd I've rebased to 4.9 as well. 
Thank you guys for the prompt replies :)

rhtyd:
@koushik-das @syed is this blocker for 4.9/4.10?

koushik-das:
Changes LGTM. As @syed mentioned there is a workaround, so won't qualify as a blocker.

pdube:
@rhtyd @koushik-das definitely a blocker IMHO. Being unable to start/reboot a VM that has 3 volumes is unacceptable in a public cloud scenario

koushik-das:
@pdube The description mentions about a way to workaround the problem, thats the reason for not calling it a blocker. If you think the issue is important for the current 4.9 release, please reply on the voting thread that is currently going on.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test centos7 xenserver-65sp2

borisstoyanov:

rhtyd:
LGTM. @karuturi this can be merged.

borisstoyanov:
Hi @syed @koushik-das, I'm trying to verify this as well, I've built an environment and I was wondering what template were you guys using? 
I've tried with this one: http://dl.openvm.eu/cloudstack/centos/vanilla/6/x86_64/CentOS-6-x86_64-vanilla-xen.vhd.bz2
but got: 
screen shot 2017-02-15 at 4 37 09 pm

do you have a template that I could register and use for the sake of the test?

koushik-das:
@borisstoyanov try using a Windows template

borisstoyanov:
@koushik-das happen to have any handy?

koushik-das:
@borisstoyanov Sorry don't have a windows template

borisstoyanov:
I've verified this with Windows HVM enabled template. Having more than 4 disks (6 data disks in my case), removing the disk on deviceId=4. After a reboot the VM boots as expected, user is able to operate with all the data disks. Just one question I have, how are the unix machines affected by this, should we address testing them?

borisstoyanov:

borisstoyanov:
@blueorangutan test centos7 vmware-60u2

syed:
@borisstoyanov  I've done my testing on a linux based HVM VM and it works as expected after the fix.

borisstoyanov:
thanks @syed

rhtyd:
LGTM. /cc @karuturi

borisstoyanov:
@blueorangutan test centos7 vmware-60u2

rhtyd:
LGTM, code and (regression) tests.

rhtyd:
@karuturi let's merge this, thanks.

rhtyd:
This is ready for merge @karuturi

anshul1886:
@rhtyd @karuturi @syed, This PR has subset of changes of #672 and #1941.  So how should we go about merging these?

syed:
@anshul1886 it looks like #672 is closed so that leaves us with #1941 , Are we targeting it for 4.10? If yes, then there are some comments there which you might want to address there. If we do make it into 4.10 then we can close this and use #1941 instead.

rhtyd:
Okay @syed either PR need to be merged to solve the issue of restart bug.

syed:
@rhtyd @anshul1886 I suggest merging this first as the changes here are very limited and tested. We can refactor the change out from #1941

karuturi:
Both #1829 and #1941 are merged and is causing build failures. I am creating new PR with the revert.

PR #1853, opened 2016-12-22
CLOUDSTACK-9696: Fixed Virtual Router deployment failing if there are…

anshul1886:
… no shared resources available

### DIFF
--- a/server/src/com/cloud/dc/dao/DedicatedResourceDaoImpl.java
+++ b/server/src/com/cloud/dc/dao/DedicatedResourceDaoImpl.java
@@ -18,7 +18,6 @@ package com.cloud.dc.dao;
 
 import java.util.List;
 
-
 import org.springframework.stereotype.Component;
 
 import com.cloud.dc.DedicatedResourceVO;
@@ -54,6 +53,7 @@ public class DedicatedResourceDaoImpl extends GenericDaoBase<DedicatedResourceVO
     protected SearchBuilder<DedicatedResourceVO> ListAllPodsSearch;
     protected SearchBuilder<DedicatedResourceVO> ListAllClustersSearch;
     protected SearchBuilder<DedicatedResourceVO> ListAllHostsSearch;
+    protected SearchBuilder<DedicatedResourceVO> ListAvailableResourcesSearch;
 
     protected SearchBuilder<DedicatedResourceVO> ListByAccountId;
     protected SearchBuilder<DedicatedResourceVO> ListByDomainId;
@@ -164,6 +164,12 @@ public class DedicatedResourceDaoImpl extends GenericDaoBase<DedicatedResourceVO
         ListAllHostsSearch.and("domainId", ListAllHostsSearch.entity().getDomainId(), Op.EQ);
         ListAllHostsSearch.and("affinityGroupId", ListAllHostsSearch.entity().getAffinityGroupId(), Op.EQ);
         ListAllHostsSearch.done();
+
+        ListAvailableResourcesSearch = createSearchBuilder();
+        ListAvailableResourcesSearch.or("accountId", ListAvailableResourcesSearch.entity().getAccountId(), Op.EQ);
+        ListAvailableResourcesSearch.or().op("nullaccountId", ListAvailableResourcesSearch.entity().getAccountId(), Op.NULL);
+        ListAvailableResourcesSearch.and("domainId", ListAvailableResourcesSearch.entity().getDomainId(), Op.IN);
+        ListAvailableResourcesSearch.cp();
 
         ListByAccountId = createSearchBuilder();
         ListByAccountId.and("accountId", ListByAccountId.entity().getAccountId(), SearchCriteria.Op.EQ);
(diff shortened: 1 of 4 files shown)


server/src/com/cloud/dc/dao/DedicatedResourceDaoImpl.java
```diff
@@ -312,6 +317,65 @@ public DedicatedResourceVO findByHostId(Long hostId) {
     }
 
     @Override
+    public List<DedicatedResourceVO> listAvailableResources(Long accountId, Long... domains) {
```
koushik-das: Please rewrite this using the standard query builder (SearchBuilder) that is available and used in the other methods in this class.

server/src/com/cloud/deploy/DeploymentPlanningManagerImpl.java
```diff
@@ -584,105 +583,64 @@ private void checkForNonDedicatedResources(VirtualMachineProfile vmProfile, Data
...
-
-                tempStorage = _dedicatedDao.searchDedicatedPods(null, vmDomainId, vmAccountId, null).first();
-
-                for(DedicatedResourceVO vo : tempStorage) {
-                    allPodsFromDedicatedID.add(vo.getPodId());
-                }
-
-                tempStorage.clear();
-                tempStorage = _dedicatedDao.searchDedicatedClusters(null, vmDomainId, vmAccountId, null).first();
+        if ((vm.getType() == VirtualMachine.Type.User && !isExplicit) || vm.getType() == VirtualMachine.Type.DomainRouter) {
```
koushik-das: Any reason for this check? checkForNonDedicatedResources() is anyway called for user and router VMs.

server/src/com/cloud/deploy/DeploymentPlanningManagerImpl.java
```diff
@@ -584,105 +583,64 @@ private void checkForNonDedicatedResources(VirtualMachineProfile vmProfile, Data
...
 
-                for(DedicatedResourceVO vo : tempStorage) {
-                    allHostsFromDedicatedID.add(vo.getHostId());
+                    if (vo.getHostId() != null) {
+                        allHostsFromDedicatedID.add(vo.getHostId());
+                    }
                 }
 
                 //Remove the dedicated ones from main list
                 allPodsInDc.removeAll(allPodsFromDedicatedID);
```
koushik-das: Can you put some explanation as to why there is a difference in checking for dedicated resources for user VM and VR? In case of user VMs dedicated resources are put in avoid list, for VR only non-dedicated resources are put in avoid list.

server/src/com/cloud/dc/dao/DedicatedResourceDaoImpl.java
anshul1886: @koushik-das Could not find a way to write the conditional statement so went with this way. Here account_id in involved in two places in condition.

server/src/com/cloud/deploy/DeploymentPlanningManagerImpl.java
anshul1886: @koushik-das Here additional isExplicit check is needed.

server/src/com/cloud/deploy/DeploymentPlanningManagerImpl.java
anshul1886: @koushik-das User can specify the affinity group for VM while deploying VM. So if user has specified the affinity group then only that VM needs to go to dedicated hosts. While in case of router user has not such option and needs to be inferred from account/domain of the user. These whole checks are mainly introduced to avoid virtual router going to dedicated resources of other account/domain.

server/src/com/cloud/dc/dao/DedicatedResourceDaoImpl.java
koushik-das: @anshul1886 Check out SearchBuilder class, it has methods like or(), op(), cp() which can be used to construct the query. Also the same field can be used in multiple places, search for reference to or(), op() etc. elsewhere in the code and you will find how to use them. One such place is QueryManagerImpl.java

server/src/com/cloud/deploy/DeploymentPlanningManagerImpl.java
koushik-das: @anshul1886 As per the code for user VMs without explicit dedication, all dedicated resources are put in 'avoid' list so deployment happens in non-dedicated resources only. This part is clear. But for VR the opposite is happening as per the code. Can you explain the VR part from code?

server/src/com/cloud/deploy/DeploymentPlanningManagerImpl.java
koushik-das: @anshul1886 Got it

server/src/com/cloud/dc/dao/DedicatedResourceDaoImpl.java
anshul1886: @koushik-das Made the changes using SearchBuilder class.

server/src/com/cloud/deploy/DeploymentPlanningManagerImpl.java
anshul1886: @koushik-das Renamed variables to make the code more understandable.

anshul1886:
@koushik-das Made the changes as per review comments.

koushik-das:
Code change LGTM.
@anshul1886 Please make sure Travis is green.

anshul1886:
Fixed conflict occurred due to typo fix of CLOUDSTACK-9783.

anshul1886:
Closing PR as issue is not seen anymore.

PR #1855, opened 2016-12-22
CLOUDSTACK-9697: Better error message in UI if user tries to shrink VM ROOT Volume size

rashmidixit:
Skip the API call altogether if the UI detects this case and throw a more user friendly message

Conflicts:
	client/WEB-INF/classes/resources/messages.properties
	client/WEB-INF/classes/resources/messages_ja_JP.properties
	client/WEB-INF/classes/resources/messages_zh_CN.properties
	ui/dictionary2.jsp


sadhugit:
tested on my setup and Looks good to me.
Need to handle on API side  so please update the Bug description  by adding UI only

rashmidixit:
@sadhugit I have updated the bug description based on your comments.

ramkatru:
tag:mergeready

karuturi:
@rashmidixit Can you rebase this to master? master doesnt have messages.properties

rashmidixit:
Need to reopen same PR on the master branch. Will create a new PR for the same.

PR #1856, opened 2016-12-22
CLOUDSTACK-9569: propagate global configuration router.aggregation.command.each.timeout to KVM agent

ustcweizhou:
The router.aggregation.command.each.timeout in global configuration is only applied on new created KVM host.
For existing KVM host, changing the value will not be effective.
We need to propagate the configuration to existing host when cloudstack-agent is connected.


core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
```diff
@@ -259,8 +260,18 @@ private Answer execute(GetDomRVersionCmd cmd) {
         return new GetDomRVersionAnswer(cmd, result.getDetails(), lines[0], lines[1]);
     }
 
+    public boolean configureHostParams(final Map<String, String> params) {
+        if (_params.get("router.aggregation.command.each.timeout") == null) {
```
murali-reddy: Why do we need this check?

murali-reddy:
@ustcweizhou this will be useful for 4.9 as well, would you mind rebasing the patch to 4.9?

core/src/com/cloud/agent/api/SetHostParamsCommand.java
```diff
@@ -0,0 +1,43 @@
...
+// under the License.
+//
+
+package com.cloud.agent.api;
+
+import java.util.Map;
+
+public class SetHostParamsCommand extends Command {
+
+    Map<String, String> params;
```
koushik-das: Map is not serializable, also HashMap that is actually used may have issues with serialization. Please check.

core/src/com/cloud/agent/resource/virtualnetwork/VirtualRoutingResource.java
ustcweizhou: the idea is: if router.aggregation.command.each.timeout is already set in agent.properties, then ignore the propagation.

ustcweizhou:
rebase with apache/4.9

borisstoyanov:
@blueorangutan package

core/src/com/cloud/agent/api/SetHostParamsCommand.java
ustcweizhou: @koushik-das luckily there is no issue on our platforms. We have some similar definitions in other Command in cloudstack actually.

rhtyd:
@blueorangutan package

rhtyd:
LGTM.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@karuturi this is ready for merge.

PR #1857, opened 2016-12-22
BUG-ID: CLOUDSTACK-8284

adwaitpatankar:
On expunge VM, initiate resource count update at domain
level. Setting accountId as null ensures resource count is done at domain level and any hierarchies below.

### DIFF
--- a/server/src/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/com/cloud/vm/UserVmManagerImpl.java
@@ -2050,7 +2050,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
             // Update Resource count
             if (vm.getAccountId() != Account.ACCOUNT_ID_SYSTEM && !rootVol.isEmpty()) {
                 _resourceLimitMgr.decrementResourceCount(vm.getAccountId(), ResourceType.volume);
-                _resourceLimitMgr.recalculateResourceCount(vm.getAccountId(), vm.getDomainId(), ResourceType.primary_storage.getOrdinal());
+                _resourceLimitMgr.recalculateResourceCount(null/*acclountId*/, vm.getDomainId(), ResourceType.primary_storage.getOrdinal());
             }
 
             // Only if vm is not expunged already, cleanup it's resources


cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 348
 Hypervisor xenserver
 NetworkType Advanced
 Passed=103
 Failed=1
 Skipped=7

**Failed tests:**
* test_routers_network_ops.py

 * test_03_RVR_Network_check_router_state Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

ustcweizhou:
@adwaitpatankar this is because _resourceLimitMgr.recalculateResourceCount only recalculate the resource count on account(if accountid is set), or domain (if accountid is not set), not both account/domain.

We have this issue before, it is fixed by replacing _resourceLimitMgr.recalculateResourceCount with
{code}
                _resourceLimitMgr.decrementResourceCount(vm.getAccountId(), ResourceType.primary_storage, rootVol.get(0).isDisplay(), new Long(rootVol.get(0).getSize()));
{code}

adwaitpatankar:
@ustcweizhou, the issue is that the resource count at domain level are not getting updated implicitly after expunge VM call. If the user tries to add a new instance, and if the primary storage resource count is not up to date, then it results in a false alarm (exception) and not allowing the user to create a guest instance even though the requested storage space is withing the limits of the domain quota.
When we do not pass accountID (as per the fix), the resource counts would be getting calculated on expungeVM event at Domain level which is the expectation as per the ticket. Also when the resource calculation is done at Domain level, all resource counts for all accounts within that domain are recalculated as well.

ustcweizhou:
@adwaitpatankar oh, I totally understand the issue and the change you made. You are right as I remember. However, I think the recalculation on other accounts in the domain is not necessary ( we also thought about this change before), so please test my suggestion in previous comment if possible.

by the way, the issue also exists in engine/storage/volume/src/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java.

kishankavala:
@adwaitpatankar Can you please try @ustcweizhou 's suggestion. That looks like a better suggestion to me.
recalculateResourceCount wont be required, if volume size is included in decrementResource count.

DaanHoogland:
@adwaitpatankar please rebase and re-open if still relevant

PR #1868, opened 2016-12-26
CLOUDSTACK-9707: While using hostid parameter, vm gets deployed on an…

anshul1886:
…other if the host

given is running out of capacity. If host id is specified the deployment should happen
on the given host and it should fail if the host is out of capacity. We are retrying
deployment on the entire zone without the given host id if we fail once. The retry,
which will retry on other hosts, should only be attempted if host id isn't given.

### DIFF
--- a/engine/api/src/org/apache/cloudstack/engine/cloud/entity/api/VirtualMachineEntity.java
+++ b/engine/api/src/org/apache/cloudstack/engine/cloud/entity/api/VirtualMachineEntity.java
@@ -106,7 +106,8 @@ public interface VirtualMachineEntity extends CloudStackEntity {
      * @param reservationId reservation id from reserve call.
      *
      */
-    void deploy(String reservationId, String caller, Map<VirtualMachineProfile.Param, Object> params) throws InsufficientCapacityException, ResourceUnavailableException;
+    void deploy(String reservationId, String caller, Map<VirtualMachineProfile.Param, Object> params, boolean deployOnGivenHost)
+            throws InsufficientCapacityException, ResourceUnavailableException;
 
     /**
      * Stop the virtual machine
--- a/engine/orchestration/src/org/apache/cloudstack/engine/cloud/entity/api/VMEntityManager.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/cloud/entity/api/VMEntityManager.java
@@ -39,7 +39,7 @@ public interface VMEntityManager {
     String reserveVirtualMachine(VMEntityVO vmEntityVO, DeploymentPlanner plannerToUse, DeploymentPlan plan, ExcludeList exclude) throws InsufficientCapacityException,
         ResourceUnavailableException;
 
-    void deployVirtualMachine(String reservationId, VMEntityVO vmEntityVO, String caller, Map<VirtualMachineProfile.Param, Object> params)
+    void deployVirtualMachine(String reservationId, VMEntityVO vmEntityVO, String caller, Map<VirtualMachineProfile.Param, Object> params, boolean deployOnGivenHost)
         throws InsufficientCapacityException, ResourceUnavailableException;
 
     boolean stopvirtualmachine(VMEntityVO vmEntityVO, String caller) throws ResourceUnavailableException;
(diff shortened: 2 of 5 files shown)


cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 323
 Hypervisor xenserver
 NetworkType Advanced
 Passed=104
 Failed=0
 Skipped=7

**Failed tests:**

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_routers_network_ops.py
test_disk_offerings.py

koushik-das:
Code changes LGTM, also verified that deployment is not retried in case of failure if host is specified

ustcweizhou:
Do we need to add a global setting to determine whether deploy to other hosts if the specified host is out of capacity ?

anshul1886:
@ustcweizhou, In my opinion we should not do that as that way we are giving them option of silently deploying VM on some host which he might not be aware of. As after specifying host he is expecting that VM will be deployed on that host. If he is ok with deploying VM on different host then he can try deploying VM in next attempt without specifying hostid.

ustcweizhou:
@anshul1886 
Considering about backwards compatibility. I would like to add the global setting, then users will have two options
(1) deploy vm fails if specified host does not have enough capacity. This is what this PR wants to do.
(2) deploy to other hosts if specified host does not have enough capacity. In this case, the specified host can be regarded as the perferred host. This is also what cloudstack works currently.

anshul1886:
@ustcweizhou, Is it looking good to you now?

ustcweizhou:
@anshul1886 
Sorry I think it would be better to add a global setting.
This is campatible with 3.X, but not with 4.X which has more users.

anshul1886:
@ustcweizhou Added the global setting allow.deploy.vm.if.deploy.on.given.host.fails to control the behaviour.

ustcweizhou:
@anshul1886 good, I will test it.
Could you please change the default value to true ?

anshul1886:
@ustcweizhou I think default value should be false. As it tells the user to make informed decision instead of misleading according to original purpose of parameter. @sateesh-chodapuneedi @koushik-das What do you guys think?

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

sateesh-chodapuneedi:
The point of specifying a host is in deploy request itself means inclination towards using that host for instance. Hence I think default value should be false. 
While global setting makes sense, rather than a global setting, it would be better to add optional parameter in API itself which indicates whether deploy should strictly try deployment on specific host, if any, or look for other hosts as well. That way more granular control in case by case perspective.

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 448
 Hypervisor xenserver
 NetworkType Advanced
 Passed=104
 Failed=1
 Skipped=7

**Failed tests:**
* test_routers_network_ops.py

 * test_01_isolate_network_FW_PF_default_routes_egress_true Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

ramkatru:
tag:mergeready

PR #1884, opened 2017-01-02
CLOUDSTACK-9699: Add global setting for enable/disable Metrics feature

rashmidixit:
The Metrics view for each type of entity basically fires APIs and calculates required values on the client end. For e.g. to display memory usage etc at the zone level, it will fetch all zones. For each zone it will fetch pods->cluster->host->VMs
For a very large Cloudstack installation this will have a major impact on the performance. 
Ideally, there should be an API which calculates all metrics in the backend and the UI should simply show the values. However, for the time, introduce a global setting called enable.metrics which will be set to false. This will cause the metrics button not to be shown on any of the pages.
If the Admin changes this to true, then the button will be visible and Metrics functionality will work as usual.

In this pull request, have also added JUnit test case for listCapabilities API.

Additionally added a test case in marvin called test_global_settings as part of the smoke tests which test the output of the listCapabilities API and checks if this setting is present or not.

Refer to [CLOUDSTACK-9699](https://issues.apache.org/jira/browse/CLOUDSTACK-9699) for more details.

### DIFF
--- a/api/test/org/apache/cloudstack/api/command/test/ListCapabilitiesCmdTest.java
+++ b/api/test/org/apache/cloudstack/api/command/test/ListCapabilitiesCmdTest.java
@@ -1,3 +1,20 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
 package org.apache.cloudstack.api.command.test;
 
 
--- a/api/test/org/apache/cloudstack/api/command/test/ListCapabilitiesCmdTest.java
+++ b/api/test/org/apache/cloudstack/api/command/test/ListCapabilitiesCmdTest.java
@@ -1,0 +1,63 @@
+package org.apache.cloudstack.api.command.test;
+
+
+import com.cloud.server.ManagementService;
+import com.cloud.utils.Pair;
+import junit.framework.Assert;
+import junit.framework.TestCase;
+import org.apache.cloudstack.api.ResponseGenerator;
+import org.apache.cloudstack.api.command.admin.config.ListCfgsByCmd;
+import org.apache.cloudstack.api.command.user.config.ListCapabilitiesCmd;
+import org.apache.cloudstack.api.response.CapabilitiesResponse;
+import org.apache.cloudstack.api.response.ConfigurationResponse;
+import org.apache.cloudstack.api.response.ListResponse;
+import org.apache.cloudstack.config.Configuration;
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class ListCapabilitiesCmdTest extends TestCase {
+
+    private ListCapabilitiesCmd listCapabilitiesCmd;
+    private ManagementService mgr;
+    private ResponseGenerator responseGenerator;
+
+    @Override
+    @Before
+    public void setUp() {
+        responseGenerator = Mockito.mock(ResponseGenerator.class);
+        mgr = Mockito.mock(ManagementService.class);
+        listCapabilitiesCmd = new ListCapabilitiesCmd();
+    }
+
+    @Test
+    public void testCreateSuccess() {
+
+        listCapabilitiesCmd._mgr = mgr;
+        listCapabilitiesCmd._responseGenerator = responseGenerator;
+
+        Map<String, Object> result = new HashMap<String, Object>();
+
+        try {
+            Mockito.when(mgr.listCapabilities(listCapabilitiesCmd)).thenReturn(result);
+        } catch (Exception e) {
+            Assert.fail("Received exception when success expected " + e.getMessage());
+        }
+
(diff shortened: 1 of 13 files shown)


api/test/org/apache/cloudstack/api/command/test/ListCapabilitiesCmdTest.java
```diff
@@ -0,0 +1,80 @@
...
+        listCapabilitiesCmd._mgr = mgr;
+        listCapabilitiesCmd._responseGenerator = responseGenerator;
+
+        Map<String, Object> result = new HashMap<String, Object>();
+
+        try {
+            Mockito.when(mgr.listCapabilities(listCapabilitiesCmd)).thenReturn(result);
+        } catch (Exception e) {
+            Assert.fail("Received exception when success expected " + e.getMessage());
+        }
```
jburwell: Why not perform lines 58-67 in in `setUp`?  Not only would this approach focus this method on the test operations, but remove the need for unnecessary class-level attributes.

Also, catching exceptions is not required because JUnit will automatically fail when exceptions are thrown.  If there are checked exceptions, add them to the `throws` of the test method to keep test methods as succinct as possible.

api/test/org/apache/cloudstack/api/command/test/ListCapabilitiesCmdTest.java
```diff
@@ -0,0 +1,80 @@
...
+        listCapabilitiesCmd = new ListCapabilitiesCmd();
+    }
+
+    @Test
+    public void testCreateSuccess() {
+
+        listCapabilitiesCmd._mgr = mgr;
+        listCapabilitiesCmd._responseGenerator = responseGenerator;
+
+        Map<String, Object> result = new HashMap<String, Object>();
```
jburwell: Please consider using ``Collections.emptyMap()`` instead of allocating a new ``Map`` to improve clarity of intention.

test/integration/smoke/test_global_settings.py
```diff
@@ -63,6 +63,33 @@ def test_UpdateConfigParamWithScope(self):
...
+        self.assertIsNotNone(
+                listCapabilities,
+                "Failed to list Capabilities"
+        )
+
+        # Verifying enable.metrics.ui is not null
+        self.assertIsNotNone(
+                listCapabilities.enablemetricsui,
+                "Failed to fetch enable.metrics.ui"
+        )
```
jburwell: Please add tests to verify that the value is set as expected.

ui/scripts/storage.js
```diff
@@ -54,6 +54,12 @@
                             label: 'label.vm.display.name'
                         }
                     },
+                    actionPreFilter: function(args){
+                        if (g_enablemetricsui)
+                            return ['add', 'viewMetrics', 'uploadVolume', 'uploadVolumefromLocal'];
+                        else
+                            return ['add', 'uploadVolume', 'uploadVolumefromLocal'];
+                    },
```
jburwell: Please remove the duplication in this function by refactoring as follows:

```
actionPreFilter: function(args){
                         actions = ['add', 'uploadVolume', 'uploadVolumefromLocal'];
                         if (g_enablemetricsui) {
                             actions.splice(1, 0, 'viewMetrics');
                         }
                         return actions;
                     },

File: ui/scripts/system.js [hunk 5]
```diff
@@ -7851,9 +7851,14 @@
...
                                         });
                                     }
-                                });
+                               });
+                            },
+                            actionPreFilter: function(args){
+                                if (g_enablemetricsui)
+                                    return ['add', 'viewMetrics'];
+                                else
+                                    return ['add'];
```
jburwell: Please remove the duplication in this function by refactoring as follows:

```
actions = ['add'];
if (g_enablemetricsui) {
   actions.append('viewMetrics');
}
return actions;
```

ui/scripts/system.js
```diff
@@ -14088,6 +14093,12 @@
                             }
                         });
                     },
+                    actionPreFilter: function(args){
+                        if (g_enablemetricsui)
+                            return ['add', 'viewMetrics'];
+                        else
+                            return ['add'];
+                    },
```
jburwell: Lines 14096-14101 appear to be a duplication of lines 7857-7860.  Can this code duplication be removed?

ui/scripts/system.js
```diff
@@ -15581,7 +15592,12 @@
                             }
                         });
                     },
-
+                    actionPreFilter: function(args){
+                        if (g_enablemetricsui)
+                            return ['add', 'viewMetrics'];
+                        else
+                            return ['add'];
+                    },
```
jburwell: Please code duplication comments above.

ui/scripts/system.js
```diff
@@ -17309,7 +17325,12 @@
...
```
jburwell: Please code duplication comments above.

ui/scripts/ui/widgets/listView.js
```diff
@@ -1922,7 +1922,14 @@
 
         // List view header actions
         if (listViewData.actions) {
+            // If a preFilter is set, then get the values. Else set this to null
+            var filteredActions = listViewData.actionPreFilter == undefined ?
+                        null : listViewData.actionPreFilter();
```
jburwell: Why not set to an empty list rather than ``null`` to avoid a potential null error?

ui/scripts/ui/widgets/listView.js
```diff
@@ -1922,7 +1922,14 @@
...
             $.each(listViewData.actions, function(actionName, action) {
+                // filter out those actions that shouldn't be shown
+                if (filteredActions != null
+                        && (filteredActions.indexOf(actionName) < 0))
+                    return false;
```
jburwell: Per coding standards, please surround all `if` blocks with curly braces.

rashmidixit:
Thank you @jburwell. I will incorporate the comments in a couple of days and resubmit.

rhtyd:
@rashmidixit I'm working on writing separate metrics view apis (backend/apis), they will improve the performance significantly. An explicit global settings is not necessary, pl. hold on this PR.

rhtyd:
I'm working on improving the metrics view feature by implementing the logic as backend APIs, will post a PR this/next week. Thanks.

rashmidixit:
@rhtyd Thanks for the update.

rhtyd:
This is not needed with PR #1944, though changes around capability refactoring can be separated.

rashmidixit:
Closing this based on the new PR  #1944.

PR #1888, opened 2017-01-04
CLOUDSTACK-9710: Switch to JRE1.8

rhtyd:
- Switches Travis to use jdk1.8
- Changes java-version to 1.8
- Change jdk/maven version to 1.8
- Switch to F5/java8 compatible library release
- Switch packaging to use jdk 1.8, and jre 1.8 in init/systemd scripts
- Switch systemvm to openjdk-8-jre

By default, openjdk-8 packages are not available on Debian 7 (wheezy); installing the pkg from jessie/testing causes several package regressions so I've used an openjdk free/opensource repository maintained by Azul, please review their terms of use: http://www.azul.com/products/zulu/zulu-terms-of-use
They are similar to what openjdk, or any other package on Debian would provide.


rhtyd:
@blueorangutan package

rhtyd:
This one is open for review, pinging for review -- @swill @karuturi @abhinandanprateek @murali-reddy @PaulAngus @nvazquez @serg38 @milamberspace @wido @syed @mike-tutkowski and others

High level major changes:
- Switches to openjdk 8 for building CloudStack.
- Switches to openjdk-8-jre for running mgmt server, usage server, agents. Test results are based on CentOS6, CentOS7 against Xen, VMware and KVM.
- A new systemvmtemplate with openjdk-8 (zulu-8 distribution of openjdk from Azul as Debian7 repos don't have openjdk-8-jre) needs to be used. A compatible 4.6.0 systemvmtemplate was built and tested, that only replaces openjdk-7 JRE with openjdk-8 JRE and was found to be compatible with ACS 4.6.0+. If StrongSwan PR is merged, we'll need to create and use a new systemvmtemplate with 4.10+.

abhinandanprateek:
lgtm on code review @rhtyd

karuturi:
code LGTM.

rhtyd:
All, for transparency this comment has the build log of the new systemvmtemplate attached:
[systemvm-buildlog.txt](https://github.com/apache/cloudstack/files/687704/systemvm-buildlog.txt)

The systemvmtemplates for 4.10 (Debian7 based with openjdk-8/zulu-8): http://packages.shapeblue.com/systemvmtemplate/custom/4.10-jre8

Notes for above systemvmtemplate:
- Newer openssl/openssh-server updates, along with other security updates
- OpenJDK-8 (zulu-8 distribution as wheezy don't have jre-8 packages yet)
- Bigger /var/log partition
- No major change wrt any other package such as apache2, openswan etc.

wido:
Overall the changes look good to me. But can you explain why for example you changed the dhcp package name? It doesn't seem related to Java 8.

 All in favor of this switch, just checking why it's in there.

rhtyd:
Thanks @wido I'll rever that line, though the dhcp-3 is a virtual package that actually install isc-dhcp-client, see https://packages.debian.org/wheezy/dhcp3-client

wido:
@rhtyd Understood, no problem, just trying to keep the changes in this PR clean. All changes should be Java 8 related.

rhtyd:
Thanks @wido I've reverted and updated the PR now, you can refresh to see the updated changes.

wido:
Looks good now @rhtyd 

LGTM for the code

mike-tutkowski:
I walked through the diffs and they LGTM. I did not actually run the code, however, because I'm in the middle of another CloudStack project and I don't have the resources available at the time being.

packaging/build-deb.sh
```diff
@@ -37,7 +37,7 @@ set -e
 # docker run -ti -v /tmp:/src ubuntu:16.04 /bin/bash -c "apt-get update && apt-get install -y dpkg-dev python debhelper openjdk-8-jdk genisoimage python-mysql.connector maven lsb-release devscripts dh-systemd python-setuptools && /src/cloudstack/packaging/build-deb.sh"
 #
 # Ubuntu 14.04
-# docker run -ti -v /tmp:/src ubuntu:14.04 /bin/bash -c "apt-get update && apt-get install -y dpkg-dev python debhelper openjdk-7-jdk genisoimage python-mysql.connector maven lsb-release devscripts dh-systemd python-setuptools && /src/cloudstack/packaging/build-deb.sh"
+# docker run -ti -v /tmp:/src ubuntu:14.04 /bin/bash -c "apt-get update && apt-get install -y dpkg-dev python debhelper openjdk-8-jdk genisoimage python-mysql.connector maven lsb-release devscripts dh-systemd python-setuptools && /src/cloudstack/packaging/build-deb.sh"
```
milamberspace: @rhtyd openjdk-8-jdk doesn't exists on Ubuntu 14.04

packaging/debian/cloudstack-agent.init
```diff
@@ -50,7 +50,7 @@ unset OPTIONS
 mkdir -m 0755 -p "$TMP"
 
 # The first existing directory is used for JAVA_HOME (if JAVA_HOME is not defined in $DEFAULT)
-JDK_DIRS="/usr/lib/jvm/java-7-openjdk-amd64 /usr/lib/jvm/java-7-openjdk-i386 /usr/lib/jvm/java-7-oracle /usr/lib/jvm/java-6-openjdk /usr/lib/jvm/java-6-openjdk-i386 /usr/lib/jvm/java-6-openjdk-amd64 /usr/lib/jvm/java-6-sun"
+JDK_DIRS="/usr/lib/jvm/java-8-openjdk-amd64 /usr/lib/jvm/java-8-openjdk-i386 /usr/lib/jvm/java-8-oracle"
```
milamberspace: @rhtyd So Ubuntu 14.04 doesn't have openjdk-8, the .deb packages for Trusty don't works too with this JDK_DIRS

milamberspace:
I've try to install this PR on a 2 sample topologies without success:
* Ubuntu 14.04.5: needs some tweaks to create the .deb packages, and after installing them, the services don't start because openjdk8 don't exists
* Ubuntu 16.04.1: building the .deb works (and the systemvm template), but cloudstack-management don't start because some symbolic links reference tomcat6 (not exists in 16.04) and the tomcat(cloudstack) config is incorrect (some classpath not found).

This is -1 for this PR with Ubuntu 14.04 and 16.04 (cloudstack debian packages). Trillian don't tests these distribs?

packaging/build-deb.sh
rhtyd: @milamberspace Ubuntu users have options to use the PPA: https://launchpad.net/~openjdk-r/+archive/ubuntu/ppa
Java7 has EOL-ed, Ubuntu users should use JRE8 now?

rhtyd:
@milamberspace thanks, yes there is some issue with Ubuntu as mgmt server so Trillian tests don't run them. Mgmt servers is not supported on ubuntu 16.04 yet; only as kvm hosts. Ubuntu users can use this ppa: https://launchpad.net/~openjdk-r/+archive/ubuntu/ppa which is widely used

packaging/debian/cloudstack-agent.init
rhtyd: Do you think adding the usage of PPA https://launchpad.net/~openjdk-r/+archive/ubuntu/ppa for openjdk-8 should be good enough for Ubuntu users?

milamberspace:
@rhtyd the PPA openjdk can be used probably (even if it's an unsupported packages without warranty), but we need to document this on installation manual (as a requirement to installation CS on Ubuntu 14.04...

Other question about Zulu on Debian 7: not supported too?

rhtyd:
@milamberspace yes Debian7 official main and backport repos don't have openjdk8 yet, even the jessie main repos don't have openjdk8 pkg yet, only jessie-backports has it. For this reason, I used zulu-8... I think zulu-8 can even be used by Ubuntu users as well, they are certified openjdk8 builds from Azul, see http://www.azul.com/downloads/zulu

The blockers for running mgmt server on Ubuntu16.04 are systemd and tomcat8/9 related support, if we can solve the issue or moving away from Tomcat to embedded-Jetty (like we've in developer mode) -- we can simplify packages for a lot of the distributions.

jburwell:
@rhtyd most Java-based systems do not specific a particular JDK/JRE as a dependency in their packages in order to allow administrators to pick the implementation they prefer (e.g. OpenJDK, Sun JDK, IBM JDK, etc).  Should we consider removing those dependencies from ours packages for the same reason?

Also, Travis supports testing [multiple Java versions](https://blog.travis-ci.com/support_for_multiple_jdks/).  If we are not discontinuing support for 1.7, we should have Travis both 1.7 and 1.8.

rhtyd:
With `cs.jdk.version` set to 1.8 for both as source and target versions, generated jars can only be run against JRE1.8. Forward compatibility is not supported (build with 1.8, run on 1.7), we can force the maven compiler plugin to target for 1.7 while set source to 1.8, but I read that it may not work sometimes and I would like to avoid that given 1.7 has EOL-ed.

Without having package/version dependencies fixated on jdk/jre1.8, sysadmins will only find out when the agents/usage server or mgmt server fails to start with an error message such as `Unsupported major.minor version 52.0`

rhtyd:
We've enough LGTMs and test results on this PR. I'll keep this PR open till end of the week to receive any other feedback, suggestions and test results. Thanks.

milamberspace:
I'm not agree with this PR because it's break the support of Ubuntu 14.04 with the default JRE (java7) provide by Ubuntu and the support of Ubuntu 16.04 (java8) the is currently partial for the cloudstack-management.
If you don't know that you need to install a Java 8 via a PPA or manually, you cannot run CS 4.10 on Ubuntu 14.04...
I would not block the PR, but I think at least the installation guide must be update to mention theses points for Ubuntu 14 (need a PPA or manual installation of Java8) and 16 (no support for management)

(Or remove the Ubuntu support entirely?)

rhtyd:
@milamberspace thanks, we can document this in the upgrade notes for Ubuntu 14.04 users to use the widely known/used PPA.

rhtyd:
JFYI, these are the docker images Trillian uses to build packages for CloudStack, they have all the dependencies and jenv (http://www.jenv.be/) installed to automatically switch between java7 or java8 on master or any other branches:
https://hub.docker.com/r/bhaisaab/centos7-cloudstack-slave/
https://hub.docker.com/r/bhaisaab/centos6-cloudstack-slave/
https://hub.docker.com/r/bhaisaab/ubuntu-cloudstack-slave/

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@milamberspace @terbolous I've added the information for Ubuntu users on release notes for 4.10 (preview) now:
http://docs.cloudstack.apache.org/projects/cloudstack-release-notes/en/4.10/upgrade/upgrade-4.9.html#java-8-jre-on-ubuntu
http://docs.cloudstack.apache.org/projects/cloudstack-release-notes/en/4.10/upgrade/upgrade_notes.html#java-8-jre-on-ubuntu

I think with this we don't have any blockers, I'll merge this PR with a final round of testing now. Thanks everyone.

milamberspace:
Thanks for the docs update.

I've tested the PR with my test topology on Ubuntu 14.04 + Openjdk PPA. With the new systemvm (generated from the PR too)
Installation works, Simple deployment works (ssvm, cpvm, RV)

LGTM.

Note : the docker build command with Ubuntu 14.04 don't works if the Openjdk8/PPA is not installed before.

rhtyd:
@blueorangutan test

rhtyd:
Thanks everyone, since we've no outstanding issues or objections and all regression tests across xen/vmware/kvm pass -- I'll proceed with merge this.

syed:
@rhtyd I tried building the latest master I have problems getting the SSVM and the console proxy up. I get the following error 

```
4 localgw=172.31.0.1 private.network.device=eth1 internaldns1=8.8.8.8 dns1=8.8.8.8 nfsVersion=null
Exception in thread "main" java.lang.UnsupportedClassVersionError: com/cloud/agent/AgentShell : Unsupported major.minor version 52.0
        at java.lang.ClassLoader.defineClass1(Native Method)
        at java.lang.ClassLoader.defineClass(ClassLoader.java:800)
        at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
        at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
        at java.net.URLClassLoader.access$100(URLClassLoader.java:71)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:361)
        at java.net.URLClassLoader$1.run(URLClassLoader.java:355)
        at java.security.AccessController.doPrivileged(Native Method)
        at java.net.URLClassLoader.findClass(URLClassLoader.java:354)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:425)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:308)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:358)
        at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:482)
```

The management server is running Java 8 but it looks like the systemvms are still running java 7. How can I upgrade the java version there?

milamberspace:
Yes you need a new systemvm with Java 8 too.

See: http://markmail.org/thread/lj7simcdfynti2ml

(you can build your own systemvm, see:
https://github.com/apache/cloudstack/blob/master/tools/appliance/README.md )

swill:
This is supposed to do it @syed: https://github.com/apache/cloudstack/pull/1888/files#diff-e8e4d31bb5d860c8e7ada3120f758ee9R98

You built a new system VM template or used a new one provided by one of the mirrors?  You have to change the global config to make the new template the default and restart ACS in order for it to be used by default.

Here are the global configs for the different hypervisors:
router.template.hyperv
router.template.kvm
router.template.lxc
router.template.ovm3
router.template.vmware
router.template.xenserver

Let us know...

swill:
@syed if you have not built the system vm templates yet, here are a couple built from master today for XenServer and KVM.

**XenServer**
Name: systemvm-xenserver-4.10.0
Description: systemvm-xenserver-4.10.0
URL: http://objects-east.cloud.ca/v1/29a3d8ee6d4048b7b20d3ca820146a35/systemvm/build-systemvm64-PR_42/systemvm64template-master-4.10.0-xen.vhd.bz2
Zone: Choose the zone where this hypervisor is used
Hypervisor: XenServer
Format: VHD
OS Type: Debian GNU/Linux 7.x (64-bit)
Extractable: no
Password Enabled: no
Public: no
Featured: no
Routing: no

**KVM**
Name: systemvm-kvm-4.10.0
Description: systemvm-kvm-4.10.0
URL: http://objects-east.cloud.ca/v1/29a3d8ee6d4048b7b20d3ca820146a35/systemvm/build-systemvm64-PR_42/systemvm64template-master-4.10.0-kvm.qcow2.bz2
Zone: Choose the zone where this hypervisor is used
Hypervisor: KVM
Format: QCOW2
OS Type: Debian GNU/Linux 7.x (64-bit)
Extractable: no
Password Enabled: no
Public: no
Featured: no
Routing: no

swill:
BTW, something that is worth noting is that even if you use a new system vm template, but the box that the vm template was built on is running Java 7, you will get the same error that @syed just reported.  I forgot to update my Jenkins slave to use Java 8 and I was able to reproduce the problem that @syed experienced with new builds of the system vm template.  Maybe we should try to put a check or something if you try to build the system vm template with the wrong version.  I am sure people will run into this because almost everyone will be using Java 7 in their Jenkins for previous ACS builds and I am sure others will forget to make the change to use Java 8 for those builds.  

Building the ACS RPMs will fail if you don't have Java 8 installed, but the system VM build will succeed even though the resulting template will not be functional...

rhtyd:
@swill yes, please install both jdk7 and jdk8, and setup `jenv` http://www.jenv.be

rhtyd:
@swill I use these docker images for building packages, they have jenv and everything else setup: https://hub.docker.com/u/bhaisaab

Don't use jenv's initialization in .bashrc or .profile, but do that in your Jenkins script:

export PATH="$HOME/.jenv/bin:$PATH"
eval "$(jenv init -)"
java -version
mvn -version

For using my CentOS6 image, you'll need to export this:
# Enable maven on bash
export PATH=/opt/rh/maven30/root/usr/bin:/opt/rh/rh-java-common/root/usr/bin:$PATH

wido:
We will have to review building the .DEB packages as well. I made a start by building in Docker and building for 14.04 and 16.04 with the systemd differences between them.

I however really don't have the time right now to dig into this and configure Jenkins to get it working properly.

serg38:
We need to change one more setting for noredist build to work with updated iControl version
in  install-non-oss.sh   
mvn install:install-file -Dfile=cloud-iControl.jar      -DgroupId=com.cloud.com.f5     -DartifactId=icontrol        -Dversion=1.0   -Dpackaging=jar
to
mvn install:install-file -Dfile=cloud-iControl.jar      -DgroupId=com.cloud.com.f5     -DartifactId=icontrol        -Dversion=12.1   -Dpackaging=jar

rhtyd:
@serg38 I've been maintaining the non-oss libs here: https://github.com/rhtyd/cloudstack-nonoss
You can get the latest v12.1 library that is Java8 compatible from above, it also includes the install-non-oss.sh script, you can simply clone the repo and run it before building/packaging (in Jenkins too), in fact Travis jobs actually do that: https://github.com/apache/cloudstack/blob/master/tools/travis/install.sh#L28

syed:
I got the master working now with the systemvm template http://packages.shapeblue.com.s3-eu-west-1.amazonaws.com/systemvmtemplate/4.10/systemvm64template-4.10-xen.vhd.bz2

@rhtyd would the systemvm template from http://jenkins.buildacloud.org/ also work? Also, we need to update the wiki https://cwiki.apache.org/confluence/display/CLOUDSTACK/How+to+build+CloudStack with the newer URLs. I would prefer it be jenkins.buildacloud.org.

rhtyd:
@syed yes, if the systemvmtemplates were built from master after this PR was merged. There is no secret sauce in the systemvmtemplate that I've uploaded on packages.shapeblue.com S3 bucket. Checkout the readme file at tools/appliance, that's more updated than any wiki page.

swill:
Any suggestions to resolve this?  I have built from master using JDK8 and have setup my `cloudstack.repo` to point to my build.  I have been doing this successfully for the past month.

My management server is on CentOS6.8.

When I run: `sudo yum -y install cloudstack-management cloudstack-usage`

It fails with:
```
Error: Package: cloudstack-management-4.10.0.0-SNAPSHOT.el7.centos.x86_64 (cloudstack)
           Requires: iptables-services
Error: Package: cloudstack-common-4.10.0.0-SNAPSHOT.el7.centos.x86_64 (cloudstack)
           Requires: python(abi) = 2.7
           Installed: python-2.6.6-66.el6_8.x86_64 (@updates)
               python(abi) = 2.6
           Available: python-2.6.6-64.el6.x86_64 (base)
               python(abi) = 2.6
           Available: python34-3.4.5-1.el6.i686 (epel)
               python(abi) = 3.4
```

rhtyd:
@swill build el7 (centos7) packages on centos7, and el6 (centos6) packages on centos6. You may use my docker images, this is what Trillian/blueorangutan uses for building (PR) packages: https://hub.docker.com/u/bhaisaab/

swill:
Wow, these builds are getting complicated now.  I will see if I can get Java8 installed on my CentOS6 Jenkins slaves so I can get my existing build pipelines working again.  Thanks...

swill:
@rhtyd I built the Java 8 version on a CentOS6 slave and I am trying to install it on a CentOS 6.8 host and I am getting a similar issue.

The details are here:  https://lists.apache.org/thread.html/d771df6a4ff07638362ab1343af69d6ab5d35a6cb4a9925b3077aa55@%3Cdev.cloudstack.apache.org%3E

This seems to be related to this PR.  I have not been able to get a Java 8 version of ACS to work yet.

swill:
I found the issue.  `jenv` was not setting the `JAVA_HOME` correctly, so the following logic in `/etc/cloudstack/management/classpath.conf` was getting hit.
```
if ([ -z "$JAVA_HOME" ] || [ ! -d "$JAVA_HOME" ]) && [ -d /usr/lib/jvm/jre-1.7.0 ]; then
     export JAVA_HOME=/usr/lib/jvm/jre-1.7.0
fi
```

I modified this section to be the following and everything is working now.
```
if ([ -z "$JAVA_HOME" ] || [ ! -d "$JAVA_HOME" ]) && [ -d /usr/lib/jvm/jre-1.8.0 ]; then
     export JAVA_HOME=/usr/lib/jvm/jre-1.8.0
fi
```

swill:
Maybe we should be defaulting this file to Java 8 now???

wido:
@swill: Yes we should! I think that was an oversight. Can you send a PR for this against master?

PR #1893, opened 2017-01-06
Improve misleading translations

niwo:


milamberspace:
@niwo We use Transifex for the translation of Web UI. Please change the localization strings on Transifex

https://www.transifex.com/ke4qqq/CloudStack_UI/410_messagesjson/

PR #1898, opened 2017-01-09
CLOUDSTACK-8862: Lock on volume before attempting to attach it.

ProjectMoon:
It is possible to attach a single volume to multiple instances by running the commands close to one another. In KVM, this can be verified by checking the `virsh dumpxml` for the instances once the volume attaches are complete. This PR adds a lock on the volume ID in the orchestration step to make sure the method is only executed one at a time for any given volume ID. So in the scenario of someone sending simultaneous volume attach requests, one should succeed and the others will fail after hitting the check added in 8aa34d048ab3e459de6d254049f3a31dde42b50d.


karuturi:
@ProjectMoon @anshul1886  you are working on the same bug. Can you please collaborate and put a single PR?

ProjectMoon:
Closed in favor of #1900

PR #1812, opened 2016-12-02
CLOUDSTACK-9650: Allow starting VMs regardless of cpu/memory cluster.…

koushik-das:
…disablethreshold setting

Introduced a global configuration flag 'cluster.threshold.enabled'. By default the flag is true.
If the value is false, then a VM can be started in a cluster even if the cluster thresholds are
crossed. However, for a new VM deployment the cluster threshold will always be honoured.

PR #1899, opened 2017-01-10
CLOUDSTACK-9650: Allow starting VMs regardless of cpu/memory cluster.…

koushik-das:
…disablethreshold setting

Fixed the scope of configuration flag 'cluster.threshold.enabled' introduced as part of PR#1812 to global

### DIFF
--- a/api/src/com/cloud/deploy/DeploymentClusterPlanner.java
+++ b/api/src/com/cloud/deploy/DeploymentClusterPlanner.java
@@ -55,7 +55,7 @@ public interface DeploymentClusterPlanner extends DeploymentPlanner {
             "true",
             "Enable/Disable cluster thresholds. If disabled, an instance can start in a cluster even though the threshold may be crossed.",
             false,
-            ConfigKey.Scope.Zone);
+            ConfigKey.Scope.Global);
 
     /**
      * This is called to determine list of possible clusters where a virtual


koushik-das:
@ustcweizhou please review

ustcweizhou:
LTGM
smoke test is not necessary

serg38:
LGTM

koushik-das:
Merging this

api/src/com/cloud/deploy/DeploymentClusterPlanner.java
```diff
@@ -55,7 +55,7 @@
             "true",
             "Enable/Disable cluster thresholds. If disabled, an instance can start in a cluster even though the threshold may be crossed.",
             false,
-            ConfigKey.Scope.Zone);
+            ConfigKey.Scope.Global);
```
rhtyd: This refers to cluster level setting, admins should be allowed to override this at the cluster level. Please change this to ConfigKey.Scope.Cluster.

rhtyd: @koushik-das Ignore, the above comment. But, I think making this global takes away the option to override this at a zone level? Why remove this?

rhtyd:
@koushik-das there are no tests, Java related code changes ought to require regression tests, even if it may appear to not need them. Let's not work towards degrading the branch health.

koushik-das:
@rhtyd Tests are present in #1812

rhtyd:
@koushik-das there are no test results on this PR, also can you explain why you've removed the option to override this setting at the zone level?

koushik-das:
@rhtyd If you had checked #1812 you wouldn't have asked these questions :)
This config was introduced in #1812 and the scope was incorrectly put as zone, the config is meant to be a global one.

rhtyd:
@koushik-das as a RM I want to ensure that all changes, irrespective of their size, are merged after a QA process.

For UI/translation/docs changes tests are not necessary. Please understand that it's difficult to review and keep track of each and every PR in detail, and help and support from the community is appreciated to make this work. We have agreed to gate the branches, and you've went ahead with merging without following proper guidelines; testing against one hypervisor which have errors that may/may-not be related to a PR are not good enough, given we've systems like Trillian to test against at least three major hypervisors.

Lastly, I disagree, I think the setting should be available per zone for admins to override this on a per-zone basis.

PR #1901, opened 2017-01-11
CLOUDSTACK-9405: add details parameter in listDomains API to reduce the execution time

ustcweizhou:
The resource limitation causes long time while execute listDomains API. The resources are not needed in some cases so that the execution time will be reduced.

### DIFF
--- a/api/src/org/apache/cloudstack/api/command/admin/domain/ListDomainsCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/admin/domain/ListDomainsCmd.java
@@ -16,10 +16,15 @@
 // under the License.
 package org.apache.cloudstack.api.command.admin.domain;
 
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
+
 import org.apache.log4j.Logger;
 
 import org.apache.cloudstack.api.APICommand;
 import org.apache.cloudstack.api.ApiConstants;
+import org.apache.cloudstack.api.ApiConstants.DomainDetails;
 import org.apache.cloudstack.api.BaseListCmd;
 import org.apache.cloudstack.api.Parameter;
 import org.apache.cloudstack.api.ResponseObject.ResponseView;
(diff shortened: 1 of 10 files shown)


serg38:
Thanks @ustcweizhou . We have been planning to address this for a while. Should we create a smoke test to verify validity of the modified listDomain call?

ustcweizhou:
@serg38 what do you want to validate if it is necessary ?

api/src/org/apache/cloudstack/api/command/admin/domain/ListDomainsCmd.java
```diff
@@ -53,6 +59,12 @@
                description = "If set to false, list only resources belonging to the command's caller; if set to true - list resources that the caller is authorized to see. Default value is false")
     private Boolean listAll;
 
+    @Parameter(name = ApiConstants.DETAILS,
+               type = CommandType.LIST,
+               collectionType = CommandType.STRING,
+               description = "comma separated list of host details requested, value can be a list of [ all, resource, min]")
```
serg38: Can you clarify what would be permitted values for details parameter? If they are only "min", resource", "all" then we might want to change the name and description of this parameter to reflect true purpose. What is the difference between 'resource' and 'all'?

ustcweizhou: @serg38 currently no difference. However if we want to add more type of details, then it will be useful. for example, we add a new detail type: stats in our branch.

the defition is copied from listhosts API (you can see host details in the decription, I will change it).

serg38:
Thanks @ustcweizhou LGTM for code review

rhtyd:
LGTM, @ustcweizhou can you add a marvin test to validate api changes?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Test LGTM, known intermittent errors seen. This can be accepted now.

PR #1904, opened 2017-01-12
CLOUDSTACK-9729: Use latest Nuage client.

fmaximus:
CloudStack root pom change to use Amazon WS 11.1.16
caused our client to fail, as it was depending on classes,
which are not not present anymore.
Latest client version uses Gson instead.

BUG-ID: CLOUDSTACK-9729
Bugfix-for: master


krissterckx:
LGTM

karuturi:
LGTM merging now

PR #1905, opened 2017-01-13
CLOUDSTACK-9738: [Vmware] Optimize vm expunge process for instances with vm snapshots

nvazquez:
## Description
It was noticed that expunging instances with many vm snapshots took a look of time, as hypervisor received as many tasks as vm snapshots instance had, apart from the delete vm task. We propose a way to optimize this process for instances with vm snapshots by sending only one delete task to hypervisor, which will delete vm and its snapshots

## Use cases

1. deleteVMsnapohsot-> no changes to current behavior
2. destroyVM with expunge=false ->  no actions to VMsnaphsot is performed at the moment. When VM cleanup thread is executed it will perform the same sequence as (3). If instance is recovered before expunged by the cleanup thread it will remain intact with VMSnapshot chain present
3. destroyVM with expunge=true:
   * Vmsnaphsot is marked with removed timestamp and state = Expunging in DB
   * VM is deleted in HW

### DIFF
--- a/api/src/com/cloud/vm/UserVmService.java
+++ b/api/src/com/cloud/vm/UserVmService.java
@@ -75,14 +75,14 @@ public interface UserVmService {
     /**
      * Destroys one virtual machine
      *
-     * @param userId
-     *            the id of the user performing the action
      * @param vmId
      *            the id of the virtual machine.
+     * @param expunge
+     *            indicates if vm should be expunged
      * @throws ConcurrentOperationException
      * @throws ResourceUnavailableException
      */
-    UserVm destroyVm(long vmId) throws ResourceUnavailableException, ConcurrentOperationException;
+    UserVm destroyVm(long vmId, boolean expunge) throws ResourceUnavailableException, ConcurrentOperationException;
 
     /**
      * Resets the password of a virtual machine.
--- a/api/src/com/cloud/vm/snapshot/VMSnapshotService.java
+++ b/api/src/com/cloud/vm/snapshot/VMSnapshotService.java
@@ -46,4 +46,11 @@ public interface VMSnapshotService {
         ConcurrentOperationException;
 
     VirtualMachine getVMBySnapshotId(Long id);
+
+    /**
+     * Delete vm snapshots only from database. Introduced as a Vmware optimization in which vm snapshots are deleted when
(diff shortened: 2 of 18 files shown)


ozhanrk:
Hi, we currently hold vm snapshots like as vm backup, even if we destroy vm with expunge. So that we could restore vm snapshot on future time. Does this pr has any effect on our scenario?

nvazquez:
Hi @ozhanrk, 

In this PR, vm snapshots get deleted when the instance gets deleted from hypervisor, on task sent by the vm cleanup thread. Can it be possible that you were referring to volume snapshots instead of vm snapshots for restoring vms?

Thanks

ozhanrk:
Hi @nvazquez;
You are right i am referring to volume based snapshots.

Thanks

nvazquez:
@ozhanrk nice, then this PR shouldn't affect your workflow

serg38:
@rhtyd @karuturi Can we run tests on this PR and merge on success ?

karuturi:
There are some conflicts after the latest merge. Can you resolve them?

nvazquez:
@karuturi sure, done!

karuturi:
@blueorangutan package

karuturi:
@blueorangutan test

karuturi:
Hi @nvazquez 
Did the behavior in step2 change with this PR or is it same as earlier?
Can you please add the testing already done?

nvazquez:
Hi @karuturi,
On Vmware, step 2 reduces tasks sent to hypervisor by sending only one task which actually deletes vm snapshots and the vm itself instead of sending one for each vm snapshot plus the one for the vm. For other hypervisors, it would work the same. I refactored this PR adding tag [Vmware] on title too.

nvazquez:
@karuturi regarding tests done, we've compared the number of delete tasks sent to Vmware hypervisor on expunging a vm with `N` vm snapshots before and after this optimization, reducing from `N+1` tasks to only `1`. Is a marvin test required for this case?

Thanks

karuturi:
Thanks for the clarification. Its a nice enhancement. LGTM for the changes.
Marvin tests are always welcome :)

nvazquez:
@karuturi thanks!

PR #1706, opened 2016-10-12
CLOUDSTACK-9500: remove the IP from the databag if it is not used

vilisseranen:
https://issues.apache.org/jira/browse/CLOUDSTACK-9500

This removes the IP from the databags ips.json and forwardingrules.json when the IP is unused.

PR #1871, opened 2016-12-27
[4.9] CLOUDSTACK-9692: Fix password server issue in redundant VRs

ustcweizhou:
The password server in RVRs has wrong parameters as the gateway of guest nics is None.
In this case, we should get the gateway from /var/cache/cloud/cmdline.
This issue is caused by commit 45642b83821ce0ecd6d4cddb76a77a2481e54d9a

PR #1907, opened 2017-01-16
Fix public IPs not being removed from the VR when deprovisioned

swill:
This PR replaces #1706.  It does not remove the IP from the database, but it does deprovision the IP correctly from the VR when the public IP is removed.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
@@ -106,6 +106,10 @@ class CsAddress(CsDataBag):
                 ip.setAddress(address)
                 logging.info("Address found in DataBag ==> %s" % address)
 
+                if not address['add'] and not ip.configured():
+                    logging.info("Skipping %s as the add flag is set to %s " % (address['public_ip'], address['add']))
+                    continue
+
                 if ip.configured():
                     logging.info(
                         "Address %s on device %s already configured", ip.ip(), dev)


systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
```diff
@@ -659,7 +663,10 @@ def is_guest_gateway(self, bag, ip):
         gw = interface.get_gateway()
         logging.info("Interface has the following gateway ==> %s", gw)
 
-        if bag['nw_type'] == "guest" and rip == gw:
+        guest_gw = self.config.cmdline().get_guest_gw()
+        logging.info("Interface has the following gateway ==> %s", guest_gw)
+
+        if bag['nw_type'] == "guest" and (rip == gw or rip == guest_gw):
```
ustcweizhou: this works for isolated networks, but not for vpc tiers.

remibergsma:
LGTM We run this code in production without issues for a few months now (https://github.com/MissionCriticalCloud/cosmic/pull/114)

systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
rafaelweingartner: This is more like a curiosity than anything else. 
What do you think about using only `return bag['nw_type'] == "guest" and (rip == gw or rip == guest_gw)` ? It does not feel that we need the conditional here.

swill: I find that makes the code a bit harder to read if we did that. Then it is not as obvious you are returning a boolean. This stuff is already complicated, for two lines of code, I don't think it makes sense to make it harder to understand. :)

rafaelweingartner: ok, got it.

ustcweizhou: @swill @rafaelweingartner 
#1871 has already been merged into 4.9 and master, so I think guest_gw is not necessary here.

swill: @ustcweizhou I am not sure that changes removes the need for this, but it looks like it could potentially.  Maybe I can run some tests to see if it can be removed given #1871 has been merged...

swill:
I have squash merged the changes that @ustcweizhou requested.  Would you mind reviewing again so we can get this important fix into 4.10.  Thanks...

@rafaelweingartner, it looks like I lost your code review in the process of doing the squashed commit.  Can you review and confirm your previous review?  

Thanks guys...

rafaelweingartner:
@swill LGTM for the changes.

The changes introduced now are different from the last ones, right? Now it is basically the addition of a log and the skipping of a processing flow if condition `not address['add'] and not ip.configured()` is met.

I am guessing this the PR changed because of the changes introduced with #1871, right?

swill:
@rafaelweingartner the only change was to remove the extra gateway verification if it was a 'guest' network because this is now handled by the code in #1871.

```
+        guest_gw = self.config.cmdline().get_guest_gw()
+        logging.info("Interface has the following gateway ==> %s", guest_gw)
+
+        if bag['nw_type'] == "guest" and (rip == gw or rip == guest_gw):
```

The rest of the code is the same as the original pull request.

rafaelweingartner:
Ok, great.
Thanks for the explanation.

swill:
This has the required reviews and both @remibergsma (sbp) and us (cloud.ca) have been using this fix in production.

kiwiflyer:
tag:mergeready

rhtyd:
LGTM @karuturi this is ready for merge

ustcweizhou:
code LGTM

PR #1909, opened 2017-01-17
CLOUDSTACK-9740 : Search for secondary IP of NIC that is attached to …

niteshsarda:
Search for secondary IP of NIC that is attached to an instance is not working.

If instance has more than one secondary IP assigned and administrator attempts to search that list, all IPs are displayed instead of being filtered by the search string.
in web UI:
CCP dashboard -> Instances -> <Instance name> -> NICs -> ”View Secondary IPs” -> Search for a IP, but the search returns all the IPs in the list

**Screenshot before applying fix :** 

screenshot before applying fix

**Screenshot after applying fix :**

screenshot after applying fix

### DIFF
--- a/api/src/com/cloud/network/NetworkService.java
+++ b/api/src/com/cloud/network/NetworkService.java
@@ -175,6 +175,8 @@ public interface NetworkService {
     /* lists the nic informaton */
     List<? extends Nic> listNics(ListNicsCmd listNicsCmd);
 
+    List<? extends NicSecondaryIp> listSecondaryNics(ListNicsCmd listNicsCmd);
+
     Map<Network.Capability, String> getNetworkOfferingServiceCapabilities(NetworkOffering offering, Service service);
 
     IpAddress updateIP(Long id, String customId, Boolean displayIp);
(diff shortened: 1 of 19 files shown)


niteshsarda:
screenshot after applying fix
screenshot before applying fix

ustcweizhou:
@niteshsarda in my opinion, what we should change is the search processing, not the process on creating nic response.
to be clear, in server/src/com/cloud/network/NetworkServiceImpl.java
```
public List<? extends Nic> listNics(ListNicsCmd cmd)
```

SudharmaJain:
@niteshsarda I see a problem here. The listNics command can be used to list both primary and secondary ips. In case if we pass nicid parameter with it, i feel we should use keywork to filter secondary nics. If only virtualmachineid is given, we should use it to filter primary IPs.

niteshsarda:
@ustcweizhou @SudharmaJain : This PR got automatically closed while pushing the updated code. So created a new PR for this issue : 
https://github.com/apache/cloudstack/pull/2126

I have incorporated all the comments from this PR in the new PR. Please check the new PR.

PR #1910, opened 2017-01-18
CLOUDSTACK-9748:VPN Users search functionality broken

Ashadeepa:
VPN Users search functionality broken 
If you try to search VPN users with it’s user name, you will not be able to search.

Fixed the same.

### DIFF
--- a/server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
+++ b/server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
@@ -582,6 +582,7 @@ public class RemoteAccessVpnManagerImpl extends ManagerBase implements RemoteAcc
     public Pair<List<? extends VpnUser>, Integer> searchForVpnUsers(ListVpnUsersCmd cmd) {
         String username = cmd.getUsername();
         Long id = cmd.getId();
+        String keyword = cmd.getKeyword();
         Account caller = CallContext.current().getCallingAccount();
         List<Long> permittedAccounts = new ArrayList<Long>();
 
@@ -596,7 +597,7 @@ public class RemoteAccessVpnManagerImpl extends ManagerBase implements RemoteAcc
 
 
         sb.and("id", sb.entity().getId(), SearchCriteria.Op.EQ);
-        sb.and("username", sb.entity().getUsername(), SearchCriteria.Op.EQ);
+        sb.and("username", sb.entity().getUsername(), SearchCriteria.Op.LIKE);
         sb.and("state", sb.entity().getState(), Op.IN);
 
         SearchCriteria<VpnUserVO> sc = sb.create();
@@ -611,6 +612,10 @@ public class RemoteAccessVpnManagerImpl extends ManagerBase implements RemoteAcc
 
         if (username != null) {
             sc.setParameters("username", username);
+        }
+
+        if (keyword!= null) {
+            sc.setParameters("username",  "%" + keyword + "%");
         }
 
         Pair<List<VpnUserVO>, Integer> result = _vpnUsersDao.searchAndCount(sc, searchFilter);


server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
```diff
@@ -613,6 +614,10 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
             sc.setParameters("username", username);
         }
 
+        if (keyword!= null) {
+            sc.setParameters("username",  "%" + keyword + "%");
+        }
+
         Pair<List<VpnUserVO>, Integer> result = _vpnUsersDao.searchAndCount(sc, searchFilter);
```
syed: What if both username and keyword are set? Shouldn't we prefer the username? Also what is `keyword` anyways?

server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
```diff
@@ -596,7 +597,7 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
 
 
         sb.and("id", sb.entity().getId(), SearchCriteria.Op.EQ);
-        sb.and("username", sb.entity().getUsername(), SearchCriteria.Op.EQ);
+        sb.and("username", sb.entity().getUsername(), SearchCriteria.Op.LIKE);
```
syed: Any reason for changing this to `LIKE` instead of `EQ` ?

server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
Ashadeepa: @syed : If both are set, we are giving preference to keyword as the search will happen using keyword query param. This is in according with other existing APIs. Keyword is the request parameter for : List by keyword functionality.

server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
Ashadeepa: @syed Used LIKE instead of EQ as username will be used in a WHERE clause to search for a specified pattern in a column using the keyword.

ustcweizhou: I vote EQ instead LIKE here.
you may add new rules here in the next part, for example.

+        if (keyword != null) {
+            final SearchCriteria<VpnUserVO> ssc = _vpnUsersDao.createSearchCriteria();
+            ssc.addOr("username", SearchCriteria.Op.LIKE, "%" + keyword + "%");
+            ssc.addOr("state", SearchCriteria.Op.EQ, "%" + keyword + "%");
+            sc.addAnd("username", SearchCriteria.Op.SC, ssc);
+        }

Ashadeepa: @ustcweizhou : Thanks but unfortunately, addOr & addAnd methods are now deprecated, so I avoided using the same.

Also, LIKE with work only when keyword is set. When only username is set,      `sc.setParameters("username", username); `
will get executed with the exact match i.e. same as how EQ will work!

SudharmaJain:
Code changes looks good. LGTM

ustcweizhou:
@Ashadeepa I just tested it. It seems it is not working as I expected if username and keyword are both set.
I suggest to use the following
```
        sb.and("username", sb.entity().getUsername(), SearchCriteria.Op.EQ);
        sb.and("keyword", sb.entity().getUsername(), SearchCriteria.Op.LIKE);
...
        if (keyword!= null) {
             sc.setParameters("keyword",  "%" + keyword + "%");
         }
```

Ashadeepa:
test_input_ss
test_output_ss
[TestResults_CLOUDSTACK-9748.txt](https://github.com/apache/cloudstack/files/779944/TestResults_CLOUDSTACK-9748.txt)

Ashadeepa:
@ustcweizhou : Hi, I had tested the same and added my test results in the file TestResults_CLOUDSTACK-9748.txt. 

As per my previous comment, when both username & keyword are set, users with username like 'keyword' will be the expected result. 

Also, your suggested won't work as "keyword" in absent in the VPNUserVO class as there is no such column named as keyword. I tried it, got an exception.

ustcweizhou:
@Ashadeepa your comments in previous comment is not what I expect.
In my opinion, if username and keyword are both set, we should search by both condition, but only one.

You might missed the line in your testing.
```
        sb.and("keyword", sb.entity().getUsername(), SearchCriteria.Op.LIKE);
```
it works fine in my testing.

Ashadeepa:
@ustcweizhou : Thanks. I have made the changes.

New PR : https://github.com/apache/cloudstack/pull/1957.

PR #1916, opened 2017-01-24
[4.9] CLOUDSTACK-9462: Build packages on Ubuntu 12.04/16.04 and support tomcat6/tomcat7

ustcweizhou:
Changes
(1) seperate debian/control and debian/rules for Ubuntu 12.04/16.04
(2) add python-setuptools in Build-Depends
(3) seperate cloudstack-management.service and cloudstack-management.default for CentOS7 and Ubuntu 16.04
(4) add server7-ssl.xml and server7-nonssl.xml in management installation
(5) link /usr/share/cloudstack-management/lib and /usr/share/cloudstack-management/bin to correct path (tomcat6 or tomcat7)
(6) link /etc/cloudstack/management/server.xml to correct file path (server-nonssl.xml or server7-nonssl.xml)
(7) remove *.zip from .gitignore to avoid build error caused by missing /vhds/test.vhd.zip

Instruction
(1) build packages on Ubuntu 12.04/16.04: packaging/build_deb.sh
    Output on Ubuntu 12.04: cloudstack-management_4.9.3.0-SNAPSHOT~precise_all.deb
    Output on Ubuntu 16.04: cloudstack-management_4.9.3.0-SNAPSHOT~xenial_all.deb
(2) setup tomcat6/tomcat7 on management server: cloudstack-setup-management --tomcat7

Already tested on Ubuntu 12.04 and Ubuntu 16.04, upgraded from 4.7.1 to 4.9.3.0-SNAPSHOT

### DIFF
--- a/debian/cloudstack-management.postinst
+++ b/debian/cloudstack-management.postinst
@@ -60,7 +60,26 @@ if [ "$1" = configure ]; then
 
     chmod 0640 ${CONFDIR}/${DBPROPS}
     chgrp cloud ${CONFDIR}/${DBPROPS}
-    invoke-rc.d tomcat6 stop || true
+
+    if [ -d "/usr/share/tomcat7" ]; then
+        rm -f /usr/share/cloudstack-management/bin
+        rm -f /usr/share/cloudstack-management/lib
+        ln -s /usr/share/tomcat7/bin /usr/share/cloudstack-management/bin
+        ln -s /usr/share/tomcat7/lib /usr/share/cloudstack-management/lib
+        invoke-rc.d tomcat7 stop || true
+        if [ ! -f "/etc/cloudstack/management/server.xml" ]; then
+            ln -s /etc/cloudstack/management/server7-nonssl.xml /etc/cloudstack/management/server.xml
+        fi
+    elif [ -d "/usr/share/tomcat6" ]; then
+        rm -f /usr/share/cloudstack-management/bin
+        rm -f /usr/share/cloudstack-management/lib
+        ln -s /usr/share/tomcat6/bin /usr/share/cloudstack-management/bin
+        ln -s /usr/share/tomcat6/lib /usr/share/cloudstack-management/lib
+        invoke-rc.d tomcat6 stop || true
+        if [ ! -f "/etc/cloudstack/management/server.xml" ]; then
+            ln -s /etc/cloudstack/management/server-nonssl.xml /etc/cloudstack/management/server.xml
+        fi
+    fi
 
     # Add jdbc MySQL driver settings to db.properties if not present
     grep -s -q "db.cloud.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS} || sed -i -e "\$adb.cloud.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS}
--- a/.gitignore
+++ b/.gitignore
@@ -51,7 +51,6 @@ tools/cli/build/
 *.jar
 *.war
 *.mar
-*.zip
 *.iso
 *.tar.gz
 *.tgz
--- a/debian/cloudstack-management.install
+++ b/debian/cloudstack-management.install
@@ -21,7 +21,8 @@
 /etc/cloudstack/management/ehcache.xml
 /etc/cloudstack/management/server-ssl.xml
 /etc/cloudstack/management/server-nonssl.xml
-/etc/cloudstack/management/server.xml
+/etc/cloudstack/management/server7-ssl.xml
+/etc/cloudstack/management/server7-nonssl.xml
 /etc/cloudstack/management/classpath.conf
 /etc/cloudstack/management/db.properties
 /etc/cloudstack/management/tomcat6-ssl.conf
@@ -33,6 +34,7 @@
 /etc/cloudstack/management/log4j-cloud.xml
 /etc/cloudstack/management/tomcat-users.xml
 /etc/cloudstack/management/context.xml
+/etc/default/cloudstack-management
 /etc/init.d/cloudstack-management
 /etc/security/limits.d/cloudstack-limits.conf
 /etc/sudoers.d/cloudstack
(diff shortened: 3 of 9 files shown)


wido:
Why the ubuntu12 control file? Just to keep Ubuntu 12.04 support in there for CloudStack Management servers?

Since that would mean you'd have to rename files when building DEB packages for older distributions.

ustcweizhou:
@wido We will use packaging/build_deb.sh to build packages. It will get the os version, then copy corresponding files (debian control/rules) according to os version  , add os version (eg, precise/xenial) to tags .
This patch is for LTS 4.9 branch. We have 4.9.2 for Ubuntu 16.04 but it does not work with tomcat7. On the other hand, we do not provide packages for ubuntu 12.04/tomcat6. This patch will make them work well.
For 4.10/master, we can remove the files for ubuntu 12.04 if we decide to not support it anymore.

debian/cloudstack-management.postinst
```diff
@@ -60,7 +60,26 @@ if [ "$1" = configure ]; then
...
+    elif [ -d "/usr/share/tomcat6" ]; then
+        rm -f /usr/share/cloudstack-management/bin
+        rm -f /usr/share/cloudstack-management/lib
+        ln -s /usr/share/tomcat6/bin /usr/share/cloudstack-management/bin
+        ln -s /usr/share/tomcat6/lib /usr/share/cloudstack-management/lib
+        invoke-rc.d tomcat6 stop || true
+        if [ ! -f "/etc/cloudstack/management/server.xml" ]; then
+            ln -s /etc/cloudstack/management/server-nonssl.xml /etc/cloudstack/management/server.xml
+        fi
+    fi
```
wido: Not sure if we this is the best way. Usually creating directories and moving symlinks is not the way to go in postinst files.

When cleaning packages or upgrading this might break things. Can you think of a different way?

debian/rules
```diff
@@ -149,7 +150,7 @@ override_dh_installinit:
 	dh_installinit -pcloudstack-management -pcloudstack-agent -pcloudstack-usage --onlyscripts --no-start
 
 override_dh_systemd_enable:
-	dh_systemd_enable -pcloudstack-agent -pcloudstack-usage
+	dh_systemd_enable -pcloudstack-management -pcloudstack-agent -pcloudstack-usage
```
wido: Shouldn't we somehow clean up any left behind script in /etc/init.d of the management server?

ustcweizhou: @wido we should if we do not support ubuntu 12.04 anymore. I prefer to remove it in 4.10. I can create another PR for it. This PR is only for 4.9 branch.

debian/cloudstack-management.postinst
ustcweizhou: @wido I was thinking of other ways but could not find. Do you have any idea?
We can remove this change and ask users to execute "cloudstack-setup-management --tomcat7" after each upgrade. It seems not wise.

I have tested the upgrade from 4.7.1 to 4.9.3.0-SNAPSHOT on ubuntu 12.04 and 16.04 without any issue.
I did not test the os upgrade or tomcat upgrade.

wido:
Good point to remove Ubuntu 12.04 in 4.10.

I didn't notice this is for 4.9. Not really paid attention to it.

I personally dislike the moving of the control file while building the deps though. Don't see another way however.

rhtyd:
@ustcweizhou can you check/fix the failures?
@blueorangutan package

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test ubuntu kvm-ubuntu

rhtyd:
@borisstoyanov can you check Trillian failures?
@ustcweizhou can you check Jenkins/Travis failures? I'm in favour of removing 12.04 support (for mgmt server) in 4.9+, since the support for Ubuntu 12.04 ends in April 2017.

ustcweizhou:
@rhtyd then I will close this PR, and create another PR for 4.10/master.
Someone who run with 4.9 LTS and want to have 16.04 systemd support, can feel free to merge this PR into their branch.

borisstoyanov:
@rhtyd I remember Trillian having some issues when using ubuntu images, but cannot find the build failure. 
Will trigger a new one to investigate. 
@blueorangutan test ubuntu kvm-ubuntu

rhtyd:
@ustcweizhou I think it would be good to add support for 16.04/systemd to run mgmt server. But Given 12.04 will EOL in next 2.5 months, let's have support for 16.04 in 4.9 as well, drop support for 4.9.

ustcweizhou:
@rhtyd As 4.9 is LTS version, it seems not good to drop ubuntu 12.04 support in next minor release (4.9.3 ?), some customers may still use ubuntu 12.04 as management server.

I will aim to add this support from 4.10.

rhtyd:
@ustcweizhou makes sense, in that case please close this PR and open a new PR for 4.10/master that adds support for cloudstack mgmt server on ubuntu 16.04, dropping support for 12.04 etc.

wido:
@ustcweizhou @rhtyd Seems good to me. Let's drop 12.04 support in 4.10

Java 8 anyway which is not supported on 12.04, so we will need to drop it.

If we just can get rid of 14.04 then I'm very happy (joke). Everything systemd :+1:

borisstoyanov:
we got this failure when installing cloudstack-management

```
root@pr1916-t848-kvm-ubuntu-mgmt1:~# apt-get install cloudstack-management
Reading package lists... Done
Building dependency tree
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 cloudstack-management : Depends: init-system-helpers (>= 1.18~) but 1.14 is to be installed
E: Unable to correct problems, you have held broken packages.
```
installing on 14.04

ustcweizhou:
@borisstoyanov  It is weird I did not specify the minimal version of init-system-helpers in debian/control. Did you build the packages on ubuntu 16.04 and install it on ubuntu 14.04 ?

by the way, could you tell me what command you use to build the packages ? dpkg-buildpackage or packaging/build-deb.sh ?

borisstoyanov:
used dpkg-buildpackage 
the docker container just for the build is 16.04
but @rhtyd reports he's able to install this package on his local 14.04

ustcweizhou:
close this as new PR for apache/4.10 : #1950

PR #1727, opened 2016-10-24
CLOUDSTACK-9539: Support changing Service offering for instance with VM Snapshots

nvazquez:
## Actual behaviour

CloudStack doesn't support changing service offering for vm instances which have vm snapshots, they should be removed before changing service offering.
## Goal

Extend actual behaviour by supporting changing service offering for vms which have vm snapshots. In that case, previously taken snapshots (if reverted) should use previous service offering, future snapshots should use the newest.
## Proposed solution:
1. Adding `service_offering_id` column on `vm_snapshots` table: This way snapshot can be reverted to original state even though service offering can be changed for vm instance.
   NOTE: Existing vm snapshots are populated on update script by:
   `UPDATE vm_snapshots s JOIN vm_instance v ON v.id = s.vm_id SET s.service_offering_id = v.service_offering_id;`
2. New vm snapshots will use instance vm service offering id as `service_offering_id`
3. Revert to vm snapshots should use vm snapshot's `service_offering_id` value.
## Example use case:
- Deploy vm using service offering A
- Take vm snapshot -> snap1 (service offering A)
- Stop vm
- Change vm service offering to B
- Revert to VM snapshot snap 1
- Start vm

It is expected that vm has service offering A after last step

PR #1918, opened 2017-01-30
Management Server UI (VM statistics page) CPU Utilized value is incorrect.

jayakarteek:
VMware CPU Utilised  value is in Mhz , in MS value is displayed directly.
If CPU utilised is 500Mhz,  UI value is displayed as 500 %

Planned to restrict changes in resource layer VmwareResource.java, but cpu allocated to VM by VMware is not same as  allocated in the service offering.

If CPU limit flag is set to false in  service offerings  the Allocated CPU can go beyond what is assigned.
So picking actual user assigned Total CPU from database for calculating percentage.  
It is expected that percentage would go beyond 100 percent.

 so changes are made in UserVmJoinDaoImpl.java

### DIFF
--- a/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -5085,21 +5085,23 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
         final String guestMemUseStr = "summary.quickStats.guestMemoryUsage";
         final String memLimitStr = "resourceConfig.memoryAllocation.limit";
         final String memMbStr = "config.hardware.memoryMB";
+        final String allocatedCpuStr = "summary.runtime.maxCpuUsage";
 
         ObjectContent[] ocs =
-                hyperHost.getVmPropertiesOnHyperHost(new String[] {"name", numCpuStr, cpuUseStr ,guestMemUseStr ,memLimitStr ,memMbStr, instanceNameCustomField});
+                hyperHost.getVmPropertiesOnHyperHost(new String[] {"name", numCpuStr, cpuUseStr ,guestMemUseStr ,memLimitStr ,memMbStr,allocatedCpuStr ,instanceNameCustomField});
         if (ocs != null && ocs.length > 0) {
             for (ObjectContent oc : ocs) {
                 List<DynamicProperty> objProps = oc.getPropSet();
                 if (objProps != null) {
                     String name = null;
                     String numberCPUs = null;
-                    String maxCpuUsage = null;
+                    double maxCpuUsage = 0;
                     String memlimit = null;
                     String memkb = null;
                     String guestMemusage = null;
                     String vmNameOnVcenter = null;
                     String vmInternalCSName = null;
+                    double allocatedCpu = 0;
                     for (DynamicProperty objProp : objProps) {
                         if (objProp.getName().equals("name")) {
                             vmNameOnVcenter = objProp.getVal().toString();
@@ -5111,13 +5113,17 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                         }else if (objProp.getName().equals(numCpuStr)) {
                             numberCPUs = objProp.getVal().toString();
                         } else if (objProp.getName().equals(cpuUseStr)) {
-                            maxCpuUsage = objProp.getVal().toString();
+                            maxCpuUsage = NumberUtils.toDouble(objProp.getVal().toString());
                         } else if (objProp.getName().equals(memLimitStr)) {
                             memlimit = objProp.getVal().toString();
                         } else if (objProp.getName().equals(memMbStr)) {
                             memkb = objProp.getVal().toString();
+                        } else if (objProp.getName().equals(allocatedCpuStr)){
+                            allocatedCpu  = NumberUtils.toDouble(objProp.getVal().toString());
                         }
                     }
+
+                    maxCpuUsage = (maxCpuUsage/allocatedCpu)*100;
                     new VirtualMachineMO(hyperHost.getContext(), oc.getObj());
                     if (vmInternalCSName != null) {
                         name = vmInternalCSName;
@@ -5184,7 +5190,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                             }
                         }
                     }
-                    vmResponseMap.put(name, new VmStatsEntry( NumberUtils.toDouble(memkb)*1024,NumberUtils.toDouble(guestMemusage)*1024,NumberUtils.toDouble(memlimit)*1024, NumberUtils.toDouble(maxCpuUsage), networkReadKBs, networkWriteKBs, NumberUtils.toInt(numberCPUs), "vm"));
+                    vmResponseMap.put(name, new VmStatsEntry( NumberUtils.toDouble(memkb)*1024,NumberUtils.toDouble(guestMemusage)*1024,NumberUtils.toDouble(memlimit)*1024, maxCpuUsage, networkReadKBs, networkWriteKBs, NumberUtils.toInt(numberCPUs), "vm"));
                 }
             }
         }


rafaelweingartner:
@jayakarteek this problem is happening only for VMware, right?

@nvazquez could you confirm this problem here? I believe you have access to an ACS system that orchestrates VMware deployment.

nvazquez:
Hi @jayakarteek @rafaelweingartner,

I confirm issue, I tested using an instance with 1 CPU 500Mhz. I'll add screenshots for cases when CPU utilization is low and high:

First, with little utilization, vSphere shows 51Mhz CPU Utilization, while on Management Server statistics shows 51% CPU utilization
cpu51
cpu512

Then, ran a loop to increase CPU utilization increasing it to ~1609Mhz and Management Server statistics shows 1609% CPU Utilization,
cpu1600
cpu16002

Hope this helps!

jayakarteek:
HI  raferweingartner, nvazquez<https://github.com/nvazquez>,

Yes as specified in mail,
The issue is specific to VMware.

--JayaKarteek

rafaelweingartner:
@nvazquez, this helped.

I checked the “VmwareResource.java” at line 5113 is obtained the CPU usage of VM. From @jayakarteek explanations I understood that the usage does not come as %; VMware systems show the value as a continuous value of CPU used in MHz.  Does not the object “objProps” (created at line 5092) contain the configured CPU for the VM? So, the proper conversion could be applied on “VmwareResource.java” at line 5186, making the POJO (VmStatsEntry) consistent; otherwise, there is a variable called “cpuUtilization” that gives the idea of percentage/ratio of usage, but in reality, it is carrying the real used valued of CPU in MHz.

I also understood the issue regarding the CPU cap, but do not you think that if users select a service offering of 500MHz and their VMs are consuming 1GHz, it would be interesting to show that the VM is using more than they would expect? The cloud provider is being very generous in this case :)

nvazquez:
@rafaelweingartner sorry for the last example, actually I think we've found a bug on change service offering, as I've been using that instance for changing its service offering when it had vm snapshots for PR #1727. Looking at `/proc/cpuinfo` I could see that CPU speed was 2.6Ghz, as vCenter shows, and not 500Mhz as listed on Management Server, so CPU speed was not properly set after changing service offering. I'll work on this.

nvazquez:
@rafaelweingartner please ommit my last comment, service offering didn't use reserve CPU, so it didn't limit CPU speed to 500Mhz, it used all the available frecuency of the CPU instead.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

rafaelweingartner:
@nvazquez no problem. I think it is ok to use everything that is available because the service offering is not restricting the CPU usage. 

@nvazquez would it be possible for you to check if the object “objProps” (created at line 5092) contains the configured CPU for the VM in “VmwareResource.java”? I think we should try to put the logic to fix that value there, instead of creating a POJO with inconsistent data.

nvazquez:
@rafaelweingartner I checked 'objProps' object and it doesn't contain the property. However, I think it can be easily retrieved by adding "summary.config.cpuReservation" on array at line 5089 (from Vmware docs: https://pubs.vmware.com/vi3/sdk/ReferenceGuide/vim.vm.Summary.ConfigSummary.html) for method `getVmPropertiesOnHyperHost`

rafaelweingartner:
That is a good idea @nvazquez, thanks for the help here ;)

What do you think guys (@nvazquez, @borisstoyanov, @jayakarteek)? if we use the "summary.config.cpuReservatio" as @nvazquez suggested we could remove the changes proposed on (UserVmJoinDaoImpl.java) and avoid having a POJO with inconsistent data. Moreover, we concentrate VMware related logic and operation in VMware classes.

jayakarteek:
Hi @nvazquez , @borisstoyanov , @rafaelweingartner ,

Pardon me for very late reply.
I tried with new property "summary.config.cpuReservation" , but it deals with the CPU reservation for the VM that is configured by VMWare.

It is different from the CPU speed, we define in CPU service offerings, as explained by below link

http://www.vfrank.org/2013/09/19/understanding-vmware-reservations-limits-and-shares/

jayakarteek:
@nvazquez , @borisstoyanov , @rafaelweingartner 

Attached screen shots. where I set CPU Reservation as 500 Mhz using VSphere-Client , and same is obtained by the  API call.

acs_vmstatistics
acs_vmstatistics2

rafaelweingartner:
@jayakarteek I did not understand your doubt.

I see by your SS that the reservation field contains the speed configured in your service offering. From what I understood so far, this is what you want right? Looking at your debug SS, I also see that you can get this property on the ACS side. It is just because of this property (the configured CPU speed for the service offering) that you moved the calculation/definition of cpuUtilization to another layer in ACS components for VMware VMs.  So, I am not seeing a problem to execute the calculations there (VmwareResource.java),  when you create the POJO with the VM usage information.

Could you elaborate it a bit more?

jayakarteek:
@rafaelweingartner 

The CPU.Reservation value by **default** is 0, Also the value is set in VMware not from the ACS UI.
this gives ambiguity about the CPU allocated and Utilised.

Use case 1 :
CPU.Reservation is 0 by default, In this case we cannot calculate % as the value is always zero.

Use case 2:
CPU.Reservation is set by VMware admin say 500Mhz.
But while creating vm-instance  using ACS UI , admin selected service offering as Medium Instance which is 1Ghz CPU. so CPU utilisation will be wrong.
In case say cloud environment has 30 instances , then Vmware admin has to manually set the CPU.Reservation value for each of the VM Instance which  is cumbersome.

rafaelweingartner:
@jayakarteek thanks for the explanation
I thought that ACS would configure those parameters automatically. The way you described it feels like ACS is not limiting the use of CPU for VMs created in VMware systems. 

I was checking the code, and there is this piece here: `com.cloud.hypervisor.vmware.util.VmwareHelper.setBasicVmConfig(VirtualMachineConfigSpec, int, int, int, int, int, String, boolean)`. At line 634 ACS is setting the CPU limit if CPU cap is enabled. Isn't that information available to be retrieved later (when we get the stats of a VM)? I know, this will only work with CPU cap. However, we could retrieve that CPU limit configuration, if it is '-1', we could then retrieve the CPU speed of the host (and use it to calculate the percentage of usage), is it possible? Or do you think it brings much trouble for the code?

@nvazquez could you also provide some feedback here?

nvazquez:
@jayakarteek @rafaelweingartner what about retrieving CPU performance metric for usage using [PerfomanceManager](https://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.PerformanceManager.html)? It is already used in `VmwareResource.getVmStats` lines 5134-5185 to retrieve vm network metrics. I think that similar as it is done for these metrics, we must query for available metrics and look for [usage CPU counter](https://pubs.vmware.com/vsphere-60/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fcpu_counters.html) which already comes as percentage for a given interval which would have to be low. This way we don't have to calculate depending on limit flag. What do you think about this approach?

rafaelweingartner:
This seems to be a great idea @nvazquez.
What do you think @jayakarteek ?

jayakarteek:
@rafaelweingartner 
we can retrieve the CPU speed of the host but the utilisation will have mismatch.
create VM with  Medium Instance having 1000 Mhz , and VM is utilising 200Mhz of CPU then utilisation is 20%.
But if we consider host CPU capacity with  2400Mhz  then utilisation in 8.3%.

@nvazquez 

I checked CPUCounter API doc, even in this case values will have some missmatch.
as per the below document with the usage formula they are using.
https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/cpu_counters.html

rafaelweingartner:
@jayakarteek from what I understood so far, there will always be a mismatch.

For instance, the way it is implemented so far, if a VM uses a service offering that provides 1000 MHz; and if CPU cap is not used, it (the VM) can use up to the real CPU of the host. If real CPU on the host is 2000 MHz, then it would show 200% of usage. Unless we want this behavior, the use of CPUCounter seems a better approach to avoid having an inconsistent POJO in the code. You said it might not be ideal because the CPUCounter (per its documentation) does not seem to consider the amount of virtual CPU allocated in MHz while calculating usage, right?

I just started this discussion because of the POJO that has an attribute that sometimes is set using percentage values and some other as continuous values. My goal was to highlight this and maybe discuss some solution.

I was reading the documentation link you sent, what are the differences between “cpuentitlement” and “reservedCapacity”; I mean, I can read the description of the parameters 

>  cpuentitlement is computed based on an ideal scenario in which all virtual machines are completely busy and the load is perfectly balanced across all hosts

but the description for me seems different from what I can understand from "entitlement" when I read "CPU entitlement" I think of CPU power that the CPU is entitled and may not be using or getting for one or other reason. However, the description may sound like something else (this value being based on the load of hosts/clusters). 

 Would not one of these two configurations have the CPU in MHz allocated for the VM?

jayakarteek:
@rafaelweingartner  
Tried using both cpuentitlement cpu reservedCapacity but retrieved values from vmware are zero.

While going through docs , I saw **summary.runtime.maxCpuUsage** property which is max cpu that is allocated per vm, this looks accurate.

If I enable CPU cap in ACS UI , the cpu allocated returned by this property matches with the service offering.
If CPU cap is not enabled then cpu allocated value is determined by VMware.


So I am going with the above property using which I will calculate the CPU % and populate the POJO.

rafaelweingartner:
Awesome! This seems a good approach to deal with the issue.
@jayakarteek and @nvazquez thanks for the effort.

@jayakarteek sorry for all of the stubbornness ;)
You really listened and discussed, and I think we reached a better solution here.

jayakarteek:
@rafaelweingartner , @nvazquez 

Review the code changes.

server/src/com/cloud/api/query/dao/UserVmJoinDaoImpl.java
```diff
@@ -196,6 +196,7 @@ public UserVmResponse newUserVmResponse(ResponseView view, String objectName, Us
             // stats calculation
             VmStats vmStats = ApiDBUtils.getVmStatistics(userVm.getId());
             if (vmStats != null) {
+
```
rafaelweingartner: I think you can remove this extra line here

nvazquez:
LGTM for testing. Tested on Vmware 6.0, I attach some screenshots for test scenarios, in which vm was deployed using service offering, infinite loop was executed to increase CPU utilization and then stopped

- Test case 1: Service offering not limiting CPU speed







- Test case 2: Service offering limiting CPU speed






@rafaelweingartner thanks for your effort in this PR
@jayakarteek can you please rebase master branch and squash your commits?

jayakarteek:
@nvazquez  @rafaelweingartner 
Done.

jayakarteek:
tag:mergeready

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -5111,13 +5113,17 @@ private VirtualMachineGuestOsIdentifier translateGuestOsIdentifier(String cpuArc
...
                         } else if (objProp.getName().equals(memLimitStr)) {
                             memlimit = objProp.getVal().toString();
                         } else if (objProp.getName().equals(memMbStr)) {
                             memkb = objProp.getVal().toString();
+                        } else if (objProp.getName().equals(allocatedCpuStr)){
+                            allocatedCpu  = NumberUtils.toDouble(objProp.getVal().toString());
                         }
                     }
+
+                    maxCpuUsage = (maxCpuUsage/allocatedCpu)*100;
```
rhtyd: Add a check that allocatedCpu is > 0.0?

rhtyd:
In order to avoid exceptions can be use a default value (if applicable) when using NumberUtils.<method>(string, default_value) (default_value can be 0 or 0.0)?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
jayakarteek: @rhtyd 
AllocatedCpu is constant and will be greater then zero.

karuturi:
@kiwiflyer Thanks for the tag :)

PR #1919, opened 2017-01-30
CLOUDSTACK-9763: set ssh public key when creating vm data

serbaut:


rafaelweingartner:
@serbaut  I would only like to make a few questions to see if I understood the issue here, before evaluating the code any further.

When you first deploy the VM, the keys are delivered to the newly deployed VM, right? The problem only happens on reboots?

Do we need to always keep sending the same access keys? Is not one time enough? Of course, if the keys are changed, we should send them; but otherwise, it feels that we should not need to keep sending them.

serbaut:
The VPC VR maintains metadata (http://docs.cloudstack.apache.org/projects/cloudstack-administration/en/4.8/virtual_machines/user-data.html) as static files in /var/www/html/metadata. When a VR is destroyed and recreated (by e.g. "restart with cleanup") this metadata is rebuilt by createVmDataCommandForVMs(). public-keys is missing from that function so it becomes empty after the rebuild and a request for latest/meta-data/public-keys no longer returns the correct key.

This PR adds public-key to the rebuild.

ustcweizhou:
I agree with @serbaut , different from password which is applied to VR only once, the public keys should be set in VR each time when we recreate a VR.

this LGTM+1

rafaelweingartner:
Thanks for the explanation @serbaut. That is exactly why I asked; so, it is not a reboot/restart per se. It is a re-deploy; the old virtual machine is destroyed and a new one is deployed (or maybe a reset of the VHD of the VM). By VM here I mean VR (which at the end of the day is a VM). 

@serbaut, could you add these explanations on the PR description and Jira ticket (https://issues.apache.org/jira/browse/CLOUDSTACK-9763)? At least for me, this was not clear.

Thanks for the fix ;)
Code LGTM

pdion891:
This is usefull when using VPC.

+1

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 746
 Hypervisor xenserver
 NetworkType Advanced
 Passed=112
 Failed=0
 Skipped=12

**Failed tests:**

**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_01_verify_libvirt
test_04_verify_guest_lspci
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vm_snapshots.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_list_ids_parameter.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_metrics_api.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_routers_network_ops.py
test_disk_offerings.py

rhtyd:
@serbaut  can you rebase against latest master

rhtyd:
@blueorangutan package

rhtyd:
LGTM.

PR #1926, opened 2017-02-02
CLOUDSTACK-9768: Time displayed for events in UI is incorrect

jayantpatil1234:
Time displayed for events in UI is incorrect. Let's say, when we login using Japanese language the time displayed in the events is GMT instead of JST. However with English language the time is JST, as expected.
Example: 
Time is displayed in the event is 10:40, if you are logged in using English language. 
Whereas, time in the event shows 19:40 If you login with Japanese language.

### DIFF
--- a/api/src/org/apache/cloudstack/api/ApiConstants.java
+++ b/api/src/org/apache/cloudstack/api/ApiConstants.java
@@ -263,6 +263,7 @@ public class ApiConstants {
     public static final String ISO_ID = "isoid";
     public static final String TIMEOUT = "timeout";
     public static final String TIMEZONE = "timezone";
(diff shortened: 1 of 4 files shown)


borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test centos6

borisstoyanov:
@blueorangutan test kvm-centos6

borisstoyanov:
@blueorangutan test

borisstoyanov:
I've checked current behavior and no matter of the language you select time of the events is displayed in the local time timezone, which I think It's the expected result here. 
LGTM on the code changes, marvin tests looking good as well.

karuturi:
LGTM. merging

PR #1700, opened 2016-10-04
CLOUDSTACK-9359: IPv6 for Basic Networking

wido:
This PR is a proposal for adding very basic IPv6 to Basic Networking. The main goal of this PR is that the API returns a valid IPv6 address over which the Instance is reachable.

The GUI will show the IPv6 address after deployment of the Instance.

screenshot from 2016-10-03 16 34 56

If the table VLAN has a proper IPv6 CIDR configured the DirectPodBasedNetworkGuru will calculate the IPv6 Address the Instance will obtain using EUI-64 and SLAAC: https://tools.ietf.org/search/rfc4862

In this case the _vlan_ table contained:

<pre>mysql> select * from vlan \G
*************************** 1. row ***************************
                 id: 1
               uuid: 90e0716c-5261-4992-bb9d-0afd3006f476
            vlan_id: vlan://untagged
       vlan_gateway: 172.16.0.1
       vlan_netmask: 255.255.255.0
        description: 172.16.0.10-172.16.0.250
          vlan_type: DirectAttached
     data_center_id: 1
         network_id: 204
physical_network_id: 200
        ip6_gateway: 2001:980:7936:112::1
           ip6_cidr: 2001:980:7936:112::/64
          ip6_range: NULL
            removed: NULL
            created: 2016-07-19 20:39:41
1 row in set (0.00 sec)

mysql></pre>

It will then log:

<pre>2016-10-04 11:42:44,998 DEBUG [c.c.n.g.DirectPodBasedNetworkGuru] (Work-Job-Executor-1:ctx-1975ec54 job-186/job-187 ctx-0d967d88) (logid:275c4961) Found IPv6 CIDR 2001:980:7936:112::/64 for VLAN 1
2016-10-04 11:42:45,009 INFO  [c.c.n.g.DirectPodBasedNetworkGuru] (Work-Job-Executor-1:ctx-1975ec54 job-186/job-187 ctx-0d967d88) (logid:275c4961) Calculated IPv6 address 2001:980:7936:112:4ba:80ff:fe00:e9 using EUI-64 for NIC 6a05deab-b5d9-4116-80da-c94b48333e5e</pre>

The template has to be configured accordingly:
- No IPv6 Privacy Extensions
- Use SLAAC
- Follow RFC4862

This is also described in: https://cwiki.apache.org/confluence/display/CLOUDSTACK/IPv6+in+Basic+Networking

The next steps after this will be:
- Security Grouping to prevent IPv6 Address Spoofing
[...]

PR #1927, opened 2017-02-02
ipv6: Set IPv6 CIDR and Gateway in 'nic' profile

wido:
Without this information a NPE might be triggered when starting a VR, SSVM or CP
as this information is read from the 'nics' table and causes a NPE.

During deployment we should set the IPv6 Gateway and CIDR for the NIC object so that
it is persisted to the database.

### DIFF
--- a/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -1523,6 +1523,8 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
             nic.setIPv4Address(profile.getIPv4Address());
             nic.setAddressFormat(profile.getFormat());
             nic.setIPv6Address(profile.getIPv6Address());
+            nic.setIPv6Cidr(profile.getIPv6Cidr());
+            nic.setIPv6Gateway(profile.getIPv6Gateway());
             nic.setMacAddress(profile.getMacAddress());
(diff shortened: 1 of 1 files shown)


wido:
This regression was introduced with PR #1700, so my bad.

Found this myself during tests. Assume nobody is using it yet as it has only been there in master.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test centos7 vmware-60u2

rhtyd:
LGTM, let's merge this blocker @karuturi

karuturi:
@rhtyd, Can @blueorangutan run basic network tests?

karuturi:
also, as @wido said, this happens only when ipv6 is set. So, its not a blocker. I will wait and see if someone can run basic network BVT.
@wido, Will running BVT verify anything with respect to ipv6? do we have any tests? 
If you or @mike-tutkowski can run the BVT on basic networking, it will be really helpful.
Thanks,
Rajani

rhtyd:
@karuturi blueorangutan by default can run all the smoke/component tests -- but by default only in an advanced zone (advanced network). If your question was around testing this in basic zone (shared network), I'll need to discuss with my team and get back to you.

karuturi:
yes, my question was about testing this PR in basic zone. Thanks.

wido:
@karuturi @rhtyd: This only applies when you have IPv6 in Basic Networking enabled and are also using it.

I can't think of another use-case where this goes bad.

karuturi:
merging this.

PR #1934, opened 2017-02-06
[CLOUDSTACK-9772] Template: perform a HEAD request to check file size from a URL

marcaurele:
Signed-off-by: Marc-Aurèle Brothier <m@brothier.org>

### DIFF
--- a/utils/src/main/java/com/cloud/utils/UriUtils.java
+++ b/utils/src/main/java/com/cloud/utils/UriUtils.java
@@ -35,8 +35,6 @@ import java.util.List;
 import java.util.ListIterator;
 import java.util.StringTokenizer;
 
-import javax.net.ssl.HttpsURLConnection;
-
 import org.apache.commons.httpclient.Credentials;
 import org.apache.commons.httpclient.HttpClient;
 import org.apache.commons.httpclient.HttpException;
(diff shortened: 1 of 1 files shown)


borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

borisstoyanov:
@marcaurele there are some marvin tests in extract and delete template, could you please have a look?

utils/src/main/java/com/cloud/utils/UriUtils.java
```diff
@@ -200,37 +198,27 @@ public static String getUpdateUri(String url, boolean encrypt) {
     }
 
     // Get the size of a file from URL response header.
-    public static Long getRemoteSize(String url) {
-        Long remoteSize = (long)0;
+    public static long getRemoteSize(String url) {
```
remibergsma: @marcaurele I think this needs to be changed back to Long, or else the method cannot be found as it is called with type Long.

marcaurele: @remibergsma if you look at all the calls to getRemoteSize, they all expect a `long`. Moreover the result is neither `null` but a long, 0 or an exception, so it's best to return a `long` directly IMO.

marcaurele:
@borisstoyanov Dunno what's wrong with the failed tests. Its' failing at https://github.com/apache/cloudstack/blob/master/plugins/storage/image/default/src/org/apache/cloudstack/storage/datastore/driver/CloudStackImageStoreDriverImpl.java#L83. The log says that the ssvm endpoint was not ready.
Could you try to launch another bueorangutan test suite?

We're running that fix in production since a week and haven't hit a problem yet.

borisstoyanov:
OK @marcaurele, if you're 100% sure it works as expected it could be that we need to redesign the test itself, I guess it needs deeper investigation of the failure itself. 
@blueorangutan test

remibergsma:
What about signed S3 urls that people may supply? If you want to register a template and have it download from a S3 bucket, you need to specify a signed url to provide the authentication. But that signature is valid for either GET or HEAD requests, not both. 

When you use S3 as secondary storage, you automatically get a GET pre-signed url from CloudStack that you cannot use to register it as a new template. Something to think about..

See: http://stackoverflow.com/questions/15717230/pre-signing-amazon-s3-urls-for-both-head-and-get-verbs

marcaurele:
@remibergsma Good point, I was aware of that difference, which I think doesn't help to make systems reliable.
Another improvement would be to remove this function and refactor the code to read sizes from the template objects in the DB or on the GET requests when downloading them.

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 752
 Hypervisor xenserver
 NetworkType Advanced
 Passed=106
 Failed=5
 Skipped=12

**Failed tests:**
* test_deploy_vm_iso.py

 * test_deploy_vm_from_iso Failed

* test_list_ids_parameter.py

 * ContextSuite context=TestListIdsParams>:setup Failing since 20 runs

* test_volumes.py

* test_ssvm.py

 * test_03_ssvm_internals Failed

* test_routers_network_ops.py

 * ContextSuite context=TestRedundantIsolateNetworks>:teardown Failing since 4 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_01_verify_libvirt
test_04_verify_guest_lspci
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vm_snapshots.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_metrics_api.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

rhtyd:
@blueorangutan package

rhtyd:
@marcaurele are you happy to further testing/merging, or you think this can break use-cases for people using URLs where HEAD may not be supported by hosting server?

marcaurele:
@rhtyd I will amend the code to support the case with S3 pre-signed URL which will potentially fail on a HEAD operation. It will *try* to perform a HEAD operation and if it does not work, it tries to do a normal GET. I'll ping you when it's updated.
cc @remibergsma @wido @borisstoyanov

marcaurele:
@rhtyd ready on my side with the fallback on a GET request if the HEAD fails

wido:
Code LGTM

With one side-note though. I still see a lot of webservers not sending Content-Length nor Content-Type (not used here)

marcaurele:
@wido ok, then I'll improve it to calculate the content size in case of a GET request based on the input stream.

marcaurele:
@wido improvement added on `GET`

rhtyd:
Update on this? @marcaurele

marcaurele:
@rhtyd I did the code update regarding @wido comment. So now it tries to download using the HEAD request and reads the information. If that does not work, for example because the URL is a pre-signed S3 download one, it performs a GET and tries to read the content length from the header. If it is not found, it will read the length out of the downloaded content.
IMO it's just waiting for @wido comment, and/or other people from the community

utils/src/main/java/com/cloud/utils/UriUtils.java
```diff
@@ -201,40 +199,53 @@ public static String getUpdateUri(String url, boolean encrypt) {
         return details;
     }
 
+    public static void main(String[] args) {
```
rhtyd: Remove `main`.

rhtyd:
@blueorangutan package

utils/src/main/java/com/cloud/utils/UriUtils.java
marcaurele: gotcha! Done

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Test LGTM.
I found the failed test case was an environment issue where user VM could not be ssh-ed into. The test code does not have a `%s` to capture the error failing with unable to translate the symbol. I'll look into the test code issue separately.

Merging this based on three code lgtm and test results. Thanks.

PR #1937, opened 2017-02-10
CLOUDSTACK-9779 : Releasing secondary guest IP fails with error VM nic Ip x.x.x.x is mapped to load balancing rule

niteshsarda:
ISSUE 
=================
Releasing secondary guest IP fails with error VM nic Ip x.x.x.x is mapped to load balancing rule

REPRO STEPS
==================
1. Create two isolated guest networks with same CIDR
2. Deploy VMs on both networks
3. Acquire secondary IP on NICs of both VMs and make sure they have the same value, user can input the IP address.
4. Configure Loadbalancing rule on one of the secondary IP address and try releasing the other secondary IP address.
5. The operation would fail

EXPECTED BEHAVIOR
==================
Secondary IP address should be released if there are no LB rules associated with it.

ACTUAL BEHAVIOR
==================
Releasing secondary IP address even if there are no LB rules associated with it.

### DIFF
--- a/server/src/com/cloud/api/query/dao/AccountJoinDaoImpl.java
+++ b/server/src/com/cloud/api/query/dao/AccountJoinDaoImpl.java
@@ -162,8 +162,8 @@ public class AccountJoinDaoImpl extends GenericDaoBase<AccountJoinVO, Long> impl
         response.setTemplateAvailable(templateAvail);
 
         // Get stopped and running VMs
-        response.setVmStopped(account.getVmStopped());
-        response.setVmRunning(account.getVmRunning());
+        response.setVmStopped(account.getVmStopped()!=null ? account.getVmStopped() : 0);
+        response.setVmRunning(account.getVmRunning()!=null ? account.getVmRunning() : 0);
 
         //get resource limits for networks
         long networkLimit = ApiDBUtils.findCorrectResourceLimit(account.getNetworkLimit(), account.getId(), ResourceType.network);


server/src/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -852,7 +852,8 @@ public boolean releaseSecondaryIpFromNic(long ipAddressId) {
                 throw new InvalidParameterValueException("Can' remove the ip " + secondaryIp + "is associate with static NAT rule public IP address id " + publicIpVO.getId());
```
ustcweizhou: line 849. The issue also happen on static nat. Could you fix it as well ?

server/src/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -852,7 +852,8 @@ public boolean releaseSecondaryIpFromNic(long ipAddressId) {
...
             }
 
-            if (_lbService.isLbRuleMappedToVmGuestIp(secondaryIp)) {
+            List<Long> lbRuleIdList = _firewallDao.listIdByNetworkAndPurposeAndNotRevoked(network.getId(), Purpose.LoadBalancing);
```
ustcweizhou: this two checks can be done by a join search in LoadBalancerVMMapDao.java

ustcweizhou:
files are also changed from 100644 → 100755 which is not necessary

server/src/com/cloud/network/NetworkServiceImpl.java
niteshsarda: Will take this up in separate ticket.

ustcweizhou: @niteshsarda I've created a PR for it: https://github.com/apache/cloudstack/pull/1947

server/src/com/cloud/network/NetworkServiceImpl.java
niteshsarda: @ustcweizhou  : As per your comment, do you expect Null and Empty check for list variable "lbRuleIdList " to be moved in LoadBalancerVMMapDao.java class ?

ustcweizhou: @niteshsarda sorry what I said is not clear. You added two methods in Daos: listIdByNetworkAndPurposeAndNotRevoked and listByLoadBalancerIdVmIdAndInstanceIp, I suggest to implement it by one method in LoadBalancerDaoImpl.java

niteshsarda:
@ustcweizhou : I have created a new PR for this issue https://github.com/apache/cloudstack/pull/1972 .

Please check. Also, I have incorporated all the comments which you have provided in this PR in the new PR.

PR #1938, opened 2017-02-10
CLOUDSTACK-9780: Fixed the default JAVA_HOME value to be Java8 if not set

swill:
Now that PR-1888 is merged, Java8 is required.  Unfortunately, the file pushed to `/etc/cloudstack/management/classpath.conf` on ACS install will default the version to Java7 instead of Java8 if JAVA_HOME is unset.  This fix sets the default to Java8 if JAVA_HOME is not set.

### DIFF
--- a/client/tomcatconf/classpath.conf.in
+++ b/client/tomcatconf/classpath.conf.in
@@ -34,8 +34,8 @@ for vendorconf in "@MSCONF@"/vendor/* ; do
 	CLASSPATH=$vendorconf:$CLASSPATH
 done
 export CLASSPATH
-if ([ -z "$JAVA_HOME" ] || [ ! -d "$JAVA_HOME" ]) && [ -d /usr/lib/jvm/jre-1.7.0 ]; then
-     export JAVA_HOME=/usr/lib/jvm/jre-1.7.0
+if ([ -z "$JAVA_HOME" ] || [ ! -d "$JAVA_HOME" ]) && [ -d /usr/lib/jvm/jre-1.8.0 ]; then
+     export JAVA_HOME=/usr/lib/jvm/jre-1.8.0
 fi
 PATH=$JAVA_HOME/bin:/sbin:/usr/sbin:$PATH
 export PATH


swill:
@rhtyd, @wido, @nvazquez, @rafaelweingartner please review...

This fixes an issue with #1888.

nvazquez:
Thanks @swill!

LGTM

rafaelweingartner:
@Swill, I have a question (actually two) :)

If “JAVA_HOME” is not set, or if “JAVA_HOME” content does not exist, we check if there is a java installation. I have a doubt about this last path we are checking. Is this path “/usr/lib/jvm/jre-1.8.0” default (some sort of standard)?

I ask because (at least on Debian, or at least the Debian I am using) whenever I install using “aptitude”, the java versions were added to folders following the pattern: “/usr/lib/jvm/java-<version>-openjdk-<processorArchiteture>/jre/”. Of course, this is for OpenJDK installed with “aptitude”.

Are we assuming that the user when installing manually (e.g. installing the JRE from Oracle) will put Java JRE files in a folder like “/usr/lib/jvm/jre-1.8.0”?

This will also only work for Java JRE 1.8.0, if users install Java JRE 1.8.1 and use the version as the name of the folder, this would not work. Of course, all can be fixed by setting the JAVA_HOME. 

Would not it be better to cause an exception and stop the deployment with a message saying that we require the “JAVA_HOME” to be set? Then, it feels that we will be consistent; otherwise, I can imagine a user complaining that she/he has installed the Java 8 (1.8.x; where x > 0), but still ACS does not work. 

What do you think?

swill:
So the short answer is that I ran into this issue when I upgraded and since `jenv` did not correctly set the `JAVA_HOME` directory, I ran into this, so I don't have a deep understanding for all the details.  I made this change in my environment and everything worked.  I understand this is not a good answer, but that is what I have.

I don't install to that path either, but for some reason it was populated in my environment.  Can you do an `$ ll /usr/lib/jvm/` and post what you get?  I am curious what it would return.  

Here is the result from my CentOS 6.8 setup.

image

The way I see it right now, people WILL run into this issue.  This at least reduces the number of people who have problems.

rafaelweingartner:
Ok, got it.
Maybe in CentOS logical links get created during the installation of Java. Or maybe during the installation of ACS. The computer  I am using is not used to run ACS, but I use Java here, and this is why I asked you.

my output: 
```
lrwxrwxrwx  1 root root   20 Dec  1 12:48 java-1.6.0-openjdk-amd64 -> java-6-ope                                                                                                                                                             njdk-amd64/
-rw-r--r--  1 root root 2387 Dec  1 12:48 .java-1.6.0-openjdk-amd64.jinfo
lrwxrwxrwx  1 root root   20 Feb  7 20:10 java-1.7.0-openjdk-amd64 -> java-7-ope                                                                                                                                                             njdk-amd64/
-rw-r--r--  1 root root 2439 Feb  7 20:10 .java-1.7.0-openjdk-amd64.jinfo
drwxr-xr-x  5 root root 4096 Feb 10 14:00 java-6-openjdk-amd64/
drwxr-xr-x  3 root root 4096 Feb 10 13:59 java-6-openjdk-common/
drwxr-xr-x  5 root root 4096 Feb 10 13:57 java-7-openjdk-amd64/
```
Anyways, I understand that you are following the current pattern. So, we may let this problem (if it turns out to be a problem) for our future selves. If people start having problems when updating to java JRE version bigger than 1.8.0.

Having said that, LGTM

borisstoyanov:
@BlueOrangutan package

borisstoyanov:
@blueorangutan package

ustcweizhou:
@swill I agree with @rafaelweingartner . I suggest to use the following setting ?
JAVA_HOME=$(readlink -f /usr/bin/java | sed "s:bin/java::")

ustcweizhou:
If someone use ubuntu 16.04/14.04, could you test #1916 as well ?

borisstoyanov:
@blueorangutan test ubuntu kvm-ubuntu

swill:
@ustcweizhou I am not comfortable with that change. If more than one java is installed and Java 8 is not the default, then this will incorrectly set java_home (if I understand correctly). 

I will leave my pr as is. It solves the problem the same way it was done in the past. If you would like to introduce a new approach and test it against the different distributions, please open a new PR with your new implementation.

ustcweizhou:
@swill in my point of view, users should set java8 as default if they have multiple versions (we can add it in user guide).
Although this PR is a ideal solution, this LGTM +1.

karuturi:
@borisstoyanov can you post ubuntu test results?

borisstoyanov:
@karuturi unfortunately we've hit an issue with the ubuntu images, didn't manage to deploy the Trillian env..

karuturi:
good point @swill. Thanks @borisstoyanov.
Since the fix is no different from the earlier version for 1.7, I am assuming it should work for 1.8 as well on ubuntu.
merging this now

PR #1948, opened 2017-02-17
[CLOUDSTACK-9793] Faster IP in subnet check

ProjectMoon:
This change removes the conversion from IPNetwork to list in one of the router scripts. This makes the router faster at processing static NAT rules, which can prevent timeouts when attaching or detaching IPs.

With the `list` conversion, it has to potentially check a list of 65536 IP strings multiple times. We assume that the comparison implemented in the IPNetwork is far more efficient. We have seen speed-up from 218 seconds to enable static NAT with 18 IPs on the router to 2 or 3 seconds by removing this cast. This also fixes a potential bug where adding IPs to a router time out because the scripts are taking too long. 218 seconds, for example, is beyond the timeout on the KVM agent for script execution, and then all enableStaticNat operations will fail.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsAddress.py
@@ -145,7 +145,7 @@ class CsInterface:
     def ip_in_subnet(self, ip):
         ipo = IPAddress(ip)
         net = IPNetwork("%s/%s" % (self.get_ip(), self.get_size()))
-        return ipo in list(net)
+        return ipo in net
 
     def get_gateway_cidr(self):
         return "%s/%s" % (self.get_gateway(), self.get_size())


borisstoyanov:
@ProjectMoon thank you for this PR, is there an ACS JIRA issue for this change, if so it'll be good to add it in the title same as the other PRs.

ProjectMoon:
We have not entered one into JIRA, no. But we can.

borisstoyanov:
Thanks @ProjectMoon 
@blueorangutan package

ProjectMoon:
Updated with the JIRA ticket.

borisstoyanov:
Thanks, @ProjectMoon I cannot run tests against it since we don't have an automated way to build the systemVM.

rhtyd:
@borisstoyanov cloud-scripts tarball in the systemvm.iso will have the router/scripts. We can test it.
@blueorangutan test

rafaelweingartner:
@ProjectMoon may I ask you a question?
The "net" object is already an array/map, right?

stefaniabergljot:
@rafaelweingartner It's not, but you can treat IPNetwork like it's a standard list object.

rafaelweingartner:
Great, nice work.
LGTM

kiwiflyer:
LGTM from borisstoyanov in a commit comment (acspr)

kiwiflyer:
tag:mergeready

rhtyd:
LGTM, @karuturi can this be merged?

PR #1928, opened 2017-02-02
Update hypervisor capabilities for xenserver 7

syed:
It looks like I missed to add the hypervisor capabilities when changing from master to 4.9. Can this be merged as well? 

@rhtyd @karuturi ?

PR #1951, opened 2017-02-17
CLOUDSTACK-9792: Add upgrade path for 4.9.3.0

rhtyd:
This adds an upgrade path from 4.9.2.0 to 4.9.3.0, this also includes changes from PR https://github.com/apache/cloudstack/pull/1928 that adds missing hypervisor capability in 4.9/4.10+. This also fixes a db-cleanup path sequence issues, with that puts 4.10 after 4.1.0, and before 4.2.0.

Once validated I can help merge this on master, since this will cause merge conflicts on fwd-merging.

Ping - @syed @karuturi @borisstoyanov @DaanHoogland @abhinandanprateek

### DIFF
--- a/setup/db/db/schema-4920to4930.sql
+++ b/setup/db/db/schema-4920to4930.sql
@@ -21,11 +21,103 @@
 
 INSERT IGNORE INTO `cloud`.`hypervisor_capabilities`(uuid, hypervisor_type, hypervisor_version, max_guests_limit, max_data_volumes_limit, storage_motion_supported) values (UUID(), 'XenServer', '7.0.0', 500, 13, 1);
 
-INSERT IGNORE INTO `cloud`.`hypervisor_capabilities`(uuid, hypervisor_type, hypervisor_version, max_guests_limit, security_group_enabled, max_data_volumes_limit, max_hosts_per_cluster, storage_motion_supported, vm_snapshot_enabled) values (UUID(), 'VMware', '6.5', 128, 0, 13, 32, 1, 1)
+INSERT IGNORE INTO `cloud`.`hypervisor_capabilities`(uuid, hypervisor_type, hypervisor_version, max_guests_limit, security_group_enabled, max_data_volumes_limit, max_hosts_per_cluster, storage_motion_supported, vm_snapshot_enabled) values (UUID(), 'VMware', '6.5', 128, 0, 13, 32, 1, 1);
 
-INSERT IGNORE INTO `cloud`.`guest_os_hypervisor` (uuid, hypervisor_type, hypervisor_version, guest_os_name, guest_os_id, created, is_user_defined) VALUES (UUID(), 'VMware', '6.5', 'windows9Guest', 257, now(), 0);
-INSERT IGNORE INTO `cloud`.`guest_os_hypervisor` (uuid, hypervisor_type, hypervisor_version, guest_os_name, guest_os_id, created, is_user_defined) VALUES (UUID(), 'VMware', '6.5', 'windows9_64Guest', 258, now(), 0);
-INSERT IGNORE INTO `cloud`.`guest_os_hypervisor` (uuid, hypervisor_type, hypervisor_version, guest_os_name, guest_os_id, created, is_user_defined) VALUES (UUID(), 'VMware', '6.5', 'windows9Server64Guest', 259, now(), 0);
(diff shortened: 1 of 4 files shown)


setup/db/db/schema-4920to4930.sql
```diff
@@ -0,0 +1,22 @@
...
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+--;
+-- Schema upgrade from 4.9.2.0 to 4.9.3.0;
+--;
+
+INSERT IGNORE INTO `cloud`.`hypervisor_capabilities`(uuid, hypervisor_type, hypervisor_version, max_guests_limit, max_data_volumes_limit, storage_motion_supported) values (UUID(), 'XenServer', '7.0.0', 500, 13, 1);
```
rhtyd: Note: On fwd merging, this can be re-executed in 4.9.2.0->4.10.0.0 upgrade path.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

rhtyd:
@borisstoyanov an explicit integration test is not necessary if Travis passes, since all changes are around db upgrade/version.

borisstoyanov:
Just want to confirm it fresh installs and upgrades the DB to 4.9.3. will cancel the run after as I'm planning to do an upgrade scenario as well. 
Thanks @rhtyd

setup/db/db/schema-4920to4930.sql
syed: @rhtyd I thought that this would be in the 4.9.2.0->4.9.3.0 path which is a subset to the 4.10.0.0 path. So in theory, it should be only executed once correct?

rhtyd:
Thanks @borisstoyanov 
@karuturi ping

rhtyd:
Requesting for review - @DaanHoogland @abhinandanprateek @karuturi @koushik-das @syed

koushik-das:
I already see schema-4920to41000.sql and the corresponding cleanup file in master. So how will this fit in the overall scheme? Is there any plans for 4.9.3?

rhtyd:
@koushik-das there is plan for 4.9.3, though the db schema change to insert/update the hypervisor capability need to go into 4.10.0 either with this PR carefully fwd-merged (which I can help with) or as a separate PR targeted only for 4.10. 4.9.x should not have unnecessary db changes, therefore 4.9.2->4.9.3 is not expected of any other db changes, with this assumption and that 4.9.3 should be worked upon in next few months this would add an upgrade path from both 4.9.2/4.9.3 to 4.10.0.

syed:
@rhtyd If I understand correctly if we add the DB changes in 4.9.2->4.9.3  then after forward merging, upgrade from 4.9.2->4.10.0 should not have any problems with idempotency correct?

rhtyd:
@syed yes there is no problem I see, with `INSERT IGNORE` we can have this db change in both paths resulting in same final state.

syed:
Thanks @rhtyd for confirming. LGTM :+1:

rhtyd:
ping @karuturi let's merge this?

rhtyd:
Ping @karuturi

setup/db/db/schema-4920to4930.sql
DaanHoogland: @syed depends on which we release first I'd say.

karuturi:
@rhtyd Can you merge this and also fwd-merge to master?

rhtyd:
@karuturi thanks, this will require some merge-conflicts on master. Will do it later today or tomorrow. Since it's not a blocker for 4.10 and 4.9.3 release work has not started, I'll assume this a low priority.

rhtyd:
@resmo @FabriceBrazier @karuturi I've moved from commits/changes from #2111 and eee1e1a6 to 4.9.3.0 upgrade path. If you're all okay with this, I can fwd-merge this as follows:

- Once merged, we'll freeze the schema changes for 4.9.3.0 (i.e. the 4.9.2.0->4.9.3.0 upgrade path)
- I'll fwd-merge this into master, fix conflicts and refactor/rename the upgrade path 4.9.2.0->4.10.0.0 to 4.9.3.0->4.10.0.0. This will ensure a simplified upgrade path as: 4.9.2.0->4.9.3.0->4.10.0.0.
- In future, when we have 4.9.3.0, 4.9.3.0 users will have an upgrade path to 4.10.x.

Pinging for additional review/comments - @wido @DaanHoogland @abhinandanprateek  and others

rhtyd:
@blueorangutan package

resmo:
LGTM

rhtyd:
@blueorangutan package

rhtyd:
I'll be testing and fwd-merging this with the approach described yesterday. /cc @karuturi

PR #980, opened 2015-10-26
[4.6] CLOUDSTACK-8992 Allow 32 disks to be attached to a KVM VM.

borisroman:
Tested manually.

Added 32 disks to a VM, booted the VM, formatted all disks and mounted them all. Also created volume group and wrote all disks full. Also checked UI for glitches, which weren't present.

2015-10-26-140843_423x583_scrot
2015-10-26-140930_505x320_scrot
2015-10-26-140950_645x122_scrot
screenshot from 2015-10-26 14 52 35

PR #1953, opened 2017-02-18
CLOUDSTACK-9794: Unable to attach more than 14 devices to a VM

sureshanaparti:
Updated hardcoded value with max data volumes limit from hypervisor capabilities.

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
@@ -584,18 +584,36 @@ public class LibvirtVMDef {
 
         /* skip iso label */
         private String getDevLabel(int devId, DiskBus bus) {
+            if (devId < 0) {
+                return "";
+            }
+
             if (devId == 2) {
                 devId++;
             }
 
-            char suffix = (char)('a' + devId);
             if (bus == DiskBus.SCSI) {
-                return "sd" + suffix;
+                return "sd" + getDevLabelSuffix(devId);
             } else if (bus == DiskBus.VIRTIO) {
-                return "vd" + suffix;
+                return "vd" + getDevLabelSuffix(devId);
             }
-            return "hd" + suffix;
+            return "hd" + getDevLabelSuffix(devId);
+        }
 
+        private String getDevLabelSuffix(int deviceIndex) {
+            if (deviceIndex < 0) {
+                return "";
+            }
+
+            int base = 'z' - 'a' + 1;
+            String labelSuffix = "";
+            do {
+                char suffix = (char)('a' + (deviceIndex % base));
+                labelSuffix = suffix + labelSuffix;
+                deviceIndex = (deviceIndex / base) - 1;
+            } while (deviceIndex >= 0);
+
+            return labelSuffix;
         }
 
         public void defFileBasedDisk(String filePath, int devId, DiskBus bus, DiskFmtType diskFmtType) {
@@ -714,11 +732,6 @@ public class LibvirtVMDef {
 
         public DiskFmtType getDiskFormatType() {
             return _diskFmtType;
-        }
-
-        public int getDiskSeq() {
-            char suffix = _diskLabel.charAt(_diskLabel.length() - 1);
-            return suffix - 'a';
         }
 
         public void setBytesReadRate(Long bytesReadRate) {
--- a/server/src/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/com/cloud/storage/VolumeApiServiceImpl.java
@@ -1433,9 +1433,9 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
         // that supported by hypervisor
         if (deviceId == null || deviceId.longValue() != 0) {
             List<VolumeVO> existingDataVolumes = _volsDao.findByInstanceAndType(vmId, Volume.Type.DATADISK);
-            int maxDataVolumesSupported = getMaxDataVolumesSupported(vm);
-            if (existingDataVolumes.size() >= maxDataVolumesSupported) {
-                throw new InvalidParameterValueException("The specified VM already has the maximum number of data disks (" + maxDataVolumesSupported + "). Please specify another VM.");
+            int maxAttachableDataVolumesSupported = getMaxDataVolumesSupported(vm) - 2; //IDs: 0 (ROOT) and 3 (CD-ROM) are reserved
+            if (existingDataVolumes.size() >= maxAttachableDataVolumesSupported) {
+                throw new InvalidParameterValueException("The specified VM already has the maximum number of data disks (" + maxAttachableDataVolumesSupported + ") attached. Please specify another VM.");
             }
         }
 
@@ -2527,7 +2527,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
 
             DataTO volTO = volFactory.getVolume(volumeToAttach.getId()).getTO();
 
-            deviceId = getDeviceId(vm.getId(), deviceId);
+            deviceId = getDeviceId(vm, deviceId);
 
             DiskTO disk = storageMgr.getDiskWithThrottling(volTO, volumeToAttach.getVolumeType(), deviceId, volumeToAttach.getPath(),
                     vm.getServiceOfferingId(), volumeToAttach.getDiskOfferingId());
@@ -2585,7 +2585,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
                     _volsDao.update(volumeToAttach.getId(), volumeToAttach);
                 }
             } else {
-                deviceId = getDeviceId(vm.getId(), deviceId);
+                deviceId = getDeviceId(vm, deviceId);
 
                 _volsDao.attachVolume(volumeToAttach.getId(), vm.getId(), deviceId);
             }
@@ -2623,7 +2623,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
             _hostDao.loadDetails(host);
             maxDataVolumesSupported = _hypervisorCapabilitiesDao.getMaxDataVolumesLimit(host.getHypervisorType(), host.getDetail("product_version"));
         }
-        if (maxDataVolumesSupported == null) {
+        if (maxDataVolumesSupported == null || maxDataVolumesSupported.intValue() <= 0) {
             maxDataVolumesSupported = 6; // 6 data disks by default if nothing
             // is specified in
             // 'hypervisor_capabilities' table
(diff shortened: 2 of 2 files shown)


remibergsma:
Thanks for the fix! 

Although I think that after id=25 stuff breaks on KVM, as the code generates a deviceName that is invalid. Id 25 results in vdz, after that it continues with caracters like `{` etc. The default for KVM seems to be 32 (as defined here: https://github.com/apache/cloudstack/blob/87ef8137534fa798101f65c6691fcf71513ac978/setup/db/db/schema-452to460.sql#L422).

See: https://github.com/apache/cloudstack/blob/master/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java#L591

So we either limit it to 25, or improve the code to generate valid device names above id 25. What do you think?

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

sureshanaparti:
@remibergsma The KVM support for max data volumes 32 was updated in the DB as part of CLOUDSTACK-8992 (PR #980). CloudStack supports till device Id 15, without this changes. Might be the limit 32 was tested manually in the VM.  As you pointed, in KVM, I see there is a problem with the device name after device id 25. As soon as the device name ends with suffix 'z' for device id 25, it should take the next suffix 'aa', 'ab',...so on. I think it is better to improve the code to generate a valid device name suffix with the given device id. As this is specific to KVM, I'll create a separate PR with these changes. Thanks.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2639,22 +2639,23 @@ private int getMaxDataVolumesSupported(UserVmVO vm) {
...
             }
             for (VolumeVO vol : vols) {
                 if (vol.getDeviceId().equals(deviceId)) {
-                    throw new RuntimeException("deviceId " + deviceId + " is used by vm" + vmId);
+                    throw new RuntimeException("deviceId " + deviceId + " is used by vm" + vm.getId());
                 }
             }
         } else {
             // allocate deviceId here
             List<String> devIds = new ArrayList<String>();
```
HrWiggles: Not part of your changes but... variable `devIds` should have type `List<Long>` instead of `List<String>`.
All that conversion from `int` to `String` and then converting from `String` to `long` seems unnecessary.
Should simply be able to do:
```
List<Long> devIds = new ArrayList<>();
for (long i = 1; i <= maxDataVolumesSupported; i++) {
    devIds.add(i);
}
devIds.remove(3L);
for (VolumeVO vol : vols) {
    devIds.remove(vol.getDeviceId());
}
if (devIds.isEmpty()) {
    throw new RuntimeException("every available deviceId already in use by vm " + vm.getId());
}
deviceId = devIds.iterator().next();
```
Note: my code above includes fixes to two other comments I made further down in the code.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2639,22 +2639,23 @@ private int getMaxDataVolumesSupported(UserVmVO vm) {
         return maxDataVolumesSupported.intValue();
     }
 
-    private Long getDeviceId(long vmId, Long deviceId) {
+    private Long getDeviceId(UserVmVO vm, Long deviceId) {
```
HrWiggles: How about adding unit tests for the method `getDeviceId(UserVmVO vm, Long deviceId)`?
Things I can currently think of to test:
- `RuntimeException` if param `deviceId` is specified as a negative value
- `RuntimeException` if param `deviceId` is specified as `0L`
- `RuntimeException` if param `deviceId` is specified as a value greater than the "max-device-id"
- `RuntimeException` if param `deviceId` is specified as reserved id `3L`
- `RuntimeException` if param `deviceId` is specified as an id that is already in use
- `RuntimeException` if param `deviceId` is specified as `null` and all device ids are in use
- returns id specified in param `deviceId` when not `null` and the id is not in use
- returns lowest available id when param `deviceId` is specified as `null`

(all of the above are from my understanding of how the method should behave)

rafaelweingartner: big 👍 for this request :)

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2639,22 +2639,23 @@ private int getMaxDataVolumesSupported(UserVmVO vm) {
...
         // allocate deviceId
-        List<VolumeVO> vols = _volsDao.findByInstance(vmId);
+        int maxDataVolumesSupported = getMaxDataVolumesSupported(vm);
```
rafaelweingartner: I think this is a good question. I would add that the ID `0` is reserved for the root device.

So, I add the question if the `maxDataVolumesSupported` already accounts for the one already reserved for the root disk. 

For instance, when configuring `getMaxDataVolumesSupported(vm)` with `6` for the hypervisor of the VM. Does that mean that the VM can have up to 7 devices? 1 root (id `0`), 1 CD-ROM (id `3`) and other 5 for extra disks/volumes.

sureshanaparti: @HrWiggles Thanks for the review. The max data volumes here is the actual hypervisor capability (which is posted in the db). The device id 3 is being reserved for something since long and I don't want that to be effected.
When _getMaxDataVolumesSupported()_ returns 6, max 5 volumes can be attached to the VM and one device reserved (might be for virtual tools/CDROM). _maxDataVolumesSupported_ specifies the data volumes limit supported by hypervisor, nothing related to _maxDeviceId_.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
sureshanaparti: @HrWiggles Will check if I can write a test for the same.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2639,22 +2639,23 @@ private int getMaxDataVolumesSupported(UserVmVO vm) {
...
+        List<VolumeVO> vols = _volsDao.findByInstance(vm.getId());
         if (deviceId != null) {
-            if (deviceId.longValue() > 15 || deviceId.longValue() == 3) {
-                throw new RuntimeException("deviceId should be 1,2,4-15");
+            if (deviceId.longValue() > maxDataVolumesSupported || deviceId.longValue() == 3) {
```
sureshanaparti: @HrWiggles Addressed.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2639,22 +2639,23 @@ private int getMaxDataVolumesSupported(UserVmVO vm) {
...
                 if (vol.getDeviceId().equals(deviceId)) {
-                    throw new RuntimeException("deviceId " + deviceId + " is used by vm" + vmId);
+                    throw new RuntimeException("deviceId " + deviceId + " is used by vm" + vm.getId());
                 }
             }
         } else {
             // allocate deviceId here
             List<String> devIds = new ArrayList<String>();
-            for (int i = 1; i < 15; i++) {
+            for (int i = 1; i < maxDataVolumesSupported; i++) {
```
sureshanaparti: @HrWiggles Thanks for pointing this. Addressed.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
sureshanaparti: @HrWiggles. Noted, not considering it for now.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2639,22 +2639,23 @@ private int getMaxDataVolumesSupported(UserVmVO vm) {
...
                 devIds.add(String.valueOf(i));
             }
             devIds.remove("3");
```
sureshanaparti: Thanks. Added this.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
sureshanaparti: @rafaelweingartner when configuring getMaxDataVolumesSupported(vm) with 6 for the hypervisor of the VM, the VM can have max 6 devices. 1 root (id 0), 1 CD-ROM (id 3) and other 4 for extra disks/volumes.

sureshanaparti:
@remibergsma @borisstoyanov Updated the KVM code to generate the valid device name above id 25.

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -716,11 +734,6 @@ public DiskFmtType getDiskFormatType() {
             return _diskFmtType;
         }
```
sureshanaparti: Removed unused method _getDiskSeq()_.

sureshanaparti:
@borisstoyanov Can you kick off Jenkins job on this PR.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

server/src/com/cloud/storage/VolumeApiServiceImpl.java
HrWiggles: @sureshanaparti If the condition really should be `i < maxDataVolumesSupported` (which would make the maximum device id returned by the method be `maxDataVolumesSupported - 1`), then the check + error message above
```
if (deviceId.longValue() <= 0 || deviceId.longValue() > maxDataVolumesSupported || deviceId.longValue() == 3) {
    throw new RuntimeException("deviceId should be 1,2,4-" + maxDataVolumesSupported);
```
need to be changed so as not to include the value of `maxDataVolumesSupported` itself.
Otherwise, when `maxDataVolumesSupported` has value `6` (for example), the method would not ever return `6` when parameter `deviceId` is specified as `null` but would return `6` when parameter `deviceId` is specified as `6` (assuming device id `6` is not already in use).  Also, the error message would state "deviceId should be 1,2,4-6" whenever parameter `deviceId` would be specified as an invalid value, which would not be correct (as `5` should be the highest valid device id).

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -584,18 +584,36 @@ public void defFileBasedDisk(String filePath, String diskLabel, DiskBus bus, Dis
 
         /* skip iso label */
         private String getDevLabel(int devId, DiskBus bus) {
```
HrWiggles: Would be great to have unit tests for either `getDevLabel(int devId, DiskBus bus)` or `getDevLabelSuffix(int deviceIndex)`, especially to test for the expected results when `devId` (or `deviceIndex`) are high enough to return a double-letter device label suffix.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
sureshanaparti: @HrWiggles Correct. I've updated the code. Thanks.

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
sureshanaparti: @HrWiggles Will add unit tests.

remibergsma:
Thanks @sureshanaparti will test it soon!

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

remibergsma:
Tested it, works as expected! LGTM :+1:

koushik-das:
@karuturi This PR is not correct and is resulting in travis CI failures. It needs to be properly fixed.

sureshanaparti:
@karuturi I'm working on the changes.

sureshanaparti:
Changes in the PR #2019.

Maximum data volumes supported field, in the _hypervisor_capabilities_ table, should be the attachable data volumes to the VM. This excludes the ROOT disk and CD-ROM. So, the hypervisor has to support max devices = maximum data volumes supported + 2 (ROOT and CD-ROM).

0(zero) index is considered in `getDeviceId()` for ROOT disk as the ROOT disk attach cmd uses this method. This has to be refactored.

PR #1956, opened 2017-02-20
CLOUDSTACK-9796 - Fix NPE in VirtualMachineManagerImpl.java

nathanejohnson:
This checks the work variable for NULL in all cases where it is
used.  Fixes CLOUDSTACK-9796.

### DIFF
--- a/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -744,14 +744,17 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
 
     protected <T extends VMInstanceVO> boolean changeState(final T vm, final Event event, final Long hostId, final ItWorkVO work, final Step step) throws NoTransitionException {
         // FIXME: We should do this better.
-        final Step previousStep = work.getStep();
-        _workDao.updateStep(work, step);
+        Step previousStep = null;
+        if (work != null) {
+            previousStep = work.getStep();
+            _workDao.updateStep(work, step);
+        }
         boolean result = false;
         try {
             result = stateTransitTo(vm, event, hostId);
             return result;
         } finally {
-            if (!result) {
+            if (!result && work != null) {
                 _workDao.updateStep(work, previousStep);
             }
         }
@@ -1507,12 +1510,13 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
             if (doCleanup) {
                 if (cleanup(vmGuru, new VirtualMachineProfileImpl(vm), work, Event.StopRequested, cleanUpEvenIfUnableToStop)) {
                     try {
-                        if (s_logger.isDebugEnabled()) {
+                        if (s_logger.isDebugEnabled() && work != null) {
                             s_logger.debug("Updating work item to Done, id:" + work.getId());
                         }
                         if (!changeState(vm, Event.AgentReportStopped, null, work, Step.Done)) {
                             throw new CloudRuntimeException("Unable to stop " + vm);
                         }
+
                     } catch (final NoTransitionException e) {
                         s_logger.warn("Unable to cleanup " + vm);
                         throw new CloudRuntimeException("Unable to stop " + vm, e);


GabrielBrascher:
Thanks for the code that prevents NULL pointer exception.
Based on code revirew and that all checks have passed LGTM.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -744,14 +744,17 @@ protected boolean checkWorkItems(final VMInstanceVO vm, final State state) throw
 
     protected <T extends VMInstanceVO> boolean changeState(final T vm, final Event event, final Long hostId, final ItWorkVO work, final Step step) throws NoTransitionException {
         // FIXME: We should do this better.
-        final Step previousStep = work.getStep();
-        _workDao.updateStep(work, step);
+        Step previousStep = null;
+        if (work != null) {
+            previousStep = work.getStep();
```
rafaelweingartner: Can " work.getStep()" return null? 
I see that you add a check at line 757 `previousStep != null`. Why would we need that check there, and not need it here (line750)?

nathanejohnson: @rafaelweingartner if work is null, previousStep will stay null.  Maybe not the clearest way to handle this, but this prevents a null work from being passed down below.

rafaelweingartner: Ok, now I think I am starting to get it.
But I am still not sure about some things here, would you mind continue discussing?

If the work is not null, you get the previous step (let’s assume it is not null) and call the method ` _workDao.updateStep(work, step)`. After this, you call ` stateTransitTo(vm, event, hostId)`. Why do we need to call ` _workDao.updateStep(work, previousStep)` again? The ` previousStep ` continues to be the same.

nathanejohnson: Most of this code I didn't write, but I can make some guesses:

_workDao.updateStep(work, previousStep) line is in the finally block, which will execute even if an exception is thrown in stateTransitTo (like NoTransitException for instance).  So if stateTransitTo a) returns a false, or b) throw an exception, then result will be false, and line 758 will run.  So if something happens that the state isn't transitioned, someone wanted the work reverted to its previous step value.  Sort of a rollback maybe?

In the case of the VM hung in starting, my desired side effect is I want stateTransitTo to be called and set the state to Stopped , i.e., Event.AgentReportStopped -> State.Stopped .  The work has already expired at this point, so it is null.  I was trying to preserve the same behavior as before when work was not null.

Sorry if this wasn't very clear.

nathanejohnson: Oh, and to your point earlier, getStep() generally shouldn't ever return a null I don't *think* , because the step column in the op_it_work table is marked not null.

rafaelweingartner: I know you did not write this code, but it seemed a good opportunity to discuss and evaluate it. 

I understood you. I already noticed the try/finally block, and this is the point I wanted to discuss. As the example you described, if an exception happens, the finally block is executed and the state is restored to a previous one (assuming that the `stateTransitTo(vm, event, hostId)` will change the step); and this makes sense in the case of an exception. However, if `NO` exception happens, the step is also reverted to a previous one (assuming that the `stateTransitTo(vm, event, hostId)` will change the step) . The `finally ` is always executed; either with successful or unsuccessful execution of `stateTransitTo(vm, event, hostId)`.

If we wanted to deal with exceptions, it would make much more sense executing the revert on a `catch` block. I think that we want/need to change the step for `null` when `stateTransitTo` return false and the `previousStep` is null . You are changing exactly that with the extra condition at line 757. 

Did you understand what I mean?

nathanejohnson: I'm not sure I'm following.  From reading the code a bit, it looks like the only scenario where a false would be returned from stateTransitTo would be where the state was not properly persisted to the db.  No exception is thrown in that case.  In the current code, false *or* exception will try to revert.  Also, currently previousStep would *probably* never be null - though if work is null this will cause another NPE on line 747 of current code.  In the PR, previousStep will *probably* only be null in the case where work is null.

What is the value of making a distinction from a false versus an exception in this case?  Me adding a check for previousStep != null is simply to make sure that work is not null above.  I could also check work for null here too, but I don't think that's what you're getting at.

rafaelweingartner: I am not asking to make a distinction between exception or not. What I tried to say is that, if the intent/purpose of the `finally` block was only to revert the step to a previous state when exceptions occur, we could do that using a `catch` block. I think the finally here is meant to revert the state of work step even if an exception does not happen, for instance when ` stateTransitTo ` returns `false`.

I think you already answered my doubt; when you said that the ` previousStep ` is most likely never to be `null`. I thought we could have cases where ` previousStep == null`, and then if the ` stateTransitTo` returns false, with the newly added check at line 757, we would not update the step back to `null` for these cases.

nathanejohnson: Do you think something like:

    if (!result && work != null) {

would be better?  Even if work.getStep() did return a null, that should have the same effect as before.  Maybe it would be more readable too?

rafaelweingartner: Yes, I think this is more readable.

nathanejohnson: Updated, thanks for the input

rafaelweingartner: I thank you

kiwiflyer:
LGTM by Rafael in commit review above (for acspr).

borisstoyanov:
@nathanejohnson thanks for this bug fix 
@blueorangutan package

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

GabrielBrascher:
Great, this one looks ready to merge.
Just for caution. Are those failures false positives?

```
test_01_create_redundant_VPC_2tiers_4VMs_4IPs_4PF_ACL 	Failure 	368.96 	test_vpc_redundant.py
test_04_rvpc_privategw_static_routes 	Failure 	315.79 	test_privategw_acl.py
test_02_list_snapshots_with_removed_data_store 	Error 	0.04 	test_snapshots.py
```

borisstoyanov:
@GabrielBrascher privategw_acl and vpc_redundant test failures have been a pain for a while, couldn't managed to fix those tests. If someone spend time and fix them would be really good. 

While the test_snapshot is already fixed, in fact we just merged the fix in Trillian as well and next run would pass.

GabrielBrascher:
Got it @borisstoyanov. Thanks!

rhtyd:
LGTM. /cc @karuturi

PR #1957, opened 2017-02-21
CLOUDSTACK-9748:VPN Users search functionality broken

Ashadeepa:
VPN Users search functionality broken
If you try to search VPN users with it’s user name, you will not be able to search.

Fixed the same.

Parent PR : https://github.com/apache/cloudstack/pull/1910

### DIFF
--- a/server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
+++ b/server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
@@ -590,6 +590,7 @@ public class RemoteAccessVpnManagerImpl extends ManagerBase implements RemoteAcc
     public Pair<List<? extends VpnUser>, Integer> searchForVpnUsers(ListVpnUsersCmd cmd) {
         String username = cmd.getUsername();
         Long id = cmd.getId();
+        String keyword = cmd.getKeyword();
         Account caller = CallContext.current().getCallingAccount();
         List<Long> permittedAccounts = new ArrayList<Long>();
 
@@ -605,6 +606,7 @@ public class RemoteAccessVpnManagerImpl extends ManagerBase implements RemoteAcc
 
         sb.and("id", sb.entity().getId(), SearchCriteria.Op.EQ);
         sb.and("username", sb.entity().getUsername(), SearchCriteria.Op.EQ);
+        sb.and("keyword", sb.entity().getUsername(), SearchCriteria.Op.LIKE);
         sb.and("state", sb.entity().getState(), Op.IN);
 
         SearchCriteria<VpnUserVO> sc = sb.create();
@@ -612,6 +614,10 @@ public class RemoteAccessVpnManagerImpl extends ManagerBase implements RemoteAcc
 
         //list only active users
         sc.setParameters("state", State.Active, State.Add);
+
+        if(keyword != null){
+            sc.setParameters("keyword", "%" + keyword + "%");
+        }
 
         if (id != null) {
             sc.setParameters("id", id);


server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
```diff
@@ -621,6 +627,10 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
             sc.setParameters("username", username);
         }
 
+        if (keyword!= null) {
```
ustcweizhou: it seems line 630 to 633 are not needed

Ashadeepa:
@ustcweizhou : My bad. Amended the changes . Thanks.

server/src/com/cloud/network/vpn/RemoteAccessVpnManagerImpl.java
Ashadeepa: @ustcweizhou : My bad. Amended the changes . Thanks.

rafaelweingartner:
@Ashadeepa why do we have 2 PRs for the same problem?
It seems that one of them can be closed.

Ashadeepa:
@rafaelweingartner : This is due to the change in my remote urls. Closing the old PR https://github.com/apache/cloudstack/issues/1910.

ustcweizhou:
tested. LGTM

btw, you can use "git push --force" to overwrite the code

SudharmaJain:
LGTM on the code changes.

rafaelweingartner:
@Ashadeepa thanks for the information.
As a tip, try to avoid opening duplicated PRs, it gets hard for reviewers to track changes and discussions.

As @ustcweizhou suggested, you could have used a push force to push the new changes to the old PR branch.

I think @syed has asked you before; what is this `keyword` parameter? The docs are not clear about it; it just says “List by keyword”. This property comes from `org.apache.cloudstack.api.BaseListCmd`. This parameter was not been used before here; so I am assuming you are repurposing it, or at least giving a use to a property that was already there.

Looking at the code, I am I see that you use the keyword value to list any username that matches it. My question is the following: is there a way to document this use case? Otherwise, only a hand full of people will know its use. I have to tell you, reading the docs (https://cloudstack.apache.org/api/apidocs-4.9/apis/listVpnUsers.html), I would never try to use the `keyword` parameter as a filter to list all users that may match a given string.

Ashadeepa:
@rafaelweingartner : Regarding that force commit, will surely keep that in mind in future.

About keyword documentation, I had gone through several cloudstack apis (E.g. https://cloudstack.apache.org/api/apidocs-4.9/apis/listVirtualMachines.html, https://cloudstack.apache.org/api/apidocs-4.9/apis/listNetworks.html, https://cloudstack.apache.org/api/apidocs-4.9/apis/listZones.html, etc) , similar thing are being used. I do agree what you are suggesting. However,  IMO, the API name "listVpnUser" gives an idea about the response we will get, i.e. listing the vpn user using a search keyword.

rafaelweingartner:
Actually, my concerns are not regarding the response. Reading the name of the method one can for sure understand what it returns. My concern is that reading the parameter name `keyword` I really have no clue that this keyword is part of a username. The `keyword` could be used to list users of domains that may match it; the same can be applied for projects. Did you understand what I mean?
I was trying to discuss if we could find a better way to document it. As you also checked, there are other methods that this situation happens. However, I really do not believe this is a good standard to follow.
Would not something like the following work as well?
```
...
sb.and("state", sb.entity().getState(), Op.IN);
sb.and().op("username", sb.entity().getUsername(), SearchCriteria.Op.EQ).or("username", "%" + sb.entity().getUsername() + "%" , SearchCriteria.Op.LIKE);
...
```

rhtyd:
LGTM (code review).
@blueorangutan package

rhtyd:
@blueorangutan package
LGTM /cc @karuturi

rhtyd:
@blueorangutan test

PR #1962, opened 2017-02-22
CLOUDSTACK-9798 - Resizing root disk fails when using SolidFire

syed:
This one is easy to reproduce:

1. Create a VM with where the root disk goes to the managed solidfire
2. Stop the VM
3. Resize the ROOT disk
4. Error

@mike-tutkowski  Would you be able to review this?

### DIFF
--- a/engine/storage/volume/src/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
+++ b/engine/storage/volume/src/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
@@ -18,56 +18,6 @@
  */
 package org.apache.cloudstack.storage.volume;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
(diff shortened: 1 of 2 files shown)


syed:
@mike-tutkowski This happens when you don't specify IOPS in the service offering for the VM

mike-tutkowski:
@syed This LGTM. I wonder, though, if we can only check in the necessary parts (i.e. not the massive rearrangements to the imports)?

mike-tutkowski:
@syed Hey - I was also working on a feature like this, but related to both XenServer and VMware (adding in VMware-specific code on the resource side). Do you need these changes for 4.10 or can I incorporate what I need from them for 4.11 into code that I'm building?

syed:
@mike-tutkowski If you already have incorporated these changes, I'll close this one. If possible, please add me as reviewer when you open the other PR

mike-tutkowski:
@syed Yes, I can include you as a reviewer for the relevant PR.

PR #1972, opened 2017-02-25
CLOUDSTACK-9779 : Releasing secondary guest IP fails with error VM nic Ip x.x.x.x is mapped to load balancing rule

niteshsarda:
ISSUE

Releasing secondary guest IP fails with error VM nic Ip x.x.x.x is mapped to load balancing rule

REPRO STEPS

Create two isolated guest networks with same CIDR
Deploy VMs on both networks
Acquire secondary IP on NICs of both VMs and make sure they have the same value, user can input the IP address.
Configure Loadbalancing rule on one of the secondary IP address and try releasing the other secondary IP address.
The operation would fail
EXPECTED BEHAVIOR

Secondary IP address should be released if there are no LB rules associated with it.

ACTUAL BEHAVIOR

Not releasing secondary IP address even if there are no LB rules associated with it.

### DIFF
--- a/server/src/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/com/cloud/vm/UserVmManagerImpl.java
@@ -4946,8 +4946,8 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         }
 
         // Check if the source and destination hosts are of the same type and support storage motion.
-        if (!(srcHost.getHypervisorType().equals(destinationHost.getHypervisorType()) && srcHost.getHypervisorVersion().equals(destinationHost.getHypervisorVersion()))) {
-            throw new CloudRuntimeException("The source and destination hosts are not of the same type and version. " + "Source hypervisor type and version: "
+        if (!(srcHost.getHypervisorType().equals(destinationHost.getHypervisorType()))) {
+            throw new CloudRuntimeException("The source and destination hosts are not of the same type. " + "Source hypervisor type and version: "
                     + srcHost.getHypervisorType().toString() + " " + srcHost.getHypervisorVersion() + ", Destination hypervisor type and version: "
                     + destinationHost.getHypervisorType().toString() + " " + destinationHost.getHypervisorVersion());
         }


engine/schema/src/com/cloud/network/dao/LoadBalancerDaoImpl.java
```diff
@@ -75,4 +84,40 @@ protected LoadBalancerDaoImpl() {
...
     }
 
+    @Override
+    public boolean isLoadBalancerRulesMappedToVmGuestIp(long instanceId, String instanceIp, long networkId, FirewallRule.Purpose purpose)
+    {
+        SearchBuilder<LoadBalancerVMMapVO> lbVmMapSearch = _loadBalancerVMMapDao.createSearchBuilder();
+        lbVmMapSearch.and("instanceIp", lbVmMapSearch.entity().getInstanceIp(),SearchCriteria.Op.EQ);
+        lbVmMapSearch.and("instanceId", lbVmMapSearch.entity().getInstanceId(), SearchCriteria.Op.EQ);
+
+        SearchBuilder<FirewallRuleVO> firewallRuleIdSearch = _firewallRulesDao.createSearchBuilder();
```
ustcweizhou: You might know LoadBalancerVO extends FirewallRuleVO, you can use LoadBalancerVO instead.

niteshsarda: @ustcweizhou  : As per suggestion, I have made changes. Please check.

ustcweizhou:
code LGTM

engine/schema/src/com/cloud/network/dao/LoadBalancerDao.java
```diff
@@ -29,4 +30,6 @@
 
     List<LoadBalancerVO> listInTransitionStateByNetworkIdAndScheme(long networkId, Scheme scheme);
 
+    boolean isLoadBalancerRulesMappedToVmGuestIp(long instanceId, String instanceIp, long networkId, FirewallRule.Purpose purpose);
```
SudharmaJain: @niteshsarda You don't need the purpose parameter here, as the implementation is specific for firewall rules. While setting the parameter with SearchBuilder the value can be set as  Purpose.LoadBalancing.

niteshsarda: @SudharmaJain : As per suggestion, removed Purpose parameter from method signature and added directly in the SearchBuilder. 

Please check.

SudharmaJain:
@niteshsarda I don't see any change in file createNetworkOffering.xml, but still there are many lines being removed and added. As there is not change in this file, please revert.

Also the marvin test case seems as the complete file was edited, even the license header. Please add only the changes that are relevant.

niteshsarda:
@ustcweizhou : Can you please check new PR for this issue and provide LGTM on that PR.

PR link : 
https://github.com/apache/cloudstack/pull/1978

niteshsarda:
@SudharmaJain Can you please check new PR for this issue. Also, I have implemented the changes which you have suggested.

PR link :
#1978

PR #1973, opened 2017-02-27
CLOUDSTACK-9795: moved logrotate from cron.daily to cron.hourly for vpcrouter

dmabry:
[BACKPORT PR1954]

moved logrotate from cron.daily to cron.hourly for vpcrouter in cloud-early-config. This brings 'vpcrouter' inline with 'router'. We are having issues with cloud.log not rotating fast enough, which filled up /var/log and ultimately caused the VR to stop functioning in such a way that it prevented new VMs from being deployed.

### DIFF
--- a/systemvm/patches/debian/config/etc/init.d/cloud-early-config
+++ b/systemvm/patches/debian/config/etc/init.d/cloud-early-config
@@ -1110,6 +1110,10 @@ EOF
       echo 0 > /var/cache/cloud/dnsmasq_managed_lease
   fi
   load_modules
+
+  #setup hourly logrotate
+  mv -n /etc/cron.daily/logrotate /etc/cron.hourly 2>&1
+
 }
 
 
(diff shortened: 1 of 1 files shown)


dmabry:
Did a git cherry-pick as suggested by @rafaelweingartner.

rafaelweingartner:
@dmabry as I posted on the other PR, I only suggested `git cherry-pick` because I thought it would preserve the commit hash. It was a misunderstanding from me. The git hash is generated based on parent commit as well as other bits of information. Therefore, even when using `git cherry-pick` the hash of commits between branches will always be different (unless a hash collision happens). 

The `git cherry-pick` is quite useful when you have to apply a bunch of changes, but in your case, I do not think you need it. The way you did before was ok.

BTW: I think a single commit to introduce these changes is enough ;)

dmabry:
Closing PR as it will be cherry-picked from master.

PR #1975, opened 2017-02-28
Fix build failure on master

koushik-das:


koushik-das:
ping @karuturi

rhtyd:
LGTM, found the build failure locally today. Introduced in `a2428508`.

PR #1979, opened 2017-03-01
CLOUDSTACK-9198: Donot allow VMs to be deplyed on host that are in di…

abhinandanprateek:
…sabled pod


borisstoyanov:
@blueorangutan package

PR #1984, opened 2017-03-03
CLOUDSTACK-9812:Update "updatePortForwardingRule" pi to include additional parameter to update the end port in case of port range

mrunalinikankariya:
Configure a PF rule Private port : Start port ; 20 ENd POrt 25 || Public Port : Start port 20 ; ENd Port : 25.
Trigger UpdatePortForwardingRule api
ApI fails with following error : " Unable to update the private port of port forwarding rule as the rule has port range "

Solution-
Port range gets modified


yvsubhash:
@mrunalinikankariya One of the commits is not related to the issue

mrunalinikankariya:
Can see two cmmits in same card. so

PR #1995, opened 2017-03-09
CLOUDSTACK-9828: GetDomRVersionCommand fails to get the correct versi…

abhinandanprateek:
…on as output

Fix tries to return the output as a single command, instead of appending output from two commands

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/get_template_version.sh
+++ b/systemvm/patches/debian/config/opt/cloud/bin/get_template_version.sh
@@ -42,5 +42,6 @@ then
     exit
 fi
 
-echo -n `cat /etc/cloudstack-release`'&'
-cat /var/cache/cloud/cloud-scripts-signature
+release=`cat /etc/cloudstack-release`
+sig=`cat /var/cache/cloud/cloud-scripts-signature`
+echo  "${release}&${sig}"


ustcweizhou:
@abhinandanprateek I woud say the change is good, but it does not really fix the issue you described.
The issue seems to be caused by missing /var/cache/cloud/cloud-scripts-signature
The file should be created in cloud-early-config. How could it be missed ?

abhinandanprateek:
@ustcweizhou the file exists and the script works on the VR. It is not able to return the complete output though via the router_proxy. The signature file is part of the template.

abhinandanprateek:
@ustcweizhou another observation is that with the same systemvm.iso and VR templates; get version works on with some hosts while consistently failing in some cases.

rhtyd:
LGTM. Got this tested on a trillian env. I'll fire one more round.
@blueorangutan package

DaanHoogland:
LGTM

ustcweizhou:
@abhinandanprateek can you spend time on what caused the issue ?
The VR will not be stopped with this PR, it will help troubleshooting the issue, good.
I never saw this issue before actually.

rhtyd:
@ustcweizhou the issue we found is due to delay in buffered stdout, where sometimes the getdomr answer may not have the 'hash' after the delimiter '&'. This is a very specific case which normally could not be reproduced. We saw this issue in a production environment and Abhi recently found the issue in a Trillian environment.

ustcweizhou:
@rhtyd thanks for your explanation.
@abhinandanprateek LGTM

rhtyd:
LGTM

jayapalu:
This issue I have seen many times  in my setup. But the problem is I am not able to reproduce the issue.

rhtyd:
@jayapalu it's a time/buffer issue and we were able to reproduce this in nested-environment (Trillian) few times.

rafaelweingartner:
@DaanHoogland @rhtyd @abhinandanprateek, even with this PR the problem keeps happening :(
How did you manage to reproduce this error?

In a VMware environment, this problem still happens (sometimes, which is even worse than if it would happen every single time). Interesting enough, after applying this PR and using only our XenServer clusters, the problem does not happen anymore (at least we were not able to reproduce it).

abhinandanprateek:
@rafaelweingartner it is not easily reproducible. We were seeing in one of our customer env and then in one of our test env  (Xenserver). Applying the patch fixed it. See if you can get meaningful info from env where it is failing.

rafaelweingartner:
@abhinandanprateek even with this change the error still happened in XenServer hosts.
I noticed the usage of CPU in Dom0 close do 90% when the error happened. I do not know if this can trigger the problem, though.

We changed a little bit the script "router_proxy.sh", and so far we have not seen the error again. We will keep monitoring to check if the error disappears, we will create a PR if everything goes well.

BTW: to get the template version from a VR running in VMware, do we have the same execution flow? I mean, using the "router_proxy.sh" in the ESX host to access the VR?

abhinandanprateek:
@rafaelweingartner The router_proxy route happens for Xen and KVM (similar to the way we manually access the VR, by logging into the host). On the other hand, VMWare VR is accessible for MS using its private ip.

PR #1999, opened 2017-03-10
CLOUDSTACK-9600: listVirtualMachines: add VPC ID to response

resmo:
Currently, the listVirtualMachines takes the vpcid as a param but does not return it in a response. It also solves expensive operation for finding out if a VM is in a VPC.

### DIFF
--- a/api/src/org/apache/cloudstack/api/response/UserVmResponse.java
+++ b/api/src/org/apache/cloudstack/api/response/UserVmResponse.java
@@ -144,6 +144,10 @@ public class UserVmResponse extends BaseResponseWithTagInformation implements Co
     @Param(description = "the ID of the service offering of the virtual machine")
     private String serviceOfferingId;
 
+    @SerializedName(ApiConstants.VPC_ID)
+    @Param(description = "the ID of the VPC the virtual machine belongs to", since = "4.11")
+    private String vpcId;
+
     @SerializedName("serviceofferingname")
     @Param(description = "the name of the service offering of the virtual machine")
     private String serviceOfferingName;
@@ -708,6 +712,7 @@ public class UserVmResponse extends BaseResponseWithTagInformation implements Co
     public void setVgpu(String vgpu) {
         this.vgpu = vgpu;
     }
+
     public void setCpuUsed(String cpuUsed) {
         this.cpuUsed = cpuUsed;
     }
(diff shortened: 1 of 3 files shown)


resmo:
@ustcweizhou 
(1) I don't fully understand that question, there are network ids on the nics https://github.com/resmo/cloudstack/blob/705336af9d03b680c3254d37376ae22b146e97e8/server/src/com/cloud/api/query/dao/UserVmJoinDaoImpl.java#L236 , I didn't change anything related to networkid.

(2) Correct me if I am wrong, but a IMHO a VM can not be part of more than one VPC, but can be connected of different VPC tiers with corresponds in networks in VPC.

ustcweizhou:
@resmo VM can be part of any isolated networks, shared networks, vpc tiers. 
You can try adding another vpc tier to an existing vm in a vpc.

Anyway, adding Vpcid in NicResponse makes more sense.

resmo:
@ustcweizhou I can add another VPC tier in the _same_ VPC to the VM, but it is still the same VPC. I am still convinced this PR is valid unless you prove otherwise.

I am ok with adding the VPC id to the nic's network response but this does not solve my issue completely.

ustcweizhou:
@resmo please create another vpc and some vpc tiers, and have a retry

ustcweizhou:
@resmo I still think vpcid should not be added to UserVmResponse, because vm can be added to multiple vpc tiers (same as networks, this is why I asked about networkid at the beginning), I just confirmed yes.

I am not clear what your issue is. Anyway, if you want to know which vpc the vm belongs to, you can still get the vpcid from NicResponse (even now, if you get networkid of nic, then you can get the vpcid from networkid via ListNetworks).

resmo:
@ustcweizhou let us discuss this on ML. I am now doing exactly this querying ListNetworks for each network for each VM which is stupid expensive operation to only find out if a VM is in a VPC.

ustcweizhou:
@resmo I agree with adding vpcid to NicResponse, then you do not need an extra API call. You only need to search NicResponse in loop.

remibergsma:
@resmo we use this i production so @ustcweizhou is right. From a test environment I just made some screenshots to demonstrate this. It's actually prety cool :-)

This VM is part of 4 VPCs:

image

Network details show VPCid:
image

resmo:
Ok, then. 

Thanks @ustcweizhou  and @remibergsma for the clarification, changing the code to add vpcid on the nic's network response.

Closing this PR in favor of a new and fresh one.

PR #1741, opened 2016-10-27
Updated StrongSwan VPN Implementation

swill:
This PR is a merge of @jayapalu changes in #872 and the changes I had to make to get the functionality working.

I have done pretty extensive testing of this code so far and we are looking to be in pretty good shape.  One thing to note is that a `Diffie-Hellman` group **is required** in order for this feature to work correctly.  It is not highlighted in the tests below, but I have shown that the `PFS` is not required for this feature to work.  In #872 I have shown a more exhaustive set of tests of this code, but I have limited this set of tests to a recommended `IKE` and `ESP` configuration in order to reduce the noise and test the other areas of functionality.

**Test Results**
I am testing this functionality by creating two VPCs with VMs in each and creating a S2S VPN connection between the two VPCs. Then I SSH into a VM in one VPC and I ping the private IP of a VM in the other VPC. Then I tear it down and try a different configuration.

_Setup_

```
VPC 1                          VPC 2               
=====                          =====               
VPN Gateway                    VPN Gateway         
VPN Customer Gateway           VPN Customer Gateway
VPN Connection        <--->    VPN Connection
 - Passive = True               - Passive = False
```

_Legend_
`SKIP` => At least one of the VPN Connections did not come up, so no test was run.
`OK` => The ping test was successful over the S2S VPN connection.
`FAIL` => The ping test failed over the S2S VPN connection.

`Passive` => Specifies if either the `<vpc_1> : <vpc_2>` sides of the VPN Connection is set to passive.
`Conn State` => Specifies the connection status of the `<vpc_1> : <vpc_2>` VPN Connection in the UI.
`Requires Reset` => If the ping test does not result in an `OK`, then a VPN Connection Reset is performed on either `<vpc_1> : <vpc_2>` sides of the VPN Connection based on which side is not showing `Connected`.  The results in the `Status` column is the final result after the reset is performed.

_Results_

```
+--------+----------------------+-------+-------+----------+----------+---------------+-----------------------------+----------------+
| Status | IKE & ESP            | DPD   | Encap | IKE Life | ESP Life | Passive       | Conn State                  | Requires Reset |
+========+======================+=======+=======+==========+==========+===============+=============================+================+
| OK     | aes128-sha1;modp1536 | True  | False | 86400    | 3600     | True : False  | Disconnected : Connected    | False : False  |
+--------+----------------------+-------+-------+----------+----------+---------------+-----------------------------+----------------+
[...]

PR #2003, opened 2017-03-13
CLOUDSTACK-9811: fixed an issue if the dev is not in the databag

swill:
Defend against the specified dev not being in the databag.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs_ip.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs_ip.py
@@ -45,7 +45,7 @@ def merge(dbag, ip):
     if ip['nw_type'] == 'control':
         dbag[ip['device']] = [ip]
     else:
-        if index != -1:
+        if index != -1 and ip['device'] in dbag and index in dbag[ip['device']]:
             dbag[ip['device']][index] = ip
         else:
             dbag.setdefault(ip['device'], []).append(ip)
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs_ip.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs_ip.py
@@ -20,15 +20,14 @@ from netaddr import *
 
 def merge(dbag, ip):
     nic_dev_id = None
-    index = -1 # a non-valid array index
     for dev in dbag:
         if dev == "id":
             continue
-        for i, address in enumerate(dbag[dev]):
+        for address in dbag[dev]:
             if address['public_ip'] == ip['public_ip']:
                 if 'nic_dev_id' in address:
                     nic_dev_id = address['nic_dev_id']
-                index = i
+                dbag[dev].remove(address)
 
     ipo = IPNetwork(ip['public_ip'] + '/' + ip['netmask'])
     if 'nic_dev_id' in ip:
@@ -45,8 +44,8 @@ def merge(dbag, ip):
     if ip['nw_type'] == 'control':
         dbag[ip['device']] = [ip]
     else:
-        if index != -1 and ip['device'] in dbag and index in dbag[ip['device']]:
-            dbag[ip['device']][index] = ip
+        if 'source_nat' in ip and ip['source_nat'] and ip['device'] in dbag and len(dbag[ip['device']]) > 0:
+            dbag[ip['device']].insert(0, ip) # make sure the source_nat ip is first (primary) on the device
         else:
             dbag.setdefault(ip['device'], []).append(ip)
 


borisstoyanov:
Thanks for this fix @swill 
@blueorangutan package

borisstoyanov:
@blueorangutan test

karuturi:
Thanks Will. Can you please add bug id(CLOUDSTACK-9811) to the PR and commit message?

ustcweizhou:
not tested it. but it seems the 'else' part also need to be changed.

swill:
@ustcweizhou the `else` case did not change from the original code: https://github.com/apache/cloudstack/pull/1741/files#diff-a7d6f7150cca74029f23c19b72ad0622L49

@karuturi I have updated the PR title and I updated the commit message to include the jira ticket.

@borisstoyanov do we have a run of these tests just against master recently to review the difference in output?

ustcweizhou:
@swill what I mean is , if you do not want to change the logic, do you need to change 'else:' to 'elif index == -1:' ?

swill:
I have to admit, I am not sure how the IP was found when looping through the list initially, but then when trying to update the index for the IP that was found, it is not there.  The only way I can think of that would cause this is if the IP <-> dev mapping is broken somehow.  It found the matching IP, but apparently that IP is not on the dev defined by the `nic_dev_id` (which comes from either: `nic_dev_id = address['nic_dev_id']` (from the databag), which can then be overridden by `nic_dev_id = ip['nic_dev_id']` (from the passed in IP) ).

Yes, I think you are right that the `else` case does not work right now.  This is because we are appending the IP to this dev, but we never removed it from the other dev it was apparently found in.  This logic is SUPPOSED to find the existing IP's index and then if found, update that index with the updated info.  

The old logic which just removed all "found" IPs and then re-added them, did not preserve the order, so the source nat IP could get reconfigured as a secondary IP on a nic instead of it being primary.  This reordering of the IPs on VR reboot caused the VPN to fail because the source nat IP was no longer the primary IP.  

So we know that removing all found IPs and then adding them does not work because it changes the order of the IPs causing the source nat IP to become a secondary IP on the nic.

The error described in CLOUDSTACK-9811 seems to be a situation where the IP is found on one nic, but then for some reason the dev which the IP is associated with is changed (I don't know why this would be happening), causing the index where the IP was found to not be valid because the IP is actually on a different dev.  

To know how to fix this correctly, we need to understand why/how an IP can be associated with one dev (index found) and then get changed to be associated with a different dev.

We need to find a way to preserve IP order while being able to update the IP configurations.

@remibergsma do you have any ideas on this?  Anyone else have ideas here?

swill:
On the Jira ticket, Wei Zhou mentioned that this could potentially happen if you have two different public IP ranges.  So if you associate a public IP which is in a different range than the source nat, a new nic will be created.  If you disassociate it, that nic will be removed.

I guess my question is.  Why is it not on the`dbag`variable then associated with that dev?  I think Wei is probably right in terms of context, but I am not sure how exactly to solve this yet.

@borisstoyanov can you confirm that you have two different public IP ranges in your setup?

borisstoyanov:
@swill we're using one IP range for the physical guest public network. I think we could set this up if required?

ustcweizhou:
@borisstoyanov are you using vpc ?

swill:
@borisstoyanov we are still trying to understand how the IP was found when looping through the `dbag` but it is not associated with the expected dev when we try to update the index which it was found on.  The only way I can see this could happen is if the dev requested it be configured on is different from the dev it is actually configured on.  I still have not figured out how that case exists.

@ustcweizhou has some good ideas on this front, but I don't think we have gotten to a point where we understand how this problem is happening.

@borisstoyanov is there a chance you can post the `ips.json` and `ip_association.json` files in your environment so we can start to understand what the config is that is causing this?

borisstoyanov:
@swill thanks, I found the ips.json, but I couldn't find the  ip_association.json on the VR. I guess it's not processed yet. Can you advise how to force it to generate? Hope this helps. 
[ips.txt](https://github.com/apache/cloudstack/files/842199/ips.txt)

borisstoyanov:
@ustcweizhou no VPC

swill:
@borisstoyanov I have a suspicion that the ip_association.json is a transient file which may be short lived. I am not sure because I have not seen it either, but it was the file being processed when you had your original error. 

I will review your ips.json and see what I can find.

swill:
This `ips.json` file confuses the hell out of me.  From what I can tell both `eth2` and `eth3` both have the ip `10.1.35.83` twice for each (for a total of the IP being configured 4 times over two interfaces).

How was this IP configured?  

```
    "eth2": [
        {
            "add": true,
            "broadcast": "10.1.63.255",
            "cidr": "10.1.35.83/19",
            "device": "eth2",
            "gateway": "10.1.63.254",
            "netmask": "255.255.224.0",
            "network": "10.1.32.0/19",
            "nic_dev_id": "2",
            "nw_type": "public",
            "one_to_one_nat": false,
            "public_ip": "10.1.35.83",
            "size": "19",
            "source_nat": false
        },
        {
            "add": true,
            "broadcast": "10.1.63.255",
            "cidr": "10.1.35.83/19",
            "device": "eth2",
            "gateway": "10.1.63.254",
            "netmask": "255.255.224.0",
            "network": "10.1.32.0/19",
            "nic_dev_id": "2",
            "nw_type": "public",
            "one_to_one_nat": false,
            "public_ip": "10.1.35.83",
            "size": "19",
            "source_nat": false
        }
    ],
    "eth3": [
        {
            "add": true,
            "broadcast": "10.1.63.255",
            "cidr": "10.1.35.83/19",
            "device": "eth3",
            "first_i_p": true,
            "gateway": "10.1.63.254",
            "netmask": "255.255.224.0",
            "network": "10.1.32.0/19",
            "new_nic": true,
            "nic_dev_id": 3,
            "nw_type": "public",
            "one_to_one_nat": false,
            "public_ip": "10.1.35.83",
            "size": "19",
            "source_nat": true,
            "vif_mac_address": "06:ee:46:00:00:03"
        },
        {
            "add": true,
            "broadcast": "10.1.63.255",
            "cidr": "10.1.35.83/19",
            "device": "eth3",
            "first_i_p": true,
            "gateway": "10.1.63.254",
            "netmask": "255.255.224.0",
            "network": "10.1.32.0/19",
            "new_nic": true,
            "nic_dev_id": 3,
            "nw_type": "public",
            "one_to_one_nat": false,
            "public_ip": "10.1.35.83",
            "size": "19",
            "source_nat": true,
            "vif_mac_address": "06:a4:80:00:00:03"
        }
    ],
```

I will see if I can get to the bottom of this, but this is pretty confusing.  Do you have a sequence of events which produced this configuration?

swill:
So...  Looking at this with a different perspective.  If there is a bug elsewhere which would cause an IP to be duplicated in the `dbag`, the old implementation would have cleaned up this bug and would have removed all the duplicates (including the real one) and then would have only added back the real one.  My implementation did not assume that I was having to clean up duplicates in the `dbag` variable and only focused on adding if it didn't exist or updating the existing entry if it did exist.  I will have to think about this more...

ustcweizhou:
the ips.json is very very weird

borisstoyanov:
yes, agree with you @swill, it seems eth3 is duplicated eth2. The VR we're trying to bring up isn't supposed to have eth3 at all...

ustcweizhou:
it seems nicDevId is not set correctly in java code.

ustcweizhou:
@borisstoyanov what's the last commit in your code ?
could you check how many nics attached to the VR in database, select * from nics where instance_id=<vr id> ?

swill:
Ya, I am not sure where the problem is stemming from.  The fact that my code is not defensive around this problem is a problem.  That is for sure.  But given my understanding of the moving parts here, I feel like there is something else going on as well.  I will see if I can get an environment up and running tomorrow so I can do some more testing on this.

borisstoyanov:
@ustcweizhou 
 - I was able to restart the network with cleanup. New VR came up with duplicated eth2 and eth3. 
screen shot 2017-03-15 at 9 48 33 am


 - Not sure what you mean by what is my last commit, currently this PR has been built and deployed on the physical hosts. 

 - There are 3 nics in the DB for that router. 
```
*************************** 8. row ***************************
            id: 8
          uuid: 9a2d5343-fbff-487e-a48d-509ae0503687
   instance_id: 4
   mac_address: 02:00:19:89:00:02
   ip4_address: 10.1.1.1
       netmask: 255.255.255.0
       gateway: NULL
       ip_type: NULL
 broadcast_uri: vlan://884
    network_id: 204
          mode: Dhcp
         state: Reserved
      strategy: Start
 reserver_name: ExternalGuestNetworkGuru
reservation_id: 2f428f23-afae-4916-a423-4a7452a33138
     device_id: 0
   update_time: 2017-03-14 13:41:23
 isolation_uri: vlan://884
   ip6_address: NULL
   default_nic: 0
       vm_type: DomainRouter
       created: 2017-03-14 09:31:29
       removed: NULL
   ip6_gateway: NULL
      ip6_cidr: NULL
  secondary_ip: 0
   display_nic: 1
*************************** 9. row ***************************
            id: 9
          uuid: bb36a0a5-ac45-45ac-976f-341e2d2c72be
   instance_id: 4
   mac_address: 0e:00:a9:fe:00:44
   ip4_address: 169.254.0.68
       netmask: 255.255.0.0
       gateway: 169.254.0.1
       ip_type: Ip4
 broadcast_uri: NULL
    network_id: 202
          mode: Static
         state: Reserved
      strategy: Start
 reserver_name: ControlNetworkGuru
reservation_id: 2f428f23-afae-4916-a423-4a7452a33138
     device_id: 1
   update_time: 2017-03-14 13:41:23
 isolation_uri: NULL
   ip6_address: NULL
   default_nic: 0
       vm_type: DomainRouter
       created: 2017-03-14 09:31:29
       removed: NULL
   ip6_gateway: NULL
      ip6_cidr: NULL
  secondary_ip: 0
   display_nic: 1
*************************** 10. row ***************************
            id: 10
          uuid: a4e7691b-b290-4bfe-905e-ddb3c5b1c466
   instance_id: 4
   mac_address: 06:bb:c8:00:00:03
   ip4_address: 10.1.35.83
       netmask: 255.255.224.0
       gateway: 10.1.63.254
       ip_type: NULL
 broadcast_uri: vlan://7
    network_id: 200
          mode: Static
         state: Reserved
      strategy: Managed
 reserver_name: PublicNetworkGuru
reservation_id: NULL
     device_id: 2
   update_time: 2017-03-14 13:41:23
 isolation_uri: vlan://7
   ip6_address: NULL
   default_nic: 1
       vm_type: DomainRouter
       created: 2017-03-14 09:31:29
       removed: NULL
   ip6_gateway: NULL
      ip6_cidr: NULL
  secondary_ip: 0
   display_nic: 1
```

ustcweizhou:
nics in database seems ok, so it looks some issue in creating ip_associations.json

swill:
Does anyone know what `first_i_p` represents in the `ips.json` dbag?  I am assuming it is the Source NAT IP, but I can't find it represented anywhere in the code (other than in a test).  If it does represent the Source NAT, then we can use it to maintain the order and keep the Source NAT IP as the primary IP on the Nic.  That was the bug I was trying to fix in my #1741  PR.

Example with: 1) source nat  2) port forwarding  3) static nat  4) load balancer
```
"eth2": [
    {
        "add": true, 
        "broadcast": "192.168.80.255", 
        "cidr": "192.168.80.8/24", 
        "device": "eth2", 
        "first_i_p": true, 
        "gateway": "192.168.80.5", 
        "netmask": "255.255.255.0", 
        "network": "192.168.80.0/24", 
        "new_nic": false, 
        "nic_dev_id": 2, 
        "nw_type": "public", 
        "one_to_one_nat": false, 
        "public_ip": "192.168.80.8", 
        "size": "24", 
        "source_nat": true, 
        "vif_mac_address": "06:00:44:00:00:23"
    }, 
    {
        "add": true, 
        "broadcast": "192.168.80.255", 
        "cidr": "192.168.80.9/24", 
        "device": "eth2", 
        "first_i_p": false, 
        "gateway": "192.168.80.5", 
        "netmask": "255.255.255.0", 
        "network": "192.168.80.0/24", 
        "new_nic": false, 
        "nic_dev_id": 2, 
        "nw_type": "public", 
        "one_to_one_nat": false, 
        "public_ip": "192.168.80.9", 
        "size": "24", 
        "source_nat": false, 
        "vif_mac_address": "06:34:d1:00:00:23"
    }, 
    {
        "add": true, 
        "broadcast": "192.168.80.255", 
        "cidr": "192.168.80.10/24", 
        "device": "eth2", 
        "first_i_p": false, 
        "gateway": "192.168.80.5", 
        "netmask": "255.255.255.0", 
        "network": "192.168.80.0/24", 
        "new_nic": false, 
        "nic_dev_id": 2, 
        "nw_type": "public", 
        "one_to_one_nat": true, 
        "public_ip": "192.168.80.10", 
        "size": "24", 
        "source_nat": false, 
        "vif_mac_address": "06:34:d1:00:00:23"
    }, 
    {
        "add": true, 
        "broadcast": "192.168.80.255", 
        "cidr": "192.168.80.11/24", 
        "device": "eth2", 
        "first_i_p": false, 
        "gateway": "192.168.80.5", 
        "netmask": "255.255.255.0", 
        "network": "192.168.80.0/24", 
        "new_nic": false, 
        "nic_dev_id": 2, 
        "nw_type": "public", 
        "one_to_one_nat": false, 
        "public_ip": "192.168.80.11", 
        "size": "24", 
        "source_nat": false, 
        "vif_mac_address": "06:34:d1:00:00:23"
    }
], 
```
@ustcweizhou @murali-reddy @remibergsma

ustcweizhou:
@swill 
first_i_p in json -> firstIP in java. This is formatted by google.gson
it is defined in api/src/com/cloud/agent/api/to/IpAddressTO.java, and used in server/src/com/cloud/network/router/CommandSetupHelper.java

so, the issue might be caused by commit baac747089ef48ea6627a6aacf27156222862352
not sure.

ustcweizhou:
@borisstoyanov I think this is caused by KVM plugin.
can you please perform the following on KVM host ?
(1) change INFO to DEBUG in /etc/cloudstack/agent/log4j-cloud.xml, and restart cloudstack-agent
(2) restart the VR (on UI), and post the StartCommand in /var/log/cloudstack/agent/agent.log
(3) post the xml definition of VR (virsh dumpxml r-XXXX-VM)
(4) post bridge information (brctl show)
(5) post /etc/cloudstack/agent/agent.properties

ustcweizhou:
@borisstoyanov 
I suspect this is caused by the name of your public interface (p55p1). 
can you please try the following change
```
diff --git a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingReso
index d925297..706f7d9 100755
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -1481,7 +1481,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv

     private String getBroadcastUriFromBridge(final String brName) {
         final String pif = matchPifFileInDirectory(brName);
-        final Pattern pattern = Pattern.compile("(\\D+)(\\d+)(\\D*)(\\d*)");
+        final Pattern pattern = Pattern.compile("(\\D+)(\\d+)(\\D*)(\\d*)(\\D*)(\\d*)");
         final Matcher matcher = pattern.matcher(pif);
         s_logger.debug("getting broadcast uri for pif " + pif + " and bridge " + brName);
         if(matcher.find()) {
@@ -1489,7 +1489,9 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
                 return BroadcastDomainType.Vxlan.toUri(matcher.group(2)).toString();
             }
             else{
-                if (!matcher.group(4).isEmpty()) {
+                if (!matcher.group(6).isEmpty()) {
+                    return BroadcastDomainType.Vlan.toUri(matcher.group(6)).toString();
+                } else if (!matcher.group(4).isEmpty()) {
                     return BroadcastDomainType.Vlan.toUri(matcher.group(4)).toString();
                 } else {
                     //untagged or not matching (eth|bond|team)#.#
```

swill:
I am considering changing my implementation to be the same as the old implementation (which removed the IP from the dbag in the initial loop of the merge), but if `source_nat` is present and it is `true` and the `len(dbag[ip['device']]) > 0` (with all the checking required), then prepend else append.   This should ensure that the source nat ip is the primary ip on the nic which the StrongSwan feature depends on.

If I do this and there is a bug elsewhere that will duplicate the IPs in the databag, then this will clean that up (as I assume it was doing before).  

Not sure this will fix @borisstoyanov's current problem, but I think it is probably a safer implementation to keep the source nat IP as the first IP.

I will update this PR with that change later today...

borisstoyanov:
Hi @swill, could you please let us know when do you expect to address the changes you have mentioned, so I could schedule the testing accordingly. Thanks.

ustcweizhou:
@borisstoyanov could you please test with my suggestion? I think it should fix the duplicated nics in VR.

remibergsma:
@swill FYI if you look for `first_i_p` as seen in the json on the Python side, then on the Java side it's called `firstIP`, as the gson lib replaces every capital with an underscore and then the lowercase letter.

swill:
Thanks @remibergsma.  I am going to use `source_nat` as it better represents what it is I am trying to establish and it seems to always be in the object where `first_i_p` does not always seem to be present (look at @borisstoyanov's output above for example).

@borisstoyanov I will work on getting you the new implementation today.

swill:
@borisstoyanov alright, this is ready for you to start testing.  Can you kick off CI on this as well?

I will be doing testing of this locally as well.  

This implementation is very similar to how it was implemented before my original change.  See: https://github.com/apache/cloudstack/pull/1741/files#diff-a7d6f7150cca74029f23c19b72ad0622L19

The only change from the original was that if the IP is a `source_nat` IP and there are already IPs associated with that dev, it will `prepend` instead of `append` so the source nat ip is set as the primary ip on the nic (required for VPN).

borisstoyanov:
not sure about the CI tests @swill, I think the easiest way to kick Travis tests is the close/reopen the PR.
I'll pick it up as soon as possible. I'll rebuild and run the smoketests. 
@blueorangutan package

borisstoyanov:
@blueorangutan test

swill:
thank you sir.  :)

borisstoyanov:
@blueorangutan test

borisstoyanov:
ping @rhtyd @DaanHoogland @abhinandanprateek @PaulAngus for review.

DaanHoogland:
code LGTM but @ustcweizhou 's patch still makes sense to me. Maybe you can add that separately? travis still fails!

borisstoyanov:
@DaanHoogland you could review it as well it addressed in https://github.com/apache/cloudstack/pull/2011/files

karuturi:
@ustcweizhou can you also review this fix? Is pr #2011 still required?

rhtyd:
LGTM, does this affect 4.9 as well @swill ? /cc @borisstoyanov

borisstoyanov:
@rhtyd I think this was introduced with the StrongSwan implementation which was 4.10 I think.

ustcweizhou:
@karuturi to be honest, I think PR#2011 fixes the issue, not this one.

swill:
I think #2011 fixes the observed issue, but this PR improves the IP ordering fix originally added to the strongswan pr.  This implementation ensures (again) that duplicate public ips will not be saved to the data bag while ensuring the source nature ip is primary on its nic. 

In short, I think both PRs are worth merging.

borisstoyanov:
@karuturi we have 3 LGTMs, I think we're good to merge this.
tag:mergeready

karuturi:
ok. Thanks everyone. I am merging this.

PR #2009, opened 2017-03-17
CLOUDSTACK-9369 Fixed Ldap regression

karuturi:
Ldap auto creation of accounts is broken due to the security fix for
CLOUDSTACK-9369.
There was an explicit check to not allow login incase the
user doesnt exist.  removed the same.

### DIFF
--- a/server/src/com/cloud/api/auth/DefaultLoginAPIAuthenticatorCmd.java
+++ b/server/src/com/cloud/api/auth/DefaultLoginAPIAuthenticatorCmd.java
@@ -166,7 +166,7 @@ public class DefaultLoginAPIAuthenticatorCmd extends BaseCmd implements APIAuthe
                     throw new CloudAuthenticationException("Unable to find the domain from the path " + domain);
                 }
                 final UserAccount userAccount = _accountService.getActiveUserAccount(username[0], domainId);
-                if (userAccount == null || !(User.Source.UNKNOWN.equals(userAccount.getSource()) || User.Source.LDAP.equals(userAccount.getSource()))) {
+                if (userAccount != null && User.Source.SAML2 == userAccount.getSource()) {
                     throw new CloudAuthenticationException("User is not allowed CloudStack login");
                 }
                 return ApiResponseSerializer.toSerializedString(_apiServer.loginUser(session, username[0], pwd, domainId, domain, remoteAddress, params),


karuturi:
@rhtyd Can you review the fix?

server/src/com/cloud/api/auth/DefaultLoginAPIAuthenticatorCmd.java
```diff
@@ -166,7 +166,7 @@ public String authenticate(String command, Map<String, Object[]> params, HttpSes
                     throw new CloudAuthenticationException("Unable to find the domain from the path " + domain);
                 }
                 final UserAccount userAccount = _accountService.getActiveUserAccount(username[0], domainId);
-                if (userAccount == null || !(User.Source.UNKNOWN.equals(userAccount.getSource()) || User.Source.LDAP.equals(userAccount.getSource()))) {
+                if (userAccount != null && User.Source.SAML2 == userAccount.getSource()) {
```
rhtyd: Why do we need to remove the check `userAccount == null`? Comparision against Source.SAML2 is fine as long as there are no other user sources other than LDAP, SAML2 and UNKNOWN. (UNKNOWN should be NATIVE though)

rhtyd:
@karuturi done, can you check Travis failure; push -f or close/reopen the PR to kick it again?

server/src/com/cloud/api/auth/DefaultLoginAPIAuthenticatorCmd.java
karuturi: in case of LDAP, account will be created if it doesnt already exist.
https://cwiki.apache.org/confluence/display/CLOUDSTACK/LDAP%3A+Trust+AD+and+Auto+Import

karuturi:
I am seeing the same Travis timeout on all the PRs. Will force push again

karuturi:
ping @rhtyd

svenvogel:
We installed your jar file 4.9.2 with the changes and test the login. all thing work correctly. +1

server/src/com/cloud/api/auth/DefaultLoginAPIAuthenticatorCmd.java
rhtyd: Thanks, let's keep this. We can revisit if we've more user type for which this may cause a security issue.

rhtyd:
LGTM.
@blueorangutan package

PR #2011, opened 2017-03-20
CLOUDSTACK-9811: fix duplicated nics on VR caused by nic name p<slot_number>p<port_number>

ustcweizhou:


borisstoyanov:
@blueorangutan package

borisstoyanov:
Hi @ustcweizhou, out of curiosity why 4.9 instead of master?

ustcweizhou:
@borisstoyanov first 4.9, then merge into master.
this patch can also be applied to master.
```
wget --no-check-certificate https://github.com/apache/cloudstack/pull/2011.patch
git am 2011.patch 
(or patch -p1 <2011.patch)
```

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

borisstoyanov:
Looks like tests were aborted, will restart now
@blueorangutan test

rhtyd:
LGTM. Can we check travis failures?

DaanHoogland:
@ustcweizhou this enables looking for a nic by a name like a1b2c3 (\D+\d+\D*\d*\D*\d*) it will also allow all shorter as long as they start with <letters><numbers> which makes sense. But to satisfy @karuturi's question in 2003, could you *briefly* describe how this prevents extra nics from being created? otherwise LGTM

borisstoyanov:
thank you guys, tests are looking good as well. 
I think we have enough LGTM to merge this now
cc: @karuturi 
tag:mergeready

ustcweizhou:
@DaanHoogland @karuturi @borisstoyanov 

Here is a short explanation why the issue happened:
Normally getBroadcastUriFromBridge method in KVM plugin will get the URI from a bridge name. for example, breth1-100 -> vlan://100, brem1-200 -> vlan://200.
However, for the bridge like brp55p1-100, it will return vlan://1 which is obviously wrong.

The method getBroadcastUriFromBridge  is used in prepareNetworkElementCommand for command IpAssocCommand  in VR. Each time when a new ip is associated, it will check the existing nics by getting the URIs from bridge name of nics. If the URIs do not contain the uri of te new ip, then a new nic will be plugged and new ip will be associated to the new nic.

The issue happened when cloudstack try to set more information of  eth2 (public nic) of a VR, as for the error described in part 1 , the URI of eth2 return wrong result (vlan://1) which is different from uri of ip (it is vlan://7 from StartCommand), so eth3 is added.

DaanHoogland:
thanks @ustcweizhou 
@karuturi can we merge this? and merge forward?

karuturi:
ok. Thanks everyone. I am merging this.

PR #1613, opened 2016-07-21
CLOUDSTACK-9436: vm_network_map table cleanup, release network resources on expunge command

nvazquez:
JIRA TICKET: https://issues.apache.org/jira/browse/CLOUDSTACK-9436

This PR replaces #1594 

Due to error on `test/integration/smoke/test_vpc_redundant.py` it was found out that `vm_network_map` table should be less aggresive on vm stop

PR #2017, opened 2017-03-23
CLOUDSTACK-9847 vm_network_map is not getting cleaned 

SudharmaJain:
While releasing nic for Create reservation strategy, vm_network_map is not getting cleaned. I have added corresponding code to perform the cleanup.


serg38:
@SudharmaJain Isn't it duplicate to PR1613 that is already merged?

SudharmaJain:
@serg38 This is different from #1613. In #1613 vm_network_map was getting cleaned only when nic reservation strategy is Start, but not for Create.

PR #1549, opened 2016-05-17
CLOUDSTACK-9348: NioConnection improvements

rhtyd:
Reopened PR with squashed changes for a re-review and testing after https://github.com/apache/cloudstack/pull/1493 and sub-sequent PRs got reverted

PR #1493, opened 2016-04-13
CLOUDSTACK-9348: Use non-blocking SSL handshake in NioConnection/Link

rhtyd:
- Uses non-blocking socket config in NioClient and NioServer/NioConnection
- Scalable connectivity from agents and peer clustered-management server
- Removes blocking ssl handshake code with a non-blocking code
- Protects from denial-of-service issues that can degrade mgmt server responsiveness
  due to an aggressive/malicious client
- Uses separate executor services for handling connect/accept events

Changes are covered the NioTest so I did not write a new test, advise how we can improve this. Further, I tried to invest time on writing a benchmark test to reproduce a degraded server but could not write it deterministic-ally (sometimes fails/passes but not always). Review, CI testing and feedback requested /cc @swill @jburwell @DaanHoogland @wido @remibergsma @rafaelweingartner @GabrielBrascher

PR #2027, opened 2017-04-05
CLOUDSTACK-9918: Activate NioTest following changes in CLOUDSTACK-9348 PR #1549

marcaurele:
The first PR #1493 re-enabled the NioTest but not the new PR #1549.

@rhtyd the test fails locally on my laptop. Is there any special configuration requirements?

### DIFF
--- a/utils/pom.xml
+++ b/utils/pom.xml
@@ -208,7 +208,6 @@
           <excludes>
             <exclude>com/cloud/utils/testcase/*TestCase*</exclude>
             <exclude>com/cloud/utils/db/*Test*</exclude>
-            <exclude>com/cloud/utils/testcase/NioTest.java</exclude>
           </excludes>
         </configuration>
       </plugin>
--- a/utils/src/main/java/com/cloud/utils/nio/NioConnection.java
+++ b/utils/src/main/java/com/cloud/utils/nio/NioConnection.java
@@ -98,10 +98,9 @@ public abstract class NioConnection implements Callable<Boolean> {
         }
         _isStartup = true;
 
-        _threadExecutor = Executors.newSingleThreadExecutor();
-        _futureTask = _threadExecutor.submit(this);
-
+        _threadExecutor = Executors.newSingleThreadExecutor(new NamedThreadFactory(this._name + "-NioConnectionHandler"));
         _isRunning = true;
+        _futureTask = _threadExecutor.submit(this);
     }
 
     public void stop() {


marcaurele:
@rhtyd the `NioTest` result is not consistent on my laptop and fails from time to time.

rhtyd:
@marcaurele the test makes use of threads to run various clients and server instances and is known to consume ports for the same, for these reasons it might fail in some test/build environments. If you look at the git history on this test file and on the pom.xml, you'll find that this test had been disabled in the past (before PR 1493) due to these issues. The test is known to fail intermittently in CI environments such as Jenkins and Travis, which is why I added to the ignored list of test with one of the PRs.

On my workstation, I've been consistently able to run and pass it, it is likely an environment issue in your laptop if this is failing inconsistently.

marcaurele:
@rhtyd I found one issue with the test and `NioConnection` class. This kind of intermittent problem are always hard to search for a root cause, but after lots of logging I finally found why. I updated the PR with the change.

If the main thread running the test is stopped there https://github.com/apache/cloudstack/blob/master/utils/src/main/java/com/cloud/utils/nio/NioConnection.java#L102 due to context switching, the flag `_isRunning` isn't switched to True by the time the NioServer connection handler start it's call loop, and it exits on the `while(_isRunning)`
https://github.com/apache/cloudstack/blob/master/utils/src/main/java/com/cloud/utils/nio/NioConnection.java#L125 directly. Therefore the server isn't listening at all and the connection cannot be made. The flag `_isRunning` must be turned `true` before submitting the task/thread.

I still digging into Nio thread handler as we are experiencing some problem in production when quite a few agents try to connect at the same time to a management server. None of them can connect.

rhtyd:
Thanks @marcaurele that's a good catch! Can you push -f and see why Travis failed, if your fix works let's reenable this test! Also, can you edit the PR and send it for 4.9?

marcaurele:
@rhtyd I moved the PR against 4.9 and rebased the changes

rhtyd:
Thanks @marcaurele could not get time to work on this, I'll review soon.

rhtyd:
@blueorangutan package

rhtyd:
LGTM :) thanks @marcaurele

rhtyd:
@marcaurele can you add/amend the jira id in the commit summary? Something like `CLOUDSTACK-xxxx: short summary`

rhtyd:
@blueorangutan test

marcaurele:
@rhtyd I edited the summary and created the corresponding Jira issue.

rhtyd:
LGTM.

PR #2044, opened 2017-04-14
CLOUDSTACK-9877 Cleanup unlinked templates

DaanHoogland:
This implements CLOUDSTACK-9877 by marking templates for gc when
1. the template they where created with is deleted from cloudstack
2. all vms that still use it were created as full clones

a full test scenario can be found in [CLOUDSTACK-9877](https://issues.apache.org/jira/browse/CLOUDSTACK-9877)


serg38:
@DaanHoogland Can you clarify few questions/concerns
1. This change will likely to break restoreVirtualMachine functions. At the very least users should be able to turn off new behavior with a global setting that will enable/disable such cleanup.
2. Does this affect a normal template cleanup that will delete both full and linked clone templates from PS when all VMs that use them are gone?
3. You introduce vmware.full.clone.template.cleanup.period as vmware specific setting. Is this only applicable to vmware?
4. Some changes are done to Vmware-Host-Scan  task that seems to be unrelated to scope of this PR.

DaanHoogland:
@serg38 
ad 1. yes they can, if you look in the ticket there is a part of description that describes this possibility. But that said I don't see how restore is going to be hindered if the vm was created using a full clone. The image is not deleted.
ad 2. No it is only marking templates for gc and not touching any cleanup process, it relies on the regular process to do the actual cleanup. The benefit is that short lived templates that will be replaced by cron jobs for instance will be cleaned without having to wait until the VMs cloned from it are all gone (on a per primary store basis btw)
ad 3. Yes so far it is I have not studied other hypervisors on this area. The code is now in the vmware plugin but not tightly coupled to the vmware API so generalizing it shoud be rather trivial.
ad 4. the host scan task was an empty placeholder with only todo. The vcente clusters are not scanned for new hosts atm. If you look at the old code you can find that the task consists of only a comment:
```
    private Runnable getHostScanTask() {
        return new Runnable() {
            @Override
            public void run() {
                // TODO scan vSphere for newly added hosts.
                // we are going to both support adding host from CloudStack UI and
                // adding host via vSphere server
                //
                // will implement host scanning later
            }
        };
    }
```

serg38:
@DaanHoogland Thanks. But for #1 the requirement for the code changes is that the template is deleted from cloudstack. In this condition  restoreVirtualMachines when template ID is not specified  will use a local template copy on PS for either full or link clone.

DaanHoogland:
ah ok, I was not aware of that @serg38 . I'll look it up, except from not enabling the feature, how would you handle this?
a. To not allow restore
b. To recognise that the disk is fully cloned and bypass the null-template (if almighty vmware allows)
c. To allow for choosing a successor-template
d. ...
Also, Is this a blocker for you?

serg38:
@DaanHoogland Not for us since we still use link clones but I am sure tons of other might be affected. Interestingly enough that with PR 1773 merged the default behavior in API is not to allow template deletion if there are active VMs and only if 'forced' flag is used it will be executed
# c is an option already but I suggest one of the following
#d introduce another global config with default value = false e.g. vmware.cleanup.fullclonedtemplate that would be a switch old/new behaviur in combination with storage.template.cleanup.enabled
#e make 0 as a default value for vmware.full.clone.template.cleanup.period and when you  interpret 0 it means that full clone cleanup thread thread never

#e seems to be a good compromise that doesn't require too many changes

DaanHoogland:
@serg38 I am confused by that last comment; storage.template.cleanup.enabled already switches between old and new behaviour. when false/default no extra work is done, when true marking of fully cloned deleted templates is done. So you #d does not make sense to me and #e would be a redundant check. Am I missing something?

serg38:
@DaanHoogland storage.template.cleanup.enabled  is a current global setting so you can't really use it to control a behavior you are about to introduce. For example, I want my template copies on PS to be cleaned up only when no VM using the template left on the particular PS. In this case we need to have  both storage.template.cleanup.enabled and storage.cleanup.enabled to be set to 'true'. If we keep logic in your PR in this case it will also start cleaning up templates copies for full clone based templates which might be unwanted in some environments.

borisstoyanov:
@blueorangutan package

borisstoyanov:
Build fails with `Could not find artifact org.apache.cloudstack:cloud-plugin-network-midonet:jar:4.10.0.0-SNAPSHOT` as it was just retired. 
@DaanHoogland can you please rebase against master, I think that would fix the error.

DaanHoogland:
ok, @serg38 you have convinced me. I do think and extra option is overkill so I'll implement the 0 minutes idea.

DaanHoogland:
@borisstoyanov rebase
@serg38 the 0 period already had that effect due to the library for scheduling but I now put a nice guard around it and made 0 the default
@borisstoyanov please not that I changed the default period from 1440 minutes to 0 have a look at https://github.com/apache/cloudstack/pull/2044/commits/e3ae25acf20f2a685627ef706bdb4527a58e17eb

serg38:
@DaanHoogland Thanks. LGTM on the code changes

borisstoyanov:
@blueorangutan package

DaanHoogland:
@blueorangutan package

borisstoyanov:

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 800
 Hypervisor xenserver
 NetworkType Advanced
 Passed=106
 Failed=4
 Skipped=12

**Failed tests:**
* test_list_ids_parameter.py

 * ContextSuite context=TestListIdsParams>:setup Failing since 27 runs

* test_volumes.py

 * test_06_download_detached_volume Failing since 3 runs

* test_routers_network_ops.py

 * test_01_RVR_Network_FW_PF_SSH_default_routes_egress_true Failing since 4 runs

 * test_02_RVR_Network_FW_PF_SSH_default_routes_egress_false Failed


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_01_verify_libvirt
test_04_verify_guest_lspci
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vm_snapshots.py
test_over_provisioning.py
test_global_settings.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_metrics_api.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

borisstoyanov:

DaanHoogland:
two LGM and clean rebase pushed; merging

rhtyd:
@DaanHoogland can you explain (1) why this PR was merged without running latest tests, (2) last test results are at least two months old, master has changed a lot recently, (3) lastly the commits are not squashed and do not have JIRA ids in them.

DaanHoogland:
sure i can: two LGM and clean rebase pushed. commitws should not be squashed and There is a hira id in the PR title, hence merge commit

DaanHoogland:
@blueorangutan package

PR #2036, opened 2017-04-11
[CLOUDSTACK-9858] Retirement of midonet plugin (build disabling)

rafaelweingartner:
Recently there have been two threads asking and discussing the “midonet” integration with Apache CloudStack (ACS) [1-2].

After quite some discussions, we noticed that despite having some people willing to use it, the plugin has never been fully developed by its vendor (Midokura). Further, nobody else has put the effort on fully testing and finishing its implementation. It seems that the plugin was incorporated into our code base without being fully finished. Moreover, I have asked around at the Midonet community, and the java client they use has changed quite a bit from the one we use.

It begs the question if it does not work, why do we advertise such integration? [3]. 
Following the component retirement process defined in [4], a vote thread was started in [5]. The community decided to retire this Midonet plugin. This task represents the first step of the retirement, which is the disabling of the plugin in CloudStack`s build process.
 
[1] http://cloudstack.markmail.org/thread/qyedle5jb2c34gsc#query:+page:1+mid:xn2zq2v3eim5vl2q+state:results
[2] http://cloudstack.markmail.org/message/rewzk4v7dgzpsxkm?q=midonet+order:date-backward&page=1#query:midonet%20order%3Adate-backward+page:1+mid:i563khxlginf6smg+state:results
[3] http://docs.cloudstack.apache.org/en/latest/networking/midonet.html
[4] https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=68720798
[5] http://markmail.org/message/qigrtfirwnmct4hr

PR #2051, opened 2017-04-18
CLOUDSTACK-9858: remove midonet plugin dependency on client packaging

rhtyd:
This fixes packaging/build regression from #2036

Pinging for review - @wido @rafaelweingartner @DaanHoogland 

@blueorangutan package

### DIFF
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -150,11 +150,6 @@
     </dependency>
     <dependency>
       <groupId>org.apache.cloudstack</groupId>
-      <artifactId>cloud-plugin-network-midonet</artifactId>
-      <version>${project.version}</version>
-    </dependency>
-    <dependency>
-      <groupId>org.apache.cloudstack</groupId>
       <artifactId>cloud-plugin-network-ssp</artifactId>
       <version>${project.version}</version>
     </dependency>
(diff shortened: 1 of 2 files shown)


wido:
The Midonet profile stays there so we can still activate it if we want to?

Only removed as a build dependency?

rafaelweingartner:
@wido this is exactly what I did in #2036 
From what it looks like I forgot to change the dependency  in "client/pom.xml". Probably that is why the Jenkins build failed there, but I was not able to retrieve the log files to check it.

Thanks @rhtyd.

rhtyd:
This breaks packaging and noredist building.

On Apr 18, 2017 8:00 PM, "Wido den Hollander" <notifications@github.com>
wrote:

The Midonet profile stays there so we can still activate it if we want to?

Only removed as a build dependency?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/apache/cloudstack/pull/2051#issuecomment-294862977>,
or mute
the thread

rhtyd:
Packaging build came out all okay, we can merge. I'll fire a test as well.

@blueorangutan test

rhtyd:
@blueorangutan test

rhtyd:
Since, this is breaking master and blocking rest of the PRs. I'll proceed with my discretion and manual test results from Trillian and local env. Another issue was found with deployment on centos6/7 for cloudstack-management that has been fixed as well.

wido:
Good @rhtyd ! Let's fix master.

PR #1766, opened 2016-11-17
 CLOUDSTACK-9598: wrong defaut gateway for the nic in non-default network

murali-reddy:
when guest VM has nic's in more than one guest network set the tag for each host in /etc/dhcphosts.txt, and use the tag to add exception in /etc/dhcpopts.txt to prevent sending default route, dns server in case if the nic is in non-default network

this was the behaviour with edithosts.sh prior to 4.6

PR #2055, opened 2017-04-20
CLOUDSTACK-9887 ignore bogus default gateway

DaanHoogland:
when a shared network is secondary the default gateway gets overwritten by a bogus one
  dnsmasq does the right thing and replaces it with its own default which is not good for us
  so check for '0.0.0.0'

### DIFF
--- a/server/src/com/cloud/network/router/CommandSetupHelper.java
+++ b/server/src/com/cloud/network/router/CommandSetupHelper.java
@@ -108,8 +108,6 @@ import com.cloud.offering.NetworkOffering;
 import com.cloud.offerings.NetworkOfferingVO;
 import com.cloud.offerings.dao.NetworkOfferingDao;
 import com.cloud.service.dao.ServiceOfferingDao;
-import com.cloud.storage.GuestOSVO;
-import com.cloud.storage.dao.GuestOSDao;
 import com.cloud.user.Account;
 import com.cloud.uservm.UserVm;
 import com.cloud.utils.Pair;
@@ -174,8 +172,6 @@ public class CommandSetupHelper {
     private VlanDao _vlanDao;
     @Inject
     private IPAddressDao _ipAddressDao;
-    @Inject
-    private GuestOSDao _guestOSDao;
 
     @Inject
     private RouterControlHelper _routerControlHelper;
@@ -221,12 +217,6 @@ public class CommandSetupHelper {
                 _networkModel.getExecuteInSeqNtwkElmtCmd());
 
         String gatewayIp = nic.getIPv4Gateway();
-        if (!nic.isDefaultNic()) {
-            final GuestOSVO guestOS = _guestOSDao.findById(vm.getGuestOSId());
-            if (guestOS == null || !guestOS.getDisplayName().toLowerCase().contains("windows")) {
-                gatewayIp = "0.0.0.0";
-            }
-        }
 
         final DataCenterVO dcVo = _dcDao.findById(router.getDataCenterId());
 
--- a/systemvm/patches/debian/config/opt/cloud/bin/cs/CsDhcp.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/cs/CsDhcp.py
@@ -89,9 +89,10 @@ class CsDhcp(CsDataBag):
                 gateway = gn.get_gateway()
             else:
                 gateway = i['gateway']
-            sline = "dhcp-option=tag:interface-%s-%s,3," % (device, idx)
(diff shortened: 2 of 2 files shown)


rhtyd:
@DaanHoogland can you fix the commit message to include the JIRA id
@blueorangutan package

DaanHoogland:
@rhtyd the PR has the ticket id so the merge commit will as well. Is that not enough?

rhtyd:
LGTM.
@blueorangutan test

rhtyd:
@karuturi can this be merged?

resmo:
LGTM, based on logic and code but could not test

rhtyd:
LGTM

ustcweizhou:
I think it is fixed with CLOUDSTACK-9598

DaanHoogland:
@ustcweizhou #1766 doesn't fix the issue completely as it still remains in 4.9

ustcweizhou:
@DaanHoogland are you testing on 4.9.2 or 4.9 branch ?

This issue is related to commit c81b3380dffa06a7134845ea97438619018b520f
I think we'd better fix it in java side.

DaanHoogland:
@ustcweizhou I am looking into this for a customer running 4.9.1
We can not do everything on the java side as the line must not be present when the gateway is not explicitely something other then 0.0.0.0. When we change that to Null we still need to not add the line. Any line of the form
dhcp-option=tag:interface-%s-%s,3,...
fot the tagged interface must be absent.

ustcweizhou:
@DaanHoogland 
To be honest, I have no idea why the gateway ip is '0.0.0.0' when the first ip in the network is applied to a non-default nic of a non-windows vm.

However, I do not think this PR fixes the issue. If the gateway ip is not set in dhcp-option, other vms in this network will not get the gateway, even the ip is on the default nic.

ustcweizhou:
@DaanHoogland  
FYI, we removed the lines to set gatewayip to 0.0.0.0 in java code in our fork. It does not bring any issue on our production in the past about one year.

DaanHoogland:
@ustcweizhou It is worth a try. Can you confirm that no config line 3 is present for the tagged interface?
i.e. a line containing the pattern
"dhcp-option=tag:interface-%s-%s,3,"

DaanHoogland:
@ustcweizhou do you have a link to that patch, you refer to?

ustcweizhou:
@DaanHoogland The gateway is set after the change, so the dhcp-option you mentioned is configured with correct gateway.

here you are
```
diff --git a/server/src/com/cloud/network/router/CommandSetupHelper.java b/server/src/com/cloud/network/router/CommandSetupHelper.java
index 54e70b4..402fe3c 100644
--- a/server/src/com/cloud/network/router/CommandSetupHelper.java
+++ b/server/src/com/cloud/network/router/CommandSetupHelper.java
@@ -108,8 +108,6 @@ import com.cloud.offering.NetworkOffering;
 import com.cloud.offerings.NetworkOfferingVO;
 import com.cloud.offerings.dao.NetworkOfferingDao;
 import com.cloud.service.dao.ServiceOfferingDao;
-import com.cloud.storage.GuestOSVO;
-import com.cloud.storage.dao.GuestOSDao;
 import com.cloud.user.Account;
 import com.cloud.uservm.UserVm;
 import com.cloud.utils.Pair;
@@ -174,8 +172,6 @@ public class CommandSetupHelper {
     private VlanDao _vlanDao;
     @Inject
     private IPAddressDao _ipAddressDao;
-    @Inject
-    private GuestOSDao _guestOSDao;

     @Inject
     private RouterControlHelper _routerControlHelper;
@@ -225,12 +221,6 @@ public class CommandSetupHelper {
                 _networkModel.getExecuteInSeqNtwkElmtCmd());

         String gatewayIp = nic.getIPv4Gateway();
-        if (!nic.isDefaultNic()) {
-            final GuestOSVO guestOS = _guestOSDao.findById(vm.getGuestOSId());
-            if (guestOS == null || !guestOS.getDisplayName().toLowerCase().contains("windows")) {
-                gatewayIp = "0.0.0.0";
-            }
-        }

         final DataCenterVO dcVo = _dcDao.findById(router.getDataCenterId());
```

DaanHoogland:
@ustcweizhou thanks for the patch, i'll look into the replace code. if the gateway line is removed completely this is good. I'll test your patch as well.

DaanHoogland:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
LGTM

rhtyd:
LGTM, based on test results and review.

rhtyd:
@ustcweizhou @DaanHoogland you're both LGTM on this as well I suppose?

PR #2062, opened 2017-04-24
CLOUDSTACK-9878: remove unnecessary VPN drops

swill:
I have removed unnecessary VPN disconnects when updating the VPN configuration (or any configuration which changes public IP addresses).  I have verified the functionality still behaves as expected with these changes to the Remote Access VPN config.

In addition, I have also added a fix to the S2S VPN to not make it drop the VPN connection when other networking config is changed.  I was not able to reproduce the problem described in https://github.com/MissionCriticalCloud/cosmic/pull/339 manually, but we have experienced similar issues in our production environment which makes me think this is a worthwhile change.  I have tested that the functionality still behaves as expected with this change, so I have added it to this PR as well.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/configure.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/configure.py
@@ -542,15 +542,14 @@ class CsSite2SiteVpn(CsDataBag):
             logging.info("Configured vpn %s %s", leftpeer, rightpeer)
             CsHelper.execute("ipsec rereadsecrets")
 
+        # This will load the new config and start the connection when needed since auto=start in the config
         CsHelper.execute("ipsec reload")
-        if not obj['passive']:
-            CsHelper.execute("sudo nohup ipsec down vpn-%s" % rightpeer)
-            CsHelper.execute("sudo nohup ipsec up vpn-%s &" % rightpeer)
         os.chmod(vpnsecretsfile, 0400)
 
     def convert_sec_to_h(self, val):
         hrs = int(val) / 3600
         return "%sh" % hrs
+
 
 class CsVpnUser(CsDataBag):
     PPP_CHAP='/etc/ppp/chap-secrets'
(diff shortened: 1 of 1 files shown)


karuturi:
@borisstoyanov Can you run trilion tests please?

karuturi:
@swill A comment on the bug says it also happens on 4.9.2. Do you think this fix will work on 4.9.2 as well?

swill:
@karuturi I would have to look at that code to know.  Openswan was removed and replaced with Strongswan in 4.10, so the implementation is different. 

If I remember correctly, they were just reloading the ipsec config in 4.9, but the code did bounce the xl2tpd service on change. In testing yesterday, stopping the xl2tpd service and starting it again was casing issues where the tunnel could not be found when it came back up. 

I can open a similar PR against 4.9, but I don't have a 4.9 environment right now, so I can't test it. Should I create a new PR for it and ask for others to validate it?

swill:
Is there a label for `run CI on this`?  I feel like that would be useful.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

swill:
Haijiao from the mailing list has confirmed that the majority of the cases are fixed, but the connection still drops when a new VPN user is added/removed.  I have tested and can reproduce the behavior.  I am working on a solution for that problem...

swill:
Regarding the Remote Access VPN connection being dropped when users are added and deleted is because the entire VPN user databag is reprocessed on each VPN user change.  This means that every user is removed and then added back again after.  Since each user is removed, if they had an active VPN session, the session is obviously dropped.

Since there is no way to modify an existing VPN user (like change password), I should be able to find a way to selectively reconfigure the relevant VPN users.  I will update this PR once I have a solution to this specific sub-problem.

swill:
I have fixed the issue around VPN user changes dropping existing VPN connections.

In working on this, I found bugs that the VPN users were not correctly being removed from the `/etc/ppp/chap-secrets` file.  This had a lot of unwanted side effects, including:
- VPN connections could be established using deleted users.
- VPN connections could not be established for a new user if a user was deleted and added again with a different password (the original password would work, but new one would not).

All of these bugs have been fixed as well as the original problem of VPN connections dropping when users are added/removed.

Now remote access VPN connections will no longer drop when network changes are made.

This PR is ready for testing again.  Please run CI against it again and I will poke the people who are manually testing this as well.  Thanks...

swill:
@borisstoyanov or @rhtyd can one of you kick of the tests?  Thanks...

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

swill:
This one should be ready to merge...

luhaijiao:
awesome,  thanks Will, we will test in 1-2 days.

karuturi:
@luhaijiao any update? Waiting on this PR for RC2

luhaijiao:
we confirmed it has fixed the issues we reported, no more VPN drops while network changes taken including VPN account delete/add.

Thanks Will !

karuturi:
Thank you @luhaijiao. RC2 is coming :)

rhtyd:
LGTM, RC2!

PR #2065, opened 2017-04-25
do not take down the VPN on public IP config changes

swill:
This fix is similar to the 4.10 fix found here: https://github.com/apache/cloudstack/pull/2062

**NOTE:** I have not tested this fix since I do not have a 4.9 environment.  We need confirmation from the people who raised the concern that this does in-fact fix their issue before this is merged.

@borisstoyanov can you run trillian tests against this please?


luhaijiao:
Hi, Will
We have tested this fix.  The remote VPN connection is now working fine while network changes taken.  
However, we noticed if we try to add/delete VPN accounts,  the connection will be disconnected again.
It seems the issue is not completely fixed yet.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
LGTM based on results. Based on test confirmation from @luhaijiao and comments from @swill and that similar changes from master I'll merge this.

PR #2095, opened 2017-05-10
CLOUDSTACK-9914: update Quota plugin to support currency values up to 5 decimal places

GabrielBrascher:
**Summary:** this commit alters column currency_value from table cloud_usage.quota_tariff to support values up to 5 decimal places. The current implementation allows up to 2 decimal places.

**Issue:** need to use more than 2 decimal places to define resources values in Quota tariff.

**Solution:** modify column currency_value from table cloud_usage.quota_tariff to support values up to 5 decimal places. Values with more than 5 decimal places will be displayed with scientific notation in the user interface.

SQL command: "ALTER TABLE `cloud_usage`.`quota_tariff` MODIFY `currency_value` DECIMAL(15,5) not null"


rafaelweingartner:
@GabrielBrascher is there any reason that you did not use the schema-4920to41000.sql? It does not seem to be required to use Java code for this change.

Also, could you make it clearer why we need more decimal places? I think a more detailed explanation in the Jira ticket would do.

GabrielBrascher:
Thanks @rafaelweingartner, I think that both points are valid, I will update them soon.

GabrielBrascher:
Updated code and Jira description.

rafaelweingartner:
@GabrielBrascher it seems that there are some conflicts here. 
Can you check?

GabrielBrascher:
Sorry. I broke this branch and couldn't recover it in time. The new PR is #2123.

PR #2096, opened 2017-05-11
Fix for CLOUDSTACK-9660

mike-tutkowski:
A root volume can be replaced by a different root volume without the VM it belongs to being expunged.

From dev@:

For example: Let’s say we have a system VM running on NFS primary storage. We then put this primary storage into maintenance mode, which creates the system VM (with the same name) on a different primary storage (we do not create a new row in the cloud.vm_instance table for this VM). While this VM works, the original root disk of the system VM remains on the original primary storage and is not destroyed by the code in StorageManagerImpl.cleanupStorage(boolean) in 4.10 because 4.10 (as shown above) only asks for non-root volumes to consider for deletion. In the 4.9 version of the code, the original root disk is cleaned up in StorageManagerImpl.cleanupStorage(boolean). The problem with 4.10 relying on a root disk always being deleted when the VM it belongs to is deleted is that in a situation like this that the system VM doesn’t get deleted at this point – it gets a new root disk that’s hosted by a different primary storage (so now it’s original root disk is stranded).

### DIFF
--- a/engine/schema/src/com/cloud/storage/dao/VolumeDao.java
+++ b/engine/schema/src/com/cloud/storage/dao/VolumeDao.java
@@ -80,7 +80,7 @@ public interface VolumeDao extends GenericDao<VolumeVO, Long>, StateDao<Volume.S
 
     List<VolumeVO> listVolumesToBeDestroyed();
 
-    List<VolumeVO> listNonRootVolumesToBeDestroyed(Date date);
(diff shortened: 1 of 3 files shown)


karuturi:
Hi @mike-tutkowski base branch should be master as RC3 will be created from master

borisstoyanov:
@blueorangutan package

borisstoyanov:

mike-tutkowski:
Closing this PR and opening one against master

PR #2098, opened 2017-05-11
Fix for CLOUDSTACK-9660

mike-tutkowski:
A root volume can be replaced by a different root volume without the VM it belongs to being expunged.

From dev@:

For example: Let’s say we have a system VM running on NFS primary storage. We then put this primary storage into maintenance mode, which creates the system VM (with the same name) on a different primary storage (we do not create a new row in the cloud.vm_instance table for this VM). While this VM works, the original root disk of the system VM remains on the original primary storage and is not destroyed by the code in StorageManagerImpl.cleanupStorage(boolean) in 4.10 because 4.10 (as shown above) only asks for non-root volumes to consider for deletion. In the 4.9 version of the code, the original root disk is cleaned up in StorageManagerImpl.cleanupStorage(boolean). The problem with 4.10 relying on a root disk always being deleted when the VM it belongs to is deleted is that in a situation like this that the system VM doesn’t get deleted at this point – it gets a new root disk that’s hosted by a different primary storage (so now it’s original root disk is stranded).

*** Update on May 19, 2017 ***

I recently discovered a new use case (in addition to the one above) that this PR fixes:

Reinstall VM

When you run a "Reinstall VM", the root disk is removed from the VM and a new one is given to the VM. The old root disk later gets cleaned up by the storage cleanup thread.

***

https://issues.apache.org/jira/browse/CLOUDSTACK-9917

### DIFF
--- a/engine/schema/src/com/cloud/storage/dao/VolumeDao.java
+++ b/engine/schema/src/com/cloud/storage/dao/VolumeDao.java
@@ -80,7 +80,7 @@ public interface VolumeDao extends GenericDao<VolumeVO, Long>, StateDao<Volume.S
 
     List<VolumeVO> listVolumesToBeDestroyed();
 
-    List<VolumeVO> listNonRootVolumesToBeDestroyed(Date date);
+    List<VolumeVO> listVolumesToBeDestroyed(Date date);
 
     ImageFormat getImageFormat(Long volumeId);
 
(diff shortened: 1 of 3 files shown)


rhtyd:
@blueorangutan package

karuturi:
code changes LGTM. @borisstoyanov can you run CI on this?

rhtyd:
@blueorangutan test

rhtyd:
Something went wrong, will rekick the tests.
@blueorangutan package

rhtyd:

karuturi:
@mike-tutkowski 
Can you take a look at the test failures?
I think the below two failures are new

> test_08_migrate_vm	Error	71.18	test_vm_life_cycle.py
> ContextSuite context=TestDeployVmRootSize>:setup	Error	0.00	test_deploy_vm_root_resize.py

rhtyd:
I also run against kvm:
@blueorangutan test

mike-tutkowski:
@karuturi 

For test_vm_life_cycle.test_08_migrate_vm, it looks like the test was run in a bad state for some reason:

jobresult : {errorcode : 431, errortext : u'Cannot migrate VM, VM is already presnt on this host, please specify valid destination host to migrate the VM'}

For test_deploy_vm_root_resize, it appears its XML config file references an OS type that doesn't exist.

karuturi:
Thanks @mike-tutkowski. Let us wait and see if there are same failures in the next run.
@rhtyd looks like the last job didn't post results. Can you run it again?

rhtyd:
@karuturi sure

karuturi:
failures werent reported in this run. merging this now.

PR #2111, opened 2017-05-17
[CLOUDSTACK-9923] Add support for vSphere 6.5

FabriceBrazier:
vSphere 6.5 is not registered in the ACS DB. Therefore, new VMs are always created with os type "other". This is a blocking point, especially when using Windows 2012 OS (boot process fails).
This fix updates the DB accordingly.

### DIFF
--- a/setup/db/db/schema-4920to41000.sql
+++ b/setup/db/db/schema-4920to41000.sql
@@ -21,6 +21,8 @@
 
 ALTER TABLE `cloud`.`domain_router` ADD COLUMN  update_state varchar(64) DEFAULT NULL;
 
+INSERT IGNORE INTO cloud.hypervisor_capabilities values (26,UUID(),'VMware','6.5',128,0,13,32,1,1);
+
 INSERT IGNORE INTO `cloud`.`guest_os` (id, uuid, category_id, display_name, created) VALUES (257, UUID(), 6, 'Windows 10 (32-bit)', now());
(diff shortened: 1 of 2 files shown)


resmo:
Since vsphere is officially supported in LTS 4.9, this should probably even go into 4.9.3 /cc @rhtyd Any thoughts?

I'll have to check again but this was probably the cause of issues we had upgrading an ESX fom 5.5 to 6.5. AFAIR we had to remove and add the host to the cluster. Let me verify again.

rhtyd:
@resmo yes, good idea
@FabriceBrazier yes, if possible can you target the PR for 4.9 branch, and move the sql changes into the 4.9.2->4.9.3 upgrade path. There is an outstanding PR #1951 which @karuturi has asked me to merge, I can fwd-merge this to master and change master's upgrade path from 4.9.2.0 to 4.9.3.0->4.10.0.0. For master, the upgrade path looks as: 4.9.2.0->4.9.3.0->4.10.0.0. After the fwd-merge, we'll freeze any changes to the 4.9.2.0->4.9.3.0 upgrade path.

rhtyd:
LGTM, otherwise.

FabriceBrazier:
Modif is done. Could you please approve?

rhtyd:
Thanks @FabriceBrazier but that's not what I meant, we wanted to get this into 4.9 branch. Look, I've picked your change as a separate commit in: #1951  see if that's okay?

FabriceBrazier:
Thanks, ok for me. I will close this ticket.

PR #2119, opened 2017-05-18
CLOUDSTACK-9925: Quota memory tariff value is for 1MB of RAM used per month (not hour)

GabrielBrascher:
The quota memory tariff description in the CloudStack UI is wrong when defines that the value is for "using 1MB or RAM for 1 hour". The quota currency values reflect the value of a resource used per month, not an hour.

Quota divides the tariff value by the number of hours a month has (30 days - 720 hours); then it calculates the credits used by a client based on the amount of resources used per hour. The memory tariff description in the interface is wrong and can guide users to configure values for an hour.


rhtyd:
LGTM.

PR #2124, opened 2017-05-21
CLOUDSTACK-9432: cluster/host dedicated to a domain is owned by the root domain

GabrielBrascher:
**Issue:** when dedicating a resource (cluster or host) to a domain, the affinity group which is created is visible to everyone rather than only to the domain that the cluster is dedicated to.

**Cause:** if a resource is not dedicated to an account but a domain, the account parameter is null. The code then uses the domain id from the system account, making the affinity group visible by all accounts.

**Proposal**: if the account is null, then it should use the domain id passed as a parameter, instead of the domain id from the system account.

**Test scenario**: a cluster is dedicated to a domain with id '3'. The affinity group domain map should have '3' as its domain id.

**Before:** id '8' links affinity group '11' to domain '1' (ROOT).
```
select * from affinity_group_domain_map;
+----+-----------+-------------------+------------------+
| id | domain_id | affinity_group_id | subdomain_access |
+----+-----------+-------------------+------------------+
|  8 |         1 |                11 |                1 |
+----+-----------+-------------------+------------------+
```
**After:** id '9' links affinity group '12' to domain '3'.
```
select * from affinity_group_domain_map;
+----+-----------+-------------------+------------------+
| id | domain_id | affinity_group_id | subdomain_access |
+----+-----------+-------------------+------------------+
|  9 |         3 |                12 |                1 |
+----+-----------+-------------------+------------------+
```

### DIFF
--- a/server/src/org/apache/cloudstack/affinity/AffinityGroupServiceImpl.java
+++ b/server/src/org/apache/cloudstack/affinity/AffinityGroupServiceImpl.java
@@ -117,7 +117,7 @@ public class AffinityGroupServiceImpl extends ManagerBase implements AffinityGro
     @DB
     @Override
     public AffinityGroup createAffinityGroup(final String accountName, final Long projectId, final Long domainId, final String affinityGroupName, final String affinityGroupType,
-        final String description) {
+            final String description) {
 
         // validate the affinityGroupType
         Map<String, AffinityGroupProcessor> typeProcessorMap = getAffinityTypeToProcessorMap();
@@ -158,7 +158,7 @@ public class AffinityGroupServiceImpl extends ManagerBase implements AffinityGro
         verifyAffinityGroupNameInUse(owner.getAccountId(), owner.getDomainId(), affinityGroupName);
         verifyDomainLevelAffinityGroupName(domainLevel, owner.getDomainId(), affinityGroupName);
 
-        AffinityGroupVO group = createAffinityGroup(processor, owner, aclType, affinityGroupName, affinityGroupType, description);
+        AffinityGroupVO group = createAffinityGroup(processor, owner, aclType, affinityGroupName, affinityGroupType, description, domainLevel, domainId);
 
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Created affinity group =" + affinityGroupName);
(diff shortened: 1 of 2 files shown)


server/src/org/apache/cloudstack/affinity/AffinityGroupServiceImpl.java
```diff
@@ -176,25 +176,31 @@ private void verifyAccessToDomainWideProcessor(Account caller, AffinityGroupProc
         }
     }
 
-    private AffinityGroupVO createAffinityGroup(final AffinityGroupProcessor processor, final Account owner, final ACLType aclType, final String affinityGroupName, final String affinityGroupType, final String description) {
+    private AffinityGroupVO createAffinityGroup(final AffinityGroupProcessor processor, final Account owner, final ACLType aclType, final String affinityGroupName, final String affinityGroupType, final String description, boolean domainLevel, Long domainId) {
+
+        Long domainIdBasedOnDomainLevel = owner.getDomainId();
```
rafaelweingartner: @GabrielBrascher what about extracting lines 181-184 to a method and then create a test case for it? 
Also a documentation describing the method seems a good idea.

server/src/org/apache/cloudstack/affinity/AffinityGroupServiceImpl.java
```diff
@@ -323,12 +330,12 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
         List<String> types = new ArrayList<String>();
 
         for (AffinityGroupProcessor processor : _affinityProcessors) {
-                if (processor.isAdminControlledGroup()) {
-                    continue; // we dont list the type if this group can be
-                              // created only as an admin/system operation.
-                }
-                types.add(processor.getType());
+            if (processor.isAdminControlledGroup()) {
```
rafaelweingartner: Are these formatting changes needed?
Are you using the CloudStack code formatter style? Or was the code not formatted properly?

GabrielBrascher: At first I tried to submit without those changes in lines which I did not touch; However, it failed to pass in check style. When I let my environment make the code formatting, then all checks passed. I can review those changes and see if I can minimize them. But some will be necessary.

rafaelweingartner: Ah, ok.
So, let the changes there

remibergsma:
Hi @GabrielBrascher thanks for the fix! Works better for me, although I think the `affinity_group` table still has IDs `1` for domain and account. Do you see that too?

server/src/org/apache/cloudstack/affinity/AffinityGroupServiceImpl.java
```diff
@@ -205,6 +208,20 @@ public AffinityGroupVO doInTransaction(TransactionStatus status) {
...
+     * @note: this method fixes a critical bug. More details in JIRA ticket CLOUDSTACK-9432.
+     */
+    protected Long getDomainIdBasedOnDomainLevel(final Account owner, boolean domainLevel, Long domainId) {
+        Long domainIdBasedOnDomainLevel = owner.getDomainId();
+        if (domainLevel) {
+            domainIdBasedOnDomainLevel = domainId;
+        }
+        return domainIdBasedOnDomainLevel;
+    }
+
```
GabrielBrascher: @rafaelweingartner thanks for the review. The code now is in a method with tests and documentation.

GabrielBrascher:
Hi @remibergsma . Thanks for the feedback!

(i) the account ID in the affinity_group table: I kept the account ID as the owner account id (in this case 1 because of the system account id) because I don't know the effect of letting it null; thus, the account ID stayed as it was being managed before this fix. Please feel free to provide any different approach if you have something in mind.

(i) the domain ID in affinity_group table: if I didn't miss anything, the domain ID should be from the dedicated domain ID (instead of 1). Line 186 `AffinityGroupVO group = new AffinityGroupVO(affinityGroupName, affinityGroupType, description, affinityGroupDomainId, owner.getId(), aclType);` creates an AffinityGroupVO with the domain id `affinityGroupDomainId`. I will test this again to be sure.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rafaelweingartner:
From me you have an LGTM.
@remibergsma what do you think here?

rhtyd:
@blueorangutan package

GabrielBrascher:
Thanks for the review @rafaelweingartner @houthuis @remibergsma @rhtyd. I think this PR is ready for merge.

rhtyd:
@GabrielBrascher okay, I'll run regression tests on master later. Kindly use 'squash merge' in future and commits should include the JIRA id in their commit summary such as `CLOUDSTACK-xxxx: short summary`...

PR #2125, opened 2017-05-22
CLOUDSTACK-9927: Root admin user should be forced to change password …

harikrishna-patnala:
The default password for the root admin in CloudStack is "password". The user is not required to change this password.

Using CloudStack with the default password is the same as using it with no password. An attacker could log onto the management UI or API and make changes to the system, delete or steal resources, and stop services.

This fix should not allow admin to continue in UI until  password is changed to something other than the default. Also, do not permit the admin to change his password back to the default one later. 
Also fixed the issue where the password check now runs only for root admin users and not others.

### DIFF
--- a/ui/scripts/installWizard.js
+++ b/ui/scripts/installWizard.js
@@ -16,18 +16,55 @@
 // under the License.
 (function($, cloudStack) {
     cloudStack.installWizard = {
+        changedFirstPassword: false,
         // Check if install wizard should be invoked
         check: function(args) {
-            $.ajax({
-                url: createURL('listZones'),
-                dataType: 'json',
-                async: true,
-                success: function(data) {
-                    args.response.success({
-                        doInstall: !data.listzonesresponse.zone
-                    });
-                }
-            });
+            if(isAdmin()) {
+                $.ajax({
+                    url: createURL('listDomains'),
+                    data: {
+                        listAll: true,
+                        level: '0'
+                    },
+                    dataType: 'json',
+                    async: false,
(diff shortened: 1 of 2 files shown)


DaanHoogland:
Though this might seem as a reasonable security feature, I have doubts. The integration automation highly depends on the password being 'password'. Are we somehow allowing behaviour where we don't change the password or change it in headless installs to a known setting?

rafaelweingartner:
@DaanHoogland this proposal only affects the UI.
The automation you run only use the API, right?

This means we will only force users to change the admin password if they try to log in using the API.
Because the changes only affect users that use the UI, I was/am a little unsure of the increase in security it might bring.

harikrishna-patnala:
Automation runs use 8096 port where we don't need any password. This should not effect test runs.

wido:
I don't really see a need for this. Isn't this up to the user to change it? Why should we force it?

rhtyd:
I agree with @wido we should not force users to change password, however in the release notes or admin/install notes we can add a section on things to do after first install.

rhtyd:
Thanks @harikrishna-patnala for the PR. I think the intention to force the admin to change the admin password is better to be suggested/enforced when `cloudstack-setup-databases` is run, i.e. during installation of the mgmt server but not via the UI. Given views from others, I'll close the PR but encourage sending another PR where the suggestion/enforcement may be performed via the `cloudstack-setup-databases` script which may display/print a warning when the default user/admin password is not changed.

PR #2084, opened 2017-05-06
Passwd speedup

DaanHoogland:

PR #2155, opened 2017-06-20
Work around for guest VM password reset failure (PR/2084)

krissterckx:
… providers.

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/configure.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/configure.py
@@ -945,8 +945,10 @@ def main(argv):
 
         if process_file in ["cmd_line.json", "vm_password.json"]:
             logging.debug("Configuring vmpassword")
-            password = CsPassword("vmpassword", config)
-            password.process()
+            # TODO fix the line below
+            # password = CsPassword("vmpassword", config)
+            # password.process()
+            iptables_change = True
 
         if process_file in ["cmd_line.json", "vm_metadata.json"]:
             logging.debug("Configuring vmdata")


systemvm/patches/debian/config/opt/cloud/bin/configure.py
```diff
@@ -945,8 +945,10 @@ def main(argv):
 
         if process_file in ["cmd_line.json", "vm_password.json"]:
             logging.debug("Configuring vmpassword")
-            password = CsPassword("vmpassword", config)
-            password.process()
+            # TODO fix the line below
+            # password = CsPassword("vmpassword", config)
```
DaanHoogland: I don't feel confident this will be backwards compatible. We are now not configuring the vm with a password.

prashanthvarma:
This is the error we see with the CsPassword class in the configure.py systemvm patch script (cloud.log on VR):

2017-06-20 12:28:47,015  configure.py main:943 Configuring Guest Network
2017-06-20 12:28:47,015  configure.py main:947 Configuring vmpassword
2017-06-20 12:28:47,015  configure.py main:1034 Exception while configuring router
  File "/opt/cloud/bin/configure.py", line 948, in main
    password = CsPassword("vmpassword", config)
NameError: global name 'CsPassword' is not defined
Tue Jun 20 12:28:53 UTC 2017 : VR config: configuation format version 1.0
Tue Jun 20 12:28:53 UTC 2017 : VR config: creating file: /var/cache/cloud/ip_associations.json
Tue Jun 20 12:28:53 UTC 2017 : VR config: create file success
Tue Jun 20 12:28:53 UTC 2017 : VR config: executing: /opt/cloud/bin/update_config.py ip_associations.json

Here is an example cloud.log in a VR on a Native ACS setup:
[cloud.log.zip](https://github.com/apache/cloudstack/files/1091905/cloud.log.zip)

systemvm/patches/debian/config/opt/cloud/bin/configure.py
prashanthvarma: @DaanHoogland This block of code also fails on Native ACS setups. Please refer my latest comment on this PR.

rhtyd:
@krissterckx @prashanthvarma if `CsPassword` is not available, should we explore why that's the case. If it's unnecessary, we can remove/comment it, though I'm not sure if this causes any backward compatibility issues, and other regressions around vm lifecycle events (shutdown, reboot etc)?

@blueorangutan package

prashanthvarma:
@rhtyd To give you more context, we started seeing password reset failures on our Nuage VSP setups due to missing iptable rules on the VR, which are not being configured when we are not using specific services being provided by the VR (e.g. LB, setting iptables_change=True). Thus, we are adding the iptables_change=True flag to the CsPassword block as we use VR as the password reset service provider.

We found out that CsPassword block always fails due to import error (Note: even for Native ACS setups, refer the above attached cloud.log), which we tried to fix and ran into other issues with that block of code. Thus, we don't think that any of the functionality is being affected by this block of code. 

Anyhow, someone should fix and analyze the impact of this block on the functionality.

krissterckx:
Hi

so just to be very clear :
The line in code in current master which we commented out

password = CsPassword("vmpassword", config)

can never really work for two reasons

- CsPassword is not imported in this file
- the constructor is wrong, as it takes databag as input now

I can leave out the comment lines and the TODO statement in the PR obviously but i wanted the author/expert eyes to look at this before doing that.

Kris

rhtyd:
@blueorangutan test matrix

ustcweizhou:
I think it is better to revert PR #2084

karuturi:
@remibergsma can you please review?

krissterckx:
Thanks for driving the work at this useful PR, but i believe it needs more soak time.

We found today this PR is breaking VR-provided Guest DNS for SDN providers also.

I would recommend taking it out 4.10 and give it some more cycles in 4.11.

thanks,

Kris

- Nuage Networks

krissterckx:
I recommend closing this PR and reverting https://github.com/apache/cloudstack/pull/2084 instead i.e give PR/2048 more soak time in 4.11.    I want to support qualifying it in the upcoming release.

It is breaking VR-provided Guest DNS for SDN providers also, as we found today.

Best,

luhaijiao:
Yes,  think we need rever #2084 and release another RC of 4.10.




在2017年06月23 03时42分, "Kris Sterckx"<notifications@github.com>写道:



I recommend closing this PR and reverting #2084 instead i.e give PR/2048 more soak time in 4.11.    I want to support qualifying it in the upcoming release.

It is breaking VR-provided Guest DNS for SDN providers also, as we found today.

Best,

krissterckx:
Closing this PR as #2084 was reverted from RC5 -> RC6.

PR #2160, opened 2017-06-26
Revert 2084

karuturi:


shivajeesharma:
Pull from apache master on 13 July 2017

PR #2172, opened 2017-07-11
Travis: Fix usage of sudo enable Trusty images

rhtyd:
Adds group to travis yml file based on:
https://blog.travis-ci.com/2017-06-19-trusty-updates-2017-Q2

This can be accepted if Travis passes. Pinging for review - @DaanHoogland @borisstoyanov @wido @karuturi @PaulAngus and others

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,10 @@
 # under the License.
 sudo: required
 dist: trusty
+group: edge
 language: java
 jdk:
-- oraclejdk7
+- openjdk7
 python:
   - "2.7"
 cache:


DaanHoogland:
looks good. so if it solves it we should fwd merge to 4.10 first and to master next.

rhtyd:
Travis is green, this can be accepted to 4.9, fwd-merge to 4.10 -> and to master. /cc @DaanHoogland @wido @karuturi @nvazquez and others

rhtyd:
We've two lgtms and Travis is passing I'll fwd merge this.

karuturi:
Hi @rhtyd 
I am seeing the below error on latest master build

"Exception in thread "main" java.lang.UnsupportedClassVersionError: XmlToHtmlConverter has been compiled by a more recent version of the Java Runtime (class file version 53.0), this version of the Java Runtime only recognizes class file versions up to 52.0"

I think javac is 1.9 and java is 1.8

rhtyd:
Thanks @karuturi let me check the 4.10 and master builds, I've set jdk version to openjdk8 which should work but I'll re-check and fix appropriately.

rhtyd:
@karuturi it seems to be a bug with Travis, while switching to openjdk8 stack javac versions return 1.9 however java version is 1.8.

rhtyd:
@karuturi I've logged a bug with Travis https://github.com/travis-ci/travis-ci/issues/8075 and opened a PR for fixing the issue for 4.10+ to use oraclejdk8 but new trusy images.

PR #2174, opened 2017-07-12
CLOUDSTACK-9996 :  bug in juniper srx Resource vlan untagged jedgement and xml element replacment.

digitalgust:
fix  juniper vlan untagged jedgement and xml element replacment.


wido:
Seems like a very simple fix. But I don't know how to test nor verify it.

Can you elaborate a bit?

digitalgust:
We have change it work in production environment , our cloudstack have three Zone , one is  Juniper SRX 550 hardware device and the other two is juniper  VSRX ( juniper srx in vm ),  the old code would be throw   Exceptions .
1.
If the public interface has marked untagged on create zone ,the vlan is
“vlan://untagged” ,but old source compare equals “untagged” , so it will raise an
Exception : Unable to parse VLAN tag: vlan://untagged
2.
%action% not replaced , error shows in code , bracket pos error. It will report syntax error when configure
outgress rule to Juniper srx .

digitalgust:
https://github.com/apache/cloudstack/pull/2178

this PR commit by  my colleague :)

we have more juniper srx usage modification will submit.

wido:
LGTM based on the code

rhtyd:
LGTM

PR #372, opened 2015-06-08
Embedded Tomcat & Jetty

rsafonseca:
Well, this spruces things up a bit :)

Some of the features/fixes (i may have forgotten a couple as this involved a lot of changes ;) )
- Embedded Tomcat support
- Embedded Jetty support
- Reduce management .deb package to roughly half the size
- Reduce common .rpm package to roughly half the size
- Transparent support for more distros like Fedora 21
- Reduced dependencies on the packages, such as mysql-connector-java which are now bundled
- Upgraded Jetty version used by Travis from oldie and discontinued morbay 6.x version to current eclipse 9.x version
- Tidy up debian packaging under /packaging and include support on package.sh
- Cleanup of a lot of duplicate files which made it harder to maintain and caused some fixes to be implemented only for some distros
- Consolidates a few configs that were different across distros and shouldn't be
- Enables running with any version of openjdk runtime and compiling with any version of openjdk devel as long as they meet the minimum requirements (Java 7+) without conflicting with other packages.

The package.sh script packages ASF Tomcat by default, you can package it with Jetty instead by passing -j or --jetty :)

There's still a couple more things to consolidate and a couple more places to remove hardcoding of configs like the cloud user, but i'll take care of that on a later PR :)

PR #2197, opened 2017-07-23
CLOUDSTACK-10012: Embedded Jetty for CloudStack

rhtyd:
This is work-in-progress WIP embedded jetty support. Feedback welcome!

Changes adapted from PR #372 and #1546

Signed-off-by: Rohit Yadav <rohit.yadav@shapeblue.com>

### DIFF
--- a/packaging/cloud-management.sh
+++ b/packaging/cloud-management.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+ . /etc/cloudstack/management/tomcat6.conf
+SERVLETENGINE='jetty'
+# For SELinux we need to use 'runuser' not 'su'
+if [ -x "/sbin/runuser" ]; then
+    SU="/sbin/runuser -s /bin/sh"
+else
+    SU="/bin/su -s /bin/sh"
+fi
+
+touch $CLOUD_PID;
+chown $CLOUD_USER. $CLOUD_PID
+# CLASSPATH munging
+if [ -n "$JSSE_HOME" ]; then
+  CLASSPATH="${CLASSPATH}:$(build-classpath jcert jnet jsse 2>/dev/null)"
+fi
+CLASSPATH="${CLASSPATH}:$(build-classpath commons-daemon 2>/dev/null)"
+
+CLOUD_HOME="/usr/share/cloudstack-management/"
+CLOUD_BASE=$CLOUD_HOME
+
+NAME="cloudstack-management"
+
+# Define the tomcat username
+CLOUD_USER="${CLOUD_USER:-cloud}"
+
+# Define the tomcat group
+CLOUD_GROUP="${CLOUD_GROUP:-`id -gn $CLOUD_USER`}"
+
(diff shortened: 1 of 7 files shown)


marcaurele:
@rhtyd What's the reason to create another (sub)module of the client, instead of adding a profile to the client module to either produce a tomcat or jetty artifact?

Would be nice to created a shaded jar including Jetty server, to have a "short" classpath (bouncycastle JCE is still required on the classpath, therefore it's not possible to create a single fat jar to run due to the signature checks). I could help you on that since it's the way we run it ourself.

rhtyd:
@marcaurele thanks, any help would be great. The important goal is to move away from distro provided tomcat, I'm open to all possible ways to do that. Having a fat jar would be great, but for the first step if that's not possible I can do without that approach.

packaging/cloud-management.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+
+ . /etc/cloudstack/management/tomcat6.conf
+SERVLETENGINE='jetty'
+# For SELinux we need to use 'runuser' not 'su'
+if [ -x "/sbin/runuser" ]; then
+    SU="/sbin/runuser -s /bin/sh"
+else
+    SU="/bin/su -s /bin/sh"
+fi
```
wido: You don't seem to use this?

packaging/cloud-management.sh
```diff
@@ -0,0 +1,58 @@
...
+# Define the tomcat log file
+CLOUD_LOG="${CLOUD_LOG:-/var/log/${NAME}-initd.log}"
+
+# Define the pid file name
+# If change is needed, use sysconfig instead of here
+export CLOUD_PID="${CLOUD_PID:-/var/run/${NAME}.pid}"
+CLASSPATH="${CLASSPATH}:${CLOUD_HOME}/lib/jetty-runner.jar"
+CLASS="org.eclipse.jetty.runner.Runner --classes /etc/cloudstack/management --path /client $CLOUD_HOME/webapps/client  --stop-port 8888 --stop-key monkeystop"
+
+if [ "$1" = "start" ]; then
```
wido: Shouldn't this just be a wrapper script that systemd can execute? As we want to start/stop everything through systemd

rhtyd:
@wido thanks, this does not work as of now and needs some further changes. I wanted to start and do this in open to gather some initial feedback and advise.

marcaurele:
@rhtyd How could we work together on a branch? My idea was to push what we have (still based on 4.4.2) and let you adapt it to make it work with upstream version since I don't have a way to test upstream correctly

rhtyd:
Thanks @marcaurele if you can share your idea as a PR, I can use it as reference and adapt as necessary. I think the general goal is to fix the problem, whatever way necessary. By moving away from tomcat (distro provided), can help in CloudStack's adoption (and as a fatjar somewhere in future).

rhtyd:
Closing this in light in #2207

PR #2200, opened 2017-07-24
[CLOUDSTACK-10015] Return storage provider with response to list storage pools

syed:
It would be useful if we return the provider name in the list storage pools response. This will be useful for example to identify different storages that are in use and their scope.


wido:
Code-wise it doesn't matter, but you seem to have optimize imports enabled. That seems like a lot of lines changed, but it didn't.

Otherwise: LGTM

rhtyd:
LGTM.
@blueorangutan package

rhtyd:
@blueorangutan test

syed:
Thanks @wido and @rhtyd

PR #2207, opened 2017-07-28
Embedded jetty inside shaded jar

marcaurele:
This PR is to remove the dependency on distro Tomcat to ship CloudStack directly with an embedded Jetty. The delivery will only be scripts, one fat jar, JCE jar and default properties files.

- [x] Remove Tomcat from requirement
- [x] Embed Jetty
- [x] Shaded jar
- [x] Modify init script (classpath must contain bouncycastle) + log4j & properties files)

### DIFF
--- a/client/src/org/apache/cloudstack/ServerDaemon.java
+++ b/client/src/org/apache/cloudstack/ServerDaemon.java
@@ -1,0 +1,205 @@
+//
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+package org.apache.cloudstack;
+
+import org.apache.commons.daemon.Daemon;
+import org.apache.commons.daemon.DaemonContext;
+import org.eclipse.jetty.jmx.MBeanContainer;
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.NCSARequestLog;
+import org.eclipse.jetty.server.NetworkTrafficServerConnector;
+import org.eclipse.jetty.server.RequestLog;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.HandlerCollection;
+import org.eclipse.jetty.server.handler.HandlerList;
+import org.eclipse.jetty.server.handler.RequestLogHandler;
+import org.eclipse.jetty.util.thread.QueuedThreadPool;
+import org.eclipse.jetty.util.thread.ThreadPool;
+import org.eclipse.jetty.webapp.Configuration;
+import org.eclipse.jetty.webapp.WebAppContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.lang.management.ManagementFactory;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+/***
+ * Daemon server class to start the embedded server, either through JSVC or directly inside a JAR.
+ * Parameter to configure the jetty server are:
+ * - jetty.port: to start jetty on the specific port (default: 8080)
+ * - jetty.host: to bind to specific interface (default: null = all)
+ * - jetty.requestlog: path to log file for requests (default: request.log)
+ */
+public class ServerDaemon implements Daemon {
+    private static final Logger logger = LoggerFactory.getLogger(ServerDaemon.class);
+    private static final String WEB_XML = "META-INF/webapp/WEB-INF/web.xml";
+    private static final String PROJECT_RELATIVE_PATH_TO_WEBAPP = "src/main/webapp";
+    private static final String REQUEST_LOG = "request.log";
+
+    private Server jettyServer;
+    private int port;
+    private String bindInterface;
+    private String requestLogFile;
+
+    public static void main(String... anArgs) throws Exception {
+        ServerDaemon csServer = new ServerDaemon();
+        csServer.init(null);
+        csServer.start();
+        csServer.join();
+    }
+
+    @Override
+    public void init(DaemonContext context) {
+        Properties props = System.getProperties();
+        setPort(Integer.parseInt(props.getProperty("jetty.port", "8080")));
+        setBindInterface(props.getProperty("jetty.host"));
+        setRequestLogFile(props.getProperty("jetty.requestlog", REQUEST_LOG));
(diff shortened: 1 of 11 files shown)


marcaurele:
@rhtyd I still have to add a couple of things. There isn't anymore any JSP for the UI, right?

client/src/org/apache/cloudstack/ServerDaemon.java
```diff
@@ -0,0 +1,205 @@
...
+        csServer.init(null);
+        csServer.start();
+        csServer.join();
+    }
+
+    @Override
+    public void init(DaemonContext context) {
+        Properties props = System.getProperties();
+        setPort(Integer.parseInt(props.getProperty("jetty.port", "8080")));
+        setBindInterface(props.getProperty("jetty.host"));
```
wido: To what does *jetty.host* default to? Reason I'm asking is IPv6. Just want to make sure we bind on '::' and not 0.0.0.0 so that IPv6 works out of the box

wido:
Very nice! Very welcome.

I think we also need a modified systemd service to make this work, right?

client/src/org/apache/cloudstack/ServerDaemon.java
marcaurele: Good point @wido but it should work since if there isn't any value it's binding to `*` which should includes IPv6, right?

rhtyd:
@marcaurele thanks yes, JSPs have been removed from UI now.

marcaurele:
@rhtyd I think I backported all necessary changes (for Ubuntu distro). Can you have a look & try?

rhtyd:
@marcaurele thanks, I'll have a look this week. I've one reservation -- if we put everything in a single jar, it might be difficult for users to apply UI changes and add custom jars, anyway to do that? Maybe add a custom path (add a classpath) for an override directory from where jars may be loaded and maybe split the UI as a separate rpm/deb?

rhtyd:
@blueorangutan package

wido:
I see you modified the init script, but as Ubuntu and CentOS have moved to systemd, shouldn't we fix that as well?

rhtyd:
@wido that should, however testing is required. I see `apidocs` build is broken that blocks us from further testing. /cc @marcaurele

rhtyd:
@marcaurele I've found the fix, can you put this commit in your branch as I cannot push changes:
```
diff --git a/tools/apidoc/build-apidoc.sh b/tools/apidoc/build-apidoc.sh
index 8ab69c8373..4e1f9a639d 100755
--- a/tools/apidoc/build-apidoc.sh
+++ b/tools/apidoc/build-apidoc.sh
@@ -47,7 +47,7 @@ fi
 
 CP=$PATHSEP/
 
-java -cp $CP$PATHSEP$TARGETJARDIR/*$PATHSEP$DEPSDIR/* com.cloud.api.doc.ApiXmlDocWriter -d "$DISTDIR" $*
+java -cp $CP$PATHSEP$TARGETJARDIR$PATHSEP$DEPSDIR com.cloud.api.doc.ApiXmlDocWriter -d "$DISTDIR" $*
 
 if [ $? -ne 0 ]
 then
diff --git a/tools/apidoc/gen_toc.py b/tools/apidoc/gen_toc.py
index a261037d80..39e1576422 100644
--- a/tools/apidoc/gen_toc.py
+++ b/tools/apidoc/gen_toc.py
@@ -180,7 +180,8 @@ known_categories = {
     'stopNetScalerVpx' : 'Load Balancer',
     'deleteServicePackageOffering' : 'Load Balancer',
     'destroyNsVpx' : 'Load Balancer',
-    'startNsVpx' : 'Load Balancer'
+    'startNsVpx' : 'Load Balancer',
+    'listElastistorInterface': 'Misc'
     }
 
 
diff --git a/tools/apidoc/pom.xml b/tools/apidoc/pom.xml
index e6804a0e78..61c1ccd593 100644
--- a/tools/apidoc/pom.xml
+++ b/tools/apidoc/pom.xml
@@ -21,8 +21,8 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
     <properties>
-        <client.config.base>../../client/target/cloud-client-ui-${project.version}/WEB-INF/</client.config.base>
-        <client.config.jars>${client.config.base}/lib</client.config.jars>
+        <client.config.base>../../client/target/</client.config.base>
+        <client.config.jars>${client.config.base}/cloud-client-ui-${project.version}.jar</client.config.jars>
         <client.config.conf>${client.config.base}/classes</client.config.conf>
     </properties>
```

rhtyd:
@marcaurele with the above fix I could get the build to work, however can you share how would you run the management server both using mvn and scripts? I'm unable to get it to work.

marcaurele:
Even if it looks like a fat/uber jar, the startup script still runs it as a normal multi jar. If you really want, you could have a single jar. Due to bouncycastle and the JCE signature check, it is not possible to run it as a single jar.

To start it, I use:
```
# normal start
java -Dlog4j.configuration=log4j-cloud.xml -cp "client/target/conf:client/target/bcprov-jdk15on-1.55.jar:client/target/cloud-client-ui-X.X.X.X-SNAPSHOT.jar" org.apache.cloudstack.ServerDaemon

# with remote debug for example and specifying a port for Jetty
java -Xdebug -Xnoagent -Xrunjdwp:transport=dt_socket,address=8787,server=y,suspend=n -Djava.net.preferIPv4Stack=true -Dlog4j.configuration=log4j-cloud.xml -Djetty.port=8080 -cp "client/target/conf:client/target/bcprov-jdk15on-1.55.jar:client/target/cloud-client-ui-X.X.X.X-SNAPSHOT.jar" org.apache.cloudstack.ServerDaemon
```

packaging/debian/init/cloud-management
```diff
@@ -103,30 +107,17 @@ if [ ! -f "$CATALINA_HOME/bin/bootstrap.jar" ]; then
...
-fi
+JAVA_OPTS="$JAVA_OPTS -Djava.io.tmpdir=$JVM_TMP"
 
-# Set juli LogManager if logging.properties is provided
-if [ -r "$CATALINA_BASE"/conf/logging.properties ]; then
-  JAVA_OPTS="$JAVA_OPTS "-Djava.util.logging.manager=org.apache.juli.ClassLoaderLogManager" "-Djava.util.logging.config.file="$CATALINA_BASE/conf/logging.properties"
+# Load extra arguments from java-opts.conf if any
+if [ -r "$MANAGEMENT_CONFIG/java-opts.conf" ]; then
+	JAVA_EXTRA_OPTS=$(cat $MANAGEMENT_CONFIG/java-opts.conf)
+	JAVA_OPTS="$JAVA_OPTS $JAVA_EXTRA_OPTS"
```
marcaurele: @rhtyd there's the inclusion of a java-opts.conf file if present that helps to add extra parameters to the classpath, like extra JARs, extra config directories, JMX...

marcaurele:
There should be an option to let maintainer edit a PR (on github), so I guess the content of it too. But it does not seem to be active on the repository (https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)

rhtyd:
@marcaurele thanks, I started working on this PR based on your branch to fix the centos7 issues: https://github.com/apache/cloudstack/pull/2226 this is on ASF remote so any committer should be able to commit or send PRs.

packaging/debian/init/cloud-management
rhtyd: Nice

marcaurele:
Ok, closing this one then in favor of #2226

rhtyd:
Thanks @marcaurele please collaborate on #2226 , feel free to directly push changes. I'll use your suggestions and incorporate them in the systemd scripts now, currently trying to fix centos packaging.

PR #2209, opened 2017-07-28
CLOUDSTACK-10017: Adding 'quick-test' attribute to list of smoketests…

borisstoyanov:
… and adding 'vpc', 'templates', 'deploy-vm', 'roles', 'service-offerings', 'storage', 'routers', 'networks', 'usage', 'accounts' tags

quick-test label aims to provide quick feedback of recent changes to the developer, instead of waiting for the whole smoketest to execute(24h+). This execution lasts about 1.5h at Trillian environments.

The categories aim to offer more efficient way to slice through tests in both smoke and component tests. Let's say as a developer you've made a change in the accounts. You can simply trigger marvin tests with tag=accounts pointing all the tests for example:
```

nosetests --with-marvin --marvin-config=[config] --hypervisor=xenserver -a tags=accounts /tests/integration/*/*.*

```
https://cwiki.apache.org/confluence/display/CLOUDSTACK/Test+Categories+and+Quick-test+in+Marvin

### DIFF
--- a/test/integration/component/maint/test_ha_pool_maintenance.py
+++ b/test/integration/component/maint/test_ha_pool_maintenance.py
@@ -144,7 +144,7 @@ class testHaPoolMaintenance(cloudstackTestCase):
         return
 
     @attr(tags=["advanced", "cl", "advancedns", "sg",
-                "basic", "eip", "simulator", "multihost"])
+                "basic", "eip", "simulator", "multihost", "storage"])
     def test_ha_with_storage_maintenance(self):
         """put storage in maintenance mode and start ha vm and check usage"""
         # Steps
--- a/test/integration/component/maint/test_vpc.py
+++ b/test/integration/component/maint/test_vpc.py
@@ -278,7 +278,7 @@ class TestVPC(cloudstackTestCase):
         sshClient.execute(command)
         return
 
-    @attr(tags=["advanced", "intervlan", "dvs", "test"], required_hardware="true")
+    @attr(tags=["advanced", "intervlan", "dvs", "test", "vpc"], required_hardware="true")
     def test_01_create_tier_Vmxnet3(self):
         """
             Test to create vpc tier with nic type as Vmxnet3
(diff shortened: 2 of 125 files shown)


rhtyd:
Good effort, @borisstoyanov can you rebase Github is such a complainer :)

borisstoyanov:
Thanks @rhtyd, I've rebased and squashed all the commits

borisstoyanov:
ping @DaanHoogland @nvazquez @PaulAngus for review

rhtyd:
I must share that as a developer I don't see any value in this, as for any change I would do I would run a full smoketest anyway, and for component/resource specific test run I can already run a single test file which are already grouped by resource/component, for example vm_lifecycle test is all related to a vm etc so I could run that individual test.

@borisstoyanov can you pull rebase against latest master?

borisstoyanov:
I've just rebased @rhtyd, a full smoketest would take day+ and this label gives you a real quick feedback if your changes brake anything fundamentally. 

Actually in theory smoketests should be really quick and should give near instant result back. Their main outcome should be yes/no answer to 'is the build testable' question (as per common testing practice). I think this is where we got this wrong and kept on adding and adding more items to the suite, instead it should be the other way around, having a really short smoketest and more extensive regression testing suite.

PaulAngus:
in theory i agree @borisstoyanov - the problem there is how do you define fundamental.
and given how large the component suite is (and how badly broken), results from component tests tend to get ignored.

We could split the tests in to 3 categories 

- a set of true smoke tests basic vm lifecycle and network tests that take just a hour or two to run
- a set of known good component tests - the remainder of smoke tests and known good component tests
- broken tests - ones that rarely, if ever pass

The broken tests can live in a separate directory, and as they get fixed, can be moved into the known good directory to be run in future.

but by hook or by crook we need to get to a point where a failure *_really_* means a code issue.

then we can be much more boolean about quality.

DaanHoogland:
@borisstoyanov please rebase and re-open if still relevant

PR #2213, opened 2017-07-31
CLOUDSTACK-10023: Windows VMs with Multiple Nics fail to access internet

jayakarteek:
REPRO STEPS:
Create a Windows VM with Multiple NICs
Ensure that only one of the NICs can actually access internet (set that as default NIC in ACS)
Try to access internet

EXPECTED BEHAVIOR:
Windows VM should be able to access internet

ACTUAL BEHAVIOR:
Windows VM is not able to access internet

Windows will experience issue if we have multiple default gateway assigned as explained in "Multiple Default Gateways Can Cause Connectivity Problems" (https://support.microsoft.com/en-us/kb/159168).  

You may want to make Windows to select certain gateway always, we can change interface metrics:
 
Go to your Network adapter properties
Select Internet Protocol v4.
Click the Advanced button on General tab.
Uncheck Automatic metric.
Enter a number higher than 10 to the Interface metric field.

The default automatic metric for interfaces is 10. The lower the metric, the higher the preference to use it. So, we have to enter metric higher than 10 for the interfaces you don't want to use. This is because if Windows has multiple interfaces connected to several networks that use DHCP, it installs default gateways for all those interfaces. By default, Windows sets the metric of the default gateways to the same, effectively leading to random selection of default gateway.

Window stand point the above mentioned issue can be resolved by setting metric for NICs inside guest OS

### DIFF
--- a/server/src/com/cloud/network/router/CommandSetupHelper.java
+++ b/server/src/com/cloud/network/router/CommandSetupHelper.java
@@ -109,6 +109,8 @@ import com.cloud.offering.NetworkOffering;
 import com.cloud.offerings.NetworkOfferingVO;
 import com.cloud.offerings.dao.NetworkOfferingDao;
 import com.cloud.service.dao.ServiceOfferingDao;
+import com.cloud.storage.GuestOSVO;
+import com.cloud.storage.dao.GuestOSDao;
 import com.cloud.user.Account;
 import com.cloud.uservm.UserVm;
 import com.cloud.utils.Pair;
@@ -173,6 +175,8 @@ public class CommandSetupHelper {
     private VlanDao _vlanDao;
     @Inject
     private IPAddressDao _ipAddressDao;
+    @Inject
+    private GuestOSDao _guestOSDao;
 
     @Inject
     private RouterControlHelper _routerControlHelper;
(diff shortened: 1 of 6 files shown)


server/src/com/cloud/network/router/CommandSetupHelper.java
```diff
@@ -216,6 +220,15 @@ public void createDhcpEntryCommand(final VirtualRouter router, final UserVm vm,
                 _networkModel.getExecuteInSeqNtwkElmtCmd());
 
         String gatewayIp = nic.getIPv4Gateway();
+        if (!nic.isDefaultNic()) {
+            final GuestOSVO guestOS = _guestOSDao.findById(vm.getGuestOSId());
+            if (guestOS == null || !guestOS.getDisplayName().toLowerCase().contains("windows")) {
+                gatewayIp = "0.0.0.0";
```
rhtyd: @jayakarteek I'm not sure it's best to put a `0.0.0.0` as gateway IP

rhtyd:
@jayakarteek let's find another way to fix this, I'm -1 on having assumptions about the type of guest VMs

DaanHoogland:
@jayakarteek please rebase and re-open if still relevant

PR #2221, opened 2017-08-03
[CLOUDSTACK-10030] - Make publicips reachable inside VPC

syed:
If a public IP is assigned to a VPC, a VM running inside that VPC cannot ping that public IP. This is due to the IPtables Nat rules set in such a way that drop any requests to the public IP from internal interfaces. I am fixing this so that internal hosts can also reach the public IP.

Reproduction:

Create a VPC
Create a network inside the VPC
Allocate a public IP
Create a VM in the network
Create a port forwarding rule enabling ICMP
ping the public IP inside the VM (this will fail)

### DIFF
--- a/systemvm/patches/debian/config/opt/cloud/bin/configure.py
+++ b/systemvm/patches/debian/config/opt/cloud/bin/configure.py
@@ -886,7 +886,7 @@ class CsForwardingRules(CsDataBag):
         self.fw.append(["filter", "", fw7])
 
     def forward_vpc(self, rule):
-        fw_prerout_rule = "-A PREROUTING -d %s/32 -i %s" % (rule["public_ip"], self.getDeviceByIp(rule['public_ip']))
+        fw_prerout_rule = "-A PREROUTING -d %s/32 " % (rule["public_ip"])
         if not rule["protocol"] == "any":
             fw_prerout_rule += " -m %s -p %s" % (rule["protocol"], rule["protocol"])
         if not rule["public_ports"] == "any":


rhtyd:
@blueorangutan package

systemvm/patches/debian/config/opt/cloud/bin/configure.py
```diff
@@ -886,7 +886,7 @@ def forward_vr(self, rule):
         self.fw.append(["filter", "", fw7])
 
     def forward_vpc(self, rule):
-        fw_prerout_rule = "-A PREROUTING -d %s/32 -i %s" % (rule["public_ip"], self.getDeviceByIp(rule['public_ip']))
+        fw_prerout_rule = "-A PREROUTING -d %s/32 " % (rule["public_ip"])
```
rhtyd: @syed should we fix `forward_vr` as well toward the same behaviour?

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test centos6 vmware-55u3

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:

rhtyd:

systemvm/patches/debian/config/opt/cloud/bin/configure.py
syed: @rhtyd This is already fixed in `forward_vr` as there is only one internal network for each VR so the IPtables rules work correctly there.

rhtyd:
LGTM, no new errors, existing errors are known intermittent errors. @syed if possible send a marvin test for this case in a separate PR. I'll merge this now.

PR #2228, opened 2017-08-08
CLOUDSTACK-10036: Decreasing timeout of failing unit test

borisstoyanov:
https://issues.apache.org/jira/browse/CLOUDSTACK-10036

This test occasionally fails on CentOS6 environments by failing to meet the 2000 milliseconds threshold. Usually it ends up executing the method for ~1100. So decreasing the timeout to 1000 would prevent it from failing.


borisstoyanov:
@blueorangutan package

borisstoyanov:
the build has passed without any failures

rhtyd:
LGTM, this is a simple unit test failure fix. I'll proceed with merging this as per of stablizing master branch against intermittent errors.

PR #2239, opened 2017-08-18
CLOUDSTACK-9993: Securing Agents Communications

rhtyd:
This introduces a new certificate authority framework that allows                  
pluggable CA provider implementations to handle certificate operations             
around issuance, revocation and propagation. The framework injects                 
itself to `NioServer` to handle agent connections securely. The                    
framework adds assumptions in `NioClient` that a keystore if available             
with known name `cloud.jks` will be used for SSL negotiations and                  
handshake.                                                                         

FS: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Secure+Agent+Communications                                                  
                                 
This includes a default 'root' CA provider plugin which creates its own            
self-signed root certificate authority on first run and uses it for                
issuance and provisioning of certificate to CloudStack agents such as              
the KVM, CPVM and SSVM agents and also for the management server for               
peer clustering.                                                                   
                                                                                   
Additional changes and notes:                                                      
- Comma separate list of management server IPs can be set to the 'host'            
  global setting. Newly provisioned agents (KVM/CPVM/SSVM etc) will get            
  radomized comma separated list to which they will attempt connection             
  or reconnection in provided order. This removes need of a TCP LB on              
  port 8250 (default) of the management server(s).                                 
- All fresh deployment will enforce two-way SSL authentication where               
  connecting agents will be required to present certificates issued                
  by the 'root' CA plugin.                                                         
- Existing environment on upgrade will continue to use one-way SSL                 
  authentication and connecting agents will not be required to present             
  certificates.                                                                    
- A script `keystore-setup` is responsible for initial keystore setup              
  and CSR generation on the agent/hosts.                                           
- A script `keystore-cert-import` is responsible for import provided               
  certificate payload to the java keystore file.                                   
- Agent security (keystore, certificates etc) are setup initially using            
  SSH, and later provisioning is handled via an existing agent connection          
  using command-answers. The supported clients and agents are limited to           
  CPVM, SSVM, and KVM agents, and clustered management server (peering).           
- Certificate revocation does not revoke an existing agent-mgmt server             
  connection, however rejects a revoked certificate used during SSL                
  handshake.                                                                       
- Older `cloudstackmanagement.keystore` is deprecated and will no longer           
  be used by mgmt server(s) for SSL negotiations and handshake. New                
  keystores will be named `cloud.jks`, any additional SSL certificates             
  should not be imported in it for use with tomcat etc. The `cloud.jks`            
  keystore is stricly used for agent-server communications.                        
- Management server keystore are validated and renewed on start up only,           
  the validity of them are same as the CA certificates.  

New APIs:                                                                          
- listCaProviders: lists all available CA provider plugins                         
- listCaCertificate: lists the CA certificate(s)                                   
- issueCertificate: issues X509 client certificate with/without a CSR              
- provisionCertificate: provisions certificate to a host                           
- revokeCertificate: revokes a client certificate using its serial                 
                                                                                   
Global settings for the CA framework:                                              
- ca.framework.provider.plugin: The configured CA provider plugin                  
- ca.framework.cert.keysize: The key size for certificate generation               
- ca.framework.cert.signature.algorithm: The certificate signature algorithm       
- ca.framework.cert.validity.period: Certificate validity in days                  
- ca.framework.cert.automatic.renewal: Certificate auto-renewal setting            
- ca.framework.background.task.delay: CA background task delay/interval            
- ca.framework.cert.expiry.alert.period: Days to check and alert expiring certificates
                                                                                   
Global settings for the default 'root' CA provider:                                
- ca.plugin.root.private.key: (hidden/encrypted) CA private key                    
- ca.plugin.root.public.key: (hidden/encrypted) CA public key                      
- ca.plugin.root.ca.certificate: (hidden/encrypted) CA certificate                 
- ca.plugin.root.issuer.dn: The CA issue distinguished name                        
- ca.plugin.root.auth.strictness: Are clients required to present certificates     
- ca.plugin.root.allow.expired.cert: Are clients with expired certificates allowed 
                                                                                   
UI changes:                                                                        
- Button to download/save the CA certificates


rhtyd:
Pinging for review @nvazquez @borisstoyanov @DaanHoogland @wido @marcaurele @GabrielBrascher @rafaelweingartner @ustcweizhou @pyr and others

@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test centos6 vmware-55u3

rhtyd:
The one test failing `create_volume` is not related to this PR but general intermittent failures, include those rvpc/rvr failures and are ignorable. 

The failure was caused due to;
```
sshClient: DEBUG:  Host: 10.1.34.164 Cmd: /sbin/fdisk -l | grep Disk Output:{'status': 'SUCCESS', 'stdin': None, 'stderr': [u"Disk /dev/sda doesn't contain a valid partition table\n"], 'stdout': [u'Disk /dev/hdb: 2147 MB, 2147483648 bytes\n', u'Disk /dev/sda: 1073 MB, 1073741824 bytes\n']}
test_01_create_volume (tests.smoke.test_volumes.TestCreateVolume): DEBUG:  Volume Size Expected 1073741824  Actual :Volume Not Found
```

wido:
What I am missing is that it's enforced/mandatory for all Agents, right?

What if somebody doesn't want to enable this and simply wants to run unencrypted. Why shouldn't we make a opt-out somewhere?

rhtyd:
@wido all agent-mgmt server connections are encrypted and SSL enabled, based on a random cert that the mgmt server creates, stores and uses from `cloudmanagement.keystore` reading/updating from ssl.keystore global setting; and when agents connect to the mgmt server they use a trust-all-manager to trust any certificate presented to them. So, all cloudstack env have the connections encrypted, however not authenticated and secured in one or two-way SSL.

For existing environments, after upgrade the auth strictness enforcement will be `'false`, however newer hosts/agents will be provisioned to use this new system to have CA certs etc stored in agents so they will have more secure SSL authentication, however, mgmt server will not perform additional trust validations and it will allow any clients (like it used to now).

wido:
Ok, good! Because I know many (like us) do not use SSH from the mgmt server to add a new KVM host, but simply generate a agent.properties and add it to the mgmt server.

rhtyd:
@wido okay, so for you environment just keep `ca.plugin.root.auth.strictness` set to `false` for both new and existing cloudstack environments

rhtyd:
@blueorangutan package

borisstoyanov:
@blueorangutan package

utils/pom.xml
```diff
@@ -72,6 +77,10 @@
       <artifactId>bcprov-jdk15on</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.bouncycastle</groupId>
```
DaanHoogland: is this specific or can it be inheritied from teh root-project

utils/src/main/java/org/apache/cloudstack/utils/security/KeyStoreUtils.java
```diff
@@ -0,0 +1,70 @@
...
+import java.io.File;
+import java.io.IOException;
+
+import com.cloud.utils.script.Script;
+import com.google.common.base.Strings;
+
+public class KeyStoreUtils {
+
+    public static String defaultTmpKeyStoreFile = "/tmp/tmp.jks";
+    public static String defaultKeystoreFile = "/cloud.jks";
```
DaanHoogland: with "/" seems dangerous/vulnerable. I'd change it to "~/" or none (and move the '/' to a directory configuration.

utils/src/main/java/org/apache/cloudstack/utils/security/KeyStoreUtils.java
```diff
@@ -0,0 +1,70 @@
...
+    public static String defaultPrivateKeyFile = "/cloud.key";
```
DaanHoogland: same for this and below

agent/src/com/cloud/agent/Agent.java
```diff
@@ -166,7 +178,8 @@ public Agent(final IAgentShell shell, final int localAgentId, final ServerResour
             throw new ConfigurationException("Unable to configure " + _resource.getName());
         }
 
-        _connection = new NioClient("Agent", _shell.getHost(), _shell.getPort(), _shell.getWorkers(), this);
+        final String host = _shell.getHost();
```
rhtyd: @DaanHoogland yes, the `getHost` is supposed to internally loop/return the next address.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -464,7 +493,7 @@ protected void processRequest(final Request request, final Link link) {
 
             for (int i = 0; i < cmds.length; i++) {
                 final Command cmd = cmds[i];
-                Answer answer;
+                Answer answer = null;
```
rhtyd: this can be removed, an uninitialized variable is `null`, it just states it explicitly.

agent/src/com/cloud/agent/AgentShell.java
```diff
@@ -107,7 +108,16 @@ public String getPod() {
 
     @Override
     public String getHost() {
-        return _host;
+        String[] hosts = _host.split(",");
+        if (_hostCounter >= hosts.length) {
+            _hostCounter = 0;
+        }
+        s_logger.info("Connecting to host: " + hosts[_hostCounter % hosts.length]);
```
rhtyd: I'll refactor this, however this prints the next 'host' the `Agent` class may attempt to connect to.

agent/src/com/cloud/agent/AgentShell.java
```diff
@@ -107,7 +108,16 @@ public String getPod() {
...
+        return hosts[_hostCounter++ % hosts.length];
+    }
+
+    public void setHost(final String host) {
```
rhtyd: For historic reasons, I simply created getters/setters for `_host`, this method is only used for writing a unit test.

agent/test/com/cloud/agent/AgentShellTest.java
```diff
@@ -44,4 +48,15 @@ public void loadProperties() throws ConfigurationException {
         Assert.assertNotNull(shell.getProperties());
         Assert.assertFalse(shell.getProperties().entrySet().isEmpty());
     }
+
+    @Test
+    public void testGetHost() {
+        AgentShell shell = new AgentShell();
+        List<String> hosts = Arrays.asList("10.1.1.1", "20.2.2.2", "30.3.3.3", "2001:db8::1");
+        shell.setHost(StringUtils.listToCsvTags(hosts));
+        for (String host : hosts) {
```
rhtyd: That's the case for now.

agent/src/com/cloud/agent/dao/impl/PropertiesStorage.java
```diff
@@ -51,6 +51,9 @@ public synchronized String get(String key) {
 
     @Override
     public synchronized void persist(String key, String value) {
+        if (!loadFromFile(_file)) {
+            s_logger.warn("Failed to load changes and then write to them");
```
rhtyd: There is nobody to read the values, I refactored existing methods to allow persisting passwords/passphrases for keystore. I can refactor this to throw an exception, however that changes the previous implementation as persisting changes is not mandatory, the agent reads stuff from cmdline (in systemvms) etc. and sometimes receives information from readycommand.

api/src/org/apache/cloudstack/alert/AlertService.java
```diff
@@ -67,6 +67,7 @@ private AlertType(short type, String name, boolean isDefault) {
         public static final AlertType ALERT_TYPE_SYNC = new AlertType((short)27, "ALERT.TYPE.SYNC", true);
         public static final AlertType ALERT_TYPE_UPLOAD_FAILED = new AlertType((short)28, "ALERT.UPLOAD.FAILED", true);
         public static final AlertType ALERT_TYPE_OOBM_AUTH_ERROR = new AlertType((short)29, "ALERT.OOBM.AUTHERROR", true);
+        public static final AlertType ALERT_TYPE_CA_CERT = new AlertType((short)31, "ALERT.CA.CERT", true);
```
rhtyd: Yep, 30 is reserved for 'host-ha'. I wanted to avoid changing the numbers.

rhtyd: Yes.

api/src/org/apache/cloudstack/api/command/admin/ca/IssueCertificateCmd.java
```diff
@@ -0,0 +1,162 @@
...
+        responseHasSensitiveInfo = false,
+        since = "4.11.0",
+        authorized = {RoleType.Admin})
+public class IssueCertificateCmd extends BaseAsyncCmd {
+    private static final Logger LOG = Logger.getLogger(IssueCertificateCmd.class);
+
+    public static final String APINAME = "issueCertificate";
+
+    @Inject
+    private CAManager caManager;
```
rhtyd: Putting inject on `BaseCmd` class would slow the bean creation and injections. I avoid such a pattern, keep classes light weight. This is the pattern we've been using for several features now, including dynamic roles, oobm, host-ha etc.

rhtyd: -do-

api/src/org/apache/cloudstack/api/command/admin/ca/IssueCertificateCmd.java
```diff
@@ -0,0 +1,162 @@
...
+        return provider;
+    }
+
+    /////////////////////////////////////////////////////
+    /////////////// API Implementation///////////////////
+    /////////////////////////////////////////////////////
+
+    @Override
+    public void execute() {
+        if (Strings.isNullOrEmpty(getCsr()) && getDomains().isEmpty()) {
```
rhtyd: preconditions might throw exceptions, however we would want to throw specific errors/api errors etc.

api/src/org/apache/cloudstack/api/command/admin/ca/IssueCertificateCmd.java
```diff
@@ -0,0 +1,162 @@
...
+        } catch (final IOException e) {
+            LOG.error("Failed to generate and convert client certificate(s) to PEM due to error: ", e);
+            throw new ServerApiException(ApiErrorCode.INTERNAL_ERROR, "Failed to process and return client certificate");
+        }
+        certificateResponse.setResponseName(getCommandName());
+        setResponseObject(certificateResponse);
+    }
+
+    @Override
+    public String getCommandName() {
```
rhtyd: We can however, it would require a codebase wide refactoration.

api/src/org/apache/cloudstack/api/response/CertificateResponse.java
```diff
@@ -0,0 +1,58 @@
...
+
+    public CertificateResponse(final String objectName) {
+        setObjectName(objectName);
+    }
+
+    public void setCertificate(final String certificate) {
+        this.certificate = certificate;
+    }
+
+    public void setPrivateKey(final String privateKey) {
```
rhtyd: The response in this case is necessary, as the issue Cert is supposed to return private key when CSR is provided. The API is restricted to root-admin usage now, this is necessary. I'll see what I can do.

core/src/com/cloud/agent/api/routing/NetworkElementCommand.java
```diff
@@ -46,6 +47,18 @@ protected NetworkElementCommand() {
         super();
     }
 
+    public void setAccessDetail(final Map<String, String> details) {
+        if (details == null) {
+            return;
+        }
+        for (final Map.Entry<String, String> detail : details.entrySet()) {
```
rhtyd: @DaanHoogland sorry don't understand the comment, the idea was to provide/set a detail by passing a map here, which may override a detail (if so asked by the developer)

engine/orchestration/src/com/cloud/agent/manager/ClusteredAgentManagerImpl.java
```diff
@@ -495,28 +495,29 @@ public SocketChannel connectToPeer(final String peerName, final SocketChannel pr
...
+                int port = Port.value();
                 try {
                     addr = InetAddress.getByName(ip);
                 } catch (final UnknownHostException e) {
                     throw new CloudRuntimeException("Unable to resolve " + ip);
                 }
                 SocketChannel ch1 = null;
                 try {
-                    ch1 = SocketChannel.open(new InetSocketAddress(addr, Port.value()));
+                    ch1 = SocketChannel.open(new InetSocketAddress(addr, port));
```
rhtyd: we can however, I avoided much changes, as we were doing something with handshake etc.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -1073,6 +1079,24 @@ public void orchestrateStart(final String vmUuid, final Map<VirtualMachineProfil
                             if (s_logger.isDebugEnabled()) {
                                 s_logger.debug("Start completed for VM " + vm);
                             }
+                            final Host vmHost = _hostDao.findById(destHostId);
```
rhtyd: I'll explore this, it's funny because it was actually very huge and I actually have refactored it a lot, and we need cmd-answers needed to do deployments etc, I decided to keep it here for closeness.

engine/schema/src/com/cloud/certificate/dao/CrlDaoImpl.java
```diff
@@ -0,0 +1,57 @@
...
+    public CrlDaoImpl() {
+        super();
+
+        CrlBySerialSearch = createSearchBuilder();
+        CrlBySerialSearch.and("certSerial", CrlBySerialSearch.entity().getCertSerial(), SearchCriteria.Op.EQ);
+        CrlBySerialSearch.done();
+    }
+
+    @Override
+    public CrlVO findBySerial(final BigInteger certSerial) {
```
rhtyd: No, I don't see a risk here. The method needs to be public to allow subsystems to find by serial. How do you think it can be a risk, usage-wise?

framework/ca/src/org/apache/cloudstack/framework/ca/CAProvider.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Map;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+
+public interface CAProvider {
+
+    /**
+     * Method returns capability of the plugin to participate in certificate issuance, revocation and provisioning
+     * @return
```
rhtyd: That's what the description says, it was repetitive so I avoided that.

framework/ca/src/org/apache/cloudstack/framework/ca/CAProvider.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Issues certificate using given CSR and other options
+     * @param csr
+     * @param domainNames
+     * @param ipAddresses
+     * @param validityDays
+     * @return
+     */
+    Certificate issueCertificate(final String csr, final List<String> domainNames, final List<String> ipAddresses, final int validityDays);
```
rhtyd: The CSR may or may-not, for systemvm/kvm agents the CSR generated may not know additional names, ips (such as storage, public, link-local, private) etc that CloudStack knows so during certificate generation we use primary domain name from CSR only, and additional details is provided by CloudStack.

framework/ca/src/org/apache/cloudstack/framework/ca/CAProvider.java
```diff
@@ -0,0 +1,93 @@
...
+     * @return returns true on success
+     */
+    boolean revokeCertificate(final BigInteger certSerial, final String certCn);
+
+    /**
+     * This method can add/inject custom TrustManagers for client connection validations.
+     * @param sslContext The SSL context used while accepting a client connection
+     * @param remoteAddress
+     * @param certMap
+     * @return
```
rhtyd: Self explanatory

framework/ca/src/org/apache/cloudstack/framework/ca/CAProvider.java
```diff
@@ -0,0 +1,93 @@
...
+     * @throws GeneralSecurityException
+     * @throws IOException
+     */
+    SSLEngine createSSLEngine(final SSLContext sslContext, final String remoteAddress, final Map<String, X509Certificate> certMap) throws GeneralSecurityException, IOException;
+
+    /**
+     * Returns the unique name of the provider
+     * @return
```
rhtyd: Self explanatory

framework/ca/src/org/apache/cloudstack/framework/ca/CAProvider.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    String getProviderName();
+
+    /**
+     * Returns description about the CA provider plugin
+     * @return
```
rhtyd: Self explanatory

framework/ca/src/org/apache/cloudstack/framework/ca/CAService.java
```diff
@@ -0,0 +1,36 @@
...
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+
+public interface CAService {
+    /**
+     * Returns a SSLEngine to be used for handling client connections
+     * @param context
+     * @param remoteAddress
+     * @return
```
rhtyd: CloudStack classes are not consume-able as libraries -- the docs are only for ACS developers. I also have a dilemma whether to even have these docs as it's quite clear what the methods accept, returns in most cases. With this, I have added the docs to provide some interface definition but not wrote it pendantically to be consumed as a library.

framework/ca/src/org/apache/cloudstack/framework/ca/Certificate.java
```diff
@@ -0,0 +1,46 @@
...
+        this.clientCertificate = clientCertificate;
+        this.privateKey = privateKey;
+        this.caCertificates = caCertificates;
+    }
+
+    public X509Certificate getClientCertificate() {
+        return clientCertificate;
+    }
+
+    public PrivateKey getPrivateKey() {
```
rhtyd: Expected, we want to return private key is it's there.

plugins/ca/root-ca/src/org/apache/cloudstack/ca/provider/RootCAProvider.java
```diff
@@ -0,0 +1,572 @@
...
+    @Inject
+    private ConfigurationDao configDao;
+    @Inject
+    private CrlDao crlDao;
+
+    ////////////////////////////////////////////////////
+    /////////////// Root CA Settings ///////////////////
+    ////////////////////////////////////////////////////
+
+    private static ConfigKey<String> rootCAPrivateKey = new ConfigKey<>("Hidden", String.class,
```
rhtyd: Yes, all `Hidden` keys are encrypted.

plugins/ca/root-ca/src/org/apache/cloudstack/ca/provider/RootCAProvider.java
```diff
@@ -0,0 +1,572 @@
...
+    @Override
+    public boolean revokeCertificate(final BigInteger certSerial, final String certCn) {
+        return true;
+    }
+
+    ////////////////////////////////////////////////////////////
+    /////////////// Root CA Trust Management ///////////////////
+    ////////////////////////////////////////////////////////////
+
+    public static final class RootCACustomTrustManager implements X509TrustManager {
```
rhtyd: It needs to be restricted to the `RootCAProvider` class, however, I made this public to write unit tests easily. I'll see if we can refactor it out completely.

plugins/ca/root-ca/src/org/apache/cloudstack/ca/provider/RootCAProvider.java
```diff
@@ -0,0 +1,572 @@
...
+            }
+            LOG.debug(builder.toString());
+        }
+
+        @Override
+        public void checkClientTrusted(final X509Certificate[] certificates, final String s) throws CertificateException {
+            if (LOG.isDebugEnabled()) {
+                printCertificateChain(certificates, s);
+            }
+            if (!authStrictness) {
```
rhtyd: Yep :)

plugins/ca/root-ca/src/org/apache/cloudstack/ca/provider/RootCAProvider.java
```diff
@@ -0,0 +1,572 @@
...
+        }
+
+        @Override
+        public X509Certificate[] getAcceptedIssuers() {
+            if (!authStrictness) {
+                return null;
+            }
+            return new X509Certificate[]{caCertificate};
+        }
+    }
```
rhtyd: I'll move this to a separate file/class.

server/src/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
```diff
@@ -125,6 +137,73 @@ public boolean processTimeout(long agentId, long seq) {
         return false;
     }
 
+    private void setupAgentSecurity(final Connection sshConnection, final String agentIp, final String agentHostname) {
```
rhtyd: Hmm, I would avoid, this is do-able however.

server/src/org/apache/cloudstack/ca/CAManagerImpl.java
```diff
@@ -0,0 +1,427 @@
...
+        if (crl != null && crl.getCertSerial().equals(certSerial)) {
+            final CAProvider provider = getCAProvider(caProvider);
+            return provider.revokeCertificate(certSerial, certCn);
+        }
+        return false;
+    }
+
+    @Override
+    @ActionEvent(eventType = EventTypes.EVENT_CA_CERTIFICATE_PROVISION, eventDescription = "provisioning certificate for host", async = true)
+    public boolean provisionCertificate(final Host host, final Boolean reconnect, final String caProvider) {
```
rhtyd: Yes, this is part of the on-boarding strategy. One can write or move to a new provider, by calling this API with the provider name arg, and then switch the value from global setting. This way onboarding can be done, without restarting mgmt server even.

server/src/org/apache/cloudstack/ca/CAManagerImpl.java
```diff
@@ -0,0 +1,427 @@
...
+        private CAManager caManager;
+        private HostDao hostDao;
+
+        public CABackgroundTask(final CAManager caManager, final HostDao hostDao) {
+            this.caManager = caManager;
+            this.hostDao = hostDao;
+        }
+
+        @Override
+        protected void runInContext() {
```
rhtyd: I would avoid, not huge.

server/src/org/apache/cloudstack/ca/CAManagerImpl.java
```diff
@@ -0,0 +1,427 @@
...
+            try {
+                if (LOG.isTraceEnabled()) {
+                    LOG.trace("CA background task is running...");
+                }
+                final DateTime now = DateTime.now(DateTimeZone.UTC);
+                final Map<String, X509Certificate> certsMap = caManager.getActiveCertificatesMap();
+                for (final Iterator<Map.Entry<String, X509Certificate>> it = certsMap.entrySet().iterator(); it.hasNext(); ) {
```
rhtyd: Nope, the other subsystems may be removing keys as we're iterating -- so we need an explicitly iterator. Removing keys from a map while iterating using the syntax you've shared may produce a NPE.

server/src/org/apache/cloudstack/ca/CAManagerImpl.java
```diff
@@ -0,0 +1,427 @@
...
+        if (configuredCaProvider == null) {
+            LOG.error("Failed to find valid configured CA provider, please check!");
+            return false;
+        }
+        return true;
+    }
+
+    @Override
+    public boolean configure(final String name, final Map<String, Object> params) throws ConfigurationException {
+        backgroundPollManager.submitTask(new CABackgroundTask(this, hostDao));
```
rhtyd: configure is called before start, the bg pool mgr's executor is started at start() which uses a dynamic size based on the submitted tasks list/collection so things need to be submitted/configured in `configure()` of any manager/service impl. See, the background polling manager implementation you'll like what I've done there :)

services/secondary-storage/controller/src/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -1118,7 +1119,7 @@ public boolean finalizeVirtualMachineProfile(VirtualMachineProfile profile, Depl
 
         StringBuilder buf = profile.getBootArgsBuilder();
         buf.append(" template=domP type=secstorage");
-        buf.append(" host=").append(ApiServiceConfiguration.ManagementHostIPAdr.value());
+        buf.append(" host=").append(StringUtils.shuffleCSVList(ApiServiceConfiguration.ManagementHostIPAdr.value()));
```
rhtyd: Yes, we want to have some randomness in case of multiple-mgmt server ips (if configured) that's still better than a static/configured value.

setup/db/server-setup.sql
```diff
@@ -27,3 +27,6 @@ INSERT INTO `cloud`.`configuration` (category, instance, component, name, value,
 
 -- Enable dynamic RBAC by default for fresh deployments
 INSERT INTO `cloud`.`configuration` (category, instance, component, name, value) VALUES ('Advanced', 'DEFAULT', 'RoleService', 'dynamic.apichecker.enabled', 'true');
+
+-- Enable RootCA auth strictness for fresh deployments
+INSERT INTO `cloud`.`configuration` (category, instance, component, name, value) VALUES ('Advanced', 'DEFAULT', 'RootCAProvider', 'ca.plugin.root.auth.strictness', 'true');
```
rhtyd: The configkey value by default is false, we want `true`

utils/pom.xml
rhtyd: We want dependency explicitly stated, as done for bcprov-jdk15on. The root pom has it as well.

utils/src/main/java/org/apache/cloudstack/utils/security/KeyStoreUtils.java
rhtyd: Good comment, however, the `/` is added to add a separator in case someone forgets to add one in the command/exec. I can remove it, it will still work as these are only file names.

utils/src/main/java/org/apache/cloudstack/utils/security/KeyStoreUtils.java
rhtyd: -do-

rhtyd:
@blueorangutan package

rhtyd:

server/src/org/apache/cloudstack/ca/CAManagerImpl.java
```diff
@@ -0,0 +1,427 @@
...
+
+    private CAProvider getConfiguredCaProvider() {
+        if (configuredCaProvider == null && caProviderMap.containsKey(CAProviderPlugin.value())) {
+            configuredCaProvider = caProviderMap.get(CAProviderPlugin.value());
+        }
+        if (configuredCaProvider == null) {
+            throw new CloudRuntimeException("Failed to find default configured CA provider plugin");
+        }
+        return configuredCaProvider;
+    }
```
nvazquez: What about refactoring method a little bit to something like
````
if (configuredCaProvider != null) return configuredCaProvider
else if (caProviderMap.containsKey(..) && caProviderMap.get(..) != null) return caProviderMap.get(..);
else throw new CloudRuntimeException("..")
````

rhtyd: @nvazquez sure, however if the second `else if`, we want to set and then return, it would be the same kind of code again.

nvazquez: You are right, sorry I ommited set line, then makes no sense to refactor in my opinion

rhtyd: @nvazquez thanks, anyway I made a change

rhtyd:
Thanks @DaanHoogland @nvazquez for reviewing

rhtyd:
Thanks @borisstoyanov , I'll keep this open for now until we receive the xenserver smoke test results. I'll merge this on satisfactory results.

rhtyd:
@blueorangutan package

rhtyd:

mlsorensen:
@rhtyd it would still be possible for @wido to use the new secure system without adding hosts via the management server, correct? They could generate a CSR on the host and call issueCertificate, then install it during the process that creates agent.properties?

rhtyd:
@mlsorensen yes if @wido wants to use the new secure system he can (1) create a keystore at /etc/cloudstack/agent/cloud.jks with a 'cloud' alias and password at agent.properties file, and (2) create a CSR using the keystore, (3) use the CSR to issue a certificate from management server and save the certificate and ca-certs and import them into the keystore, (4) add/start the agent to a zone/cluster.
The utility scripts will be installed as part of `cloudstack-common` package and will be put in systemvm.iso for systemvms, and available at: `/usr/share/cloudstack-common/scripts/util/{keystore-cert-import, keystore-setup}`

I'll document the usage of these scripts, in admin docs or somewhere, briefly here's how these scripts work:

`keystore-setup` <the agent/db properties filepath> <the target keystore filepath> <the keystore password> <ks/cert validity in number of days> <csr filepath>
This script will save and output the csr, save the keystore passphrase in the properties file and also create a `cloud.jks.new` keystore file.

After the certificate (ca+client) are created, they are stored in the keystore file using `keystore-import-cert` script:
`keystore-cert-import` <properties filepath> <final keystore filename, give cloud.jks not cloud.jks.new> <application mode: ssh|agent, only in ssh mode the agent/cloud service is restarted, ex. for systemvms> <the certificate filepath> <the certificate content/base-64 with newlines replaced by `^` and spaces replaced by `~`> <the ca-cert filepath> <the ca-cert content, with char-replacements same as the cert content> <the private key file path, this is useful only when csr was not used/created> <the private key content, with same char-replacement scheme>

Running the import-script would save the certs in the keystore file `cloud.jks.new` and rename it to `cloud.jks`. The `.new` is used to act as a two-phase commit approach, so in case during provisioning (for example, a live system using provisionCertificate API) if anything goes bad, it won't affect an existing keystore. Lastly, in addition to the keystore file, the certificates and keys are also stored in usual x509/pem formats with `chmod 600` applied to them, at the agent conf directory for usage with other services.

mlsorensen:
Thanks. I realize the initial question was more along the lines of "will this break how we do things now?", but it's good to know that it can be incorporated into that mode of operation as well if people choose.

cloudmonger:
### ACS CI BVT Run
 **Sumarry:**
 Build Number 1149
 Hypervisor xenserver
 NetworkType Advanced
 Passed=107
 Failed=5
 Skipped=12

**Failed tests:**
* test_list_ids_parameter.py

 * ContextSuite context=TestListIdsParams>:setup Failing since 53 runs

* test_routers_network_ops.py

 * test_01_isolate_network_FW_PF_default_routes_egress_true Failing since 14 runs

 * test_02_isolate_network_FW_PF_default_routes_egress_false Failing since 141 runs

 * test_01_RVR_Network_FW_PF_SSH_default_routes_egress_true Failing since 136 runs

 * test_02_RVR_Network_FW_PF_SSH_default_routes_egress_false Failing since 136 runs


**Skipped tests:**
test_vm_nic_adapter_vmxnet3
test_01_verify_libvirt
test_04_verify_guest_lspci
test_static_role_account_acls
test_3d_gpu_support
test_deploy_vgpu_enabled_vm

**Passed test suits:**
test_portable_publicip.py
test_vm_snapshots.py
test_over_provisioning.py
test_global_settings.py
test_router_dnsservice.py
test_scale_vm.py
test_service_offerings.py
test_loadbalance.py
test_routers.py
test_reset_vm_on_reboot.py
test_network.py
test_router_dns.py
test_non_contigiousvlan.py
test_login.py
test_deploy_vm_iso.py
test_public_ip_range.py
test_multipleips_per_nic.py
test_metrics_api.py
test_regions.py
test_affinity_groups.py
test_network_acl.py
test_pvlan.py
test_volumes.py
test_nic.py
test_deploy_vm_root_resize.py
test_resource_detail.py
test_secondary_storage.py
test_vm_life_cycle.py
test_disk_offerings.py

rhtyd:
The XenServer env related failure was caused by centos7 based packging/dependencies issues where pyopessl.crypto may not have a `load_publickey` method available. I've put a fix in the marvin test to skip when pyopenssl library does not export such a method, with that all tests pass now:

=== TestName: test_issue_certificate_with_csr | Status : SUCCESS ===

=== TestName: test_issue_certificate_without_csr | Status : SUCCESS ===

=== TestName: test_list_ca_certificate | Status : SUCCESS ===

=== TestName: test_list_ca_providers | Status : SUCCESS ===

=== TestName: test_provision_certificate | Status : SUCCESS ===

=== TestName: test_revoke_certificate | Status : SUCCESS ===

With enough test results and code reviews, this is ready for merging. I'll wait for Travis to go green and then merge this feature.

PR #2253, opened 2017-08-30
[CLOUDSTACK-10061][Managed Storage] When starting a VM, make sure it has the correct volume access group

syed:
This can happen when you stop a VM in one cluster and start a VM in another cluster. When the VM starts in a new cluster, we don't add a new VAG and hence it fails to start. This PR ensures that we call `grantAccess` to the VM that gets started which will fix the access issue.

cc: @mike-tutkowski

### DIFF
--- a/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
@@ -1369,6 +1369,20 @@ public class VolumeOrchestrator extends ManagerBase implements VolumeOrchestrati
             if (task.type == VolumeTaskType.NOP) {
                 pool = (StoragePool)dataStoreMgr.getDataStore(task.pool.getId(), DataStoreRole.Primary);
                 vol = task.volume;
+                // For a zone-wide managed storage, it is possible that the VM can be started in another
+                // cluster. In that case make sure that the volume in in the right access group cluster.
+                if (pool.isManaged()) {
+                    long oldHostId = vm.getVirtualMachine().getLastHostId();
+                    long hostId = vm.getVirtualMachine().getHostId();
+                    if (oldHostId != hostId) {
+                        Host oldHost = _hostDao.findById(oldHostId);
+                        Host host = _hostDao.findById(hostId);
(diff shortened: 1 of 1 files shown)


engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
```diff
@@ -1369,6 +1369,13 @@ public void prepare(VirtualMachineProfile vm, DeployDestination dest) throws Sto
             if (task.type == VolumeTaskType.NOP) {
                 pool = (StoragePool)dataStoreMgr.getDataStore(task.pool.getId(), DataStoreRole.Primary);
                 vol = task.volume;
+                if (pool.isManaged()) {
```
mike-tutkowski: @syed I am wondering if we should instead call revokeAccess on the volume for the old cluster and grantAccess on the volume for the new cluster. That way the storage plug-in doesn't need to have any new, special logic around automatically removing the volume from an existing VAG when adding it to a new VAG. It also provides a level of symmetry in CloudStack's core code around grant/revokeAccess (the core code never gets into a grantAccess + grantAccess scenario - it's always grantAccess + revokeAccess + grantAccess + revokeAccess, etc.).

What do you think?

syed: I think you are right @mike-tutkowski It definitely simplifies things.

syed:
@mike-tutkowski I've changed the logic to remove any dependence on the storage driver.

GabrielBrascher:
@syed can you please create a JIRA ticket and append the JIRA ID with the PR title?

mike-tutkowski:
@syed This code LGTM. Thanks!

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

rhtyd:
LGTM, the other failures are not related to this PR and are known issues handled in separate PRs.

rhtyd:
@syed can we get a JIRA id for this, this is ready for merge.

syed:
@rhtyd @GabrielBrascher Added JIRA ID
@mike-tutkowski Thanks for the review!

rhtyd:
LGTM, test LGTM. Merging now.

PR #2254, opened 2017-08-31
CLOUDSTACK-10058: Error while opening the Settings tab in Secondary storage

nitin-maharana:
Issue Description:
=============
https://issues.apache.org/jira/browse/CLOUDSTACK-10058

Root Cause:
=========
Some global parameters contains NULL value, where the code doesn't handle NULL check.
So it fails with an exception. Hence nothing appears on the field(ERROR).

Solution:
=======
Added required NULL check.

Snapshots:
========
Before
before

After
after


rhtyd:
LGTM
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Additional review requested @karuturi @DaanHoogland @nvazquez @niteshsarda  @yvsubhash and others

nvazquez:
LGTM

PR #2255, opened 2017-08-31
CLOUDSTACK-10059: Dashboard ignores resource value that is less than 1%

nitin-maharana:
Description:
=========
https://issues.apache.org/jira/browse/CLOUDSTACK-10059

Root Cause:
=========
The API returns the percentage value as floating point number but In UI it is converted to floor value of that number.
Hence for values less than 1%, the value is being converted to 0% and it doesn't show the value in the UI.

Solution:
=======
Removed the conversion, now it represents as it is(Floating point value).

Snapshots:
========
Before
before

After
after


sateesh-chodapuneedi:
I believe this PR could be merged as this UI specific change, with Jenkins and Travis both are green.

rhtyd:
LGTM, merging now.

PR #2270, opened 2017-09-20
CLOUDSTACK-10085 : Upload volume from local fails when global config max.account.seconday.storage is set to -1

niteshsarda:
**ISSUE**

Upload volume from local fails when global config max.account.seconday.storage is set to -1

**STEPS TO REPRODUCE**

1. Download any volume on your local machine.
2. Change value of global configuration max.account.seconday.storage to -1.
3. Restart Management server.
4. Try to upload a volume from UI using "Upload from Local" option. Upload will fail.

**FIX**

Implementation of -1 option for unlimited secondary storage usage, which is absent in current code.

### DIFF
--- a/services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
+++ b/services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
@@ -35,9 +35,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-
 import javax.naming.ConfigurationException;
-
 import com.amazonaws.services.s3.model.S3ObjectSummary;
 import com.cloud.agent.api.Answer;
 import com.cloud.agent.api.CheckHealthAnswer;
@@ -76,6 +74,7 @@ import com.cloud.exception.InternalErrorException;
 import com.cloud.exception.InvalidParameterValueException;
 import com.cloud.host.Host;
 import com.cloud.host.Host.Type;
+import com.cloud.configuration.Resource;
 import com.cloud.hypervisor.Hypervisor.HypervisorType;
 import com.cloud.resource.ServerResourceBase;
 import com.cloud.storage.DataStoreRole;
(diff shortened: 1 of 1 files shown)


mrunalinikankariya:
Code LGTM

niteshsarda:
@GabrielBrascher : I have rebase this PR against latest master and all checks are passing as well. 
Can you please merge this PR ?

GabrielBrascher:
Thanks @niteshsarda but I am not the person you need now ;)

ustcweizhou:
LGTM
we have the same fix in production for more than one year.

services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
```diff
@@ -2809,7 +2808,7 @@ private synchronized void checkSecondaryStorageResourceLimit(TemplateOrVolumePos
         int accountDirSizeInGB = getSizeInGB(accountTemplateDirSize + accountSnapshotDirSize + accountVolumeDirSize);
         int defaultMaxAccountSecondaryStorageInGB = Integer.parseInt(cmd.getDefaultMaxAccountSecondaryStorage());
 
-        if ((accountDirSizeInGB + contentLengthInGB) > defaultMaxAccountSecondaryStorageInGB) {
+        if (defaultMaxAccountSecondaryStorageInGB != Resource.RESOURCE_UNLIMITED && (accountDirSizeInGB + contentLengthInGB) > defaultMaxAccountSecondaryStorageInGB) {
```
rafaelweingartner: @niteshsarda thanks for the fix.
To improve this PR to an even higher level, how do you feel about extracting the condition of the `if` to a method? 

Then, this method can receive test cases and Java doc explaining when it returns true or false. 
This can facilitate code maintenance in the future and also to reduce the bar for newcomers.

niteshsarda: @rafaelweingartner : In my opinion, we don not need a separate method for this code because both the conditions in "if " loop are pretty simple to understand.
Also, we are using Java doc very rarely in our code, so adding Java doc for this straight forward condition will not be much useful.

ustcweizhou: I agree with @niteshsarda

rafaelweingartner: No problem, but I disagree with this kind of thinking "we are using Java doc very rarely in our code, so adding Java doc for this straight forward condition will not be much useful".

This is what makes our code base terrible in same (most) places and in my opinion we should improve bit by bit. Code  (easy to maintain, easy to be extended by newcomers, and easy to read) is a culture that we have to create.

DaanHoogland: I do agree with @rafaelweingartner to some extend but I don't think we should stop this fix because of it. Also the extend is that I agree that extracting the if in a separate method but that we should always try to not need any comments (including javadoc)
I am merging this.

rafaelweingartner: @DaanHoogland agreed with you ;)

Sometimes people do not understand when I suggest improvements. In this case for example, my intention was not to stop the addition of the code, but rather to improve it, and to show this perspective to the contributor. If I am against a code being merged, I will for sure tell that directly.

Thanks for merging.

PR #2272, opened 2017-09-22
CLOUDSTACK-10012: ServerDaemon customization

marcaurele:
- add option to allow loading another webapp (cloudstack-ui for example)
- remove `jetty` in param keywords as it's better to stay agnostic


marcaurele:
cc @rhtyd

PR #2273, opened 2017-09-25
CLOUDSTACK-10090:createPortForwardingRule api call accepts 'halt' as …

mrunalinikankariya:
…Protocol which Stops VR

When we run the createPortForwardingRule API with input as Protocol as halt the PF rule is added however Halt is executed on VR. Hence the VR is stopped.

Following entry added to Firewall_Rules table and VirtualRouter went to halt(stopped)
mysql> select * from firewall_rules where id = 7 

*************************** 1. row ***************************
id: 7
uuid: XXXXXXXXXXXXXXXXXXXXXXXXXXX
ip_address_id: 13
start_port: 222
end_port: 222
state: Revoke
protocol: `halt`
purpose: PortForwarding
account_id: 2
domain_id: 1
network_id: 208
xid: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
created: 2017-09-04 04:48:16
icmp_code: NULL
icmp_type: NULL
related: NULL
type: User
vpc_id: NULL
traffic_type

### DIFF
--- a/server/src/com/cloud/network/element/VirtualRouterElement.java
+++ b/server/src/com/cloud/network/element/VirtualRouterElement.java
@@ -23,6 +23,8 @@ import java.util.Map;
 import java.util.Set;
 
 import javax.inject.Inject;
+
+import com.cloud.utils.net.NetUtils;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Qualifier;
 import com.cloud.network.router.NetworkHelper;
@@ -568,7 +570,10 @@ NetworkMigrationResponder, AggregatedCommandExecutor, RedundantResource, DnsServ
         capabilities.put(Service.SourceNat, sourceNatCapabilities);
 
         capabilities.put(Service.StaticNat, null);
-        capabilities.put(Service.PortForwarding, null);
+
+        final Map<Capability, String> portForwardingCapabilities = new HashMap<Capability, String>();
+        portForwardingCapabilities.put(Capability.SupportedProtocols, NetUtils.TCP_PROTO + "," + NetUtils.UDP_PROTO);
+        capabilities.put(Service.PortForwarding, portForwardingCapabilities);
 
         return capabilities;
     }


server/src/com/cloud/network/element/VirtualRouterElement.java
```diff
@@ -568,7 +570,10 @@ public static String getHAProxyStickinessCapability() {
         capabilities.put(Service.SourceNat, sourceNatCapabilities);
 
         capabilities.put(Service.StaticNat, null);
-        capabilities.put(Service.PortForwarding, null);
+
+        final Map<Capability, String> portForwardingCapabilities = new HashMap<Capability, String>();
+        portForwardingCapabilities.put(Capability.SupportedProtocols, NetUtils.TCP_PROTO + "," + NetUtils.UDP_PROTO);
```
remibergsma: Why UDP? HAProxy cannot do UDP loadbalancing.

jayapalu: @remibergsma 
The changes here are for PF not for LB. Did I get you correctly ?

jayapalu:
for PF protocol validation code changes LGTM

server/src/com/cloud/network/element/VirtualRouterElement.java
remibergsma: Never mind

niteshsarda:
Tested the fix and it seems to be working fine. **Test LGTM.**

**Before Applying fix :**

If createPortForwardingRule api is executed with Protocol as "Halt", then router goes into halt state.

**After Applying fix :**

If createPortForwardingRule api is executed with Protocol as "Halt", then it throws an error. Below is the screenshot : 

screenshot

mrunalinikankariya:
tag:This is Ready to Merge

mrunalinikankariya:
@rhtyd Could you merge this please

rhtyd:
@mrunalinikankariya can you explain where/how this method is used and if putting a hardcoded protocol list can have undesirable effects, for example what if not both `tcp,udp` don't need to be port-forwarded.

For merging, this needs to be tested first.
@blueorangutan package

rhtyd:
@blueorangutan test

mrunalinikankariya:
@rhtyd 
Here we are hardcoding the Port Forwarding capabilities which is used to validate the type of protocol while creating port forwarding rule. If the capabilities list is empty no validation is done and hence any value is accepted as the protocol type. With this change now only TCP and UDP protocol type are allowed for port forwarding rule creation. Hope this answers your question

rhtyd:
Thanks @mrunalinikankariya for the explanation, however, I'm wondering if this might break use-cases where you would want to port-forward non-tcp/non-udp traffic such as `esp`, `ah`. Instead, can we add list of keywords that should not be allowed to be added to the protocol option? /cc @PaulAngus @DaanHoogland @nvazquez @borisstoyanov @wido @swill and others

jayapalu:
@rhtyd We can add list of allowed/supported protocols in the port forwarding capabilities rest will be blocked. If esp, ah is used in PF these can be added as well.

rhtyd:
@jayapalu @mrunalinikankariya looks like there are several regression failures, this cannot be accepted as such, please refactor the logic to throw exception only for unsupported keywords/protocols

jayapalu:
@rhtyd  @mrunalinikankariya Test failures are not related these changes. If we look at the changes there are cpvm, ssvm, iso failures which are unrelated to it. There are test setup failures as well.

rhtyd:
Okay @jayapalu rekicking tests.
@blueorangutan test

rhtyd:
Tests LGTM, given the provided feedback and explainations I'll merge this now. Thanks @mrunalinikankariya and @jayapalu

PR #2274, opened 2017-09-26
CLOUDSTACK-10096 Can't reset integration.api.port and usage.sanity.ch…

SudharmaJain:
…eck.interval back to null

Fix for CLOUDSTACK-7931 enforces a valid integer value to be configured for integration.api.port and usage.sanity.check.interval. These global configs can't be reset back to null(default).

While trying to reset following errors is observed. 

image
image

But 0 and negative are accepted.

image
image

Add changes to handle 0 and negative values as disabled for api.integration.port.
For usage.sanity.check.interval 0 and negative values are already handled. So no code changes are required. Only updated config description.

### DIFF
--- a/server/src/com/cloud/api/ApiServer.java
+++ b/server/src/com/cloud/api/ApiServer.java
@@ -336,6 +336,7 @@ public class ApiServer extends ManagerBase implements HttpRequestHandler, ApiSer
             final ConfigurationVO apiPortConfig = values.get(0);
             if (apiPortConfig.getValue() != null) {
                 apiPort = Integer.parseInt(apiPortConfig.getValue());
+                apiPort = (apiPort <= 0) ? null : apiPort;
             }
         }
 
--- a/server/src/com/cloud/configuration/Config.java
+++ b/server/src/com/cloud/configuration/Config.java
@@ -562,7 +562,7 @@ public enum Config {
             "The interval (in milliseconds) when host stats are retrieved from agents.",
             null),
     HostRetry("Advanced", AgentManager.class, Integer.class, "host.retry", "2", "Number of times to retry hosts for creating a volume", null),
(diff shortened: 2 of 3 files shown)


nitin-maharana:
@SudharmaJain, Thanks for the fix. Code LGTM.
I would suggest instead of selecting (<= 0) as the disabled state, we should allow users to give null value and consider it as the disabled state. I think this happens with all the parameters which expect an integer value but initially, the value would be null. Once changed to an integer value, it doesn't allow to change it back to null.

SudharmaJain:
@nitin-maharana As per the code fix for CLOUDSTACK-7931, we are restricting valid Integer/Float.  These are only two configs of type Integer/Float which are default initialized with null. Rest of the configs with default as null are Strings, which can be reset.

niteshsarda:
Tested the fix and it seems to be working fine.
**Test LGTM.**

SudharmaJain:
tag:MergeReady

SudharmaJain:
@rhtyd Conflicts resolved.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

SudharmaJain:
tag:MergeReady

rhtyd:
@SudharmaJain can you fix the conflicts?

rhtyd:
@SudharmaJain ping?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

SudharmaJain:
Test failures are not related to my changes.

rhtyd:
@blueorangutan test

rhtyd:
LGTM and tests LGTM. Merging this based on code reviews and test results.

PR #2275, opened 2017-09-26
CLOUDSTACK-9975: Allow customizing system VM templates for SSVM and Console Proxy

GabrielBrascher:
## Description

Currently, it is possible to change the template used by virtual routers, but other system VMs do not have the same feature. The virtual router template is configured according to the respective global settings parameters: router.template.hyperv, router.template.kvm, router.template.lxc, router.template.xenserver, router.template.ovm, router.template.vmware.

This PR allows users to configure templates for storage system VMs (SSVMs) and console proxy system VMs with parameters similar with the virtual router template configuration:  ssvm.template.<virtualization-tool> and consoleproxy.template.<virtualization-tool>

If a parameter is null then we use the default system VM.

This feature was useful in a practical scenario where the console proxy system VM template was customized.

## Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
+++ b/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
@@ -32,6 +32,8 @@ import javax.naming.ConfigurationException;
 import org.apache.cloudstack.agent.lb.IndirectAgentLB;
 import org.apache.cloudstack.context.CallContext;
 import org.apache.cloudstack.engine.orchestration.service.NetworkOrchestrationService;
+import org.apache.cloudstack.framework.config.ConfigKey;
+import org.apache.cloudstack.framework.config.Configurable;
 import org.apache.cloudstack.framework.config.dao.ConfigurationDao;
 import org.apache.cloudstack.framework.security.keys.KeysManager;
 import org.apache.cloudstack.framework.security.keystore.KeystoreDao;
@@ -153,7 +155,7 @@ import com.google.gson.GsonBuilder;
 // Starting, HA, Migrating, Running state are all counted as "Open" for available capacity calculation
 // because sooner or later, it will be driven into Running state
 //
-public class ConsoleProxyManagerImpl extends ManagerBase implements ConsoleProxyManager, VirtualMachineGuru, SystemVmLoadScanHandler<Long>, ResourceStateAdapter {
+public class ConsoleProxyManagerImpl extends ManagerBase implements ConsoleProxyManager, VirtualMachineGuru, SystemVmLoadScanHandler<Long>, ResourceStateAdapter, Configurable {
     private static final Logger s_logger = Logger.getLogger(ConsoleProxyManagerImpl.class);
 
     private static final int DEFAULT_CAPACITY_SCAN_INTERVAL = 30000; // 30 seconds
@@ -249,6 +251,19 @@ public class ConsoleProxyManagerImpl extends ManagerBase implements ConsoleProxy
     private KeystoreDao _ksDao;
     @Inject
     private KeystoreManager _ksMgr;
+
+    private static final ConfigKey<String> ConsoleProxyTemplateXenServer = new ConfigKey<String>("Console Proxy", String.class, "consoleproxy.template.xenserver",
+            "SystemVM Template (XenServer)", "Name of the console proxy template on Xenserver.", true);
+    private static final ConfigKey<String> ConsoleProxyTemplateKVM = new ConfigKey<String>("Console Proxy", String.class, "consoleproxy.template.kvm", "SystemVM Template (KVM)",
+            "Name of the console proxy template on KVM.", true);
+    private static final ConfigKey<String> ConsoleProxyTemplateVMware = new ConfigKey<String>("Console Proxy", String.class, "consoleproxy.template.vmware",
+            "SystemVM Template (vSphere)", "Name of the console proxy template on VMware.", true);
+    private static final ConfigKey<String> ConsoleProxyTemplateHyperV = new ConfigKey<String>("Console Proxy", String.class, "consoleproxy.template.hyperv",
+            "SystemVM Template (HyperV)", "Name of the console proxy template on Hyperv.", true);
+    private static final ConfigKey<String> ConsoleProxyTemplateLxc = new ConfigKey<String>("Console Proxy", String.class, "consoleproxy.template.lxc", "SystemVM Template (LXC)",
+            "Name of the console proxy template on LXC.", true);
+    private static final ConfigKey<String> ConsoleProxyTemplateOvm3 = new ConfigKey<String>("Console Proxy", String.class, "consoleproxy.template.ovm3", "SystemVM Template (Ovm3)",
+            "Name of the console proxy template on Ovm3.", true);
 
     public class VmBasedAgentHook extends AgentHookBase {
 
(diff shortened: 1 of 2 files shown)


wido:
Very nice and interesting! It might even be that in the future we should leverage this even more. This SSVM template is now difficult to manage since it has to be able to handle 3 different scenarios.

rhtyd:
Interesting PR, I'll ping couple of colleagues who may have more operational insights -- @PaulAngus @DagSontebo and others.

rhtyd:
@blueorangutan package

rafaelweingartner:
@GabrielBrascher can you rebase and fix conflicts?

GabrielBrascher:
Sorry by the delay @rafaelweingartner, I will do it soon. Thanks!

rafaelweingartner:
@GabrielBrascher can you fix the conflicts please?

rhtyd:
This PR may be closed, the global setting names may be confusing but the `router.xxx` global settings apply for cpvm, ssvm too.

DaanHoogland:
functionally this is good. I definately would want a marvin test for this that at least proves the success case so we can track regressions if we merge it.

rhtyd:
@GabrielBrascher thanks for replying. When we do update new systemvmtemplate with every release, it does get used for all systemvms and not just only routers so I disagree what you share. We can test of course, see this upgrade path for example where only the new template's name is updated for the `router.xxx` global settings: https://github.com/apache/cloudstack/blob/master/engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41000to41100.java#L175

rhtyd:
I verified that changing systemvmtemplate by following the changes that the upgrade path does, indeed changes the systemvmtemplate for all vrs and systemvms (cpvm/ssvm). This may be closed @GabrielBrascher. If you'd prefer, please perform a test yourself, thanks.

GabrielBrascher:
Thanks, @rhtyd. I used that code in CloudStack 4.9.3.0 when it was requested to edit system VM template but without that code, it was updated only templates for virtual routers. After that, I created a PR for the master.

I will review all and perform a test here as well to make sure.

rhtyd:
@GabrielBrascher perhaps you can explore refactoring of the systemvmtemplate upgrade code we've as part of the upgrade path, as an service layer API that may also be used via registerTemplate to make it easier to register/use a new systemvmtemplate.

DennisKonrad:
Hi @GabrielBrascher, @rhtyd,

what's the status of this? The feature looks interesting. Would be nice to specify different templates.

Also I have the feeling you both misunderstood each other. Rhtyd talked about changing the SystemVM-template which leaves CS no other option as to deploy the new template when a systemvm needs to be deployed.

Gabriel talked about the possibility to specify another template to use as CP or SSVM i think. I think that are two different things.

GabrielBrascher:
That is indeed the main goal, @DennisKonrad. Thanks for pointing that!

DennisKonrad:
@rhtyd What do you think? Maybe you could clarify on what you disagree with in this request or where/why you want to see changes to this?

PaulAngus:
@DennisKonrad I'm told by our guys that the router global setting in fact DOES change the temple for the ssvm and cpvm. Which would unfortunately make this redundant. 
Some of the ShapeBlue guys are working on a discrete API to update/change the system VMs template at any time.

DennisKonrad:
@PaulAngus I think so too. But the feature here allows for _different templates_ instead of having the same template for router, ssvm and console proxy.

Feel free to correct me on this :)

GabrielBrascher:
@PaulAngus @rhtyd this PR allows to have different templates for Virtual Router (VR), Console Proxy VM (CPVM), and Storage System VM (SSVM); thus, a feature that we all know it is not yet implemented for CloudStack. I do not see any technical reason for blocking this PR.

Cases where this can be useful:

1. **VR** has template _A_, **SSVM** has template _B_, and **CPVM** has template _C_;
2. **VR** has template _A_, **SSVM** has template _A_, and **CPVM** has template _B_;
3. **VR** has template _A_, **SSVM** has template _B_, and **CPVM** has template _A_; and
4. **VR** has template _A_, **SSVM** has template _B_, and **CPVM** has template _B_;

I created this PR at 2017 because it was a **critical** feature for a data center in Brazil. For sure this would help multiple CloudStack's users, @DennisKonrad is one of them.

Additionally, I experienced (more than once) a case where it was not possible to configure all system VMs with the same template (**VR** with template _A_, **SSVM** template _A_, **CP** template _A_). I will explain myself.

The method [findSystemVMReadyTemplate](https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/engine/schema/src/main/java/com/cloud/storage/dao/VMTemplateDaoImpl.java#L857) is called during the creation of [SSVM](https://github.com/apache/cloudstack/blob/1fa4f10a9ea65f156464da1671a4dbf62cc88e06/services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java#L645) and [CPVM](https://github.com/apache/cloudstack/blob/1fa4f10a9ea65f156464da1671a4dbf62cc88e06/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java#L645). 

```
    @Override
    public VMTemplateVO findSystemVMReadyTemplate(long zoneId, HypervisorType hypervisorType) {
        SearchCriteria<VMTemplateVO> sc = readySystemTemplateSearch.create();
        sc.setParameters("templateType", Storage.TemplateType.SYSTEM);
        sc.setParameters("state", VirtualMachineTemplate.State.Active);
        sc.setJoinParameters("tmplHyper", "type", Host.Type.Routing);
        sc.setJoinParameters("tmplHyper", "zoneId", zoneId);
        sc.setJoinParameters("vmTemplateJoinTemplateStoreRef", "downloadState", VMTemplateStorageResourceAssoc.Status.DOWNLOADED);

        // order by descending order of id
        List<VMTemplateVO> tmplts = listBy(sc, new Filter(VMTemplateVO.class, "id", false, null, null));

        if (tmplts.size() > 0) {
            if (hypervisorType == HypervisorType.Any) {
                return tmplts.get(0);
            }
            for (VMTemplateVO tmplt : tmplts) {
                if (tmplt.getHypervisorType() == hypervisorType) {
                    return tmplt;
                }
            }

        }
        return null;
    }
```

Thus, SSVMs and CPs templates are selected by descending order of id from table vm_template where the template type is of 'SYSTEM', the state is as 'Active', type of 'Routing', zone id as zoneId, has been downloaded and the hypervisor of the hypervisorType. In some cases more than one templates fit the select (multiple templates of 'SYSTEM', 'Active', 'Routing', same zone Id, same hypervisor), thus it returns the template that has the bigger Id, not the one that is configured for the router. Furthermore, if a root admin registers a new template for the system VM this template will not be owned by the system (type is of 'ROUTING', not 'SYSTEM'). Only during upgrades, the templates that are marked as 'ROUTING' are converted to 'SYSTEM'.

Note that for the [VR](https://github.com/apache/cloudstack/blob/1fa4f10a9ea65f156464da1671a4dbf62cc88e06/server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java#L474) the template is selected as follows. This **DOES NOT** happen for SSVM and CPVM:
```
final String templateName = retrieveTemplateName(hType, routerDeploymentDefinition.getDest().getDataCenter().getId());
final VMTemplateVO template = _templateDao.findRoutingTemplate(hType, templateName);
```

In conclusion, I am not seeing any technical reason to block this PR. What do you think guys? @DaanHoogland @dhlaluku @nvazquez @dagsonstebo @wido

PaulAngus:
thanks @DennisKonrad OK, now that I understand why you're doing this, I can see what you're trying to achieve.
IMO what you are trying to achieve is great, but there is a lot _around_ that the feature that would need to take into account before it would be 'safe' for users who are not so familiar with your design intent.

for instance (off the top of my head):
Now that there would potentially be no 'main/default' template, how do we define which system VMs & VRs need updating, when a new system vm template is 'required' for a new release. we check one now, what would we check in future?
Would we potentially need a 'requires update' state for cpvm and ssvm?
If you you've already thought of of this (and maybe more), awesome! then could you maybe share your FS to allay my paranoia .

GabrielBrascher:
That is a good point @PaulAngus.

What I have in mind is to keep system VM upgrade as it is on current documentation, adding just one point regarding the fact that the user should set the upgraded template for all system VMs [1], or update his customized templates based on the new template.

[1] VR, SSVMs and CPVMs templates configured with the upgraded template; or set SSVM and CPVM as null and the VR template with the upgraded template.

DennisKonrad:
@PaulAngus I agree with @GabrielBrascher that the default way of upgrading should stay the same. If no 'personalized' templates are set, the vr router template should be used.

The "required update" idea seems more of an enhancement to me that would be great even regardles of this pr here ;) If a user does not upgrade his svm's according to the documentation that might not be a problem (if lucky).
But a friendy reminder (in the UI) would be nice.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@GabrielBrascher can you triage this yourself?
Is it still applicable?
Do we want it in 4.15?

GabrielBrascher:
@DaanHoogland thanks for the ping. I think that it is more a matter of working with the systemvm upgrade process. 

Considering that it was not a top priority and 4.13.1/4.14.0 were on tigh schedule I did not invevst much time on this one. I will take another look on it as soon as 4.14 is released and we faced the python3 migration :)

PR #2288, opened 2017-10-10
CLOUDSTACK-10107: For VMware VMs add devices without unit number

rhtyd:
When VMs are deployed or nics are plugged, using a static unit number
may cause device configuration errors. This fixes a previous limitation
that more than 7 nics/networks could not be added to a VM.

Pinging for review @sureshanaparti @resmo @GabrielBrascher @PaulAngus @nvazquez @DaanHoogland @borisstoyanov and others

Per the API docs, `unitNumber` need not be set: https://www.vmware.com/support/developer/converter-sdk/conv55_apireference/vim.vm.device.VirtualDevice.html

@blueorangutan package

### DIFF
--- a/vmware-base/src/com/cloud/hypervisor/vmware/util/VmwareHelper.java
+++ b/vmware-base/src/com/cloud/hypervisor/vmware/util/VmwareHelper.java
@@ -89,7 +89,7 @@ public class VmwareHelper {
     }
 
     public static VirtualDevice prepareNicOpaque(VirtualMachineMO vmMo, VirtualEthernetCardType deviceType, String portGroupName,
-            String macAddress, int deviceNumber, int contextNumber, boolean conntected, boolean connectOnStart) throws Exception {
+            String macAddress, int contextNumber, boolean connected, boolean connectOnStart) throws Exception {
 
         assert(vmMo.getRunningHost().hasOpaqueNSXNetwork());
 
@@ -123,18 +123,17 @@ public class VmwareHelper {
 
         VirtualDeviceConnectInfo connectInfo = new VirtualDeviceConnectInfo();
         connectInfo.setAllowGuestControl(true);
-        connectInfo.setConnected(conntected);
+        connectInfo.setConnected(connected);
         connectInfo.setStartConnected(connectOnStart);
         nic.setAddressType("Manual");
         nic.setConnectable(connectInfo);
         nic.setMacAddress(macAddress);
-        nic.setUnitNumber(deviceNumber);
         nic.setKey(-contextNumber);
         return nic;
     }
 
     public static VirtualDevice prepareNicDevice(VirtualMachineMO vmMo, ManagedObjectReference morNetwork, VirtualEthernetCardType deviceType, String portGroupName,
-            String macAddress, int deviceNumber, int contextNumber, boolean conntected, boolean connectOnStart) throws Exception {
+            String macAddress, int contextNumber, boolean connected, boolean connectOnStart) throws Exception {
 
         VirtualEthernetCard nic;
         switch (deviceType) {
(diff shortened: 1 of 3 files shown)


rhtyd:

rhtyd:
@blueorangutan package

DaanHoogland:
LGTM based on the code

rhtyd:

resmo:
LGTM

vmware-base/src/com/cloud/hypervisor/vmware/util/VmwareHelper.java
```diff
@@ -210,16 +208,13 @@ public static VirtualDevice prepareDvNicDevice(VirtualMachineMO vmMo, ManagedObj
...
         connectInfo.setAllowGuestControl(true);
-        connectInfo.setConnected(conntected);
+        connectInfo.setConnected(connected);
         connectInfo.setStartConnected(connectOnStart);
         nic.setAddressType("Manual");
         nic.setConnectable(connectInfo);
         nic.setMacAddress(macAddress);
-
-        nic.setUnitNumber(deviceNumber);
         nic.setKey(-contextNumber);
```
rafaelweingartner: @rhtyd I know this is not your code, but there is something that called my attention (maybe you know why this is the way it is). At first I thought it was some dusty on my laptop screen... When the method `nic.setKey` is called, the code does `-contextNumber`, converting the positive integer to a negative one. I looked at the docs and the key is a unique ID of the NIC, but I reading the docs I did not see any reasons to use negative value as IDs.

rafaelweingartner:
@rhtyd may I ask you something? I was looking at the code, and I see that you removed a variable called `deviceNumber` that did not seem to be used at the method `prepareDvNicDevice`. However, looking at the changes I did not understand how they modify the limit of NICs to be attached into a VM when using VMware hypervisor. Is it related to that `setKey(30)` that you removed there?

BTW: long time I am not able to review or work much with you guys ;)

rhtyd:
@blueorangutan package

rhtyd:
Looks like some trillian env issue, I'll rekick tests shortly.

vmware-base/src/com/cloud/hypervisor/vmware/util/VmwareHelper.java
rhtyd: @rafaelweingartner when I was trying to fix the bug, I thought too why we're using negative values however, I could not find a reason. For backward compatibilities sake and to reduce our test scopes, I chose to keep it that way. IMO it's not a problem as long as the key is unique, being negative/positive is not a problem.

rhtyd:
@rafaelweingartner thanks, good to see you reviewing. The issue is that during starting of a VM, we used hard coded nic unit numbers that caused the backend (vcenter side) to fail sometimes (well most the times, when number of nics > 7). I looked at the docs, and the unit number is not a mandatory attribute to use when adding new (nic) devices, which is the fix for the bug. In the specific method, we're still passing `deviceNumber` to be used as key when plugging in a dvswitch `nic` device, I removed the `setKey(30)` as `setKey` is again called, its usage is redundant and unnecessary.

rhtyd:
@blueorangutan package

rhtyd:

vmware-base/src/com/cloud/hypervisor/vmware/util/VmwareHelper.java
rafaelweingartner: I agree with you, I only asked to see if there were some other hidden reasons.
Thanks for the explanations.

borisstoyanov:
Newly added test is passing
```
Test to add and update added nic to a virtual machine ... === TestName: test_01_nic | Status : SUCCESS ===
ok
Test to adding multiple nics to a VMware VM and restarting VM ... === TestName: test_03_nic_multiple_vmware | Status : SUCCESS ===
ok

Ran 2 tests in 1372.192s

OK

```

PR #2290, opened 2017-10-10
WIP: add track id to async api calls

DaanHoogland:
## Description

API calls should be logging a more appropriate tracking id. It already exists but seems not be used consistently enough.

## Types of changes

### DIFF
--- a/api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java
@@ -333,6 +333,7 @@ public class DeployVMCmd extends BaseAsyncCreateCustomIdCmd implements SecurityG
                     } catch (NumberFormatException e) {
                         throw new InvalidParameterValueException("Unable to translate and find entity with networkId: " + ips.get("networkid"));
                     }
+                    // so no network found but a valid number as network id; are we assuming exceptions thrown or lucky guessing?
                 }
                 String requestedIp = ips.get("ip");
                 String requestedIpv6 = ips.get("ipv6");
--- a/api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java
@@ -314,7 +314,8 @@ public class DeployVMCmd extends BaseAsyncCreateCustomIdCmd implements SecurityG
 
     public Map<Long, IpAddresses> getIpToNetworkMap() {
         if ((networkIds != null || ipAddress != null || getIp6Address() != null) && ipToNetworkList != null) {
-            throw new InvalidParameterValueException("NetworkIds and ipAddress can't be specified along with ipToNetworkMap parameter");
+            throw new InvalidParameterValueException(getInjectedJobId(),
+                    "NetworkIds and ipAddress can't be specified along with ipToNetworkMap parameter");
         }
         LinkedHashMap<Long, IpAddresses> ipToNetworkMap = null;
         if (ipToNetworkList != null && !ipToNetworkList.isEmpty()) {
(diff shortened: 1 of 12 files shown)


DaanHoogland:
@blueorangutan package

api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java
```diff
@@ -333,6 +333,7 @@ public boolean getStartVm() {
                     } catch (NumberFormatException e) {
                         throw new InvalidParameterValueException("Unable to translate and find entity with networkId: " + ips.get("networkid"));
                     }
+                    // so no network found but a valid number as network id; are we assuming exceptions thrown or lucky guessing?
```
rafaelweingartner: @DaanHoogland so you are the one that makes the enlightened comments all around ACS code base?! ahahaha

If the network ID is invalid (it was not found), it means that something else further down the deployment process will break, right? If this is the case, I think we could stop the processing here and throw a runtime exception.

DaanHoogland: I don't know yet, maybe an uncaught CloudRuntimeException is thrown somewhere else, but you are right, we can throw it here.

server/src/com/cloud/api/ApiServer.java
```diff
@@ -540,7 +540,7 @@ public String handleRequest(final Map params, final String responseType, final S
                         throw new CloudRuntimeException("No APICommand annotation found for class " + cmdClass.getCanonicalName());
                     }
 
-                    BaseCmd cmdObj = (BaseCmd)cmdClass.newInstance();
+                    BaseCmd cmdObj = (BaseCmd)cmdClass.getConstructor(cmdClass).newInstance();
```
rhtyd: @DaanHoogland what's the difference in doing `newInstance` directly, than getting the constructor first?

server/src/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4600,14 +4602,18 @@ public UserVm createVirtualMachine(DeployVMCmd cmd) throws InsufficientCapacityE
 
         DataCenter zone = _entityMgr.findById(DataCenter.class, zoneId);
         if (zone == null) {
-            throw new InvalidParameterValueException("Unable to find zone by id=" + zoneId);
+            CloudRuntimeException ex =  new InvalidParameterValueException("Unable to find zone by id=" + zoneId);
```
rhtyd: Does it make sense to instead, get/set the context/jobid in the exception constructor if CallContext is available to it?

api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java
rhtyd: Let's remove comments, unless they're around newly added code or explain existing code?

server/src/com/cloud/api/ApiServer.java
DaanHoogland: it is deprecated in version 9.

api/src/org/apache/cloudstack/api/command/user/vm/DeployVMCmd.java
DaanHoogland: sure, this comment should not add value any more once we decide how to handle this case.
At this point we have not found a network but are having a networkId. I think this should be unreachable.
I will refactor untill the code makes sense to me.

server/src/com/cloud/vm/UserVmManagerImpl.java
DaanHoogland: thanks, yes that's why I created the CloudRuntimeException(String callId, String message) but i haven't created a InvalidParameterValueException(String callId, String message) yet.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

server/src/com/cloud/api/ApiServer.java
DaanHoogland: ok, stupifd mistake here; i added the class self to the list of parameters :(

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

PR #2293, opened 2017-10-11
CLOUDSTACK-10047: DVSwitch fixes and improvements

rhtyd:
This adds a minor feature to accepts security policies while creating network offering. Changes:
- Deployed network will have security policies from the network offering
  applied on the port group (in vmware environment)
- Global settings as fallback when security policies are not defined for a network
  offering
- Default promiscuous mode security policy set to REJECT as it's the default
  for standard/default vswitch

This also allows admins to define a network with vlan range such as vlan://200-400
and use the range to configure vlan-trunking with the range for a portgroup
in dvswitch.

VLAN overlap checks are performed for:
- isolated network against existing shared and isolated networks
- dedicated vlan ranges for the physical/public network for the zone
- shared network against existing isolated network

Notes:
- No vlan-range overlap checks are performed when creating shared networks
- Multiple vlan id/ranges should include the vlan:// scheme prefix


rhtyd:
Notes:
- Portgroup creation and update is triggered primarily by StartCommand for a network (when a VM is started, here VM could be either a guest VM or a VR, therefore VM start and VR start triggered by network deployment or network restart with cleanup=true).
- While processing StartCommand, the subsystem reconfigures the nics/portgroups/disks etc which is not possible when simply doing a VM reboot/restart from CloudStack, which is why portgroups changes are not applied/propagated when doing VM reboot via CloudStack which send a Vm reboot to vmware.
- The global settings will apply for (cloud.public, cloud.guest and cloud.private) portgroups, while if network offering has the security policies defined the defined policies will be applied to cloud.guest (i.e. the guest/vm network specific portgroup) otherwise that will too use the global settings
- The feature is primarily intended for new networks and/or network-offerings, and if old/existing network offerings are used then global settings will be used. 
- The security policies/behaviour is same as vswitches, i.e. promiscuous mode set to false, and others set to true. By default, dvswitch would previously set promiscuous mode to true which would be both an overhead and security issue.

rhtyd:
Pinging for review @PaulAngus @nvazquez @DaanHoogland @borisstoyanov @DagSonstebo @resmo @rafaelweingartner @sureshanaparti and others
@blueorangutan package

rhtyd:
Thanks @resmo fixed :)

rhtyd:

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
```diff
@@ -138,7 +143,24 @@ protected VirtualMachineTO toVirtualMachineTO(VirtualMachineProfile vmProfile) {
             if(vm.getType() == VirtualMachine.Type.NetScalerVm) {
                 nicProfile.setBroadcastType(BroadcastDomainType.Native);
             }
-            nics[i++] = toNicTO(nicProfile);
+            NicTO nicTo = toNicTO(nicProfile);
+            final NetworkVO network = _networkDao.findByUuid(nicTo.getNetworkUuid());
+            if (network != null) {
+                final Map<NetworkOffering.Detail, String> details = networkOfferingDetailsDao.getNtwkOffDetails(network.getNetworkOfferingId());
+                if (details != null) {
```
rafaelweingartner: What about reducing the cyclomatic complexicity here?
it is a matter of inverting the conditional. Instead of `if (true){doSomething}`, we can do `if(!false){continue} doSomething`
This would enable to remove one if inception

utils/src/main/java/com/cloud/utils/UriUtils.java
```diff
@@ -391,4 +391,57 @@ public static InputStream getInputStreamFromUrl(String url, String user, String
...
     }
+
+    /**
+     * Expands a given vlan URI to a list of vlan IDs
+     * @param vlanAuthority the URI part without the vlan:// scheme
+     * @return returns list of vlan integer ids
+     */
+    public static List<Integer> expandVlanUri(final String vlanAuthority) {
+        final List<Integer> expandedVlans = new ArrayList<>();
+        if (vlanAuthority == null || vlanAuthority.isEmpty()) {
```
rafaelweingartner: what about `StringUtils.isBlank` here?

utils/src/main/java/com/cloud/utils/UriUtils.java
```diff
@@ -391,4 +391,57 @@ public static InputStream getInputStreamFromUrl(String url, String user, String
...
+            return expandedVlans;
+        }
+        for (final String vlanPart: vlanAuthority.split(",")) {
+            if (vlanPart == null || vlanPart.isEmpty()) {
```
rafaelweingartner: what about `StringUtils.isBlank` here?

utils/src/main/java/com/cloud/utils/UriUtils.java
```diff
@@ -391,4 +391,57 @@ public static InputStream getInputStreamFromUrl(String url, String user, String
...
+                Integer start = NumbersUtil.parseInt(range[0], -1);
+                Integer end = NumbersUtil.parseInt(range[1], -1);
+                if (start <= end && end > -1 && start > -1) {
+                    while (start <= end) {
+                        expandedVlans.add(start++);
+                    }
+                }
+            } else {
+                final Integer value = NumbersUtil.parseInt(range[0], -1);
+                if (value > -1) {
```
rafaelweingartner: what about a debug message here to say why we are rejecting this value, and displaying the value of range[0], in case `range[0]` is not a number and the method returns -1

utils/src/main/java/com/cloud/utils/UriUtils.java
```diff
@@ -391,4 +391,57 @@ public static InputStream getInputStreamFromUrl(String url, String user, String
...
+    /**
+     * Checks if given vlan URI authorities overlap
+     * @param vlanRange1
+     * @param vlanRange2
+     * @return true if they overlap
+     */
+    public static boolean checkVlanUriOverlap(final String vlanRange1, final String vlanRange2) {
+        final List<Integer> vlans1 = expandVlanUri(vlanRange1);
+        final List<Integer> vlans2 = expandVlanUri(vlanRange2);
+        if (vlans1 == null || vlans2 == null) {
```
rafaelweingartner: if one of them is null, the result is `true`? This means that they overlap?

engine/schema/src/com/cloud/dc/dao/DataCenterVnetDaoImpl.java
```diff
@@ -100,29 +101,42 @@ public void lockRange(long dcId, long physicalNetworkId, Integer start, Integer
...
     public int countZoneVlans(long dcId, boolean onlyCountAllocated) {
         SearchCriteria<Integer> sc = onlyCountAllocated ? countAllocatedZoneVlans.create() : countZoneVlans.create();
         sc.setParameters("dc", dcId);
         return customSearch(sc, null).get(0);
     }
 
+    private List<DataCenterVnetVO> findOverlappingVnets(final long dcId, final Long physicalNetworkId, final String vnet) {
+        final List<Integer> searchVnets = UriUtils.expandVlanUri(vnet);
+        final List<DataCenterVnetVO> overlappingVnets = new ArrayList<>();
+        if (searchVnets != null && searchVnets.size() > 0) {
```
rafaelweingartner: what about inverting this conditional?
`if( searchVnets == null || searchVnets.size() == 0){return overlappingVnets;}`
This helps to reduce the number of IFs inside of IFs

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
rhtyd: We're protecting against NPEs and setting details to a `NicTO` object when those details are available. We cannot simply `continue` as the value is being set at lines 161, 163.

utils/src/main/java/com/cloud/utils/UriUtils.java
rhtyd: Fixed with `Strings.isNullOrEmpty`, thanks

utils/src/main/java/com/cloud/utils/UriUtils.java
rhtyd: Fixed with `Strings.isNullOrEmpty`, thanks

utils/src/main/java/com/cloud/utils/UriUtils.java
rhtyd: This is used for vlan range checks etc, vlans are always > -1.

utils/src/main/java/com/cloud/utils/UriUtils.java
rhtyd: We're checking if two vlan ranges (comma separated ranges or values, such as `100-200,300` or `20-30` etc) overlap, if any of them when expanded (i.e. `1-3` expands to `1,2,3`) is null, i.e. there is no overlap.

engine/schema/src/com/cloud/dc/dao/DataCenterVnetDaoImpl.java
rhtyd: Fixed, thanks.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
rafaelweingartner: You are right about the details, but I think I clicked on the wrong line (sorry for the mistake). I was talking about the ` if (network != null) ` check. if ` if (network == null)` you do not do anything, then it is the same as using a continue. I mean, you do a `nics[i++] = nicTo;`, but that can go inside the condition as well. Duplicated lines are not that great, so let's see what else could be done...

We can do something else, we could extract lines 149-160 to a method; this would improve the readability of the code and enable unit tests and Java docs.

utils/src/main/java/com/cloud/utils/UriUtils.java
rafaelweingartner: I did not understand the explanations :(

Anyways, looking at the code of `expandVlanUri`, I did not see a way for it to return null. Worst case scenario it returns an empty list. Do we need this check?

rhtyd: Yes, we need this check.

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
rhtyd: The block converts `NicProfile` to `NicTO`, the idea of the changes is to override certain security settings on dvswitch's portgroups if they already don't have the settings, using global settings.

utils/src/main/java/com/cloud/utils/UriUtils.java
rafaelweingartner: Do you mind explaining why if the method `expandVlanUri` returns an empty list in the worst case?

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
rafaelweingartner: Thanks for both the explanations and for the effort!

BTW: your explanation is great to enrich our code base. This explanation would be awesome in a Javadoc, so people in the future can understand why we are doing this without needing to deeply inspect the code.

utils/src/main/java/com/cloud/utils/UriUtils.java
rhtyd: In general, this case won't be hit but the implementation was done to be defensive of all sorts of inputs, including nulls. If vlanRange provided to `expandVlanUri` had issues such as non-integers or negative values, then you'll get empty list. Again, this was defensive programming.

rafaelweingartner: I understand that, but I think we should not over code or engineer. If we worry about the method `expandVlanUri` returning null, we can do something else to catch this.

What about a test case for `expandVlanUri ` that fails if it returns null? Then, we can remove this check. On thing is to be defensive when a `null` case can happen, the other is to code expecting someone to make a mistake in the future (for that it is better to write unit test cases).

utils/src/test/java/com/cloud/utils/UriUtilsTest.java
```diff
@@ -57,4 +60,45 @@ public void getUpdateUri() {
...
+        Assert.assertTrue(UriUtils.checkVlanUriOverlap("10-30", "25-35"));
+    }
+
+    @Test
+    public void testVlanUriNoOverlap() {
+        Assert.assertFalse(UriUtils.checkVlanUriOverlap("10-30,45,50,12,31", null));
+        Assert.assertFalse(UriUtils.checkVlanUriOverlap("10-30,45,50,12,31", ""));
+        Assert.assertFalse(UriUtils.checkVlanUriOverlap("10-30,45,50,12,31", "32"));
+        Assert.assertFalse(UriUtils.checkVlanUriOverlap("10,22,111", "12"));
+        Assert.assertFalse(UriUtils.checkVlanUriOverlap("100-200", "30-40,50,201-250"));
```
rhtyd: @rafaelweingartner unit tests for respective methods in questions are in this file, please see all the above lines.

utils/src/main/java/com/cloud/utils/UriUtils.java
rhtyd: I think we're over-discussing, not over-engineering :) I'll ping you on respective unit tests.

rafaelweingartner: Ah that as well ;)

I am sorry to bother, but if I do not understand something I keep asking until I can move along.

utils/src/test/java/com/cloud/utils/UriUtilsTest.java
rafaelweingartner: I had already seen these tests, I did not say anything because in my option they are properly written ;)

So, about that `null` case, in my option it is already covered, if for some reason someone alters the method to return null, your test cases will catch it (this is great!). That is why I was saying you do not need those null checks.

rhtyd: Sure, sometimes we do want to over-engineer™ :)

rafaelweingartner: Ahaha, sometimes we do, but we should not. 

I totally understand when we do, I also exaggerate sometimes, that is why I find it is great a review process to get another set of eyes to look at the problem and the code.

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
```diff
@@ -138,7 +143,24 @@ protected VirtualMachineTO toVirtualMachineTO(VirtualMachineProfile vmProfile) {
...
+                    if (!details.containsKey(NetworkOffering.Detail.PromiscuousMode)) {
```
fmaximus: You might also use putIfAbsent here instead, which was added in Java 8

utils/src/main/java/com/cloud/utils/UriUtils.java
fmaximus: You could also use !Collections.disjoint(vlans1, vlans2)

api/src/org/apache/cloudstack/api/command/admin/network/CreateNetworkCmdByAdmin.java
```diff
@@ -48,6 +51,13 @@ public String getVlan() {
         return vlan;
     }
 
+    public Boolean getBypassVlanOverlapCheck() {
```
sgoeminn: Is there a reason why we do return a Boolean instead of boolean? (this function can never return null)

engine/schema/src/com/cloud/dc/dao/DataCenterVnetDaoImpl.java
```diff
@@ -100,29 +103,43 @@ public void lockRange(long dcId, long physicalNetworkId, Integer start, Integer
...
         sc.setParameters("dc", dcId);
         return customSearch(sc, null).get(0);
     }
 
-    @Override
-    public List<DataCenterVnetVO> findVnet(long dcId, long physicalNetworkId, String vnet) {
+    private List<DataCenterVnetVO> findOverlappingVnets(final long dcId, final Long physicalNetworkId, final String vnet) {
+        final List<Integer> searchVnets = UriUtils.expandVlanUri(vnet);
+        final List<DataCenterVnetVO> overlappingVnets = new ArrayList<>();
+        if (searchVnets == null || searchVnets.size() == 0) {
```
sgoeminn: You could use searchVnets.isEmpty() instead of size()==0?

vmware-base/src/com/cloud/hypervisor/vmware/mo/HypervisorHostHelper.java
```diff
@@ -501,12 +510,18 @@ public static void updatePortProfile(VmwareContext context, String ethPortProfil
             // No doubt about this, depending on vid=null to avoid lots of code below
             vid = null;
         } else {
+            if (vlanId != null) {
+                vlanId = vlanId.replace("vlan://", "");
+            }
             networkName = composeCloudNetworkName(namePrefix, vlanId, secondaryvlanId, networkRateMbps, physicalNetwork);
 
-            if (vlanId != null && !UNTAGGED_VLAN_NAME.equalsIgnoreCase(vlanId)) {
+            if (vlanId != null && !UNTAGGED_VLAN_NAME.equalsIgnoreCase(vlanId) && !vlanId.contains(",") && !vlanId.contains("-")) {
```
sgoeminn: && !vlanId.contains(",") && !vlanId.contains("-") could be replaced by !StringUtils.containsAny(vlanId, ",-")?

vmware-base/src/com/cloud/hypervisor/vmware/mo/HypervisorHostHelper.java
```diff
@@ -501,12 +510,18 @@ public static void updatePortProfile(VmwareContext context, String ethPortProfil
...
                 createGCTag = true;
                 vid = Integer.parseInt(vlanId);
             }
+            if (vlanId != null && (vlanId.contains(",") || vlanId.contains("-"))) {
```
sgoeminn: could be replaced by  StringUtils.containsAny(vlanId, ",-")? Maybe also extract it to a separate boolean because it's checked multiple times?

utils/src/main/java/com/cloud/utils/UriUtils.java
rhtyd: @fmaximus fixed.

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
rhtyd: Fixed. Thanks, good tip.

api/src/org/apache/cloudstack/api/command/admin/network/CreateNetworkCmdByAdmin.java
rhtyd: Yes, because cmd classes set API args to fields using reflections etc. It's easier to not use native type `boolean`. When arg is not sent part of the API request, it would be set to null; setting null to `boolean` will throw an exception.

engine/schema/src/com/cloud/dc/dao/DataCenterVnetDaoImpl.java
rhtyd: Thanks, fixed.

vmware-base/src/com/cloud/hypervisor/vmware/mo/HypervisorHostHelper.java
rhtyd: Fixed, thanks.

vmware-base/src/com/cloud/hypervisor/vmware/mo/HypervisorHostHelper.java
rhtyd: Fixed, thanks.

rhtyd:
I've incorporated feedback from code review, given this has enough LGTMs and test results, I'll merge this as soon as Travis goes green. Thanks everyone for your feedback, review and testings.

PR #2295, opened 2017-10-12
CLOUDSTACK-10109: Enable dedication of public IPs to SSVM and CPVM

nvazquez:
JIRA Ticket: https://issues.apache.org/jira/browse/CLOUDSTACK-10109

### Description
This feature allow admins to dedicate a range of public IP addresses to the SSVM and CPVM, such that they can be subject to specific external firewall rules. The option to dedicate a public IP range to the System VMs (SSVM & CPVM) is added to the `createVlanIpRange` API method and the UI.

### Solution
Global setting 'system.vm.public.ip.reservation.mode.strictness' is added to determine if the use of the system VM reservation is strict (when true) or preferred (false), false by default.
When a range has been dedicated to System VMs, CloudStack should apply IPs from that range to
the public interfaces of the CPVM and the SSVM depending on global setting's value:
- If the global setting is set to false: then CloudStack will use any unused and unreserved public IP
addresses for system VMs only when the pool of reserved IPs has been exhausted
 - If the global setting is set to true: then CloudStack will fail to deploy the system VM when the pool
of reserved IPs has been exhausted, citing the lack of available IPs.

### UI Changes
Under Infrastructure -> Zone -> Physical Network -> Public -> IP Ranges, button 'Account' label is refactored to 'Set reservation':

When that button is clicked, dialog displayed is also refactored, including a new checkbox 'System VMs' which indicates if range should be dedicated for CPVM and SSVM, and a note indicating its usage:

When clicking on button for any created range, UI dialog displayed indicates whether IP range is dedicated for system vms or not:

PR #2298, opened 2017-10-16
CLOUDSTACK-9620: Enhancements for managed storage

mike-tutkowski:
Allowed zone-wide primary storage based on a custom plug-in to be added via the GUI in a KVM-only environment (previously this only worked for XenServer and VMware)

Added support for root disks on managed storage with KVM

Added support for volume snapshots with managed storage on KVM

Enabled creating a template directly from a volume (i.e. without having to go through a volume snapshot) on KVM with managed storage

Only allowed the resizing of a volume for managed storage on KVM if the volume in question is either not attached to a VM or is attached to a VM in the Stopped state

Included support for Reinstall VM on KVM with managed storage

Enabled offline migration on KVM from non-managed storage to managed storage and vice versa

Included support for online storage migration on KVM with managed storage (NFS and Ceph to managed storage)

Added support to download (extract) a managed-storage volume to a QCOW2 file

When uploading a file from outside of CloudStack to CloudStack, set the min and max IOPS, if applicable.

Included support for the KVM auto-convergence feature

The compression flag was actually added in version 1.0.3 (1000003) as opposed to version 1.3.0 (1003000) (changed this to reflect the correct version)

On KVM when using iSCSI-based managed storage, if the user shuts a VM down from the guest OS (as opposed to doing so from CloudStack), we need to pass to the KVM agent a list of applicable iSCSI volumes that need to be disconnected.

Added a new Global Setting: kvm.storage.live.migration.wait

For XenServer, added a check to enforce that only volumes from zone-wide managed storage can be storage motioned from a host in one cluster to a host in another cluster (cannot do so at the time being with volumes from cluster-scoped managed storage)

Don’t allow Storage XenMotion on a VM that has any managed-storage volume with one or more snapshots.

Enabled for managed storage with VMware: Template caching, create snapshot, delete snapshot, create volume from snapshot, and create template from snapshot

Added an SIOC API plug-in to support VMware SIOC

When starting a VM that uses managed storage in a cluster other than the one it last was running in, we need to remove the reference to the iSCSI volume from the original cluster.

Added the ability to revert a volume to a snapshot

Enabled cluster-scoped managed storage

Added support for VMware dynamic discovery

### DIFF
--- a/core/src/com/cloud/agent/api/MigrateCommand.java
+++ b/core/src/com/cloud/agent/api/MigrateCommand.java
@@ -53,10 +53,6 @@ public class MigrateCommand extends Command {
         return migrateStorage != null ? new HashMap<>(migrateStorage) : new HashMap<String, MigrateDiskInfo>();
     }
 
-    public boolean isMigrateStorage() {
-        return migrateStorage != null && !migrateStorage.isEmpty();
-    }
-
     public void setAutoConvergence(boolean autoConvergence) {
         this.autoConvergence = autoConvergence;
     }
@@ -100,11 +96,7 @@ public class MigrateCommand extends Command {
 
             @Override
             public String toString() {
-                switch(this) {
-                    case FILE: return "file";
-                    case BLOCK: return "block";
-                    default: throw new IllegalArgumentException();
-                }
+                return name().toLowerCase();
             }
         }
 
@@ -113,11 +105,7 @@ public class MigrateCommand extends Command {
 
             @Override
             public String toString() {
-                switch(this) {
-                    case QCOW2: return "qcow2";
-                    case RAW: return "raw";
-                    default: throw new IllegalArgumentException();
-                }
+                return name().toLowerCase();
             }
         }
 
@@ -126,11 +114,7 @@ public class MigrateCommand extends Command {
 
             @Override
             public String toString() {
-                switch(this) {
-                    case FILE: return "file";
-                    case DEV: return "dev";
-                    default: throw new IllegalArgumentException();
-                }
+                return name().toLowerCase();
             }
         }
 
--- a/core/src/com/cloud/agent/api/MigrateCommand.java
+++ b/core/src/com/cloud/agent/api/MigrateCommand.java
@@ -19,15 +19,20 @@
 
 package com.cloud.agent.api;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import com.cloud.agent.api.to.VirtualMachineTO;
 
 public class MigrateCommand extends Command {
-    String vmName;
-    String destIp;
-    String hostGuid;
-    boolean isWindows;
-    VirtualMachineTO vmTO;
-    boolean executeInSequence = false;
+    private String vmName;
+    private String destIp;
+    private Map<String, MigrateDiskInfo> migrateStorage;
+    private boolean autoConvergence;
+    private String hostGuid;
+    private boolean isWindows;
+    private VirtualMachineTO vmTO;
+    private boolean executeInSequence = false;
 
     protected MigrateCommand() {
     }
@@ -38,6 +43,26 @@ public class MigrateCommand extends Command {
         this.isWindows = isWindows;
         this.vmTO = vmTO;
         this.executeInSequence = executeInSequence;
+    }
+
+    public void setMigrateStorage(Map<String, MigrateDiskInfo> migrateStorage) {
+        this.migrateStorage = migrateStorage;
+    }
+
+    public Map<String, MigrateDiskInfo> getMigrateStorage() {
+        return migrateStorage != null ? new HashMap<>(migrateStorage) : new HashMap<String, MigrateDiskInfo>();
+    }
+
+    public boolean isMigrateStorage() {
+        return migrateStorage != null && !migrateStorage.isEmpty();
+    }
+
+    public void setAutoConvergence(boolean autoConvergence) {
+        this.autoConvergence = autoConvergence;
+    }
+
+    public boolean isAutoConvergence() {
+        return autoConvergence;
     }
 
     public boolean isWindows() {
@@ -67,5 +92,80 @@ public class MigrateCommand extends Command {
     @Override
     public boolean executeInSequence() {
         return executeInSequence;
+    }
+
+    public static class MigrateDiskInfo {
+        public enum DiskType {
+            FILE, BLOCK;
+
+            @Override
+            public String toString() {
+                switch(this) {
+                    case FILE: return "file";
+                    case BLOCK: return "block";
+                    default: throw new IllegalArgumentException();
+                }
+            }
+        }
+
+        public enum DriverType {
+            QCOW2, RAW;
+
+            @Override
+            public String toString() {
+                switch(this) {
+                    case QCOW2: return "qcow2";
+                    case RAW: return "raw";
+                    default: throw new IllegalArgumentException();
+                }
+            }
+        }
+
+        public enum Source {
+            FILE, DEV;
+
+            @Override
+            public String toString() {
+                switch(this) {
+                    case FILE: return "file";
+                    case DEV: return "dev";
+                    default: throw new IllegalArgumentException();
+                }
+            }
+        }
+
+        private final String serialNumber;
+        private final DiskType diskType;
+        private final DriverType driverType;
+        private final Source source;
+        private final String sourceText;
+
+        public MigrateDiskInfo(final String serialNumber, final DiskType diskType, final DriverType driverType, final Source source, final String sourceText) {
+            this.serialNumber = serialNumber;
+            this.diskType = diskType;
+            this.driverType = driverType;
+            this.source = source;
+            this.sourceText = sourceText;
+        }
+
+        public String getSerialNumber() {
+            return serialNumber;
+        }
+
+        public DiskType getDiskType() {
+            return diskType;
+        }
+
+        public DriverType getDriverType() {
+            return driverType;
+        }
+
+        public Source getSource() {
+            return source;
+        }
+
+        public String getSourceText() {
+            return sourceText;
+        }
     }
 }
--- a/core/src/com/cloud/agent/api/StartAnswer.java
+++ b/core/src/com/cloud/agent/api/StartAnswer.java
@@ -29,6 +29,10 @@ public class StartAnswer extends Answer {
 
     VirtualMachineTO vm;
     String hostGuid;
+    // key = an applicable IQN (ex. iqn.1998-01.com.vmware.iscsi:name1)
+    // value = a Map with the following data:
+    //   key = PATH or IMAGE_FORMAT (defined above)
+    //   value = Example if PATH is key: UUID of VDI; Example if IMAGE_FORMAT is key: DiskTO.VHD
     private Map<String, Map<String, String>> _iqnToData;
 
     protected StartAnswer() {
--- a/core/src/com/cloud/agent/api/StartAnswer.java
+++ b/core/src/com/cloud/agent/api/StartAnswer.java
@@ -24,9 +24,12 @@ import java.util.Map;
 import com.cloud.agent.api.to.VirtualMachineTO;
 
 public class StartAnswer extends Answer {
+    public static final String PATH = "path";
+    public static final String IMAGE_FORMAT = "imageFormat";
+
     VirtualMachineTO vm;
     String hostGuid;
-    Map<String, String> _iqnToPath;
+    private Map<String, Map<String, String>> _iqnToData;
 
     protected StartAnswer() {
     }
@@ -61,11 +64,11 @@ public class StartAnswer extends Answer {
         return hostGuid;
     }
 
-    public void setIqnToPath(Map<String, String> iqnToPath) {
-        _iqnToPath = iqnToPath;
+    public void setIqnToData(Map<String, Map<String, String>> iqnToData) {
+        _iqnToData = iqnToData;
     }
 
-    public Map<String, String> getIqnToPath() {
-        return _iqnToPath;
+    public Map<String, Map<String, String>> getIqnToData() {
+        return _iqnToData;
     }
 }
--- a/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -39,6 +39,7 @@ import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.log4j.Logger;
 
 import org.apache.cloudstack.affinity.dao.AffinityGroupVMMapDao;
@@ -567,7 +568,7 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         // Clean up volumes based on the vm's instance id
         volumeMgr.cleanupVolumes(vm.getId());
 
-        if (hostId != null && targets != null && targets.size() > 0) {
+        if (hostId != null && CollectionUtils.isNotEmpty(targets)) {
             removeDynamicTargets(hostId, targets);
         }
 
@@ -612,23 +613,27 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
 
         HostVO hostVO = _hostDao.findById(hostId);
 
-        if (hostVO != null && hostVO.getHypervisorType() == HypervisorType.VMware) {
-            List<VolumeVO> volumes = _volsDao.findByInstance(vmId);
+        if (hostVO == null || hostVO.getHypervisorType() != HypervisorType.VMware) {
+            return targets;
+        }
 
-            if (volumes != null) {
-                for (VolumeVO volume : volumes) {
-                    StoragePoolVO storagePoolVO = _storagePoolDao.findById(volume.getPoolId());
+        List<VolumeVO> volumes = _volsDao.findByInstance(vmId);
 
-                    if (storagePoolVO != null && storagePoolVO.isManaged()) {
-                        Map<String, String> target = new HashMap<>();
+        if (CollectionUtils.isEmpty(volumes)) {
+            return targets;
+        }
 
-                        target.put(ModifyTargetsCommand.STORAGE_HOST, storagePoolVO.getHostAddress());
-                        target.put(ModifyTargetsCommand.STORAGE_PORT, String.valueOf(storagePoolVO.getPort()));
-                        target.put(ModifyTargetsCommand.IQN, volume.get_iScsiName());
+        for (VolumeVO volume : volumes) {
+            StoragePoolVO storagePoolVO = _storagePoolDao.findById(volume.getPoolId());
 
-                        targets.add(target);
-                    }
-                }
+            if (storagePoolVO != null && storagePoolVO.isManaged()) {
+                Map<String, String> target = new HashMap<>();
+
+                target.put(ModifyTargetsCommand.STORAGE_HOST, storagePoolVO.getHostAddress());
+                target.put(ModifyTargetsCommand.STORAGE_PORT, String.valueOf(storagePoolVO.getPort()));
+                target.put(ModifyTargetsCommand.IQN, volume.get_iScsiName());
+
+                targets.add(target);
             }
         }
 
@@ -1419,19 +1424,21 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
 
         List<VolumeVO> volumes = _volsDao.findByInstance(vm.getId());
 
-        if (volumes != null) {
-            for (VolumeVO volume : volumes) {
-                StoragePoolVO storagePool = _storagePoolDao.findById(volume.getPoolId());
+        if (CollectionUtils.isEmpty(volumes)) {
+            return volumesToDisconnect;
+        }
 
-                if (storagePool != null && storagePool.isManaged()) {
-                    Map<String, String> info = new HashMap<>(3);
+        for (VolumeVO volume : volumes) {
+            StoragePoolVO storagePool = _storagePoolDao.findById(volume.getPoolId());
 
-                    info.put(DiskTO.STORAGE_HOST, storagePool.getHostAddress());
-                    info.put(DiskTO.STORAGE_PORT, String.valueOf(storagePool.getPort()));
-                    info.put(DiskTO.IQN, volume.get_iScsiName());
+            if (storagePool != null && storagePool.isManaged()) {
+                Map<String, String> info = new HashMap<>(3);
 
-                    volumesToDisconnect.add(info);
-                }
+                info.put(DiskTO.STORAGE_HOST, storagePool.getHostAddress());
+                info.put(DiskTO.STORAGE_PORT, String.valueOf(storagePool.getPort()));
+                info.put(DiskTO.IQN, volume.get_iScsiName());
+
+                volumesToDisconnect.add(info);
             }
         }
 
--- a/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -85,6 +85,7 @@ import com.cloud.agent.api.ClusterVMMetaDataSyncAnswer;
 import com.cloud.agent.api.ClusterVMMetaDataSyncCommand;
 import com.cloud.agent.api.Command;
 import com.cloud.agent.api.MigrateCommand;
+import com.cloud.agent.api.ModifyTargetsCommand;
 import com.cloud.agent.api.PingRoutingCommand;
 import com.cloud.agent.api.PlugNicAnswer;
 import com.cloud.agent.api.PlugNicCommand;
@@ -114,6 +115,7 @@ import com.cloud.agent.manager.Commands;
 import com.cloud.agent.manager.allocator.HostAllocator;
 import com.cloud.alert.AlertManager;
 import com.cloud.capacity.CapacityManager;
+import com.cloud.configuration.Config;
 import com.cloud.dc.ClusterDetailsDao;
 import com.cloud.dc.ClusterDetailsVO;
 import com.cloud.dc.DataCenter;
@@ -536,6 +538,8 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
 
         final Long hostId = vm.getHostId() != null ? vm.getHostId() : vm.getLastHostId();
 
+        List<Map<String, String>> targets = getTargets(hostId, vm.getId());
+
         if (volumeExpungeCommands != null && volumeExpungeCommands.size() > 0 && hostId != null) {
             final Commands cmds = new Commands(Command.OnError.Stop);
 
@@ -562,6 +566,10 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
 
         // Clean up volumes based on the vm's instance id
         volumeMgr.cleanupVolumes(vm.getId());
+
+        if (hostId != null && targets != null && targets.size() > 0) {
+            removeDynamicTargets(hostId, targets);
+        }
 
         final VirtualMachineGuru guru = getVmGuru(vm);
         guru.finalizeExpunge(vm);
@@ -597,6 +605,60 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
             s_logger.debug("Expunged " + vm);
         }
 
+    }
+
+    private List<Map<String, String>> getTargets(Long hostId, long vmId) {
+        List<Map<String, String>> targets = new ArrayList<>();
+
+        HostVO hostVO = _hostDao.findById(hostId);
+
+        if (hostVO != null && hostVO.getHypervisorType() == HypervisorType.VMware) {
+            List<VolumeVO> volumes = _volsDao.findByInstance(vmId);
+
+            if (volumes != null) {
+                for (VolumeVO volume : volumes) {
+                    StoragePoolVO storagePoolVO = _storagePoolDao.findById(volume.getPoolId());
+
+                    if (storagePoolVO != null && storagePoolVO.isManaged()) {
+                        Map<String, String> target = new HashMap<>();
+
+                        target.put(ModifyTargetsCommand.STORAGE_HOST, storagePoolVO.getHostAddress());
+                        target.put(ModifyTargetsCommand.STORAGE_PORT, String.valueOf(storagePoolVO.getPort()));
+                        target.put(ModifyTargetsCommand.IQN, volume.get_iScsiName());
+
+                        targets.add(target);
+                    }
+                }
+            }
+        }
+
+        return targets;
+    }
+
+    private void removeDynamicTargets(long hostId, List<Map<String, String>> targets) {
+        ModifyTargetsCommand cmd = new ModifyTargetsCommand();
+
+        cmd.setTargets(targets);
+        cmd.setApplyToAllHostsInCluster(true);
+        cmd.setAdd(false);
+        cmd.setTargetTypeToRemove(ModifyTargetsCommand.TargetTypeToRemove.DYNAMIC);
+
+        sendModifyTargetsCommand(cmd, hostId);
+    }
+
+    private void sendModifyTargetsCommand(ModifyTargetsCommand cmd, long hostId) {
+        Answer answer = _agentMgr.easySend(hostId, cmd);
+
+        if (answer == null) {
+            String msg = "Unable to get an answer to the modify targets command";
+
+            s_logger.warn(msg);
+        }
+        else if (!answer.getResult()) {
+            String msg = "Unable to modify target on the following host: " + hostId;
+
+            s_logger.warn(msg);
+        }
     }
 
     @Override
@@ -1073,8 +1135,10 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
 
                     startAnswer = cmds.getAnswer(StartAnswer.class);
                     if (startAnswer != null && startAnswer.getResult()) {
-                        handlePath(vmTO.getDisks(), startAnswer.getIqnToPath());
+                        handlePath(vmTO.getDisks(), startAnswer.getIqnToData());
+
                         final String host_guid = startAnswer.getHost_guid();
+
                         if (host_guid != null) {
                             final HostVO finalHost = _resourceMgr.findHostByGuid(host_guid);
                             if (finalHost == null) {
@@ -1254,8 +1318,8 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
     }
 
     // for managed storage on XenServer and VMware, need to update the DB with a path if the VDI/VMDK file was newly created
-    private void handlePath(final DiskTO[] disks, final Map<String, String> iqnToPath) {
-        if (disks != null && iqnToPath != null) {
+    private void handlePath(final DiskTO[] disks, final Map<String, Map<String, String>> iqnToData) {
+        if (disks != null && iqnToData != null) {
             for (final DiskTO disk : disks) {
                 final Map<String, String> details = disk.getDetails();
                 final boolean isManaged = details != null && Boolean.parseBoolean(details.get(DiskTO.MANAGED));
@@ -1264,12 +1328,31 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
                     final Long volumeId = disk.getData().getId();
                     final VolumeVO volume = _volsDao.findById(volumeId);
                     final String iScsiName = volume.get_iScsiName();
-                    final String path = iqnToPath.get(iScsiName);
 
-                    if (path != null) {
-                        volume.setPath(path);
+                    boolean update = false;
 
-                        _volsDao.update(volumeId, volume);
+                    final Map<String, String> data = iqnToData.get(iScsiName);
+
+                    if (data != null) {
+                        final String path = data.get(StartAnswer.PATH);
+
+                        if (path != null) {
+                            volume.setPath(path);
+
+                            update = true;
+                        }
+
+                        final String imageFormat = data.get(StartAnswer.IMAGE_FORMAT);
+
+                        if (imageFormat != null) {
+                            volume.setFormat(ImageFormat.valueOf(imageFormat));
+
+                            update = true;
+                        }
+
+                        if (update) {
+                            _volsDao.update(volumeId, volume);
+                        }
                     }
                 }
             }
@@ -1331,10 +1414,35 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         }
     }
 
+    private List<Map<String, String>> getVolumesToDisconnect(VirtualMachine vm) {
+        List<Map<String, String>> volumesToDisconnect = new ArrayList<>();
+
+        List<VolumeVO> volumes = _volsDao.findByInstance(vm.getId());
+
+        if (volumes != null) {
+            for (VolumeVO volume : volumes) {
+                StoragePoolVO storagePool = _storagePoolDao.findById(volume.getPoolId());
+
+                if (storagePool != null && storagePool.isManaged()) {
+                    Map<String, String> info = new HashMap<>(3);
+
+                    info.put(DiskTO.STORAGE_HOST, storagePool.getHostAddress());
+                    info.put(DiskTO.STORAGE_PORT, String.valueOf(storagePool.getPort()));
+                    info.put(DiskTO.IQN, volume.get_iScsiName());
+
+                    volumesToDisconnect.add(info);
+                }
+            }
+        }
+
+        return volumesToDisconnect;
+    }
+
     protected boolean sendStop(final VirtualMachineGuru guru, final VirtualMachineProfile profile, final boolean force, final boolean checkBeforeCleanup) {
         final VirtualMachine vm = profile.getVirtualMachine();
         StopCommand stpCmd = new StopCommand(vm, getExecuteInSequence(vm.getHypervisorType()), checkBeforeCleanup);
         stpCmd.setControlIp(getControlNicIpForVM(vm));
+        stpCmd.setVolumesToDisconnect(getVolumesToDisconnect(vm));
         final StopCommand stop = stpCmd;
         try {
             Answer answer = null;
@@ -2103,6 +2211,12 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         try {
             final boolean isWindows = _guestOsCategoryDao.findById(_guestOsDao.findById(vm.getGuestOSId()).getCategoryId()).getName().equalsIgnoreCase("Windows");
             final MigrateCommand mc = new MigrateCommand(vm.getInstanceName(), dest.getHost().getPrivateIpAddress(), isWindows, to, getExecuteInSequence(vm.getHypervisorType()));
+
+            String autoConvergence = _configDao.getValue(Config.KvmAutoConvergence.toString());
+            boolean kvmAutoConvergence = Boolean.parseBoolean(autoConvergence);
+
+            mc.setAutoConvergence(kvmAutoConvergence);
+
             mc.setHostGuid(dest.getHost().getGuid());
 
             try {
@@ -2176,32 +2290,48 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         final Map<Volume, StoragePool> volumeToPoolObjectMap = new HashMap<>();
 
         for (final VolumeVO volume : allVolumes) {
-            final Long poolId = volumeToPool.get(Long.valueOf(volume.getId()));
+            final Long poolId = volumeToPool.get(volume.getId());
             final StoragePoolVO destPool = _storagePoolDao.findById(poolId);
             final StoragePoolVO currentPool = _storagePoolDao.findById(volume.getPoolId());
             final DiskOfferingVO diskOffering = _diskOfferingDao.findById(volume.getDiskOfferingId());
 
             if (destPool != null) {
-                // Check if pool is accessible from the destination host and disk offering with which the volume was
-                // created is compliant with the pool type.
-                if (_poolHostDao.findByPoolHost(destPool.getId(), host.getId()) == null || destPool.isLocal() != diskOffering.getUseLocalStorage()) {
-                    // Cannot find a pool for the volume. Throw an exception.
-                    throw new CloudRuntimeException("Cannot migrate volume " + volume + " to storage pool " + destPool + " while migrating vm to host " + host +
-                            ". Either the pool is not accessible from the host or because of the offering with which the volume is created it cannot be placed on " +
-                            "the given pool.");
-                } else if (destPool.getId() == currentPool.getId()) {
-                    // If the pool to migrate to is the same as current pool, the volume doesn't need to be migrated.
-                } else {
-                    volumeToPoolObjectMap.put(volume, destPool);
+                if (currentPool.isManaged()) {
+                    if (destPool.getId() == currentPool.getId()) {
+                        volumeToPoolObjectMap.put(volume, currentPool);
+                    }
+                    else {
+                        throw new CloudRuntimeException("Currently, a volume on managed storage can only be 'migrated' to itself.");
+                    }
+                }
+                else {
+                    // Check if pool is accessible from the destination host and disk offering with which the volume was
+                    // created is compliant with the pool type.
+                    if (_poolHostDao.findByPoolHost(destPool.getId(), host.getId()) == null || destPool.isLocal() != diskOffering.getUseLocalStorage()) {
+                        // Cannot find a pool for the volume. Throw an exception.
+                        throw new CloudRuntimeException("Cannot migrate volume " + volume + " to storage pool " + destPool + " while migrating vm to host " + host +
+                                ". Either the pool is not accessible from the host or because of the offering with which the volume is created it cannot be placed on " +
+                                "the given pool.");
+                    } else if (destPool.getId() == currentPool.getId()) {
+                        // If the pool to migrate to is the same as current pool, the volume doesn't need to be migrated.
+                    } else {
+                        volumeToPoolObjectMap.put(volume, destPool);
+                    }
                 }
             } else {
                 if (currentPool.isManaged()) {
-                    volumeToPoolObjectMap.put(volume, currentPool);
+                    if (currentPool.getScope() == ScopeType.ZONE) {
+                        volumeToPoolObjectMap.put(volume, currentPool);
+                    }
+                    else {
+                        throw new CloudRuntimeException("Currently, you can only 'migrate' a volume on managed storage if its storage pool is zone wide.");
+                    }
                 } else {
                     // Find a suitable pool for the volume. Call the storage pool allocator to find the list of pools.
 
                     final DiskProfile diskProfile = new DiskProfile(volume, diskOffering, profile.getHypervisorType());
-                    final DataCenterDeployment plan = new DataCenterDeployment(host.getDataCenterId(), host.getPodId(), host.getClusterId(), host.getId(), null, null);
+                    final DataCenterDeployment plan = new DataCenterDeployment(host.getDataCenterId(), host.getPodId(), host.getClusterId(),
+                            host.getId(), null, null);
 
                     final List<StoragePool> poolList = new ArrayList<>();
                     final ExcludeList avoid = new ExcludeList();
@@ -3588,6 +3718,12 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         try {
             final boolean isWindows = _guestOsCategoryDao.findById(_guestOsDao.findById(vm.getGuestOSId()).getCategoryId()).getName().equalsIgnoreCase("Windows");
             final MigrateCommand mc = new MigrateCommand(vm.getInstanceName(), dest.getHost().getPrivateIpAddress(), isWindows, to, getExecuteInSequence(vm.getHypervisorType()));
+
+            String autoConvergence = _configDao.getValue(Config.KvmAutoConvergence.toString());
+            boolean kvmAutoConvergence = Boolean.parseBoolean(autoConvergence);
+
+            mc.setAutoConvergence(kvmAutoConvergence);
+
             mc.setHostGuid(dest.getHost().getGuid());
 
             try {
--- a/engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
@@ -19,6 +19,16 @@
 package org.apache.cloudstack.storage.motion;
 
 import com.cloud.agent.AgentManager;
+import com.cloud.agent.api.Answer;
+import com.cloud.agent.api.storage.CopyVolumeAnswer;
+import com.cloud.agent.api.storage.CopyVolumeCommand;
+import com.cloud.agent.api.MigrateAnswer;
+import com.cloud.agent.api.MigrateCommand;
+import com.cloud.agent.api.ModifyTargetsAnswer;
+import com.cloud.agent.api.ModifyTargetsCommand;
+import com.cloud.agent.api.PrepareForMigrationCommand;
+import com.cloud.agent.api.storage.MigrateVolumeAnswer;
+import com.cloud.agent.api.storage.MigrateVolumeCommand;
 import com.cloud.agent.api.to.DataStoreTO;
 import com.cloud.agent.api.to.DataTO;
 import com.cloud.agent.api.to.DiskTO;
@@ -33,24 +43,35 @@ import com.cloud.host.HostVO;
 import com.cloud.host.dao.HostDao;
 import com.cloud.host.dao.HostDetailsDao;
 import com.cloud.hypervisor.Hypervisor.HypervisorType;
-import com.cloud.server.ManagementService;
 import com.cloud.storage.DataStoreRole;
 import com.cloud.storage.DiskOfferingVO;
 import com.cloud.storage.Snapshot;
 import com.cloud.storage.SnapshotVO;
 import com.cloud.storage.Storage.ImageFormat;
+import com.cloud.storage.StorageManager;
+import com.cloud.storage.VMTemplateVO;
 import com.cloud.storage.VolumeDetailVO;
+import com.cloud.storage.Volume;
 import com.cloud.storage.VolumeVO;
 import com.cloud.storage.dao.DiskOfferingDao;
+import com.cloud.storage.dao.GuestOSCategoryDao;
+import com.cloud.storage.dao.GuestOSDao;
 import com.cloud.storage.dao.SnapshotDao;
 import com.cloud.storage.dao.SnapshotDetailsDao;
 import com.cloud.storage.dao.SnapshotDetailsVO;
+import com.cloud.storage.dao.VMTemplateDao;
 import com.cloud.storage.dao.VolumeDao;
 import com.cloud.storage.dao.VolumeDetailsDao;
 import com.cloud.utils.NumbersUtil;
+import com.cloud.utils.db.GlobalLock;
 import com.cloud.utils.exception.CloudRuntimeException;
+import com.cloud.vm.VirtualMachine;
 import com.cloud.vm.VirtualMachineManager;
+import com.cloud.vm.VMInstanceVO;
+import com.cloud.vm.dao.VMInstanceDao;
+
 import com.google.common.base.Preconditions;
+
 import org.apache.cloudstack.engine.subsystem.api.storage.ChapInfo;
 import org.apache.cloudstack.engine.subsystem.api.storage.ClusterScope;
 import org.apache.cloudstack.engine.subsystem.api.storage.CopyCommandResult;
@@ -81,46 +102,58 @@ import org.apache.cloudstack.storage.command.CopyCommand;
 import org.apache.cloudstack.storage.command.ResignatureAnswer;
 import org.apache.cloudstack.storage.command.ResignatureCommand;
 import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
+import org.apache.cloudstack.storage.datastore.db.SnapshotDataStoreDao;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
 import org.apache.cloudstack.storage.to.PrimaryDataStoreTO;
 import org.apache.cloudstack.storage.to.VolumeObjectTO;
 import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang.math.NumberUtils;
 import org.apache.log4j.Logger;
+
 import org.springframework.stereotype.Component;
 
 import javax.inject.Inject;
+
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
-import java.util.concurrent.ExecutionException;
+import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 @Component
 public class StorageSystemDataMotionStrategy implements DataMotionStrategy {
     private static final Logger LOGGER = Logger.getLogger(StorageSystemDataMotionStrategy.class);
     private static final Random RANDOM = new Random(System.nanoTime());
+    private static final int LOCK_TIME_IN_SECONDS = 300;
+    private static final String OPERATION_NOT_SUPPORTED = "This operation is not supported.";
 
     @Inject private AgentManager _agentMgr;
     @Inject private ConfigurationDao _configDao;
     @Inject private DataStoreManager dataStoreMgr;
     @Inject private DiskOfferingDao _diskOfferingDao;
+    @Inject private GuestOSCategoryDao _guestOsCategoryDao;
+    @Inject private GuestOSDao _guestOsDao;
     @Inject private ClusterDao clusterDao;
     @Inject private HostDao _hostDao;
     @Inject private HostDetailsDao hostDetailsDao;
-    @Inject private ManagementService _mgr;
     @Inject private PrimaryDataStoreDao _storagePoolDao;
     @Inject private SnapshotDao _snapshotDao;
+    @Inject private SnapshotDataStoreDao _snapshotDataStoreDao;
     @Inject private SnapshotDetailsDao _snapshotDetailsDao;
+    @Inject private VMInstanceDao _vmDao;
+    @Inject private VMTemplateDao _vmTemplateDao;
     @Inject private VolumeDao _volumeDao;
     @Inject private VolumeDataFactory _volumeDataFactory;
     @Inject private VolumeDetailsDao volumeDetailsDao;
     @Inject private VolumeService _volumeService;
     @Inject private StorageCacheManager cacheMgr;
     @Inject private EndPointSelector selector;
+
     @Override
     public StrategyPriority canHandle(DataObject srcData, DataObject destData) {
         if (srcData instanceof SnapshotInfo) {
@@ -136,9 +169,40 @@ public class StorageSystemDataMotionStrategy implements DataMotionStrategy {
             return StrategyPriority.HIGHEST;
         }
 
+        if (srcData instanceof VolumeInfo && destData instanceof VolumeInfo) {
+            VolumeInfo srcVolumeInfo = (VolumeInfo)srcData;
+
+            if (isVolumeOnManagedStorage(srcVolumeInfo)) {
+                return StrategyPriority.HIGHEST;
+            }
+
+            VolumeInfo destVolumeInfo = (VolumeInfo)destData;
+
+            if (isVolumeOnManagedStorage(destVolumeInfo)) {
+                return StrategyPriority.HIGHEST;
+            }
+        }
+
+        if (srcData instanceof VolumeInfo && destData instanceof TemplateInfo) {
+            VolumeInfo srcVolumeInfo = (VolumeInfo)srcData;
+
+            if (isVolumeOnManagedStorage(srcVolumeInfo)) {
+                return StrategyPriority.HIGHEST;
+            }
+        }
+
         return StrategyPriority.CANT_HANDLE;
     }
 
+    private boolean isVolumeOnManagedStorage(VolumeInfo volumeInfo) {
+        long storagePooldId = volumeInfo.getDataStore().getId();
+        StoragePoolVO storagePoolVO = _storagePoolDao.findById(storagePooldId);
+
+        return storagePoolVO.isManaged();
+    }
+
+    // canHandle returns true if the storage driver for the DataObject that's passed in can support certain features (what features we
+    // care about during a particular invocation of this method depend on what type of DataObject was passed in (ex. VolumeInfo versus SnapshotInfo)).
     private boolean canHandle(DataObject dataObject) {
         Preconditions.checkArgument(dataObject != null, "Passing 'null' to dataObject of canHandle(DataObject) is not supported.");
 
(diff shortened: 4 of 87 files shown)


mike-tutkowski:
Managed-storage regression test runs:

[results.txt](https://github.com/apache/cloudstack/files/1388769/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388770/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388771/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388772/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388773/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388776/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388777/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388778/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388779/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388780/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/1388781/results.txt)

syed:
>  Don’t allow Storage XenMotion on a VM that has any managed-storage volume with one or more snapshots.

Is this VM snapshots on XenServer or snapshots on SolidFire?

mike-tutkowski:
That is actually snapshots on the storage system (SolidFire snapshots, in this case). The problem is that we have to copy the volume from a source to a new destination volume (due to the way Storage XenMotion works) and can’t reproduce the snapshots on the destination side.

mike-tutkowski:
Although, @syed, now that you mention it, we should probably prohibit Storage XenMotion when either existing backend snapshots (volume snapshots) or existing VM snapshots are present. I believe either snapshot type will be a problem in this scenario.

This, of course, doesn't mean we have to prohibit them forever. Perhaps we will have a good solution in the future. However, at least for this release, they should both be prohibited.

I have taken a note to implement the extract check. Thanks!

mike-tutkowski:
@syed I investigated the VM-snapshots issue. As it turns out, there is already logic in place to prohibit Storage XenMotion from occurring if the VM has any VM snapshots on it.

rhtyd:
@blueorangutan package

rhtyd:
@mike-tutkowski is this ready for prime time review and testing, or still in progress?

mike-tutkowski:
Hi @rhtyd - This is essentially ready to go minus a small bit of VMware work around iSCSI dynamic discovery that I've been adding to it.

I would say it would be great if people would review it now and just kind of know there might be a couple more VMware updates to it over the next week or so. There shouldn't be much code added to it. The extra VMware code is something I worked on for a customer last month for the 4.6 branch and am now merging to this PR.

rhtyd:
@mike-tutkowski thanks for replying, looking forward to your changes. Once you're done, please ping people on this PR and ask explicitly. I may find some time to review by end of next week.

rhtyd:
@mike-tutkowski ping me when you want review/tests to be run, thanks.

mike-tutkowski:
Sounds good, @rhtyd - it should be pretty soon. I needed to mainly attend to other work over the past week or so, so I didn't have a chance to finalize what I was working on for VMware here (dynamic discovery). It should probably be ready early next week. Right now I am running my tests and making some final touch-up tweaks.

mike-tutkowski:
@rhtyd @syed @swill Just an update: I have added that final feature (related to VMware dynamic discovery), so this PR is now completely ready for review. Thanks!

rhtyd:
Thanks @mike-tutkowski I'll do review and help run tests next week. Happy holidays!

mike-tutkowski:
Thanks, @rhtyd!

rhtyd:
@mike-tutkowski I've resolved conflict issue, will run some tests. Please check my resolution.
@blueorangutan package

borisstoyanov:
@blueorangutan test

plugins/api/solidfire-sioc/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+  software distributed under the License is distributed on an
+  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  KIND, either express or implied. See the License for the
+  specific language governing permissions and limitations
+  under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <artifactId>cloud-plugin-api-solidfire-sioc</artifactId>
+  <name>Apache CloudStack Plugin - API SolidFire (SIOC)</name>
```
rhtyd: @mike-tutkowski what is SIOC, can this be under `plugins/solidfire/...` than under `plugin/api`?

plugins/api/solidfire-sioc/src/org/apache/cloudstack/api/command/admin/solidfire/UpdateSiocInfoCmd.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.cloudstack.context.CallContext;
+import org.apache.cloudstack.solidfire.SolidFireSiocManager;
+
+import com.cloud.user.Account;
+
+@APICommand(name = "updateSiocInfo", responseObject = ApiUpdateSiocInfoResponse.class, description = "Update SIOC info",
+        requestHasSensitiveInfo = false, responseHasSensitiveInfo = false)
+public class UpdateSiocInfoCmd extends BaseCmd {
+    private static final Logger s_logger = Logger.getLogger(UpdateSiocInfoCmd.class.getName());
+    private static final String s_name = "updatesiocinforesponse";
```
rhtyd: Please refactor using APINAME pattern, see recent API implementations such as the `*RoleCmd.java`.

plugins/api/solidfire-sioc/src/org/apache/cloudstack/api/command/admin/solidfire/UpdateSiocInfoCmd.java
```diff
@@ -0,0 +1,97 @@
...
+import org.apache.cloudstack.api.Parameter;
+import org.apache.cloudstack.api.response.StoragePoolResponse;
+import org.apache.cloudstack.api.response.ZoneResponse;
+import org.apache.cloudstack.api.response.solidfire.ApiUpdateSiocInfoResponse;
+import org.apache.cloudstack.context.CallContext;
+import org.apache.cloudstack.solidfire.SolidFireSiocManager;
+
+import com.cloud.user.Account;
+
+@APICommand(name = "updateSiocInfo", responseObject = ApiUpdateSiocInfoResponse.class, description = "Update SIOC info",
```
rhtyd: @mike-tutkowski please add `version`, `authorized` etc. See the `*RoleCmd.java` as an example.

rhtyd:
@mike-tutkowski I've left some comments, my only major concern is the placement of the `SIOC API plug-in for VMware SIOC` otherwise LGTM.

plugins/api/solidfire-sioc/pom.xml
mike-tutkowski: @rhtyd From what I recall, I have enabled this to only run with the -Dnoredist flag. It is an API plug-in, so I just placed it with the other API plug-ins that I found. I'm fine with moving it, though, if you think that's the better way to go. Let me know. Thanks!

rhtyd: @mike-tutkowski I think `plugins/api` is for hosting cloudstack's API related plugin such as rate limiting plugin and the api discovery plugin. Perhaps you can consider moving this to `plugin/integrations` (if it's related to integration to external system/service sioc?) or `plugins/storage` or perhaps a more appropriate location?

rhtyd:
@mike-tutkowski can you fix the conflicts and address outstanding issues?
Additional code review is also requested.

mike-tutkowski:
Hi @rhtyd Sorry for the delay in responding. I believe I have addressed all of your points except one (which I wanted to run past you before making any changes here).

The one remaining issue is with where to place the SIOC code.

SIOC is short for VMware Storage I/O Control.

Initially I placed it under plugins/api/solidfire-sioc because 1) I wanted it to extend the standard CloudStack API and I thought all API plug-ins needed to be placed in this location and 2) originally there was SolidFire-specific code in here (hence the naming of some folders).

As it turns out, in working with the customer for whom I originally developed this code (the SIOC code, that is), I was able to extract all SolidFire-specific references.

How's about this? If I go ahead and remove references to SolidFire from this SIOC code (ex. folder naming, file naming, etc.), would you be OK with me placing it here: plugins/api/sioc?

core/src/com/cloud/agent/api/StartAnswer.java
```diff
@@ -24,9 +24,12 @@
 import com.cloud.agent.api.to.VirtualMachineTO;
 
 public class StartAnswer extends Answer {
+    public static final String PATH = "path";
+    public static final String IMAGE_FORMAT = "imageFormat";
+
     VirtualMachineTO vm;
     String hostGuid;
-    Map<String, String> _iqnToPath;
+    private Map<String, Map<String, String>> _iqnToData;
```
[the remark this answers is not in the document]
mike-tutkowski: I could go either way on it. It's just a Map whose key (a String) returns another Map.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -226,10 +310,12 @@ public void copyAsync(DataObject srcData, DataObject destData, Host destHost, As
 
                 if (canHandleSrc) {
                     String errMsg = "This operation is not supported (DataStoreCapabilities.STORAGE_SYSTEM_SNAPSHOT " +
-                            "not supported by source storage plug-in). " + getSrcDataStoreMsg(srcData);
+                            "not supported by source storage plug-in). " + getDestDataStoreMsg(destData);
```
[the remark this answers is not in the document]
mike-tutkowski: I can add a comment. Basically, there are three if statements in a row. The first one covers 'if (canHandleSrc && canHandleDest)'. The second one covers 'if (canHandleDest)' (so, at this point, you know that both canHandleSrc && canHandleDest can't resolve to true). The final one is 'if (canHandleSrc)' (so, we are saying that our storage plug-in can handle the source side, but not the destination side, which is a problem).

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -242,15 +328,99 @@ public void copyAsync(DataObject srcData, DataObject destData, Host destHost, As
...
 
+                invokeCallback(errMsg, callback);
+
                 throw new UnsupportedOperationException(errMsg);
             }
 
             handleCreateVolumeFromTemplateBothOnStorageSystem((TemplateInfo)srcData, (VolumeInfo)destData, callback);
 
+            return;
+        } else if (srcData instanceof VolumeInfo && destData instanceof VolumeInfo) {
```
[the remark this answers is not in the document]
mike-tutkowski: I can investigate a refactor of that method. IntelliJ is also complaining about its inability to analyze it properly.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -684,40 +1237,154 @@ private void handleCreateVolumeFromSnapshotBothOnStorageSystem(SnapshotInfo snap
...
+
+            Map<String, String> destDetails = getVolumeDetails(destVolumeInfo);
+
+            copyCommand.setOptions2(destDetails);
+
+            copyCmdAnswer = (CopyCmdAnswer)_agentMgr.send(hostVO.getId(), copyCommand);
+        }
+        catch (CloudRuntimeException | AgentUnavailableException | OperationTimedoutException ex) {
+            String msg = "Failed to copy image : ";
+
```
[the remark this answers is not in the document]
mike-tutkowski: I'm not sure I understand what you mean here, @DaanHoogland. Do you mean the indentation under line 1373? Perhaps you are referring to line 1374 being blank? If it's 1374 being blank, I guess I just lean toward using empty lines as a way to not cram all the code together into one, unreadable block.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -684,40 +1237,154 @@ private void handleCreateVolumeFromSnapshotBothOnStorageSystem(SnapshotInfo snap
...
+            LOGGER.warn(msg, ex);
+
+            throw new CloudRuntimeException(msg + ex.getMessage());
```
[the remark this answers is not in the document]
mike-tutkowski: In a prior code review (for a different PR), the reviewer didn't like us having to have a "throws" clause with a bunch of exception types all over the place. His suggestion was to catch the specific exception, log it so we can find the details, then convert it to a CloudRuntimeException (using the message of the exception) and throw that (since that's a type of RuntimeException and, as such, doesn't have to be declared in the "throws" clause).

plugins/hypervisors/vmware/src/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -1426,19 +1998,66 @@ private Answer attachVolume(Command cmd, DiskTO disk, boolean isAttach, boolean
         } catch (Throwable e) {
             if (e instanceof RemoteException) {
                 s_logger.warn("Encounter remote exception to vCenter, invalidate VMware session context");
+
                 hostService.invalidateServiceContext(null);
             }
 
             String msg = "";
-            if (isAttach)
+
```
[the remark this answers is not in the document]
mike-tutkowski: I tend to separate variables from the code that operates on them with line spacing. I don't like reading people's code that is completely crammed together, so I don't write it that way. I'm sure some like the extra spacing and others don't.

server/src/com/cloud/configuration/Config.java
```diff
@@ -191,6 +191,22 @@
             "10800",
             "In second, timeout for download template to primary storage",
             null),
+    KvmStorageOfflineMigrationWait(
```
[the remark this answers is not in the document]
mike-tutkowski: I have a question about that. I have this config key (and corresponding logic to use it) implemented in a custom version of 4.6 for a customer and this customer is looking to migrate to 4.11. If I change this from being in Config to being in ConfigKey, will that cause problems for them when they migrate to 4.11?

core/src/com/cloud/agent/api/StartAnswer.java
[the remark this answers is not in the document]
mike-tutkowski: With regards to multiple PRs: Originally I had it that way (VMware in one PR, KVM in another, and XenServer in a third). Unfortunately, since 4.10 took a very long to to come out, it became a big pain to constantly rebase and maintain all of those PRs. As such, I merged the code into this one to simplify PR maintenance and to minimize the chances of making mistakes with rebasing due to having to perform that action so many times (and having to make so many manual merge fixes along the way). At this point, I think it's safest to stay with this one, large PR.

Feel free to provide your code style and maintainability suggestions. I'm happy to hear what you have to say.

mike-tutkowski:
@rhtyd Just an FYI that I went ahead and refactored the VMware SIOC code to remove references to SolidFire. It is now located in plugins/api/vmware-sioc.

server/src/com/cloud/configuration/Config.java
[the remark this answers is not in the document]
DaanHoogland: No, the config item with the same key will be more dynamic and will be read and written to the same record.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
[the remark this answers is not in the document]
DaanHoogland: I don't mind extra whitelines, even though I think these are noise but I don't like methods that don't fit in my 24 line terminal (it's actually 30x180)

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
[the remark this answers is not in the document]
rafaelweingartner: This is a funny cause. I get amused imagining the thought process of a person that code these things...
Documentation could help. However, I would first suggesting extracting lines 191-227 to a method. Then, instead of starting with `if (canHandleSrc && canHandleDest)`, you can do the following:
if (!canHandleSrc){Throw exception regarding source not being able to handle the request}
if (!canHandleDest){Throw exception regarding destination not being able to handle the request }

Then, if you pass those two IFs, you do not need a third one to check `if (canHandleSrc && canHandleDest)`

server/src/com/cloud/configuration/Config.java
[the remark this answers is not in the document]
mike-tutkowski: Sounds good, @DaanHoogland. I can migrate to the ConfigKey approach for that key.

I'll handle that and refactor the copyAsync method.

rhtyd:
@mike-tutkowski can you resolve the conflicts, also see any outstanding issue
@DaanHoogland @rafaelweingartner are you lgtm on this?

mike-tutkowski:
@rhtyd @DaanHoogland I have refactored the createAsync method in StorageSystemDataMotionStrategy.java. I still need to make use of the ConfigKey class instead of the Config class, as Daan requested. I should be able to do this tomorrow (well, technically, later today.

DaanHoogland:
@mike-tutkowski @rafaelweingartner the questions are:
Are we comfortable with this code as is?
Do we want to go with it in 4.11?
How much effort do we put in after releasing this?
I can not fully assess this as it is so big. And the argument of cyclic complexity of both methods and classes is weighing heavily.

btw @mike-tutkowski I'll stop bugging you about the noisy whitelines but I will remove a lot of them at my next chance ;)

mike-tutkowski:
@DaanHoogland Ideally, yes, 4.11 (I have several customers looking for this to go into a LTS release).

As for support: I don't plan on going anywhere anytime soon. :) Also, I have a bunch of new integration tests (as part of this PR) to test out the new features (as well as whatever tests @rhtyd runs as part of the PR process).

Feel free to change white space as you see fit. :) I do that kind of thing when I'm looking through existing code, too.

rhtyd:
@blueorangutan package

DaanHoogland:
So early bird @mike-tutkowski did you address the complexity question @rafaelweingartner and I had?

rafaelweingartner:
@DaanHoogland comfortable I am not. However, this is an interesting feature; and I understand how hard it is to create and maintain something like (conflict may happen...). If everything is working and tests passing, I would be ok with it.

One thing we must keep in mind though. We need to start looking for better code (code for people). I know that it was like this before and @mike-tutkowski is only working on top of it.

mike-tutkowski:
@DaanHoogland I've been searching for comments from @rafaelweingartner, but I don't see any regarding complexity.

I will say that 90% of this code exists in prior (private) branches that customers of mine have been using throughout much of 2017.

Here are some videos that demonstrate a subset of the new functionality:

**KVM enhancements regarding managed storage:**

https://www.youtube.com/watch?v=-tExEvkMJYs&index=19&list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&t=2508s

**VMware enhancements regarding managed storage:**

https://www.youtube.com/watch?v=UbiAPP9nyEM&index=20&list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&t=538s

**KVM volume migration (NFS or Ceph to managed storage):**

https://www.youtube.com/watch?v=Oy4IsTzqN1c&t=27s&list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&index=21

**Online storage migration in KVM (NFS or Ceph to managed storage):**

https://www.youtube.com/watch?v=VbnMwYYkRnQ&t=342s&list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&index=22

**Extract contents of managed-storage volume on KVM to QCOW2 file:**

https://www.youtube.com/watch?v=mW3hZLKymsI&t=3s&list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&index=23

**Upload contents of QCOW2 file to managed-storage volume on KVM:**

https://www.youtube.com/watch?v=_c3OSluMkxc&t=164s&list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&index=24

mike-tutkowski:
@rhtyd Once we do have the requisite number of LGTMs, I'd like to re-run all of the managed-storage automated tests (existing and new ones) to verify they all still pass (since I have made changes to the code per reviewer requests). The automated tests will probably take me a couple days to execute.

rafaelweingartner:
@mike-tutkowski my comment was under one of Daan’s where the validations are inverted in class “StorageSystemDataMotionStrategy.java”. Here follows what I said there:

> Documentation could help. However, I would first suggesting extracting lines 308-327 to a method. Then, instead of starting with if (canHandleSrc && canHandleDest), you can do the following:
>
> if (!canHandleSrc){Throw exception regarding source not being able to handle the request}
> if (!canHandleDest){Throw exception regarding destination not being able to handle the request }
> 
> Then, if you pass those two IFs, you do not need a third one to check if (canHandleSrc && canHandleDest)
>

mike-tutkowski:
@rafaelweingartner I went ahead and updated it to the following:

    private void handleCopyAsyncForSnapshotToVolume(boolean canHandleSrc, SnapshotInfo srcSnapshotInfo, VolumeInfo destVolumeInfo,
                                                    AsyncCompletionCallback<CopyCommandResult> callback) {
        boolean canHandleDest = canHandle(destVolumeInfo);

        if (canHandleSrc && canHandleDest) {
            if (srcSnapshotInfo.getDataStore().getId() == destVolumeInfo.getDataStore().getId()) {
                handleCreateVolumeFromSnapshotBothOnStorageSystem(srcSnapshotInfo, destVolumeInfo, callback);
            } else {
                String errMsg = "To perform this operation, the source and destination primary storages must be the same.";

                handleError(errMsg, callback);
            }
        } else if (canHandleDest) {
            handleCreateVolumeFromSnapshotOnSecondaryStorage(srcSnapshotInfo, destVolumeInfo, callback);
        }
        else {
            handleError(OPERATION_NOT_SUPPORTED, callback);
        }
    }

Basically, the code needs to throw an exception in two cases: [if both canHandleSrc and canHandleDest are false] or [if canHandleSrc is true, but canHandleDest is false].

rafaelweingartner:
@mike-tutkowski thanks for the effort. I would still prefer a cleaner version though.

This block of code is the case that it is better to start with the conditions that generate exceptions `!canHandleDest` and `!canHandleSrc`, and then you can create the normal flow without requiring to add it inside a conditional. 

Anyways as I said here previously, I understand if you guys do not have time and energy to improve this now. However, we should start paying more attention to these legacy codes that sometimes are badly structured.

mike-tutkowski:
@rafaelweingartner Sorry, I've probably been staring at code "too long" this afternoon and missed what you were saying at first. You probably mean something like this:

    private void handleCopyAsyncForSnapshotToVolume(SnapshotInfo srcSnapshotInfo, VolumeInfo destVolumeInfo,
                                                    AsyncCompletionCallback<CopyCommandResult> callback) {
        boolean canHandleDest = canHandle(destVolumeInfo);

        if (!canHandleDest) {
            handleError(OPERATION_NOT_SUPPORTED, callback);
        }

        boolean canHandleSrc = canHandle(srcSnapshotInfo);

        if (canHandleSrc) {
            if (srcSnapshotInfo.getDataStore().getId() == destVolumeInfo.getDataStore().getId()) {
                handleCreateVolumeFromSnapshotBothOnStorageSystem(srcSnapshotInfo, destVolumeInfo, callback);
            } else {
                String errMsg = "To perform this operation, the source and destination primary storages must be the same.";

                handleError(errMsg, callback);
            }
        } else {
            handleCreateVolumeFromSnapshotOnSecondaryStorage(srcSnapshotInfo, destVolumeInfo, callback);
        }
    }

rafaelweingartner:
That happens :)
Yes, that is what I meant. Actually, you can even go one step further to take this code out of the conditional:
```
 if (srcSnapshotInfo.getDataStore().getId() == destVolumeInfo.getDataStore().getId()) {
            handleCreateVolumeFromSnapshotBothOnStorageSystem(srcSnapshotInfo, destVolumeInfo, callback);
        } else {
            String errMsg = "To perform this operation, the source and destination primary storages must be the same.";

            handleError(errMsg, callback);
        }
```
you can use `if(!canHandleSrc)`, then throw the appropriated exception. 
Then, you have your nice chunk of happy day (everything working and handling things as they should) code out of IFs

mike-tutkowski:
@rafaelweingartner I might be a be confused on that suggestion. If you !canHandleSrc, then you want to invoke handleCreateVolumeFromSnapshotOnSecondaryStorage.

canHandleSrc being false just means a non-managed-storage driver is responsible for that primary storage. We can still use it, but it has to be handled a special way.

mike-tutkowski:
I've added the following comment to the canHandle method:

    // canHandle returns true if the storage driver for the DataObject that's passed in can support certain features (what features we
    // care about during a particular invocation of this method depend on what type of DataObject was passed in (ex. VolumeInfo versus SnapshotInfo).

Perhaps the method name is misleading. It might be interpreted to mean we can't do anything with this object if canHandle returns false. In reality, if canHandle returns true, we know you support certain features and we handle you a particular way. If it returns false, we might throw an exception or we might be able to handle you another way.

mike-tutkowski:
As an example:

        boolean canHandleSrc = canHandle(srcSnapshotInfo);

        if (canHandleSrc) {
            if (srcSnapshotInfo.getDataStore().getId() == destVolumeInfo.getDataStore().getId()) {
                handleCreateVolumeFromSnapshotBothOnStorageSystem(srcSnapshotInfo, destVolumeInfo, callback);
            } else {
                String errMsg = "To perform this operation, the source and destination primary storages must be the same.";

                handleError(errMsg, callback);
            }

In this case, canHandle returns true because srcSnapshotInfo's storage driver does support certain features for snapshots. However, the canHandle method doesn't know if the destination volume (that we want to copy our snapshot to) is in the same primary storage or not. As it turns out, right now we only support this operation if the two objects (the snapshot and volume) are in the same storage pool. It might come down to the naming of canHandle is a bit vague.

rafaelweingartner:
I understood that. It is good the way you did as well. What I meant previously was the following.
You have this block of code:
```
  if (canHandleSrc) {
        if (srcSnapshotInfo.getDataStore().getId() == destVolumeInfo.getDataStore().getId()) {
            handleCreateVolumeFromSnapshotBothOnStorageSystem(srcSnapshotInfo, destVolumeInfo, callback);
        } else {
            String errMsg = "To perform this operation, the source and destination primary storages must be the same.";

            handleError(errMsg, callback);
        }
    } else {
        handleCreateVolumeFromSnapshotOnSecondaryStorage(srcSnapshotInfo, destVolumeInfo, callback);
    }
```

This previous block is the same as:
```
  if (!canHandleSrc) {
        handleCreateVolumeFromSnapshotOnSecondaryStorage(srcSnapshotInfo, destVolumeInfo, callback);
    }

 if (srcSnapshotInfo.getDataStore().getId() == destVolumeInfo.getDataStore().getId()) {
            handleCreateVolumeFromSnapshotBothOnStorageSystem(srcSnapshotInfo, destVolumeInfo, callback);
        } else {
            String errMsg = "To perform this operation, the source and destination primary storages must be the same.";

            handleError(errMsg, callback);
        }
```

Sorry the bad indentation here. So, the idea was to remove one IF inside the other. At least for me, the least nested conditions and operations, the better to read the code.

mike-tutkowski:
Sure, @rafaelweingartner. If you feel that's easier to read, I can write it that way. I'll send a new commit up soon. Thanks for your time!

mike-tutkowski:
Sure, @rafaelweingartner. If you feel that's easier to read, I can write it that way. I'll send a new commit up soon. Thanks for your time!

rafaelweingartner:
@mike-tutkowski thanks for your hard work!

mike-tutkowski:
@rafaelweingartner Oh, it's no problem. There is so much code here to implement all of these new features and it all has to work alongside the old code, so sometimes it's easier to change as little as possible from the old code to "assure" yourself you haven't broken any of it along the way. Of course, as you do this, the new code placed on top of the old code ends up adding complexity. At some point, it's best to "suck it up," sit down, and rewrite a bunch of it to support the entire feature set, but be less complicated. To do this, though, it's nice to have a large set of automated tests to catch anything you might have broken in the code during the rewrite. I do have a lot of new tests to catch issues in the new code, but I've never been super confident in CloudStack's automated tests related to its legacy features.

server/src/com/cloud/configuration/Config.java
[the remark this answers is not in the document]
mike-tutkowski: @DaanHoogland I changed the KvmStorageOfflineMigrationWait and KvmStorageOnlineMigrationWait keys from using the Config approach to using the ConfigKey approach.

DaanHoogland:
I do not trust travis, and jenkins has disk problems. I restarted the travis run  that failed just to see. If it succeeds I'll start 'our' integrations as well.

mike-tutkowski:
Yes, Travis seems to (apparently) randomly fail. Sometimes I’ve made trivial changes (like change comments), forced an update, and Travis ended up failing the new commit where it was OK with the prior commit.

On Jan 6, 2018, at 1:46 AM, dahn <notifications@github.com<mailto:notifications@github.com>> wrote:


I do not trust travis, and jenkins has disk problems. I restarted the travis run that failed just to see. If it succeeds I'll start 'our' integrations as well.

rhtyd:
Travis is green. @DaanHoogland are you lgtm on it?
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland:
@rhtyd if the tests come back alright I am alright with it but
@mike-tutkowski I plead that you have a look at simplifying the code and extract as much as possible into separate methods/helper classes on your next iteration.

rhtyd:
@mike-tutkowski can you fix runtime issue, management server fails to start with: /cc @DaanHoogland 
```
2018-01-07 18:27:32,548 ERROR [o.a.c.s.m.w.CloudStackContextLoaderListener] (main:null) (logid:) Failed to start CloudStack
java.io.IOException: Resource [jar:file:/usr/share/cloudstack-management/lib/cloudstack-4.11.0.0-SNAPSHOT.jar!/META-INF/cloudstack/sioc/module.properties] is expected to exist at [classpath:META-INF/cloudstack/vmware-sioc/module.properties] please ensure the name property is correct
        at org.apache.cloudstack.spring.module.model.impl.DefaultModuleDefinition.checkNameMatchesSelf(DefaultModuleDefinition.java:108)

```
Due to this issue and without a fix, we can not further test/review this.

DaanHoogland:
closing to trigger jenkins again

DaanHoogland:
jenkins do your thing, please.

mike-tutkowski:
That looks like a silly config-file mistake. I made a change to that file due to reviewer comments and didn’t re-run the management server. Once I’m back at my computer today, I can fix that and let you know when that’s done.

On Jan 7, 2018, at 3:39 AM, Rohit Yadav <notifications@github.com<mailto:notifications@github.com>> wrote:


@mike-tutkowski<https://github.com/mike-tutkowski> can you fix runtime issue, management server fails to start with: /cc @DaanHoogland<https://github.com/daanhoogland>

2018-01-07 18:27:32,548 ERROR [o.a.c.s.m.w.CloudStackContextLoaderListener] (main:null) (logid:) Failed to start CloudStack
java.io.IOException: Resource [jar:file:/usr/share/cloudstack-management/lib/cloudstack-4.11.0.0-SNAPSHOT.jar!/META-INF/cloudstack/sioc/module.properties] is expected to exist at [classpath:META-INF/cloudstack/vmware-sioc/module.properties] please ensure the name property is correct
        at org.apache.cloudstack.spring.module.model.impl.DefaultModuleDefinition.checkNameMatchesSelf(DefaultModuleDefinition.java:108)



Due to this issue and without a fix, we can not further test/review this.

rhtyd:
Thanks @mike-tutkowski

mike-tutkowski:
@rhtyd I needed to rename a folder from "sioc" to "vmware-sioc" and now the management server starts just fine. I've updated the PR with the new commit. Thanks!

mike-tutkowski:
@rhtyd Once we get a successful run of (standard) tests back, I can then re-run all of the managed-storage regression tests.

rhtyd:
@blueorangutan package

rhtyd:
Travis keeps failing, please check @mike-tutkowski

rhtyd:
@blueorangutan test

mike-tutkowski:
Hi @rhtyd: I notice that the following PRs from our 4.11 milestone list are also failing on the same Travis test as mine: 2128, 2146, 2362, 2389, 2393. Perhaps we have a problem with that test?

rhtyd:
Thanks @mike-tutkowski I've created a PR #2394 to fix Travis failures. I've kicked a test run in the background now. Meanwhile, can you run solidfire tests against your PR and ensure that they are all passing, and share results as well?

mike-tutkowski:
@rhtyd Has something changed with regards to how we need to set up a new cloud? Neither my SSVM nor my CPVM was successful at starting up its agent.

From cloud.log in SSVM, I see the following:

Mon Jan  8 20:52:24 UTC 2018 Incompleted parameters STORAGE_IP:, STORAGE_NETMASK:, STORAGE_CIDR:. Cannot setup storage network
Mon Jan  8 20:52:24 UTC 2018 Not setting up route of RFC1918 space to 10.117.40.126 befause 10.117.40.27 is RFC1918.

Perhaps there is a new Global Setting I need to configure?

mike-tutkowski:
@rhtyd This is the system VM template that's in use: https://builds.cloudstack.org/view/master/job/build-master-systemvm/lastSuccessfulBuild/artifact/tools/appliance/dist/systemvm64template-master-4.11.0.440-xen.vhd.bz2

mike-tutkowski:
@rhtyd I also should note that this is a Basic Zone (I think a lot of our automated tests (maybe all) use an Advanced Zone).

rhtyd:
@mike-tutkowski can you try the new debian9 systemvmtemplates, get from http://dl.rohityadav.cloud/debian9/ or build yourself using qemu+packer using tools/appliance/build.sh

mike-tutkowski:
@rhtyd Using your system VM templates fixed it - thanks!

rhtyd:
@mike-tutkowski can you look at several of the failures, many could be env related. Also fix the conflict?

mike-tutkowski:
@rhtyd I was looking at those tests when I saw the failures came in. It seems pretty unlikely that any of this code would be triggering errors in areas like VPC, affinity groups, etc. Maybe we should see if we are getting similar errors on other PRs?

I can fix the file conflict soon. Thanks!

rhtyd:
Thanks @mike-tutkowski I'll rekick tests, meanwhile continue and share solidfire related regression testing.
@blueorangutan package

rhtyd:
@blueorangutan test matrix

mike-tutkowski:
@rhtyd Just so I get a feeling for the test failures with regards to other PRs: Are other PRs failing with issues that don't seem directly related to them?

With regards to my progress running the managed-storage tests, I'm about 1/3 of the way through.

So far, I only see one failure. It is possibly related to commit 1d0f212. The problem concerns SAN-assisted snapshots. This is when we take a SAN snapshot, but we don't want to keep the snapshot on the SAN, so we migrate it to secondary storage (the idea here is to minimize the impact on the source volume that we are taking a volume snapshot on by not copying from that volume to secondary storage, but rather going through a SAN snapshot).

In any event, we can't create a template from this kind of snapshot anymore because we can't find a hypervisor host to perform the copy from secondary to primary storage.

I have this noted and plan to investigate it soon. In the meanwhile, I've just continued running the tests and all has been well with them.

Once I have all of the tests run, I can post results.

rhtyd:
Hi @mike-tutkowski I've tried to fix some failures, you can compare failures against baseline master branch from #2376 (I'll kick another round tonight). Given high failure rate and that we're already past the cut/freeze, do you think the PR would be stable and ready for merge?

rhtyd:
@mike-tutkowski please fix the conflicts

mike-tutkowski:
@rhtyd So far for me, I'm only having trouble with one managed-storage test. I'm investigating it. As for the general failing tests, I'm confused why things like stopping the CPVM, listing items on secondary storage, etc. would be impacted by anything in this PR. I should be able to upload the results of my managed-storage tests this evening. Thanks!

mike-tutkowski:
@rhtyd @rafaelweingartner I have located the problem with my one integration test. The issue stems from a bit of code in https://github.com/apache/cloudstack/pull/2315.

I went ahead and put comments in that PR under the applicable code. Here is the gist:

HostDaoImpl.java:
 +        long snapshotId = srcData.getId();
 +        StoragePoolVO storagePoolVO = storagePoolDao.findStoragePoolForSnapshot(snapshotId);
      @mike-tutkowski
mike-tutkowski Pending
This is not finding a connected host in the managed-storage situation.

I've examined the cloud.storage_pool_host_ref table and there is, in fact, a reference to every host in my zone and my primary storage.

By the way, what would happen if I removed the primary storage in question (managed or non managed) after having created a volume snapshot from a volume on it? It seems like this logic wouldn't work because the storage pool won't be connected to any hosts.

com.cloud.utils.exception.CloudRuntimeException: Could not find a host connected to the storage pool [storagepool=4].
at com.cloud.host.dao.HostDaoImpl.findHostToOperateOnSnapshotBasedOnStoragePool(HostDaoImpl.java:1189)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:338)
at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:197)
at com.cloud.utils.db.TransactionContextInterceptor.invoke(TransactionContextInterceptor.java:34)
at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:174)
at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:92)
at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212)
at com.sun.proxy.$Proxy86.findHostToOperateOnSnapshotBasedOnStoragePool(Unknown Source)
at com.cloud.hypervisor.XenServerGuru.getCommandHostDelegation(XenServerGuru.java:207)
at com.cloud.hypervisor.HypervisorGuruManagerImpl.getGuruProcessedCommandTargetHost(HypervisorGuruManagerImpl.java:76)
at org.apache.cloudstack.storage.RemoteHostEndPoint.sendMessage(RemoteHostEndPoint.java:120)
at org.apache.cloudstack.storage.motion.AncientDataMotionStrategy.createTemplateFromSnapshot(AncientDataMotionStrategy.java:507)
at org.apache.cloudstack.storage.motion.AncientDataMotionStrategy.copyAsync(AncientDataMotionStrategy.java:451)
at org.apache.cloudstack.storage.motion.DataMotionServiceImpl.copyAsync(DataMotionServiceImpl.java:68)
at org.apache.cloudstack.storage.motion.DataMotionServiceImpl.copyAsync(DataMotionServiceImpl.java:73)
at org.apache.cloudstack.storage.image.TemplateServiceImpl.copyAsync(TemplateServiceImpl.java:732)
at org.apache.cloudstack.storage.image.TemplateServiceImpl.createTemplateFromSnapshotAsync(TemplateServiceImpl.java:738)
at com.cloud.template.TemplateManagerImpl.createPrivateTemplate(TemplateManagerImpl.java:1591)
at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:338)
at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:197)
at org.apache.cloudstack.network.contrail.management.EventUtils$EventInterceptor.invoke(EventUtils.java:107)
at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:174)
at com.cloud.event.ActionEventInterceptor.invoke(ActionEventInterceptor.java:51)
at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:174)
at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:92)
at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212)
at com.sun.proxy.$Proxy194.createPrivateTemplate(Unknown Source)
at org.apache.cloudstack.api.command.admin.template.CreateTemplateCmdByAdmin.execute(CreateTemplateCmdByAdmin.java:43)
at com.cloud.api.ApiDispatcher.dispatch(ApiDispatcher.java:150)
at com.cloud.api.ApiAsyncJobDispatcher.runJob(ApiAsyncJobDispatcher.java:108)
at org.apache.cloudstack.framework.jobs.impl.AsyncJobManagerImpl$5.runInContext(AsyncJobManagerImpl.java:581)
at org.apache.cloudstack.framework.jobs.impl.AsyncJobManagerImpl$5.run(AsyncJobManagerImpl.java:529)
at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
at java.util.concurrent.FutureTask.run(FutureTask.java:266)
at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
at java.lang.Thread.run(Thread.java:745)

mike-tutkowski:
@rhtyd @rafaelweingartner Let me provide a bit more background on this use case.

1) This only applies to XenServer (for the time being).

2) Normally when you create a volume snapshot on managed storage, it is simply stored on primary storage as a SAN snapshot.

However, there is a second option when using managed storage with XenServer: You can temporarily create the volume snapshot as a SAN snapshot and then the SAN snapshot gets copied to secondary storage to remain on secondary storage like any other volume snapshot. This is called a SAN-assisted snapshot. The purpose is so we are not copying data from the virtual disk to secondary storage and having the host write new data to it simultaneously as that can lead to performance issues. Instead, we take a SAN snapshot of the volume housing the virtual disk. We then mount the SAN snapshot and copy they hypervisor snapshot to secondary storage. The original VDI's performance is therefore not impacted during this process, but we still end up with a volume snapshot (albeit a full copy) on secondary storage.

Now, let's say you then go to create a template from this volume snapshot. As we can see in my previous comment, the logic from that PR can't find a host to perform the copy from secondary storage to primary storage and, as such, the operation fails.

I've looked in the cloud.storage_pool_host_ref table and there is, in fact, a reference to every host in my zone and my primary storage.

Aside from this concern, I wonder what would happen (for managed or non-managed) storage if the storage pool the volume snapshot originated from was later removed. I think we would fail to find a host in that case, as well.

Also, looking at HostDaoImpl.createSqlFindHostConnectedToStoragePoolToExecuteCommand, I see code that is not correct:

        if (storagePoolVO.isManaged()) {
            joinForManagedStorage = " join cluster_details cd on cd.cluster_id = c.id and cd.name = 'supportsResign' and cd.value = 'true' ";
        }

It is not necessary to meet this restriction in this particular case. We only need to check this value if we want to perform a UUID/VDI resignature operation on XenServer (which we don't here).

rafaelweingartner:
@mike-tutkowski I really do not understand much of managed storage in ACS. That join was added after some discussions. So, let's see if I understood what you are saying... we do not need that extra join in case of managed storage, is that it?

mike-tutkowski:
@rafaelweingartner It looks like the method that contains that code (createSqlFindHostConnectedToStoragePoolToExecuteCommand) is only called from HostDaoImpl.findHostToOperateOnSnapshotBasedOnStoragePool.

In this case, we are just trying to find a host to perform a copy off the volume snapshot from secondary to a new template on secondary. As such, you shouldn't really care for managed storage if the compute cluster meets the restrictions of your cluster_details clause (no UUID resigning is required here).

Even more concerning to me, though, is what happens if the storage pool (managed or non managed) no longer exists? For example, let's say you created a volume snapshot from a volume in storage pool 1 and later you removed storage pool 1. You then go to create a template from this volume snapshot, but you can't find a connected host because storage pool 1 no longer exists (it shouldn't have any connected hosts listed in cloud.storage_pool_host_ref).

rafaelweingartner:
Hmm... as far as I remember we are not filtering the storage_pool table by `removed = null`, so it will use the `cluster_id` of the `storage_pool`, which should be there to find the hosts where this storage pool was connected to.

rafaelweingartner:
Oh.. damm it... I just tested..
The method `org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDaoImpl.findStoragePoolForSnapshot(long)` will return null because of `findById` only returns entries that have `removed = null`

I can open a PR to fix it, but can you guide me in respect to that extra join. Do we need it? or we can just ignore it?

mike-tutkowski:
@rafaelweingartner Sure, I'm happy to help.

A couple things:

1) If you are just looking for a host to perform the operation (in my use case, it's a copy of a volume snapshot that's on secondary storage to a new template, which is also, as always, on secondary storage), then you don't need that join on the cluster_details table in HostDaoImpl.createSqlFindHostConnectedToStoragePoolToExecuteCommand. We only need to check that field in cluster_details when we'd like to perform a UUID resignature of an SR and/or a VDI on XenServer, which we don't here.

2) I'm curious why we don't just do the following: Look up the hypervisor type of the volume snapshot in cloud.snapshots. Find any host of that hypervisor type in the applicable zone and use it to perform the operation. This would solve the problem of having to find a storage pool that may no longer be in use. In any event, once the volume snapshot is on secondary storage, we don't really care what storage pool it came from, right? I think we just care what hypervisor type it came from.

rafaelweingartner:
Reviewing the comments there, there is something you said:

> On XenServer, making use of a volume snapshot that resides on primary storage needs to leverage a XenServer host that supports UUID resigning

I believe that it is because of these cases where snapshots is in primary storage (is that a thing?). We might have not understood each other there at that moment.

So, can I simply get any host of the environment that is of the same type as the snapshot to execute the command?

mike-tutkowski:
@rafaelweingartner Yeah, sorry if I wasn't clear about that back when we were looking at your PR.

For XenServer + managed storage, there are two approaches to taking volume snapshots:

1) The volume snapshot resides on primary storage.

2) The volume snapshot goes through a SAN snapshot and ends up on secondary storage.

In use case one, a snapshot is created on the SAN (fast and space efficient, but it's not technically a backup). To make use of it (i.e. to create a volume from a volume snapshot or a template from a volume snapshot) on XenServer, we can leverage UUID resigning (if the applicable XenServer service pack is installed). If UUID resigning is not available (because the service pack isn't installed), then we just perform a copy of the snapshot when creating a volume or a template from the snapshot. XenServer has a UUID for each SR and each VDI. You cannot mount two SRs or VDIs with the same UUID at the same time. If you're interested, here's a video of mine demoing this feature:

https://www.youtube.com/watch?v=YQ3pBeL-WaA&index=13&list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&t=1362s

In any event, use case 2 allows you to temporarily use a SAN snapshot, but end up with a standard volume snapshot on secondary storage when all is said and done. It is this use case that is failing.

I believe if you just look at the hypervisor type of the volume snapshot and pick any host in the applicable zone of the volume snapshot to perform the operation that all should be well.

mike-tutkowski:
@rafaelweingartner What I noted above should solve not only this problem, but also the more general problem of what would happen if the storage pool in question were removed after a volume snapshot was created from it and you wanted to use that volume snapshot.

rafaelweingartner:
hmm.. but, if I get a random host of the same hypervisor in the zone, will not it break your case 1 (sometimes)?

mike-tutkowski:
@rafaelweingartner To be clear, when the volume snapshot is stored on managed storage, the StorageSystemDataMotionStrategy handles the operation when you want to create a volume or a template from the volume snapshot (so you don't have to worry about that use case).

rafaelweingartner:
Ah, now it is clear....
Tomorrow I will open a PR with these changes, and will ping you then

mike-tutkowski:
@rafaelweingartner That sounds good. Just an FYI that I will be traveling and won't be able to check my e-mail until late afternoon Eastern time. Thanks!

mike-tutkowski:
@rhtyd Managed-storage test results:

TestAddRemoveHosts:
[results.txt](https://github.com/apache/cloudstack/files/1620564/results.txt)

TestManagedSystemVMs:
[results.txt](https://github.com/apache/cloudstack/files/1620565/results.txt)

TestVMMigrationWithStorage:
[results.txt](https://github.com/apache/cloudstack/files/1620567/results.txt)

TestVMSnapshots:
[results.txt](https://github.com/apache/cloudstack/files/1620568/results.txt)

TestVolumes (no UUID resigning):
[results.txt](https://github.com/apache/cloudstack/files/1620569/results.txt)

TestVolumes (UUID resigning):
[results.txt](https://github.com/apache/cloudstack/files/1620570/results.txt)

TestSnapshots (Tests 3 and 4 failed, more on that later):
[results.txt](https://github.com/apache/cloudstack/files/1620573/results.txt)

TestSnapshots (Test 3 passed after fixing an environment issue that caused it to fail previously):
[results.txt](https://github.com/apache/cloudstack/files/1620574/results.txt)

The only test that continue to fail is 4 from TestSnapshots. @rafaelweingartner is working on a fix in a separate PR that should address this issue.

mike-tutkowski:
@rhtyd I see the following Travis test failed (not sure why HA should be impacted by anything in this PR):

test_ha_configure_enabledis | Failure | 603.629 | test_hostha_simulator      |
| able_across_clusterzones    |         |         |                            |

mike-tutkowski:
@rhtyd The weird part is that Travis was green until I manually fixed a merge conflict in VmwareResource.java (most of which was pretty trivial to fix).

mike-tutkowski:
Since we seem to have decided to put https://github.com/apache/cloudstack/pull/2402 into a maintenance release for 4.11, I can report that all of my managed-storage integration tests have successfully passed with the one exception that will be corrected when https://github.com/apache/cloudstack/pull/2402 goes in. That particular managed-storage feature (the one who's test fails due to https://github.com/apache/cloudstack/pull/2402) is currently only being used by one customer and this customer uses a private branch of CloudStack, so I can always provide them the fix, if need be.

mike-tutkowski:
@rhtyd Please let me know if I should be investigating any of those test failures of if you believe they are inapplicable. Many/most seem pretty random to be related to this PR. I would normally be OK with holding off on a PR until the next release, but - in this case - 4.11 is an LTS and I have several customers interested in it due to that characteristic. Thanks!

mike-tutkowski:
@rhtyd Just following up on Travis: I see https://github.com/apache/cloudstack/pull/2402 has the same failure as this PR, which probably indicates a problem with that test since the code in the two PRs is quite different.

rhtyd:
Mike see the smoketests from #2376 for baseline master failures. It's too risky to accept big feature  prs where we're seeing a lot failures. You can appreciate I've been engaging with all the PR authors, including extending the freeze/cut date. If we have consensus to further delay the release l by waiting for outstanding 4.11 milestone  PRs I don't have a problem.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

mike-tutkowski:
@rhtyd I will wait for the tests to complete and then I can compare any failures against #2376.

Since it looks like the Travis failure is a common one (same problem for #2402, for example), I won't worry about that one.

mike-tutkowski:
@rhtyd Am I missing something or are there really dozens and dozens of failures/errors being reported from tests for #2376?

rhtyd:
@mike-tutkowski ignore travis failure for now (known and not caused by your PR, it intermittently passes and I'll look into it in a separate PR). I see some new failures caused by this PR, please verify and advise:

XenServer:
test_02_edit_template | Failure | 90.10 | test_templates.py
test_change_service_offering_for_vm_with_snapshots | Failure | 293.60 | test_vm_snapshots.py

I'll update this when kvm, vmware tests are back.

rhtyd:
@mike-tutkowski see ONLY the smoketest failures on #2376 at the very end. Smoke tests failures are the only which mentions they are smoketests (about 66-67 tests).

rhtyd:
@mike-tutkowski this is the only outstanding feature PR towards the 4.11 milestone. If we can demonstrate that no new smoketest regressions are caused by this PR, we can accept this PR. You can reference the latest results on master from "smoketest" results on PR #2376 .

Support for reviewing, testing appreciated @rafaelweingartner @DaanHoogland @wido and others

core/src/com/cloud/agent/api/MigrateCommand.java
```diff
@@ -40,6 +45,26 @@ public MigrateCommand(String vmName, String destIp, boolean isWindows, VirtualMa
...
+    public void setMigrateStorage(Map<String, MigrateDiskInfo> migrateStorage) {
+        this.migrateStorage = migrateStorage;
+    }
+
+    public Map<String, MigrateDiskInfo> getMigrateStorage() {
+        return migrateStorage != null ? new HashMap<>(migrateStorage) : new HashMap<String, MigrateDiskInfo>();
+    }
+
+    public boolean isMigrateStorage() {
+        return migrateStorage != null && !migrateStorage.isEmpty();
```
rafaelweingartner: You can use MapUtils.isNotEmpty here.
BTW, `migrateStorage` is a quite misleading variable name. It feels boolean, but it is a map. Can you  change it to something more descriptive?

core/src/com/cloud/agent/api/MigrateCommand.java
```diff
@@ -68,4 +93,79 @@ public String getHostGuid() {
...
         return executeInSequence;
     }
+
+    public static class MigrateDiskInfo {
+        public enum DiskType {
+            FILE, BLOCK;
+
+            @Override
+            public String toString() {
+                switch(this) {
```
rafaelweingartner: This does not make much sense. It is an enum, you will not be able to "instantiate" anything different from your constants. You can simply return this.name.toLowerCase, if you only want the name of the enum in lower case

core/src/com/cloud/agent/api/MigrateCommand.java
```diff
@@ -68,4 +93,79 @@ public String getHostGuid() {
...
+                }
+            }
+        }
+
+        public enum DriverType {
+            QCOW2, RAW;
+
+            @Override
+            public String toString() {
+                switch(this) {
```
rafaelweingartner: The same as before here

core/src/com/cloud/agent/api/MigrateCommand.java
```diff
@@ -68,4 +93,79 @@ public String getHostGuid() {
...
+                    case QCOW2: return "qcow2";
+                    case RAW: return "raw";
+                    default: throw new IllegalArgumentException();
+                }
+            }
+        }
+
+        public enum Source {
+            FILE, DEV;
```
rafaelweingartner: also here

core/src/com/cloud/agent/api/StartAnswer.java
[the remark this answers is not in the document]
rafaelweingartner: What is this `iqnToData`? What is the key of the first map? And the key of the second map?

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -563,6 +567,10 @@ protected void advanceExpunge(VMInstanceVO vm) throws ResourceUnavailableExcepti
         // Clean up volumes based on the vm's instance id
         volumeMgr.cleanupVolumes(vm.getId());
 
+        if (hostId != null && targets != null && targets.size() > 0) {
```
rafaelweingartner: You can use `hostId != null && CollectionUtils.isNotEmpty(targets)`.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -599,6 +607,60 @@ protected void advanceExpunge(VMInstanceVO vm) throws ResourceUnavailableExcepti
...
 
+    private List<Map<String, String>> getTargets(Long hostId, long vmId) {
+        List<Map<String, String>> targets = new ArrayList<>();
+
+        HostVO hostVO = _hostDao.findById(hostId);
+
+        if (hostVO != null && hostVO.getHypervisorType() == HypervisorType.VMware) {
+            List<VolumeVO> volumes = _volsDao.findByInstance(vmId);
+
+            if (volumes != null) {
```
rafaelweingartner: what if you invert this one here?
`if (CollectionUtils.isEmpty(volumes))`, then you return targets.
This would allow to remove the other for and IF statement from one conditional.

Rule of thumb: the less nested things, the better to read.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -1331,10 +1414,35 @@ public boolean getExecuteInSequence(final HypervisorType hypervisorType) {
         }
     }
 
+    private List<Map<String, String>> getVolumesToDisconnect(VirtualMachine vm) {
+        List<Map<String, String>> volumesToDisconnect = new ArrayList<>();
+
+        List<VolumeVO> volumes = _volsDao.findByInstance(vm.getId());
+
+        if (volumes != null) {
```
rafaelweingartner: Here you can also do the same, if volumes is empty, you can return right away the empty `volumesToDisconnect`, thus removing the next operations from a IF statement

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -179,123 +242,320 @@ private boolean canHandle(DataObject dataObject) {
...
+                hostVO = getHostInCluster(srcStoragePoolVO.getClusterId());
+            }
+            else {
+                hostVO = getHost(srcVolumeInfo.getDataCenterId(), hypervisorType, false);
+            }
+
+            volumePath = copyVolumeToSecondaryStorage(srcVolumeInfo, destVolumeInfo, hostVO,
+                    "Unable to copy the volume from managed storage to secondary storage");
+        }
+        catch (Exception ex) {
```
rafaelweingartner: Do you need this generic exception here?

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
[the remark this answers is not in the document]
rafaelweingartner: I like the idea of switching from checked to unchecked exception, but I would prefer using  throw new CloudRuntimeException(ex), because this would give me the full stack to the cause of the problem. 

If we re-throw an exception, we do not need to log it right away...

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -916,19 +1987,45 @@ else if (dataObj instanceof SnapshotInfo) {
...
+        ResignatureAnswer answer;
+
+        GlobalLock lock = GlobalLock.getInternLock(dataStore.getUuid());
+
+        if (!lock.lock(LOCK_TIME_IN_SECONDS)) {
+            String errMsg = "Couldn't lock the DB (in performResignature) on the following string: " + dataStore.getUuid();
+
+            LOGGER.warn(errMsg);
+
+            throw new CloudRuntimeException(errMsg);
```
rafaelweingartner: This is not specific to this PR, but why are we logging and throwing exceptions?  Isn't throwing enough (as long as someone does not hide the exception)?

mike-tutkowski:
@rhtyd I will implement @rafaelweingartner's code suggestions and then take a look at the handful of tests that are listed as failed. Thanks!

core/src/com/cloud/agent/api/MigrateCommand.java
mike-tutkowski: Thanks for catching this. I was initially returning something different than the simple lowercase version of the name of the enum, but then I went back to that simple approach and didn't change the code to reflect that simplicity.

core/src/com/cloud/agent/api/MigrateCommand.java
mike-tutkowski: Same as above

core/src/com/cloud/agent/api/MigrateCommand.java
mike-tutkowski: Same as above

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
mike-tutkowski: Done

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
mike-tutkowski: Done. My initial approach here was due to the fact that some developers/reviewers don't like multiple return points in a method. Personally, I'm fine with what you suggest here and have updated the code.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
mike-tutkowski: Same as above

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
mike-tutkowski: The KVM migration can occasionally return an exception that extends RuntimeException. I wanted to lean on the safe side: Catch any non-system exception here.

engine/storage/datamotion/src/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
mike-tutkowski: There are a couple reasons I believe CloudStack takes this approach:

1) Code was (every now and then) swallowing exceptions, so this at least provides logging at the point where we convert from a checked exception to a non-checked exception.

2) In some cases, the message of the exception is presented to the end user and putting in a more readable string can help people. I went ahead and changed the constructor that's used in several places to be CloudRuntimeException(String message, Throwable t).

rafaelweingartner: Got it. I thought that was the reason, but we should strive to force people not to swallow exception anymore.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
rafaelweingartner: Got it. I have seen reviewers also saying that. I personally disagree.... specially in method where we might have few conditionals, and by doing this check and return, we can avoid nesting them.

Thanks for your work!

rhtyd:
@mike-tutkowski thanks, please also fix any conflicts? (fyi I fixed conflict this morning after merging config drive PR on your PR). After fixing ping me, I can kick another test round.

We now have 0 smoketest failures on KVM, VMware and XenServer with Trillian and 0 failures with Travis with this PR: https://github.com/apache/cloudstack/pull/2403

I assessed the failures, you'll need to look at following failures: (it's possible these failures are not caused by your PR but environment related)

XenServer:
test_change_service_offering_for_vm_with_snapshots | Failure | 293.60 | test_vm_snapshots.py

KVM:
test_router_dhcphosts | Failure | 44.53 | test_router_dhcphosts.py
ContextSuite context=TestRouterDHCPHosts>:teardown | Error | 73.03 | test_router_dhcphosts.py
test_01_list_sec_storage_vm | Failure | 0.07 | test_ssvm.py
test_05_stop_ssvm | Failure | 62.65 | test_ssvm.py
test_03_deploy_vm_wrong_checksum | Error | 20.55 | test_templates.py
ContextSuite context=TestCreateTemplateWithDirectDownload>:teardown | Error | 32.88 | test_templates.py
test_04_extract_template | Failure | 128.27 | test_templates.py
ContextSuite context=TestISOUsage>:setup | Error | 0.00 | test_usage.py
test_06_download_detached_volume | Failure | 137.57 | test_volumes.py
test_07_resize_fail | Failure | 4.23 | test_volumes.py

VMware:
test_01_list_sec_storage_vm | Failure | 0.11 | test_ssvm.py
test_05_stop_ssvm | Failure | 83.51 | test_ssvm.py
test_04_extract_template | Failure | 132.45 | test_templates.py
ContextSuite context=TestISOUsage>:setup | Error | 0.00 | test_usage.py
test_change_service_offering_for_vm_with_snapshots | Failure | 351.93 | test_vm_snapshots.py
test_01_create_volume | Failure | 150.18 | test_volumes.py
test_06_download_detached_volume | Failure | 177.71 | test_volumes.py

mike-tutkowski:
@rhtyd

An update with regards to the three XenServer test failures:

**test_02_edit_template | Failure | 90.10 | test_templates.py**:
I'm still looking into this one. I don't see off hand while this should behave differently from master with regards to this bit of logic.

**test_change_service_offering_for_vm_with_snapshots | Failure | 293.60 | test_vm_snapshots.py**:
I see there is new logic when changing the service offering of a stopped VM to invoke resizeVolume and pass in new performance values (i.e. min and max IOPS). resizeVolume was seeing that there is a VM snapshot and throwing an exception in this situation. I have changed the logic so that it doesn't do this when you are not looking to change the size of the volume in question.

**test_07_resize_fail | Failure | 41.45 | test_volumes.py**:
As a result of a change in logic in resizeVolume, we no longer check to see if you passed in a custom disk size when you specified a non-custom disk offering. Instead of silently performing another action (the code was just using the disk size of the new disk offering), I have changed the logic to throw an exception if you passed in a custom size with a non-custom disk offering.

rhtyd:
@mike-tutkowski can you also resolve the conflicts?

mike-tutkowski:
I should be able to do so pretty soon. I’m not at my computer right now.

On Jan 12, 2018, at 5:34 PM, Rohit Yadav <notifications@github.com<mailto:notifications@github.com>> wrote:


@mike-tutkowski<https://github.com/mike-tutkowski> can you also resolve the conflicts?

mike-tutkowski:
@rhtyd

**XenServer**:

test_volumes.py::test_07_resize_fail: **Fixed**

test_vm_snapshots.py::test_change_service_offering_for_vm_with_snapshots: **Fixed**

**KVM**:

test_router_dhcphosts.py::test_router_dhcphosts: **Doesn’t seem related to PR**: AssertionError: Ping to outside world from VM should be successful
test_router_dhcphosts.py::ContextSuite context=TestRouterDHCPHosts>:teardown: **Related to the problem in the issue above**

test_ssvm.py::test_01_list_sec_storage_vm: **Doesn’t seem related to PR**: AssertionError: Check gateway with that of corresponding ip range
test_ssvm.py::test_05_stop_ssvm: **Doesn’t seem related to this PR**: AssertionError: Check gateway with that of corresponding ip range

test_volumes.py::test_07_resize_fail: **Fixed** (same fix as first issue in XenServer section)

**Still investigating for KVM**

test_templates.py::test_03_deploy_vm_wrong_checksum
test_templates.py::ContextSuite context=TestCreateTemplateWithDirectDownload>:teardown
test_templates.py::test_04_extract_template

test_usage.py::ContextSuite context=TestISOUsage>:setup

test_volumes.py::test_06_download_detached_volume

**VMware**:

test_ssvm.py::test_01_list_sec_storage_vm: **Doesn’t seem related to PR**: AssertionError: Check gateway with that of corresponding ip range
test_ssvm.py::test_05_stop_ssvm: **Doesn’t seem related to PR**: AssertionError: Check gateway with that of corresponding ip range

test_vm_snapshots.py::test_change_service_offering_for_vm_with_snapshots: **Fixed** (same fix as second issue in XenServer section)

**Still investigating for VMware**

test_templates.py::test_04_extract_template

test_usage.py::ContextSuite context=TestISOUsage>:setup

test_volumes.py::test_01_create_volume
test_volumes.py::test_06_download_detached_volume

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

mike-tutkowski:
@rhtyd I don't seem to see any issues that would stem from those remaining tests (the ones I listed in the "Still investigating" category). As such, I'll wait for the currently running tests to complete and go from there. Thanks!

rhtyd:
@mike-tutkowski can you have a look at errors from last smoketest run?

rhtyd:
@mike-tutkowski /cc @rafaelweingartner @DaanHoogland 
I relooked at each of the errors, the only outstanding issues are with vmware test results, rest have been addressed in #2403:

test_change_service_offering_for_vm_with_snapshots | Failure | 299.25 | test_vm_snapshots.py

Based on my analysis of the logs and test results, all the kvm and xenserver failures are not related to this PR. The `test_public_ip_range.py` test adds a fake public IP range which if not removed causes new ssvm/cpvm to pick a fake public IP that causes iso/template/volume download and setup related failures, the new tests (failing sometimes now) were introduced in #2295.

If you can comment on that one failing vmware test (see above) we can accept this - please see that asap @mike-tutkowski @DaanHoogland @rafaelweingartner.

mike-tutkowski:
@rhtyd Thanks, Rohit! By the way, I have just finished manually walking through the steps in test_vm_snapshots.py::test_change_service_offering_for_vm_with_snapshots and they have always worked (I've tried it three times now).

rhtyd:
Okay @mike-tutkowski I'll take your word on it and request that you work with us in case we find regression post-merge. I'll cut RC1 tomorrow, but given our experience it's unlikely that RC1 graduates to a release which will get attention from community to start testing it and buy you some time to fix smoketest/regression and solidfire related tests failures. Do you agree and willing to work?

mike-tutkowski:
@rhtyd I’ll pretty much be running tests on the RC all week and will be available to work on any regressions, should they be found, as well.

On Jan 14, 2018, at 11:28 AM, Rohit Yadav <notifications@github.com<mailto:notifications@github.com>> wrote:


Okay @mike-tutkowski<https://github.com/mike-tutkowski> I'll take your word on it and request that you work with us in case we find regression post-merge. I'll cut RC1 tomorrow, but given our experience it's unlikely that RC1 graduates to a release which will get attention from community to start testing it and buy you some time to fix smoketest/regression and solidfire related tests failures. Do you agree and willing to work?

rhtyd:
Fantastic @mike-tutkowski, thanks for your support. I'll merge the PR as soon as XenServer test results are back (maybe in next hour).

rhtyd:
Tests LGTM. Merging this based on two code reviews and test results. Post merging I'll kick a smoketest round and work with @mike-tutkowski and others to fix any regressions. Thanks everyone for their contributions and work (of course the author!).

PR #2302, opened 2017-10-26
CLOUDSTACK-10122: Unrelated error message

nvazquez:
Simply remove unrelated text from error message

### DIFF
--- a/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
@@ -1409,21 +1409,16 @@ public class VolumeOrchestrator extends ManagerBase implements VolumeOrchestrati
                 if (value != null && value) {
                     cloneType = UserVmCloneType.full;
                 }
-                try {
-                    UserVmCloneSettingVO cloneSettingVO = _vmCloneSettingDao.findByVmId(vm.getId());
-                    if (cloneSettingVO != null){
-                        if (! cloneSettingVO.getCloneType().equals(cloneType.toString())){
-                            cloneSettingVO.setCloneType(cloneType.toString());
-                            _vmCloneSettingDao.update(cloneSettingVO.getVmId(), cloneSettingVO);
-                        }
-                    }
-                    else {
-                        UserVmCloneSettingVO vmCloneSettingVO = new UserVmCloneSettingVO(vm.getId(), cloneType.toString());
-                        _vmCloneSettingDao.persist(vmCloneSettingVO);
+                UserVmCloneSettingVO cloneSettingVO = _vmCloneSettingDao.findByVmId(vm.getId());
+                if (cloneSettingVO != null){
+                    if (! cloneSettingVO.getCloneType().equals(cloneType.toString())){
+                        cloneSettingVO.setCloneType(cloneType.toString());
+                        _vmCloneSettingDao.update(cloneSettingVO.getVmId(), cloneSettingVO);
                     }
                 }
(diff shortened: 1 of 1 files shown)


nvazquez:
Thanks @GabrielBrascher!

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
```diff
@@ -1423,7 +1423,7 @@ public void prepare(VirtualMachineProfile vm, DeployDestination dest) throws Sto
                     }
                 }
                 catch (Throwable e){
-                    s_logger.debug("[NSX_PLUGIN_LOG] ERROR: " + e.getMessage());
```
rafaelweingartner: @nvazquez thanks for the work!
Sorry, but I have to ask, do we need this try/catch block?
It seems that the code at 1413-1423 does not throw checked exceptions. Is there some kind of runtime exception that might happen in between 1413-1423 that we want to "hide"/swallow?

marcaurele: The log level should be moved to `warn`or `error` so that the message get caught in the logs when not running in debug. Putting error message as debug should only be done if they are irrelevant.

rhtyd:
LGTM, let's change the logging level to `warn` or `error`.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
rafaelweingartner: @marcaurele, do you think we need this try/catch? If we need, what about a more specific one, what exception we want to catch here? The way it is, we are catching much more than Pokemons here...

nvazquez: Thanks @rafaelweingartner @marcaurele, sorry for late response.
@rafaelweingartner good observation, I think we can either remove try/catch block or throw a checked exception which will be caught on `VirtualMachineManagerImpl#orchestrateStart`, what do you think?
@marcaurele you're right, done, thanks

rafaelweingartner: I would remove the try/catch block. I really did not seen any relevant exception in that piece of code to be dealt if there.

marcaurele: Agree with @rafaelweingartner - it should be removed all together. It would be hard anyway to track the problem since the log output will only output the message and not the stack trace. If we're trying to catch any error, the stack is very important to understand the issue. This would normally be done at a higher level if the exception isn't caught here.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
```diff
@@ -1423,7 +1423,7 @@ public void prepare(VirtualMachineProfile vm, DeployDestination dest) throws Sto
...
+                    s_logger.error("ERROR: " + e.getMessage());
```
rhtyd: LGTM

marcaurele:
@nvazquez can you change the PR to remove the try/catch block all together?

nvazquez:
Done. Thanks guys for your input @rhtyd @borisstoyanov @rafaelweingartner @marcaurele @GabrielBrascher

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #2303, opened 2017-10-30
CLOUDSTACK-10123: any entity should use GMT timestamp values

marcaurele:
Depending on the timezone you're running CS (before GMT timezones) you could experience that some jobs are marked as failed since the parent job got a null result despite its child job having successfully done the job. The child job got deleted by the `CleanupTask` ahead of time, due to a missing datetime conversion to GMT timezone.

Jobs are failing with this message: `Job failed with un-handled exception`

The fix intends to correct any datetime used in the code that should be using the GMT timezone instead of the local one since all DB datetime should be stored at GMT.


rhtyd:
LGTM, pending testing.
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #2306, opened 2017-11-03
CLOUDSTACK-10129: Allow navigation from VRs to network, instances, owners etc.

rhtyd:
The following navigation filtering is supported:
- Instances connected to a network, Networks -> view instances
- View account/owner and instances from a VR, Infra -> VR -> View account (if it's not project or system), view instances, view network(s) (Network -> list of VRs/tab already exists)
- Show redundant state for VPC VRs in parenthesis at Infra -> VRs -> list of VPC VRs

Pinging for review @DaanHoogland @wido @PaulAngus @nvazquez @borisstoyanov @DagSonstebo @glennwagner @rafaelweingartner @GabrielBrascher @fmaximus @ustcweizhou @resmo @NuxRo and others

### Network to Instances:
screenshot from 2017-11-03 15-00-43
screenshot from 2017-11-03 15-00-49

### VR listing
screenshot from 2017-11-03 15-01-43

### VR to instances, owner account and networks
screenshot from 2017-11-03 15-02-25
screenshot from 2017-11-03 15-02-31
screenshot from 2017-11-03 15-02-42

### DIFF
--- a/ui/scripts/system.js
+++ b/ui/scripts/system.js
@@ -31,6 +31,7 @@
 
         if (router.projectid) routerType = _l('label.project');
         if (router.vpcid) routerType = _l('label.vpc');
+        if ("isredundantrouter" in router && router.isredundantrouter) routerType = routerType + " (" + router.redundantstate + ")";
 
         return $.extend(router, {
             routerType: routerType
@@ -9599,6 +9600,23 @@
                             },
                             detailView: {
                                 name: 'label.virtual.appliance.details',
+                                viewAll: [{
+                                    label: 'label.account',
+                                    path: 'accounts',
+                                    preFilter: function(args) {
+                                        if (args.context.routers[0].projectid)
+                                            return false;
+                                        if (args.context.routers[0].account == 'system')
+                                            return false;
+                                        return true;
+                                    }
+                                }, {
+                                    label: 'label.networks',
+                                    path: 'network',
+                                }, {
+                                    label: 'label.instances',
+                                    path: 'instances'
+                                }],
                                 actions: {
                                     start: {
                                         label: 'label.action.start.router',
--- a/ui/scripts/accounts.js
+++ b/ui/scripts/accounts.js
@@ -137,6 +137,19 @@
                             });
                         }
 
+                        if ("routers" in args.context) {
+                            if ("account" in args.context.routers[0]) {
+                                $.extend(data, {
+                                    name: args.context.routers[0].account
+                                });
+                            }
+                            if ("domainid" in args.context.routers[0]) {
+                                $.extend(data, {
+                                    domainid: args.context.routers[0].domainid
+                                });
+                            }
+                        }
+
                         $.ajax({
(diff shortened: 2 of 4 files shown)


DaanHoogland:
creating an new experiment env with this as it was developed in an env used for other things
@blueorangutan package

DaanHoogland:
just discussed with @rhtyd 
holding off for now as he is adding some more informational links regarding VRs and networks

rhtyd:
@blueorangutan package

Thanks @resmo @borisstoyanov, testing UI should be enough.

ui/scripts/system.js
```diff
@@ -31,6 +31,7 @@
 
         if (router.projectid) routerType = _l('label.project');
         if (router.vpcid) routerType = _l('label.vpc');
+        if ("isredundantrouter" in router && router.isredundantrouter) routerType = routerType + " (" + router.redundantstate + ")";
```
nitin-maharana: @rhtyd, Thanks for adding this.
Can we replace these three if statements with if..else? Or do we override the value in this order?

ui/scripts/system.js
```diff
@@ -9599,6 +9600,23 @@
                             },
                             detailView: {
                                 name: 'label.virtual.appliance.details',
+                                viewAll: [{
+                                    label: 'label.account',
+                                    path: 'accounts',
+                                    preFilter: function(args) {
+                                        if (args.context.routers[0].projectid)
+                                            return false;
+                                        if (args.context.routers[0].account == 'system')
```
nitin-maharana: Is it for System VMs? If so, then it won't appear on this page.

ui/scripts/system.js
rhtyd: @nitin-maharana I'm not sure how could all these three disjoint if conditions be refactored into if..else, what do you have in mind?

ui/scripts/system.js
rhtyd: For shared networks, VR's account owner is 'system' we don't want to show 'View account' otherwise it fails as you cannot listAccounts for account=system.

ui/scripts/system.js
nitin-maharana: I'm also not sure. Because in all three statements we set a value to routerType variable according to the condition. If we don't overwrite the variable in the above order, then I think we can put them in if..else condition.

wido:
My JavaScript knowledge is really low. Can't help here :)

ui/scripts/system.js
rhtyd: @nitin-maharana thanks, here I'm not assuming that redundancy of the router is hard-coded VR being a VPC one only, the if statements are therefore not nested. In future, if we've redundant isolated/shared network VRs I would want the redundant state to be reflected (in parenthesis). I think it's alright to keep as such, unless of course you have a suggestion we can incorporate.

rhtyd:
Thanks @wido, how do you feel about the overall look and feel, navigation/filtering aspects of this small UI feature.

PR #2307, opened 2017-11-06
packging: Raise compat mode to 9

wido:
We are already compatible with that mode and should solve these
warnings:

<pre>
dh: Compatibility levels before 9 are deprecated (level 7 in use)
   debian/rules override_dh_auto_clean
make[1]: Entering directory '/home/wido/repos/cloudstack'
dh_auto_clean
dh_auto_clean: Compatibility levels before 9 are deprecated (level 7 in use)
rm -f replace.properties.tmp
make[1]: Leaving directory '/home/wido/repos/cloudstack'
   dh_clean
dh_clean: Compatibility levels before 9 are deprecated (level 7 in use)
 debian/rules build
make: 'build' is up to date.
 fakeroot debian/rules binary
dh binary --with python2,systemd
dh: Compatibility levels before 9 are deprecated (level 7 in use)
</pre>


rhtyd:
LGTM, @wido can you comment if this may break packaging on old ubuntu/debian distros? (i.e. I would expect this works with 14.04+).
I'll kick some packaging+testing.
@blueorangutan package

rhtyd:
@blueorangutan package

wido:
@rhtyd I checked the building of the packages and it works for me. Installed and ran just fine.

As far as I know the *compat* file is only for building, the end result does not change.

rhtyd:
Packaging LGTM, thanks @wido

PR #2309, opened 2017-11-07
HOLD - CLOUDSTACK-10132: Multiple Management Servers Support for agents

nvazquez:
JIRA Ticket: https://issues.apache.org/jira/browse/CLOUDSTACK-10132

Functional Specification: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Multiple+Management+Servers+Support+on+agents

This feature allows the direct agents to support multiple management servers. With this feature, a load balancer may no longer be necessary. The CloudStack administrator is responsible for setting a list of management servers and an algorithm in which to sort them, to the management server, using global configurations. The management server is responsible for applying algorithm to the management server list and propagate it to the agents. 

### Overview
A new global configuration is added: 'direct.agent.lb.algorithm'. Possible values of this global configuration are: 'static', 'roundrobin' and 'shuffle', corresponding to the algorithm that would be applied to the list of comma separated hosts defined on 'host' global configuration before sending them to the agents.
- 'static': No modification is made to the original 'host' list
- 'roundrobin': Rotate items one by one once the list is read
- 'shuffle': Randomly sort the list

A central utility is reading values from both global configurations on management server startup and is responsible of propagating list to each agent. In propagated list, the first element acts as the main host. Load balancing takes place by prepending the management server each agent should connect to, to the propagated actual list.

Once agents connect to management server, they receive a comma separated string of management servers. This value is compared to their actual 'host' configuration, and in case they differ, 'host' property is updated and persisted on 'agent.properties' file with the newly received string.

Reconnection logic between agent and management server is refactored as well. If agent is not able to establish a connection to its actual management server after some timeout interval, it attempts connection to the next element on its 'host' list.

Also, a background task is added to the agents to attempt reconnection to their main management server in pursuit of balancing. The interval of this background task is set on 'hosts.lb.background.interval' property on 'agent.properties' file. It can be optionally disabled, by setting property to 0

### Example
Supposing an environment in which there are 3 management servers: A, B and C and 3 KVM agents.
Setting 'host' = 'A,B,C', agents will receive lists depending on 'direct.agent.lb' value:
- 'static': Each agent will receive the list: 'A,B,C'
- 'roundrobin': First agent receives: 'A,B,C', second agent receives: 'B,C,A', third agent receives: 'C,B,A'
- 'shuffle': Each agent will receive a list in random order.


borisstoyanov:
@blueorangutan package

borisstoyanov:
@nvazquez thanks for this PR, I'll run marvin tests

borisstoyanov:
@blueorangutan test

agent/src/com/cloud/agent/AgentShell.java
```diff
@@ -107,18 +109,37 @@ public String getPod() {
     }
 
     @Override
-    public String getHost() {
+    public String getNextHost() {
         final String[] hosts = _host.split(",");
```
rhtyd: We can use getHosts() here.

agent/src/com/cloud/agent/AgentShell.java
```diff
@@ -173,6 +194,19 @@ public void setPersistentProperty(String prefix, String name, String value) {
             _storage.persist(name, value);
     }
 
+    @Override
+    public boolean setPropertyAndPersist(String property, String value) {
```
rhtyd: Don't use it, there is another way to do that --> call setPersistentProperty, see the method is just above your implementation.

utils/src/main/java/com/cloud/utils/PropertiesUtil.java
```diff
@@ -193,4 +195,19 @@ public static Properties loadFromFile(final File file)
         return properties;
     }
 
+    public static void persist(Properties properties, File file, String key, String value) throws IOException {
```
rhtyd: This can be removed as well. See how the keystore passphrase is persisted, follow that pattern. Hope this helps @nvazquez

nvazquez:
Thanks @rhtyd for pointing out pattern for persistent properties, I've addressed your comments

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@nvazquez @borisstoyanov I've made and pushed some additional edge cases and bugfixes now. In my test env it seemed okay. Please continue reviewing and testing.
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

nvazquez:
@rhtyd @borisstoyanov please find new additions, kindly review/test

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

agent/src/com/cloud/agent/Agent.java
```diff
@@ -699,16 +719,97 @@ public void processResponse(final Response response, final Link link) {
...
         if (ready.getHostId() != null) {
             setId(ready.getHostId());
         }
-        s_logger.info("Ready command is processed: agent id = " + getId());
 
+        final List<String> msList = ready.getMsList();
+        if (msList != null && msList.size() > 0) {
+            final List<String> hosts = Arrays.asList(_shell.getHosts());
+            s_logger.info("Received management servers list: " + msList + ", current list: " + hosts);
+            if (isReceivedListUpdated(hosts, msList)) {
```
rhtyd: It's debatable if we want to update the `host` list when order has changed, I think if the hosts are same (but order is different) we should ignore. In my testing I was able to reproduce an infinite switching behaviour when preferred host (the first in list) was not the one the agent was connected to leading to period disconnections (with just two hosts in the list).

agent/src/com/cloud/agent/Agent.java
```diff
@@ -699,16 +719,97 @@ public void processResponse(final Response response, final Link link) {
...
+                persistNewManagementServersList(msList);
+                if (!_shell.getConnectedHost().equals(msList.get(0))) {
+                    attemptReconnectionToPrimaryHost(msList.get(0));
```
rhtyd: Please remove the attempt to reconnect from here, the responsibiilty of doing that lies with the background task and not the ready command handler.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -990,4 +1091,53 @@ public void doTask(final Task task) throws TaskExecutionException {
...
+                 final String[] msList = _shell.getHosts();
+                 String preferredHost  = msList[index];
+                 final String connectedHost = _shell.getConnectedHost();
+                 if (s_logger.isTraceEnabled()) {
+                     s_logger.trace("Running host LB checker task, connected host=" + connectedHost + ", preferred host=" + preferredHost);
+                 }
+                 final long wait = _shell.getFailbackPollingWait();
+                 while (!preferredHost.equals(connectedHost)) {
+                     if (_link != null) {
+                         boolean isHostUp = true;
```
rhtyd: Optional: The code to check if the host is reachable on a port/address can we refactored.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -990,4 +1091,53 @@ public void doTask(final Task task) throws TaskExecutionException {
...
+             try {
+                 int index = 0;
+                 final String[] msList = _shell.getHosts();
+                 String preferredHost  = msList[index];
+                 final String connectedHost = _shell.getConnectedHost();
+                 if (s_logger.isTraceEnabled()) {
+                     s_logger.trace("Running host LB checker task, connected host=" + connectedHost + ", preferred host=" + preferredHost);
+                 }
+                 final long wait = _shell.getFailbackPollingWait();
+                 while (!preferredHost.equals(connectedHost)) {
```
rhtyd: Let's not have a while loop here. @nvazquez if you're not aware, when the agent is disconnected the `reconnect()` logic already kicks and is responsible for performing reconnections. This timer/task should not share the same responsibility, instead it should check if the preferred host (first in the list) is available and if the agent is already connected, only then it should disconnect by performing a reconnect on the `_link` (if it's not null for example).

Bear in mind, we've two assumptions here (1) the first host in the list is the preferred one, (2) the timer task is responsible for reconnecting to the preferred host.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -990,4 +1091,53 @@ public void doTask(final Task task) throws TaskExecutionException {
...
+                             if (_inProgress.get() == 0) {
+                                 if (s_logger.isDebugEnabled()) {
+                                     s_logger.debug("Preferred host is found to be reachable, trying to reconnect to: " + preferredHost);
+                                 }
+                                 _shell.resetHostCounter();
+                                 reconnect(_link);
+                             }
+                         }
+                     }
+                     index = (index + 1) % msList.length;
```
rhtyd: @nvazquez don't change the preferred host to some other host, it's because we assume the first in the list is the `preferred` host. For example, consider this use-case when using the `static` algorithm the first item in the list could be IP of the LB (netscalar/haproxy etc) and the use of this mechanism is to ensure HA for the mgmt server when this LB is down. If we change the preferred host, it breaks our use-case.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -990,4 +1091,53 @@ public void doTask(final Task task) throws TaskExecutionException {
...
+                             socket.connect(new InetSocketAddress(preferredHost, _shell.getPort()), 5000);
+                         } catch (final IOException e) {
+                             isHostUp = false;
+                             if (s_logger.isDebugEnabled()) {
+                                 s_logger.debug("Host: " + preferredHost + " is not reachable, trying to reconnect to next host...");
+                             }
+                         }
+                         if (isHostUp && _link != null) {
+                             long initialTime = System.currentTimeMillis();
+                             while (System.currentTimeMillis() - initialTime < wait && _inProgress.get() > 0) {
```
rhtyd: @nvazquez please avoid writing procedural code, in `reconnect` remaining tasks handling already exists. We should not have any timeouts/limits on these tasks, i.e. don't kill/purge outstanding tasks. Instead, one way to rewrite this could be to perform no operations when there are any tasks in `_inProgress`, and let the HostLBCheckerTask retry in the next round.

agent/conf/agent.properties
```diff
@@ -30,6 +30,19 @@ workers=5
...
+# one) and will attempt to reconnect to the preferred host when it's connected
+# to one of the secondary/backup hosts.
+# The timer task is scheduled when agent starts for the first time, any change
+# in this setting requires restarting the agent.
+host.lb.interval=0
+
+# The time in seconds in which the agent will wait for active commands to be completed
+# before connecting to preferred host
+# Any change in this setting requires restarting the agent.
+failback.polling.wait=10
```
rhtyd: @nvazquez this can be removed, see my comments below.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -699,16 +719,97 @@ public void processResponse(final Response response, final Link link) {
...
+        _shell.setPersistentProperty(null, "host", newHosts);
+        _shell.resetHostCounter();
+        s_logger.info("Saved new management servers list: " + msList);
+    }
 
+    /**
+     * Try reconnecting to preferred management server after new list is received.
+     * If connection is not performed, host counter is reset
+     */
+    private void attemptReconnectionToPrimaryHost(String preferred) {
```
rhtyd: This can be removed, duplicates code too.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -699,16 +719,97 @@ public void processResponse(final Response response, final Link link) {
...
+     * Checks if received list is different to actual management server list (in order and size)
+     * @param actual actual list
+     * @param received received list
+     * @return true if list is updated, false if not
+     */
+    private boolean isReceivedListUpdated(List<String> actual, List<String> received) {
+        if (actual.size() != received.size()) {
+            return true;
+        }
+        for (int i = 0; i < received.size(); i++) {
```
rhtyd: We can simply using the previous implementation to simply check if the existing setting differs in size or `set`.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -990,4 +1091,53 @@ public void doTask(final Task task) throws TaskExecutionException {
...
+                             while (System.currentTimeMillis() - initialTime < wait && _inProgress.get() > 0) {
+                                 s_logger.debug("Waiting for active commands to be completed");
+                                 Thread.sleep(5000l);
+                             }
+                             if (_inProgress.get() == 0) {
+                                 if (s_logger.isDebugEnabled()) {
+                                     s_logger.debug("Preferred host is found to be reachable, trying to reconnect to: " + preferredHost);
+                                 }
+                                 _shell.resetHostCounter();
+                                 reconnect(_link);
```
rhtyd: @nvazquez just so we're on the same page, when performing a reconnect after resting the host counter, `reconnect` will start getting host from the first in the list, i.e. the preferred one.

agent/src/com/cloud/agent/AgentShell.java
```diff
@@ -107,18 +111,48 @@ public String getPod() {
...
+    public String getNextHost() {
+        final String[] hosts = getHosts();
         if (_hostCounter >= hosts.length) {
             _hostCounter = 0;
         }
-        final String host = hosts[_hostCounter % hosts.length];
+        connectedHost = hosts[_hostCounter % hosts.length];
         _hostCounter++;
-        return host;
+        s_logger.info("Connecting to host: " + connectedHost);
```
rhtyd: This method only returns the next possible host in list, however, this does not mean we're connected to the host yet. This might cause threading issues, and in rare cases the connectedHost might not be the connected host. Reconsider how this could be re-written, perhaps the connectedHost can be updated using one of the `Task.CONNECT` handlers.

agent/src/com/cloud/agent/AgentShell.java
```diff
@@ -291,6 +325,11 @@ protected boolean parseCommand(final String[] args) throws ConfigurationExceptio
             _properties.setProperty("guid", _guid);
         }
 
+        String val = getProperty(null, "host.lb.interval");
+        hostLBTimerInterval = (Strings.isNullOrEmpty(val) ? 0L : Long.valueOf(val));
+        String wait = getProperty(null, "failback.polling.wait");
```
rhtyd: Remove this.

agent/src/com/cloud/agent/IAgentShell.java
```diff
@@ -22,33 +22,64 @@
...
 
-    public String getVersion();
+    String getVersion();
+
+    void setHosts(String hosts);
+
+    /**
+     * Resets host counter to start from the beginning
+     */
+    void resetHostCounter();
```
rhtyd: Remove comments, or add them for all the methods on this interface for consistency.

server/src/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -800,6 +804,7 @@ public Discoverer getMatchingDiscover(final Hypervisor.HypervisorType hypervisor
 
                 }
                 s_logger.info("server resources successfully discovered by " + discoverer.getName());
+                agentMSLB.propagateManagementServerListToAgents();
```
rhtyd: @nvazquez this is not desirable, this will cause the list to be sent to all the hosts whenever any single agent connects.

server/src/org/apache/cloudstack/agent/mslb/AgentMSLBServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+                    .collect(Collectors.toList());
+        }
+        return null;
+    }
+
+    @Override
+    public void propagateManagementServerListToAgents() {
+        List<HostVO> hosts = getOrderedRunningIndirectAgentIds();
+        if (hosts != null) {
+            for (HostVO h : hosts) {
```
rhtyd: @nvazquez I don't think it's a good idea, `resource claim and ownership` is not within the scope of this feature. This method will send command/answers to all the hosts (even those not managed by the `this` management server) and may cause disconnections. In case of multiple management servers, this will cause all the management servers to send multiple requests to all the agents. (note, when you do agent.send(), if the agent is not connected to `this` management server, it will use clustered mgmt server impl to send the request to the forwarding agent which is via other mgmt servers).

agent/src/com/cloud/agent/Agent.java
nvazquez: Done, thanks

agent/src/com/cloud/agent/Agent.java
nvazquez: Thanks @rhtyd, I assumed background task should attempt reconnection to any available host starting from the preferred one at the beggining of the list and trying in order on 'host' list. I refactored this background task to attempt reconnection only to preferred host as you suggested, I think it simplifies the feature

agent/src/com/cloud/agent/Agent.java
nvazquez: Done, thanks! Agreed

agent/src/com/cloud/agent/Agent.java
nvazquez: Right, my bad. Now it makes sense as timer task will attempt reconnection to the first host in the list

agent/src/com/cloud/agent/AgentShell.java
nvazquez: Done, thanks! Please checkout `hostToConnect` and `connectedHost` relationship using `updateConnectedHost()` method on connection to a host.

nvazquez:
@rhtyd thanks for your valuable review! I refactored code addressing most of your comments. There are quite few in which I think we should get aligned on and once we get aligned I could refactor as well! Thanks!

borisstoyanov:
@blueorangutan package

agent/src/com/cloud/agent/Agent.java
DaanHoogland: That seems to be a design issue then. When the order has changed the (ordered) list should be updated. It is debatable if the connection should be reset, though.

agent/src/com/cloud/agent/IAgentShell.java
DaanHoogland: I do not feel consistency is an issue here as much as the comment not giving more information than the methodname, so yes, please remove.

borisstoyanov:
@blueorangutan package

agent/src/com/cloud/agent/Agent.java
```diff
@@ -138,12 +143,14 @@ public int value() {
     private String _keystoreSetupPath;
     private String _keystoreCertImportPath;
 
+    private Long connectionTimeout = 120000L;
```
rhtyd: @nvazquez This is not used anywhere, can be removed.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -216,6 +223,7 @@ public String getPod() {
     }
 
     protected void setLink(final Link link) {
+        s_logger.info("Setting up link!");
```
rhtyd: I had put this for logging/debugging, may be removed. @nvazquez

agent/src/com/cloud/agent/Agent.java
```diff
@@ -265,6 +273,8 @@ public void start() {
                 s_logger.info("Attempted to connect to the server, but received an unexpected exception, trying again...");
             }
         }
+        _shell.updateConnectedHost();
```
rhtyd: @nvazquez to be on safer/deterministic side, can this method accept the `host` String and set that instead of `hostToConnect`?

agent/src/com/cloud/agent/Agent.java
```diff
@@ -699,16 +721,102 @@ public void processResponse(final Response response, final Link link) {
...
 
+        final List<String> msList = ready.getMsList();
+        if (msList != null && msList.size() > 0) {
+            final List<String> hosts = Arrays.asList(_shell.getHosts());
+            s_logger.info("Received management servers list: " + msList + ", current list: " + hosts);
+            if (isReceivedListUpdated(hosts, msList)) {
+                persistNewManagementServersList(msList);
+                _shell.resetHostCounter();
+                if (hasPrimaryHostChanged(hosts, msList)) {
+                    attemptReconnectionToPrimaryHost();
```
rhtyd: @nvazquez this can be removed, the ReadyCommand handler should not be responsible for reconnection/switching.

agent/src/com/cloud/agent/Agent.java
```diff
@@ -699,16 +721,102 @@ public void processResponse(final Response response, final Link link) {
...
+        _shell.setHosts(newHosts);
+        _shell.setPersistentProperty(null, "host", newHosts);
+        s_logger.info("Saved new management servers list: " + msList);
+    }
+
+    /**
+     * Try reconnecting to preferred management server after new list is received.
+     * If connection is not performed, host counter is reset
+     */
+    private void attemptReconnectionToPrimaryHost() {
```
rhtyd: @nvazquez this method duplicates code and is not necessary, may be removed.

server/src/org/apache/cloudstack/agent/mslb/AgentMSLBServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        List<String> msList = Arrays.asList(msServerAddresses.replace(" ", "").split(","));
+        AgentMSLBAlgorithm algorithm = getAgentMSLBAlgorithm();
+        return algorithm.getMSList(msList, orderedHostIds, hostId);
+    }
+
+    protected List<Long> getOrderedRunningHostIds(Long dcId) {
+        List<HostVO> hosts = resourceManager.listAllHostsInOneZoneByType(Host.Type.Routing, dcId);
+        if (hosts != null) {
+            return hosts.stream()
+                    .filter(x -> x.getHypervisorType().equals(Hypervisor.HypervisorType.KVM) &&
```
rhtyd: This should not be restricted just for KVM, but for any connected (indirect) agent.

agent/src/com/cloud/agent/Agent.java
nvazquez: @rhtyd I decided to take that path as when agent is launched, the next host is set on constructor and then from `start()` method we couldn't know which host the agent is trying to connect to unless we do the same as in AgentShell with `hostToConnect` but on Agent class. Which approach do you think it would be best?

nvazquez:
@rhtyd listeners for global settings update and new hosts added. Also shuffle is back to "normal" behavior. Lists will be propagated to agents only when 'hosts' or 'connected.agent.mslb.algorithm' values change or new hosts added, using new command-answer pattern instead of `ReadyCommand`. Please kindly review changes

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

DaanHoogland:
just want to see jenkins succeed

rhtyd:
Will open a new PR, closing this one.

rhtyd:
New PR https://github.com/apache/cloudstack/pull/2469/files

PR #2314, opened 2017-11-08
CLOUDSTACK-10136: Fix RemoteHostEndPoint thread growth

rhtyd:
This fixes the following:
- Unchecked thread growth in RemoteEndHostEndPoint
- Potential NPE while finding EP for a storage/scope

Unbounded thread growth can be reproduced with following findings:
- Every unreachable template would produce 6 new threads (in a single
ScheduledExecutorService instance) spaced by 10 seconds
- Every reachable template url without the template would produce 1 new
thread (and one ScheduledExecutorService instance), it errors out quickly without
causing more thread growth.
- Every valid url will produce upto 10 threads as the same ep (endpoint
instance) will be reused to query upload/download (async callback)
progresses.

Every RemoteHostEndPoint instances creates its own
ScheduledExecutorService instance which is why in the jstack dump, we
see several threads that share the prefix RemoteHostEndPoint-{1..10}
(given poolsize is defined as 10, it uses suffixes 1-10).

This fixes the discovered thread leakage with following notes:
- Instead of ScheduledExecutorService instance, a cached pool could be
used instead and was implemented, and with `static` scope to be reused
among other future RemoteHostEndPoint instances.
- It was not clear why we would want to wait when we've Answers returned
from the remote EP, and therefore a scheduled/delayed Runnable was
not required at all for processing answers. ScheduledExecutorService
was therefore not really required, moved to ExecutorService instead.
- Another benefit of using a cached pool is that it will shutdown
threads if they are not used in 60 seconds, and they get re-used for
future runnable submissions.
- Caveat: the executor service is still unbounded, however, the use-case
that this method is used for short jobs to check upload/download
progresses fits the case here.
- Refactored CmdRunner to not use/reference objects from parent class.

Screenshots showing deterministic thread growth for template with an invalid/unreachable URL:
screenshot from 2017-11-08 13-40-59

Screenshot showing threads transitioning from waiting->stopped (and re-use) with this fix:
screenshot from 2017-11-08 14-49-10

To verify, the following can be tried:
- Before applying this fix, in a test environment register two template such that (1) one has a reachable IP/domain but the resource does not exist (causing 404) and (2) the second template uses a domain/IP that is not reachable at all
- Thread growths can be checked using: `jstack -l <mgmt server PID> | grep RemoteHostEndPoint`, or using a visual tool such as VisualVM etc.
- With the fix + restart, the mgmt server will reattempt to download those template, and a humungous thread growth won't be seen and after say 2-4 minutes all the threads should shutdown, and  `jstack -l <mgmt server PID> | grep RemoteHostEndPoint` will show no threads.

Pinging for review - @DaanHoogland @nvazquez @borisstoyanov @PaulAngus @wido @mlsorensen @marcaurele and others

@blueorangutan package

### DIFF
--- a/engine/storage/src/org/apache/cloudstack/storage/RemoteHostEndPoint.java
+++ b/engine/storage/src/org/apache/cloudstack/storage/RemoteHostEndPoint.java
@@ -18,17 +18,15 @@
  */
 package org.apache.cloudstack.storage;
 
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.ScheduledExecutorService;
-import java.util.concurrent.TimeUnit;
 
 import javax.inject.Inject;
-
-import org.apache.log4j.Logger;
 
 import org.apache.cloudstack.engine.subsystem.api.storage.EndPoint;
 import org.apache.cloudstack.framework.async.AsyncCompletionCallback;
 import org.apache.cloudstack.managed.context.ManagedContextRunnable;
+import org.apache.log4j.Logger;
 
 import com.cloud.agent.AgentManager;
 import com.cloud.agent.Listener;
@@ -54,9 +52,11 @@ import com.cloud.vm.dao.SecondaryStorageVmDao;
 
 public class RemoteHostEndPoint implements EndPoint {
     private static final Logger s_logger = Logger.getLogger(RemoteHostEndPoint.class);
+
     private long hostId;
     private String hostAddress;
     private String publicAddress;
+
     @Inject
     AgentManager agentMgr;
     @Inject
@@ -65,10 +65,10 @@ public class RemoteHostEndPoint implements EndPoint {
     protected SecondaryStorageVmDao vmDao;
     @Inject
     protected HostDao _hostDao;
-    private ScheduledExecutorService executor;
+
+    private static ExecutorService executorService = Executors.newCachedThreadPool(new NamedThreadFactory("RemoteHostEndPoint"));
 
     public RemoteHostEndPoint() {
-        executor = Executors.newScheduledThreadPool(10, new NamedThreadFactory("RemoteHostEndPoint"));
     }
 
     private void configure(Host host) {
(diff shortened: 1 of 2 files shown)


rhtyd:
@blueorangutan test

borisstoyanov:
LGTM, tests does not show any regressions related to this change.

engine/storage/src/org/apache/cloudstack/storage/RemoteHostEndPoint.java
```diff
@@ -65,10 +65,10 @@
     protected SecondaryStorageVmDao vmDao;
     @Inject
     protected HostDao _hostDao;
-    private ScheduledExecutorService executor;
+
+    private static ExecutorService executorService = Executors.newCachedThreadPool(new NamedThreadFactory("RemoteHostEndPoint"));
```
DaanHoogland: bug fix 1 :+1:

engine/storage/src/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -106,16 +106,19 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
...
-        if (scope.getScopeType() == ScopeType.HOST) {
-            sbuilder.append(" and h.id = ");
-            sbuilder.append(scope.getScopeId());
-        } else if (scope.getScopeType() == ScopeType.CLUSTER) {
-            sbuilder.append(" and h.cluster_id = ");
-            sbuilder.append(scope.getScopeId());
-        } else if (scope.getScopeType() == ScopeType.ZONE) {
-            sbuilder.append(" and h.data_center_id = ");
-            sbuilder.append(scope.getScopeId());
+        if (scope != null) {
```
DaanHoogland: bug fix 2 :+1:

rhtyd:
Thanks @borisstoyanov and @DaanHoogland for reviewing, merged now.

PR #1176, opened 2015-12-04
CLOUDSTACK-9025: Fixed can't create usable template from snapshot in Xenserver and Vmware

anshul1886:
https://issues.apache.org/jira/browse/CLOUDSTACK-9025

Fix also reverts below commit as below solution making assumption about hypervisor which are not applicable in case of XenServer and VmWare

Revert "CLOUDSTACK-8964: Can't create template or volume from snapshot"

This reverts commit ccf5d75cfbe769b34c021ab3653ed318cae25933.

Testing:

Able to deploy VM successfully from template created from linked clone snapshot on XenServer.

PR #1124, opened 2015-11-25
CID-1338387: Deletion of method endPointSelector.selectHypervisorHost

rafaelweingartner:
Following the discussions and analysis presented on PR #1056 create by @DaanHoogland
This PR is intended to push those changes that were discussed there regarding the of endPointSelector.selectHypervisorHost method.

PR #1056, opened 2015-11-10
CID-1338387: remove logicless execution cycles

DaanHoogland:
I have removed the paramoia null check and created a specific exception as i think we always should for any specific condition. This would require careful consideration each time on whether and how specific the condition is of course ;)

PR #2315, opened 2017-11-09
A comprehensive solution for CLOUDSTACK-9025 and CLOUDSTACK-10128.

rafaelweingartner:
The first PR(#1176) intended to solve #CLOUDSTACK-9025 was only tackling the problem for CloudStack deployments that use single hypervisor types (restricted to XenServer). Additionally, the lack of information regarding the solution (documentation, test cases and description in PRs and Jira ticket) led the code to be removed in #1124 after a long discussion and analysis in #1056. That piece of code seemed logicless.  It would receive a hostId and then change that hostId for other hostId of the zone without doing any check; it was not even checking the hypervisor and storage in which the host was plugged into.

The problem reported in #CLOUDSTACK-9025 is caused by partial snapshots that are taken in XenServer. This means, we do not take a complete snapshot, but a partial one that contains only the modified data. This requires rebuilding the VHD hierarchy when creating a template out of the snapshot. The point is that the first hostId received is not a hostId, but a system VM ID(SSVM). That is why the code in #1176 fixed the problem for some deployment scenarios, but would cause problems for scenarios where we have multiple hypervisors in the same zone. We need to execute the creation of the VHD that represents the template in the hypervisor, so the VHD chain can be built using the parent links.

This commit changes the method com.cloud.hypervisor.XenServerGuru.getCommandHostDelegation(long, Command). From now on we replace the hostId that is intended to execute the “copy command” that will create the VHD of the template according to some conditions that were already in place. The idea is that starting with XenServer 6.2.0 hotFix ESP1004 we need to execute the command in the hypervisor host and not from the SSVM. Moreover, the method was improved making it readable and understandable; it was also created test cases assuring that from XenServer 6.2.0 hotFix ESP1004 and upward versions we change the hostId that will be used to execute the “copy command”.

Furthermore, we are not selecting a random host from a zone anymore. A new method was introduced in the HostDao called “findHostConnectedToSnapshotStoragePoolToExecuteCommand”, using this method we look for a host that is in the cluster that is using the storage pool where the volume from which the Snaphost is taken of. By doing this, we guarantee that the host that is connected to the primary storage where all of the snapshots parent VHDs are stored is used to create the template.

This PR was testes with ACS 4.9.2.0, primary storage using NFS and Xenserver hosts 6.5.0

This also solves #CLOUDSTACK-10128

### DIFF
--- a/api/src/com/cloud/hypervisor/HypervisorGuru.java
+++ b/api/src/com/cloud/hypervisor/HypervisorGuru.java
@@ -33,7 +33,7 @@ import com.cloud.vm.VirtualMachineProfile;
 
 public interface HypervisorGuru extends Adapter {
     static final ConfigKey<Boolean> VmwareFullClone = new ConfigKey<Boolean>("Advanced", Boolean.class, "vmware.create.full.clone", "true",
-                        "If set to true, creates guest VMs as full clones on ESX", false);
+            "If set to true, creates guest VMs as full clones on ESX", false);
     HypervisorType getHypervisorType();
 
     /**
@@ -45,7 +45,7 @@ public interface HypervisorGuru extends Adapter {
     VirtualMachineTO implement(VirtualMachineProfile vm);
 
     /**
-     * Give hypervisor guru opportunity to decide if certain command needs to be delegated to other host, mainly to secondary storage VM host
+     * Gives hypervisor guru opportunity to decide if certain commands need to be delegated to another host, for instance, we may have the opportunity to change from a system VM (is considered a host) to a real host to execute commands.
      *
      * @param hostId original hypervisor host
      * @param cmd command that is going to be sent, hypervisor guru usually needs to register various context objects into the command object
--- a/engine/schema/src/com/cloud/host/dao/HostDao.java
+++ b/engine/schema/src/com/cloud/host/dao/HostDao.java
@@ -19,6 +19,8 @@ package com.cloud.host.dao;
 import java.util.Date;
 import java.util.List;
 
+import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
+
 import com.cloud.host.Host;
 import com.cloud.host.Host.Type;
 import com.cloud.host.HostVO;
@@ -72,14 +74,6 @@ public interface HostDao extends GenericDao<HostVO, Long>, StateDao<Status, Stat
 
     List<HostVO> findHypervisorHostInCluster(long clusterId);
 
-    /**
-     * @param type
-     * @param clusterId
-     * @param podId
-     * @param dcId
-     * @param haTag TODO
-     * @return
-     */
     List<HostVO> listAllUpAndEnabledNonHAHosts(Type type, Long clusterId, Long podId, long dcId, String haTag);
 
     List<HostVO> findByDataCenterId(Long zoneId);
@@ -103,4 +97,16 @@ public interface HostDao extends GenericDao<HostVO, Long>, StateDao<Status, Stat
     List<HostVO> listByType(Type type);
 
     HostVO findByIp(String ip);
+
+    /**
+     * This method will look for a host that is of the same hypervisor and same zone as the storage pool where the volume of the Snapshot is stored.
+     * <ul>
+     * <li>(this is applicable only for XenServer) If the storage pool is managed, then we will look for a host that has the property 'supportsResign' in cloud.cluster_details
+     * <li>We give priority to 'Enabled' hosts, but if no 'Enabled' hosts are found, we use 'Disabled' hosts
+     * <li>If no host is found, we throw a runtime exception
+     * </ul>
+     *
+     * Side note: this method is currently only used in XenServerGuru; therefore, it was designed to meet XenServer deployment scenarios requirements.
+     */
+    HostVO findHostToOperateOnSnapshotBasedOnStoragePool(StoragePoolVO storagePoolVO);
 }
--- a/engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
+++ b/engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
@@ -30,13 +30,15 @@ import javax.annotation.PostConstruct;
 import javax.inject.Inject;
 import javax.persistence.TableGenerator;
 
-import com.cloud.configuration.ManagementServiceConfiguration;
+import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
+import org.apache.commons.lang3.StringUtils;
 import org.apache.log4j.Logger;
 import org.springframework.stereotype.Component;
 
 import com.cloud.agent.api.VgpuTypesInfo;
 import com.cloud.cluster.agentlb.HostTransferMapVO;
 import com.cloud.cluster.agentlb.dao.HostTransferMapDao;
+import com.cloud.configuration.ManagementServiceConfiguration;
 import com.cloud.dc.ClusterVO;
 import com.cloud.dc.dao.ClusterDao;
 import com.cloud.gpu.dao.HostGpuGroupsDao;
@@ -305,11 +307,11 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
         }
         HostTransferSearch.and("id", HostTransferSearch.entity().getId(), SearchCriteria.Op.NULL);
         UnmanagedDirectConnectSearch.join("hostTransferSearch", HostTransferSearch, HostTransferSearch.entity().getId(), UnmanagedDirectConnectSearch.entity().getId(),
-            JoinType.LEFTOUTER);
+                JoinType.LEFTOUTER);
         ClusterManagedSearch = _clusterDao.createSearchBuilder();
         ClusterManagedSearch.and("managed", ClusterManagedSearch.entity().getManagedState(), SearchCriteria.Op.EQ);
-        UnmanagedDirectConnectSearch.join("ClusterManagedSearch", ClusterManagedSearch, ClusterManagedSearch.entity().getId(), UnmanagedDirectConnectSearch.entity()
-            .getClusterId(), JoinType.INNER);
+        UnmanagedDirectConnectSearch.join("ClusterManagedSearch", ClusterManagedSearch, ClusterManagedSearch.entity().getId(), UnmanagedDirectConnectSearch.entity().getClusterId(),
+                JoinType.INNER);
         UnmanagedDirectConnectSearch.done();
 
         DirectConnectSearch = createSearchBuilder();
@@ -388,7 +390,8 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
 
         ClusterManagedSearch = _clusterDao.createSearchBuilder();
         ClusterManagedSearch.and("managed", ClusterManagedSearch.entity().getManagedState(), SearchCriteria.Op.EQ);
-        ClustersForHostsNotOwnedByAnyMSSearch.join("ClusterManagedSearch", ClusterManagedSearch, ClusterManagedSearch.entity().getId(), ClustersForHostsNotOwnedByAnyMSSearch.entity().getClusterId(), JoinType.INNER);
+        ClustersForHostsNotOwnedByAnyMSSearch.join("ClusterManagedSearch", ClusterManagedSearch, ClusterManagedSearch.entity().getId(),
+                ClustersForHostsNotOwnedByAnyMSSearch.entity().getClusterId(), JoinType.INNER);
 
         ClustersForHostsNotOwnedByAnyMSSearch.done();
 
@@ -520,18 +523,7 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
         return clusters;
     }
 
-    /*
-     * Returns a list of all cluster Ids
-     */
-    private List<Long> listAllClusters() {
-        SearchCriteria<Long> sc = AllClustersSearch.create();
-        sc.setParameters("managed", Managed.ManagedState.Managed);
-
-        List<Long> clusters = _clusterDao.customSearch(sc, null);
-        return clusters;
-    }
-
-    /*
+    /**
      * This determines if hosts belonging to cluster(@clusterId) are up for grabs
      *
      * This is used for handling following cases:
@@ -656,7 +648,7 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
         SearchCriteria<HostVO> sc = UnmanagedApplianceSearch.create();
         sc.setParameters("lastPinged", lastPingSecondsAfter);
         sc.setParameters("types", Type.ExternalDhcp, Type.ExternalFirewall, Type.ExternalLoadBalancer, Type.BaremetalDhcp, Type.BaremetalPxe, Type.TrafficMonitor,
-            Type.L2Networking, Type.NetScalerControlCenter);
+                Type.L2Networking, Type.NetScalerControlCenter);
         List<HostVO> hosts = lockRows(sc, null, true);
 
         for (HostVO host : hosts) {
@@ -790,10 +782,8 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
     @Override
     public List<HostVO> findLostHosts(long timeout) {
         List<HostVO> result = new ArrayList<HostVO>();
-        String sql =
-                "select h.id from host h left join  cluster c on h.cluster_id=c.id where h.mgmt_server_id is not null and h.last_ping < ? and h.status in ('Up', 'Updating', 'Disconnected', 'Connecting') and h.type not in ('ExternalFirewall', 'ExternalLoadBalancer', 'TrafficMonitor', 'SecondaryStorage', 'LocalSecondaryStorage', 'L2Networking') and (h.cluster_id is null or c.managed_state = 'Managed') ;";
-        try (
-                TransactionLegacy txn = TransactionLegacy.currentTxn();
+        String sql = "select h.id from host h left join  cluster c on h.cluster_id=c.id where h.mgmt_server_id is not null and h.last_ping < ? and h.status in ('Up', 'Updating', 'Disconnected', 'Connecting') and h.type not in ('ExternalFirewall', 'ExternalLoadBalancer', 'TrafficMonitor', 'SecondaryStorage', 'LocalSecondaryStorage', 'L2Networking') and (h.cluster_id is null or c.managed_state = 'Managed') ;";
+        try (TransactionLegacy txn = TransactionLegacy.currentTxn();
                 PreparedStatement pstmt = txn.prepareStatement(sql);) {
             pstmt.setLong(1, timeout);
             try (ResultSet rs = pstmt.executeQuery();) {
@@ -887,8 +877,7 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
     @Override
     @DB
     public List<RunningHostCountInfo> getRunningHostCounts(Date cutTime) {
-        String sql =
-            "select * from (" + "select h.data_center_id, h.type, count(*) as count from host as h INNER JOIN mshost as m ON h.mgmt_server_id=m.msid "
+        String sql = "select * from (" + "select h.data_center_id, h.type, count(*) as count from host as h INNER JOIN mshost as m ON h.mgmt_server_id=m.msid "
                 + "where h.status='Up' and h.type='SecondaryStorage' and m.last_update > ? " + "group by h.data_center_id, h.type " + "UNION ALL "
                 + "select h.data_center_id, h.type, count(*) as count from host as h INNER JOIN mshost as m ON h.mgmt_server_id=m.msid "
                 + "where h.status='Up' and h.type='Routing' and m.last_update > ? " + "group by h.data_center_id, h.type) as t " + "ORDER by t.data_center_id, t.type";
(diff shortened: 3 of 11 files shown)


engine/schema/src/com/cloud/host/dao/HostDao.java
```diff
@@ -72,14 +72,6 @@
...
     List<HostVO> findHypervisorHostInCluster(long clusterId);
 
-    /**
-     * @param type
-     * @param clusterId
-     * @param podId
-     * @param dcId
-     * @param haTag TODO
-     * @return
-     */
```
DaanHoogland: great removal, not all javadoc is helpful ;)

engine/schema/src/com/cloud/host/dao/HostDao.java
```diff
@@ -103,4 +95,13 @@
...
     HostVO findByIp(String ip);
+
+    /**
+     * This method will look for a host that is connected to the storage pool where the volume of the Snapshot is stored.
+     * <ul>
+     * <li>If the storage pool found for the volume of the snapshotId has more than a host, we will choose one randomly;
+     * <li>If no host is found, we throw a runtime exception
+     * </ul>
+     */
+    HostVO findHostConnectedToSnapshotStoragePoolToExecuteCommand(long snapshotId);
```
DaanHoogland: can i propose a shorter name? like findHostToOperateOnSnapshot(long id);

engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
```diff
@@ -792,11 +791,8 @@ public void loadHostTags(HostVO host) {
     @Override
     public List<HostVO> findLostHosts(long timeout) {
         List<HostVO> result = new ArrayList<HostVO>();
-        String sql =
-                "select h.id from host h left join  cluster c on h.cluster_id=c.id where h.mgmt_server_id is not null and h.last_ping < ? and h.status in ('Up', 'Updating', 'Disconnected', 'Connecting') and h.type not in ('ExternalFirewall', 'ExternalLoadBalancer', 'TrafficMonitor', 'SecondaryStorage', 'LocalSecondaryStorage', 'L2Networking') and (h.cluster_id is null or c.managed_state = 'Managed') ;";
-        try (
-                TransactionLegacy txn = TransactionLegacy.currentTxn();
-                PreparedStatement pstmt = txn.prepareStatement(sql);) {
+        String sql = "select h.id from host h left join  cluster c on h.cluster_id=c.id where h.mgmt_server_id is not null and h.last_ping < ? and h.status in ('Up', 'Updating', 'Disconnected', 'Connecting') and h.type not in ('ExternalFirewall', 'ExternalLoadBalancer', 'TrafficMonitor', 'SecondaryStorage', 'LocalSecondaryStorage', 'L2Networking') and (h.cluster_id is null or c.managed_state = 'Managed') ;";
```
DaanHoogland: not sure if I like all automatic refactors. 👎

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/XenServerGuru.java
```diff
@@ -58,7 +59,9 @@
 import com.cloud.vm.dao.UserVmDao;
 
 public class XenServerGuru extends HypervisorGuruBase implements HypervisorGuru, Configurable {
-    private final Logger LOGGER = Logger.getLogger(XenServerGuru.class);
+
+    private Logger logger = Logger.getLogger(getClass());
```
DaanHoogland: why not static final LOG?

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/XenServerGuru.java
```diff
@@ -167,35 +170,61 @@ public boolean trackVmHostChange() {
...
+            return defaultHostToExecuteCommands;
+        }
+        boolean isXenServer610 = StringUtils.equals(hypervisorVersion, "6.1.0");
+        if (isXenServer610) {
+            logger.debug("We are returning the default host to execute commands because the hypervisor version is 6.1.0.");
+            return defaultHostToExecuteCommands;
+        }
+        String snapshotHotFixVersion = hostCandidateToExecutedCommand.getDetail(XenserverConfigs.XS620HotFix);
+        boolean isXenServer620 = StringUtils.equals(hypervisorVersion, "6.2.0");
+        if (isXenServer620 && !StringUtils.equalsIgnoreCase(XenserverConfigs.XSHotFix62ESP1004, snapshotHotFixVersion)) {
```
DaanHoogland: Do we have to take into account updates beyond that hotfix version?

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
```diff
@@ -194,6 +193,13 @@ protected VirtualMachineTO toVirtualMachineTO(VirtualMachineProfile vmProfile) {
     }
 
     @Override
+    /**
+     * The basic implementation assumes that the initial "host" defined to execute the command is the host that is in fact going to execute it.
+     * However, subclasses can extend this behavior, changing the host that is going to execute the command in runtime.
+     * The first element of the 'Pair' indicates if the hostId has been changed; this means, if you change the hostId, but you do not inform this action in the return 'Pair' object, we will use the original "hostId".
+     *
+     * Side note: it seems that the 'hostId' received here is normally the ID of the SSVM that has an entry at the host table. Therefore, this methods gives the opportunity to change from the SSVM to a real host to execute a command.
+     */
```
DaanHoogland: ok, some javadoc does make sense

borisstoyanov:
Thanks for this PR @rafaelweingartner, let me run some testing on it. 
@blueorangutan package

borisstoyanov:

anshul1886:
Also see the fix in  https://github.com/apache/cloudstack/pull/2317 which I am working on.

engine/schema/src/com/cloud/host/dao/HostDao.java
rafaelweingartner: yep, I actually do not consider this a documentation ;)

engine/schema/src/com/cloud/host/dao/HostDao.java
rafaelweingartner: Sure that is a good name, I will change it.

engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
rafaelweingartner: You mean, to break into multiple lines? 
I actually did not notice this, probably I left my save actions applying in all lines, instead of only modified lines.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/XenServerGuru.java
rafaelweingartner: The result is the same.

The variable was not declared as "static", but its name was LOGGER as a static variable. I only changed because at first I removed it as it was not being used. Then, when I added some logs, I re-added the variable and I created this way. I can change to static final.

I can change to static if you think it is better for some reason.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/XenServerGuru.java
rafaelweingartner: I have no idea :)
I only re-worked the code, the logic here is the same. I made sure everything was kept the same with test cases.
The result of this code is the same as the replace one. The difference is that I am not using those if/if/if/if/if.....

server/src/com/cloud/hypervisor/HypervisorGuruBase.java
rafaelweingartner: yep, this is the Java doc type I keep telling people about; this type of documentation in the code is important (for the ones that do not have great memory as me, and for newcomers).

It took me quite a long time to understand the purpose of the method "getCommandHostDelegation" that was overridden in some classes.

rafaelweingartner:
@DaanHoogland I did not understand what you mean with more is needed?

I did not add code here, I only re-worked the logic to make it more clear. The only thing I changed is that instead of selecting a random host of the zone where the snapshot is located, I am locating a host based on the storage of the volume of the snapshot that is being converted to template.

DaanHoogland:
@rafaelweingartner nothing more is needed. @borisstoyanov did start the relevant tests

engine/schema/src/com/cloud/host/dao/HostDao.java
```diff
@@ -103,4 +95,13 @@
...
     HostVO findByIp(String ip);
+
+    /**
+     * This method will look for a host that is connected to the storage pool where the volume of the Snapshot is stored.
+     * <ul>
+     * <li>If the storage pool found for the volume of the snapshotId has more than a host, we will choose one randomly;
+     * <li>If no host is found, we throw a runtime exception
+     * </ul>
+     */
+    HostVO findHostToOperateOnSnapshot(long snapshotId);
```
anshul1886: @rafaelweingartner,  Command only goes to SSVM when only secondary storage is involved so no role of storage pool here so better to think of some other way to find host.

engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
```diff
@@ -1194,4 +1178,26 @@ public HostVO findByIp(final String ipAddress) {
         sc.setParameters("type", type);
         return listBy(sc);
     }
-}
+
+    String sqlFindHostConnectedToSnapshotStoragePoolToExecuteCommand = "select h.id from snapshots s join volumes v on v.id = s.volume_id "
+            + "join storage_pool pool on pool.id = v.pool_id join cluster c on pool.cluster_id = c.id join host h on h.cluster_id = c.id "
```
anshul1886: This method will not work in scenarios when volume or storage pool associated to snapshot is deleted. Also there is no need to consider storage pool as only secondary storage is involved. Any random hypervisor host in zone will do. Scenarios where storage pool is involved are already handled in core code and don't go via command delegation route. Also don't use sql queries instead write queries using Daos.

anshul1886:
It seems some of my old review comments are missing, so added again.

api/src/com/cloud/hypervisor/HypervisorGuru.java
```diff
@@ -45,7 +45,7 @@
     VirtualMachineTO implement(VirtualMachineProfile vm);
 
     /**
-     * Give hypervisor guru opportunity to decide if certain command needs to be delegated to other host, mainly to secondary storage VM host
+     * Give hypervisor guru opportunity to decide if certain command needs to be delegated to other host, mainly from the secondary storage VM host to a real hypervisor host
```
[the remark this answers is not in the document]
rafaelweingartner: But still, this method gives the opportunity to change the default host that is intended to execute a command right?
what about this message to help people understand the purpose of this method:
> Give hypervisor guru opportunity to decide if certain commands need to be delegated to another host, for instance, we may have the opportunity to change from a system VM (is considered a host) to a real host to execute commands.

engine/schema/src/com/cloud/host/dao/HostDao.java
rafaelweingartner: I did not understand this, can you explain what you mean with some more details?

engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
rafaelweingartner: I am using a DAO (Data access object), did you mean using a JPA queries (with ACS' home made JPA implementation)?

But if volume is deleted, how can we find its snapshots? I mean, aren't the snapshots of a volume deleted as well?

Are you sure that we use only the files in the secondary storage to build the template from a snapshot? I was under the impression that it was being used the primary storage as well.

rafaelweingartner:
@anshul1886 could you look at some of my questions? There are some things I did not understand.

api/src/com/cloud/hypervisor/HypervisorGuru.java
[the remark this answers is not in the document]
anshul1886: This message looks ok.

engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
anshul1886: Yes, we only use secondary storage to build the template from snapshot. This might not be applicable to managed store. But that's also available to all hosts in zone so any random host in a zone will do.

Also we can delete volume without deleting its snapshot (backup). We only delete VM snapshots when we delete VM.

engine/schema/src/com/cloud/host/dao/HostDao.java
anshul1886: For unmanaged storage snapshots goes on secondary storage. Secondary storage is accessible to all hosts in zone. So any random host in a zone will do. Once the snapshot is created for unmanaged storage it's independent of volume and storage pool. I believe managed store is zone wide so it is available to all and any random host in zone. Also it's handling is bit different. Confirm behaviour with @mike-tutkowski.

engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
rafaelweingartner: @anshul1886  I tested this solution here to recover a volume from a snapshot of a volume that has been deleted, and it works. As you can see I am not filtering the volumes that have been deleted. 

I also tested what you said that we only use the secondary storage to build the volume from snapshots, and it is right. This means, any XenServer of the Zone can execute this command then. I will update my select to reflect that.

Regarding JPA, I have mixed feeling about the JPA implementation ACS uses. It is not a standard one and it requires too much code to achieve the same thing we do with SQL. Also, It does not seem to be a standard in ACS to do everything with JPA. I would rather use Spring-data with a commercial JPA implementation (EclipseLink, Hibernate, OpenJPA), but this requires some structural changes.

engine/schema/src/com/cloud/host/dao/HostDao.java
rafaelweingartner: @anshul1886  Any host at all?
I mean, if I create a snapshot of a volume that is being used in a VM in a XenServer host. Can I use a Vmware ESXi host to transform this snapshot into a volume to be used by a XenServer VM?

rafaelweingartner:
@anshul1886 I changed the methods to reflect your suggestions.
Thanks for the help here.

engine/schema/src/com/cloud/host/dao/HostDao.java
mike-tutkowski: Currently managed storage is just supported at the zone level. I do have code in mind that would allow it to be confined to cluster scope, though. You make a good point that such a configuration (since snapshots for managed storage are stored on primary storage) would make the use of snapshots outside of that cluster somewhat complicated.

rafaelweingartner: @mike-tutkowski, following @anshul1886 I changed this code to look for any host in the zone that is of the same hypervisor type of the cluster where the original volume of the snapshot resides on. 

So, what do you guys think? let´s use only hosts of the cluster where the volume of the snapshot has been taken resides on? For me, it seems an easy way to cope with these different scenarios that we have.

mike-tutkowski: One workaround I already have in place, though, is the following (currently only supported on XenServer): You can pass in a parameter when taking the volume snapshot that tells us to put the volume snapshot on secondary storage instead of keeping it on primary storage. If, however, you have already created the snapshot on primary storage, then you'd need to create a volume from it and then snapshot that volume to secondary storage.

rafaelweingartner: So, this requires that we use a host that has access to the primary storage where the snapshot resides?

mike-tutkowski: On XenServer, making use of a volume snapshot that resides on primary storage needs to leverage a XenServer host that supports UUID resigning. UUID resigning is available as a patch for XenServer (@syed created it in 2016). When the CSMS connects with a XenServer host, it checks to see if this patch is installed. The results of the check are stored in cloud.host_details (the supportsResign property). If any host in the cluster supports UUID resigning, then the "convenience" property, also called supportsResign, in cloud.cluster_details is set to true. That being the case, we need to find one of these hosts in the zone when leveraging a managed snapshot on XenServer.

mike-tutkowski: "So, this requires that we use a host that has access to the primary storage where the snapshot resides?"

Correct, but currently we only support zone-wide managed storage, so all hosts in the zone should have access to that primary storage.

rafaelweingartner: @mike-tutkowski I will take your comments into consideration. This new requirement will change things a bit, but at least we will make a comprehensive solution for snapshots in XenServer deployments

As soon as I have an update, I will notify you guys

rafaelweingartner:
@mike-tutkowski how can I check if managed storage is enabled/being used?

Per your explanation, in case of managed storage I can take a look at the "convenience" property called 'supportsResign' in cloud.cluster_details to look for a host to execute the command. However, I did not 
understand how to check if a storage is managed.

Can I use the 'managed' property at the storage_pool table?

mike-tutkowski:
@rafaelweingartner Yes, you are correct: Use the 'managed' property of the storage pool in question.

Also, the supportsResign property is really only relevant for XenServer. All versions of ESXi supports UUID resigning by default and KVM doesn't require such support. (Note: Managed storage is only currently supported for XenServer, VMware, and KVM).

rafaelweingartner:
@mike-tutkowski I am now taking into consideration if the storage pool is managed or not.
I first load the storage pool where the volume of the snapshot is storage, and then I look for a host that I can use to execute the operation. If the storage pool is of managed type, I am going to check the parameter you mentioned (‘supportsResign’) in the cluster_details table; I am also expecting that property to be set as true if the host supports resign UUIDs.

Thanks for your help here ;)

P.S. As you mentioned, this changes only affect snapshots in XenServer environments.

mike-tutkowski:
I’m looking at the logic in StorageSystemSnapshotStrategy.java and it actually flows like this:

if (volumeInfo.getFormat() == ImageFormat.VHD) {
    HostVO hostVO = getHost(volumeInfo.getId());

    computeClusterSupportsVolumeClone = clusterDao.getSupportsResigning(hostVO.getClusterId());
}
else if (volumeInfo.getFormat() == ImageFormat.OVA || volumeInfo.getFormat() == ImageFormat.QCOW2) {
    computeClusterSupportsVolumeClone = true;
}
else {
    throw new CloudRuntimeException("Unsupported format");
}

rafaelweingartner:
So, it seems to be safe to take the same assumption. Meaning that, all of the hosts will support this feature if the cluster has the flag ‘SupportsResigning’ as true.

mike-tutkowski:
Correct, if the XenServer cluster supports UUID resigning (per the cloud.cluster_details supportsResign property), then all hosts in that cluster should support UUID resigning.

engine/schema/src/com/cloud/host/dao/HostDaoImpl.java
anshul1886: @rafaelweingartner SQL queries are hard to follow and is not recommended way. See https://github.com/apache/cloudstack/pull/2317 for example.

anshul1886:
@rafaelweingartner I hope one thing is clear here that in case of non-managed storage pool snapshots reside on secondary storage and can't be used to find host as once the snapshot is created its independent of storage pool and storage pool can be deleted without deleting snapshots.

anshul1886:
@rafaelweingartner Also take care, host which is getting used is enabled. Only when no enabled host is available use disabled host.

rafaelweingartner:
@anshul1886 it is clear to me now that for non-managed storage we do not need the primary storage to re-create a volume from a snapshot.

Regarding the host needing to be enabled, I agree. I only have some doubts about using hosts disabled. If there are no enabled hosts available I think we could simply throw an exception. I think we should not use disabled hosts.

anshul1886:
@rafaelweingartner Since we are not allocating any resources on those hosts we can use them for these kind of purposes only if no enabled host is available. I believe we are using disabled hosts for these kind of purposes elsewhere also.

rafaelweingartner:
@anshul1886 ok, I will do that.

rafaelweingartner:
@anshul1886 Donw

rhtyd:
Ping - can we have +1/-1 on code review?

rafaelweingartner:
@rhtyd if this helps for anything, I have this solution running in production for over a month now.
Let's see what other have to say now

mike-tutkowski:
@rhtyd I haven't actually examined the code for this PR. I was serving mainly as a source of answers for questions regarding how this feature might interact with managed storage. Let me know if we still need reviewers. At the time being, I've been racing pretty hard to finish up another project, so I probably couldn't spend the necessary time until, say, mid next week.

rhtyd:
@blueorangutan package

rhtyd:

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Ignoring known issues, tesrs lgtm. @anshul1886 and others are we lgtm on this?

rhtyd:
@rafaelweingartner can you fix conflicts on this?

rafaelweingartner:
@rhtyd Done.

rhtyd:
@blueorangutan package

rhtyd:
Ping - can we have additional review on it thanks.
@blueorangutan package

rhtyd:
@rafaelweingartner can you fix the conflicts?
@DaanHoogland @mike-tutkowski can you help review this?

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/XenServerGuru.java
```diff
@@ -167,35 +170,62 @@ public boolean trackVmHostChange() {
 
     @Override
     public Pair<Boolean, Long> getCommandHostDelegation(long hostId, Command cmd) {
```
DaanHoogland: in terms of cyclic complexity this method doesn't improve

rafaelweingartner: Actually, it depends on how you count. Normally you would need to build a decision graph to calculate. However, the most import thing in my opinion is the removal of 3/4 conditionals one inside the other. And More specifically this block:
```
if (hypervisorVersion != null && !hypervisorVersion.equalsIgnoreCase("6.1.0")) {
-                        if (!(hypervisorVersion.equalsIgnoreCase("6.2.0") &&
-                                !(snapshotHotFixVersion != null && snapshotHotFixVersion.equalsIgnoreCase(XenserverConfigs.XSHotFix62ESP1004)))) {
-                            return new Pair<Boolean, Long>(Boolean.TRUE, new Long(host.getId()));
-                        }
-                    }
````

Those negations altogether do not compute well in my mind...

rafaelweingartner:
@rhtyd conflicts solved

rhtyd:
@blueorangutan package

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/XenServerGuru.java
DaanHoogland: all the conditional returns compensate more then fully the complexity of the nested conditionals, so i think the complexity grew. But of course you are right and we should actually measure (lazy mode on - quickly)

rafaelweingartner: Yes you are right!
I have seen people counting the return as a point for the final metric; but, I am not sure yet if they count or not. In my view the decision point is in the conditional that enabled the return to be executed and not in the return per se.

borisstoyanov:
@blueorangutan test

mike-tutkowski:
I've read through the code and the existing comments about it. I don't think I have anything to add above and beyond what's already been discussed here.

LGTM

DaanHoogland:
merged as both positive test results and two lgtm where added

PR #2316, opened 2017-11-10
CLOUDSTACK-10137: Only link log4j if not already present

marcaurele:
When reinstalling the package, cloudstack-management fails to install properly due to the `log4j.xml` already present.

### DIFF
--- a/debian/cloudstack-management.postinst
+++ b/debian/cloudstack-management.postinst
@@ -57,7 +57,9 @@ if [ "$1" = configure ]; then
     chgrp cloud ${CONFDIR}/${DBPROPS}
     chown -R cloud:cloud /var/log/cloudstack/management
 
-    ln -s ${CONFDIR}/log4j-cloud.xml ${CONFDIR}/log4j.xml
+    if [ ! -e "$CONFDIR/log4j.xml" ]; then
+        ln -s ${CONFDIR}/log4j-cloud.xml ${CONFDIR}/log4j.xml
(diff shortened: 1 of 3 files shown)


marcaurele:
@rhtyd to be confirmed in your upstream environment.

debian/cloudstack-management.postinst
```diff
@@ -57,7 +57,9 @@ if [ "$1" = configure ]; then
     chgrp cloud ${CONFDIR}/${DBPROPS}
     chown -R cloud:cloud /var/log/cloudstack/management
 
-    ln -s ${CONFDIR}/log4j-cloud.xml ${CONFDIR}/log4j.xml
+    if [ ! -e "$CONFDIR/log4j.xml" ]; then
+        ln -s ${CONFDIR}/log4j-cloud.xml ${CONFDIR}/log4j.xml
```
rhtyd: @marcaurele perhaps we can fix it as `ln -sf`?

rhtyd:
@marcaurele thanks, I went ahead and fixed similar issues for centos6 and centos7 as well. Will fire some tests.
@blueorangutan package

debian/cloudstack-management.postinst
marcaurele: Yep

rhtyd:
Tests LGTM, this is a packaging only change so no need for regression testing. Was able to install and upgrade.

PR #2319, opened 2017-11-11
CLOUDSTACK-10138: Load br_netfilter in security_group management script

bwsw:
When setting

sysctl -w net.bridge.bridge-nf-call-arptables=1
sysctl -w net.bridge.bridge-nf-call-iptables=1
sysctl -w net.bridge.bridge-nf-call-ip6tables=1

/usr/share/cloudstack-common/scripts/vm/network/security_group.py
doesn't check that br_netfilter is load.

### DIFF
--- a/scripts/vm/network/security_group.py
+++ b/scripts/vm/network/security_group.py
@@ -38,8 +38,11 @@ iptables = Command("iptables")
 bash = Command("/bin/bash")
 ebtables = Command("ebtables")
 driver = "qemu:///system"
+br_netfilter_module = "br_netfilter"
+
 cfo = configFileOps("/etc/cloudstack/agent/agent.properties")
 hyper = cfo.getEntry("hypervisor.type")
+
 if hyper == "lxc":
     driver = "lxc:///"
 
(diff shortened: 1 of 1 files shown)


bwsw:
I suppose it also should be added to 4.10.1

ustcweizhou:
What if module br_netfilter does not exist ? for example on Ubuntu 12.04 or CentOS 7 ?

As I know, the issue only happens on Ubuntu 16.04.

Actually we fixed it by chef
```
execute 'modprobe br_netfilter' do
  command 'if [[ $(modinfo br_netfilter 2>1 /dev/null) ]]; then /sbin/modprobe br_netfilter; fi'
  only_if { node['platform_version'] == '16.04' }
end
```

bwsw:
Hello, 14.04 is affected too. 

Ok, generally, it either should be handled by 
- script (may be smarter) or 
-s etup of agent or 
- documentation (which is absent). 

I think the check should be imroved. Ok, I haven't realized that it may be absent, but this means that it also can be either compiled-in or disabled, and your check is also weak (for abstract user).

bwsw:
I also propose alternative approach based on installation script for debian/ubuntu:
https://github.com/apache/cloudstack/pull/2321

scripts/vm/network/security_group.py
```diff
@@ -61,6 +64,15 @@ def execute(cmd):
     logging.debug(cmd)
     return bash("-c", cmd).stdout
 
+def load_kernel_module(module_name):
+    module_proc_file = "/proc/modules"
+    idx = execute("grep -E '^%s' %s | cut -f1 -d' '" % (module_name, module_proc_file)).find(module_name)
+    if idx == -1:
+        logging.debug("Module %s is absent. Load it." % (module_name))
+        execute("modprobe %s" % (module_name))
```
rhtyd: Can this throw an error, should we try-catch it?

rhtyd:
@blueorangutan package

scripts/vm/network/security_group.py
bwsw: Hi, please, take a look at #2321, I think the solution when module is installed during dpkg install is better than this one and this PR should be declined.

rhtyd:
This PR may be closed since the other PR was merged.

PR #2321, opened 2017-11-12
CLOUDSTACK-10138: Load br_netfilter in security_group management script

bwsw:
The implementation does check and fix during installation process (Ubuntu 14.04/16.04)


ustcweizhou:
this LGTM

rhtyd:
LGTM, this is the post install script code. This change cannot be automatically verified, installation of packages is enough.
Manually tested on ubuntu. Merging now.

PR #2320, opened 2017-11-11
CLOUDSTACK-10178: Hotfixes to make 4.10 work

bwsw:
This PR implements several fixes which are vital for CS 4.10 to work:
1. Fixes absent IPv6 network definition bugs for basic zone which lead to exceptions on KVM agent if it uses SGs (management and agent affected)
2. Fixes the case when template is created from a snapshot (CLOUDSTACK-10140, merged https://github.com/apache/cloudstack/pull/2322)
3. Fixes ubuntu/debian br_netfilter dependency (CLOUDSTACK-10138, merged CLOUDSTACK-10138: Load br_netfilter in security_group management script)
4. Fixes quota plugin bug (https://github.com/apache/cloudstack/pull/2326)

PR #2322, opened 2017-11-13
CLOUDSTACK-10140: Fix for when template is created from snapshot template.properties are corrupted

bwsw:


rafaelweingartner:
@bwsw can you provide more details about this issue?
I read the #CLOUDSTACK-10140 Jira ticket and also the code, but I am still not understanding what is being done here.

I have some doubts here, for instance in `NfsSecondaryStorageResource.java` you stopped writing the size to the property, but I do not understand why.

bwsw:
Hi, @rafaelweingartner 
Basically, I started from this one:

https://github.com/apache/cloudstack/pull/2320/files#diff-71901db20e3d4adb6711e69c1c1ca2ba

it was found by @rhtyd and I fixed Id from 1 to actual problem. I don't include it into PR because master already includes it. Next I went to debugging and opened the issue (CLOUDSTACK-10140) you saw in Jira.

```
root@cs2-secstorage-nfs1:/secondary/template/tmpl/2/209# cat template.properties 
uniquename=209-2-3accb15b-8647-356b-9791-9e2f1ec36b6b
filename=bbab2601-050d-4c18-ae35-00f2ba44de28.qcow2
size=3254517760
qcow2.filename=bbab2601-050d-4c18-ae35-00f2ba44de28.qcow2
qcow2.virtualsize=10737418240
public=true
id=1
```

ID is fixed by code from 1st diff, but other information is incorrect including size (I don't see it in my snippet but AFAIK "size" field was incorrect and included something like UUID).

Next, https://issues.apache.org/jira/secure/ViewProfile.jspa?name=weizhou user provided the patch and everything started to work fine.

I suppose that 

```
long size = _storage.getSize(destFileFullPath);
``` 

works incorrectly , but the fix works the next way:
We write basic info on step one (first try block). Next, we configure format processor
```
processor = new QCOW2Processor();
```

which actually does the rest of things correctly.  Basically, I think if basic format is incorrect the processor code stumbles upon it and unable to write the rest of the format. 

Frankly speaking, I didn't dive into details and it's a shame but It takes a lot of time to move across all the ACS code, because I'm not a "real" java developer. But this patch really works. I patched the 4.10, compiled, deployed into Agent, recreated SSVM and templates started to work properly. Basically, I recommend to consult Wei Zhou about his patch.

rafaelweingartner:
Thanks for the information.
Can you also explain something else? What do you mean with template.properties?
Is that it was not being created? Or maybe the information contained in the file was not accurate?

ustcweizhou:
@rafaelweingartner  yes.
the content of template.properties is not correct, if the template is create from snapshot.
After restarting ssvm, the template will be removed from database but still exist on secondary storage, because the template.properties is not correct.

We had this issue a year ago and fixed it by the patch I posted.
It is a regression bug caused by commit bb2c02d14478ec3c1256f8739dd3e9a09deacb10 

In short, the template.properties should be saved before going to the next step.
```
+                }
+                try {
```

rafaelweingartner:
@ustcweizhou thanks!
Now thinks are a little bit more clear.

rhtyd:
@bwsw do a `git pull --rebase <asf/upstream remo> master` please

rafaelweingartner:
@bwsw your rebase did not work because you applied a rebase on a merge that you had already done previously, right?

It is easier to checrry-pick the first two commits on top of a clean master branch of yours and then push it here.

bwsw:
@rhtyd done
@rafaelweingartner thanks a lot.

services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
```diff
@@ -521,12 +521,8 @@ protected Answer copySnapshotToTemplateFromNfsToNfs(CopyCommand cmd, SnapshotObj
                     bufferWriter.write("uniquename=" + destData.getName());
                     bufferWriter.write("\n");
                     bufferWriter.write("filename=" + fileName);
-                    bufferWriter.write("\n");
-                    long size = _storage.getSize(destFileFullPath);
-                    bufferWriter.write("size=" + size);
```
rhtyd: @bwsw should we not write `size`?

rhtyd:
LGTM
@blueorangutan package

rhtyd:
@blueorangutan test

services/secondary-storage/server/src/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
rhtyd: Seems noj issue, ignore.

rhtyd:
Lgtm, ignoring intermittent known failures.

PR #2313, opened 2017-11-08
CLOUDSTACK-10135 ACL rules order is not maintained for ACL_OUTBOUND i…

yvsubhash:
…n VPC VR

Repro steps
1.Create a vpc with super cidr(172.16.0.0/16)
2. created a custom acl with at least 3 ACL_OUTBOUND rules with number oder like 15, 10, 20
3. Create a tier with the above ACL
4.Deploy an instance in the tier
5.In router the ACL rules wont be as per the sequence number order

PR #2323, opened 2017-11-13
CLOUDSTACK-10043: fix restore default drop for egress rules in ACL

resmo:
While verifying the fix for CLOUDSTACK-10135 in #2313 we found a regression.

~At least to version 4.5.x (verified in 4.5.3) every ACL hat a default drop for egress (ACL_OUTBOUND). This commit restores this behavior (regression fix)~

Ok, after some more investigation and  help of @fmaximus it seems the 4.5 implementation is wrong as well. It does not add a deny egress all but a deny egress all if there is any egress rule (even a deny), but anyway. 

This is a not a bug fix anymore but a default deny.

/cc @rhtyd @yvsubhash


rhtyd:
@blueorangutan package

rhtyd:
@resmo are you planning to work on option 3 (from dev@ ML) or go with this change (drop by default) ?

resmo:
I would say, close this PR and I am going for the option 3.

PR #2327, opened 2017-11-16
CLOUDSTACK-10129: Show account, network info in VR list view

rhtyd:
This shows the owner account and network of a VR in the VR list view,
and for VPCs shows the VPC name and redundant state of the VPC rVR.

screenshot from 2017-11-16 10-30-45

Pinging for review @PaulAngus @borisstoyanov @resmo @nvazquez @DaanHoogland @wido @ustcweizhou @rafaelweingartner  and others

### DIFF
--- a/ui/scripts/system.js
+++ b/ui/scripts/system.js
@@ -29,9 +29,19 @@
     var mapRouterType = function (index, router) {
         var routerType = _l('label.menu.system');
 
-        if (router.projectid) routerType = _l('label.project');
-        if (router.vpcid) routerType = _l('label.vpc');
-        if ("isredundantrouter" in router && router.isredundantrouter) routerType = routerType + " (" + router.redundantstate + ")";
+        if (router.projectid) {
+            routerType = _l('label.project');
+            router.account = router.project;
+        }
+
+        if (router.vpcid) {
+            routerType = _l('label.vpc');
+            router.guestnetworkname = router.vpcname;
+        }
+
+        if ("isredundantrouter" in router && router.isredundantrouter) {
+            router.guestnetworkname = router.guestnetworkname + " (" + router.redundantstate + ")";
+        }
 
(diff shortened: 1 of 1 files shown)


wido:
LGTM based on the code

ustcweizhou:
@rhtyd I suggest to show 'Network' or 'VPC' in 'Type' field, and show the network/vpc name in 'Network' field.

rhtyd:
@ustcweizhou thanks, I see your point. The router type is not something return by the API, but set by the UI limited to VPC, Project and System. I think it may be a good idea to show network/vpc/project name in another column, probably not under the 'network' column?

rhtyd:
I made some more changes considering suggestions from @ustcweizhou and now, we show project names in Account tab and VPC name (with redundant router state in parenthesis) in network column: (this is not perfect but useful in knowing what VR belongs to what account/project or vpcs)
screenshot from 2017-11-16 16-45-51

rhtyd:
Can you re-review @wido @ustcweizhou @borisstoyanov thanks.

ui/scripts/system.js
```diff
@@ -29,9 +29,19 @@
...
+            routerType = _l('label.project');
+            router.account = router.project;
+        }
+
+        if (router.vpcid) {
+            routerType = _l('label.vpc');
+            router.guestnetworkname = router.vpcname;
+        }
+
+        if ("isredundantrouter" in router && router.isredundantrouter) {
```
rafaelweingartner: I believe you can use directly `router.isredundantrouter` here without needing to check if the property `isredundantrouter` exists in the object `router`; `undefined `is interpreted as `false`.

rhtyd: Thanks, I get defensive writing code. I guess this is not be necessary here.

resmo:
LGTM

PR #2328, opened 2017-11-16
CLOUDSTACK-10012: Upgrade Jetty to 9.3.x

rhtyd:
Upgrades Jetty to Java8 compatible version

Pinging for review @marcaurele @nvazquez @borisstoyanov @DaanHoogland @wido and others

@blueorangutan package

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -123,7 +123,7 @@
     <cs.groovy.version>2.4.12</cs.groovy.version>
     <cs.nitro.version>10.1</cs.nitro.version>
     <cs.wiremock.version>2.8.0</cs.wiremock.version>
-    <cs.jetty.version>9.2.22.v20170606</cs.jetty.version>
+    <cs.jetty.version>9.3.22.v20171030</cs.jetty.version>
   </properties>
 
   <distributionManagement>


wido:
What would be the easiest way to test? Just create a DEB and run it?

marcaurele:
@wido The problem upgrading to jetty 9.3 or 9.4 is the wiremock dependency which forces the use of jetty 9.2. The dependency is only used by cloudian module. I was looking to push an upgrade to 9.4 along with the new gzip handler and a context redirect. I haven't found the time to find a solution for the wiremock dependency.

- https://github.com/tomakehurst/wiremock/issues/407
- https://github.com/tomakehurst/wiremock/issues/644
- https://groups.google.com/forum/#!topic/wiremock-user/0W600q57pl0

rhtyd:
@wido yes, building a package and testing that management server works as expected is a valid regression test.
Thanks @marcaurele -- last time I tried using Jetty 9.4.x, I faced issues around server configuration which is why I dropped to using 9.2.x for the fatjar PR. The wiremock stuff is used only for writing unit tests for the cloudian plugin, I don't see how they are related to Jetty/running mgmt server. I'll close this PR in favour of your PR #2329, thanks.

PR #2332, opened 2017-11-17
[CLOUDSTACK-10156] Fix Coverity new problems CID(1349987, 1349986, 1347248)

rafaelweingartner:


plugins/database/quota/src/org/apache/cloudstack/api/command/QuotaStatementCmd.java
```diff
@@ -119,10 +118,6 @@ public String getCommandName() {
 
     @Override
     public long getEntityOwnerId() {
-        Long accountId = _accountService.getActiveAccountByName(accountName, domainId).getAccountId();
```
rhtyd: This can be kept to know who made the API call

rhtyd:
LGTM

plugins/database/quota/src/org/apache/cloudstack/api/command/QuotaStatementCmd.java
rafaelweingartner: Ah, so that is what this is for.
I will create a log then.

plugins/database/quota/src/org/apache/cloudstack/api/command/QuotaStatementCmd.java
```diff
@@ -120,9 +119,7 @@ public String getCommandName() {
     @Override
     public long getEntityOwnerId() {
         Long accountId = _accountService.getActiveAccountByName(accountName, domainId).getAccountId();
-        if (accountId == null) {
-            return CallContext.current().getCallingAccount().getId();
-        }
+        s_logger.debug(String.format("Account [id=%d, name=%s, domain=%d] is executing method: %s ", accountId, accountName, domainId, getClass().getSimpleName()));
```
rhtyd: Logging is not necessary here.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Tests lgtm, fixed redundant logging. Will merge.

plugins/database/quota/src/org/apache/cloudstack/api/command/QuotaStatementCmd.java
```diff
@@ -120,8 +119,8 @@ public String getCommandName() {
     @Override
     public long getEntityOwnerId() {
         Long accountId = _accountService.getActiveAccountByName(accountName, domainId).getAccountId();
```
rafaelweingartner: @rhtyd this will never be null. The method `getAccountId` returns a primitive long. Therefore, this check that you added is not needed. That is why when you asked to preserve the call `_accountService.getActiveAccountByName(accountName, domainId).getAccountId();` I wrote a code that was logging the value. Otherwise it would be a  call that did not make sense.

If you take a look at the first commit and compare it with HEAD, you are going to see the following code in HEAD:

>   if (accountId == null) {
>             return CallContext.current().getCallingAccount().getId();
>        }
>   return Account.ACCOUNT_ID_SYSTEM;

`accountId` is never null, that is why I removed that block, and left only the ` return Account.ACCOUNT_ID_SYSTEM;` code. If you think that we have to return the value of `accountService.getActiveAccountByName(accountName, domainId).getAccountId();` we can simply return it as it is a primitive long and will never be `null`

rhtyd: Sure, let's return long value of accountId.

rafaelweingartner:
@rhtyd sure I can create a Jira ticket. I have some remarks in the code you are introducing.

I will create the Jira while we check that point there.

plugins/database/quota/src/org/apache/cloudstack/api/command/QuotaStatementCmd.java
rafaelweingartner: Great. I will change this then

rafaelweingartner:
@rhtyd done!

rhtyd:
@rafaelweingartner I've fixed a minor spacing/styling issue. LGTM merging.

rafaelweingartner:
Thanks @rhtyd !
I am using ACS eclipse code formatter, I will check next time if it has some misconfiguration, or if I disabled something in the save actions.

PR #1555, opened 2016-05-20
Add Java Default Certificat Authorities into the keystore if using a custom cert SSL

milamberspace:
Related to CLOUDSTACK-1475

PR #2333, opened 2017-11-20
Add Java Default Certificat Authorities into the keystore if using a custom cert SSL

svenvogel:
change defaultJavaKeyStoreFile always to the correct path. paths are changed since any version of java.

e.g. /usr/lib/jvm/zulu-8-amd64/jre/lib/security/cacerts

Old pull request #1555


systemvm/scripts/config_ssl.sh
```diff
@@ -120,7 +120,7 @@ customCACert=
 publicIp=
 hostName=
 keyStore=$(dirname $0)/certs/realhostip.keystore
-defaultJavaKeyStoreFile=/etc/ssl/certs/java/cacerts
+defaultJavaKeyStoreFile="$(jrunscript -e 'java.lang.System.out.println(java.lang.System.getProperty("java.home"));')/lib/security/cacerts"
```
rhtyd: LGTM for 4.10, however for 4.11 we're going back to openjdk-8 provided in Debian "Stretch" where this may not be needed. Maybe you want to send this PR for 4.10?

svenvogel: Hi @rhtyd 
yes we can do it as fix for 4.10. no problem. is it clear that 4.11 will use the old path and for longer time or do we need a more ultimate fix?

Sven

svenvogel: how we can change that to 4.10?

rafaelweingartner:
@svenvogel you need to change the target branch. Currently, you opened this PR against master. If you wnt to propose this to 4.10 you can open a PR against 4.10 branch.

svenvogel:
@rafaelweingartner changed to 4.10

rafaelweingartner:
@svenvogel now the problem is that you are bringing a lot of changes from master to 4.10.
Can you rebase only your commits on top of 4.10?

rhtyd:
@svenvogel can you rebase against latest master and see if you can reproduce against master which has moved to openjdk8 and debian9 based systemvmtemplate.

rhtyd:
Thanks @svenvogel I checked your original change: https://github.com/apache/cloudstack/pull/2333/commits/53b626a9b7a34c83ce2e80274923433f5bcf2772

This is not applicable with the master/4.11 systemvmtemplate which uses openjdk8 than zulu8, and the default cacerts are indeed at `/etc/ssl/certs/java/cacerts`. Closing this PR, on this remark.

For 4.10, please re-submit a new PR or re-open this after rebasing against 4.10 and fixing the base branch to 4.10. Thanks.

PR #2337, opened 2017-11-25
CLOUDSTACK-10157: Wrong notification while migration.

nitin-maharana:
Steps to Reproduce:
===============
1. Create a cluster with three hosts, two of them with GPU cards NVIDIA GRID K1 and GRID K2 and one normal host.
2. Deploy a VM with GRID K2 vGPU.
3. Stop the VM and try starting it on a different host. (Expect it to fail as there is only one host supporting that particular vGPU type)
4. The log shows it as failed operation, VM remains in the stopped state, yet the notification shows "Start Instance" as successful and no error message displays.

Root Cause:
=========
Earlier, it was failing with ArrayIndexOutOfBoundsException, when the list is empty and accessing the first element.
The error was only observed in Log, but was not showing in UI as it was not throwing any exception.
Hence the API call was in turn successful.

Solution:
=======
Added the empty check before sending device details.
Which says either the required GPU device is not available or out of capacity.

### DIFF
--- a/server/src/com/cloud/resource/ResourceManagerImpl.java
+++ b/server/src/com/cloud/resource/ResourceManagerImpl.java
@@ -29,6 +29,10 @@ import java.util.Map;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
+import org.apache.commons.lang.ObjectUtils;
+import org.apache.log4j.Logger;
+import org.springframework.stereotype.Component;
+
 import org.apache.cloudstack.api.ApiConstants;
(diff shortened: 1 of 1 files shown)


nitin-maharana:
Thanks, @rafaelweingartner !!

rhtyd:
LGTM, simple NPE check.

rhtyd:
Given our current regression testing system does not have GPU capabilties, I'll use travis test results as verification of the build/code, two reviews/lgtms and merge this PR.

PR #1664, opened 2016-08-27
CLOUDSTACK-8676 Deploy user instance from vm snapshot for VMware hypervisor

sateesh-chodapuneedi:
Currently, ACS provides the ability to deploy a VM from a template or ISO. However, ACS does not provide the ability to deploy a VM(s) directly from a VM snapshot.

VM snapshots are stored in the primary storage and have a hierarchical or parent/child relationship. The requirement would be to provide the ability to deploy user instances from selected VM snapshots. Additionally, any VM snapshot in the hierarchy can be deployed concurrently.

Even though this can be supported and applicable to all hypervisors, to start with this feature is supported only for VMware hypervisor.

Also this feature supports creation of volume from VM snapshots. Details of new APIs added as part of this feature can be found in feature specification document below.

Feature specification is at https://cwiki.apache.org/confluence/display/CLOUDSTACK/Deploy+instance+from+VM+snapshot

There is a separate PR raised by @nitt10prashant  that contains marvin tests for this feature.

PR #1320, opened 2016-01-08
CLOUDSTACK-8968: UI icon over VM snapshot to deploy user instance

nitin-maharana:
Added a new Icon in Instance page to launch the VM from the snapshot.

A new icon over VM snapshot object, which upon invoked will open a deploy user instance wizard, where there is no choice to choose zone and ISO or template to deploy the user instance from.

1) A new icon to indicate deploy user instance from VM snapshot
2) The new icon is placed as an operation over each VM snapshot object.
3) Clicking this icon will show the wizard to deploy user instance, which is same as wizard for normal user instance except that zone selection page and ISO/template selection page would not be present in this wizard.

PR #2338, opened 2017-11-25
[WIP] CLOUDSTACK-8968: UI icon over VM snapshot to deploy user instance.

nitin-maharana:
Added a new Icon in the Instance as well as storage page to launch the VM from the snapshot.

A new icon over VM snapshot object, which upon invoked will open a deploy user instance wizard, where there is no choice to choose zone and ISO or template to deploy the user instance from.

The new icon is placed as an operation over each VM snapshot object.
Clicking this icon will show the wizard to deploy user instance, which is same as the wizard for normal user instance except that zone selection page and ISO/template selection page would not be present in this wizard.

It only supports in the case of VMware.

Other changes:
Maintained indentation in storage.js
Added the dictionary values according to google translate.

This PR is dependent on parent PR #1664.
Reference PR: #1320


DaanHoogland:
@rhtyd seems like good functionality; implement in primate and close here?

DaanHoogland:
@shwstppr @rhtyd @svenvogel this might best be considered for primate and closed here?

rhtyd:
Hi, the PR got auto-closed due to no known source repository during the recent renaming of the default branch to `main`. If you wish kindly submit a new PR.

PR #2340, opened 2017-11-26
CLOUDSTACK-10106: GPU/vGPU Support on VMWare

nitin-maharana:
Please refer the ticket for description and FS.

Ticket: [CLOUDSTACK-10106](https://issues.apache.org/jira/browse/CLOUDSTACK-10106)

Note:
====
Currently, It supports NVIDIA GRID K1 & GRID K2 cards on ESXi 6.0 U2 and later version.


plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2254,6 +2270,46 @@ protected void configureVideoCard(VirtualMachineMO vmMo, VirtualMachineTO vmSpec
         }
     }
 
+    /**
+     * Sets GPU card details to the one provided in detail vgpu.type (if provided) on {@code vmConfigSpec}.
+     * @param vmMo
```
rafaelweingartner: @nitin-maharana I think that these @param without much explanations can be removed.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -2254,6 +2270,46 @@ protected void configureVideoCard(VirtualMachineMO vmMo, VirtualMachineTO vmSpec
...
+                s_logger.error(msg);
+                throw new Exception(msg);
+            }
+
+            pciPassthroughDevice = hostMo.prepareDirectPciPassthroughDevice(hostPciDeviceInfo);
+        } else {
+            pciPassthroughDevice = hostMo.prepareSharedPciPassthroughDevice(vGpuProfile);
+        }
+
+        if (pciPassthroughDevice != null) {
```
rafaelweingartner: Is it possible for this `pciPassthroughDevice ` to be null after the previous IF/ELSE?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -3348,9 +3404,35 @@ protected Answer execute(StopCommand cmd) {
         // i.e., i-x-y. This is the internal VM name.
         VmwareContext context = getServiceContext();
         VmwareHypervisorHost hyperHost = getHyperHost(context);
+
         try {
             VirtualMachineMO vmMo = hyperHost.findVmOnHyperHost(cmd.getVmName());
+
             if (vmMo != null) {
```
rafaelweingartner: What about extracting this code from 3411-3429 to a method? It might also be good to Document it, and write some unit test cases

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -4766,6 +4848,31 @@ protected Answer execute(UnregisterNicCommand cmd) {
         }
     }
 
+    /**
+     * GetGPUStatsCommand is used to collect the GPU/vGPU details of all the
+     * GPU enabled hosts.
+     * Do Nothing.
+     * @param cmd
```
rafaelweingartner: These @param and @result without much description you remove them

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -4766,6 +4848,31 @@ protected Answer execute(UnregisterNicCommand cmd) {
...
+     */
+    protected Answer execute(final GetGPUStatsCommand cmd) {
+        final VmwareContext context = getServiceContext();
+        final VmwareHypervisorHost hyperHost = getHyperHost(context);
+        final HostMO hostMo = new HostMO(context, hyperHost.getMor());
+
+        HashMap<String, HashMap<String, VgpuTypesInfo>> groupDetails = new HashMap<String, HashMap<String, VgpuTypesInfo>>();
+
+        try {
+            groupDetails = hostMo.getGPUGroupDetails();
```
rafaelweingartner: What about using the code of line 4873 right after this line?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -5101,14 +5209,32 @@ private void fillHostNetworkInfo(VmwareContext serviceContext, StartupRoutingCom
...
+        final VmwareHypervisorHost hyperHost = getHyperHost(context);
+        final HostMO hostMo = new HostMO(context, hyperHost.getMor());
+
+        try {
+            final HashMap<String, HashMap<String, VgpuTypesInfo>> groupDetails = hostMo.getGPUGroupDetails();
+            cmd.setGpuGroupDetails(groupDetails);
+            if (groupDetails != null && !groupDetails.isEmpty()) {
+                cmd.setHostTags("GPU");
+            }
+        } catch (final Exception e) {
```
rafaelweingartner: Can't we use a more specific exception here?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
nitin-maharana: @rafaelweingartner, I just cross-checked, It's not possible because we initialize the object in hostMo prepareDirectPciPassthroughDevice and prepareSharedPciPassthroughDevice If/Else statement. It's redundant but I feel it's safe to check. Do you want me to remove?  Thanks!!

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
nitin-maharana: Sure, I would do that.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
nitin-maharana: You mean this statement: return new GetGPUStatsAnswer(cmd, groupDetails);?

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
nitin-maharana: Sure. I will modify this.

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
rafaelweingartner: Well, I understand the idea to code avoiding/deflecting null pointer exceptions. However, cases like this when it is not possible to happen a null pointer exception. I would not do the check/conditional; otherwise, it feels that for every time we use an object we would have to add a conditional checking if that object is a null value or not.

I would not force (suggest strongly) you to remove this because it does not harm the code, but I called your attention to this matter. I know that it might seem nice that we are deflecting null pointers, but if they do not happen at all, we are not deflecting anything ;)

It is a personal choice (so the power is yours here).

plugins/hypervisors/vmware/src/com/cloud/hypervisor/vmware/resource/VmwareResource.java
rafaelweingartner: Yes this one. Then, you could do directly inside the Try/catch:
` HashMap<String, HashMap<String, VgpuTypesInfo>> groupDetails = hostMo.getGPUGroupDetails()`

This would avoid the creation of a hashmap that is subsequently just dropped for the garbage collector.

rhtyd:
@nitin-maharana can you address outstanding comments, then I can initiate testing for you.

nitin-maharana:
@rhtyd Done!!

rhtyd:
Re-review requested.
@blueorangutan package

vmware-base/src/com/cloud/hypervisor/vmware/mo/HostMO.java
```diff
@@ -1184,4 +1206,261 @@ public ManagedObjectReference waitForPortGroup(String networkName, long timeOutM
...
+                                    }
+                                }
+                            }
+                        }
+                    }
+
+                    // If GRID K1, then search for only K1 vGPU types. Same for GRID K2.
+                    // The remaining capacity of one type affects other vGPU type capacity.
+                    // Each GPU should always contain one type of vGPU VMs.
+                    if ((groupName.equals("NVIDIAGRID K1") && vgpuType.equals("grid_k100")) || (groupName.equals("NVIDIAGRID K2") && vgpuType.equals("grid_k200"))) {
```
rhtyd: @nitin-maharana can you refactor this code by assigning the groupName against an index where `graphicInfo.getVm().size()` should be used instead of static/hard-coded `remainingCapacities` values.

vmware-base/src/com/cloud/hypervisor/vmware/mo/HostMO.java
```diff
@@ -1184,4 +1206,261 @@ public ManagedObjectReference waitForPortGroup(String networkName, long timeOutM
...
+                        remainingCapacities[3] -= 2l;
+                        remainingCapacities[4] -= ((long)graphicInfo.getVm().size());
+                    }
+                }
+            }
+        }
+
+        //Pattern.matches("grid_k100|grid_k1[2468]0[q]", gpuType)
+        for (final String gpuType : sharedPassthruGpuTypes) {
+            if ((groupName.equals("NVIDIAGRID K1") && gpuType.equals("grid_k100")) || (groupName.equals("NVIDIAGRID K2") && gpuType.equals("grid_k200"))) {
```
rhtyd: @nitin-maharana it's probably worth exploring refactoring each supported gfx card type as it's own class wrapping the behaviour of what gpuCapacity to put?

rhtyd:
@nitin-maharana I've left some remarks and I feel the code may be further refactored, thanks.

vmware-base/src/com/cloud/hypervisor/vmware/mo/HostMO.java
nitin-maharana: @rhtyd, I didn't get you properly. I guess you are saying to replace the multiple remainingCapacities assignments present in the if..else statement with a single assignment using index variable by matching the grid type. Am I correct?

Note:
Actually, the reason why I hardcoded those assignments because unlike XenServer, VMware doesn't provide libraries to get the remaining capacities of each vGPU type. To obtain that I used the VM size, which says how many VMs of the particular vGPU type present in the card. Accordingly, it reduces the capacity of each vGPU type(One vGPU type affects the capacity of other vGPU types).

rhtyd: @nitin-maharana my suggestion is around refactoring, all the ifs specific to a graphics card type/name could be wrapped to a class where the behavior settings/values could be moved. It will make it easier in future to add more graphics cards, maintain code etc.

rhtyd:
@nitin-maharana can you rebase against latest master, thanks.

rhtyd:
@nitin-maharana can you rebase against latest master, thanks.

nitin-maharana:
@rhtyd, I am sorry, somehow I missed your comment. Rebased against latest master. Thank You!

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test centos7 vmware-65

rhtyd:
Test LGTM, but I think the code could be further refactored. I'm unable to find time to do that @nitin-maharana but please see parts of the code where you see several ifs-elses, see how you can improve that and also add more tests?

DennisKonrad:
@rhtyd @nitin-maharana @rafaelweingartner 
What do you think about this pull request? This kind of feature (vGPU) will only become more relevant in the future for all of us.

Is there are way to proceed getting things like this into the master?

onitake:
I think GPU passthrough should be much more general than this.
Only supporting a handful of particular NVIDIA products is not very future-proof.
See here: https://www.nvidia.com/en-us/data-center/graphics-cards-for-virtualization/

This article explains different approaches: https://blogs.vmware.com/apps/2018/07/using-gpus-with-virtual-machines-on-vsphere-part-1-overview.html

While the pull request is focused on NVIDIA GRID tech, where GPUs are shared between VMs, PCI passthrough for dedicating a single GPU to a single VM should also be possible.

onitake:
And here's a list of current GPUs supported by NVIDIA GRID: https://www.nvidia.com/object/vgpu-certified-servers.html

Kepler isn't even mentioned any more...

DaanHoogland:
closing this as it is picked up in #3527 (and it needs lo0ve there as well)

rhtyd:
Hi, the PR got auto-closed due to no known source repository or some other Github-specific issue during the recent renaming of the default branch to `main`. If you wish kindly submit a new PR.

PR #2341, opened 2017-11-27
CLOUDSTACK-10160: Fix typo in Libvirt XML definition for Virtio-SCSI

wido:
The attribute for the XML element 'controller' should be 'model' and
not 'mode'.

Source: https://libvirt.org/formatdomain.html#elementsControllers

  A scsi controller has an optional attribute model, which is one of
  'auto', 'buslogic', 'ibmvscsi', 'lsilogic', 'lsisas1068', 'lsisas1078',
  'virtio-scsi' or 'vmpvscsi'.

In the current state a regular SCSI device is attached and not a Virtio-SCSI
device.

Afterwards this is what *lspci* reports:

<pre>
root@ub-scsi:~# lspci |grep -i scsi
00:09.0 SCSI storage controller: Red Hat, Inc Virtio SCSI
root@ub-scsi:~# 
</pre>

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
@@ -1411,10 +1411,10 @@ public class LibvirtVMDef {
         public String toString() {
             StringBuilder scsiBuilder = new StringBuilder();
 
-            scsiBuilder.append(String.format("<controller type='scsi' index='%d' mode='virtio-scsi'>\n", this.index ));
+            scsiBuilder.append(String.format("<controller type='scsi' index='%d' model='virtio-scsi'>\n", this.index ));
             scsiBuilder.append(String.format("<address type='pci' domain='0x%04X' bus='0x%02X' slot='0x%02X' function='0x%01X'/>\n",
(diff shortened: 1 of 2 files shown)


wido:
Preferably this one would be merged to 4.10 as well so that we could fix it in a hotfix there.

Otherwise 4.11

rhtyd:
@wido the 4.10 maintainer may choose to backport (cherry-pick it)
@blueorangutan package

wido:
Thanks @rhtyd !

I've also added a UnitTest in a second commit to make sure this is tested.

rhtyd:
@blueorangutan test

wido:
@nathanejohnson: Not to point a finger, but without this it never worked properly. Although it seemed to work as VirtIO-SCSI it was actually just plain SCSI.

nathanejohnson:
@wido when using linux, it recognizes as a VIRTIO-Scsi HBA, and when spinning up Windows guests with the Virtio SCSI drivers installed it recognizes the storage appropriately.  That doesn't "prove" it's actually using it as virtio-scsi, but it's possible when the "model" is not specified (or incorrectly specified) it might default to "auto", which might do the right thing based on OS support?  I'm guessing here. 

screen shot 2017-11-27 at 10 34 15 am

rhtyd:
Lgtm, any further improvements may be submitted in separate PRs. Merging now, thanks.

PR #2345, opened 2017-11-29
CLOUDSTACK-10164: bugfix allow users to create a VPC through the UI.

sgoeminn:
https://issues.apache.org/jira/browse/CLOUDSTACK-10164

### DIFF
--- a/plugins/network-elements/nuage-vsp/src/com/cloud/network/manager/NuageVspManagerImpl.java
+++ b/plugins/network-elements/nuage-vsp/src/com/cloud/network/manager/NuageVspManagerImpl.java
@@ -57,6 +57,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


rafaelweingartner:
@sgoeminn can you provide a bit more details regarding the problem? Does it only affect 4.10? Specific to a deployment scenario?

sgoeminn:
@rafaelweingartner The problem only occurs in current master. So 4.10 is not affected. To my knowledge all native (non-nuage) CloudStack deployments are affected. During testing another feature this bug came to the surface. So apparently the listNuageVspDomainTemplat cmd was called by the ui when the create vpc dialog was opened. This command failed when no nuage vsp device was present. As a consequence the ui did not show the dialog. So currently it's not possible to create a vpc through the UI with a native CloudStack deployment. This bugfix, adds robustness to the ui that in case the cmd fails the dialog will appear nonetheless . Furthermore I also changed the listNuageVspDomainTemplateCmd to always return an empty result when there is no nuage zone.

plugins/network-elements/nuage-vsp/src/com/cloud/network/manager/NuageVspManagerImpl.java
```diff
@@ -1168,7 +1178,7 @@ private Long getNuageVspHostId(long physicalNetworkId) {
             return config.getHostId();
         }
 
-        throw new CloudRuntimeException("There is no Nuage VSP device configured on physical network " + physicalNetworkId);
```
rafaelweingartner: What about keeping the message here as a log entry?

rafaelweingartner:
Thanks for the details!

What do you think about calling the `listNuageVspDomainTemplateCmd ` when there is no Nuage zone? Shouldn't we only call that method when ACS is deployed with Nuage cofigurations?

sgoeminn:
@rafaelweingartner Thanks, I agree it would not be necessary if there is no nuage zone, however to find that out we would probably need to do more API calls. Making the listNuageVsp.. function returning an empty result fast looked like a better more performance wise solution to me?

rafaelweingartner:
Got it. 
I asked that because I noticed that we have an Ajax inside a success function of another Ajax call. I checked your Java code and I am ok with it; however, the JavaScript is something that needs love, but I know that you did not create that, it was already like this before ;)

rhtyd:
Lgtm, test lgtm as well. I was able to reproduce the issue on latest master. Merging based on nuage plugin tests and Travis, changes are specific to nuage plugin and trillian tests are not necessary.  Thanks.

PR #2348, opened 2017-11-30
[CLOUDSTACK-10196] Remove ejb-api 3.0 dependency

rafaelweingartner:
This PR is proposing the removal of "ejb-api" 3.0 JAR. We use spring to manage objects creation and injection. Thus, spring beans do not need the annotation @Local. The @Component or the bean declaration in the XML is enough to wire the object in Spring life-cycle.

### DIFF
--- a/server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
+++ b/server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
@@ -27,7 +27,6 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Random;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -35,7 +34,6 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 
-import javax.ejb.ConcurrentAccessException;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
@@ -61,7 +59,6 @@ import com.cloud.agent.api.SecurityGroupRulesCmd;
 import com.cloud.agent.api.SecurityGroupRulesCmd.IpPortAndProto;
 import com.cloud.agent.manager.Commands;
 import com.cloud.api.query.dao.SecurityGroupJoinDao;
-import com.cloud.api.query.vo.SecurityGroupJoinVO;
 import com.cloud.configuration.Config;
 import com.cloud.domain.dao.DomainDao;
 import com.cloud.event.ActionEvent;
@@ -95,7 +92,6 @@ import com.cloud.utils.Pair;
 import com.cloud.utils.component.ManagerBase;
 import com.cloud.utils.concurrency.NamedThreadFactory;
 import com.cloud.utils.db.DB;
-import com.cloud.utils.db.Filter;
 import com.cloud.utils.db.GlobalLock;
 import com.cloud.utils.db.Transaction;
 import com.cloud.utils.db.TransactionCallback;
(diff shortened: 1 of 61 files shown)


rafaelweingartner:
Ah yeah, actually that depends.. that is why I mentioned that @component or XML declaration is enough. @Local is from a whole different framework that may be seen as an alternative to Spring, and that we do not use.

@marcaurele thanks for the review!

rhtyd:
Good cleanup, we'll need some extended testing on this one.

rhtyd:
@rafaelweingartner can you fix the conflicts?

rafaelweingartner:
@rhtyd done!

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
LGTM, test LGTM as well. The failures are known issues. Based on reviews and test results, I'll merge this.

server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
```diff
@@ -437,10 +434,6 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
                     }
                 }
             });
-
-            for (Long vmId : affectedVms) {
```
rhtyd: @rafaelweingartner why have you removed this?

server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
```diff
@@ -1207,20 +1183,6 @@ private void cleanupUnfinishedWork() {
         }
     }
 
-    private void processScheduledWork() {
```
rhtyd: The usage of this method is commented, should we remove that too or keep this?

rhtyd:
@rafaelweingartner can you address outstanding queries, following which we can merge this.

server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
rafaelweingartner: For me, code commented, dead codes, or code not used should always be deleted.

server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
rafaelweingartner: Because it was not making much sense to me; for every VM ID in `affectedVms` it executes an asynchronous task  `_executorPool.schedule(new WorkerThread(), delayMs, TimeUnit.MILLISECONDS);`. The odd part is that the `vmId` is not used. All of the tasks are scheduled to be executed are the same.

Looking at com.cloud.network.security.SecurityGroupManagerImpl.work() closely, it seems that it is indeed used for something. I was expecting it to be a generic task where we send something to be computed on. 

So, there this idea of security group work, and the guys that implemented this, probably wanted to execute/process a number of work that is equal to the number of `affectedVms`. It would have been clear if done in some other way… I will not refactor this.
I will re-added this bit again.

@rhtyd  thanks for spotting that!

rafaelweingartner:
@rhtyd I replied your inquiries.
Can you take a look at them?

Thanks for the feedback here!

server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
rhtyd: +1

rhtyd:
@rafaelweingartner thanks, no need to squash yourself. I'll squash merge using Github. I'll re-review the changes, and kick another round of tests.

server/src/com/cloud/network/security/SecurityGroupManagerImpl.java
rhtyd: Alright, see line 211 in CleanupThread, maybe remove the commented method as well?

rhtyd:
LGTM
@blueorangutan package

rhtyd:
@blueorangutan test

rafaelweingartner:
@rhtyd commits squashed.
Thanks!

rhtyd:
@rafaelweingartner do we have a jira ID for this?

rafaelweingartner:
Hmm, no...
I will create one now

rafaelweingartner:
@rhtyd done

rhtyd:
Merged based on review and test results.

PR #2346, opened 2017-11-29
Add XenServer 7.1 and 7.2 interoperablility

PaulAngus:
Add XenServer 7.1 and 7.2 interoperablility
Add hypervisor to capabilities list and add guest OS mappings (copied from 7.0.0 mappings initially)
Deprecated mappings not added

links to
https://issues.apache.org/jira/browse/CLOUDSTACK-9839

PR #2365, opened 2017-12-18
CLOUDSTACK-10197: Rename xentools iso for XenServer 7.0+

khos2ow:
The xentools iso has been renamed from xs-tools to guest-tools
starting from XenServer 7.0.

This PR is the follow up of #2346

@Reviewers: The main change is in `getIsoVDIByURL` method [line ~2589] the rest is autoformatting.

### DIFF
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
@@ -536,7 +536,7 @@ public class XcpServerDiscoverer extends DiscovererBase implements Discoverer, L
             id = _tmpltDao.getNextInSequence(Long.class, "id");
             VMTemplateVO template =
                     VMTemplateVO.createPreHostIso(id, isoName, isoName, ImageFormat.ISO, true, true, TemplateType.PERHOST, null, null, true, 64, Account.ACCOUNT_ID_SYSTEM,
-                            null, "xen-pv-drv-iso", false, 1, false, HypervisorType.XenServer);
+                            null, "XenServer Tools Installer ISO (xen-pv-drv-iso)", false, 1, false, HypervisorType.XenServer);
             _tmpltDao.persist(template);
         } else {
             id = tmplt.getId();
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
@@ -2592,9 +2592,10 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
         String mountpoint = null;
         if (isoURL.startsWith("xs-tools")) {
             try {
-                final Set<VDI> vdis = VDI.getByNameLabel(conn, isoURL);
+                final String actualIsoURL = actualIsoTemplate(conn);
+                final Set<VDI> vdis = VDI.getByNameLabel(conn, actualIsoURL);
                 if (vdis.isEmpty()) {
-                    throw new CloudRuntimeException("Could not find ISO with URL: " + isoURL);
+                    throw new CloudRuntimeException("Could not find ISO with URL: " + actualIsoURL);
                 }
                 return vdis.iterator().next();
 
@@ -2628,6 +2629,22 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
         } else {
             throw new CloudRuntimeException("Could not find ISO with URL: " + isoURL);
         }
+    }
+
+    private String actualIsoTemplate(final Connection conn) throws BadServerResponse, XenAPIException, XmlRpcException {
+        final Host host = Host.getByUuid(conn, _host.getUuid());
+        final Host.Record record = host.getRecord(conn);
+        final String xenBrand = record.softwareVersion.get("product_brand");
+        final String xenVersion = record.softwareVersion.get("product_version");
+        final String[] items = xenVersion.split("\\.");
+
+        // guest-tools.iso for XenServer version 7.0+
+        if (xenBrand.equals("XenServer") && Integer.parseInt(items[0]) >= 7) {
+            return "guest-tools.iso";
+        }
+
+        // xs-tools.iso for older XenServer versions
+        return "xs-tools.iso";
     }
 
     public String getLabel() {
@@ -3882,9 +3899,10 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
             final String templateName = iso.getName();
             if (templateName.startsWith("xs-tools")) {
                 try {
-                    final Set<VDI> vdis = VDI.getByNameLabel(conn, templateName);
+                    final String actualTemplateName = actualIsoTemplate(conn);
+                    final Set<VDI> vdis = VDI.getByNameLabel(conn, actualTemplateName);
                     if (vdis.isEmpty()) {
-                        throw new CloudRuntimeException("Could not find ISO with URL: " + templateName);
+                        throw new CloudRuntimeException("Could not find ISO with URL: " + actualTemplateName);
                     }
                     return vdis.iterator().next();
                 } catch (final XenAPIException e) {


rhtyd:
Lgtm
@blueorangutan package

rhtyd:

rhtyd:
I see new iso related failures, rest are known issues. Can you check @khos2ow

khos2ow:
@rhtyd is there a way to see the error log?

rhtyd:
@khos2ow yes, download the marvin logs from the test above, extract and see the logs for that test to get hints etc.

khos2ow:
@rhtyd damn! I haven't seen that link!!! thanks.

rhtyd:
@khos2ow can you fix the build failures?

khos2ow:
@rhtyd should be fine now, there was a code-style issue which I fixed it.

rhtyd:

rhtyd:
@khos2ow the new changes causes management server to fail on start, in a fresh installation. I get:

```
2017-12-21 12:47:53,542 INFO  [c.c.h.x.r.XenServerConnectionPool] (main:null) (logid:) XenServer Connection Pool Configs: sleep.interval.on.error=10000
2017-12-21 12:47:53,568 DEBUG [c.c.a.m.AgentManagerImpl] (main:null) (logid:) Registering listener XcpServerDiscoverer with id 25
2017-12-21 12:47:53,590 DEBUG [c.c.u.d.T.Transaction] (main:null) (logid:) Rolling back the transaction: Time = 4 Name =  persist; called by -TransactionLegacy.rollback:889-TransactionLegacy.removeUpTo:832-TransactionLegacy.close:656-TransactionContextInterceptor.invoke:36-ReflectiveMethodInvocation.proceed:174-ExposeInvocationInterceptor.invoke:92-ReflectiveMethodInvocation.proceed:185-JdkDynamicAopProxy.invoke:212-$Proxy109.persist:-1-XcpServerDiscoverer.createXsToolsISO:550-XcpServerDiscoverer.configure:500-CloudStackExtendedLifeCycle$3.with:115
2017-12-21 12:47:53,591 WARN  [o.a.c.s.m.c.ResourceApplicationContext] (main:null) (logid:) Exception encountered during context initialization - cancelling refresh attempt: org.springframework.context.ApplicationContextException: Failed to start bean 'cloudStackLifeCycle'; nested exception is com.cloud.utils.exception.CloudRuntimeException: DB Exception on: com.mysql.jdbc.JDBC4PreparedStatement@1f60c30f: INSERT INTO vm_template (vm_template.id, vm_template.format, vm_template.unique_name, vm_template.name, vm_template.public, vm_template.featured, vm_template.type, vm_template.url, vm_template.hvm, vm_template.bits, vm_template.created, vm_template.account_id, vm_template.checksum, vm_template.display_text, vm_template.enable_password, vm_template.guest_os_id, vm_template.bootable, vm_template.prepopulate, vm_template.cross_zones, vm_template.hypervisor_type, vm_template.extractable, vm_template.source_template_id, vm_template.state, vm_template.template_tag, vm_template.uuid, vm_template.sort_key, vm_template.enable_sshkey, vm_template.size, vm_template.update_count, vm_template.updated, vm_template.dynamically_scalable) VALUES (201, 'ISO', null, null, 1, 1, 'PERHOST', null, 1, 64, '2017-12-21 12:47:53', 1, null, _binary'xen-pv-drv-iso', 0, 1, 0, 0, 0, 'XenServer', 1, null, 'Active', null, _binary'a17c3900-50a7-4847-ab10-bdf309245aaf', 0, 0, null, 0, null, 0)
2017-12-21 12:47:53,592 WARN  [o.e.j.w.WebAppContext] (main:null) (logid:) Failed startup of context o.e.j.w.WebAppContext@49c43f4e{/client,file:///usr/share/cloudstack-management/webapp/,UNAVAILABLE}{/usr/share/cloudstack-management/webapp}
org.springframework.context.ApplicationContextException: Failed to start bean 'cloudStackLifeCycle'; nested exception is com.cloud.utils.exception.CloudRuntimeException: DB Exception on: com.mysql.jdbc.JDBC4PreparedStatement@1f60c30f: INSERT INTO vm_template (vm_template.id, vm_template.format, vm_template.unique_name, vm_template.name, vm_template.public, vm_template.featured, vm_template.type, vm_template.url, vm_template.hvm, vm_template.bits, vm_template.created, vm_template.account_id, vm_template.checksum, vm_template.display_text, vm_template.enable_password, vm_template.guest_os_id, vm_template.bootable, vm_template.prepopulate, vm_template.cross_zones, vm_template.hypervisor_type, vm_template.extractable, vm_template.source_template_id, vm_template.state, vm_template.template_tag, vm_template.uuid, vm_template.sort_key, vm_template.enable_sshkey, vm_template.size, vm_template.update_count, vm_template.updated, vm_template.dynamically_scalable) VALUES (201, 'ISO', null, null, 1, 1, 'PERHOST', null, 1, 64, '2017-12-21 12:47:53', 1, null, _binary'xen-pv-drv-iso', 0, 1, 0, 0, 0, 'XenServer', 1, null, 'Active', null, _binary'a17c3900-50a7-4847-ab10-bdf309245aaf', 0, 0, null, 0, null, 0)
        at org.springframework.context.support.DefaultLifecycleProcessor.doStart(DefaultLifecycleProcessor.java:186)
        at org.springframework.context.support.DefaultLifecycleProcessor.access$200(DefaultLifecycleProcessor.java:52)
        at org.springframework.context.support.DefaultLifecycleProcessor$LifecycleGroup.start(DefaultLifecycleProcessor.java:358)
        at org.springframework.context.support.DefaultLifecycleProcessor.startBeans(DefaultLifecycleProcessor.java:159)
        at org.springframework.context.support.DefaultLifecycleProcessor.onRefresh(DefaultLifecycleProcessor.java:123)

```

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
```diff
@@ -529,13 +540,12 @@ public boolean processCommands(long agentId, long seq, Command[] commands) {
     }
 
     private void createXsToolsISO() {
-        String isoName = "xs-tools.iso";
-        VMTemplateVO tmplt = _tmpltDao.findByTemplateName(isoName);
+        VMTemplateVO tmplt = _tmpltDao.findByTemplateName(_toolsIsoName);
```
khos2ow: The problem is at this line `_toolsIsoName` is not initialized yet (basically `cretaeXsToolsISO` is executed before newly added `determineToolsIsoName` method). My other idea is at this line we can register both `xs-tools.iso` and `guest-tools.iso` and either:
- later on (for example in `createServerResource` method) delete/deactivate the _wrong_ iso 
- present both of them on UI with different names and let the user select the _correct_ iso based on the version of their Xen

@rhtyd @wido @syed @rafaelweingartner @swill any comment, suggestion, concern?

rafaelweingartner: I would start with a different question. Do we need to check which XenServers versions are used? I mean, Can’t we simply persist both ISOs in the database? Without needing to use that “determine” method.

It seems that as long as we have both entries in the database (or new ones in the far far future ahead of us), your logic in ` CitrixResourceBase.java ` would take care of the rest. I would recommend one thing though. You have the strings "guest-tools" and "xs-tools" in two different classes (at least the ones I am seeing in this PR). What about extracting them to an enum?

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
```diff
@@ -529,13 +540,12 @@ public boolean processCommands(long agentId, long seq, Command[] commands) {
...
         Long id;
         if (tmplt == null) {
             id = _tmpltDao.getNextInSequence(Long.class, "id");
             VMTemplateVO template =
-                    VMTemplateVO.createPreHostIso(id, isoName, isoName, ImageFormat.ISO, true, true, TemplateType.PERHOST, null, null, true, 64, Account.ACCOUNT_ID_SYSTEM,
+                    VMTemplateVO.createPreHostIso(id, _toolsIsoName, _toolsIsoName, ImageFormat.ISO, true, true, TemplateType.PERHOST, null, null, true, 64, Account.ACCOUNT_ID_SYSTEM,
```
rafaelweingartner: I do not know what you think about the use of "_" here to denote object attributes, but I do not see any benefit on using it. As a matter of fact, I see most PRs removing them.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
```diff
@@ -2590,7 +2590,9 @@ public SR getISOSRbyVmName(final Connection conn, final String vmName) {
     public VDI getIsoVDIByURL(final Connection conn, final String vmName, final String isoURL) {
         SR isoSR = null;
         String mountpoint = null;
-        if (isoURL.startsWith("xs-tools")) {
+        // XenServer 7.0+ => guest-tools.iso
+        // XenServer [other] => xs-tools.iso
+        if (isoURL.startsWith("xs-tools") || isoURL.startsWith("guest-tools")) {
```
rafaelweingartner: Is it possible for `isoURL` to be null? If not, it is ok the way it is. Otherwise, it might be a good idea to use StringUtils.startsWith

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
khos2ow: Yes we do, that's the whole point of this PR. The tools ISO has been renamed. So the question will be we _auto detect_ it or provide the options for user to select.

+1 on extracting to enum though, will do that.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
khos2ow: I, honestly, don't like it and have never used it like this, but wanted to follow the same pattern in the code base.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
khos2ow: +1 I doubt it will be null, but checking for null almost always is good. I'll add it.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
rafaelweingartner: It is always good to create a new trend :)

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
rafaelweingartner: Can't we create a more transparent mode? I mean, the user sees just the "Xen-server-tools", and then underneath we select the right name according to the XenServer version.

khos2ow: That's what I was trying to do but the way I did it was wrong, hence the first option mentioned above:
 > we can register both xs-tools.iso and guest-tools.iso and ... later on (for example in createServerResource method) delete/deactivate the wrong iso

rhtyd:
@khos2ow do make suitable changes and let me know when you next want to kick tests.

khos2ow:
@rhtyd I will, I'm trying to figure out a good way to do it. Let you know when done.

rhtyd:
Thanks @khos2ow any updates?

khos2ow:
@rhtyd @rafaelweingartner the new change:
- registers both iso names at `ComponentLifecycle`#`configure` time
- and later on `createServerResource` removes the _wrong_ one

Any comments, suggestions?

rafaelweingartner:
What are the assumptions here?
I mean, is it possible to have clusters/hosts with different versions of XenServer at the same time?
If so, we should work with both naming conventions at the same time, right?

What about something different? The user does not need to know the version of XenServer the cloud providers use. We could present a “dummy” ISO option called “XenServer tools. Then, underneath, when plugging the ISO to the VM, we check the host the VM is running and select/find/discover the right VDI that represent the XenServer tools ISO to be connected to the VM (according to host’s XenServer version). This would require only a single entry for any type of XenServer tools.

khos2ow:
@rafaelweingartner I don't know if we can have a clusters/hosts with different XS versions, but in any case I changed it again as follow
- register only *one* iso with `name` of **xs-tools.iso** and `display_text` of **XenServer Installer ISO (xen-pv-drv-iso)**
- get the actual iso name (either _xs-tools.iso_ or _guest-tools.is_) on the fly based on version of `com.xensource.xenapi.Host.Record`

in the future, if the name of iso changes again, only another `else` block will be needed in `CitrixResourceBase#actualIsoTemplate`

rhtyd:
@blueorangutan package

rhtyd:

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
```diff
@@ -2592,9 +2592,10 @@ public VDI getIsoVDIByURL(final Connection conn, final String vmName, final Stri
         String mountpoint = null;
         if (isoURL.startsWith("xs-tools")) {
```
rafaelweingartner: This ISO URL remained the same?

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
```diff
@@ -536,7 +536,7 @@ private void createXsToolsISO() {
             id = _tmpltDao.getNextInSequence(Long.class, "id");
             VMTemplateVO template =
                     VMTemplateVO.createPreHostIso(id, isoName, isoName, ImageFormat.ISO, true, true, TemplateType.PERHOST, null, null, true, 64, Account.ACCOUNT_ID_SYSTEM,
-                            null, "xen-pv-drv-iso", false, 1, false, HypervisorType.XenServer);
+                            null, "XenServer Tools Installer ISO (xen-pv-drv-iso)", false, 1, false, HypervisorType.XenServer);
```
rafaelweingartner: This will create an entry for deployments that do not have it yet using this new "standardized name". However, for environment that already had it from previous versions, the name (display_text) will not be updated, right?

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
```diff
@@ -2630,6 +2631,22 @@ public VDI getIsoVDIByURL(final Connection conn, final String vmName, final Stri
         }
     }
 
+    private String actualIsoTemplate(final Connection conn) throws BadServerResponse, XenAPIException, XmlRpcException {
```
rafaelweingartner: Would you mind creating a documentation for the method? Then, you do not need comments in line.
Also, this is a good method to create a unit test.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
khos2ow: That is correct, I can explicitly add `tmplt.setDisplayText("XenServer Tools Installer ISO (xen-pv-drv-iso)");` at line 544 (couple lines down here)

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
khos2ow: Yes it did. I had an idea to change the iso name (url here refers to _name_) to something general - like `xenserver-tools` and manage the same thing in `actualIsoTemplate`, but _xs-tools.iso_ is used in some many places (specially in some test cases) which I ended up not changing this. Potentially it only needs a wiki/user guide doc only.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/discoverer/XcpServerDiscoverer.java
rafaelweingartner: I think this is the right thing to do, so we get consistency.
You have to save the `tmplt` as well right? So, you update the data in the database

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
khos2ow: I will the documentation, but I really don't think unit testing on this specific method will be possible. Because it needs a live XenServer Host for the method to connect to (both `Host.getByUuid(conn, _host.getUuid());` and `host.getRecord(conn);`) and I'm not sure if they can be Mocked in some way.

rafaelweingartner: We do not need a real host. We only want to test the logic. Therefore, you can mock the method calls. PowerMockito and Mockito are there for this kind of job.

rhtyd:
@khos2ow can you address outstanding issues?

rhtyd:
@blueorangutan package

khos2ow:
@rhtyd I don't think the failed tests are related to the change. The change was in attaching ISO. Can you re-run the test?

rhtyd:
@blueorangutan package

rhtyd:
@khos2ow I've kicked another test round against xenserver 7.1

rhtyd:

rhtyd:
Test LGTM for 7.1 comparing baseline results against #2376 I'll merge this after xs-6.5sp1 results are back.

PR #2370, opened 2017-12-20
CLOUDSTACK-9595: Fix regression introduced in #1762

rhtyd:
The `assignDedicateIpAddress` previously had marked the newly fetched
IP as allocated but now it does not do that. This fails for VPCs
where SNATs IP are retained as allocating and not allocated after
creation.

@yvsubhash @rafaelweingartner and please review.
@blueorangutan package

### DIFF
--- a/server/src/com/cloud/network/IpAddressManagerImpl.java
+++ b/server/src/com/cloud/network/IpAddressManagerImpl.java
@@ -291,8 +291,8 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
 
     SearchBuilder<IPAddressVO> AssignIpAddressSearch;
     SearchBuilder<IPAddressVO> AssignIpAddressFromPodVlanSearch;
-    private final Object _allocatedLock = new Object();
-    private final Object _allocatingLock = new Object();
+    private static final Object allocatedLock = new Object();
+    private static final Object allocatingLock = new Object();
 
     static Boolean rulesContinueOnErrFlag = true;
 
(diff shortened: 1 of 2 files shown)


server/src/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -936,8 +942,8 @@ public PublicIp assignDedicateIpAddress(Account owner, final Long guestNtwkId, f
                 displayIp = vpc.isDisplay();
             }
 
-            return fetchNewPublicIp(dcId, null, null, owner, VlanType.VirtualNetwork, guestNtwkId, isSourceNat, false, null, false, vpcId, displayIp);
-
+            ip = fetchNewPublicIp(dcId, null, null, owner, VlanType.VirtualNetwork, guestNtwkId, isSourceNat, true, null, false, vpcId, displayIp);
```
rafaelweingartner: You can return directly the call to `fetchNewPublicIp`.

I have not checked. What is that boolean flag you changed there?

rhtyd: @rafaelweingartner actually there is a `ip` variable that is used in `finally` which is why I assign and then return see the code below this line.

rafaelweingartner: ok, no I saw it.
Thanks

rhtyd:
@blueorangutan test

yvsubhash:
@rhtyd  Code changes are fine LGTM for code

rhtyd:
Thanks for confirming @yvsubhash, I'll merge the fix once test results arrive.

rhtyd:
Tests LGTM, the new vpc vpn failure is due to create vpn user failing. I'll merge this based on tests and two code reviews.

PR #2372, opened 2017-12-20
Cloudstack-10201: catalina log rotate

Slair1:
Receive following message in "var/spool/mail/root:

### /etc/cron.daily/logrotate:

error: skipping "/var/log/cloudstack/management/catalina.out" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.

### As error states, group is "cloud":

[root@labcloud01b ~]# ls -l /var/log/cloudstack/
total 64
drwxrwx---. 2 root root 6 Dec 18 23:41 ipallocator
drwxrwx---. 2 root cloud 28672 Dec 20 14:35 management
drwxrwx---. 2 root cloud 12288 Dec 20 00:15 usage

Need to update "cloudstack-catalina.logrotate" with su info


rhtyd:
@Slair1 current master and starting 4.11, we'll have embedded Jetty so please check if this is still relevant for current master?

Slair1:
Thanks @rhtyd you are correct, this is no longer an issue with embedded Jetty.  The file I'm editing is still in master, but isn't installed any longer by cloud.spec

PR #2373, opened 2017-12-21
CLOUDSTACK-10202:createSnapshotPolicy API create multiple entries in DB for same volume.

niteshsarda:
**ISSUE :**
createSnapshotPolicy API create multiple entries in DB for same parameters, if multiple threads are executed in parallel.

**STEPS TO REPRODUCE :**
1. Created a new machine having root and data disk.
2. Make sure that no existing snapshot policy is present for the volume.
3. Execute multiple threads in parallel for createSnapshotPolicy API having all required parameters exactly same.
4. Verify table snapshot_policy in DB, will get multiple entries for same policy.
5. Once again execute same multiple threads, by changing any API parameter, will see that existing entries are getting modified in DB and no new entries are added.

### DIFF
--- a/server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
+++ b/server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
@@ -17,6 +17,7 @@
 package com.cloud.storage.snapshot;
 
 import com.cloud.agent.api.Answer;
+import com.cloud.utils.db.GlobalLock;
 import com.cloud.agent.api.Command;
 import com.cloud.agent.api.DeleteSnapshotsDirCommand;
 import com.cloud.alert.AlertManager;
@@ -785,6 +786,7 @@ public class SnapshotManagerImpl extends MutualExclusiveIdsManagerBase implement
     public SnapshotPolicyVO createPolicy(CreateSnapshotPolicyCmd cmd, Account policyOwner) {
         Long volumeId = cmd.getVolumeId();
         boolean display = cmd.isDisplay();
+        SnapshotPolicyVO policy = null;
         VolumeVO volume = _volsDao.findById(cmd.getVolumeId());
         if (volume == null) {
             throw new InvalidParameterValueException("Failed to create snapshot policy, unable to find a volume with id " + volumeId);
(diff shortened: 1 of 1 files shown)


server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
```diff
@@ -854,15 +856,18 @@ public SnapshotPolicyVO createPolicy(CreateSnapshotPolicyCmd cmd, Account policy
...
         }
-        SnapshotPolicyVO policy = _snapshotPolicyDao.findOneByVolumeInterval(volumeId, intvType);
-        if (policy == null) {
-            policy = new SnapshotPolicyVO(volumeId, cmd.getSchedule(), timezoneId, intvType, cmd.getMaxSnaps(), display);
-            policy = _snapshotPolicyDao.persist(policy);
-            _snapSchedMgr.scheduleNextSnapshotJob(policy);
-        } else {
-            try {
+
+        final GlobalLock createSnapshotPolicyLock = GlobalLock.getInternLock("createSnapshotPolicy");
```
marcaurele: Would be more effective to append the volume id in the lock name to avoid locking the call for all snapshot policy.

marcaurele:
I'm not convinced that such a fix is needed. In what scenario do you need to send multiple identical requests?

server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
niteshsarda: @marcaurele : Implemented the changes as per your suggestion.

niteshsarda:
@marcaurele : One of our customer have reported this scenario, where they have multiple management servers and when parallel request comes for snapshot policy creation for same volume, then multiple entries are created in DB for same volume. 
As per the code, only one snapshot policy entry should be created in database for each volume. Whenever, another request of snapshot policy comes for the same volume, then existing entry should be updated. Same behavior is observed, when volume have existing snapshot policy and parallel threads are executed, then only single entry is getting updated every time. This is because, we have lock available in code in else block. Whereas, in case of first snapshot policy creation for volume, DB lock is not available, hence it creates multiple DB entries.
As multiple management servers might send API requests in parallel for same volume, we should handle this from our code and allow to create only one entry in DB for each volume.

Also, I apologize for confusing title of the PR, I have updated the same, to make it more clear and relevant to the fix.

rafaelweingartner:
I have experienced the same problem some time ago. I did not manage to reproduce it though. One of the volumes had two "daily snapshots" in the database.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
```diff
@@ -854,15 +855,17 @@ public SnapshotPolicyVO createPolicy(CreateSnapshotPolicyCmd cmd, Account policy
...
         }
-        SnapshotPolicyVO policy = _snapshotPolicyDao.findOneByVolumeInterval(volumeId, intvType);
-        if (policy == null) {
-            policy = new SnapshotPolicyVO(volumeId, cmd.getSchedule(), timezoneId, intvType, cmd.getMaxSnaps(), display);
-            policy = _snapshotPolicyDao.persist(policy);
-            _snapSchedMgr.scheduleNextSnapshotJob(policy);
-        } else {
-            try {
+
+        policy = _snapshotPolicyDao.acquireInLockTable(volumeId);
```
marcaurele: You can request a lock for the snapshot policy table for existing entries and by giving a snapshot policy ID only, not the ID of a volume: `policy = _snapshotPolicyDao.acquireInLockTable(policyId);`
In your case you don't have a policy id yet when creating a new one. You'll have to use the global lock.

niteshsarda: @marcaurele : Changed the code as per your suggestion. Now added a code for using global lock along with volume id added in the lock name.
Can you please review the new code ?

server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
```diff
@@ -854,15 +856,18 @@ public SnapshotPolicyVO createPolicy(CreateSnapshotPolicyCmd cmd, Account policy
...
-        SnapshotPolicyVO policy = _snapshotPolicyDao.findOneByVolumeInterval(volumeId, intvType);
-        if (policy == null) {
-            policy = new SnapshotPolicyVO(volumeId, cmd.getSchedule(), timezoneId, intvType, cmd.getMaxSnaps(), display);
-            policy = _snapshotPolicyDao.persist(policy);
-            _snapSchedMgr.scheduleNextSnapshotJob(policy);
-        } else {
-            try {
+
+        final GlobalLock createSnapshotPolicyLock = GlobalLock.getInternLock("createSnapshotPolicy_" + volumeId);
+        createSnapshotPolicyLock.lock(5);
```
marcaurele: The return value is not checked (`boolean`) which should handle the case when the lock could not be acquired.

niteshsarda: @marcaurele : Do we just need to add the debug logger if lock is not acquired ?
Please confirm.

marcaurele: If the lock is not acquired you have to stop processing the request and return a failure to the user. Look at other part of the code using that lock mechanism.

niteshsarda: @marcaurele : Incorporated the changes suggested by you for handling the condition of lock not acquired. In case if lock is not acquired, null value will be returned.
In CreateSnapshotPolicyCmd class, inside execute method from where this method is called, we have already handled the null condition, and in case of null value exception will be thrown.

Please review the latest code.

server/src/com/cloud/storage/snapshot/SnapshotManagerImpl.java
```diff
@@ -854,33 +856,39 @@ public SnapshotPolicyVO createPolicy(CreateSnapshotPolicyCmd cmd, Account policy
...
                 }
+            } finally {
+                createSnapshotPolicyLock.unlock();
             }
 
+            // TODO - Make createSnapshotPolicy - BaseAsyncCreate and remove this.
+            CallContext.current().putContextParameter(SnapshotPolicy.class, policy.getUuid());
+            return policy;
+        } else {
+            s_logger.debug("Unable to acquire lock for creating snapshot policy of volume : " + volume.getName());
```
marcaurele: This log should be move to the `warn` level

niteshsarda: @marcaurele : Changed the log level to warn.

Please check latest code.

niteshsarda:
@rhtyd : Can you please re-run the test cases again ?

rhtyd:
sure @niteshsarda 
@blueorangutan package

niteshsarda:
@rhtyd : Thank you.

rhtyd:
Re-kick as commit was pushed after last pkg build run
@blueorangutan package

niteshsarda:
@rhtyd : Can you please re-run the test cases again as results for previous execution is not posted yet?

rhtyd:
Tests haven't been kicked yet.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Lgtm. Merging this based on 2 lgtms and test results.

PR #2378, opened 2017-12-27
CLOUDSTACK-10205 make LinkDomainToLdap return UUID instead of internal id

DaanHoogland:
the internal id is not usefull to the user. It is a bug to return it instead of a uuid.
In the process of fixing the above "name" was deprecated in favour of "ldap_domain".

### DIFF
--- a/api/src/org/apache/cloudstack/api/ApiConstants.java
+++ b/api/src/org/apache/cloudstack/api/ApiConstants.java
@@ -705,6 +705,7 @@ public class ApiConstants {
 
     public static final String HAS_ANNOTATION = "hasannotation";
     public static final String LAST_ANNOTATED = "lastannotated";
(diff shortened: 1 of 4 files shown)


rhtyd:
@DaanHoogland I've fixed the title based on description and code (returns internal id than uuid). Actually, the APIs do support listing by internal id (long). I'll kick some tests.
@blueorangutan package

DaanHoogland:
@rhtyd I changed the title back. it describes the situation as it is implemented not as it was (that is more a title of an issue reported)

rhtyd:
@blueorangutan test

rhtyd:
Tests LGTM, additional review requested for merging this.

rhtyd:
@nvazquez @borisstoyanov can you review this?

rhtyd:
Merging this based on code reviews (2x) lgtms and test results.

PR #837, opened 2015-09-16
CLOUDSTACK-8855 Improve Error Message for Host Alert State and reconnect host API.

bvbharatk:
earlier we were eating up exceptions from the lower layer which resulted in improper error messages. fixed partially by throwing exceptions and catching them at the appropriate layer.

This also fixes the host alters. Earlier in the host alerts window we are listing only the alters, this dose not tell why the host when to alert state. This is fixed by adding the alert message to the host alerts.
(changes in AlertManagerImpl.java). 

tested manually.

PR #2380, opened 2017-12-30
CLOUDSTACK-8855 Improve Error Message for Host Alert State and reconnect host API.

DaanHoogland:
this closes #837

### DIFF
--- a/engine/components-api/src/com/cloud/agent/AgentManager.java
+++ b/engine/components-api/src/com/cloud/agent/AgentManager.java
@@ -16,7 +16,6 @@
 // under the License.
 package com.cloud.agent;
 
-import com.cloud.utils.exception.CloudRuntimeException;
 import org.apache.cloudstack.framework.config.ConfigKey;
 
 import com.cloud.agent.api.Answer;
@@ -37,7 +36,7 @@ import com.cloud.resource.ServerResource;
  */
 public interface AgentManager {
     static final ConfigKey<Integer> Wait = new ConfigKey<Integer>("Advanced", Integer.class, "wait", "1800", "Time in seconds to wait for control commands to return",
-        true);
+            true);
 
     public enum TapAgentsAction {
         Add, Del, Contains,
@@ -144,7 +143,7 @@ public interface AgentManager {
 
     public void pullAgentOutMaintenance(long hostId);
 
-    void reconnect(long hostId) throws CloudRuntimeException, AgentUnavailableException;
+    void reconnect(long hostId) throws AgentUnavailableException;
 
     void rescan();
 
--- a/engine/components-api/src/com/cloud/agent/AgentManager.java
+++ b/engine/components-api/src/com/cloud/agent/AgentManager.java
@@ -16,6 +16,7 @@
 // under the License.
 package com.cloud.agent;
 
+import com.cloud.utils.exception.CloudRuntimeException;
 import org.apache.cloudstack.framework.config.ConfigKey;
 
 import com.cloud.agent.api.Answer;
@@ -143,7 +144,7 @@ public interface AgentManager {
 
     public void pullAgentOutMaintenance(long hostId);
 
-    boolean reconnect(long hostId);
+    void reconnect(long hostId) throws CloudRuntimeException, AgentUnavailableException;
 
     void rescan();
 
--- a/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
@@ -180,28 +180,18 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
     @Inject
     ResourceManager _resourceMgr;
 
-    protected final ConfigKey<Integer> Workers = new ConfigKey<Integer>("Advanced", Integer.class, "workers", "5",
-                    "Number of worker threads handling remote agent connections.", false);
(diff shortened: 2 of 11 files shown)


DaanHoogland:
@rafaelweingartner I can't add you as a reviewer. Are you member of the apache organisation?

rhtyd:
@DaanHoogland can you fix the build failures?

rafaelweingartner:
@DaanHoogland that is probably because I am not in the Gitbox experiment.
It is great that you picked this up, I was thinking on doing that myself, but with my current workload I would not have time to do this now.

I would say that here I see the same problem as in #837. It feels so simple to create a new column in "AlertVO" table to add the content. I am not going to repeat the discussion we had there here. So, as long as others are ok with it.

With the current state of the PR I give a -0.

DaanHoogland:
@rafaelweingartner @marcaurele @rhtyd I am not picking this up. I just 're-opened' this because there is no movement. Please send me your PR if you want to see anything changed.

rafaelweingartner:
@DaanHoogland ok...
I will try to squeeze this in next week.

rafaelweingartner:
@DaanHoogland I tried to push these changes to your branch, but it seems that I do not have permissions to do so. Shall I open a PR for your branch?

DaanHoogland:
Hey @rafaelweingartner If I new how i would a few seconds ago have given you access, you know that ;)
please send PRs so I am forced to look at them.

rafaelweingartner:
@DaanHoogland PR created.
@marcaurele do you have some other code change suggestion to apply here?

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -994,33 +994,28 @@ public Answer easySend(final Long hostId, final Command cmd) {
...
         if (host == null || host.getRemoved() != null) {
-            s_logger.warn("Unable to find host " + hostId);
-            return false;
+            throw new CloudRuntimeException("Unable to find host " + hostId);
         }
 
         if (host.getStatus() == Status.Disconnected) {
-            s_logger.info("Host is already disconnected, no work to be done");
-            return true;
+            throw new CloudRuntimeException("Host is already disconnected, no work to be done");
```
marcaurele: This looks wrong to me, why throwing an exception when nothing has to be done?

plugins/network-elements/netscaler/src/com/cloud/network/element/NetscalerElement.java
```diff
@@ -759,7 +759,11 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
         });
         HostVO host = _hostDao.findById(lbDeviceVo.getHostId());
 
-        _agentMgr.reconnect(host.getId());
+        try {
+            _agentMgr.reconnect(host.getId());
+        } catch (Exception e ) {
+            s_logger.info("failed to reconnect host "+host);
```
marcaurele: `warn`

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
rafaelweingartner: I agree on this one. And besides that, if we take a closer look, the guy that proposed this change, was replacing the return of boolean values with exceptions to indicate some error. It makes sense that a false return would generate an exception while trying to reconnect as the case at line 1002. This one was returning true, so it feels that we should not throw an exception here.

rafaelweingartner:
@marcaurele thanks for your feedback here. I already created a PR to Daan's branch.

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -994,33 +980,29 @@ public Answer easySend(final Long hostId, final Command cmd) {
...
-            s_logger.warn("Unable to find host " + hostId);
-            return false;
+        if (host.getRemoved() != null) {
+            throw new CloudRuntimeException("Host has already been removed: " + hostId);
         }
 
         if (host.getStatus() == Status.Disconnected) {
-            s_logger.info("Host is already disconnected, no work to be done");
-            return true;
+            s_logger.debug("Host is already disconnected, no work to be done: " + hostId);
```
rafaelweingartner: @DaanHoogland I made a mistake here. Can you fix it for me?
You need to add a `return;`, right after line 994

engine/components-api/src/com/cloud/agent/AgentManager.java
```diff
@@ -143,7 +143,7 @@
 
     public void pullAgentOutMaintenance(long hostId);
 
-    boolean reconnect(long hostId);
+    void reconnect(long hostId) throws AgentUnavailableException;
```
DaanHoogland: this change is breaking things. We need to look at it. @rafaelweingartner CAManagerImpl.java needs the boolean. should we take this to 4.12? I'm not taking the time now.

rafaelweingartner: Hmm.. Let me try to look into it today...

rafaelweingartner: Is the problem during compilation or execution?

rafaelweingartner: Never mind. Found the problem. I will work on it.

rafaelweingartner: I fixed the problem and opened a PR to send this whole contribution. This makes things easier for me.
PR: #2387

PR #2384, opened 2018-01-03
CLOUDSTACK-10210 cleanup testfile

DaanHoogland:
cleaning up in the @After-method so it gets clean in case of exception as well

### DIFF
--- a/services/secondary-storage/server/test/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResourceTest.java
+++ b/services/secondary-storage/server/test/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResourceTest.java
@@ -32,6 +32,7 @@ import org.powermock.core.classloader.annotations.PrepareForTest;
 import org.powermock.modules.junit4.PowerMockRunner;
 
 import java.io.BufferedWriter;
+import java.io.File;
 import java.io.FileWriter;
 import java.io.StringWriter;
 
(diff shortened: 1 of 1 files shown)


rhtyd:
The changes are restricted to only the unit test, Travis test results/confirmation should be enough for QA-ing the PR.

rhtyd:
Additional review requested, test LGTM.

rhtyd:
@rafaelweingartner do we have your +1/lgtm from ML applicable here?

services/secondary-storage/server/test/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResourceTest.java
```diff
@@ -49,16 +51,22 @@ public void setUp() {
         resource = new NfsSecondaryStorageResource();
     }
 
+    String filename = "testfile";
+    @After
+    public void cleanup() {
+        File remnance = new File(filename);
+        remnance.deleteOnExit();
```
marcaurele: `exists()` check?

marcaurele:
typo in nickname ^^, cc @DaanHoogland

services/secondary-storage/server/test/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResourceTest.java
DaanHoogland: sure, if that is needed. I didn't expect it to be an issue but we can test that ;)

DaanHoogland:
@marcaurele I want to be absolutely sure that the registration for deletion is not skipped by an  exception and I have never tried to to it before a creation. I wouldn't know if that actually works. A JVM (-library) might be so 'clever' as to ignore the registration altogether if it finds it doesn't exist.

services/secondary-storage/server/test/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResourceTest.java
marcaurele: In case someone run in an IDE only 1 of the class tests, it will cause an error. It's best to do the cleanup inside the test method where the file is created.

marcaurele:
From the API docs:
> Deletion will be attempted only for normal termination of the virtual machine, as defined by the Java Language Specification.

The safest bet IMO is to put the delete in the test method which creates the file, and to use `delete` instead of `deleteOnExit`.

DaanHoogland:
@marcaurele I don't agree it would have to have a try-finaly construct to be sure the deletion happens, which is semantically the same as using the @After method.

marcaurele:
@DaanHoogland semantically it's not the same for me as there is more than 1 test in the class. The `@after` will be run too if you only run the second test of that file.

DaanHoogland:
@marcaurele valid point except that the other test is disabled and will never be run. Also my thinking was that the file might be used in more tests. I'll refactor anyway for you exists remark so i might as well add the finally construct.

rafaelweingartner:
@DaanHoogland first, if the other test is disabled, what about deleting it?

Second, I also prefer as mark said, delete (delete, not deleteOnExit) the file right after the assertion. If someone is creating something in the FS in one test method and lazily using in other....well, that is something we have to change.

If the file is used in more methods and only created at that one, it will be deleted anyway, right? I have not checked the docs, but I have the impression the @Before, and @After I executed each and respectively before and after each test method.

DaanHoogland:
@marcaurele @rafaelweingartner how about this one?

rafaelweingartner:
You are using a try/finally because you believe an exception might happen? 
If an exception happen in `swiftWriteMetadataFile` I believe the file will not get written, right? Or, is it possible for that method to throw an IOException while partially creating the file?

DaanHoogland:
@rafaelweingartner I have no idea and I want to be sure. Any I/O operation is suspect and probe to exceptions (not just the creation but also setting rights after creation)

rafaelweingartner:
@DaanHoogland Thanks for the explanations!

rhtyd:
Merging this based on code lgtms and Travis job1 (which run unit tests) test result.

PR #2390, opened 2018-01-05
[CLOUDSTACK-10214] Unable to remove local primary storage

rafaelweingartner:
When enabling the use of local storage ACS will automatically load all local storage configured in the Host and start using them as primary storage to deploy user VMs (if the service offering allows to do so). However, if the operator wants to remove the local storage ACS will throw an exception saying that the removal of local storage is not allowed.Therefore, if one wants to remove a local storage, he/she needs to do a manual intervention in the database and hosts.

This limitation was removed as it was only a logical restriction.

### DIFF
--- a/api/src/main/java/com/cloud/storage/StorageService.java
+++ b/api/src/main/java/com/cloud/storage/StorageService.java
@@ -90,23 +90,16 @@ public interface StorageService {
 
     boolean deleteSecondaryStagingStore(DeleteSecondaryStagingStoreCmd cmd);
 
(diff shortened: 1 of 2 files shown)


rafaelweingartner:
@blueorangutan package

rafaelweingartner:
@borisstoyanov is it possible to understand why the Debian package is failing here?

borisstoyanov:
I've just re-run it and it succeed. @rafaelweingartner

rafaelweingartner:
@blueorangutan package

rafaelweingartner:
@blueorangutan package

PR #2388, opened 2018-01-05
CLOUDSTACK-10212: Update Netmask/Gateway when Changing IPv4 address

wido:
This can otherwise cause problems in Basic Networking where multiple
IPv4 ranges are configured in a POD.

Signed-off-by: Wido den Hollander <wido@widodh.nl>

PR #2393, opened 2018-01-08
CLOUDSTACK-10217: Clean up old MAC addresses from DHCP lease file

wido:
When the IPv4 address of a Instance changes we need to make sure the
old entry is removed from the DHCP lease file on the Virtual Router
otherwise the Instance will still get the old lease.

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs_dhcp.py
+++ b/systemvm/debian/opt/cloud/bin/cs_dhcp.py
@@ -25,12 +25,18 @@ def merge(dbag, data):
         if data['ipv4_address'] in dbag:
             del(dbag[data['ipv4_address']])
     else:
-        remove_key = None
+        remove_keys = set()
         for key, entry in dbag.iteritems():
             if key != 'id' and entry['host_name'] == data['host_name']:
-                remove_key = key
+                remove_keys.add(key)
                 break
-        if remove_key is not None:
+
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

wido:
I understand your comment @DaanHoogland , but I have no idea on how to test this :)

We ran into this issue combined with #2388 and then needed to fix this in the Python code, this is what it resulted in.

rhtyd:
@wido PR #2388 is not marked for 4.11, are you able to reproduce the fix for this PR with a new 4.11 debian9 systemvmtemplate?

wido:
I see I forgot the label on #2388 @rhtyd , my bad!

The cs_dhcp.py file in this case hasn't changed with the Debian 9 template so it's still valid.

DaanHoogland:
@wido is #2388 a prerequisite for this?

wido:
@DaanHoogland No, it's not. Changing the IP 'works' in 4.10/master, but it doesn't change the gateway.

I discovered these duplicate MACs during #2388

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Test lgtm, and code lgtm. Considering Daan's review as lgtm, merging based on two lgtms, and purely on test results.

PR #2397, opened 2018-01-09
CLOUDSTACK-10221: Allow IPv6 when creating a Basic Network

wido:
Since CloudStack 4.10 Basic Networking supports IPv6 and thus
should be allowed to be specified when creating a network.


nitin-maharana:
Code LGTM.

PR #2398, opened 2018-01-09
CLOUDSTACK-10222: Clean previous snaphosts from primary storage when …

khos2ow:
…taking new one

When user creates a snapshot (manual or recurring), snapshot remains on
the primary storage, even if the snapshot is transferred successfully to
secondary storage. This is causing issues because XenServer can only hold
a limited number of snapshots in its VDI chain, preventing the user from
creating new snapshots after some time, when too many old snapshots are
present on the primary storage.

### DIFF
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
@@ -559,11 +559,11 @@ public class Xenserver625StorageProcessor extends XenServerStorageProcessor {
                     physicalSize = Long.parseLong(tmp[1]);
                     finalPath = folder + File.separator + snapshotBackupUuid + ".vhd";
                 }
-
-                final String volumeUuid = snapshotTO.getVolume().getPath();
-
-                destroySnapshotOnPrimaryStorageExceptThis(conn, volumeUuid, snapshotUuid);
             }
+
+            // remove every snapshot except this one from primary storage
+            final String volumeUuid = snapshotTO.getVolume().getPath();
+            destroySnapshotOnPrimaryStorageExceptThis(conn, volumeUuid, snapshotUuid);
 
             final SnapshotObjectTO newSnapshot = new SnapshotObjectTO();
             newSnapshot.setPath(finalPath);
@@ -577,12 +577,13 @@ public class Xenserver625StorageProcessor extends XenServerStorageProcessor {
             s_logger.info("New snapshot physical utilization: "+physicalSize);
 
             return new CopyCmdAnswer(newSnapshot);
-        } catch (final Types.XenAPIException e) {
-            details = "BackupSnapshot Failed due to " + e.toString();
-            s_logger.warn(details, e);
         } catch (final Exception e) {
-            details = "BackupSnapshot Failed due to " + e.getMessage();
+            final String reason = e instanceof Types.XenAPIException ? e.toString() : e.getMessage();
+            details = "BackupSnapshot Failed due to " + reason;
             s_logger.warn(details, e);
+
+            // remove last bad primary snapshot when exception happens
+            destroySnapshotOnPrimaryStorage(conn, snapshotUuid);
         }
 
         return new CopyCmdAnswer(details);
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/XenServerStorageProcessor.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/XenServerStorageProcessor.java
@@ -1065,7 +1065,7 @@ public class XenServerStorageProcessor implements StorageProcessor {
         return false;
     }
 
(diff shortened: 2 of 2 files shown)


plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -576,13 +578,20 @@ public Answer backupSnapshot(final CopyCommand cmd) {
             s_logger.info("New snapshot details: " + newSnapshot.toString());
             s_logger.info("New snapshot physical utilization: "+physicalSize);
 
+            successful = true;
+
             return new CopyCmdAnswer(newSnapshot);
         } catch (final Types.XenAPIException e) {
             details = "BackupSnapshot Failed due to " + e.toString();
             s_logger.warn(details, e);
         } catch (final Exception e) {
```
rafaelweingartner: It seems that this generic exception can be removed.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -576,13 +578,20 @@ public Answer backupSnapshot(final CopyCommand cmd) {
...
             details = "BackupSnapshot Failed due to " + e.getMessage();
             s_logger.warn(details, e);
+        } finally {
+            if (!successful) {
```
rafaelweingartner: What about removing this finally and boolean control variable? You only need to worry about exception cases (normal workflow is already addressed at 567-568). So, you can do something like this in the catch block: 
```
destroySnapshotOnPrimaryStorage(conn, snapshotUuid);
```

BTW: the method `destroySnapshotOnPrimaryStorage(Conn, String)` does seem to need that boolean return and we also could remove its catch of `Exception`.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
khos2ow: Actually not, there are multiple other exceptions can be thrown which this block should catch them. The difference is that `Types.XenAPIException` doesn't have a meaningful `getMessage()` method. I'll combine them together though.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
khos2ow: When the two `catch` block merged together I'll move this from `finally` to `catch`.

khos2ow:
@rafaelweingartner updated the PR.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -559,12 +559,12 @@ public Answer backupSnapshot(final CopyCommand cmd) {
                     physicalSize = Long.parseLong(tmp[1]);
                     finalPath = folder + File.separator + snapshotBackupUuid + ".vhd";
                 }
-
-                final String volumeUuid = snapshotTO.getVolume().getPath();
```
rafaelweingartner: This is where I was getting confused. I am inspecting the code in master, and we were only removing other snapshots, but the last one when it was not a full clone. You changed so we remove all but the last one always.

sateesh-chodapuneedi:
cc @anshul-gangwar @harikrishna-patnala for review

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
khos2ow: Yes, some secondary storage (in our case Swift) cannot take partial snapshots, so we noticed the issue.

rhtyd:

mike-tutkowski:
@rhtyd @rafaelweingartner @DaanHoogland I believe we should consider this PR for the first maintenance release of 4.11.

khos2ow:
@mike-tutkowski Rohit already added this to 4.11.1 milestone, unless you meant something else.

mike-tutkowski:
No, I wasn’t aware that he had done so. That’s great. :-)

On Jan 14, 2018, at 9:56 PM, Khosrow Moossavi <notifications@github.com<mailto:notifications@github.com>> wrote:


@mike-tutkowski<https://github.com/mike-tutkowski> Rohit already added this to 4.11.1 milestone, unless you meant something else.

khos2ow:
@rhtyd @rafaelweingartner @wido @DaanHoogland Since 4.11 is out can we get these PR going again?

rafaelweingartner:
All tests passed. I already reviewed and approved the proposed changes. I believe we need someone else to review and approve this as well.

DaanHoogland:
hm, i gues it already had. merging

PR #2400, opened 2018-01-10
CLOUDSTACK-8900 listLdapUsers with listType=new only filters users for logged-in user account

SudharmaJain:
Added an ldap user 'ldapuser1' to admin account. Now on clicking 'Add LDAP Account' button, UI displays the ldap users other than 'ldapuser1'.  

image

image

Added a new admin account as 'admin2'. Login as admin2 and clicking on  'Add LDAP Account' displays all the ldap users including 'ldapuser1'. 

## Expected behavior

List Ldapusers with listtype=new should filter already added users irrespective of logged in user.

## Before the Fix

image
## After the Fix
image


DaanHoogland:
@SudharmaJain if 'admin2' is not allowed to see the ACS account of 'ldapuser1', than 'ldapuser1' should not be filtered out, should it? it should just fail to add it if that was attempted in the same domain. I'm a bit worried about the list all functionality in terms of security.

SudharmaJain:
@DaanHoogland Both admin and admin2 are root admins. So I don't see any restrictions on these accounts. Also when adding of user fails that will also reveal as the user is already added.

niteshsarda:
Tested this fix and it seems to be working as per expectation.
**Test LGTM.**

DaanHoogland:
@SudharmaJain

The point remains, if they were not root admins they would not be allowed to see the user(, maybe?). Other then that, how about when trying to add a user with the same name to another domain. It should be possible to add the same user twice as long as they are not in the same domain.

rafaelweingartner:
Ping @SudharmaJain

GabrielBrascher:
Ping @SudharmaJain.

DaanHoogland:
@SudharmaJain please rebase and re-open if still relevant

PR #2404, opened 2018-01-12
[CLOUDSTACK-10230] User should not be able to use removed “Guest OS type”

rafaelweingartner:
Users are able to change the OS type of VMs to a “Guest OS type” that has been removed. This becomes a security issue when we try to force users to use HVM VMs (Meltdown/Spectre thing). A removed “guest os type” should not be usable by any users in the cloud.

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -306,8 +306,14 @@ import com.cloud.vm.snapshot.dao.VMSnapshotDao;
 public class UserVmManagerImpl extends ManagerBase implements UserVmManager, VirtualMachineGuru, UserVmService, Configurable {
     private static final Logger s_logger = Logger.getLogger(UserVmManagerImpl.class);
 
+    /**
+     * The number of seconds to wait before timing out when trying to acquire a global lock.
+     */
     private static final int ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_COOPERATION = 3;
-    private static final long GB_TO_BYTES = 1024 * 1024 * 1024;
+    /**
+     * The number of bytes in a GiB.
+     */
+    private static final long GiB_TO_BYTES = 1024 * 1024 * 1024;
 
     @Inject
     private EntityManager _entityMgr;
@@ -2256,7 +2262,6 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         @Override
         protected void runInContext() {
             GlobalLock scanLock = GlobalLock.getInternLock("vmIpFetch");
-
             try {
                 if (scanLock.lock(ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_COOPERATION)) {
                     try {
@@ -3292,7 +3297,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
             if (rootDiskSize <= 0) {
                 throw new InvalidParameterValueException("Root disk size should be a positive number.");
             }
-            size = rootDiskSize * GB_TO_BYTES;
+            size = rootDiskSize * GiB_TO_BYTES;
         } else {
             // For baremetal, size can be null
             Long templateSize = _templateDao.findById(template.getId()).getSize();
@@ -3312,7 +3317,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                     throw new InvalidParameterValueException("VM Creation failed. Volume size: " + diskSize + "GB is out of allowed range. Max: " + customDiskOfferingMaxSize
                             + " Min:" + customDiskOfferingMinSize);
                 }
-                size += diskSize * GB_TO_BYTES;
+                size += diskSize * GiB_TO_BYTES;
             }
             size += _diskOfferingDao.findById(diskOfferingId).getDiskSize();
         }
@@ -3786,7 +3791,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
     {
         // rootdisksize must be larger than template.
         if ((rootDiskSize << 30) < templateVO.getSize()) {
-            Long templateVOSizeGB = templateVO.getSize() / GB_TO_BYTES;
+            Long templateVOSizeGB = templateVO.getSize() / GiB_TO_BYTES;
             String error = "Unsupported: rootdisksize override is smaller than template size " + templateVO.getSize() + "B (" + templateVOSizeGB + "GB)";
             s_logger.error(error);
             throw new InvalidParameterValueException(error);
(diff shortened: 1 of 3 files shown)


rafaelweingartner:
@blueorangutan package

rafaelweingartner:
@blueorangutan package

rafaelweingartner:
@borisstoyanov can you run tests here?

DaanHoogland:
yes, he can ;)
@blueorangutan test

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -305,254 +295,220 @@
...
 import com.cloud.vm.snapshot.dao.VMSnapshotDao;
-import com.cloud.storage.snapshot.SnapshotApiService;
-import com.cloud.storage.VMTemplateStorageResourceAssoc;
+
 
 public class UserVmManagerImpl extends ManagerBase implements UserVmManager, VirtualMachineGuru, UserVmService, Configurable {
     private static final Logger s_logger = Logger.getLogger(UserVmManagerImpl.class);
 
-    private static final int ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_COOPERATION = 3; // 3 seconds
+    private static final int ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_COOPERATION = 3;
```
DaanHoogland: why not leave a // in seconds
usefull comment for a change ;)

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -3703,9 +3654,9 @@ private String generateHostName(String uuidName) {
     }
 
     private UserVmVO commitUserVm(final DataCenter zone, final VirtualMachineTemplate template, final String hostName, final String displayName, final Account owner,
-                                  final Long diskOfferingId, final Long diskSize, final String userData, final Account caller, final Boolean isDisplayVm, final String keyboard,
-                                  final long accountId, final long userId, final ServiceOfferingVO offering, final boolean isIso, final String sshPublicKey, final LinkedHashMap<String, NicProfile> networkNicMap,
-                                  final long id, final String instanceName, final String uuidName, final HypervisorType hypervisorType, final Map<String, String> customParameters, final Map<String, Map<Integer, String>> extraDhcpOptionMap, final Map<Long, DiskOffering> dataDiskTemplateToDiskOfferingMap) throws InsufficientCapacityException {
+            final Long diskOfferingId, final Long diskSize, final String userData, final Account caller, final Boolean isDisplayVm, final String keyboard,
```
DaanHoogland: hm, i actually liked the alignment of the parameters. good old days ;)

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -6458,4 +6423,4 @@ private boolean checkStatusOfVolumeSnapshots(long vmId, Volume.Type type) {
         }
         return false;
     }
-}
+}
```
DaanHoogland: it is hard to find functional changes in this file. the new 2380-2424 are it, are they?

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
rafaelweingartner: Agreed.  I probably removed without thinking. 
I re-introduced this documentation.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
rafaelweingartner: methods with hundreds of parameters... the glory of the old days...

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
rafaelweingartner: Yeas, it is.

That is why I am not doing like this anymore. I am always splitting up the PR into multiple commits. It was one of your suggestions some time ago, which I liked. However, I created this PR before that.

The changes are basically at lines 2380-2424.

rafaelweingartner:
@DaanHoogland thanks for your time reviewing this PR. I re-introduced the documentation of `ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_COOPERATION` variable as you suggested.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -306,8 +306,14 @@
...
 
+    /**
+     * The number of seconds to wait before timing out when trying to acquire a global lock.
+     */
     private static final int ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_COOPERATION = 3;
-    private static final long GB_TO_BYTES = 1024 * 1024 * 1024;
+    /**
+     * The number of bytes in a GiB.
+     */
+    private static final long GiB_TO_BYTES = 1024 * 1024 * 1024;
```
DaanHoogland: this one is actually self documenting!

rafaelweingartner: Sure, removed.

DaanHoogland: i enjoyed the renaming of it very much though ;)

DaanHoogland: come to think about it, and don't take this as criticism on your naming to much, I think constants like this one should be appended with "_FACTOR" to be completely self documenting.

rafaelweingartner: no problem. I just renamed to `GiB`, because the previous `GB` prefix that was being used is not accurate. 

Sure the wording "factor" might be more appropriate, but I think we should look for a different solution. We have these constant all over our code base. We should look for a "MetricUtils/UnitUtils" or something like that to convert between different units (bytes to KB, KiB, MB, MiB, GB, GiB, TB, TiB, hertz and Ghz and so on)

DaanHoogland: yes, my remark was about the self documenting quality of the constant name; it is not GiB_TO_BYTES, as it cannot push all information contained in a GiB into a small number of bytes as anyone as ignorant as myself might think. The contant is and can only be used as a factor to convert from GiB to a number of B so it should be called GiB_TO_B_FACTOR or some such name. having said that, another implementator would choose to have a MetricUtil.convertGiBtoB() method.
/me is not a bigot on that difference.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

rafaelweingartner:
Something must have happened with the CI system. This PR's changes would not cause these failures.

borisstoyanov:
@blueorangutan test

PaulAngus:
@blueorangutan test

rhtyd:

rafaelweingartner:
This PR might be interesting. However, due to the amount of changes (I refactored the whole method), I preferred to introduced it only on master.

PR #2405, opened 2018-01-13
CLOUDSTACK-10146 checksum in java instead of script

DaanHoogland:


agent/src/com/cloud/agent/direct/download/DirectTemplateDownloaderImpl.java
```diff
@@ -149,36 +154,16 @@ public DirectTemplateInformation getTemplateInformation() {
...
-            ChecksumValue providedChecksum = new ChecksumValue(checksum);
-            String algorithm = providedChecksum.getAlgorithm();
-            String checksumCommand = "echo '%s %s' | %s -c --quiet";
-            String cmd = String.format(checksumCommand, providedChecksum.getChecksum(), downloadedFilePath, getChecksumCommandFromAlgorithm(algorithm));
-            int result = Script.runSimpleBashScriptForExitValue(cmd);
-            return result == 0;
+            try {
+                return DigestHelper.check(checksum, new FileInputStream(downloadedFilePath));
+            } catch (IOException e) {
+                throw new CloudRuntimeException("could not check sum for file: " + downloadedFilePath);
```
marcaurele: missing second argument to view the actual exception message/stack trace:
`file: " + downloadedFilePath, e)`

DaanHoogland: yes, I agree. though people have been pushing towards not including root causes in exception traces i will add it!

DaanHoogland: come to think of it it does not make as much sense in the NoSuchAlgorithmException case. Doe you agree @marcaurele ?

marcaurele: IMO if the exception is due to some system/programming problem, the stack should be given to give a clear hint to people where to look for the error in the code. If the exception is aimed at the end user (API end user), then it does not make sens as it is a purely parameter, state,... error. So i'll keep the root cause for the NoSuchAlgorithmException.

DaanHoogland: ok, I thought the error is purely the chosen algorithnm so don't include the root cause.

borisstoyanov:
@blueoranbutan package

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

PR #2376, opened 2017-12-24
[4.11] Smoketest Health Check

yadvr:
This is a dummy PR to run smoke/component tests on master branch towards the 4.11 release.
@blueorangutan package

PR #2407, opened 2018-01-15
CLOUDSTACK-10227: Add delay before reverting VM snapshot

rhtyd:
As discovered and discussed in #2376, adding some delay after stopping
the VM and reverting VM snapshot passes the
`test_change_service_offering_for_vm_with_snapshots` test case. The
suspect here is userVMDao or background vmsync that may not update
the VM state to PowerOff.

Pinging for review @mike-tutkowski @rafaelweingartner @DaanHoogland and others.
Manually verified with QA:
```
=== TestName: test_change_service_offering_for_vm_with_snapshots | Status : SUCCESS ===
ok

Ran 1 test in 501.359s
OK
```

I think since this only changes the test file, it may be accepted only with at least code reviews. It's not a blocker, but fixing this smoketest will avoid getting incorrect failures.


rhtyd:
Thanks @DaanHoogland @borisstoyanov . I'll merge this now and start RC1.

PR #2413, opened 2018-01-19
CLOUDSTACK-10117 scoped setting have nullable values

DaanHoogland:


rhtyd:
@blueorangutan package

rhtyd:
Please review/lgtm -- @nvazquez @borisstoyanov and others
@blueorangutan test

rhtyd:
This PR has changes in db upgrade path, to avoid adding a new db upgrade path b/w 4.11.0->4.11.1, and 4.11.1 or 4.11.0->4.12.0 (-snapshot) I think we should consider this merge towards 4.11.0 milestone and to make it into RC2. /cc @wido @rafaelweingartner @nvazquez @mike-tutkowski and others - please review

rafaelweingartner:
I did not understand why this change will avoid us having an upgrade path between 4.11 - 4.12.
@DaanHoogland can you provide some extra details here?

rhtyd:
@rafaelweingartner this will add complexity that's all, ideally, we should n't (or at least try not to) have db changes in minor/dot releases.

rafaelweingartner:
Ah, now I got it.
You say, add this now, so we do not need to create upgrade paths for the minor version 4.11.1.0
Is that it?

rhtyd:
Yes @rafaelweingartner that's it.

rafaelweingartner:
It seems to be something simple. However, I have no clues why you are changing a field in the database from not null to null able. I checked the Jira ticket, and it is a generic one regarding a feature that you were/are developing.

It is always good to provide these details either in the Jira ticket you create for the PR, or in the PR description itself. I also checked the commit, but it does not provide information regarding the necessity for this change and the impacts it may cause.

DaanHoogland:
@rafaelweingartner it is a generic quality of settings that they are null able. I booboo'd here by making them non null. The code is really just part of the general functionality.

@rhtyd I will remove 4.11.1 milestone as it would have to be a alter table statement if it needs to go in there.

mike-tutkowski:
Based on the code and the comments in this PR, LGTM.

rhtyd:
Tests LGTM, Merging this based on code reviews and test result.

PR #2414, opened 2018-01-19
[CLOUDSTACK-10241] Duplicated file SRs being created in XenServer pools

rafaelweingartner:
Due to a race condition between multiple management servers, in some rare cases, CloudStack is creating multiple file SRs to the same secondary folder. This causes a problem when introducing the SR to the XenServer pools, as “there will be VDIs with duplicated UUIDs“. The VDIs are the same, but they are seen in different SRs, and therefore cause an error.

The solution to avoid race conditions between management servers is to use a deterministic srUuid for the file SR to be created (we are leaving XenServer with the burden of managing race conditions). The UUID is based on the SR file path and is generated using  `UUID.nameUUIDFromBytes`. Therefore, if there is an SR with the generated UUID, this means that some other management server has just created it. An exception will occur and it will contain a message saying 'Db_exn.Uniqueness_constraint_violation'. In these unlikely events, we catch the exception and use the method `retrieveAlreadyConfiguredSrWithoutException` to get the SR that has already been created for the given mount point.

### DIFF
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
@@ -198,7 +198,7 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
 
     private final static int BASE_TO_CONVERT_BYTES_INTO_KILOBYTES = 1024;
 
-    protected static final XenServerConnectionPool ConnPool = XenServerConnectionPool.getInstance();
+    private static final XenServerConnectionPool ConnPool = XenServerConnectionPool.getInstance();
     // static min values for guests on xenserver
     private static final long mem_128m = 134217728L;
 
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
@@ -243,7 +243,7 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
     protected String _guestNetworkName;
     protected int _heartbeatInterval = 60;
     protected int _heartbeatTimeout = 120;
-    protected final XsHost _host = new XsHost();
+    protected XsHost _host = new XsHost();
     protected String _instance; // instance name (default is usually "VM")
     protected boolean _isOvs = false;
     protected String _linkLocalPrivateNetworkName;
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
@@ -37,6 +37,7 @@ import com.xensource.xenapi.SR;
 import com.xensource.xenapi.Task;
 import com.xensource.xenapi.Types;
 import com.xensource.xenapi.Types.BadServerResponse;
+import com.xensource.xenapi.Types.StorageOperations;
 import com.xensource.xenapi.Types.XenAPIException;
 import com.xensource.xenapi.VDI;
 import org.apache.cloudstack.storage.command.CopyCmdAnswer;
@@ -45,6 +46,8 @@ import org.apache.cloudstack.storage.to.PrimaryDataStoreTO;
 import org.apache.cloudstack.storage.to.SnapshotObjectTO;
 import org.apache.cloudstack.storage.to.TemplateObjectTO;
 import org.apache.cloudstack.storage.to.VolumeObjectTO;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.apache.xmlrpc.XmlRpcException;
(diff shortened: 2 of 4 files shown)


plugins/hypervisors/xenserver/src/test/java/com/cloud/hypervisor/xenserver/resource/XenServer625ResourceTest.java
```diff
@@ -30,7 +31,7 @@
 
     @Before
     public void beforeTest() {
-        super.citrixResourceBase = new Xenserver625Resource();
+        super.citrixResourceBase = Mockito.spy(new Xenserver625Resource());
```
DaanHoogland: i don't see this spy being used, why create it?

plugins/hypervisors/xenserver/src/test/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessorTest.java
```diff
@@ -0,0 +1,459 @@
...
+    @Test
+    public void makeDirectoryTestCallHostPlugingReturningNull() {
+        boolean makeDirectoryReturn = configureAndExecuteMakeDirectoryMethodForTest(pathMock, null);
+
+        assertFalse(makeDirectoryReturn);
+    }
+
+    @Test
+    public void makeDirectoryTestCallHostPlugingReturningSomething() {
+        boolean makeDirectoryReturn = configureAndExecuteMakeDirectoryMethodForTest(pathMock, "Soemthing");
```
DaanHoogland: Soemthing on purpose or Something intended?

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
```diff
@@ -17,7 +17,52 @@
 package com.cloud.hypervisor.xenserver.resource;
```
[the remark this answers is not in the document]
rafaelweingartner: Sorry for that. It is a small change at lien 244. I removed a final keyword so I would be able to Spy an object to create unit test cases. I removed the formatting. It is my Eclipse, even though I configured it to format only edited lines, it keeps applying the standard on everything.

plugins/hypervisors/xenserver/src/test/java/com/cloud/hypervisor/xenserver/resource/XenServer625ResourceTest.java
rafaelweingartner: Oh, this one... It is because there is a hierarchy of test classes.. :(.. the property `citrixResourceBase` needs to be either a mock or spy object because of the testes executed in the super class. Therefore, needed to create a spy object whenever attributing an object to this property.

plugins/hypervisors/xenserver/src/test/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessorTest.java
rafaelweingartner: Ah this one. It can actually be any string. instead of writing a fictitious directory, which though it would be misleading. Therefore, I decided to write anything... I will try to change this...

rafaelweingartner:
@DaanHoogland no problem! Thanks for the feedback so far. 
Have you tried opening on Chrome? For me, it seems to work just fine.

rafaelweingartner:
@DaanHoogland I managed to separate the formatting from the code change itself. Now it should be easier to review.

@khos2ow, @marcaurele and others, this might be of your interest as well (if you run XenServer).

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -65,90 +69,181 @@ public Xenserver625StorageProcessor(final CitrixResourceBase resource) {
...
+    private void mountNfs(Connection conn, String remoteDir, String localDir) {
         if (localDir == null) {
             localDir = "/var/cloud_mount/" + UUID.nameUUIDFromBytes(remoteDir.getBytes());
         }
-
-        final String results = hypervisorResource.callHostPluginAsync(conn, "cloud-plugin-storage", "mountNfsSecondaryStorage", 100 * 1000, "localDir", localDir, "remoteDir",
-                remoteDir);
-
-        if (results == null || results.isEmpty()) {
+        String result = hypervisorResource.callHostPluginAsync(conn, "cloud-plugin-storage", "mountNfsSecondaryStorage", 100 * 1000, "localDir", localDir, "remoteDir", remoteDir);
```
DaanHoogland: final not allowed?

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -65,90 +69,181 @@ public Xenserver625StorageProcessor(final CitrixResourceBase resource) {
...
-
             throw new CloudRuntimeException(errMsg);
         }
-
-        return true;
     }
 
-    protected boolean makeDirectory(final Connection conn, final String path) {
-        final String result = hypervisorResource.callHostPlugin(conn, "cloud-plugin-storage", "makeDirectory", "path", path);
+    protected boolean makeDirectory(Connection conn, String path) {
```
DaanHoogland: why removing the final?

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
rafaelweingartner: Of course, they are. However, here and in a lot of places in ACS, there is an indiscriminate use of the final keyword. It does not harm, but it is something I have to read when reviewing/coding, which does not aggregate/help. So, I remove when they are not needed.

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
rafaelweingartner: Because it is useless in this context.

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
DaanHoogland: ok, short method with low cyclic complexity; makes sense

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
DaanHoogland: :+1:

DaanHoogland:
@blueorangutan test

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/Xenserver625StorageProcessor.java
```diff
@@ -65,90 +69,181 @@ public Xenserver625StorageProcessor(final CitrixResourceBase resource) {
...
+        try {
+            return retrieveAlreadyConfiguredSr(conn, srPath);
+        } catch (XenAPIException | XmlRpcException e) {
+            throw new CloudRuntimeException("Unexpected exception while trying to retrieve an already configured file SR for path: " + srPath);
+        }
+    }
 
-            final String msg = "createFileSR failed! due to the following: " + ex.toString();
+    /**
+     *  This method will check if there is an already configured file SR for the given path. If by any chance we find more an one SR with the same name (mount point path) we throw a runtime exception because this situation should never happen.
```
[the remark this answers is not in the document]
rafaelweingartner: Thanks @GabrielBrascher. Typo fixed.

rhtyd:
Thanks @rafaelweingartner since this is a bug fix, do you think it may be useful for 4.11 LTS?

rafaelweingartner:
@rhtyd I would rather not. It is not a simple rebase to 4.11.
I just tried and a lot of conflicts appeared, which would require some extra work to solve.

rhtyd:
okay @rafaelweingartner, I'll review this soon.

rhtyd:

rafaelweingartner:
@rhtyd do you want me to wait a little longer on this one?

rafaelweingartner:
@rhtyd I will wait more 24 hours before merging this one to give you some extra time in cause you have objections. Tests results are green (tests executed by @DaanHoogland 14 days ago), and the reviews are positive.

rhtyd:

rafaelweingartner:
I would leave it only for 4.12 due to its size. Only we seemed to have this issue so far.

PR #2415, opened 2018-01-19
CLOUDSTACK-10244: KVM online storage migration fails and corrupts disk

mike-tutkowski:
This is an issue that was introduced in the code-review process of https://github.com/apache/cloudstack/pull/2298.

What is probably the simplest solution (and what this PR does) is to store the result of MapUtils.isNotEmpty(mapMigrateStorage) before passing mapMigrateStorage to the replaceStorage method.

The replaceStorage method can mutate the data in the passed-in map, thus leading to a different result for MapUtils.isNotEmpty(mapMigrateStorage) later on (and causing an issue for online storage migration).

If we just store the value up front and refer to it later, we, of course, get the same (boolean) result and online storage migration works properly.

A more long-term solution (and one that I plan on implementing for 4.12) would be to make a shallow copy of the map inside of replaceStorage so that replaceStorage does not actually mutate the data in the original map. While that isn't a difficult solution to implement, I feel the one I have implemented in this PR is safer as we are in the release-candidate process already.

Here is a link to the JIRA ticket:

https://issues.apache.org/jira/browse/CLOUDSTACK-10244

I have tested the code and validated that it works.

This use case only applies to online storage migration from NFS or Ceph to managed storage, which is a new feature for 4.11.

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
@@ -132,8 +132,9 @@ public final class LibvirtMigrateCommandWrapper extends CommandWrapper<MigrateCo
             vmsnapshots = libvirtComputingResource.cleanVMSnapshotMetadata(dm);
 
             Map<String, MigrateCommand.MigrateDiskInfo> mapMigrateStorage = command.getMigrateStorage();
+            boolean migrateStorage = MapUtils.isNotEmpty(mapMigrateStorage);
 
-            if (MapUtils.isNotEmpty(mapMigrateStorage)) {
+            if (migrateStorage) {
                 xmlDesc = replaceStorage(xmlDesc, mapMigrateStorage);
             }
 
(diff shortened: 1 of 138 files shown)


rhtyd:
@mike-tutkowski can you change the base of the PR to 4.11 branch?

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
```diff
@@ -132,8 +132,9 @@ Use VIR_DOMAIN_XML_SECURE (value = 1) prior to v1.0.0.
             vmsnapshots = libvirtComputingResource.cleanVMSnapshotMetadata(dm);
 
             Map<String, MigrateCommand.MigrateDiskInfo> mapMigrateStorage = command.getMigrateStorage();
+            boolean migrateStorage = MapUtils.isNotEmpty(mapMigrateStorage);
```
wido: Would it not be safer to make this boolean a final one? That way you are 100% sure it's not changed anywhere.

mike-tutkowski: @wido Sounds good. I can do that and then put the code on the branch that @rhtyd asked.

mike-tutkowski:
@rhtyd Will I need to open a new PR to change the base of the PR? Thanks!

rhtyd:
@mike-tutkowski can you try the edit button on this pr?

mike-tutkowski:
@rhtyd @wido It looks like it took another commit along for the ride when I put this PR on top of a new base. I'll just go ahead and open a new PR and comment on its number here.

mike-tutkowski:
Closing this PR in favor of https://github.com/apache/cloudstack/pull/2416

PR #2028, opened 2017-04-05
CLOUDSTACK-9853: Add support for Secondary IPv6 Addresses and Subnets

wido:
This commit adds support for passing IPv6 Addresses and/or Subnets as
Secondary IPs.

This is groundwork for CLOUDSTACK-9853 where IPv6 Subnets have to be
allowed in the Security Groups of Instances to we can add DHCPv6
Prefix Delegation.

Use ; instead of : for separating addresses, otherwise it would cause
problems with IPv6 Addresses.

Signed-off-by: Wido den Hollander <wido@widodh.nl>

PR #2418, opened 2018-01-22
CLOUDSTACK-10242: Properly parse incoming rules to Sec Group

wido:
With merge of PR #2028 the separator for lines to the Security Group
Python script changed from : to ; to support IPv6 addresses.

This broke certain situations where rules were parsed improperly.

This commit fixes that

### DIFF
--- a/scripts/vm/network/security_group.py
+++ b/scripts/vm/network/security_group.py
@@ -951,16 +951,15 @@ def parse_network_rules(rules):
   if rules is None or len(rules) == 0:
     return ret
 
-  lines = rules.split(';')[:-1]
+  lines = rules.split('NEXT;')[:-1]
   for line in lines:
-    tokens = line.split(':', 4)
-    if len(tokens) != 5:
+    tokens = line.split(';', 3)
+    if len(tokens) != 4:
       continue
 
-    ruletype = tokens[0]
-    protocol = tokens[1]
-    start = int(tokens[2])
-    end = int(tokens[3])
+    ruletype, protocol = tokens[0].split(':')
+    start = int(tokens[1])
+    end = int(tokens[2])
     cidrs = tokens.pop();
 
     ipv4 = []


scripts/vm/network/security_group.py
```diff
@@ -951,16 +951,15 @@ def parse_network_rules(rules):
   if rules is None or len(rules) == 0:
     return ret
 
-  lines = rules.split(';')[:-1]
+  lines = rules.split('NEXT;')[:-1]
```
rhtyd: @wido what is the `NEXT` here? Can you give a sample of the rule/commands processed by this code?

rhtyd:
I've an outstanding question, other LGTM pending test results.
@blueorangutan package

scripts/vm/network/security_group.py
wido: <pre>I:icmp;-1;-1;0.0.0.0/0,NEXT;I:tcp;0;65535;0.0.0.0/0,NEXT;I:udp;0;65535;0.0.0.0/0,NEXT;</pre>

Sure, this is how it looks like. The NEXT word is the last word before a new rule/line starts

DaanHoogland: I think this extremely ugly solution is acceptable but i see a possible problem with the preceeding , it spells ',NEXT;' could it be ',NEXT,', instead? and if so, why not just ','? (read at your own risk)

rhtyd: I think having the `,` is fine, @DaanHoogland see line 967/968, the cidrs could be comma separated, so splitting on `NEXT;` looks okay. The `NEXT` separator is added in `SecurityGroupRulesCmd.java`

wido:
I agree with you @DaanHoogland that's it not very elegant. But the way we pass rules to the Python file isn't my favorite either.

The 'NEXT' keyword was the only thing I could split on elegantly without running into troubles.

rhtyd:
Test LGTM, merging based on code reviews and test results.

rhtyd:
@wido please send bugfixes to 4.11 branch, this issue is marked blocker and should be send for 4.11 branch. I've rebased your branch, but pay attention to base-branch when sending PR in future.

rhtyd:
@blueorangutan test

rhtyd:
LGTM, merging based on test results and reviews. The failing tests were caused by a failure in test_public_ip_range.py that introduces a fake public ip range used by systemvms, causing failure on download urls of template/iso/volumes.

PR #2419, opened 2018-01-22
CLOUDSTACK-10245: Fix password server

fmaximus:
In case of isolated networks where VR is not the gateway, both self.config.is_vpc() and self.config.is_router() are false,
but self.config.is_dhcp() is true.
Moved the password server logic to the `if has_metadata` block,
as this is valid for all 3 systemvm types.


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Merging this based on test results and code reviews.

PR #2421, opened 2018-01-22
CLOUDSTACK-10243: Do not use wait() on Python subprocess

wido:
This might (and does block) in certain situations on the VR as
also explained in the Python documentation:

https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait

  Warning This will deadlock when using stdout=PIPE and/or stderr=PIPE
  and the child process generates enough output to a pipe such that
  it blocks waiting for the OS pipe buffer to accept more data.
  Use communicate() to avoid that.

Using the check_output function handles most of this for us and
also provides better error handling.

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsHelper.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsHelper.py
@@ -183,13 +183,19 @@ def get_hostname():
 
 def execute(command):
     """ Execute command """
-    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-    p.wait()
-    rc = p.returncode
+    returncode = -1
+    try:
+        logging.info("Executing: %s" % command)
+        result = subprocess.check_output(command, shell=True)
+        returncode = 0
+        return result.splitlines()
+    except subprocess.CalledProcessError as e:
+        logging.error(e)
+        returncode = e.returncode
+    finally:
+        logging.debug("Executed: %s - exitstatus=%s " % (command, returncode))
 
-    logging.debug("Executed: %s - exitstatus=%s " % (command, rc))
-    result = p.communicate()[0]
-    return result.splitlines()
+    return list()
 
 
 def save_iptables(command, iptables_file):


wido:
@rhtyd My colleagues told me that they didn’t see this happen with 4.9 and now see it in 4.10.

This file has not changed between 4.10 and 4.11

borisstoyanov:
Thanks for this fix @wido 
@blueorangutan package

rhtyd:
@wido how do we reproduce a failure or timeout/delay you may have observed, and does this apply for 4.11, if so please change the base branch and let's target this for 4.11.1.0

wido:
I changed the base against 4.11 @rhtyd 

If that passes I'll draft a new one for master so that we also fix it there.

Using check_output() is much better then wait() according to the Python docs.

I don't know 100% sure on how to reproduce it, but we see it happen on VRs which spawn VMs daily and they all block on:

<pre>ip rule show</pre>

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

rhtyd:
Thanks @wido I tried to reproduce it, but could n't reproduce the cpu consumption issue. Between 4.10 and 4.11, the systemvmtemplate has changed (along with python and ip/net tools), so can you team help run some tests?

wido:
Ok @rhtyd ! The problem is that this doesn't happen on small scale, you need to spawn multiple VMs to run into the issue.

We currently 

The point is that we use p.wait() in the Python code there the Python code already say that should not be used.

What this PR does is change the inner workings for the execute() method so that it uses check_output() which is recommended.

rhtyd:
Thanks for the explanation @wido given this is not a blocker, and we cannot accept this in 4.11.0.0-rc2 based on speculative observation, so I've moved it to 4.11.1 milestone. Let's get this tested and merged post 4.11.0.0 release, if you disagree let's discuss on dev@.

wido:
Fine with me @rhtyd !

rhtyd:
LGTM, merging this based on test results and reviews.

PR #2427, opened 2018-01-23
CLOUDSTACK-10252: Delete dnsmasq leases file on restart

rhtyd:
Delete dnsmasq's leases file when dnsmasq is restarted to avoid it
use old ip-mac-address-vm mapping leases.

Pinging for review - @DaanHoogland @borisstoyanov @nvazquez @wido @rafaelweingartner and others.

@blueorangutan package

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
@@ -49,14 +49,20 @@ class CsDhcp(CsDataBag):
             self.add(self.dbag[item])
         self.write_hosts()
 
-        if self.cloud.is_changed():
-            self.delete_leases()
-
         self.configure_server()
 
-        restart_dnsmasq = self.conf.commit()
-        self.cloud.commit()
+        restart_dnsmasq = False
+
+        if self.conf.commit():
+            restart_dnsmasq = True
+
+        if self.cloud.commit():
+            restart_dnsmasq = True
+
         self.dhcp_opts.commit()
+
+        if restart_dnsmasq:
+            self.delete_leases()
 
         if not self.cl.is_redundant() or self.cl.is_master():
             if restart_dnsmasq:


rhtyd:
@blueorangutan test

systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
```diff
@@ -49,15 +49,21 @@ def process(self):
...
         self.configure_server()
 
-        restart_dnsmasq = self.conf.commit()
-        self.cloud.commit()
+        restart_dnsmasq = False
+
+        if self.conf.commit():
+            restart_dnsmasq = True
+
+        if self.cloud.commit():
```
resmo: repeat 2 times commit, why?

resmo:
I am skeptical, at the first view, are the "old" entries rewritten?

systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
rhtyd: @resmo The commit is called on two different files, it's the same old code refactored differently.

rhtyd:
@resmo the values from json/dictionary are written on commit.

systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
DaanHoogland: I see conf.commit and cloud.commit. Those are two different objects, @resmo . How is this suspect?

rhtyd:
@DaanHoogland that's right, the aim of the PR was scoped only to fix the blocker. Perhaps for 4.11.1+ we can aim to improve changing the leases file instead of removing it, how it used to be done before it was removed:
https://github.com/apache/cloudstack/commit/73c0242df355a29ffdddd3134f6288a63dd3f1df#diff-3ab4e2009108bf42111df9efc71f0f03R98

systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
resmo: I missed that. never mind.

DaanHoogland:
Yes, well that bit of code does dnsmasq's job. I think we should merge this and then find a way to call a release on a single lease by mac or ip.

LGTM ^^

resmo:
sounds like a plan. 
LGTM

rhtyd:
Thanks, merging this based on code reviews and test results.

PR #2433, opened 2018-01-26
CLOUDSTACK-10268: Fix and enhance package script

khos2ow:
- new flag `-T, --use-timestamp` to use `timestamp` when POM version contains SNAPSHOT
  - in the final artifacts (jar) name
  - in the final package (rpm, deb) name
  - in `/etc/cloudstack-release` file of SystemVMs
  - in the Management Server > About dialog
- if there's a "branding" string in the POM version (e.g. `x.y.z.a-NAME[-SNAPSHOT]`),
the branding name will be used in the final generated pacakge name such as following:
  - `cloudstack-management-x.y.z.a-NAME.NUMBER.el7.centos.x86_64`
  - `cloudstack-management_x.y.z.a-NAME-NUMBER~xenial_all.deb`
- branding string can be overriden with newly added `-b, --brand` flag
- handle the new format version for VR version
- fix long opts (they were broken)
- tolerate and show a warning message for unrecognized flags
- usage help reformat


wido:
As the 4.11 branch is closed for now and this is not a blocker I think this one should be against master and thus be included in 4.12

khos2ow:
@wido I agree, it's not a blocker for 4.11 at all. The sole reason I based it off 4.11 was that we're going to rebase against it internally. I guess it would be finevfor this PR to go in 4.11.1 if not 4.11.0 
/cc @syed @swill @pdion891 what do you suggest guys?

DaanHoogland:
i'd say 4.11.1 is fine for this. It is not production code and with some bad will could even be considered a fix. @rhtyd ?

rafaelweingartner:
Why `4.11` and not master directly? It looks more like an enhancement than a bug fix to go into a released(ish) version.

rhtyd:
@khos2ow please move the base branch to master, I think what you're trying to achieve may be possible without these changes. Let me test/confirm them and get back to you later this week.

khos2ow:
Updated the base to `master`.

DaanHoogland:
@khos2ow i think something went wrong there. github doesn't like :(

khos2ow:
@DaanHoogland anything except Jenkins failure?

DaanHoogland:
@khos2ow "This branch cannot be rebased safely"

DaanHoogland:
@blueorangutan package

packaging/package.sh
```diff
@@ -67,22 +75,32 @@ function packaging() {
             exit 2
         fi
     fi
+
     VERSION=$(cd ../; $MVN org.apache.maven.plugins:maven-help-plugin:2.1.1:evaluate -Dexpression=project.version | grep --color=none '^[0-9]\.')
+    BASEVER=$(echo "$VERSION" | sed 's/-SNAPSHOT//g')
```
rhtyd: -1 we would want to keep the SNAPSHOT in the package names.

packaging/package.sh
```diff
@@ -67,22 +75,32 @@ function packaging() {
...
+    REALVER=$(echo "$BASEVER" | cut -d '-' -f 1)
+    BRAND=$(echo "$BASEVER" | cut -d '-' -f 2)
```
rhtyd: Instead of this, why not pass the branch as a new parameter that is used for the DEFREL option? Or, use sed on the package.sh script to replace the DEFREL option with the branch parameter?

packaging/centos63/cloud.spec
```diff
@@ -24,11 +24,16 @@
 Name:      cloudstack
 Summary:   CloudStack IaaS Platform
 #http://fedoraproject.org/wiki/PackageNamingGuidelines#Pre-Release_packages
+%if "%{?_brand}" != ""
```
rhtyd: The spec changes may be removed.

rhtyd:
FYI - here's how blueorangutan does centos7 packaging, with jenkins: (the BUILD_ID can be `shapeblue0`, or any name you want to provide)

```
export MINOR=${BUILD_ID}
export VERSION=$(grep '^  <version>' pom.xml| cut -d'>' -f2 |cut -d'<' -f1)         
export BASE_VERSION=$(echo $VERSION | awk -F . '{print $1"."$2}')  
[... other build code snipped ...]                                                   
sed -i "s/^.*DEFREL=.*$/  DEFREL='-D_rel ${MINOR}'/g" package.sh                   
bash -x package.sh -p noredist -d centos7     
```

rafaelweingartner:
@rhtyd I believe that is the point of these PR. So, one does not need to hack around the script, but instead the script use the `pom.xml` version and branding. Therefore, you could have your `shapeblue0` custom version in the pom, and then move forward with the customization and easy versioning of the customization.

rafaelweingartner:
Answering @rhtyd regarding the result of the script. The standard is:
cloudstack-<artifact>-<version>-<brand>.<-r parameter>.<el6 || el7>.x86_64.rpm

rhtyd:
@rafaelweingartner please re-read, my first recommendation was indeed to introduce a new parameter to the script that takes in the brand/custom suffix string. The artifact version should be read from pom.xml without removing the SNAPSHOT (or any string), the brand/custom string can include dates etc.

rafaelweingartner:
I read it. I find the idea of providing a parameter interesting, but it is also interesting to load everything out of the pom.

rhtyd:
@rafaelweingartner disagree, the current package.sh script produces the centos7 rpms with el7.centos suffix, you can check:
http://cloudstack.apt-get.eu/centos7/4.10/
http://packages.shapeblue.com/cloudstack/upstream/centos7/4.10/
However, advising the PR author to remove el7.centos and just el7 is another argument.

rafaelweingartner:
It is not a point to agree or disagree. I am not advising anything, I am presenting only the standard name of RPM packages that this PR can create.

rhtyd:
FYI, centos7 package failed (blueorangutan gave false positive):
```
+ bash -x package.sh -p noredist -o rhel7 -d centos7
+ TARGETDISTRO=
+ SIM=
+ PACKAGEVAL=
+ RELEASE=
+ '[' -n -p ']'
+ case "$1" in
+ echo 'Packaging CloudStack...'
Packaging CloudStack...
+ PACKAGEVAL=noredist
+ echo noredist
noredist
+ '[' noredist == oss -o noredist == OSS ']'
+ '[' noredist == noredist -o noredist == NOREDIST ']'
+ PACKAGEVAL=noredist
+ shift 2
+ '[' -n -o ']'
+ case "$1" in
+ echo 'Error: Unrecognized option'
Error: Unrecognized option
+ usage
+ cat
Usage: ./package.sh -d DISTRO [OPTIONS]...
Package CloudStack for specific distribution and provided options.

If there's a "branding" string in the POM version (e.g. x.y.z.a-NAME[-SNAPSHOT]), the branding name will
be used in the final generated pacakge like: cloudstack-management-x.y.z.a-NAME.NUMBER.el7.centos.x86_64
```

For several package/build scripts to continue to work, please don't remove the -o option, and keep changes in the package.sh script restricted to accepting the brand string only.

rafaelweingartner:
Where is this `-o` option that you are talking about. I am not seeing it in the diff here.
Is it related to the removal of line 122 in file `packaging/package.sh`?

rafaelweingartner:
I also built ACS using this script again for centos7.  I used `./package.sh -p oss -d centos7 -r 1`, and everything works. Why do we need a `-o rhel7 ` when we already have `-d centos7`?

packaging/package.sh
khos2ow: on this line I'm trying to get BRANDING string and not removing SNAPSHOT from final package name. That's don in line 94 and 97 (`$(date +%s)` instead of `SNAPSHOT`).
And I believe this will be a good change because it will prevent accidental artifacts overridden on top of each other.

khos2ow:
@rhtyd @rafaelweingartner 
> Please see outstanding comments, what you're trying to achieve may be done in two ways - (a) introduce a new parameter that is used for the DEFREL option (as provided), (b) use sed to replace the value for DEFREL in your build script/process

As Rafael mentioned, this was exactly the point of my PR, to prevent hacking around packaging and have predictable package name based on POM version (most probably if one needs to have a branding to be able to differentiate between custom built and official built).

My first approach was actually to have another flag (`-b, --brand NAME`) but I removed it right away, because I, myself, forgot to update my Jenkins job although I had changed the POM version.

> I find the idea of providing a parameter interesting, but it is also interesting to load everything out of the pom.

This makes sense too. I guess for sake of completeness we can still have `-b, --brand` flag and this will force override the _auto detected_ branding from POM version.

> For several package/build scripts to continue to work, please don't remove the -o option, and keep changes in the package.sh script restricted to accepting the brand string only.

I did find that `-o` flag was removed in #46 and it wasn't doing anything at all for more than three years and because of the flaw in reading `ARGS` it was never caught. (i.e. this section `*) shift ;;` was silently ignoring unknown flag rather than breaking the build).

And btw, this PR doesn't change anything on the suffix of package name, for example `el7.centos.x86_64` part is unchanged in the scope of this PR.

packaging/package.sh
```diff
@@ -155,31 +170,30 @@ while [ $# -gt 0 ] ; do
...
             ;;
+
         -r | --release)
             RELEASE=$2
-            shift
+            shift 2
             ;;
-        -)
+
+        -*|*)
```
rhtyd: @khos2ow this change, fails the packaging job when -o is passed.

khos2ow: Correct. Shouldn't we update packaging job to have _correct_ flags rather than changing package.sh script to be able to tolerate _incorrect_ flags?

rhtyd: The issue is to support the same jenkins jobs and scripts to work with packaging older releases. I think the script should ignore unknown flags and their values but not fail otherwise.

khos2ow:
@rhtyd @rafaelweingartner @DaanHoogland I changed the PR to be more backward-compatible as follow:
- new `-T, --use-timestamp` flag to use `timestamp` instead of SNAPSHOT
- silently ignore unknown flags (e.g. `-o`) and show a warning that in future building will be broken
- new `-b, --brand string` flag. which will override "branding" string found in POM version, or to add branding if POM doesn't have it
-

packaging/package.sh
```diff
@@ -155,31 +181,51 @@ while [ $# -gt 0 ] ; do
...
-packaging "$PACKAGEVAL" "$SIM" "$TARGETDISTRO" "$RELEASE"
+if [ -n "$unrecognized_flags" ]; then
+    echo "Warning: Unrecognized option(s) found \" ${unrecognized_flags}\""
+    echo "         You're advised to fix your build job scripts and prevent using these"
+    echo "         flags, as in the future release(s) they will break packaging script."
+    echo ""
+fi
 
+echo "Packaging CloudStack..."
+packaging "$PACKAGEVAL" "$SIM" "$TARGETDISTRO" "$RELEASE" "$BRANDING"
```
rhtyd: Should you also pass `$SNAPSHOT_TIMESTAMP` as `$6` to `packaging`?

packaging/package.sh
```diff
@@ -67,22 +79,38 @@ function packaging() {
...
+        BRAND=$(echo "$BASEVER" | cut -d '-' -f 2)
+
+        if [ "$REALVER" != "$BRAND" ]; then
+            DEFBRN="-D_brand -$BRAND"
+            BRAND="${BRAND}."
+        else
+            BRAND=""
+        fi
+    fi
+
```
rhtyd: If $5 or branding is passed, then the above code is not needed. `$5` may be used as argument for `-D_rel`, why introduce a new `-D_brand`, that way changes to the spec files may be removed as well.

rhtyd:
Thanks @khos2ow appreciate the effort, I think there are few minor issues left otherwise okay. Let me kick a test packaging job and see how the packages turn out.
@blueorangutan package

packaging/package.sh
khos2ow: not really. `SNAPSHOT_TIMESTAMP` is globally accessible. On the other words we didn't need to pass all the rest input params to `packaging` function too.

packaging/package.sh
khos2ow: - If you specifically mention `--brand NAME` as input flag, it will use this _NAME_ no matter what, otherwise it will try to extract any "branding" name from POM version (hence the else block).
- If pass `$5` directly to `-D_rel` it will wipe the value of `-r, --release integer` that's why it will be appended to `-D_rel` and not replace it.
- We still do need `-D_brand` for the cases in which POM version does have branding string in them. The folder in `/dist/RPM/...` will have a folder name as the full POM version but `_maventag` in spec wouldn't have it, so the packaging will fail.

rafaelweingartner:
@khos2ow for the efforts here. I liked this since the beginning and now it is more flexible... great job +1.
BTW: will it still generate packages following: `cloudstack-<artifact>-<version>-<brand>.<-r parameter>.<el6 || el7>.x86_64.rpm`

khos2ow:
@rafaelweingartner actually it will generate: `cloudstack-<artifact>-<version>-<brand>.<-r parameter || timestamp>.<el6 || el7>.centos.x86_64.rpm`

rafaelweingartner:
Awesome, thanks for the  explanation.

khos2ow:
@rhtyd @rafaelweingartner Can we proceed with this PR if no other change required?

rafaelweingartner:
Well I already approved this PR. For me it is everything ok. It is even nicer that you provided more features and extensions/override points.

rhtyd:
I'm -1 on it, we need a round of testing to confirm changes which I'll kick
@blueorangutan package

khos2ow:
@rhtyd sure, go for it. I'm pretty sure it would be fine.

DaanHoogland:
what is your -1 based on @rhtyd ?

DaanHoogland:
jenkins failed

DaanHoogland:
rekicking jenkins

rhtyd:
@DaanHoogland see comments above on why I initially was -1 on it.

Will investigate bo failure separately, but found this from last run. LGTM based on the built packages:
screenshot from 2018-02-07 09-51-25

rhtyd:
@DaanHoogland @khos2ow in the built packages ^^, it removes the `-SNAPSHOT` from rpm package names only but not in the Debian package names. I think we should keep whatever is in the pom.xml's version, but if that's what needed I would not block the PR.

DaanHoogland:
I am neutral on this as i don't oversee the consequences, i think

rafaelweingartner:
@rhtyd I believe it should be consistent across different OSs. This looks like a "bug". @khos2ow can you take a look into this issue that was reported?

rhtyd:
Yes, I think we should not remove anything from the version as declared in the pom.xml file, i.e. keep the -SNAPSHOT don't sed/remove it @khos2ow.

khos2ow:
@rhtyd can you paste the command you ran that produced the screenshot you attached?
If you don't do any _hacking_ around it and just pass the information and let the script manages it, you should be getting `-SNAPSHOT` in the final package name (without `--use-timestamp`). The sed/remove is to get the real version without the snapshot, and it has been there in the script since long time ago.

Our custom-versioned POM is `4.12.0.0-co-SNAPSHOT` and with:

Flags | Artifacts
------- | -----------
`--distribution centos7` | `cloudstack-management-4.12.0.0-co.SNAPSHOT.el7.centos.x86_64.rpm`
`--distribution centos7 --use-timestamp` | `cloudstack-management-4.12.0.0-co.1516920012.el7.centos.x86_64.rpm`
`--distribution centos7 --release 65` | `cloudstack-management-4.12.0.0-co.SNAPSHOT65.el7.centos.x86_64.rpm`
`--distribution centos7 --use-timestamp  --release 67` | `cloudstack-management-4.12.0.0-co.151802073867.el7.centos.x86_64.rpm`

@rafaelweingartner completely forgot about Debian packages! ;) I'll look into it.

khos2ow:
@rafaelweingartner @DaanHoogland @rhtyd I updated the scripts and description of the PR again.
Now when `--use-timestamp` is specified it will be applied 1) final name of the jar 2) final name of artifacts (rpm, deb) 3) ACS version itself

rafaelweingartner:
Thansk @khos2ow!
Are we waiting for something else here? Everything seems to be in order.

rafaelweingartner:
@rhtyd what is your stance here?

rhtyd:
@blueorangutan package

tools/build/setnextversion.sh
```diff
@@ -16,60 +16,125 @@
...
+Usage: setnextversion.sh --version string [OPTIONS]...
+Set the next version of CloudStack in the POMs.
+
+Mandatory arguments:
+   -v, --version string                    Set the next version to be applied
+
+Optional arguments:
+   -b, --branch string                     Set the branch to update the version into (default "master")
+   -s, --sourcedir string                  Set the source directory to clone repo into (default "$sourcedir")
+   -n, --no-commit                         Apply only the version change and don't git commit them (default "false")
```
rhtyd: @khos2ow @rafaelweingartner please update the wiki as well.

utils/src/main/java/com/cloud/maint/Version.java
```diff
@@ -19,59 +19,86 @@
 
 package com.cloud.maint;
 
+import java.util.regex.Pattern;
+
+import org.apache.commons.lang.StringUtils;
+
 public class Version {
+    public static void main(String[] args) {
```
rhtyd: Remove main() @khos2ow

rhtyd:
@rafaelweingartner I'm okay with the change subject to testing. However, I see several new changes in files not related to packaging.

tools/build/setnextversion.sh
khos2ow: Where in wiki? The only place I could find that references `setnextversion.sh` is https://cwiki.apache.org/confluence/display/CLOUDSTACK/Release+Procedure .

rafaelweingartner: As far as I know we do not have a wiki page to document this script. It might be interesting to create one though.

rafaelweingartner:
@rhtyd the other changes you were talking about, are they one "Version.java"?

@khos2ow can you explain them further?

utils/src/main/java/com/cloud/maint/Version.java
```diff
@@ -19,59 +19,82 @@
...
     /**
      * Compares two version strings and see which one is higher version.
-     * @param ver1
-     * @param ver2
-     * @return positive if ver1 is higher.  negative if ver1 is lower; zero if the same.
+     *
+     * @param version1
+     * @param version2
+     *
+     * @return positive if version1 is higher.  negative if version2 is lower; zero if the same.
```
DaanHoogland: this comment is wrong:
version1 higher
version2 lower

DaanHoogland:
@khos2ow I spotted a strangeness in the javadoc in Version.java. Can you take a look?

utils/src/main/java/com/cloud/maint/Version.java
DaanHoogland: please change

khos2ow: Nice finding!

khos2ow: done.

khos2ow:
@rhtyd @rafaelweingartner my goal in this PR was to be able to support

- custom branding in POM version or provide it on the fly on CLI
- concept of timestamp for SNAPSHOT versions (i.e. development cycles)

in final artifacts (rpm, deb, systemvm-templates) and to achieve this I ended up modifying some files

- `Version.java` for instance to correctly handle `x.y.z.a-brand-[SNAPSHOT|timestamp]`
  - for MS to detect if VR needs to be upgraded
  - shown in `/etc/cloudstack-release`
- `setnextversion.sh` to add `--no-commit` flag to change the version on the fly (from SNAPSHOT to `timestamp`, if specified) in packaging script to have consistency between deb, rpm versions and ACS version itself (which is read from final jar package version)

rafaelweingartner:
@khos2ow thanks for the explanations. So, you needed to change that because of the changes you did on packaging, which would generate new version numbers pattern for code versions, right?

@DaanHoogland are you still -1 here?

rafaelweingartner:
@DaanHoogland thanks for the update here. So, should we update for you to run the tests as well?

DaanHoogland:
No, I don't think production code has been touched. maybe test installs?

khos2ow:
@rafaelweingartner yes, that's correct.

rafaelweingartner:
Well.. that version.java was changed, and I think that affects ACS rode when it deals with system VMs. What do you think?

khos2ow:
I'm +1 on testing (may be individually on different environments). I don't think automated test would give us much needed information anyway (it helps regardless anyway).
What I would do is:

- build with `--use-timestamp`
  - once RPMs
  - once DEBs
- build systemvm. preferably with
```bash
VERSION=$(mvn -q -Dexec.executable="echo" -Dexec.args='${project.version}' --non-recursive org.codehaus.mojo:exec-maven-plugin:1.3.1:exec)
TIMESTAMP=$(date +%s)
VERSION=$(echo $VERSION | sed 's/\-SNAPSHOT/\-'${TIMESTAMP}'/g')
./build.sh systemvmtemplate $VERSION $BUILD_NUMBER
```
- manually deploy the whole thing

What to look for:

- .rpm or .deb file to have a name like `cloudstack-xxx-<epoch>.[deb|rpm]`
- when logged in the UI > About Cloudstack should have the full version `4.12.0.0-<epoch>`
- when ssh in the systemvm `cat /etc/cloudstack-release` should have the full version `4.12.0.0-<epoch>`
- from UI > Infrastructure > VR should detect correctly to upgrase
  - when upgraded should show the full version in the UI (`4.12.0.0-<epoch>`)

DaanHoogland:
good point @rafaelweingartner! @khos2ow can you add unit tests for the class.
Please both note that there is a 'new' class called CloudStackVersion as well. Not sure if we should consider that in this work.
@blueorangutan test

rafaelweingartner:
@khos2ow I will test the package generation in both centoOs7 and CentOs6. I see that @DaanHoogland already started some other tests, so this should cover if the changes in version check

@DaanHoogland I do think unit-test cases are appreciated. If @khos2ow is willing to create them, it would be great.

rafaelweingartner:
Regarding the version thing....
@khos2ow what about deleting that `version.java` class and using only the "CloudStackVersion"? It is quite confusing having these two objects around.

khos2ow:
@rafaelweingartner @DaanHoogland I did consider using `CloudstackVersion.java` but wanted to be less disruptive and keep `Version.java` alive, although it can easily be deleted in favor of CloudstackVersion because there are only a handful of places which use it.

rafaelweingartner:
Please  do the honors and delete it :)

DaanHoogland:
yes @khos2ow , please replace either one or the other or vice versa. don't forget the test class as well, if you will.

borisstoyanov:
@blueorangutan package

khos2ow:
@DaanHoogland @rafaelweingartner can you take a look at the cleanup commit?
I think I leave it as this and don't squash them, what do you say?

rafaelweingartner:
Everything seems to be ok. I would leave the commits separated to facilitate reviews.

DaanHoogland:
The code in itself looks good @khos2ow, except that i am missing a link to the db. How and where is the connection between CloudstackVersion and VersionVO? Not sure if we need to worry to much but it is a reason for extensive testing.
@blueorangutan package

DaanHoogland:
@blueorangutan test

khos2ow:
@DaanHoogland there's no relation between `VersionVo` and `CloudStackVersion`. The latter is just a java representation of a _software version_ which is only one field from all the history of versions captured in VersionVo.

rafaelweingartner:
@DaanHoogland would you like to run the integration tests before we merge this one? Now everything seems to be ready for testing and then merging.

borisstoyanov:
@blueorangutan test

rafaelweingartner:
@borisstoyanov do you have the tests results for this PR?

borisstoyanov:
@rafaelweingartner the environment has failed, let me re-run it.
@blueorangutan test

DaanHoogland:
@khos2ow you can do packaging, only test is restricted to us due to resource scheduling/limitations
@blueorangutan package

khos2ow:
@DaanHoogland I see, good to know!

khos2ow:
@DaanHoogland is there a way to see the error of packaging? I believe this is related to #2485

DaanHoogland:
I'm afraid this is not public and the error is a git clone/fetch problem in this case;
ERROR: Timeout after 10 minutes
ERROR: Error cloning remote repo 'origin'
nothing to do with your change, i suppose. Let's see if insanity helps:
@blueorangutan package

DaanHoogland:
@blueorangutan test

borisstoyanov:
@blueorangutan test

DennisKonrad:
I think this introduced an error within
Home > Infrastructure > Virtual Routers

If we try to see the VRs there we get the following error:

org.apache.cloudstack.utils.CloudStackVersion.parse(String) passed 0, but requires a value in the format of int.int.int(.int)(-<legacy patch>)

Can someone confirm this?

rafaelweingartner:
what ACS versionare you using?

DennisKonrad:
ACS version is the master as of today

rhtyd:
Can you open an issue @DennisKonrad

DennisKonrad:
@rhtyd 
Did so: #2917

PR #2437, opened 2018-01-30
[CLOUDSTACK-10257]Create template/volume does not allow to specify HVM requirements

rafaelweingartner:
When creating a template from a snapshot of a ROOT volume, which has a PV operating system, ACS creates the template with flag HVM=1, which may cause problems to deploy the aforementioned template in environments that do not have hosts that support HVM. When creating templates from Snapshots or volumes an option to specify if the template that is going to be created needs HVM or not should be presented.


DaanHoogland:
and some javascript refactoring along the way :+1:

rafaelweingartner:
It is not as much refactoring as it is extracting duplicated code to a common method.

BTW, the JavaScript part of ACS looks like created by someone that was charging by the lines.... I have never seen so many repeated code in my whole life.

rafaelweingartner:
This PR will create add the HVM option as illustrated in the following image. The option will appear in both, create template from snapshot and create template from volume buttons.

image

rafaelweingartner:
@khos2ow would you mind reviewing this one?

rafaelweingartner:
Thanks @khos2ow.
@DaanHoogland are you ok here?

PR #2442, opened 2018-02-01
CLOUDSTACK-10147 Disabled Xenserver Cluster can still deploy VM's

houthuis:
ENVIRONMENT 
===================== 
XenServer Version : 6.2 , 7

ISSUE 
================== 
Disabled Xenserver Cluster can still deploy VM's , hosts in the cluster are still active

Repro. steps followed 
================== 
Disabled Cluster from UI. 
Deploy a new VM

Expected Behavior 
===============
After disabling the cluster , the hosts should be disabled. and no VM's can be deployed

Note:
it's the same results for XenServer or simulator, can't deploy on disabled hosts, but can deploy on disabled cluster

Solution:

Added a check to skip disabled clusters when selecting a host to deploy on.
Deploying on a disabled cluster will now result in a InsufficientServerCapacityException, if no enabled clusters are found.

i didn't want to propagate disabling a cluster down to the hosts, because then you would have to enable all the hosts again when you enable the cluster, and we won't know which hosts should be left in a disabled state

### DIFF
--- a/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
+++ b/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
@@ -1040,6 +1040,11 @@ StateListener<State, VirtualMachine.Event, VirtualMachine> {
         for (Long clusterId : clusterList) {
             ClusterVO clusterVO = _clusterDao.findById(clusterId);
 
+            if (clusterVO.getAllocationState() == Grouping.AllocationState.Disabled) {
+                s_logger.debug("Cannot deploy in disabled cluster " + clusterId + ", skipping this cluster");
+                avoid.addCluster(clusterVO.getId());
+            }
+
             if (clusterVO.getHypervisorType() != vmProfile.getHypervisorType()) {
                 s_logger.debug("Cluster: " + clusterId + " has HyperVisorType that does not match the VM, skipping this cluster");
                 avoid.addCluster(clusterVO.getId());


rafaelweingartner:
@houthuis thanks for the contributions. However, I have a doubt here...
Not even root administrators can deploy VMs on disabled clusters/hosts?

houthuis:
@rafaelweingartner no not on disabled hosts, but i was able to reproduce on disabled clusters

rafaelweingartner:
I think I did not express my self properly... I meant, do we want to block root admins to deploy VMs on disabled clusters/hosts? It feels that they should be able to do so (for testing purposes for instance).

DaanHoogland:
@rafaelweingartner isn't the point of disabling that no new deployments happen on that host?

rafaelweingartner:
The first instinct is to say yes... However, can't this help administrators while debugging problems? I mean, otherwise they need to enable a host/cluster to test something, which will enable these resources for end users.

DaanHoogland:
but that's what tags are for aren't they?

rafaelweingartner:
I never used tags this way... but then you need to have tags on all of your service offerings, right?
And then you need to remove the tags of hosts the you want to isolate. Or maybe custom service offerings to system administrators with tags that are not used by service offerings of end users...

BTW: for me, this is indifferent, I am not an admin, I only wanted to lift this question to see your feedback.

borisstoyanov:
@blueorangutan package

borisstoyanov:

DaanHoogland:
@houthuis did you ask the submitter (of CLOUDSTACK-10147) about the validity of the functionality?

rafaelweingartner:
@borisstoyanov and @DaanHoogland I asked about the use of disabled clusters by root admins, because when I was working on PR #2315 people told me to use disabled hosts when there was no enabled ones to execute some storage related commands. 

I personally agree with you two that when something is disabled, it should not be used at all, but as long as we are already doing this for some other features, I think it would be interesting to allow a similar behavior here, at least for root admins.

rafaelweingartner:
@DaanHoogland are you ok with this one?
The tests passed, and everything seems to be fine. The contributor is not responding our inquiries.  
I am willing to merge it.

DaanHoogland:
@houthuis @rafaelweingartner I am fine with this.

rafaelweingartner:
I will merge then. Thanks @DaanHoogland, @borisroman and @houthuis

glennwagner:
LGTM , Thanks guys for the work.

DaanHoogland:
Hihi, mister Writer pulled into the discussion here once again ;)

PR #2445, opened 2018-02-05
CLOUDSTACK-10218: fix for forced network update in a nuage network

sgoeminn:
We had issues with the broadcast uri not being updated correctly when updating a network offering. This issue was introduced as part of the previous migration PR in which nuage code was added/refactored to allow the VR to take another ip then the .2 ip.

### DIFF
--- a/plugins/network-elements/nuage-vsp/src/com/cloud/network/guru/NuageVspGuestNetworkGuru.java
+++ b/plugins/network-elements/nuage-vsp/src/com/cloud/network/guru/NuageVspGuestNetworkGuru.java
@@ -468,11 +468,10 @@ public class NuageVspGuestNetworkGuru extends GuestNetworkGuru implements Networ
             HostVO nuageVspHost = _nuageVspManager.getNuageVspHost(network.getPhysicalNetworkId());
             VspNetwork vspNetwork = _nuageVspEntityBuilder.buildVspNetwork(vm.getVirtualMachine().getDomainId(), network);
 
-            boolean isUpgradeNetworkOffering = vm.getType() == VirtualMachine.Type.DomainRouter && vspNetwork.getVirtualRouterIp()
+            boolean vrAddedToNuage = vm.getType() == VirtualMachine.Type.DomainRouter && vspNetwork.getVirtualRouterIp()
                                                                                           .equals("null");
-            if (isUpgradeNetworkOffering) {
-                //In case of upgrade network offering
-                NetworkOffering offering = _ntwkOfferingDao.findById(network.getNetworkOfferingId());
+            if (vrAddedToNuage) {
+                //In case a VR is added due to upgrade network offering - recalculate the broadcast uri before using it.
                 _nuageVspManager.updateBroadcastUri(network);
                 network = _networkDao.findById(network.getId());
                 vspNetwork = _nuageVspEntityBuilder.buildVspNetwork(vm.getVirtualMachine().getDomainId(), network, null);
--- a/plugins/network-elements/nuage-vsp/src/com/cloud/network/guru/NuageVspGuestNetworkGuru.java
+++ b/plugins/network-elements/nuage-vsp/src/com/cloud/network/guru/NuageVspGuestNetworkGuru.java
@@ -255,7 +255,7 @@ public class NuageVspGuestNetworkGuru extends GuestNetworkGuru implements Networ
             VpcDetailVO detail = _vpcDetailsDao.findDetail(network.getVpcId(), NuageVspManager.nuageDomainTemplateDetailName);
             if (detail != null && network.getNetworkACLId() != null) {
                 s_logger.error("Pre-configured DT are used in combination with ACL lists. Which is not supported.");
-                throw new IllegalArgumentException("CloudStack ACLs are not supported with Nuage Preconfigured Domain Template");
+                throw new IllegalArgumentException("CloudStack ACLs are not supported with Nuage Pre-configured Domain Template");
             }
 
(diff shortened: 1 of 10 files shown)


plugins/network-elements/nuage-vsp/src/com/cloud/network/manager/NuageVspManager.java
```diff
@@ -119,8 +121,8 @@
 
     /**
      * Associates a Nuage Vsp domain template with a
-     * @param cmd
-     * @return
+     * @param cmd Associate cmd which contains all the data
+     * @return true on success otherwise exception is thrown
```
DaanHoogland: if only true can be returned why return? it is then pass or exception, right?

plugins/network-elements/nuage-vsp/src/com/cloud/network/guru/NuageVspGuestNetworkGuru.java
```diff
@@ -470,14 +468,14 @@ public void reserve(NicProfile nic, Network network, VirtualMachineProfile vm, D
             HostVO nuageVspHost = _nuageVspManager.getNuageVspHost(network.getPhysicalNetworkId());
             VspNetwork vspNetwork = _nuageVspEntityBuilder.buildVspNetwork(vm.getVirtualMachine().getDomainId(), network);
 
-            if (vm.getType() == VirtualMachine.Type.DomainRouter && vspNetwork.getVirtualRouterIp().equals("null")) {
+            boolean isUpgradeNetworkOffering = vm.getType() == VirtualMachine.Type.DomainRouter && vspNetwork.getVirtualRouterIp()
```
DaanHoogland: please explain; if the vm is a router and the network has no ip, then we are upgrading a network offering? I am missing the logic.

sgoeminn: @DaanHoogland if a VR is spinned in a network that currently has no virtual router ip (vspNetwork.getVRip==null), it means that there was no virtual router before and that the network offering is changed.

plugins/network-elements/nuage-vsp/src/com/cloud/network/manager/NuageVspManager.java
sgoeminn: @DaanHoogland that's correct. I will change it.

sgoeminn:
@DaanHoogland Thanks for reviewing. I tried to address your comments and update the PR description to explain the issue.

plugins/network-elements/nuage-vsp/src/com/cloud/network/guru/NuageVspGuestNetworkGuru.java
DaanHoogland: isn't it possible that the old offering had a virtual router as well? it seems to me that only oldOffering.equals(newOffering) is a proper test here.

DaanHoogland: @sgoeminn how about this question?

sgoeminn: @DaanHoogland if the old offering had a VR and the new has, in that case the broadcast uri does not need to be updated because the VR ip does not change.

DaanHoogland: How about when more then one provider type is used in an install? Can we guarantee this call is only called when? either the name isUpgradeNetworkOffering is deceiving or the test is incomplete. Is this meant to be a 'only if migrating from VR to nuage implementation' check?

sgoeminn: You are right, the name of the boolean is a bit misleading. It is in fact a VRAddedToNuage check (for example when there was no user data in the network and a user wants to add the VR as user data provider). I will change it tomorrow.

sgoeminn: @DaanHoogland I changed it!

plugins/network-elements/nuage-vsp/src/com/cloud/network/guru/NuageVspGuestNetworkGuru.java
```diff
@@ -468,11 +468,10 @@ public void reserve(NicProfile nic, Network network, VirtualMachineProfile vm, D
             HostVO nuageVspHost = _nuageVspManager.getNuageVspHost(network.getPhysicalNetworkId());
             VspNetwork vspNetwork = _nuageVspEntityBuilder.buildVspNetwork(vm.getVirtualMachine().getDomainId(), network);
 
-            boolean isUpgradeNetworkOffering = vm.getType() == VirtualMachine.Type.DomainRouter && vspNetwork.getVirtualRouterIp()
+            boolean VRAddedToNuage = vm.getType() == VirtualMachine.Type.DomainRouter && vspNetwork.getVirtualRouterIp()
```
DaanHoogland: ok, now i hate myself but this one should be called 'vrAddedToNuage'...

sgoeminn: @DaanHoogland sorry, my bad. I changed it.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

rhtyd:
LGTM, merging based on code reviews and test results.

PR #2449, opened 2018-02-09
CLOUDSTACK-10278 idempotent column addition

ernjvr:
this script has been modified to add a stored procedure that allows a developer to call it in order to add a column in an idempotent manner to a table. This means that running the same script more than once will not cause an error trying to add the same column again. It will have not effect. The script needs to tested by running an upgrade from 410 to 411 and then re-run the upgrade to test if it is idempotency.

### DIFF
--- a/engine/schema/resources/META-INF/db/schema-41000to41100.sql
+++ b/engine/schema/resources/META-INF/db/schema-41000to41100.sql
@@ -585,9 +585,9 @@ CREATE TABLE IF NOT EXISTS `cloud`.`domain_details` (
 
 CALL `cloud`.`IDEMPOTENT_ADD_COLUMN`('cloud.ldap_configuration','domain_id', 'BIGINT(20) DEFAULT NULL');
 CALL `cloud`.`IDEMPOTENT_ADD_COLUMN`('cloud.ldap_trust_map','account_id', 'BIGINT(20) DEFAULT 0');
-ALTER TABLE cloud.ldap_trust_map DROP FOREIGN KEY fk_ldap_trust_map__domain_id;
-DROP INDEX uk_ldap_trust_map__domain_id ON cloud.ldap_trust_map;
-CREATE UNIQUE INDEX uk_ldap_trust_map__bind_location ON ldap_trust_map (domain_id, account_id);
+CALL `cloud`.`IDEMPOTENT_DROP_FOREIGN_KEY`('cloud.ldap_trust_map','fk_ldap_trust_map__domain_id');
+CALL `cloud`.`IDEMPOTENT_DROP_INDEX`('uk_ldap_trust_map__domain_id','cloud.ldap_trust_map');
+CALL `cloud`.`IDEMPOTENT_CREATE_UNIQUE_INDEX`('uk_ldap_trust_map__bind_location','cloud.ldap_trust_map', '(domain_id, account_id)');
 
 CREATE TABLE IF NOT EXISTS `cloud`.`netscaler_servicepackages` (
   `id` bigint unsigned NOT NULL AUTO_INCREMENT COMMENT 'id',
@@ -607,5 +607,5 @@ CREATE TABLE IF NOT EXISTS `cloud`.`external_netscaler_controlcenter` (
   PRIMARY KEY  (`id`)
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
-ALTER TABLE `cloud`.`sslcerts` ADD COLUMN `name` varchar(255) NULL default NULL COMMENT 'Name of the Certificate';
-ALTER TABLE `cloud`.`network_offerings` ADD COLUMN `service_package_id` varchar(255) NULL default NULL COMMENT 'Netscaler ControlCenter Service Package';
+CALL `cloud`.`IDEMPOTENT_ADD_COLUMN`('cloud.sslcerts','name', 'varchar(255) NULL default NULL COMMENT ''Name of the Certificate'' ');
+CALL `cloud`.`IDEMPOTENT_ADD_COLUMN`('cloud.network_offerings','service_package_id', 'varchar(255) NULL default NULL COMMENT ''Netscaler ControlCenter Service Package'' ');
--- a/engine/schema/resources/META-INF/db/schema-41000to41100.sql
+++ b/engine/schema/resources/META-INF/db/schema-41000to41100.sql
@@ -19,8 +19,35 @@
 -- Schema upgrade from 4.10.0.0 to 4.11.0.0
 --;
 
+--;
+-- Stored procedure to do idempotent column add;
+--;
+
+DROP PROCEDURE IF EXISTS `cloud`.`IDEMPOTENT_ADD_COLUMN`;
+
+CREATE PROCEDURE `cloud`.`IDEMPOTENT_ADD_COLUMN` (
+		IN in_table_name VARCHAR(100)
+    , IN in_column_name VARCHAR(100)
+    , IN in_column_definition VARCHAR(100)
+)
+BEGIN
+
+    DECLARE CONTINUE HANDLER FOR 1060 BEGIN END;
+
+	SET @ddl = CONCAT('ALTER TABLE ', in_table_name);
+	SET @ddl = CONCAT(@ddl, ' ', 'ADD COLUMN') ;
+	SET @ddl = CONCAT(@ddl, ' ', in_column_name);
+	SET @ddl = CONCAT(@ddl, ' ', in_column_definition);
+
+	PREPARE stmt FROM @ddl;
+	EXECUTE stmt;
+	DEALLOCATE PREPARE stmt;
+
+END;
+
 -- Add For VPC flag
-ALTER TABLE cloud.network_offerings ADD COLUMN for_vpc INT(1) NOT NULL DEFAULT 0;
+CALL `cloud`.`IDEMPOTENT_ADD_COLUMN`('cloud.network_offerings','for_vpc', 'INT(1) NOT NULL DEFAULT 0');
+
 UPDATE cloud.network_offerings o
 SET for_vpc = 1
 where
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@fmaximus adding you to reviewers as the report on users@ list is about for_vpc in networkoffering

DaanHoogland:
@rhtyd I think this warrants a 4.11.0.1 (as opposed to 4.11.1.0) , right?

rhtyd:
@ernjvr can you change the base branch of your PR to 4.11, the issue may be caused to select set of users who may have the structural changes in the existing environment. Instead of a new release, such users can be advised to use `jar uf <mgmt server jar> <updated files>` to update the schema changes/files. We can aim to include this fix for future 4.11.1.0

ustcweizhou:
this seems unnecessary.

ernjvr:
@rhtyd changed base branch to 4.11

DaanHoogland:
@ustcweizhou did you see the report on users@?

DaanHoogland:
@ernjvr  you need to really reapply this commit on 4.11 to avoid pulling in those newer commits.

DaanHoogland:
@rhtyd for avoiding an extra update script it makes sense to release it as 4.11.0.1 otherwise i am fine with any version, even 4.12

DaanHoogland:
@ernjvr some error reported by travis on the sql

borisstoyanov:
@ernjvr looks like you got ~5500 line changes, this does not look right.. maybe you submitted the PR when your branch was not in complete sync with latest master? Can you please look into that and submit your changes only?

ustcweizhou:
@DaanHoogland @rhtyd 
I would suggest to use Flyway instead current upgrade path from 4.11.

DaanHoogland:
@ustcweizhou I agree that something like that would be preferable but someone has to take the heat on implementing that. And nobody is taking the time for such a technical/non-functional improvement.

ustcweizhou:
@rhtyd @ernjvr @DaanHoogland 
As I know, this sql can be imported by mysql command, but can not be called in java.

A feasible way is to add a method in performDataMigration in engine/schema/src/com/cloud/upgrade/dao/

DaanHoogland:
@ustcweizhou I ran into that issue as well. adding the required items from java seems a feasible way to make them idempotent indeed. 
another solution would be to query the table description before adding

ernjvr:
@borisstoyanov yes, it's the way com/cloud/utils/db/ScriptRunner.java reads the META-INF/db/schema-41000to41100.sql file. It reads each line at a time and executes the sql as soon as it reads in a line that ends with a semicolon delimeter at the end. my stored procedure had many lines ending with semicolons. So it executed an incomplete stored procedure definition. I fixed by moving all lines that end with a semicolon onto one line, so that it can read in and execute the whole stored proc definition.

borisstoyanov:
Thanks @ernjvr, 
@blueorangutan package

borisstoyanov:
@blueorangutan package

ernjvr:
@DaanHoogland @rhtyd @fmaximus new changes pushed. will you please review?

borisstoyanov:
@blueorangutan test

rafaelweingartner:
After this script is merger, we will have to enforce that everyone creating scripts to alter tables to use the so called "IDEMPOTENT_ADD_COLUMN", is that the case?

engine/schema/resources/META-INF/db/schema-41000to41100.sql
```diff
@@ -19,8 +19,54 @@
 -- Schema upgrade from 4.10.0.0 to 4.11.0.0
 --;
 
+--;
+-- Stored procedure to do idempotent column add;
+--;
+DROP PROCEDURE IF EXISTS `cloud`.`IDEMPOTENT_ADD_COLUMN`;
```
rhtyd: @ernjvr /cc @DaanHoogland @rafaelweingartner  - is it possible to refactor these methods out to a Java/sql file, and automatically include them in all sql files so that these procedures may be reused (somehow)? Or, perhaps explore migrating to something else that provides these abstractions, perhaps flywaydb etc?

rafaelweingartner: I would also rather migrate to something like flywaydb too. However, I think it would required quite more work than this one here. What do you think? I have done this analysis before, but I was not having much time, and the discussion died [1]. Maybe we could pick this topic up again?

[1] http://markmail.org/message/2dxtcnqx2euq5tl2

ernjvr: @rafaelweingartner yes, we will have to make sure that everyone creating new scripts to add a new column, drop a foreign key or create/drop indexes, call these stored procedures, to ensure their SQL statements are idempotent. 
@rhtyd These 'drop procedure' statements are placed before each 'create procedure' statement only. So when the upgrade script runs the procedures are created but not dropped after the script completes. Therefore the stored procedures will still exist in the database. So they will be available (reusable) to future versions of schema-nn-nn.sql migration scripts. I think these 'drop/create procedure' statements also need to be added to the setup/db/create-schema.sql script to make them available to new cs installs, including the simulator.

rafaelweingartner:
Folks, what are the thoughts here? use this ad-hoc approach? Work a bit harder to a more comprehensive solution using a framework/system to control DB data model upgrades?

ernjvr:
Hi Rafael, guys. Any resolution regarding this PR? Perhaps this PR as an interim solution until someone has an opportunity to do a more comprehensive solution?

DaanHoogland:
@ernjvr I think you want to address @rafaelweingartner instead of Rafael. Better change on reaction.

rafaelweingartner:
@ernjvr I would go with this one. It does not mean that I like this type of solution. It is a palliative one, and we already have so much of these things around. However, I do understand that to change the database upgrade scheme that we use now is something way more complex (at least to do something good). I would expect us to change the DB schema upgrade method in a future ACS 5 for instance.

ustcweizhou:
@ernjvr have you tested it ?
If you import this sql file into mysql or mariadb, it might work.
However, In my experience, java does not accept procedure in sql. It uses CallableStatement
https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#prepareCall(java.lang.String)

ernjvr:
Thanks for the tip @DaanHoogland. Understood @rafaelweingartner. Yes @ustcweizhou, I tested with both mysql and java. In my java test class, I invoked the com/cloud/utils/db/ScriptRunner.java to execute the engine/schema/resources/META-INF/db/schema-41000to41100.sql script, since this same class is invoked during the upgrade to execute these schema-xxx.sql scripts.

DaanHoogland:
Discussion seems to go stale, are we now done with all comments @ernjvr ?
@blueorangutan package

rafaelweingartner:
I believe so, but we still need to find a better way technical and protocol to better manage data model changes in the database.

DaanHoogland:
@blueorangutan test

DaanHoogland:
@fmaximus @rafaelweingartner @marcaurele @rhtyd @ustcweizhou , can we please have further reviews for this? (or mpre discussion) Of couse a flyway kind of solution need to be our long term solution.

rafaelweingartner:
I think we can go forward with this PR. I notice a Flyway DB PR, but that is not all. We need a protocol to create/maintain scripts (especially scripts created in minor versions). Also, in my opinion, if we are going with flyway, we need to lock down on a version, and remove all legacy thing.

marcaurele:
@ernjvr what are the benefits of using such store procedure versus using the explicit SQL syntax `IF NOT EXISTS` / `IF EXISTS` during `ALTER TABLE` queries?

marcaurele:
https://mariadb.com/kb/en/library/alter-table/#add-column

ernjvr:
@marcaurele using these stored procedures adds functionality independent of any database version. I did consider the IF NOT EXISTS option. According to the link you supplied it is only available from version 10.3.0 but the according to the installation guide "CloudStack has been tested with MySQL 5.1 and 5.5. These versions are included in RHEL/CentOS and Ubuntu." The stored procedures also address foreign keys and indexes.

ernjvr:
@blueorangutan package

ernjvr:
@blueorangutan test

ernjvr:
@blueorangutan package

ernjvr:
@blueorangutan test

ernjvr:
@blueorangutan package

ernjvr:
@blueorangutan test

ernjvr:
@rhtyd @rafaelweingartner do you approve?

rafaelweingartner:
I am ok with the changes here.

rafaelweingartner:
If this one is merged, we will need another PR to organize things at the master version. To create the upgrade path from 4.11.1.0 to 4.12

rhtyd:
Will re-kick test.
@blueorangutan test

ernjvr:
Upgrade Test Result: I installed a version 4.6 environment, created a few networks: isolated, shared, vpc and created vms on each. I proceeded with the upgrade to 4.11 using this PR code. After the upgrade i had to add missing braces to some environment variables in the /usr/lib/systemd/system/cloudstack-management.service file and also had to configure the /etc/default/cloudstack-management file with the correct 4.11 environment variables. After starting the management server and running the migrate-dynamicroles.py I was able to log in to the UI and observe that the vms were back up and running. A successful upgrade.

DaanHoogland:
merging based on LGTMs and test reports

rafaelweingartner:
@DaanHoogland are you going to introduce the changes here in master as well?

rafaelweingartner:
@DaanHoogland here we have the same "test_04_rvpc_network_garbage_collector_nics" test appearing.

DaanHoogland:
@rafaelweingartner the changes should be forward merged to master and i am procastrinating that task.
I see that test was not on my eyes as being unstable yet. maybe something got merged ;)

rafaelweingartner:
I noticed that you did not forward the merge of this PR. I actually put this in my task list, but if you are going to do I will wait for you.

rafaelweingartner:
I closed the revert PR because after re-reading this one I got confused. Did we have an upgrade problem between 4.10 and 4.11, and that is why this PR was created?

DaanHoogland:
@rafaelweingartner this was fixing a 4.10 to 4.11 upgrade and yes it should go into 4.11.1 and not 4.11 but the upgrade will still fail for people with backported features So please revert the revert

DaanHoogland:
@rafaelweingartner also the upgrade from 4.11 to 4.11.1 has been created in #2490 . it isn't merged yet :(

rhtyd:
@rafaelweingartner I've fwd merge after apply 4.9.3.1 on 4.10->4.11->master. I'll discuss (in coming week) about 4.11.1.0 release that will most likely be before 4.12.0.0, in which case we can rename the paths as follows:
4.11.0.0->4.11.1.0->4.12.0.0 on master branch.

rafaelweingartner:
@DaanHoogland I did not revert the code. I only created the rever PR, but after thinking I closed it.
I thought this was a little confusing because people using 4.11 already would not have this script executed. Therefore, this can cause problems when upgrading to 4.12, if we use the procedures created by this PR in 4.12.

Thanks for the feedback guys.
Wish you a happy Easter.

PR #2452, opened 2018-02-12
CLOUDSTACK-10285: Fix db upgrade failure for 4.10.0.0 users

rhtyd:
4.10.0.0 users when upgrade to 4.11.0.0 may face db related
discrepancies due to some PRs that got merged without moving their sql
changes to 4.10->4.11 upgrade path. The 4.10.0.0 users can run those
missing sql statements manually and then upgrade to 4.11.0.0, since a
workaround like this is possible this ticket is not marked a blocker. In
4.11.1.0+, we'll move those changes from 4.9.3.0->4.10.0.0 upgrade path
to 4.10.0.0->4.11.0.0 upgrade path. Ideally we should not be doing this,
but this will fix issues for a future 4.10.0.0 user who may want to
upgrade to 4.11.1.0 or 4.12.0.0+.

The failure hit is missing column `service_package_id`.
Issue was caused primarily due to `884606f77be7621944d2f315fd8eabe8884c45d9`.

Pinging for review - @PaulAngus @borisstoyanov @DaanHoogland @wido @karuturi @sateesh-chodapuneedi and others.

@blueorangutan package


engine/schema/resources/META-INF/db/schema-41000to41100.sql
```diff
@@ -546,3 +546,24 @@ ALTER TABLE cloud.ldap_trust_map ADD COLUMN account_id BIGINT(20) DEFAULT 0;
...
+  `id` bigint unsigned NOT NULL AUTO_INCREMENT COMMENT 'id',
+  `uuid` varchar(255) UNIQUE,
+  `username` varchar(255) COMMENT 'username of the NCC',
+  `password` varchar(255) COMMENT 'password of NCC',
+  `ncc_ip` varchar(255) COMMENT 'IP of NCC Manager',
+  `num_retries` bigint unsigned NOT NULL default 2 COMMENT 'Number of retries in ncc for command failure',
+  PRIMARY KEY  (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
+
+ALTER TABLE `cloud`.`sslcerts` ADD COLUMN `name` varchar(255) NULL default NULL COMMENT 'Name of the Certificate';
```
ustcweizhou: Is it possible that `name` field already exists in `sslcerts` table ?
If yes, upgrade will fail here.

rhtyd: The assumption is it won't exist in before 4.11.0.0 release. Users of 4.9.3.0 or earlier will get this change when they upgrade to 4.11.0.0 but 4.10.0.0 users won't. See the dev/user ML for the discussion/proposal. But I see what you mean, the query is not idempotent which perhaps @ernjvr and @borisstoyanov may help improve.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

engine/schema/resources/META-INF/db/schema-41000to41100.sql
ustcweizhou: @rhtyd thanks for your explanation. 
LGTM now.

ernjvr: if pull request https://github.com/apache/cloudstack/pull/2449 gets approved and merged, we can make this add column statement idempotent afterwards

rhtyd:
Thanks @ustcweizhou for your lgtm. I'll wait for others for their feedback/review.
@ernjvr I think we can merge this PR first and you can modify/make the changes idempotent using the macros you've defined in your PR.

rhtyd:
Merged based on code reviews and test results.

PR #2453, opened 2018-02-14
Isolated network can not define cidr

ganlei333:
Problem file: network.js
In the 744 lines of code there is duplicate access to the isolated guest network gateway, the subnet settings are not getting correctly, causing the api to fail to get the subnet mask setting of the isolated guest network. Will be overridden by the zone's default subnet mask setting.


DaanHoogland:
@ganlei333 how do you want this . Any 4.11 commit should be merged forward anyway. I don't think this should happen with a PR

rhtyd:
@ganlei333 did you open a PR from 4.11 branch, your description does not match the diff

PR #2455, opened 2018-02-15
CLOUDSTACK-10292:Hostname in metadata when using external DNS is inco…

mrunalinikankariya:
In the current implementation, both the local-hostname and instance-id in the metadata are pointing to same thing when VM is deployed in a network with Network Offering that has no services. 
The metadata for the VM is presented in configdrive.iso for a VM having network with network offering that has no services. However the metadata doesn't have any info around the name of the VM user has passed when deploying the VM, instead the local_hostname.txt in the metadata refers to VM internal name "i-12-254-VM" doesn't it needs to be the name of the VM user has passed when creating the VM?

### DIFF
--- a/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -2526,7 +2526,7 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
                     final String zoneName = _dcDao.findById(vm.getDataCenterId()).getName();
                     boolean isWindows = _guestOSCategoryDao.findById(_guestOSDao.findById(vm.getGuestOSId()).getCategoryId()).getName().equalsIgnoreCase("Windows");
 
-                    vmData = _networkModel.generateVmData(userVm.getUserData(), serviceOffering, zoneName, vm.getInstanceName(), vm.getId(),
+                    vmData = _networkModel.generateVmData(userVm.getUserData(), serviceOffering, zoneName, vm.getInstanceName(), vm.getHostName(), vm.getId(),
                             (String) profile.getParameter(VirtualMachineProfile.Param.VmSshPubKey), (String) profile.getParameter(VirtualMachineProfile.Param.VmPassword), isWindows);
                     String vmName = vm.getInstanceName();
                     String configDriveIsoRootFolder = "/tmp";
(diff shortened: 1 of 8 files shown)


mrunalinikankariya:
travis build failure doesn't seem to be related to changes. Getting timeout error

DaanHoogland:
there is an error in test_primary_storage, not a timeout.
restarting the travis job anyhow

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -2526,7 +2526,7 @@ private void orchestrateMigrateWithStorage(final String vmUuid, final long srcHo
                     final String zoneName = _dcDao.findById(vm.getDataCenterId()).getName();
                     boolean isWindows = _guestOSCategoryDao.findById(_guestOSDao.findById(vm.getGuestOSId()).getCategoryId()).getName().equalsIgnoreCase("Windows");
 
-                    vmData = _networkModel.generateVmData(userVm.getUserData(), serviceOffering, zoneName, vm.getInstanceName(), vm.getId(),
+                    vmData = _networkModel.generateVmData(userVm.getUserData(), serviceOffering, zoneName, vm.getInstanceName(), vm.getHostName(), vm.getId(),
```
rafaelweingartner: @mrunalinikankariya you want to send the name set by the user for the given VM, right? They are the parameters "displayname" and "name" that the user can enter when deploying a VM. 

Even though this "hostName" contains that parameter "name", I get the feeling that it is not the right one to use. It looks like an inconsistent sent of value that can and should be fixed in the future. If you look into the VM's POJO, we have hostId and hostName; for sanity, I would consider them the same thing. Therefore, hostId is the id of the host where the VM is running, and the hostName is the name of the host where the VM is running.

Therefore, I believe you should use "displayName" or "name" attributes instead of the "hostName".

mrunalinikankariya: There is no displayname parameter in VirtualMachine.java. Also hostName refers to name parameter only

rafaelweingartner: Of course there is. I am not talking about the object you create we the call `generateVmData`. I am talking about the object from which you are retrieving data `vm`. That object is an instance of `UserVmVO`, which extends `VMInstanceVO`.

rhtyd:
@mrunalinikankariya can you rebase against 4.11, and edit/send the PR against 4.11? /cc @PaulAngus @ernjvr @DaanHoogland

rhtyd:
I've took some changes and ported them here: https://github.com/apache/cloudstack/pull/2645/files
I'll close this PR now.

PR #2456, opened 2018-02-15
[CLOUDSTACK-10293] Single view network ACL rules listing

rafaelweingartner:
The ACL rules editing/addition page is not user-friendly. Users are not able to see in a single view all of the detail of the ACL rule (they need to use a scroll bar on the horizontal). The problem becomes worse when there are a considerable number of rules. Therefore, we are proposing the following changes:

1- Instead of using the table to create new ACL, we can create a button like the one presented in attached pictures, where users can click, and then a modal popup would appear and users would be able to create the new ACL there. This is similar to the workings of the ACL edit button.
2 - Remove the ability to add new ACL via table where they are presented. All ACLs should be entered via the “New ACL” button. Therefore, the section “Add ACL” would be removed as well;
3 - Move the action section of the list ACL table to the most left position;

These changes would reduce the information in the table and facilitate users to add new rules and easily edit them as well.

The previous described changes will create a page similar to the following:
image


rhtyd:
Screenshots?

rafaelweingartner:
Yes, it is a screen shot. I just put it to show people how the page will become.

rafaelweingartner:
How do we proceed with this one? Will reviewers feedback be enough? No code was changed in the server side.

ui/scripts/vpc.js
```diff
@@ -436,6 +401,38 @@
                 isOptional: true,
                 isEditable: true
             },
+            'icmptype': {
+                edit: true,
+                label: 'ICMP.type',
+                desc: 'Please specify -1 if you want to allow all ICMP types',
```
nvazquez: How about adding internationalization to desc field? (the same for icmpcode below)

ui/scripts/vpc.js
```diff
@@ -1333,35 +1321,65 @@
...
+                                                    if(jQuery('#details-tab-aclRules').siblings('div.toolbar').children('div.add').size() === 0){
+                                                        var $addAclRuleDivButton = jQuery('<div>').addClass('button add');
+                                                        var $spanAddAclRuleButtonMessage = jQuery('<span>').html(_l('label.add.ACL'));
+                                                        
+                                                        $addAclRuleDivButton.html($spanAddAclRuleButtonMessage);
+                                                        $addAclRuleDivButton.click(function(){
+                                                        	cloudStack.dialog.createForm({
+                                                        		 form: {
+                                                                     title: 'label.add.rule',
+                                                                     desc: 'Create a new ACL rule',
```
nvazquez: How about using internationalization here as well?

ui/scripts/vpc.js
```diff
@@ -1333,35 +1321,65 @@
...
+                                                                	 var data = argsLocal.data;
+                                                                	 data.aclid = argsLocal.context.aclLists[0].id;
+                                                                	 if(data.protocol != 'icmp'){
+                                                                		 data.icmpcode = undefined;
+                                                                		 data.icmptype  = undefined;
+                                                                	 }
+                                                                	 if(data.protocol != 'protocolnumber'){
+                                                                		 data.protocolnumber = undefined;
+                                                                	 }
+                                                                	 if(data.protocol === 'all'){
```
nvazquez: Shouldn't be safer to perform these checks on the server side depending on the protocol parameter for createNetworkACL API method instead of undefining parameters to be sent?

rafaelweingartner: Yes, and they are in fact already performed there. The point is that if some of these combinations are sent to the server side, such as ICMP codes and TCP protocol, the server side will throw an exception. That is why I am cleaning up undesired information here (based on the protocol selected).

ui/scripts/vpc.js
rafaelweingartner: Sure, I was maintaining everything as before, but why not improve this one as well.
I will do that in a sec.

ui/scripts/vpc.js
rafaelweingartner: Done

rafaelweingartner:
Thanks for your review @nvazquez !
I have made the changes you suggested.

ui/scripts/vpc.js
nvazquez: Got it. Thanks for clarifying this

nvazquez:
Thanks @rafaelweingartner! Now code LGTM

GabrielBrascher:
Thanks, @rafaelweingartner! Code and Screenshots LGTM.

rafaelweingartner:
Two LGTMs, and as it is only a change in JavaScript there is no need for the integration tests. I will merge this one then.

PR #2459, opened 2018-02-18
CLOUDSTACK-10297: Isolation network defined subnet mask failure

ganlei333:
Isolation network defined subnet mask failure
The problem file is that the UI's network.js 744th line has missed the mask and repeats the gateway


DaanHoogland:
this is not to go in, it is not based on the right version and takes in loads of unwanted stuff in its slur.

PR #2463, opened 2018-02-20
[CLOUDSTACK-10302] Create database path upgrade from 4.11.0.0 to 4.12.0.0

rafaelweingartner:


rhtyd:
@blueorangutan package

rhtyd:
@rafaelweingartner another comment - should this have a JIRA id, or just a general thought for several administrative-ish work should we have either a catchall jira ticket or decide to have no ticket for such tasks at all?

rafaelweingartner:
@rhtyd I have no arguments against or pro Jira tickets for this specific case. I can create to use here if we decided to do so.
To create the other upgrade path, I believe a different PR would be better, what do you think?

wido:
LGTM

DaanHoogland:
let's not have jira tickets if we don't need them.
@blueorangutan test

rafaelweingartner:
@DaanHoogland I agree, but this topic needs to be discussed out of Github in our mailing list. There is also another PR of mine that I only created because someone asked me .

PR #2465, opened 2018-02-21
CLOUDSTACK-10232: SystemVMs and VR to run as HVM on XenServer

khos2ow:
Publishing boot args both to grub and xenstore-data and let
cloud-early-config decides if the VM is in PV or HVM mode
to read from correct source.

### DIFF
--- a/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
+++ b/plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
@@ -283,6 +283,11 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
 
     protected StorageSubsystemCommandHandler storageHandler;
 
+    private static final String XENSTORE_DATA_IP = "vm-data/ip";
+    private static final String XENSTORE_DATA_GATEWAY = "vm-data/gateway";
+    private static final String XENSTORE_DATA_NETMASK = "vm-data/netmask";
+    private static final String XENSTORE_DATA_CS_INIT = "vm-data/cloudstack/init";
+
     public CitrixResourceBase() {
     }
 
@@ -1286,10 +1291,7 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
             if(guestOsDetails.containsKey("xenserver.dynamicMax")){
                 recommendedMemoryMax = Long.valueOf(guestOsDetails.get("xenserver.dynamicMax")).longValue();
             }
-
         }
-
-
 
         if (isDmcEnabled(conn, host) && vmSpec.isEnableDynamicallyScaleVm()) {
             // scaling is allowed
@@ -1312,7 +1314,6 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
             vmr.memoryStaticMin = vmSpec.getMinRam();
             vmr.memoryStaticMax = vmSpec.getMaxRam();
             vmr.memoryDynamicMin = vmSpec.getMinRam();
-            ;
             vmr.memoryDynamicMax = vmSpec.getMaxRam();
 
             vmr.VCPUsMax = (long) vmSpec.getCpus();
@@ -1326,17 +1327,15 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
             NicTO mgmtNic = vmSpec.getNics()[0];
             if(mgmtNic != null ) {
                 Map<String, String> xenstoreData = new HashMap<String, String>(3);
-                xenstoreData.put("vm-data/ip", mgmtNic.getIp().toString().trim());
-                xenstoreData.put("vm-data/gateway", mgmtNic.getGateway().toString().trim());
-                xenstoreData.put("vm-data/netmask", mgmtNic.getNetmask().toString().trim());
+                xenstoreData.put(XENSTORE_DATA_IP, mgmtNic.getIp().toString().trim());
+                xenstoreData.put(XENSTORE_DATA_GATEWAY, mgmtNic.getGateway().toString().trim());
+                xenstoreData.put(XENSTORE_DATA_NETMASK, mgmtNic.getNetmask().toString().trim());
                 vmr.xenstoreData = xenstoreData;
             }
         }
 
         final VM vm = VM.create(conn, vmr);
-        if (s_logger.isDebugEnabled()) {
-            s_logger.debug("Created VM " + vm.getUuid(conn) + " for " + vmSpec.getName());
-        }
+        s_logger.debug("Created VM " + vm.getUuid(conn) + " for " + vmSpec.getName());
 
         final Map<String, String> vcpuParams = new HashMap<String, String>();
 
@@ -1368,12 +1367,18 @@ public abstract class CitrixResourceBase implements ServerResource, HypervisorRe
 
         final String bootArgs = vmSpec.getBootArgs();
         if (bootArgs != null && bootArgs.length() > 0) {
+            // send boot args for PV instances
             String pvargs = vm.getPVArgs(conn);
             pvargs = pvargs + vmSpec.getBootArgs().replaceAll(" ", "%");
-            if (s_logger.isDebugEnabled()) {
-                s_logger.debug("PV args are " + pvargs);
-            }
             vm.setPVArgs(conn, pvargs);
+            s_logger.debug("PV args are " + pvargs);
+
+            // send boot args into xenstore-data for HVM instances
+            Map<String, String> xenstoreData = new HashMap<>();
+
+            xenstoreData.put(XENSTORE_DATA_CS_INIT, bootArgs);
+            vm.setXenstoreData(conn, xenstoreData);
+            s_logger.debug("HVM args are " + bootArgs);
         }
 
         if (!(guestOsTypeName.startsWith("Windows") || guestOsTypeName.startsWith("Citrix") || guestOsTypeName.startsWith("Other"))) {
--- a/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
+++ b/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
@@ -42,7 +42,14 @@ hypervisor() {
   grep -q QEMU /var/log/messages && echo "kvm" && return 0
 
   [ -d /proc/xen ] && mount -t xenfs none /proc/xen
-  [ -d /proc/xen ] && echo "xen-domU" && return 0
+  if [ -d /proc/xen ]; then
+    $(dmesg | grep -q "Xen HVM")
+    if [ $? -eq 0 ]; then  # 1=PV,0=HVM
+      echo "xen-hvm" && return 0
+    else
+      echo "xen-pv" && return 0
+    fi
+  fi
 
   vmware-checkvm &> /dev/null && echo "vmware" && return 0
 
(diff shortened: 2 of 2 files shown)


khos2ow:
Ping for review: @DaanHoogland @rafaelweingartner @rhtyd @wido

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -42,7 +42,14 @@ hypervisor() {
   grep -q QEMU /var/log/messages && echo "kvm" && return 0
 
   [ -d /proc/xen ] && mount -t xenfs none /proc/xen
-  [ -d /proc/xen ] && echo "xen-domU" && return 0
+  if [ -d /proc/xen ]; then
+    `dmesg | grep -q "Xen HVM"`
```
rhtyd: For consistency, can you use this instead $(command here)?

rhtyd:
@blueorangutan package

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
```diff
@@ -1368,12 +1368,23 @@ public VM createVmFromTemplate(final Connection conn, final VirtualMachineTO vmS
...
             vm.setPVArgs(conn, pvargs);
+
+            // send boot args into xenstore-data for HVM instances
+            Map<String, String> xenstoreData = new HashMap<>();
+
+            xenstoreData.put("vm-data/cloudstack/init", bootArgs);
+            vm.setXenstoreData(conn, xenstoreData);
+
+            if (s_logger.isDebugEnabled()) {
+                s_logger.debug("HVM args are " + bootArgs);
```
wido: Is this if statement still required to check if debug is enabled?

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
khos2ow: sure.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
khos2ow: Technically it can be removed, but as its javadoc says: `This function is intended to lessen the computational cost of disabled log debug statements.` and the fact that they are used all over the place I added the same if block.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
```diff
@@ -1368,12 +1368,23 @@ public VM createVmFromTemplate(final Connection conn, final VirtualMachineTO vmS
...
             pvargs = pvargs + vmSpec.getBootArgs().replaceAll(" ", "%");
             if (s_logger.isDebugEnabled()) {
                 s_logger.debug("PV args are " + pvargs);
             }
             vm.setPVArgs(conn, pvargs);
+
+            // send boot args into xenstore-data for HVM instances
+            Map<String, String> xenstoreData = new HashMap<>();
+
+            xenstoreData.put("vm-data/cloudstack/init", bootArgs);
```
syed: Could you please move the `vm-data/cloudstack/init` string into a class constant.

khos2ow: sure.

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
rafaelweingartner: In this type of system such as ACS, I believe it is worth the cost of a tiny bit of more processing to remove an If statement. Specially in cases such as this one, when you already have others IFs where this one is nested.

khos2ow: I didn't get your point! Are your pro on this specific IF block or against it?

rafaelweingartner: What I meant is that, in a Web API drive system such as ACS, everything is synchronous (or almost everything). Therefore, the 1 nanoseconds of improvement that the use of this IF conditional can bring is negligible. This means that the code (the code used to create this optimization) is not worth it (in my perspective).

I normally see people using this approach of checking log level before logging when creating log messages using String.format, which uses quite some processing. However, even in those cases, I do not see much need to check the log level before logging (in most cases).

I just commented to give some context on the use of this is<LogLevel>Enabled method. The only time I used this approach of checking log level was when I developed some standalone optimization algorithms that had a time constraint to be executed. Therefore, there was no space to waste clock cycles creating a log message that is not going to be used.

In summary; if I were coding, I would not use it. However, we have these things spread all over our code base. I normally remove them, but if you want to use, the choice is yours.

pdion891:
BTW we are doing some test of this on our side, to make sure we keep support of VR running as PV on legacy deployments and support now default HVM VR on xenserver 7.1+...

So far the xenstore method work well but still need few tests under high workload, VR reboot, reset,...  we should have more news in a week, our QA will only focus on xenserver 7.1 including 7.1.1(cumulative updates)

plugins/hypervisors/xenserver/src/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java
khos2ow: Makes totally sense to me. Removed the IF condition.

rafaelweingartner:
@khos2ow and @pdion891 thanks for this one!
BTW, when you mean a load test, you are only talking about reboot/reset? are you planning a load test in the sense of multiple VRs in the same host, to see how XenServer will deal with all of them using this approach?

rafaelweingartner:
Will this one require new system VMs to be built?

khos2ow:
Yes it will require new SystemVM template.

rafaelweingartner:
I see you targeted the PR against 4.11 branch. So, a system VM for 4.11, and another one for 4.11.1.

khos2ow:
I don't think we need a template for 4.11.0.0 (because there's no java and cloud-early-config to handle it) rather than only new template for 4.11.0.1.

rafaelweingartner:
4.11 already requires a new template, right? So, people that upgraded to it already had to install a new one given the changes that are in 4.11. Now, we would need to create a new one for 4.11.1.0 , given the changes in this PR, right?

khos2ow:
Correct. The new template will be mandatory only if they are working against XenServer though, but we need to generate one anyway.
BTW, the POM version is `4.11.0.1-SNAPSHOT` and not `4.11.1.0-SNAPSHOT`, if that needs to be fixed.

rafaelweingartner:
Hmm..Then, I think there is something wrong. As far as I know, the version format is x.y.z.s, where x.y are used to identify the major version; .z is the minor, and .s is the security version bit (in case we have a security version for the x.y.z).

khos2ow:
It got changed in 6ffbce6. May be it was supposed to be `4.11.1.0-SNAPSHOT`?

syed:
@khos2ow @rafaelweingartner Given the changes here, I don't think we need to rebuild an new systemvm template. The changes will get deployed using the `systemvm.iso`

rafaelweingartner:
That is what I would expect. If you take a look at the millstone for the next release of 4.11, you can see that it is 4.11.1.

@syed, so the scripts all come from systemvm.iso. That is great, then we do not need to built a new one and deal with all of the bureaucracy that it would require.

khos2ow:
@syed are you sure? I don't think with only systemvm.iso, `cloud-early-config` will be updated with new code.

syed:
Yes I’m sure, unless that logic changed in the time since I last looked

DaanHoogland:
@blueorangutan test

pdion891:
@rafaelweingartner  for the load test, I want to make sure that on a heavily loaded xenserver, will not affect data availability from xenstore at the creation of a VR. We are also running some benchmark on the VR between HVM and PV to see if moving to HVM will improve VR performance.

@khos2ow,  in 4.11 systemvm is using Debian 9 correct ? if so, it's not what we are currently testing, we are currently focusing on the 4.10 template Debian 7

khos2ow:
@syed awesome, thanks!

@pdion891 yes that's correct, it uses Debian 9 since 4.11. Would you say there will be a lot of difference - in terms of performance - between Debian 7 and 9? In terms of functionality should be fine though.

rafaelweingartner:
@pdion891 do you have a feedback regarding the result of tests conducted by you guys?
I have used this in a test environment to check the backwards compatibility with system VMs PV, and everything is ok with this use case.

khos2ow:
@rafaelweingartner we've completed the "before" benchmark (iperf test). the "after" applying patches on Xen hasn't been done yet. I will post the result here when finished. /cc @pdion891

rafaelweingartner:
@khos2ow as soon as you have some final results with these tests, please tell us, so we can evaluate the next step to be taken here.

khos2ow:
@rafaelweingartner I will. The reason for it's not done yet was that we wanted to partially upgrade Xen cluster and do one round of functional test and then fully upgrade the cluster and then re-run iperf tests.

rafaelweingartner:
@khos2ow do you have a final feedback here?
@borisstoyanov can you confirm that the tests were successful?

Everything seems to be ready to merge this one.

rhtyd:
LGTM, merging this.

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -64,10 +71,17 @@ config_guest() {
...
-     xen-domU|xen-hvm)
+     xen-pv)
           cat /proc/cmdline > $CMDLINE
           sed -i "s/%/ /g" $CMDLINE
           ;;
+     xen-hvm)
+          if [ ! -f /usr/sbin/xenstore-read ]; then
+            log_it "ERROR: xentools not installed, cannot found xenstore-read" && exit 5
+          fi
+          /usr/sbin/xenstore-read vm-data/cloudstack/init > /var/cache/cloud/cmdline
```
rhtyd: @khos2ow in 4.11 systemvmtemplate, the xenstore-read is at /usr/bin, can you check/comment why /usr/sbin was used? I've however, fixed this blocker in the stabilization PR: https://github.com/shapeblue/cloudstack/commit/3dbb2e6b4b98cba9d59b241bd0a8c2a051131cbc

khos2ow: That's strange, I really don't know what's going on. I just checked dozens of random VMs, on various Xenserver version (7.x.x+ though) and all of them had `xenstore-read` in both places! I guess using `/usr/bin` would be safer. or even check both places then log an error?
BTW I wrote a comment on your PR as well.

rhtyd: @khos2ow I think you're missing a big point and I'm wondering how you're testing your changes. This script runs inside CloudStack's systemvmtemplate that is based on Debian 9 for 4.11+. For the change in 4.11 branch, the test/changes must be performed on Debian9 based guest vm only. In our built template, the path to `xenstore-read` is at `/usr/bin`. Regarding the `virt-what` comment, again the version of virt-what is specific to the Debian 9 based systemvmtemplate that would return xen-hvm in my tests across XenServer 6.5-7.x versions. I'll therefore keep my changes that are tested against 4.11 systemvmtemplate. Temporarily hosted here are latest 4.11 templates if you want to try them: https://lab.yadav.cloud/systemvmtemplates/4.11/

khos2ow: Thanks @rhtyd for the update, that's exactly how I have been testing, I'm gonna double check on my side once again, but if you've done the testing and validated the changes are good that's really good news.

PR #2466, opened 2018-02-22
CLOUDSTACK-10305: Rare race condition in KVM migration

nvazquez:
JIRA Ticket: https://issues.apache.org/jira/browse/CLOUDSTACK-10305

There is a race condition in the monitoring of the migration process on KVM. If the monitor wakes up in the tight window after the migration succeeds, but before the migration thread terminates, the monitor will get a LibvirtException “Domain not found: no domain with matching uuid” when checking on the migration status. This in turn causes CloudStack to sync the VM state to stop, in which it issues a defensive StopCommand to ensure it is correctly synced.

Fix: Prevent `LibvirtException: "Domain not found"` caused by the call to `dm.getInfo()`

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
@@ -172,13 +172,21 @@ public final class LibvirtMigrateCommandWrapper extends CommandWrapper<MigrateCo
 
                 // pause vm if we meet the vm.migrate.pauseafter threshold and not already paused
                 final int migratePauseAfter = libvirtComputingResource.getMigratePauseAfter();
-                if (migratePauseAfter > 0 && sleeptime > migratePauseAfter && dm.getInfo().state == DomainState.VIR_DOMAIN_RUNNING ) {
-                    s_logger.info("Pausing VM " + vmName + " due to property vm.migrate.pauseafter setting to " + migratePauseAfter+ "ms to complete migration");
+                if (migratePauseAfter > 0 && sleeptime > migratePauseAfter) {
+                    DomainState state = null;
                     try {
-                        dm.suspend();
+                        state = dm.getInfo().state;
                     } catch (final LibvirtException e) {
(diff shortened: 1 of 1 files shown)


rafaelweingartner:
So, the idea to solve this rare case, is just ignore the exception that happens when you reach such conditions? Will everything else be ok, even if that exception happens? Also, if the conditions happens, you do not need to suspend the VM?

nvazquez:
Hi @rafaelweingartner, let me answer your questions by explaining the issue in a better way:

This is issue is reported under these conditions:
- VM migration has physically finished
- Migration thread executor is still not marked as terminated

From the migration monitoring process, as VM has been migrated, that domain cannot be found (we should get the destination domain from the migration thread later) and the LibvirtException is thrown: `LibvirtException “Domain not found: no domain with matching uuid”`.
So basically the idea is just logging the issue but not suspending the VM as the migration has already been performed, but the migration thread is not marked as finished. Once the migrarion thread is marked as finished, the migration monitoring process (while loop) ends and the migration command wrapper would continue its execution.
Regarding your last question, we won't need to suspend the VM in this case as the migration has already been performed. VM suspension will take place whenever the migration is in progress and those conditions are met

remibergsma:
Thanks @nvazquez, was just investigating this issue when I saw your PR. Looks pretty much what we saw as well. Will test it :)

nvazquez:
Thanks @rafaelweingartner your reviewing!
That's great @remibergsma, thanks, much appreaciated!

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

borisstoyanov:
@nvazquez can you add JIRA item please

borisstoyanov:
@nvazquez how can we change the value of vm.migrate.pauseafter in order to test this?

nvazquez:
@borisstoyanov JIRA ticket added. That setting can be set on agent.properties file

borisstoyanov:
Thanks for this fix @nvazquez, I've managed to test it out. On master (without the fix) I've set the vm.migrate.pauseafter to 1 and did a migration, the issue is reproduce every time you try to migrate and the VM is put to stopped state, when applying the fix and using the same settings I was able to migrate successfully the VM around and got nothing in the management.

nvazquez:
Thanks @borisstoyanov! @rhtyd can we merge this one?

PR #2468, opened 2018-02-26
[4.11] VR: minor fixes

resmo:
Minor fixes for new VR template after testings.

### DIFF
--- a/systemvm/debian/etc/systemd/system/cloud-postinit.service
+++ b/systemvm/debian/etc/systemd/system/cloud-postinit.service
@@ -1,7 +1,7 @@
 [Unit]
 Description=CloudStack post-patching init script
 After=cloud-early-config.service network.target local-fs.target
-Before=ssh.service
+Before=ssh.service apache2.service
 Requires=network.service
 
 [Install]
--- a/systemvm/debian/etc/rsyslog.conf
+++ b/systemvm/debian/etc/rsyslog.conf
@@ -92,7 +92,7 @@ news.notice			-/var/log/news/news.notice
 #
 # Emergencies are sent to everybody logged in.
 #
-*.emerg				*
+*.emerg				:omusrmsg:*
 
 #
 # I like to have messages displayed on the console, but only on a virtual
--- a/tools/appliance/systemvmtemplate/scripts/finalize.sh
+++ b/tools/appliance/systemvmtemplate/scripts/finalize.sh
@@ -24,6 +24,12 @@ function configure_misc() {
   echo "cloud:`openssl rand -base64 32`" | chpasswd
 }
 
+function configure_rundisk_size() {
+  # Debian's default is 10% of total RAM. This is too low when total is 256M.
+  # See https://manpages.debian.org/stretch/initscripts/tmpfs.5.en.html
+  echo "tmpfs /run tmpfs nodev,nosuid,size=20%,mode=755 0 0" >> /etc/fstab
+}
+
 function configure_sudoers() {
   cat >/etc/sudoers <<END
 Defaults	env_reset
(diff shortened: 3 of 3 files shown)


systemvm/debian/etc/systemd/system/cloud-postinit.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=CloudStack post-patching init script
 After=cloud-early-config.service network.target local-fs.target
-Before=ssh.service
+Before=ssh.service apache2.service
```
rhtyd: This may require a new systemvm template, as dynamic patching may not change the process dependency graph. (needs testing/verification).

DaanHoogland:
code lgtm, is this touched by any integration tests (apache2 availibility)?

systemvm/debian/etc/systemd/system/cloud-postinit.service
rafaelweingartner: Does this mean that 4.11.1 will require another system VM?

rhtyd: It depends, we can publish a new systemvm if needed. Those facing issues may use the new updated template. There seems to be a new (minor?) linux security patch as well that could get included as well.

rhtyd:
@resmo can you put something in the summary to describe what issues were found and how your changes address them (including details about env, specific hypervisors etc) and how we can test/evaluate them?

@blueorangutan package

systemvm/debian/etc/systemd/system/cloud-postinit.service
rhtyd: I checked the code again, the cloud-early-config service runs first that patches the vr/systemvm using the new systemvm.iso (that will contain this change) and does a systemctl daemon-reload before cloud-postinit could run. In that case, it's possible that a new systemvmtemplate is not needed. We'll need to verify this manually, or via the systemd flame graph (@resmo can you share/attach the `systemd-analyse plot > svg`? or some other log/output)

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
A new template will need to be created to test this PR, the only issue it could cause are page/memory thrashing (page-swaps) for redundant routers as that's the only VR type that almost ends up consuming all the RAM (tested/seen in kvm and vmware). @resmo can you provide some test results, such as RAM consumption for VPC or isolated RVRs on vmware etc?

rhtyd:
Merging this based on manual tests, LGTM.

systemvm/debian/etc/systemd/system/cloud-postinit.service
rhtyd: @resmo post-merging this PR I found that sometimes isolated network VRs get stuck when starting due to unable to restart apache2 via postinit script. Can you advise why apache2 need to be started after postinit script runs?

resmo: We see systemd errors unable to start apache on the first boot, It seemed to be related to missing network setup but after postinit script ran, apache were "restarted" and apache was accessible again.

It is really hard to predict the dependency tree in this scripting construct...

rhtyd: You can see the dependency graph for systemd, via systemd-analyze (plot it or analyze the critical-chain etc, refer docs for more info). Apache2 is disabled by default, so during first boot no errors should be seen. Can you share when you saw those errors, did that cause any service related failure for you, also where you saw these errors (in console or logs?). The cloud-postinst script will handle any failure and start apache2 is required, any initial error/warning may be ignored.

resmo: I meant the dependency caused in the scripts not systemd.

However, I am ok to revert this if it causes serious issues.

PR #2472, opened 2018-03-01
CLOUDSTACK-10310 Fix KVM reboot on storage issue

Slair1:
If the KVM heartbeat file can't be written to, the host is rebooted, and thus taking down all VMs running on it.  The code does try 5x times before the reboot, but the there is not a delay between the retires, so they are 5 simultaneous retries, which doesn't help.  Standard SAN storage HA operations or quick network blip could cause this reboot to occur.

Some discussions on the dev mailing list revealed that some people are just commenting out the reboot line in their version of the CloudStack source.

A better option would be have it sleep between tries so it isn't 5x almost simultaneous tries.  Plus, instead of rebooting, the cloudstack-agent could just be stopped on the host instead.  This will cause alerts to be issued and if the host is disconnected long-enough, depending on the HA code in use, VM HA could handle the host failure.

The built-in reboot of the host seemed drastic


scripts/vm/hypervisor/kvm/kvmheartbeat.sh
```diff
@@ -155,10 +155,10 @@ then
   exit 0
 elif [ "$cflag" == "1" ]
 then
-  /usr/bin/logger -t heartbeat "kvmheartbeat.sh rebooted system because it was unable to write the heartbeat to the storage."
+  /usr/bin/logger -t heartbeat "kvmheartbeat.sh stopped cloudstack-agent because it was unable to write the heartbeat to the storage."
   sync &
   sleep 5
-  echo b > /proc/sysrq-trigger
+  systemctl stop cloudstack-agent
```
borisstoyanov: @Slair1 are you sure this will work with CentOS6 and debian KVM hosts? Same goes with the change above

Slair1: Good thought, I can change that to “service” for older version compatibility

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -115,22 +115,33 @@ protected void runInContext() {
                     }
 
                     String result = null;
-                    for (int i = 0; i < 5; i++) {
+                    // Try 3 times, but sleep in between tries to ensure it isn't a short lived transient error
+                    for (int i = 1; i <= 4; i++) {
```
DaanHoogland: the comment says try three times but i count 4 iterations. either the comment or the loop-control needs to change

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -115,22 +115,33 @@ protected void runInContext() {
...
                         Script cmd = new Script(s_heartBeatPath, _heartBeatUpdateTimeout, s_logger);
                         cmd.add("-i", primaryStoragePool._poolIp);
                         cmd.add("-p", primaryStoragePool._poolMountSourcePath);
                         cmd.add("-m", primaryStoragePool._mountDestPath);
                         cmd.add("-h", _hostIP);
                         result = cmd.execute();
                         if (result != null) {
-                            s_logger.warn("write heartbeat failed: " + result + ", retry: " + i);
+                            s_logger.warn("write heartbeat failed: " + result + ", retry: " + i + " of 4");
```
DaanHoogland: if this 4 in the log message is the same one as the one in the control statement you want to extract to a constant/field/variable

bwsw: I disabled that behaviour in our 4.3 completely.

Slair1:
Comments should be addressed - i also repushed and squashed commits

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -34,7 +34,8 @@
     protected static String s_heartBeatPath;
     protected long _heartBeatUpdateTimeout = 60000;
     protected long _heartBeatUpdateFreq = 60000;
-    protected long _heartBeatUpdateMaxRetry = 3;
+    protected long _heartBeatUpdateMaxTries = 5;
```
rafaelweingartner: What about externalizing these two information as parameters for operators to customize?

scripts/vm/hypervisor/kvm/kvmheartbeat.sh
```diff
@@ -155,10 +155,10 @@ then
   exit 0
 elif [ "$cflag" == "1" ]
 then
-  /usr/bin/logger -t heartbeat "kvmheartbeat.sh rebooted system because it was unable to write the heartbeat to the storage."
+  /usr/bin/logger -t heartbeat "kvmheartbeat.sh stopped cloudstack-agent because it was unable to write the heartbeat to the storage."
   sync &
   sleep 5
-  echo b > /proc/sysrq-trigger
+  service cloudstack-agent stop
```
rhtyd: Please start a discussion on dev@, this change to simply stop the agent may not be acceptable for all cases.

rafaelweingartner: Instead, what about creating a global parameter to determine this behavior? Then, the administrator would be able to change this easily.

rhtyd: Good idea @rafaelweingartner, perhaps a cluster scope global parameter. However, such a setting need to be propagated when agent/client connects to mgmt server (say via ReadyCommand) or overridable in agent.properties file.

rafaelweingartner: Yes, the externalization for a global settings would need more work, but I think it is worth it.
We could help/guide @Slair1 if he does not where to start then.

wido:
Can VM HA handle such a failure? Let's say you have two NFS mounts on that host and once fails. You only stop the Agent, but other VMs keep running.

Afaik that can still cause a corruption where a VM is spawned twice in a cluster because the other host can't know if that VM is still running or not.

Our investigators aren't that smart. They rely on ICMP and such which might be blocked.

Slair1:
@wido My opinion is that the ability to start a VM on two hosts in the cluster is a bigger issue.  We run KVM and just enabled libvirt's lockd to stop that from being able to happen.  Maybe this PR could be changed to be an option, just reboot host or stop agent.

https://libvirt.org/locking-lockd.html

scripts/vm/hypervisor/kvm/kvmheartbeat.sh
rhtyd: Maybe you can sent the PR in two parts, the first half that increases the heartbeat retry threshold to `5` LGTM. The second part could be the configuration fix, global setting or setting in agent.properties to decide whether to reboot kvm host or stop agent.

rhtyd:
@DaanHoogland @rafaelweingartner @wido @bwsw - do you think stopping the agent can cause a bigger regression than the historic behaviour of a host reboot? I did not follow any discussion (if any) had on dev ML?

wido:
@rhtyd Both cases are not ideal. Because if you stop the Agent due to a single Primary Storage failure you leave VMs running there which still might be doing I/O and to start them a second time somewhere else.

The hard reboot isn't a very nice either, so you have to deal with two bad things.

Slair1:
@wido @DaanHoogland @rhtyd 
Thoughts on me rebasing this to 4.11?

DaanHoogland:
@Slair1  validating it will be much easier. we can forward merge but 4.11 is the about to be the current LTS version.

Slair1:
Closing and rebasing to 4.11 in PR #2722

PR #2473, opened 2018-03-01
CLOUDSTACK-10309 Add option on if to VM HA power-on a OOB-shut-off-VM

Slair1:
When a user shuts down their VM from the guest OS (and VM HA is enabled), the VM just powers itself back on.  Our environment is on KVM hosts.

CloudStack does not know the difference between a VM failing or being shutdown from within the guest OS.

This is a major pain point for all our users - especially since they don't pay for VMs when they are shutoff.  It is not intuitive for end-users to understand why they can't shutdown VMs from within the guest OS.  Especially when they all come from (non-cloudstack) VMware and Hyper-V environments where this is not an issue.

However, if a host fails, we need VM HA to still work.

This PR that creates a configuration option "ha.vm.restart.hostup".  With this option set to false, if CloudStack sees a VM shutdown out-of-band, but the host it was on is still *online*, then it won't power the VM back on.  The logic is that since the host is online, it was most likely shutdown from the guest OS.

For when a host actually fails, standard VM HA logic takes over and powers on VMs (if they have VM HA enabled) if the host they were on fails.

If that "ha.vm.restart.hostup" option is true (the default to match current functionality), it works like always, and even in-guest shutdowns of VMs causes CloudStack to power back on the VM.

### DIFF
--- a/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -359,6 +359,9 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
             Integer.class, "vm.job.report.interval", "60",
             "Interval to send application level pings to make sure the connection is still working", false);
 
+    static final ConfigKey<Boolean> HaVmRestartHostUp = new ConfigKey<Boolean>("Advanced", Boolean.class, "ha.vm.restart.hostup", "true",
+            "If an out-of-band stop of a VM is detected and its host is up, then power on the VM", true);
+
     ScheduledExecutorService _executor = null;
 
     protected long _nodeId;
@@ -3708,7 +3711,7 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
     public ConfigKey<?>[] getConfigKeys() {
         return new ConfigKey<?>[] {ClusterDeltaSyncInterval, StartRetry, VmDestroyForcestop, VmOpCancelInterval, VmOpCleanupInterval, VmOpCleanupWait,
                 VmOpLockStateRetry,
-                VmOpWaitInterval, ExecuteInSequence, VmJobCheckInterval, VmJobTimeout, VmJobStateReportInterval, VmConfigDriveLabel};
+                VmOpWaitInterval, ExecuteInSequence, VmJobCheckInterval, VmJobTimeout, VmJobStateReportInterval, VmConfigDriveLabel, HaVmRestartHostUp};
     }
 
     public List<StoragePoolAllocator> getStoragePoolAllocators() {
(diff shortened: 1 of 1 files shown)


engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -359,6 +359,9 @@ public void setHostAllocators(final List<HostAllocator> hostAllocators) {
             Integer.class, "vm.job.report.interval", "60",
             "Interval to send application level pings to make sure the connection is still working", false);
 
+    static final ConfigKey<Boolean> HaVmRestartHostUp = new ConfigKey<Boolean>("Advanced", Boolean.class, "ha.vm.restart.hostup", "true",
```
DaanHoogland: I would expect to see the new ConfigKey be added to the return of getConfigKeys()

Slair1: Agreed, not sure how that got missed here.

Slair1: fixed

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -3853,7 +3856,7 @@ private void handlePowerOffReportWithNoPendingJobsOnVM(final VMInstanceVO vm) {
         case Stopped:
         case Migrating:
             s_logger.info("VM " + vm.getInstanceName() + " is at " + vm.getState() + " and we received a power-off report while there is no pending jobs on it");
-            if(vm.isHaEnabled() && vm.getState() == State.Running && vm.getHypervisorType() != HypervisorType.VMware && vm.getHypervisorType() != HypervisorType.Hyperv) {
+            if(vm.isHaEnabled() && vm.getState() == State.Running && HaVmRestartHostUp.value() && vm.getHypervisorType() != HypervisorType.VMware && vm.getHypervisorType() != HypervisorType.Hyperv) {
```
rafaelweingartner: @Slair1 would you mind extracting these conditions to a method (e.g. `shouldRestartVirtualMachine`)? With your addition the if condition will have 5 elements, which is a lot. Extracting this bit to a method enabled documentation and unit test cases for these different cases.

wido:
Nice PR! I agree with @rafaelweingartner and @DaanHoogland , but I can live without that.

But yes, a method would make testing easier.

LGTM based on code

DaanHoogland:
@Slair1 @rafaelweingartner , can you come to agreement (maybe Rafael's sugestion here or in a separate PR?)

rafaelweingartner:
It is ok the code as is. However, it might have been interesting to see the feedback of the author.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
Slair1: Yea, that sounds like a good idea, i am booked up here for a while, but it sounds like a good idea

Slair1:
I've never created a Marvin Test before, but can look at test_vm_life_cycle.py to try and figure it out

rafaelweingartner:
I actually meant unit tests. I think they would fit better the method extraction. They would also be easier to develop. Once you get to create them, if you are unsure how to proceed, just ping me. I can help you.

rhtyd:
LGTM, @DaanHoogland @rafaelweingartner are we okay on the changes and to accept without any test?

rhtyd:
FYI - I've changed PR's base branch to validate change from Travis, and trillian. Once this is tested and is okay we can move it back to 4.9 branch.
@blueorangutan package

DaanHoogland:
@rhtyd @rafaelweingartner , i just revisited the code and stand by my the lgtm.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rafaelweingartner:
I do believe it needs code extraction and unit tests, but you want to merge it. Let's proceed and merge.

rhtyd:
Okay @rafaelweingartner I'll merge after smoketests are back.

borisstoyanov:
@blueorangutan test

rhtyd:
Test and code LGTM.

PR #2474, opened 2018-03-01
CLOUDSTACK-10246 Fix Host HA and VM HA issues

Slair1:
The HA logic just does not work.  VM's with HA enabled would never restart after a host failure.  Had to re-do most of that logic.  There are comments inline with the code, but down below is the general updated logic.  Sorry for the long notes... 

PS. We enabled virtlockd (https://libvirt.org/locking-lockd.html) and highly recommend it, otherwise you can have VM HA start a VM on multiple hosts.

We are running KVM FYI.

- If host-agent is unreachable, handleDisconnectWithInvestigation() is called as always.
- The investigators are called to see what happened, which is one of the following two scenarios.  (If it isn't one of the two below, then the host just came back UP, or another status was returned and that is also logged.  But the two scenarios below are what needed updated the most)

**If the investigators find the host is UP, but just the agent is unreachable**
The host is put into DISCONNECTED status.  It will stay in this status and the PingTimeouts will continue to call handleDisconnectWithoutInvestigation() periodically.  It will stay in DISCONNECTED status until the AlertWait config option expires.  If the AlertWait time eventually is hit, and the investigators are still just reporting that the host is DISCONNECTED and not DOWN.  Then we'll put the host into ALERT state and we'll stay there until the investigators say the host is UP or the investigators say the host is DOWN.  If the host goes DOWN, then VM HA will be initiated.

**If the investigators find the host is DOWN**
Then VM HA is initiated...

**VirtualNetworkApplianceManagerImpl.java**
The file VirtualNetworkApplianceManagerImpl.java is edited for a related VM HA problem.  When a Host is determined to be DOWN, CloudStack attempts to VM HA any affected routers.  The problem is, when the host is determined to be down, by code referenced above, the host may not actually be DOWN.  On KVM for example, the host is considered DOWN if the agent is stopped on the KVM host for too long.  In that case, the VMs could still be running just fine...  However when we think the host is DOWN, VM HA runs on the router and as part of that it unallocates/cleans-up the router and it's 169.x.x.x control IP is unallocated.  Then after it cleans it up, it tries to power on the router on another host, and as part of that it allocates a NEW 169.x.x.x control IP and writes that to the DB.  However, since the router isn't actually down (we just think the host is down) the VM HA fails as the vRouter is currently still running on the problem host.  

Next, in this example, when the host agent is back online again, it sends a power report to the mgmt servers, and the management servers think the router was powered-on OOB.  However, the GUI will not show a control IP for the vRouter, and the DB will have the NEW control IP it tried to allocated during the failed VM HA event.  Thus, leaving us unable to communicate with the vRouter.

This PR does a simple check that we can still communicate with the vRouter after any OOB power-on occurs.  If we can, then we have the correct control IP in the DB and we're good - so we do nothing.  If we can't communicate with the vRouter after the OOB power-on, we do a reboot of the vRouter to fix it.

### DIFF
--- a/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
@@ -793,6 +793,17 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
             s_logger.warn("Can't find host with " + hostId);
             nextStatus = Status.Removed;
         } else {
+            final ResourceState resourceState = host.getResourceState();
+            if (host.getType() == Host.Type.Routing && event == Event.ShutdownRequested && resourceState != ResourceState.Maintenance && removeAgent) {
+                // The host agent is shutting itself down gracefully, the host isn't in maintenance and we are removing the host's monitoring task.
+                final DataCenterVO dcVO = _dcDao.findById(host.getDataCenterId());
+                final HostPodVO podVO = _podDao.findById(host.getPodId());
+                final String hostDesc = "[name: " + host.getName() + " (id:" + host.getId() + "), availability zone: " + dcVO.getName() + ", pod: " + podVO.getName() + "]";
+                final String hostShortDesc = "Host " + host.getName() + " (id:" + host.getId() + ")";
+                s_logger.warn(hostShortDesc + " has disconnected with event " + event + ",  but was in resource state " + resourceState + ", not in Maintenance.  Agent Monitor is also being removed.");
+                _alertMgr.sendAlert(AlertManager.AlertType.ALERT_TYPE_HOST, host.getDataCenterId(), host.getPodId(), hostShortDesc + " agent is shutting down",
+                                "The agent for host " + hostDesc + " has disconnected with event " + event + ",  but was in resource state " + resourceState + ", not in Maintenance.  Host should be put into Maintenance before shutting down the host agent.");
+            }
             final Status currentStatus = host.getStatus();
             if (currentStatus == Status.Down || currentStatus == Status.Alert || currentStatus == Status.Removed) {
                 if (s_logger.isDebugEnabled()) {
@@ -809,16 +820,18 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
                 }
 
                 if (s_logger.isDebugEnabled()) {
-                    s_logger.debug("The next status of agent " + hostId + "is " + nextStatus + ", current status is " + currentStatus);
+                    s_logger.debug("The next status of agent " + hostId + " is " + nextStatus + ", current status is " + currentStatus);
                 }
             }
         }
 
-        if (s_logger.isDebugEnabled()) {
-            s_logger.debug("Deregistering link for " + hostId + " with state " + nextStatus);
+        if (removeAgent) {
+            if (s_logger.isDebugEnabled()) {
+                s_logger.debug("Deregistering link for " + hostId + " with state " + nextStatus);
+            }
+            removeAgent(attache, nextStatus);
         }
 
-        removeAgent(attache, nextStatus);
         // update the DB
         if (host != null && transitState) {
             disconnectAgent(host, event, _nodeId);
@@ -830,6 +843,7 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
     protected boolean handleDisconnectWithInvestigation(final AgentAttache attache, Status.Event event) {
         final long hostId = attache.getId();
         HostVO host = _hostDao.findById(hostId);
+        boolean removeAgent = false;
 
         if (host != null) {
             Status nextStatus = null;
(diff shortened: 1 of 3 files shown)


engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -843,72 +846,103 @@ protected boolean handleDisconnectWithInvestigation(final AgentAttache attache,
...
+                return true;
+            }
+
             if (nextStatus == Status.Alert) {
-                /* OK, we are going to the bad status, let's see what happened */
-                s_logger.info("Investigating why host " + hostId + " has disconnected with event " + event);
+                /* Our next Agent transition state is Alert
+                 * Let's see if the host down or why we had this event
+                 */
+                s_logger.info("Investigating why host " + hostShortDesc + " has disconnected with event " + event);
```
DaanHoogland: 👍 good improvement, but though it is only (a comment and) a log statement, this entails an interface of the system. the ecosystem may query logs for the text and no longer find the hostId thus not being able to take mitigating actions any more. I'd rather see a less destructive change like 'hostId + " (" + hostShortDesc + ") "'

We may get away with it but it does require extensive testing by the whole community :/.

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -843,72 +846,103 @@ protected boolean handleDisconnectWithInvestigation(final AgentAttache attache,
...
 
                 Status determinedState = investigate(attache);
                 // if state cannot be determined do nothing and bail out
                 if (determinedState == null) {
                     if ((System.currentTimeMillis() >> 10) - host.getLastPinged() > AlertWait.value()) {
-                        s_logger.warn("Agent " + hostId + " state cannot be determined for more than " + AlertWait + "(" + AlertWait.value() + ") seconds, will go to Alert state");
+                        s_logger.warn("State for " + hostShortDesc + " could not be determined for more than " + AlertWait + "(" + AlertWait.value() + ") seconds, will go to Alert state");
```
DaanHoogland: for warn message the above is even more true

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -843,72 +846,103 @@ protected boolean handleDisconnectWithInvestigation(final AgentAttache attache,
...
-                } else if (determinedState == Status.Up) {
-                    /* Got ping response from host, bring it back */
-                    s_logger.info("Agent is determined to be up and running");
+                if (determinedState == Status.Up) {
+                    // Got ping response from host, bring it back
+                    s_logger.info(hostShortDesc + " is up again");
                     agentStatusTransitTo(host, Status.Event.Ping, _nodeId);
-                    return false;
                 } else if (determinedState == Status.Disconnected) {
-                    s_logger.warn("Agent is disconnected but the host is still up: " + host.getId() + "-" + host.getName());
```
DaanHoogland: why must this statement be removed?

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -843,72 +846,103 @@ protected boolean handleDisconnectWithInvestigation(final AgentAttache attache,
...
+                    // Investigation says host isn't down, just disconnected
                     if (currentStatus == Status.Disconnected) {
+                        // Last status was disconnected, only switch status if AlertWait has passed
```
DaanHoogland: Can you extract these bits of code in methods, please?

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -843,72 +846,103 @@ protected boolean handleDisconnectWithInvestigation(final AgentAttache attache,
...
             } else {
-                s_logger.debug("The next status of agent " + host.getId() + " is not Alert, no need to investigate what happened");
+                s_logger.info("The next status of host " + hostShortDesc + " is " + nextStatus + ", no need to investigate what happened");
             }
         }
-        handleDisconnectWithoutInvestigation(attache, event, true, true);
-        host = _hostDao.findById(hostId); // Maybe the host magically reappeared?
+
+        handleDisconnectWithoutInvestigation(attache, event, true, removeAgent);
+        host = _hostDao.findById(hostId); // We may have transitioned the status - refresh
```
DaanHoogland: protected boolean handleDisconnectWithInvestigation(final AgentAttache attache, Status.Event event) just grew from 85 to 118 lines. Can you please make it a bit more modular?

server/src/com/cloud/network/router/VirtualNetworkApplianceManagerImpl.java
```diff
@@ -340,6 +340,7 @@
     private ScheduledExecutorService _executor;
     private ScheduledExecutorService _checkExecutor;
     private ScheduledExecutorService _networkStatsUpdateExecutor;
+    private ExecutorService _routerOobStartExecutor;
```
DaanHoogland: we are trying to get rid of these _'s . no need to adhere to this old convention. In fact you may want to rename the others as well (in a separate commit for review ease) (this is not a request for change, just a suggestion)

server/src/com/cloud/network/router/VirtualNetworkApplianceManagerImpl.java
```diff
@@ -2587,7 +2589,13 @@ public boolean postStateTransitionEvent(final StateMachine2.Transition<VirtualMa
                 event == VirtualMachine.Event.FollowAgentPowerOnReport &&
                 newState == VirtualMachine.State.Running &&
                 isOutOfBandMigrated(opaque)) {
-            s_logger.debug("Virtual router " + vo.getInstanceName() + " is powered-on out-of-band");
+            /* Since vRouter appears to be powered-on OOB, make sure we can talk to router
+             * If we can't talk to it, we need to reboot it to get it managed correctly
+             * This is needed for example when a host agent goes down and comes back up,
+             * we would have done a failed HA event on the router and end up having our controlIP out-of-sync
+             */
+             s_logger.info("Router " + vo.getInstanceName() + " (ID:" + vo.getId() + ") is powered-on out-of-band, checking if can send CheckRouterCommand to router");
```
DaanHoogland: :+1: completely agree with this change, the ecosystem shouldn't rely on debug logging anyway. friendliness to the community dictates diligent testing anyway.

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
Slair1: @DaanHoogland good thought, I didn’t think about that.  However, the hostShortDesc does include the hostId as part of it.  So maybe it’s ok?

final String hostShortDesc = "Host " + host.getName() + " (id:" + host.getId() + ")";

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
Slair1: I removed it because it was extraneous.  A similar but more detailed log entry is performed for every case that the host is in Disconnected.  It’s just a little hard to see that while reading through the changes.

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
Slair1: Yea, that sounds good

engine/orchestration/src/com/cloud/agent/manager/AgentManagerImpl.java
DaanHoogland: yeah, maybe. intesive testing still required...

borisstoyanov:
@blueorangutan package

borisstoyanov:
@Slair1 what issues are fixed, do we have marvin tests for them? If not I think it'll be good to add them.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

DaanHoogland:
@Slair1 are you going to modularise the handleDisconnectWithInvestigation method?

Slair1:
Yea, i can work to modularize this, i unfortunately don't have the time at the moment, but can later.

On the tests, do you mean Unit Tests?  I've never wrote a unit test before, but agree it would be good to have

borisstoyanov:
@Slair1 can we also have marvin tests to cover these fixes ?

DaanHoogland:
@Slair1 I have no easy way to put this. You have four (4) PRs out and all fail all travis runs. It must be that travis doesn't like your github handle, or so. Can you think of something you have/do that might cause this. In the same period other PRs have passed travis runs...?

PR #2475, opened 2018-03-03
[CLOUDSTACK-10314] Add Text-Field to each ACL Rule

rafaelweingartner:
It is interesting to have a text field (e.g. CHAR-256) added to each ACL rule, which allows to enter a "reason" for each FW Rule created. This is valuable for customer documentation, as well as best practice for an evidence towards auditing the system.

Attached are some pictures showing how the UI looks like with this new field.
tablewithrasonfield

editaclwithreasonfield

### DIFF
--- a/api/src/main/java/com/cloud/network/vpc/NetworkACLItem.java
+++ b/api/src/main/java/com/cloud/network/vpc/NetworkACLItem.java
@@ -24,6 +24,7 @@ import org.apache.cloudstack.api.InternalIdentity;
 
 public interface NetworkACLItem extends InternalIdentity, Identity, Displayable {
 
+    @Override
     String getUuid();
 
     Action getAction();
@@ -51,7 +52,7 @@ public interface NetworkACLItem extends InternalIdentity, Identity, Displayable 
     Integer getSourcePortStart();
 
     /**
-     * @return last port of the source prot range.  If this is null, that means only one port is mapped.
+     * @return last port of the source port range.  If this is null, that means only one port is mapped.
      */
     Integer getSourcePortEnd();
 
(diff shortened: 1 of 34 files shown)


DaanHoogland:
@rafaelweingartner did you have a look at the annotations feature for adding comments? These allow for audit trails as well.

rafaelweingartner:
@DaanHoogland yes I did, but I did not quite understand their purpose. I was not sure if the goal was to be used as the tag system we have in ACS for host and storage. Also, they are quite limited in the UI if you need to write something more than a few words.

rafaelweingartner:
So, that option has nothing to do with ACS tagging system?

rafaelweingartner:
I was checking that tag thing there, and it is a bit too generic in my opinion. I could change the UI to use those fields (I would also need to increase the size of those fields), but I do prefer a not so generic solution as this one, where the description/history field becomes part of the ACL rule API.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

rafaelweingartner:
@khos2ow and @marcaurele  can you guys give your opinions here?

rafaelweingartner:
That is an interesting suggestion. I have not thought about "description" to name that field.

marcaurele:
It would be helpful for SG rules too for example, for IP aliases, snapshots...

marcaurele:
Yeah, it wasn't aimed to this PR, more as a general idea.

I'm ok with this change. If you think `description` is better than `reason` and you're willing to modify your PR, let us know otherwise we'll merge it.

rafaelweingartner:
I am still thinking about the change you mentioned. As soon as I have some feedback I tell you guys.

rafaelweingartner:
I am going to leave it as is. If we change the "reason" field name, I would need to change all references (JS, CSS, and labels) to make a more concise change. Maybe, if we feel that description is better, I can change that in the near future with another PR.

PR #2476, opened 2018-03-05
CLOUDSTACK-10317: Fix SNAT rules for additional public nics

rhtyd:
This allows networks with additional public nics to have correct SNAT iptables rules applied on configuration.

Pinging for review - @ustcweizhou @PaulAngus @DagSonstebo and others.

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsAddress.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsAddress.py
@@ -388,7 +388,7 @@ class CsIP:
             self.fw.append(["mangle", "",
                             "-A VPN_%s -j RETURN" % self.address['public_ip']])
             self.fw.append(["nat", "",
-                            "-A POSTROUTING -o eth2 -j SNAT --to-source %s" % self.address['public_ip']])
+                            "-A POSTROUTING -o %s -j SNAT --to-source %s" % (self.dev, self.address['public_ip'])])
             self.fw.append(["mangle", "",
                             "-A PREROUTING -i %s -m state --state NEW " % self.dev +
                             "-j CONNMARK --set-xmark %s/0xffffffff" % self.dnum])
--- a/systemvm/debian/opt/cloud/bin/cs/CsAddress.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsAddress.py
@@ -388,7 +388,7 @@ class CsIP:
             self.fw.append(["mangle", "",
                             "-A VPN_%s -j RETURN" % self.address['public_ip']])
             self.fw.append(["nat", "",
-                            "-A POSTROUTING -o %s -j SNAT --to-source %s" % (self.dev, self.address['public_ip'])])
+                            "-A POSTROUTING -o %s -j SNAT --to-source %s" % (self.dev, self.cl.get_eth2_ip())])
             self.fw.append(["mangle", "",
                             "-A PREROUTING -i %s -m state --state NEW " % self.dev +
                             "-j CONNMARK --set-xmark %s/0xffffffff" % self.dnum])


rhtyd:
Same logic in vpc: https://github.com/apache/cloudstack/blob/4.11/systemvm/debian/opt/cloud/bin/cs/CsAddress.py#L480

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

systemvm/debian/opt/cloud/bin/cs/CsAddress.py
```diff
@@ -388,7 +388,7 @@ def fw_router(self):
             self.fw.append(["mangle", "",
                             "-A VPN_%s -j RETURN" % self.address['public_ip']])
             self.fw.append(["nat", "",
-                            "-A POSTROUTING -o eth2 -j SNAT --to-source %s" % self.address['public_ip']])
+                            "-A POSTROUTING -o %s -j SNAT --to-source %s" % (self.dev, self.address['public_ip'])])
```
rafaelweingartner: The attribute `self.dev` is already populated with the correct device name?

rhtyd: Yes @rafaelweingartner, the logic is same what is being used in vpc routers, see - https://github.com/apache/cloudstack/blob/4.11/systemvm/debian/opt/cloud/bin/cs/CsAddress.py#L584

rafaelweingartner: Thanks for the explanation.

rhtyd:
@blueorangutan test

systemvm/debian/opt/cloud/bin/cs/CsAddress.py
rhtyd: Thanks for confirming that we indeed need to set the device/interface in the rule @ustcweizhou. I think it makes sense to have the default packets routed to the default public snat ip. I'll let others comment on it.

DagSonsteboSB: Hi @ustcweizhou - in your scenario - does this not mean you end up SNATing on the wrong interface? So lets say you have eth2/VLAN10/10.100.1.10/24 and eth3/VLAN20/10.100.2.20/24. With your method that would SNAT to

- A POSTROUTING -o eth3 -j SNAT --to-source 10.100.1.10

So in other words the IP address would be right for the initial source NAT, but it would end up being translated on the wrong interface?

ustcweizhou: @DagSonsteboSB yes, I think the outgoing IP for the vms without static nat should be the source nat IP, in your example 10.100.1.10

In my point of view, the rules applied on ethernet other than eth2 (for example eth3) are not needed at all, however it is not harmful to have them.

rhtyd: Thanks all for your discussion and testing, I'll test @ustcweizhou 's fix now.

rhtyd:
@blueorangutan package

systemvm/debian/opt/cloud/bin/cs/CsAddress.py
DagSonsteboSB: You are right @ustcweizhou - just tested this. Routing happens *after SNAT* hence translating all non-static NATed traffic to the source IP address on eth2 works even if this means a SNAT rule against eth3, eth4, etc.

rhtyd:
@blueorangutan test

rafaelweingartner:
It seems that everything is ok here, can we proceed with the merge?

PR #2478, opened 2018-03-05
[CLOUDSTACK-10318] Bug on sorting ACL rules list in chrome

rafaelweingartner:
There is a bug on ACL rules list when a user creates more than 10 rules.

Steps to reproduce this issue:
- Create more than 10 ACL rules.
- Access the ACL rules list page.
- You will see that they are not sorted as they should by the ACL rule number

### DIFF
--- a/ui/scripts/vpc.js
+++ b/ui/scripts/vpc.js
@@ -1325,7 +1325,7 @@
 
                                                     if(items){
                                                         items.sort(function(a, b) {
-                                                            return a.number >= b.number;
+                                                            return a.number - b.number;
                                                         }).map(function(acl) {
                                                             if (parseInt(acl.protocol)) { // protocol number
                                                                 acl.protocolnumber = acl.protocol;


rafaelweingartner:
Thanks @DaanHoogland

borisstoyanov:
@rafaelweingartner can you upload some screenshots? Since it's a UI change I don't see point of running integration tests.

rafaelweingartner:
I am not adding/removing anything in the UI. I am neither changing/replacing anything visible. I am only changing a sort function that was not following the standards, and that is why it was not working in chrome.

I mean, the screenshot will not prove much. It is a static image. I can provide you with two images, one with the problem and other with the problem solved if you really think they are important.

borisstoyanov:
Thanks @rafaelweingartner, please show us what was the issue and how the fix looks. Since we don't require any test results on this, let us have some justification as screenshots please.

rafaelweingartner:
Ok. there you go.
Before:
before

After:
after

rafaelweingartner:
What happens there is that Google Chrome only sorts blocks of 10 elements when the sort function return Boolean values. To fix it, we need to return a numeric value indicating if the first object is smaller (-), equal (0), or (+) greater than the second one.

PR #2479, opened 2018-03-07
CLOUDSTACK-10278: Introduce Flyway migration on CloudStack database

ustcweizhou:
@rhtyd @DaanHoogland @rafaelweingartner @blueorangutan


engine/schema/pom.xml
```diff
@@ -48,5 +48,10 @@
       <artifactId>cloud-framework-db</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.flywaydb</groupId>
+      <artifactId>flyway-core</artifactId>
+      <version>5.0.7</version>
```
rhtyd: @ustcweizhou can you move the version to root level pom.

engine/schema/src/main/java/com/cloud/upgrade/DatabaseUpgradeChecker.java
```diff
@@ -631,6 +635,27 @@ public void check() {
...
             lock.releaseRef();
         }
+
+        s_logger.info("Running Flyway migration on Cloudstack database");
+        Properties dbProps = DbProperties.getDbProperties();
+        final String cloudUsername = dbProps.getProperty("db.cloud.username");
+        final String cloudPassword = dbProps.getProperty("db.cloud.password");
+        final String cloudHost = dbProps.getProperty("db.cloud.host");
+        final int cloudPort = Integer.parseInt(dbProps.getProperty("db.cloud.port"));
+        final String dbUrl = "jdbc:mysql://" + cloudHost + ":" + cloudPort + "/cloud";
```
rhtyd: @ustcweizhou we also have cloud_usage database, how do we handle changes for that?

ustcweizhou: @rhtyd this is a good question. To be honest, I never considered it.
Actually not only me, all sql files in engine/schema/resources/META-INF/db/ do not consider it...

DaanHoogland:
Good work Wei, Can you add a link to prereq reading and maybe a short usage description?

ustcweizhou:
flyway get started
https://flywaydb.org/getstarted/how

DaanHoogland:
@ustcweizhou I don't think that is going to propagate the use and understanding of this code a lot, but soit. Let's grow it over time.

rhtyd:
Nice, looks like many more changes. @ustcweizhou do you intend to get cloudstack completely move to flyway in next major/lts 4.13 release? If so, please start a discussion on dev list, I would like to engage and discuss both upgrade and migration paths from old system to new system. cc @PaulAngus

rhtyd:
cc @nvazquez @PaulAngus @DagSonsteboSB

ustcweizhou:
@rhtyd although I merged @fmaximus 's changes to my repo, I think it would be better to keep the upgrade procedure before 4.13 (fresh install/older version to 4.12), and use flyway migration from 4.12 to 4.13 upgrade.

rhtyd:
Let's revisit this in the next major release, 4.14+

weizhouapache:
@rhtyd @DaanHoogland @PaulAngus 
can we restart the discussion on flyway migration after 4.14.0.0 release ?

DaanHoogland:
@weizhouapache I would be very pleased. It is one of those technical debt projects that keep lying around.
2. gson
3. log4j
4. ...

let's list them in the issues labelled technical-debt-project and see what we can move on.

engine/schema/src/main/java/com/cloud/upgrade/DatabaseUpgradeChecker.java
```diff
@@ -631,6 +635,27 @@ public void check() {
...
+
+        try {
+            Flyway flyway = new Flyway();
+            flyway.setDataSource(dbUrl, cloudUsername, cloudPassword);
+            flyway.setTable("cloudstack_schema_version");
+            flyway.setBaselineOnMigrate(true);
+            flyway.setBaselineVersionAsString("4.11");
```
weizhouapache: To be tested: 
will SQL starts with `4.12` or `4.20` work ?

PR #2481, opened 2018-03-08
CLOUDSTACK-10320 - Invalid pair for response object breaking response parsing

marcaurele:
Under some circumstances, the API is returning an invalid response, for simplicity I will expose the JSON case. The API response on a `listVirtualMachines` can be this string:
```
{ "listvirtualmachinesresponse" :  ] } }
```

To understand how this is possible, assume you have more than one management server and one is processing the destroy of a virtual machine in the account X which is the only one it has. Another process is returning the result of `listVirtualMachines` for that same account X. During the listVM command, the result set is fetch with a `searchAndDistinctCount` due to the view (https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java#L1024). This is done through 2 queries in the GenericDao https://github.com/apache/cloudstack/blob/master/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java#L1323 and if you encounter the right conditions, the VM will be marked as removed in between those 2 queries. This results in having a Pair result with at least one object but a count of 0. Then following how is done the serialization of the response at https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/api/response/ApiResponseSerializer.java#L86 you will reach the case where your output is the one previously mentioned.

To overcome this issue, there isn't a true fix but only a better pair response to ensure a correct response formatting. If the result set contains at least something, the count cannot be 0 but we cannot guess the correct answer, but only state it has at least one element.

### DIFF
--- a/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
+++ b/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
@@ -1315,6 +1315,10 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
     public Pair<List<T>, Integer> searchAndCount(final SearchCriteria<T> sc, final Filter filter) {
         List<T> objects = search(sc, filter, null, false);
         Integer count = getCount(sc);
+        // Count cannot be less than the result set but can be higher due to pagination, see CLOUDSTACK-10320
+        if (count < objects.size()) {
+            count = objects.size();
+        }
         return new Pair<List<T>, Integer>(objects, count);
     }
 
@@ -1323,6 +1327,11 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
     public Pair<List<T>, Integer> searchAndDistinctCount(final SearchCriteria<T> sc, final Filter filter) {
         List<T> objects = search(sc, filter, null, false);
         Integer count = getDistinctCount(sc);
+        // Count cannot be 0 if there is at least a result in the list, see CLOUDSTACK-10320
+        if (count == 0 && !objects.isEmpty()) {
+            // Cannot assume if it's more than one since the count is distinct vs search
+            count = 1;
+        }
         return new Pair<List<T>, Integer>(objects, count);
     }
 
@@ -1331,6 +1340,11 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
     public Pair<List<T>, Integer> searchAndDistinctCount(final SearchCriteria<T> sc, final Filter filter, final String[] distinctColumns) {
         List<T> objects = search(sc, filter, null, false);
         Integer count = getDistinctCount(sc, distinctColumns);
+        // Count cannot be 0 if there is at least a result in the list, see CLOUDSTACK-10320
+        if (count == 0 && !objects.isEmpty()) {
+            // Cannot assume if it's more than one since the count is distinct vs search
+            count = 1;
+        }
         return new Pair<List<T>, Integer>(objects, count);
     }
 


framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1323,6 +1327,11 @@ protected void addJoins(StringBuilder str, Collection<JoinBuilder<SearchCriteria
     public Pair<List<T>, Integer> searchAndDistinctCount(final SearchCriteria<T> sc, final Filter filter) {
         List<T> objects = search(sc, filter, null, false);
         Integer count = getDistinctCount(sc);
+        // Count cannot be 0 if there is at least a result in the list, see CLOUDSTACK-10320
+        if (count == 0 && !objects.isEmpty()) {
```
rafaelweingartner: Is it possible for that `objects` variable there to assume a null value? If not, everything seems to be in order here. Otherwise, it might be better to use CollectionUtils.

marcaurele: `search` always return a list, never `null`.

DaanHoogland:
nice find, dirty hack, though maybe justified. Isn't the proper solution to put both queries in a transaction, @marcaurele ?

rafaelweingartner:
@blueorangutan package

DaanHoogland:
@blueorangutan test

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1323,6 +1327,11 @@ protected void addJoins(StringBuilder str, Collection<JoinBuilder<SearchCriteria
...
+            // Cannot assume if it's more than one since the count is distinct vs search
+            count = 1;
```
rhtyd: why not put count = objects.size() ?

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1331,6 +1340,11 @@ protected void addJoins(StringBuilder str, Collection<JoinBuilder<SearchCriteria
     public Pair<List<T>, Integer> searchAndDistinctCount(final SearchCriteria<T> sc, final Filter filter, final String[] distinctColumns) {
         List<T> objects = search(sc, filter, null, false);
         Integer count = getDistinctCount(sc, distinctColumns);
+        // Count cannot be 0 if there is at least a result in the list, see CLOUDSTACK-10320
+        if (count == 0 && !objects.isEmpty()) {
+            // Cannot assume if it's more than one since the count is distinct vs search
```
rhtyd: Why not put count = objects.size() ?

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1315,6 +1315,10 @@ protected void addJoins(StringBuilder str, Collection<JoinBuilder<SearchCriteria
     public Pair<List<T>, Integer> searchAndCount(final SearchCriteria<T> sc, final Filter filter) {
         List<T> objects = search(sc, filter, null, false);
         Integer count = getCount(sc);
+        // Count cannot be less than the result set but can be higher due to pagination, see CLOUDSTACK-10320
+        if (count < objects.size()) {
```
rhtyd: This logic may be used in changes below.

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
marcaurele: Because the search is mostly done against views which don't return distinct rows due to the *not so smart* queries fetching the details rows. Look at the APIHelper with iterate on duplicate row entries, but only append the details.

marcaurele: Moreover, the search might return a page size list, while the count should be global: listVirtualMachines returns a count of 234 and you have 50 entries in the search based on the `page` and `pageSize`.

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
marcaurele: See comment before

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
marcaurele: I will change how it's handled

marcaurele:
@DaanHoogland you're right and I looked again on how to achieve this isolation between queries, with a `repeatable-read` but so far I didn't manage to have it correctly working. I read that for innodb, the default isolation in a transaction is the `repeatable-read` but my local debugging tests did not work as expected. I will dig more to find a better solution and not a hacky one.

marcaurele:
The default isolation level in CS is `read-committed` (https://github.com/apache/cloudstack/blob/master/framework/db/src/main/java/com/cloud/utils/db/TransactionLegacy.java#L1044) but not `repeatable-read`, except if you explicitly set it in the `db.properties` file. If the default value is change, and I modify the PR to use a transaction around those 2 SQL read operations, the issue is fixed, but...

Changing the default isolation level will certainly have an impact on performance on the database (https://www.percona.com/blog/2015/01/14/mysql-performance-implications-of-innodb-isolation-modes/). So, I'll change the PR to only have a repeatable-read isolation for those set of methods, and will try to measure how it performs.

If people are willing to test in their environment (quite loaded) the global isolation level change, it would be interesting to hear the performance outcomes.

DaanHoogland:
Good experiment @marcaurele , I have no stress test env at my disposal so i can not verify. I am worried that my question is blocking your fix. Let's try to separate the fix and the improvement.
Is the transaction failing as a solution to the bug if the isolation level isn't changed as well?
* If it is, let's merge as is.
* If it isn't (but performing badly) let's merge it as transaction.
In both cases we can create a ticket to track it further, ok?

marcaurele:
I don't think we're in a hurry to fix the problem. I will investigate on our platform and maybe try to narrow down the isolation level to those queries only. It will take one or two weeks to get the result. Anyway, I don't see another option to fix the problem with the same requirements (total count, pagination).
The other solution would be to get rid of the count value in the API response, making this change obsolete. It should be a point to keep in mind for the next API version.

marcaurele:
@DaanHoogland bad news: it didn't last very long before I had to rollback the isolation level due to DB dead locks (at least in our branch based from 4.4.2). So it's the only quick fix & hacky solution for now which doesn't add another issue.

PR #2482, opened 2018-03-08
CLOUDSTACK-10321: CPU Cap for KVM

nvazquez:
JIRA Ticket: https://issues.apache.org/jira/browse/CLOUDSTACK-10321

## Context
Libvirt has the concept of a 'period' and a 'quota'. If the period is 10000 microseconds, a quote can be set, say 5000 and the VM CPU will be limited to only using 5000 microseconds of CPU every 10000 microseconds. 
This patch basically looks at the host CPU MHz (say 2000 MHz), looks at the VM's configured MHz (say 500 MHz), comes up with a percent (25%), then uses that against the period to come up with a quota number (2500). Additionally, it only applies appropriately when the service offering has the CPU cap option set/checkmarked.

## Solution
These changes are restricted to the KVM hypervisor plugin, where creation of a VM (spec) will consider CPU cap (max speed) defined in the service offering to define and set quota/limits in the VM’s spec. This will set quota based on a fixed/pre-defined period value, and add respective flag/nodes in the VM/domain XML as documented here: https://libvirt.org/formatdomain.html#elementsCPUTuning 

## Behavior
- Setting the CPU cap option on a service offering must cause the VM created with that CPU offering, to run at the MHz value set until the core it is running on becomes contended (under KVM)
- If the CPU cap option is set in a service offering and the core that a VM is running on becomes contended, then (under KVM) the VM should receive the appropriate MHz according to its and the other VMs MHz values translated as 'shares' values.  (this is the current behaviour for contended and uncontended cores).

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -1993,12 +1993,19 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             int period = CpuTuneDef.DEFAULT_PERIOD;
             int quota = (int) (period * cpuQuotaPercentage);
             if (quota < CpuTuneDef.MIN_QUOTA) {
+                s_logger.info("Calculated quota (" + quota + ") below the minimum (" + CpuTuneDef.MIN_QUOTA + ") for VM domain " + vmTO.getUuid() + ", setting it to minimum " +
+                        "and calculating period instead of using the default");
                 quota = CpuTuneDef.MIN_QUOTA;
                 period = (int) ((double) quota / cpuQuotaPercentage);
+                if (period > CpuTuneDef.MAX_PERIOD) {
+                    s_logger.info("Calculated period (" + period + ") exceeds the maximum (" + CpuTuneDef.MAX_PERIOD +
+                            "), setting it to the maximum");
+                    period = CpuTuneDef.MAX_PERIOD;
+                }
             }
             ctd.setQuota(quota);
             ctd.setPeriod(period);
-            s_logger.info("Setting quota=" + quota + ", period=" + period + " to VM domain " + vmTO.getId());
+            s_logger.info("Setting quota=" + quota + ", period=" + period + " to VM domain " + vmTO.getUuid());
         }
     }
 
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -1984,6 +1984,24 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
         return uuid;
     }
 
+    /**
+     * Set quota and period tags on 'ctd' when CPU limit use is set
+     */
+    protected void setQuotaAndPeriod(VirtualMachineTO vmTO, CpuTuneDef ctd) {
+        if (vmTO.getLimitCpuUse() && vmTO.getCpuQuotaPercentage() != null) {
+            Double cpuQuotaPercentage = vmTO.getCpuQuotaPercentage();
+            int period = CpuTuneDef.DEFAULT_PERIOD;
+            int quota = (int) (period * cpuQuotaPercentage);
+            if (quota < CpuTuneDef.MIN_QUOTA) {
+                quota = CpuTuneDef.MIN_QUOTA;
+                period = (int) ((double) quota / cpuQuotaPercentage);
+            }
+            ctd.setQuota(quota);
+            ctd.setPeriod(period);
+            s_logger.info("Setting quota=" + quota + ", period=" + period + " to VM domain " + vmTO.getId());
+        }
+    }
+
     public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
         final LibvirtVMDef vm = new LibvirtVMDef();
         vm.setDomainName(vmTO.getName());
@@ -2059,6 +2077,9 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             } else {
                 ctd.setShares(vmTO.getCpus() * vmTO.getSpeed());
             }
+
+            setQuotaAndPeriod(vmTO, ctd);
+
             vm.addComp(ctd);
         }
 
--- a/server/src/com/cloud/hypervisor/KVMGuru.java
+++ b/server/src/com/cloud/hypervisor/KVMGuru.java
@@ -82,17 +82,21 @@ public class KVMGuru extends HypervisorGuruBase implements HypervisorGuru {
             if (host == null) {
                 throw new CloudRuntimeException("Host with id: " + vm.getHostId() + " not found");
             }
-            s_logger.debug("Limiting CPU usage for VM " + vm.getId() + " on host " + host.getId());
+            s_logger.debug("Limiting CPU usage for VM: " + vm.getUuid() + " on host: " + host.getUuid());
             double hostMaxSpeed = getHostCPUSpeed(host);
             double maxSpeed = getVmSpeed(to);
             try {
                 BigDecimal percent = new BigDecimal(maxSpeed / hostMaxSpeed);
                 percent = percent.setScale(2, RoundingMode.HALF_DOWN);
+                if (percent.compareTo(new BigDecimal(1)) == 1) {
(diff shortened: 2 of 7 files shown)


borisstoyanov:
@blueorangutan package

wido:
Nice PR! Shouldn't the XML parser for the VMDef however not be updated as well and have additional tests?

nvazquez:
Thanks @wido! Do you mean the toString() method for CpuTuneDef that should not be updated? Can you propose any additional test?

@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@borisstoyanov failing test passes on my local environment:

````
nicolas@nicolas-M4700:~/cloudstack$ cat /tmp//MarvinLogs/test_service_offerings_MOB691/results.txt 
Test CPU Cap on KVM ... === TestName: test_01_service_offering_cpu_limit_use | Status : SUCCESS ===
ok
Test to create service offering ... === TestName: test_01_create_service_offering | Status : SUCCESS ===
ok
Test to update existing service offering ... === TestName: test_02_edit_service_offering | Status : SUCCESS ===
ok
Test to delete service offering ... === TestName: test_03_delete_service_offering | Status : SUCCESS ===
ok
Test to change service to a small capacity ... === TestName: test_04_change_offering_small | Status : SUCCESS ===
ok

Ran 5 tests in 887.243s

OK
````
New test uses information on 'host_details' table to ssh into the KVM host running the VM. From BlueOrangutan logs we can see that ssh cannot be performed:
````
2018-03-10 22:03:33,761 - CRITICAL - EXCEPTION: test_01_service_offering_cpu_limit_use: ['Traceback (most recent call last):\n', '  File "/usr/lib64/python2.7/unittest/case.py", line 369, in run\n    testMethod()\n', '  File "/usr/lib/python2.7/site-packages/marvin/lib/decoratorGenerators.py", line 30, in test_wrapper\n    return test(self, *args, **kwargs)\n', '  File "/marvin/tests/smoke/test_service_offerings.py", line 549, in test_01_service_offering_cpu_limit_use\n    ssh_host = self.get_ssh_client(host_id, host_ip_address, host_username, host_password, 10)\n', '  File "/marvin/tests/smoke/test_service_offerings.py", line 437, in get_ssh_client\n    self.fail("Unable to create ssh connection: " % e)\n', 'TypeError: not all arguments converted during string formatting\n']
````

**Comment by rhtyd** [comment]
@nvazquez `test_01_service_offering_cpu_limit_use` looks like a regression, can you check?

File: server/src/com/cloud/hypervisor/KVMGuru.java [hunk 1]
```diff
@@ -52,10 +59,49 @@ protected KVMGuru() {
...
+    * Set VM CPU quota percentage with respect to host CPU on 'to' if CPU limit option is set
+    * @param to vm to
+    * @param vmProfile vm profile
+    */
+    protected void setVmQuotaPercentage(VirtualMachineTO to, VirtualMachineProfile vmProfile) {
+        if (to.getLimitCpuUse()) {
+            VirtualMachine vm = vmProfile.getVirtualMachine();
+            HostVO host = _hostDao.findById(vm.getHostId());
+            if (host == null) {
+                throw new CloudRuntimeException("Host with id: " + vm.getHostId() + " not found");
```
rafaelweingartner: instead of ID, what about UUID here?
Normally the operator is used to the UUID as the ID, and not the internal db ID.

server/src/com/cloud/hypervisor/KVMGuru.java
```diff
@@ -52,10 +59,49 @@ protected KVMGuru() {
...
+            }
+            s_logger.debug("Limiting CPU usage for VM " + vm.getId() + " on host " + host.getId());
```
rafaelweingartner: Same here, for both VMs and hosts

server/src/com/cloud/hypervisor/KVMGuru.java
```diff
@@ -52,10 +59,49 @@ protected KVMGuru() {
...
+            double hostMaxSpeed = getHostCPUSpeed(host);
+            double maxSpeed = getVmSpeed(to);
+            try {
+                BigDecimal percent = new BigDecimal(maxSpeed / hostMaxSpeed);
+                percent = percent.setScale(2, RoundingMode.HALF_DOWN);
+                to.setCpuQuotaPercentage(percent.doubleValue());
+                s_logger.debug("Host " + host.getId() + " max CPU speed = " + hostMaxSpeed + "MHz, VM " + vm.getId() +
```
rafaelweingartner: Here as well.

server/src/com/cloud/hypervisor/KVMGuru.java
```diff
@@ -52,10 +59,49 @@ protected KVMGuru() {
...
+                        "max CPU speed = " + maxSpeed + "MHz. Setting CPU quota percentage as: " + percent.doubleValue());
+            } catch (NumberFormatException e) {
+                s_logger.error("Error calculating vm quota percentage, it wll not be set: " + e.getMessage());
```
rafaelweingartner: Is this CPU quota limit set by VM?
What about logging the VM UUID that caused this error here?

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -1984,6 +1984,24 @@ protected String getUuid(String uuid) {
...
+    /**
+     * Set quota and period tags on 'ctd' when CPU limit use is set
+     */
+    protected void setQuotaAndPeriod(VirtualMachineTO vmTO, CpuTuneDef ctd) {
+        if (vmTO.getLimitCpuUse() && vmTO.getCpuQuotaPercentage() != null) {
+            Double cpuQuotaPercentage = vmTO.getCpuQuotaPercentage();
+            int period = CpuTuneDef.DEFAULT_PERIOD;
+            int quota = (int) (period * cpuQuotaPercentage);
+            if (quota < CpuTuneDef.MIN_QUOTA) {
+                quota = CpuTuneDef.MIN_QUOTA;
```
rafaelweingartner: what about some logging and documentation explaining why and when this happen?

The logginf to show this change is useful for operators to understand why the value they configure is not being used.

server/src/com/cloud/hypervisor/KVMGuru.java
```diff
@@ -52,10 +59,49 @@ protected KVMGuru() {
...
+            if (host == null) {
+                throw new CloudRuntimeException("Host with id: " + vm.getHostId() + " not found");
+            }
+            s_logger.debug("Limiting CPU usage for VM " + vm.getId() + " on host " + host.getId());
+            double hostMaxSpeed = getHostCPUSpeed(host);
+            double maxSpeed = getVmSpeed(to);
+            try {
+                BigDecimal percent = new BigDecimal(maxSpeed / hostMaxSpeed);
+                percent = percent.setScale(2, RoundingMode.HALF_DOWN);
+                to.setCpuQuotaPercentage(percent.doubleValue());
```
rafaelweingartner: Is the maximum speed of the VM limited to the maximum speed of a core in the host? Otherwise, we can have values bigger than 100% here.

server/src/com/cloud/hypervisor/KVMGuru.java
nvazquez: Agreed on the others below, but actually we won't have the host UUID at this point, only the ID reference on the VM. One way to get it could be to search host including the removed ones. Do you think we should do that?

server/src/com/cloud/hypervisor/KVMGuru.java
nvazquez: Done, thanks

server/src/com/cloud/hypervisor/KVMGuru.java
nvazquez: Done, thanks

server/src/com/cloud/hypervisor/KVMGuru.java
nvazquez: Done, thanks

server/src/com/cloud/hypervisor/KVMGuru.java
nvazquez: Solved, thanks

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

server/src/com/cloud/hypervisor/KVMGuru.java
nvazquez: No problem. Thanks @rafaelweingartner for your review!

borisstoyanov:
@blueorangutan test

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

DaanHoogland:
@rafaelweingartner is this lgty?

rafaelweingartner:
Yes!

PR #2485, opened 2018-03-12
Bump the version of Debian net-installer to 9.4.0

khos2ow:
The new Debian 9.4.0 just released (March 10th).

### DIFF
--- a/tools/appliance/systemvmtemplate/template.json
+++ b/tools/appliance/systemvmtemplate/template.json
@@ -38,8 +38,8 @@
       "disk_interface": "virtio",
       "net_device": "virtio-net",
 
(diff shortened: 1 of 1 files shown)


khos2ow:
@rhtyd @rafaelweingartner @DaanHoogland @wido unless we still want to remain on 9.3.0, which needs to be pointed to one of the mirrors, such as [this](http://sft.if.usp.br/debian-cd/9.3.0/)?

rhtyd:
Built and tested locally. For 4.11, we're not going to change the systemvmtemplate for just this change. However, there are some PRs from @resmo and others that may require that. I'll fire tests separately.

PR #2477, opened 2018-03-05
Smaller systemvm images using sparse disks.

PaulAngus:

PR #2488, opened 2018-03-15
smaller systemvm images using sparse disks.

rhtyd:
This is from @PaulAngus 's PR https://github.com/apache/cloudstack/pull/2477 towards 4.11.

### DIFF
--- a/tools/appliance/build.sh
+++ b/tools/appliance/build.sh
@@ -271,9 +271,9 @@ function xen_server_export() {
   local result=$?
   set -e
   if [ ${result} == 0 ]; then
-    qemu-img convert -f qcow2 -O raw "dist/${appliance}" img.raw
-    vhd-util convert -s 0 -t 1 -i img.raw -o stagefixed.vhd
-    faketime '2010-01-01' vhd-util convert -s 1 -t 2 -i stagefixed.vhd -o "${appliance_build_name}-xen.vhd"
+    cp "dist/${appliance}" "dist/${appliance}-xen"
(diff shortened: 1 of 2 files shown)


tools/appliance/build.sh
```diff
@@ -271,9 +271,9 @@ function xen_server_export() {
   local result=$?
   set -e
   if [ ${result} == 0 ]; then
-    qemu-img convert -f qcow2 -O raw "dist/${appliance}" img.raw
-    vhd-util convert -s 0 -t 1 -i img.raw -o stagefixed.vhd
-    faketime '2010-01-01' vhd-util convert -s 1 -t 2 -i stagefixed.vhd -o "${appliance_build_name}-xen.vhd"
+    cp "dist/${appliance}" "dist/${appliance}-xen"
+    vhd-util convert -s 0 -t 1 -i "dist/${appliance}-xen" -o stagefixed.vhd
+    faketime '2010-01-01' qemu-img convert -S 512 -O vpc -o subformat=dynamic stagefixed.vhd "${appliance_build_name}-xen.vhd"
```
rhtyd: @PaulAngus can you confirm if this could break template on XenServer, as qemu-img is used to convert the staged vhd file to `vpc` output format that corresponds to HyperV vhd?

rhtyd:
@PaulAngus looks like there is no solution for now available with linux tools to export a compressed vhd file. A slightly semi-automated way could be to export template using xenserver windows/export tools.

As commented, closing this PR as the exported vhd is of `vpc` type that is compatible with HyperV but not XenServer.

PR #2496, opened 2018-03-20
[CLOUDSTACK-10332] Users are not able to change/edit the protocol of an ACL rule 

rafaelweingartner:
## Description

Users should be able to edit an ACL rule completely. Therefore, they must be able to change the protocol type and others configs of an ACL rules.

In addition, it is not just the protocol that users are not able to change. For instance, after defining ports, or reason/description for the rule, users are not able to set those values back to null. The same happens for ICMP code and type.

We will introduce a new parameter called "partialUpdate", which will have its default value as true to maintain backward compatibility. When this parameter is set to false, we will consider only the parameters sent, and not the parameters we already have in the database to change and validate the ACL rule data. This allows us to update parameters already set back to null, and to completely change an ACL rule.

https://issues.apache.org/jira/browse/CLOUDSTACK-10332

* Create an ACL for ICMP
* Click on edit and change the protocol to TCP
* An error will happen when saving the rule.

## Types of changes

## How Has This Been Tested?

Compile, and deployed ACS with the proposed changes and verified them manually. Moreover, I have unit tests covering this issue, which are able to catch the problem.

used Unit testing and a test ACS environment running with XenServer 6.5, and NFS server as primary and secondary storage.

## Checklist:

@blueorangutan package

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
+++ b/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
@@ -309,7 +309,7 @@ public class ApiConstants {
     public static final String USERNAME = "username";
     public static final String USER_SECURITY_GROUP_LIST = "usersecuritygrouplist";
     public static final String USE_VIRTUAL_NETWORK = "usevirtualnetwork";
-    public static final String Update_IN_SEQUENCE ="updateinsequence";
+    public static final String Update_IN_SEQUENCE = "updateinsequence";
     public static final String VALUE = "value";
     public static final String VIRTUAL_MACHINE_ID = "virtualmachineid";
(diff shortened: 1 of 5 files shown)


borisstoyanov:
@blueorangutan test

rafaelweingartner:
@blueorangutan package

borisstoyanov:
@blueorangutan test

rafaelweingartner:
Now the errors changed. Are they intermittent ones?

borisstoyanov:
@rafaelweingartner we've seen these errors with other PRs, I think they are not related

rafaelweingartner:
@borisstoyanov thanks!
So, are you ok with this PR?

PR #2498, opened 2018-03-21
CLOUDSTACK-10327: Do not invalidate the session when API command not found

olivierlemasle:
## Description

CloudStack SSO (using `security.singlesignon.key`) does not work anymore with CloudStack 4.11, since commit 9988c26, which introduced a regression due to a refactoring: every API request that is not "validated" generates the same error (401 - Unauthorized) and invalidates the session.

However, CloudStack UI executes a call to `listConfigurations` [in method `bypassLoginCheck`](https://github.com/apache/cloudstack/blob/1c99fd73881938/ui/scripts/cloudStack.js#L172). A non-admin user does not have the permissions to execute this request, which causes an error 401:

```
{"listconfigurationsresponse":{"uuidList":[],"errorcode":401,"errortext":"unable to verify user credentials and/or request signature"}}
```

The session (already created by SSO) is then invalidated and the user cannot access to CloudStack UI (error "Session Expired").

Before 9988c26 (up to CloudStack 4.10), an error 432 was returned (and ignored):

```
{"errorresponse":{"uuidList":[],"errorcode":432,"cserrorcode":9999,"errortext":"The user is not allowed to request the API command or the API command does not exist"}}
```

Even if the call to `listConfigurations` was removed, another call to [`listIdps`](https://github.com/apache/cloudstack/blob/1c99fd73881938/ui/scripts/cloudStack.js#L192) also lead to an error 401 for user accounts if the SAML plugin is not enabled.

This pull request aims to fix the SSO issue, by restoring errors 432 (instead of 401 + invalidate session) for commands not available. However, if an API command is explicitly denied using ACLs or if the session key is incorrect, it still generates an error 401 and invalidates the session.

## Types of changes

## How Has This Been Tested?

Compiled and ran CloudStack.

- API commands not explicitly allowed for a role (e.g. `listConfigurations` for a user) generate 432.
- API commands not enabled (e.g. `listIdps` with SAML disabled) generate 432.
- non-existing API commands (e.g. `fooBar`) generate 432

However, API calls with session key removed generate 401 and invalidate session.
SSO is ok.

## Checklist:

### DIFF
--- a/server/src/com/cloud/api/ApiServer.java
+++ b/server/src/com/cloud/api/ApiServer.java
@@ -34,6 +34,7 @@ import com.cloud.exception.PermissionDeniedException;
 import com.cloud.exception.RequestLimitException;
 import com.cloud.exception.ResourceAllocationException;
 import com.cloud.exception.ResourceUnavailableException;
+import com.cloud.exception.UnavailableCommandException;
 import com.cloud.user.Account;
 import com.cloud.user.AccountManager;
 import com.cloud.user.DomainManager;
@@ -958,6 +959,9 @@ public class ApiServer extends ManagerBase implements HttpRequestHandler, ApiSer
         } catch (final RequestLimitException ex) {
             s_logger.debug(ex.getMessage());
             throw new ServerApiException(ApiErrorCode.API_LIMIT_EXCEED, ex.getMessage());
+        }  catch (final UnavailableCommandException ex) {
+            s_logger.debug(ex.getMessage());
+            throw new ServerApiException(ApiErrorCode.UNSUPPORTED_ACTION_ERROR, ex.getMessage());
         } catch (final PermissionDeniedException ex) {
             final String errorMessage = "The given command '" + commandName + "' either does not exist, is not available" +
                     " for user, or not available from ip address '" + remoteAddress + "'.";
--- a/api/src/main/java/com/cloud/exception/UnavailableCommandException.java
+++ b/api/src/main/java/com/cloud/exception/UnavailableCommandException.java
@@ -1,0 +1,36 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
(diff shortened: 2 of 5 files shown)


rhtyd:
@olivierlemasle can you edit the PR and rebase against 4.11 branch?

olivierlemasle:
@rhtyd Ok, I've rebased the PR against 4.11 and updated the base branch accordingly.

server/src/com/cloud/api/ApiServer.java
```diff
@@ -958,6 +959,9 @@ private boolean commandAvailable(final InetAddress remoteAddress, final String c
         } catch (final RequestLimitException ex) {
             s_logger.debug(ex.getMessage());
             throw new ServerApiException(ApiErrorCode.API_LIMIT_EXCEED, ex.getMessage());
+        }  catch (final UnavailableCommandException ex) {
+            s_logger.debug(ex.getMessage());
+            throw new ServerApiException(ApiErrorCode.UNSUPPORTED_ACTION_ERROR, ex.getMessage());
```
rhtyd: Why not throw ServerApiException with suitable http error code so as to not cause log out in the UI. Simply handle permission denied exception here? (or the line below)

rhtyd:
Thanks @olivierlemasle. I've one comment, since the last change around this was made by Wido, let's ask @wido for review.

wido:
I don't see anything odd in this PR, looks good to me. I didn't catch this situation.

rhtyd:
@blueorangutan package

server/src/com/cloud/api/ApiServer.java
olivierlemasle: @rhtyd I'm not sure I understand your suggestion. The Http error code is specified here, in line 964). The `ServerApiException` is caught on `ApiServlet` which generates the HTTP response.

borisstoyanov:
@blueorangutan test

borisstoyanov:
@blueorangutan test

rafaelweingartner:
@borisstoyanov here we also have that same `test_04_restart_network_wo_cleanup` failure.

rhtyd:
Merged based on test results and code review.

PR #2503, opened 2018-03-21
Support multiple volume access groups per compute cluster

mike-tutkowski:
## Description
Previously, the SolidFire storage plug-in for managed storage had a 1:1 mapping between a compute cluster and a volume access group (VAG). A VAG is a type of ACL in the SolidFire cluster.

This PR expands that support to be 1:M.

## Types of changes

## How Has This Been Tested?
I confirmed after the code changes that you can now have multiple volume access groups support a single compute cluster. Also, you can still have a single volume access group support a compute cluster.

## Checklist:
- [ ] All new and existing tests passed.

@blueorangutan package

### DIFF
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
@@ -1654,18 +1654,18 @@ public class StorageSystemDataMotionStrategy implements DataMotionStrategy {
         details.put(ModifyTargetsCommand.STORAGE_HOST, storagePool.getHostAddress());
         details.put(ModifyTargetsCommand.STORAGE_PORT, String.valueOf(storagePool.getPort()));
 
-        ModifyTargetsCommand modifyTargetsCommand = new ModifyTargetsCommand();
+        ModifyTargetsCommand cmd = new ModifyTargetsCommand();
 
         List<Map<String, String>> targets = new ArrayList<>();
 
         targets.add(details);
 
-        modifyTargetsCommand.setTargets(targets);
-        modifyTargetsCommand.setApplyToAllHostsInCluster(true);
-        modifyTargetsCommand.setAdd(add);
-        modifyTargetsCommand.setTargetTypeToRemove(ModifyTargetsCommand.TargetTypeToRemove.DYNAMIC);
+        cmd.setTargets(targets);
+        cmd.setApplyToAllHostsInCluster(true);
+        cmd.setAdd(add);
+        cmd.setTargetTypeToRemove(ModifyTargetsCommand.TargetTypeToRemove.DYNAMIC);
 
-        return modifyTargetsCommand;
+        return cmd;
     }
 
     private List<String> sendModifyTargetsCommand(ModifyTargetsCommand cmd, long hostId) {
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
@@ -145,6 +145,7 @@ public class VmwareStorageProcessor implements StorageProcessor {
 
     private static final Logger s_logger = Logger.getLogger(VmwareStorageProcessor.class);
     private static final int DEFAULT_NFS_PORT = 2049;
+    private static final int SECONDS_TO_WAIT_FOR_DATASTORE = 120;
 
     private final VmwareHostService hostService;
     private boolean _fullCloneFlag;
@@ -2602,8 +2603,7 @@ public class VmwareStorageProcessor implements StorageProcessor {
     }
 
     private void waitForAllHostsToSeeDatastore(List<Pair<ManagedObjectReference, String>> lstHosts, DatastoreMO dsMO) throws Exception {
-        long secondsToWait = 120;
-        long endWaitTime = System.currentTimeMillis() + secondsToWait * 1000;
+        long endWaitTime = System.currentTimeMillis() + SECONDS_TO_WAIT_FOR_DATASTORE * 1000;
 
         boolean isConditionMet = false;
 
@@ -2621,7 +2621,7 @@ public class VmwareStorageProcessor implements StorageProcessor {
     private boolean verifyAllHostsSeeDatastore(List<Pair<ManagedObjectReference, String>> lstHosts, DatastoreMO dsMO) throws Exception {
         int numHostsChecked = 0;
 
-        for (Pair<ManagedObjectReference, String> host: lstHosts) {
+        for (Pair<ManagedObjectReference, String> host : lstHosts) {
             ManagedObjectReference morHostToMatch = host.first();
             HostMO hostToMatchMO = new HostMO(dsMO.getContext(), morHostToMatch);
 
@@ -2641,8 +2641,7 @@ public class VmwareStorageProcessor implements StorageProcessor {
     }
 
     private void waitForAllHostsToMountDatastore(List<Pair<ManagedObjectReference, String>> lstHosts, DatastoreMO dsMO) throws Exception {
-        long secondsToWait = 120;
-        long endWaitTime = System.currentTimeMillis() + secondsToWait * 1000;
+        long endWaitTime = System.currentTimeMillis() + SECONDS_TO_WAIT_FOR_DATASTORE * 1000;
 
         boolean isConditionMet = false;
 
@@ -2657,13 +2656,39 @@ public class VmwareStorageProcessor implements StorageProcessor {
         }
     }
 
-    private boolean verifyAllHostsMountedDatastore(List<Pair<ManagedObjectReference, String>> lstHosts, DatastoreMO dsMO) throws Exception {
-        int numHostsChecked = 0;
+    private void waitForAllHostsToMountDatastore2(List<HostMO> lstHosts, DatastoreMO dsMO) throws Exception {
+        long endWaitTime = System.currentTimeMillis() + SECONDS_TO_WAIT_FOR_DATASTORE * 1000;
 
-        for (Pair<ManagedObjectReference, String> host: lstHosts) {
+        boolean isConditionMet = false;
+
+        while (System.currentTimeMillis() < endWaitTime && !isConditionMet) {
+            Thread.sleep(5000);
+
+            isConditionMet = verifyAllHostsMountedDatastore2(lstHosts, dsMO);
+        }
+
+        if (!isConditionMet) {
+            throw new CloudRuntimeException("Not all hosts mounted the datastore");
+        }
+    }
+
+    private boolean verifyAllHostsMountedDatastore(List<Pair<ManagedObjectReference, String>> lstHosts, DatastoreMO dsMO) throws Exception {
+        List<HostMO> hostMOs = new ArrayList<>(lstHosts.size());
+
+        for (Pair<ManagedObjectReference, String> host : lstHosts) {
             ManagedObjectReference morHostToMatch = host.first();
             HostMO hostToMatchMO = new HostMO(dsMO.getContext(), morHostToMatch);
 
+            hostMOs.add(hostToMatchMO);
+        }
+
+        return verifyAllHostsMountedDatastore2(hostMOs, dsMO);
+    }
+
+    private boolean verifyAllHostsMountedDatastore2(List<HostMO> lstHosts, DatastoreMO dsMO) throws Exception {
+        int numHostsChecked = 0;
+
+        for (HostMO hostToMatchMO : lstHosts) {
             List<DatastoreHostMount> datastoreHostMounts = dsMO.getHostMounts();
 
             for (DatastoreHostMount datastoreHostMount : datastoreHostMounts) {
@@ -2753,6 +2778,16 @@ public class VmwareStorageProcessor implements StorageProcessor {
         for (Pair<ManagedObjectReference, String> host : hosts) {
             HostMO hostMO = new HostMO(dsMO.getContext(), host.first());
 
+            List<HostMO> hostMOs = new ArrayList<>(1);
+
+            hostMOs.add(hostMO);
+
+            mountVmfsDatastore2(dsMO, hostMOs);
+        }
+    }
+
+    private void mountVmfsDatastore2(DatastoreMO dsMO, List<HostMO> hosts) throws Exception {
+        for (HostMO hostMO : hosts) {
             if (!isDatastoreMounted(dsMO, hostMO)) {
                 HostStorageSystemMO hostStorageSystemMO = hostMO.getHostStorageSystemMO();
 
@@ -2760,11 +2795,15 @@ public class VmwareStorageProcessor implements StorageProcessor {
                     hostStorageSystemMO.mountVmfsVolume(getDatastoreUuid(dsMO, hostMO));
                 }
                 catch (InvalidStateFaultMsg ex) {
-                    List<Pair<ManagedObjectReference, String>> currentHosts = new ArrayList<>(1);
+                    s_logger.trace("'" + ex.getClass().getName() + "' exception thrown: " + ex.getMessage());
 
-                    currentHosts.add(host);
+                    List<HostMO> currentHosts = new ArrayList<>(1);
 
-                    waitForAllHostsToMountDatastore(currentHosts, dsMO);
+                    currentHosts.add(hostMO);
+
+                    s_logger.trace("Waiting for host " + hostMO.getHostName() + " to mount datastore " + dsMO.getName());
+
+                    waitForAllHostsToMountDatastore2(currentHosts, dsMO);
                 }
             }
         }
@@ -2772,12 +2811,29 @@ public class VmwareStorageProcessor implements StorageProcessor {
 
     private void unmountVmfsDatastore(VmwareContext context, VmwareHypervisorHost hyperHost, String datastoreName,
                                       List<Pair<ManagedObjectReference, String>> hosts) throws Exception {
-        ManagedObjectReference morDs = HypervisorHostHelper.findDatastoreWithBackwardsCompatibility(hyperHost, datastoreName);
-        DatastoreMO dsMO = new DatastoreMO(context, morDs);
-
         for (Pair<ManagedObjectReference, String> host : hosts) {
             HostMO hostMO = new HostMO(context, host.first());
 
+            List<HostMO> hostMOs = new ArrayList<>(1);
+
+            hostMOs.add(hostMO);
+
+            unmountVmfsDatastore2(context, hyperHost, datastoreName, hostMOs);
+        }
+    }
+
+    private void unmountVmfsDatastore2(VmwareContext context, VmwareHypervisorHost hyperHost, String datastoreName,
+                                       List<HostMO> hosts) throws Exception {
+        ManagedObjectReference morDs = HypervisorHostHelper.findDatastoreWithBackwardsCompatibility(hyperHost, datastoreName);
+        DatastoreMO dsMO = new DatastoreMO(context, morDs);
+
+        for (HostMO hostMO : hosts) {
+            unmountVmfsVolume(dsMO, hostMO);
+        }
+    }
+
+    private void unmountVmfsVolume(DatastoreMO dsMO, HostMO hostMO) throws Exception {
+        if (isDatastoreMounted(dsMO, hostMO)) {
             HostStorageSystemMO hostStorageSystemMO = hostMO.getHostStorageSystemMO();
 
             hostStorageSystemMO.unmountVmfsVolume(getDatastoreUuid(dsMO, hostMO));
@@ -2902,6 +2958,20 @@ public class VmwareStorageProcessor implements StorageProcessor {
 
                         if (rescan) {
                             rescanAllHosts(hosts, true, false);
+
+                            List<HostInternetScsiHbaStaticTarget> targetsToAdd = new ArrayList<>();
+
+                            targetsToAdd.addAll(getTargets(staticTargetsForHost));
+                            targetsToAdd.addAll(getTargets(dynamicTargetsForHost));
+
+                            for (HostInternetScsiHbaStaticTarget targetToAdd : targetsToAdd) {
+                                HostDatastoreSystemMO hostDatastoreSystemMO = host.getHostDatastoreSystemMO();
+                                String datastoreName = waitForDatastoreName(hostDatastoreSystemMO, targetToAdd.getIScsiName());
+                                ManagedObjectReference morDs = hostDatastoreSystemMO.findDatastoreByName(datastoreName);
+                                DatastoreMO datastoreMO = new DatastoreMO(host.getContext(), morDs);
+
+                                mountVmfsDatastore2(datastoreMO, hosts);
+                            }
                         }
                     }
                     catch (Exception ex) {
@@ -2924,26 +2994,9 @@ public class VmwareStorageProcessor implements StorageProcessor {
 
                 if (targetsToRemove.size() > 0) {
                     if (isRemoveAsync) {
-                        new Thread(() -> {
-                            try {
-                                addRemoveInternetScsiTargetsToAllHosts(false, targetsToRemove, hosts);
-
-                                rescanAllHosts(hosts, true, false);
-                            } catch (Exception ex) {
-                                s_logger.warn(ex.getMessage());
-                            }
-                        }).start();
+                        new Thread(() -> handleRemove(targetsToRemove, host, hosts)).start();
                     } else {
-                        executorService.submit(new Thread(() -> {
-                            try {
-                                addRemoveInternetScsiTargetsToAllHosts(false, targetsToRemove, hosts);
-
-                                rescanAllHosts(hosts, true, false);
-                            }
-                            catch (Exception ex) {
-                                s_logger.warn(ex.getMessage());
-                            }
-                        }));
+                        executorService.submit(new Thread(() -> handleRemove(targetsToRemove, host, hosts)));
                     }
                 }
             }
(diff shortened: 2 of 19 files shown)


plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -2684,6 +2701,30 @@ private boolean verifyAllHostsMountedDatastore(List<Pair<ManagedObjectReference,
...
+
+            for (DatastoreHostMount datastoreHostMount : datastoreHostMounts) {
+                ManagedObjectReference morHost = datastoreHostMount.getKey();
+                HostMO hostMO = new HostMO(dsMO.getContext(), morHost);
+
+                if (hostMO.getHostName().equals(hostToMatchMO.getHostName())) {
+                    if (datastoreHostMount.getMountInfo().isMounted() && datastoreHostMount.getMountInfo().isAccessible()) {
+                        numHostsChecked++;
+                    }
+                    else {
```
rafaelweingartner: What about a log here to show the admin when a data storage is either not mounted or not accessible?

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -2770,6 +2811,25 @@ private void mountVmfsDatastore(DatastoreMO dsMO, List<Pair<ManagedObjectReferen
...
+    private void mountVmfsDatastore2(DatastoreMO dsMO, List<HostMO> hosts) throws Exception {
+        for (HostMO hostMO : hosts) {
+            if (!isDatastoreMounted(dsMO, hostMO)) {
+                HostStorageSystemMO hostStorageSystemMO = hostMO.getHostStorageSystemMO();
+
+                try {
+                    hostStorageSystemMO.mountVmfsVolume(getDatastoreUuid(dsMO, hostMO));
+                }
+                catch (InvalidStateFaultMsg ex) {
+                    List<HostMO> currentHosts = new ArrayList<>(1);
```
rafaelweingartner: In these cases when there is an exception captured, I think you can log a message in TRACE or DEBUG to say that an exception of type X has been caught, and this triggered some processing. In this case, the "InvalidStateFaultMsg" is captured and this triggers a code to wait until the datastores are mounted. The message can be logged in DEBUG and the exception in TRACE level.

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -2778,9 +2838,25 @@ private void unmountVmfsDatastore(VmwareContext context, VmwareHypervisorHost hy
...
-            HostStorageSystemMO hostStorageSystemMO = hostMO.getHostStorageSystemMO();
+            if (isDatastoreMounted(dsMO, hostMO)) {
+                HostStorageSystemMO hostStorageSystemMO = hostMO.getHostStorageSystemMO();
+
+                hostStorageSystemMO.unmountVmfsVolume(getDatastoreUuid(dsMO, hostMO));
+            }
+        }
+    }
+
+    private void unmountVmfsDatastore2(VmwareContext context, VmwareHypervisorHost hyperHost, String datastoreName,
```
rafaelweingartner: out of curiosity. Why do you have two methods with the same name? I mean, I am seeing these new methods ending in 2. What is the difference with the original ones?

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -2902,6 +2978,20 @@ public void handleTargets(boolean add, ModifyTargetsCommand.TargetTypeToRemove t
 
                         if (rescan) {
```
rafaelweingartner: can you extract the body of this IF to a method?

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -2956,6 +3033,61 @@ public void handleTargets(boolean add, ModifyTargetsCommand.TargetTypeToRemove t
         }
     }
 
+    private String waitForDatastoreName(HostDatastoreSystemMO hostDatastoreSystemMO, String iqn) throws Exception {
+        long secondsToWait = 120;
```
rafaelweingartner: Are you planning to externalize this variable?

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -2956,6 +3033,61 @@ public void handleTargets(boolean add, ModifyTargetsCommand.TargetTypeToRemove t
...
+                String datastoreName = waitForDatastoreName(host.getHostDatastoreSystemMO(), target.getIScsiName());
+
+                unmountVmfsDatastore2(host.getContext(), host, datastoreName, hosts);
+            }
+
+            addRemoveInternetScsiTargetsToAllHosts(false, targetsToRemove, hosts);
+
+            rescanAllHosts(hosts, true, false);
+        }
+        catch (Exception ex) {
```
rafaelweingartner: Do you need this catch all here? Can't you use a more specific exception?

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
```diff
@@ -344,17 +354,72 @@ public static SolidFireAccount getAccount(SolidFireConnection sfConnection, Stri
         }
     }
 
-    public static void hostAddedToOrRemovedFromCluster(long hostId, long clusterId, boolean added, String storageProvider,
-            ClusterDao clusterDao, ClusterDetailsDao clusterDetailsDao, PrimaryDataStoreDao storagePoolDao,
-            StoragePoolDetailsDao storagePoolDetailsDao, HostDao hostDao) {
+    private static boolean isCloudStackOnlyVag(SolidFireConnection sfConnection, SolidFireVag sfVag) {
+        long[] volumeIds = sfVag.getVolumeIds();
+
+        if (volumeIds == null || volumeIds.length == 0) {
```
rafaelweingartner: what about using `org.apache.commons.lang3.ArrayUtils.isEmpty(long[])` here?

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
```diff
@@ -397,65 +508,290 @@ public static void hostAddedToOrRemovedFromCluster(long hostId, long clusterId,
...
 
-    public static long placeVolumeInVolumeAccessGroup(SolidFireConnection sfConnection, long sfVolumeId, long storagePoolId,
-                                                      String vagUuid, List<HostVO> hosts, ClusterDetailsDao clusterDetailsDao) {
-        if (hosts == null || hosts.isEmpty()) {
-            throw new CloudRuntimeException("There must be at least one host in the cluster.");
+    // Put the host in an existing VAG or create a new one (only create a new one if all existing VAGs are full (i.e. 64 hosts max per VAG) and if
+    // creating a new VAG won't exceed 4 VAGs for the computer cluster).
+    // If none of the hosts in the cluster are in a VAG, then leave this host out of a VAG.
+    // Place applicable volume IDs in VAG, if need be (account of volume starts with SF_CS_ACCOUNT_PREFIX).
+    private static void handleVagForHost(SolidFireUtil.SolidFireConnection sfConnection, List<SolidFireUtil.SolidFireVag> sfVags, Host host, HostDao hostDao) {
```
rafaelweingartner: Can you use proper java doc notation? Those notations there `//` are for code comments.

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
```diff
@@ -397,65 +508,290 @@ public static void hostAddedToOrRemovedFromCluster(long hostId, long clusterId,
...
+                SolidFireUtil.createVag(sfConnection, "CloudStack-" + UUID.randomUUID().toString(),
+                        new String[]{host.getStorageUrl()}, getVolumeIds(sfConnection, sfVags, host, hostDao));
+            }
         }
+    }
 
-        long lVagId;
+    // Make use of the volume access group (VAG) of a random host in the cluster. With this VAG, collect all of its volume IDs that are for
+    // volumes that are in SolidFire accounts that are for CloudStack.
+    private static long[] getVolumeIds(SolidFireUtil.SolidFireConnection sfConnection, List<SolidFireUtil.SolidFireVag> sfVags,
```
rafaelweingartner: The same here regarding Java docs

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
mike-tutkowski: @rafaelweingartner We could do that, but it's not technically a problem at this point in execution. For example, when we create a datastore on an iSCSI target, it can take up to 30 seconds or so for each host in the cluster to mount the new VMFS volume. When this code returns false, it is just telling the calling code that the VMFS volume is not mounted on all hosts in the cluster. The calling code will wait a bit longer and then re-invoke this method to see if the VMFS volume is now mounted on all hosts in the cluster. At two minutes of trying and failing, the calling code stops invoking this method and just fails the operation.

rafaelweingartner: I know it is not a problem, but it felt like a good thing to log. 
I mean, I did not check others methods that use this block of code to see if they would log meaningful messages in case of a false return here. Reading your explanation it seems that when the operation fails there is some sort of logging to help administrators to understand the problem.  If that is the case, I agree that we do not need to add anything here.

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
mike-tutkowski: It's because I couldn't name these methods the same due to erasure "issues" with Generics.

For example:

private void mountVmfsDatastore(DatastoreMO dsMO, List<Pair<ManagedObjectReference, String>> hosts) throws Exception

private void mountVmfsDatastore2(DatastoreMO dsMO, List<HostMO> hosts) throws Exception

The Java compiler makes the second parameter in each of those methods look the same, so it complained that I had two mountVmfsDatastore methods. To solve the problem, I just changed the second method's name a little bit.

By asking this question, however, you did remind me of something I had planned to do, but forgot. I wanted to have the mountVmfsDatastore make use of the mountVmfsDatastore2 method to remove redundancy. I went ahead and implemented that in my sandbox. It will be in the next commit I push for this PR. I did the same for the unmountVmfsDatastore method.

mike-tutkowski: Same story for verifyAllHostsMountedDatastore and verifyAllHostsMountedDatastore2. I plan to make verifyAllHostsMountedDatastore leverage verifyAllHostsMountedDatastore2.

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
mike-tutkowski: I hadn't planned to. In almost every case, the datastore will mount in under 30 seconds. 120 is really a super worst-case scenario. If we have waited that long and not see the datastore mounted yet, just fail. The normal case is the datastore will mount in 5 - 10 seconds and we won't get anywhere near 120 seconds.

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
mike-tutkowski: So, that's an interesting question. In this case, these comments are aimed at a developer who is looking for help in understanding the logic of the method as opposed to a developer who is looking to simply invoke the method to have some work done for him/her. I'm OK with using JavaDoc, if you feel that makes more sense here, but I figured I would clarify to you first what the intent of these comments is. Please let me know if you'd still like to see this in JavaDoc form. Thanks!

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
mike-tutkowski: Done. It will be in the next commit.

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
mike-tutkowski: One way I could make it more clear that it's a constant is to declare the variable final. Would you prefer that I add that keyword to it?

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
rafaelweingartner: You have a clear definition of the difference between Javadocs and java comments. That is awesome 👍 

I asked because I have seen people trying to write documentation, but they used comment notation instead. If you believe it is ok with Java comment notation, I am ok with it ;)

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
rafaelweingartner: I think it is just fine as is.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
rafaelweingartner: Have you pushed the commit?

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
DaanHoogland: just to butt in here; it is used at line 2661 as well. Is there any reason those two variables with the same name would have different values? I suspect not and if I'm right I'd like to see a constant.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -1654,18 +1654,18 @@ private ModifyTargetsCommand getModifyTargetsCommand(long storagePoolId, String
         details.put(ModifyTargetsCommand.STORAGE_HOST, storagePool.getHostAddress());
         details.put(ModifyTargetsCommand.STORAGE_PORT, String.valueOf(storagePool.getPort()));
 
-        ModifyTargetsCommand modifyTargetsCommand = new ModifyTargetsCommand();
+        ModifyTargetsCommand cmd = new ModifyTargetsCommand();
```
DaanHoogland: how unjavaish of you @mike-tutkowski ;)

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
```diff
@@ -300,7 +311,7 @@ public static String getValue(String keyToMatch, String url, boolean throwExcept
     }
 
     public static String getSolidFireAccountName(String csAccountUuid, long csAccountId) {
-        return "CloudStack_" + csAccountUuid + "_" + csAccountId;
+        return SF_CS_ACCOUNT_PREFIX + csAccountUuid + "_" + csAccountId;
```
DaanHoogland: :) this method could be renamed to compileSolidFireAccountName(..) or formatSolidFireAccountName(..), right? as has been done in recent cleanups in other parts of the code.

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/SolidFireSharedPrimaryDataStoreLifeCycle.java
```diff
@@ -245,7 +243,7 @@ public DataStore initialize(Map<String, Object> dsInfos) {
 
         details.put(SolidFireUtil.VOLUME_ID, String.valueOf(sfVolume.getId()));
 
-        parameters.setUuid(iqn);
+        parameters.setUuid(UUID.randomUUID().toString());
```
DaanHoogland: so this is the only function change in this file amongst a lot of good cleanup. I known not enough of solidfire to judge its merits. Is there any chance you can split reformatting and functional changes to better my review capabilities?

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
mike-tutkowski: We could. The idea is really just to get the String representation of the SolidFire account for the CloudStack account in question. I'm definitely open to changing the method name, if you think it would help. The word "compile" is OK, but it typically invokes thoughts of a code compiler to me.

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/SolidFireSharedPrimaryDataStoreLifeCycle.java
mike-tutkowski: So, honestly, here's the thing (and this may not be a good answer, I understand):

We waste so much time with PRs in verifying that failed tests are, in fact, false calls that I just decided to coalesce this minor update with the other changes. I know it's ideal to have each and every individual unit of functionality, in theory, checked in separately, but, in practice, it's a huge time drain.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
mike-tutkowski: It seems to fit in better with the general pattern I have observed in the codebase (using cmd instead of a longer name). :)

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
mike-tutkowski: Done

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
rafaelweingartner: What about "create", "generate", "build", "construct", or something similar?

mike-tutkowski: So, here's how it's used:

Let's say we need to create a volume.

We know the CloudStack account the volume is associated with. We use this info when calling SolidFireUtil.getSolidFireAccountName to find the name of the SolidFire account that should support this CloudStack account (such a SolidFire account may not exist at this point).

We then use the String that comes back from SolidFireUtil.getSolidFireAccountName and look up the SolidFire account (making an API call to the SolidFire SAN).

If it returns a SolidFire account, we use it. Otherwise, we create a SolidFire account with that name and use it.

rafaelweingartner: The method in question (`getSolidFireAccountName`) does not create anything, right? At least I could not see anything there. It basically defines/generates a name that you then use do to some processing (I believe the processing you just described).

rafaelweingartner: That sounds a bit better. 

I think that @DaanHoogland mentioned the `get` prefix because we had some other PRs normalizing the methods used to retrieve POJOs attributes. However, if `get` is being used as a synonym for `retrieve` that should also work, which seems to be this case. English is my second language so, I might (probably am) mistaken.

mike-tutkowski:
I doubt this Jenkins error has anything to do with this PR:

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running streamer.ByteBufferTest
Tests run: 400, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.194 sec - in streamer.ByteBufferTest
Running streamer.BaseElementTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.231 sec - in streamer.BaseElementTest
Running common.ClientTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 0.575 sec - in common.ClientTest
Running rdpclient.MockServerTest
Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.055 sec <<< FAILURE! - in rdpclient.MockServerTest
testIsMockServerCanUpgradeConnectionToSsl(rdpclient.MockServerTest)  Time elapsed: 1.043 sec  <<< ERROR!
javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate)
	at sun.security.ssl.Handshaker.activate(Handshaker.java:529)
	at sun.security.ssl.SSLSocketImpl.kickstartHandshake(SSLSocketImpl.java:1492)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1361)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1413)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1397)
	at rdpclient.MockServerTest.testIsMockServerCanUpgradeConnectionToSsl(MockServerTest.java:166)

Error in mock server: Received fatal alert: handshake_failure
javax.net.ssl.SSLHandshakeException: Received fatal alert: handshake_failure
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:192)
	at sun.security.ssl.Alerts.getSSLException(Alerts.java:154)
	at sun.security.ssl.SSLSocketImpl.recvAlert(SSLSocketImpl.java:2038)
	at sun.security.ssl.SSLSocketImpl.readRecord(SSLSocketImpl.java:1135)
	at sun.security.ssl.SSLSocketImpl.performInitialHandshake(SSLSocketImpl.java:1385)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1413)
	at sun.security.ssl.SSLSocketImpl.startHandshake(SSLSocketImpl.java:1397)
	at streamer.debug.MockServer.run(MockServer.java:122)

Results :

Tests in error: 
  MockServerTest.testIsMockServerCanUpgradeConnectionToSsl:166 » SSLHandshake No...

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/SolidFireSharedPrimaryDataStoreLifeCycle.java
DaanHoogland: I am not talking separate PRs, @mike-tutkowski . Only separate commits.

DaanHoogland:
I have never agreed with that, and i still oppose it but i'm the minority. In reviewing (but also in backtracking problems) it is very convenient to separate them. And yes, we can always squash our complete history if we so desire ;)

rafaelweingartner:
@mike-tutkowski the idea of separating into commits into feature/functional changes and (formatting or cleanups) is very good. @DaanHoogland convinced me to do that, and it helps reviewers a lot. Then, when the PR is ready to be merged, we squash+merge.

BTW: there is that e-mail thread @DaanHoogland started, and then we stopped discussing [1]. Please share your follow-ups and ideas there, so we can continue this debate and convert it into a method that can satisfy our needs.

[1] https://lists.apache.org/thread.html/9852436322eae7a3158fb916226e81155a3082fb134034a9555f75ba@%3Cdev.cloudstack.apache.org%3E

mike-tutkowski:
@rafaelweingartner @DaanHoogland I went ahead and sent my comments to dev@ regarding that e-mail thread.

In the meanwhile, I have pushed another commit on top of the prior one for this PR. Per my e-mail to dev@, I think we should end up squashing the commits of a PR prior to merging it (use the commits as a temporary history during the life of the PR).

plugins/storage/volume/solidfire/src/main/java/org/apache/cloudstack/storage/datastore/util/SolidFireUtil.java
rafaelweingartner: Thinking about the `String.format` method, where you send the "pattern" and then the parameters to create a new String. I do believe that the `format` as a prefix looks good.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

mike-tutkowski:
All managed-storage tests have passed. Any test that notes a failure was due to an environmental issue. Once the environmental issue was corrected, the test was re-run and passed.

[results.txt](https://github.com/apache/cloudstack/files/2199427/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199428/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199430/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199431/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199432/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199433/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199434/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199436/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199438/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199439/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199440/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199441/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199442/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199444/results.txt)
[results.txt](https://github.com/apache/cloudstack/files/2199445/results.txt)

mike-tutkowski:
I added more integration testing and squashed the prior commits into one in preparation for merging this PR.

mike-tutkowski:
I also do not believe any of the regression tests that failed are related to this PR.

@DaanHoogland @rafaelweingartner Any further comments on this PR? If not, perhaps one of you can give a LGTM. That will make two and I can merge this PR.

rafaelweingartner:
I do think everything is fine. However, I do suggest you using in your next PRs what @DaanHoogland proposed regarding formatting commits.

mike-tutkowski:
Sounds good. I have two other open PRs at the time being, so I can look into it for them.

PR #2506, opened 2018-03-23
CLOUDSTACK-10341: Reduce systemvmtemplate size, install nftables

rhtyd:
## Description

This reduces systemvmtemplate size by 600MB and installs nftables,
updates iptables.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [x] Cleanup (Code refactoring and cleanup, that may add test cases)

## Screenshots (if appropriate):

screenshot from 2018-03-23 13-32-44
screenshot from 2018-03-23 13-33-03

## How Has This Been Tested?

Manually built using packer and tested locally by running smoketests.

## Checklist:

- [x] I have read the [CONTRIBUTING](https://github.com/apache/cloudstack/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] All new and existing tests passed.

PR #2508, opened 2018-03-23
CLOUDSTACK-9114: Reduce VR downtime during network restart

rhtyd:
## Description

Every time there is a major CloudStack version that requires a new systemvmtemplate, admins need to restart VRs which can take few (sometimes several) minutes. The aim is to make it easier to perform a rolling restart (or blue-green deployment) of the network with acceptable downtime. In this strategy, a new VR is provisioned before the old VR is destroyed/cleaned for non-redundant isolated VRs. For redundant networks, a rolling replacement of routers is performed.

The network downtime is usually the short-time (less than 10s) after the new VR starts (nics are up) and old VR is destroyed. The downtime may be affected by several factors as the new VR has a different mac address, stale ARP cache may take upto 30seconds (default `gc_timeout` on Linux) to renew in guests making egress traffic downtime slightly higher than ingress traffic (ingress, i.e. public network requesting access to VR's public ip). As an illustration, ingress traffic such as a http request or ssh request to the VR's public IP that port-forwards to guest VM may face lower downtime than egress traffic (say a ping or curl from guest VM to public Internet). Among several environment factors, the following interval/timeouts affect how the old VR's ip-mac address is cached:

    /proc/sys/net/ipv4/neigh/default/gc_interval
    /proc/sys/net/ipv4/neigh/default/gc_stale_time
    /proc/sys/net/ipv4/route/gc_interval
    /proc/sys/net/ipv4/route/gc_timeout

To make things faster, a gratituous ARP notification is sent by the new VR (again) after the old VR is successfully destroyed.

Project Goals:
- [x] Reduce systemvmtemplate size, see #2506. Smaller the image size, faster it will be to copy template to primary storage to provision new VR.
- [x] Rolling reboot/deployment of VRs for isolated network with non-rVRs
- [x] Rolling reboot/deployment of VRs for isolated network with rVRs
- [x] Rolling reboot/deployment of VRs for VPCs with non-rVRs and rVRs

Future:
- Speed up iptables rules application 
- Move away from iptables, move to nftables
- Move away from Python2 codebase to Python3 or something else like Go? TBD

## Types of changes

### Non-Redundant Isolated Network VRs

Rolling restart in action, a new VR is deployed lifing the VR as a redundant VR:
screenshot from 2018-03-23 22-36-27

The old VR is stopped and destroyed:
screenshot from 2018-03-23 22-36-53

The new VR's redundancy is removed and is reprogrammed:
screenshot from 2018-03-23 22-37-02

Meanwhile, the following screenshot shows ping drops from a guest VM, in a non-redundant isolated network: (usually 2-10 ping drops observed)
screenshot from 2018-03-23 17-53-46

### Redundant Isolated Network VR

Rolling restart starts by stopping old backup VR:
screenshot from 2018-03-23 22-54-56

New backup VR deployed:
screenshot from 2018-03-23 22-55-10

Old master stopped, new backup becomes new master:
screenshot from 2018-03-23 22-55-56

Old master destroyed and another backup VR deployed:
screenshot from 2018-03-23 22-56-07

Redundant VRs after rollling reboot:
screenshot from 2018-03-23 22-57-11

In the meanwhile, usually 0-4 ping drops observed in guest VM:
screenshot from 2018-03-23 22-57-40

## How Has This Been Tested?

Tested with following, benchmarks show rounded off avg/max downtime seconds for restarting non-redundant isolated VRs when `cleanup=true`:

Env | This PR | 4.11.0.0 | 4.9.3.0
--- | --- | --- | ---
KVM CentOS7        | 8s   | 50s | 70s
KVM CentOS6        | 10s   | 60s | 80s
XenServer 71          | 8s   | 70s | NA
XenServer 65sp1    | 20s | 70s | 100s
VMware 65              | 10s | 120s | NA
VMware 55u3          | 20s | 160s | 250s

** The downtime seconds may be higher in real-world environment under heavy load. Redundant routers show a downtime of 0-4 seconds.
*** All used a CentOS7 based management server, tested in environments deployed by [Trillian](https://github.com/shapeblue/Trillian)

Ingress traffic was tested using ssh and http services hosted on the guest VM. Example, sshping like utility I wrote for calculating ingress traffic downtime: (number of fail count = downtime seconds)

```
sshping() {                                                                                                                                              
  ip=$1                                                                            
  total_count=0                                                                    
  fail_count=0                                                                     
  while true; do                                                                   
    start_time=$(date +%s)                                                         
    timeout 1.5 sshpass -p 'password' ssh -o StrictHostKeyChecking=no root@$1 "uptime" > /dev/null 2>&1
    if [[ $? -gt 0 ]]; then                                                        
        fail_count=$((fail_count+1))                                               
    fi                                                                             
    end_time=$(date +%s)                                                           
    diff=$(($end_time - $start_time))                                              
    if [[ $diff -eq 0 ]]; then                                                     
        sleep 1                                                                    
    fi                                                                             
                                                                                   
    total_count=$((total_count+1))                                                 
    echo "Fail/Total = $fail_count/$total_count"                                   
  done                                                                             
}  

```

## Checklist:

@blueorangutan package

### DIFF
--- a/api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
@@ -57,6 +57,9 @@ public class RestartNetworkCmd extends BaseAsyncCmd {
     @Parameter(name = ApiConstants.CLEANUP, type = CommandType.BOOLEAN, required = false, description = "If cleanup old network elements")
     private Boolean cleanup;
 
+    @Parameter(name = ApiConstants.MAKEREDUNDANT, type = CommandType.BOOLEAN, required = false, description = "Turn the network into a network with redundant routers.", since = "4.11.1")
+    private Boolean makeRedundant;
+
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
     /////////////////////////////////////////////////////
@@ -77,6 +80,13 @@ public class RestartNetworkCmd extends BaseAsyncCmd {
         return true;
     }
 
+    public Boolean getMakeRedundant() {
+        if (makeRedundant != null) {
+            return makeRedundant;
+        }
+        return true;
+    }
+
     /////////////////////////////////////////////////////
     /////////////// API Implementation///////////////////
     /////////////////////////////////////////////////////
@@ -92,7 +102,7 @@ public class RestartNetworkCmd extends BaseAsyncCmd {
 
     @Override
     public void execute() throws ResourceUnavailableException, ResourceAllocationException, ConcurrentOperationException, InsufficientCapacityException {
-        boolean result = _networkService.restartNetwork(this, getCleanup());
+        boolean result = _networkService.restartNetwork(this, getCleanup(), getMakeRedundant());
         if (result) {
             SuccessResponse response = new SuccessResponse(getCommandName());
             setResponseObject(response);
--- a/api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
@@ -77,14 +77,14 @@ public class RestartNetworkCmd extends BaseAsyncCmd {
         if (cleanup != null) {
             return cleanup;
         }
-        return true;
+        return false;
     }
 
     public Boolean getMakeRedundant() {
         if (makeRedundant != null) {
             return makeRedundant;
         }
-        return true;
+        return false;
     }
 
     /////////////////////////////////////////////////////
--- a/api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
@@ -55,10 +55,10 @@ public class RestartNetworkCmd extends BaseAsyncCmd {
     private Long id;
 
     @Parameter(name = ApiConstants.CLEANUP, type = CommandType.BOOLEAN, required = false, description = "If cleanup old network elements")
-    private Boolean cleanup;
+    private Boolean cleanup = false;
 
     @Parameter(name = ApiConstants.MAKEREDUNDANT, type = CommandType.BOOLEAN, required = false, description = "Turn the network into a network with redundant routers.", since = "4.11.1")
-    private Boolean makeRedundant;
+    private Boolean makeRedundant = false;
 
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
@@ -74,17 +74,11 @@ public class RestartNetworkCmd extends BaseAsyncCmd {
     }
 
     public Boolean getCleanup() {
-        if (cleanup != null) {
-            return cleanup;
-        }
-        return false;
+        return cleanup;
     }
 
     public Boolean getMakeRedundant() {
-        if (makeRedundant != null) {
-            return makeRedundant;
-        }
-        return false;
+        return makeRedundant;
     }
 
     /////////////////////////////////////////////////////
--- a/api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
@@ -49,10 +49,10 @@ public class RestartVPCCmd extends BaseAsyncCmd {
     private Long id;
 
     @Parameter(name = ApiConstants.CLEANUP, type = CommandType.BOOLEAN, required = false, description = "If cleanup old network elements")
-    private Boolean cleanup;
+    private Boolean cleanup = false;
 
     @Parameter(name = ApiConstants.MAKEREDUNDANT, type = CommandType.BOOLEAN, required = false, description = "Turn a single VPC into a redundant one.")
-    private Boolean makeredundant;
+    private Boolean makeredundant = false;
 
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
@@ -63,17 +63,11 @@ public class RestartVPCCmd extends BaseAsyncCmd {
     }
 
     public Boolean getCleanup() {
-        if (cleanup != null) {
-            return cleanup;
-        }
-        return false;
+        return cleanup;
     }
 
     public Boolean getMakeredundant() {
-        if (makeredundant != null) {
-            return makeredundant;
-        }
-        return false;
+        return makeredundant;
     }
 
     /////////////////////////////////////////////////////
--- a/api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
@@ -66,14 +66,14 @@ public class RestartVPCCmd extends BaseAsyncCmd {
         if (cleanup != null) {
             return cleanup;
         }
-        return true;
+        return false;
     }
 
     public Boolean getMakeredundant() {
         if (makeredundant != null) {
             return makeredundant;
         }
-        return true;
+        return false;
     }
 
     /////////////////////////////////////////////////////
--- a/api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
+++ b/api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
@@ -51,7 +51,7 @@ public class RestartVPCCmd extends BaseAsyncCmd {
     @Parameter(name = ApiConstants.CLEANUP, type = CommandType.BOOLEAN, required = false, description = "If cleanup old network elements")
     private Boolean cleanup;
 
-    @Parameter(name = ApiConstants.MAKEREDUNDANTE, type = CommandType.BOOLEAN, required = false, description = "Turn a single VPC into a redundant one.")
+    @Parameter(name = ApiConstants.MAKEREDUNDANT, type = CommandType.BOOLEAN, required = false, description = "Turn a single VPC into a redundant one.")
     private Boolean makeredundant;
 
     /////////////////////////////////////////////////////
--- a/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -2902,9 +2902,13 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
         final List<DomainRouterVO> oldRouters = _routerDao.findByNetwork(network.getId());
 
         // Deploy a new router
-        network.setRollingRestart(true);
+        if (oldRouters.size() > 0) {
+            network.setRollingRestart(true);
+        }
         implementNetworkElements(dest, context, network, offering, providersToImplement);
-        network.setRollingRestart(false);
+        if (oldRouters.size() > 0) {
+            network.setRollingRestart(false);
+        }
 
         // For redundant network wait for 3*advert_int+skew_seconds for VRRP to kick in
         if (network.isRedundant() || (oldRouters.size() == 1 && oldRouters.get(0).getIsRedundantRouter())) {
@@ -2918,9 +2922,12 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
             _routerService.destroyRouter(oldRouter.getId(), context.getAccount(), context.getCaller().getId());
         }
 
-        // Add a new backup router for redundant network
         if (network.isRedundant()) {
+            // Add a new backup router for redundant network
             implementNetworkElements(dest, context, network, offering, providersToImplement);
+        } else {
+            // Re-apply rules for non-redundant network
+            implementNetworkElementsAndResources(dest, context, network, offering);
         }
 
         return validateNewRouters(_routerDao.findByNetwork(network.getId()));
--- a/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -38,15 +38,11 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
-import com.cloud.utils.StringUtils;
-import org.apache.log4j.Logger;
-
 import org.apache.cloudstack.acl.ControlledEntity.ACLType;
 import org.apache.cloudstack.context.CallContext;
 import org.apache.cloudstack.engine.cloud.entity.api.db.VMNetworkMapVO;
 import org.apache.cloudstack.engine.cloud.entity.api.db.dao.VMNetworkMapDao;
 import org.apache.cloudstack.engine.orchestration.service.NetworkOrchestrationService;
-import org.apache.cloudstack.framework.config.ConfigDepot;
 import org.apache.cloudstack.framework.config.ConfigKey;
 import org.apache.cloudstack.framework.config.ConfigKey.Scope;
 import org.apache.cloudstack.framework.config.Configurable;
@@ -54,8 +50,7 @@ import org.apache.cloudstack.framework.config.dao.ConfigurationDao;
 import org.apache.cloudstack.framework.messagebus.MessageBus;
 import org.apache.cloudstack.framework.messagebus.PublishScope;
 import org.apache.cloudstack.managed.context.ManagedContextRunnable;
-import org.apache.cloudstack.region.PortableIpDao;
-
+import org.apache.log4j.Logger;
 
 import com.cloud.agent.AgentManager;
 import com.cloud.agent.Listener;
@@ -87,7 +82,6 @@ import com.cloud.deploy.DataCenterDeployment;
 import com.cloud.deploy.DeployDestination;
 import com.cloud.deploy.DeploymentPlan;
 import com.cloud.domain.Domain;
-import com.cloud.event.dao.UsageEventDao;
 import com.cloud.exception.ConcurrentOperationException;
 import com.cloud.exception.ConnectionException;
 import com.cloud.exception.InsufficientAddressCapacityException;
@@ -119,6 +113,7 @@ import com.cloud.network.Networks.TrafficType;
 import com.cloud.network.PhysicalNetwork;
 import com.cloud.network.PhysicalNetworkSetupInfo;
 import com.cloud.network.RemoteAccessVpn;
+import com.cloud.network.VpcVirtualNetworkApplianceService;
 import com.cloud.network.addr.PublicIp;
 import com.cloud.network.dao.AccountGuestVlanMapDao;
 import com.cloud.network.dao.AccountGuestVlanMapVO;
@@ -128,7 +123,6 @@ import com.cloud.network.dao.IPAddressVO;
 import com.cloud.network.dao.NetworkAccountDao;
 import com.cloud.network.dao.NetworkAccountVO;
 import com.cloud.network.dao.NetworkDao;
-import com.cloud.network.dao.NetworkDetailsDao;
 import com.cloud.network.dao.NetworkDomainDao;
 import com.cloud.network.dao.NetworkDomainVO;
 import com.cloud.network.dao.NetworkServiceMapDao;
@@ -141,7 +135,6 @@ import com.cloud.network.dao.PhysicalNetworkTrafficTypeVO;
 import com.cloud.network.dao.PhysicalNetworkVO;
 import com.cloud.network.dao.RemoteAccessVpnDao;
 import com.cloud.network.dao.RemoteAccessVpnVO;
-import com.cloud.network.dao.VpnUserDao;
 import com.cloud.network.element.AggregatedCommandExecutor;
 import com.cloud.network.element.DhcpServiceProvider;
 import com.cloud.network.element.DnsServiceProvider;
@@ -183,6 +176,7 @@ import com.cloud.user.User;
 import com.cloud.user.dao.AccountDao;
 import com.cloud.utils.NumbersUtil;
 import com.cloud.utils.Pair;
+import com.cloud.utils.StringUtils;
 import com.cloud.utils.UuidUtils;
 import com.cloud.utils.component.AdapterBase;
 import com.cloud.utils.component.ManagerBase;
@@ -201,9 +195,9 @@ import com.cloud.utils.db.TransactionStatus;
 import com.cloud.utils.exception.CloudRuntimeException;
 import com.cloud.utils.fsm.NoTransitionException;
 import com.cloud.utils.fsm.StateMachine2;
+import com.cloud.utils.net.Dhcp;
 import com.cloud.utils.net.NetUtils;
 import com.cloud.vm.DomainRouterVO;
-import com.cloud.utils.net.Dhcp;
 import com.cloud.vm.Nic;
 import com.cloud.vm.Nic.ReservationStrategy;
 import com.cloud.vm.NicExtraDhcpOptionVO;
@@ -287,11 +281,11 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
     @Inject
     VMNetworkMapDao _vmNetworkMapDao;
     @Inject
-    DomainRouterDao _rotuerDao;
+    DomainRouterDao _routerDao;
     @Inject
     RemoteAccessVpnDao _remoteAccessVpnDao;
     @Inject
-    VpnUserDao _vpnUserDao;
+    VpcVirtualNetworkApplianceService _routerService;
 
     List<NetworkGuru> networkGurus;
 
@@ -369,17 +363,9 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
     @Inject
     NetworkACLManager _networkACLMgr;
     @Inject
-    UsageEventDao _usageEventDao;
-    @Inject
     NetworkModel _networkModel;
     @Inject
     NicSecondaryIpDao _nicSecondaryIpDao;
-    @Inject
-    PortableIpDao _portableIpDao;
-    @Inject
-    ConfigDepot _configDepot;
-    @Inject
-    NetworkDetailsDao _networkDetailsDao;
 
     protected StateMachine2<Network.State, Network.Event, Network> _stateMachine;
     ScheduledExecutorService _executor;
@@ -1147,29 +1133,7 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
         }
         // get providers to implement
         final List<Provider> providersToImplement = getNetworkProviders(network.getId());
-        for (final NetworkElement element : networkElements) {
-            if (providersToImplement.contains(element.getProvider())) {
-                if (!_networkModel.isProviderEnabledInPhysicalNetwork(_networkModel.getPhysicalNetworkId(network), element.getProvider().getName())) {
-                    // The physicalNetworkId will not get translated into a uuid by the reponse serializer,
-                    // because the serializer would look up the NetworkVO class's table and retrieve the
-                    // network id instead of the physical network id.
-                    // So just throw this exception as is. We may need to TBD by changing the serializer.
-                    throw new CloudRuntimeException("Service provider " + element.getProvider().getName() + " either doesn't exist or is not enabled in physical network id: "
-                            + network.getPhysicalNetworkId());
-                }
-
-                if (s_logger.isDebugEnabled()) {
-                    s_logger.debug("Asking " + element.getName() + " to implement " + network);
-                }
-
-                if (!element.implement(network, offering, dest, context)) {
-                    final CloudRuntimeException ex = new CloudRuntimeException("Failed to implement provider " + element.getProvider().getName() + " for network with specified id");
-                    ex.addProxyObject(network.getUuid(), "networkId");
-                    throw ex;
-                }
-            }
-        }
-
+        implementNetworkElements(dest, context, network, offering, providersToImplement);
 
         //Reset the extra DHCP option that may have been cleared per nic.
         List<NicVO> nicVOs = _nicDao.listByNetworkId(network.getId());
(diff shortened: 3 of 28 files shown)


rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

resmo:
looks promising! Can not wait to test it, +1

DaanHoogland:
looks good @rhtyd

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

server/src/com/cloud/network/element/VirtualRouterElement.java
```diff
@@ -227,7 +227,7 @@ public boolean implement(final Network network, final NetworkOffering offering,
         final List<DomainRouterVO> routers = routerDeploymentDefinition.deployVirtualRouter();
 
         int routerCounts = 1;
-        if (offering.getRedundantRouter()) {
+        if (offering.getRedundantRouter() || network.isRollingRestart()) {
```
DaanHoogland: neat solution

rhtyd: Unlikely we'll have more than 2 routers for redundancy. However, the code has an internal limit of 5 (after which it logs warning/errors that a network has more than 5 routers, which can happen in some edge cases).

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2838,8 +2828,27 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
         s_logger.debug("Restarting network " + networkId + "...");
 
         final ReservationContext context = new ReservationContextImpl(null, null, callerUser, callerAccount);
+        final NetworkOffering offering = _networkOfferingDao.findByIdIncludingRemoved(network.getNetworkOfferingId());
 
         if (cleanup) {
+            // Destroy non-running VRs
+            for (final DomainRouterVO router : _routerDao.findByNetwork(network.getId())) {
+                if (router.getState() != VirtualMachine.State.Running) {
```
rafaelweingartner: Should VRs in `Migrating` or `Starting` be destroyed as well?

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2838,8 +2828,27 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+                    s_logger.debug("Destroying " + router + " as it is not in Running state");
+                    _routerService.destroyRouter(router.getId(), context.getAccount(), context.getCaller().getId());
+                }
+            }
+            // Perform rolling restart for Virtual Routers
+            final List<DomainRouterVO> routers = _routerDao.findByNetwork(network.getId());
+            if (routers != null && !routers.isEmpty()) {
```
rafaelweingartner: You can use `CollectionUtils.isNotEmpty` here

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2838,8 +2828,27 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
         s_logger.debug("Restarting network " + networkId + "...");
 
         final ReservationContext context = new ReservationContextImpl(null, null, callerUser, callerAccount);
+        final NetworkOffering offering = _networkOfferingDao.findByIdIncludingRemoved(network.getNetworkOfferingId());
 
         if (cleanup) {
```
rafaelweingartner: What about extracting this IF body to a method? This allows proper documentation (instead of code comments), and unit tests.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2876,82 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
         }
     }
 
+    @Override
+    public VirtualRouter findExpendableRouterForRollingRestart(final List<? extends VirtualRouter> routers) {
+        VirtualRouter expendableRouter = null;
+        if (routers != null && routers.size() > 1) {
```
rafaelweingartner: You invert the logic here. `if (router == null || router.size <=1)`, then return null;
This would allow you to remove one nesting level.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2876,82 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+
+        // Create a new router
+        network.setRollingRestart(true);
+        implementNetworkElements(dest, context, network, offering, providersToImplement);
+        network.setRollingRestart(false);
+
+        if (network.isRedundant()) {
+            try {
+                // For redundant network wait for 3*(1+skew_seconds) for VRRP to kick in
+                Thread.sleep(10000L);
```
rafaelweingartner: I am reading the comment, and it uses a variable there, but the code uses a hard-coded value. Is this right?

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2876,82 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+            } catch (final InterruptedException ignored) {
```
rafaelweingartner: Please, do not hide exceptions. 
You should at least log them.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2876,82 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+            }
+            if (expendableRouter == null) {
+                expendableRouter = routers.get(routers.size() - 1);
+            }
+        }
+        return expendableRouter;
+    }
+
+    private boolean rollingRestartRouters(final NetworkVO network, final NetworkOffering offering, final List<DomainRouterVO> oldRouters, final ReservationContext context) throws ResourceUnavailableException, ConcurrentOperationException, InsufficientCapacityException {
+        final Account callerAccount = CallContext.current().getCallingAccount();
```
rafaelweingartner: Can you break this method into smaller ones? It is quite big already.

server/src/com/cloud/network/vpc/VpcManagerImpl.java
```diff
@@ -1508,19 +1519,33 @@ public boolean restartVpc(final long vpcId, final boolean cleanUp, final boolean
...
+                if (_vpcDao.update(vpc.getId(), entity)) {
+                    vpc = entity;
+                }
 
                 // If the offering and redundant column are changing, force the
                 // clean up.
                 forceCleanup = true;
             }
 
             if (forceCleanup) {
```
rafaelweingartner: Can you extract the body of this IF to a method?

server/src/com/cloud/network/vpc/VpcManagerImpl.java
```diff
@@ -2435,4 +2460,66 @@ public boolean isSrcNatIpRequired(long vpcOfferingId) {
         final Map<Network.Service, Set<Network.Provider>> vpcOffSvcProvidersMap = getVpcOffSvcProvidersMap(vpcOfferingId);
         return vpcOffSvcProvidersMap.get(Network.Service.SourceNat).contains(Network.Provider.VPCVirtualRouter);
     }
+
+    private boolean rollingRestartVpc(final Vpc vpc, final List<DomainRouterVO> oldRouters, final ReservationContext context) throws ResourceUnavailableException, ConcurrentOperationException, InsufficientCapacityException {
```
rafaelweingartner: This method has a good chunk of code duplicated with `rollingRestartRouters`. Would mind extracting the code common code to methods and then reuse them?

rhtyd:
@blueorangutan package

api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
```diff
@@ -57,6 +57,9 @@
     @Parameter(name = ApiConstants.CLEANUP, type = CommandType.BOOLEAN, required = false, description = "If cleanup old network elements")
     private Boolean cleanup;
 
+    @Parameter(name = ApiConstants.MAKEREDUNDANTE, type = CommandType.BOOLEAN, required = false, description = "Turn the network into a network with redundant routers.", since = "4.11.1")
```
nitin-maharana: @rhtyd, Thanks for adding this functionality. I think there is a typo in ApiConstants name, E at the end. Would be great if we change the name of existing name to MAKE_RENDUNDANT, I just checked the usage is also very less.

rhtyd: Thanks for spotting @nitin-maharana, will fix the typo

api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
```diff
@@ -77,6 +80,13 @@ public Boolean getCleanup() {
         return true;
     }
 
+    public Boolean getMakeRedundant() {
+        if (makeRedundant != null) {
+            return makeRedundant;
+        }
+        return true;
```
nitin-maharana: If we don't pass anything, I think it should be false, it would continue with normal restart instead of this blue-green deployment.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2849,89 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+        // For redundant network wait for 3*advert_int+skew_seconds for VRRP to kick in
+        if (network.isRedundant() || (oldRouters.size() == 1 && oldRouters.get(0).getIsRedundantRouter())) {
+            try {
+                Thread.sleep(10000L);
+            } catch (final InterruptedException ignored) {}
+        }
+
+        // Destroy old routers
+        for (final DomainRouterVO oldRouter : oldRouters) {
+            _routerService.destroyRouter(oldRouter.getId(), context.getAccount(), context.getCaller().getId());
```
nitin-maharana: Here, before validating the new routers, we are destroying the old ones. If the new routers don't come up properly, we won't have old routers to roll back.

ui/scripts/network.js
```diff
@@ -1100,11 +1100,23 @@
                                         });
                                         args.$form.find('.form-item[rel=cleanup]').find('input').attr('checked', 'checked'); //checked
                                         args.$form.find('.form-item[rel=cleanup]').css('display', 'inline-block'); //shown
+                                        args.$form.find('.form-item[rel=makeredundant]').find('input').attr('checked', 'checked'); //checked
+                                        args.$form.find('.form-item[rel=makeredundant]').css('display', 'inline-block'); //shown
+
+                                        if (Boolean(args.context.networks[0].redundantrouter)) {
```
nitin-maharana: As I understand, for RVR also the rolling upgrade will be applicable, any reason we hide the button.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
rhtyd: @nitin-maharana if you read the code again we're deploying a new VR first. If it fails, the method will fail at that point and won't fallthrough to this code that destroys the old VRs.

api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
rhtyd: Will fix that.

ui/scripts/network.js
rhtyd: This code will show option to make network redundant during restart, we hide the checkbox/label for networks that are already redundant.

nitin-maharana: @rhtyd, When we say redundant networks, that means it already contains the new upgraded VR(after Blue-Green deployment), we created after a restart? If that is the case, I think it would be great, if we give a different name instead of redundant. Because there is a chance we will misinterpret as RVR.

rhtyd: @nitin-maharana I'm not sure where the confusion is, a redundant network already has rVRs (backup+master) therefore no need to show the label/checkbox to user. This is same as the option you get when you restart VPC (see try that, restart VPC with 4.11+).

rhtyd: @nitin-maharana I'm not sure what the confusion is. The UI shows the option to makeRedundant a network which is not redundant, otherwise the checkbox is not shown. This is same as what you see in case of VPC restart button/dialog. Thanks.

api/src/org/apache/cloudstack/api/command/user/network/RestartNetworkCmd.java
rhtyd: Fixed.

server/src/com/cloud/network/vpc/VpcManagerImpl.java
rhtyd: The body of IF is rather small already?

systemvm/debian/opt/cloud/bin/configure.py
```diff
@@ -957,6 +957,7 @@ def process(self):
         acls.process()
 
         acls = CsAcl('firewallrules', self.config)
+        acls.flushAllowAllEgressRules()
```
rhtyd: FYI - this fixes the issue of egress traffic failure on network start.

rhtyd:
@blueorangutan package

server/src/com/cloud/network/vpc/VpcManagerImpl.java
rafaelweingartner: Yes it is, if you consider only the body of the IF, but eh method is already quite big.  The more we extract, the more unit tests we enable. Moreover, it becomes easier to read the code.

rhtyd:
@ustcweizhou @resmo @DaanHoogland @rafaelweingartner @wido @nitin-maharana @nvazquez @NuxRo this is ready for review, please help review and test. Packages of this PR are mirrored here: https://lab.yadav.cloud/testing/vr-downtime
I'll open up test/stabilization fixes in a separate FR for 4.11 branch.

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2849,89 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+        for (final VirtualRouter router : remainingRouters) {
+            if (router.getRedundantState() == VirtualRouter.RedundantState.BACKUP) {
+                backupRouter = router;
+            }
+        }
+        if (backupRouter == null) {
+            backupRouter = routers.get(routers.size() - 1);
+        }
+        if (backupRouter != null) {
+            _routerService.destroyRouter(backupRouter.getId(), context.getAccount(), context.getCaller().getId());
```
DaanHoogland: it seems this will be called even when a redundant backup router is required in this network. Can you explain how this works?

systemvm/debian/opt/cloud/templates/keepalived.conf.templ
```diff
@@ -31,9 +31,9 @@ vrrp_instance inside_network {
     nopreempt
 
     advert_int 1
-    garp_master_delay 1
-    garp_master_repeat 10
-    garp_master_refresh 5
+    #garp_master_delay 2
```
DaanHoogland: we can delete commented code

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
rhtyd: The purpose of removal of expendable router in `destroyExpendableRouters` is to find and destroy VRs of network where destroying has no side effect on the network. We first try to find and destroy bad routers (in non-running states), then we see how many routers are left if they are less than 2 (i.e. not redundant routers) we break/exit the method. Otherwise, we see if any of those two routers are backup router which can be destroyed. This method is shared by both the vpc and non-vpc routers.

systemvm/debian/opt/cloud/templates/keepalived.conf.templ
rhtyd: Fixed.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
DaanHoogland: Is there a reason for this method to be public? it can be called at anytime on a redundant network at which time it would destroy the backup.

rhtyd: Yes, the reason is that this common logic is used at two places one where restart of isolated networks are handled and another vpc mgr class that handles the restart of VPCs. @rafaelweingartner advised to refactor the common logic, I did as much I could.

rhtyd:
Ping - @ustcweizhou @resmo @rafaelweingartner @wido @nitin-maharana @nvazquez @NuxRo review and testing is requested.
Packages of this PR are mirrored here: https://lab.yadav.cloud/testing/vr-downtime
SystemVM template can be used from here: https://lab.yadav.cloud/systemvmtemplates/4.11/

server/src/com/cloud/network/vpc/VpcManagerImpl.java
```diff
@@ -2435,4 +2444,43 @@ public boolean isSrcNatIpRequired(long vpcOfferingId) {
...
+        }
+        startVpc(vpc, dest, context);
+        if (oldRouters.size() > 0) {
+            vpc.setRollingRestart(false);
+        }
+
+        // For redundant vpc wait for 3*advert_int+skew_seconds for VRRP to kick in
+        if (vpc.isRedundant() || (oldRouters.size() == 1 && oldRouters.get(0).getIsRedundantRouter())) {
+            try {
+                Thread.sleep(10000L);
```
GabrielBrascher: What do you think about extracting the value `10000L` to a static variable?

rhtyd: Fixed.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -1216,6 +1180,32 @@ public void implementNetworkElementsAndResources(final DeployDestination dest, f
         }
     }
 
+    protected void implementNetworkElements(final DeployDestination dest, final ReservationContext context, final Network network, final NetworkOffering offering, final List<Provider> providersToImplement)
```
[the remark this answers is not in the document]
rhtyd: Fixed.

rhtyd:
@blueorangutan package

rhtyd:
@GabrielBrascher I've addressed your comment, please re-review and lgtm if you're okay with them. Thanks.

rhtyd:
@blueorangutan test matrix

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
DaanHoogland: got it @rhtyd. I still dread this public method being called from places that do not intent what it was programmed to do. As in this method a backup router will be destroyed no matter what. How about adding some javadoc:
/**
  * This method will first destroy and router in state Stopped, Error, Shutdowned or Unknown
  * Afterthat it will destroy any backup router or, if it can find non the last router in the list
  */

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

rhtyd:
@blueorangutan test matrix

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rafaelweingartner:
@rhtyd do I need this PR in order to use the system vms 4.11.1.0 that you generated?

rhtyd:
@rafaelweingartner no, you'll need the new 4.11.1.0 systemvmtemplate for 4.11.1+ anyway. I created and posted the repo and systemvmtemplate for anyone wanting to test this PR.

rhtyd:
@blueorangutan test

rafaelweingartner:
I am using the system vm templates that you posted, but for some reason they are starting but not connecting to the management server. I am also using this PR.

rhtyd:
@rafaelweingartner if you're using xenserver for testing, please use this PR against latest 4.11 branch which got a xs blocker fix. If you debug your hypervisor host or mgmt server at what stage is it failing to patch or start the systemvms? Also what's your test env like?

rafaelweingartner:
Never mind. It was a problem with my hosts. Everything was working just fine with this PR and the new system vms 👍

rhtyd:
@rafaelweingartner great, looking forward to your review and test results.

api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
```diff
@@ -66,14 +66,14 @@ public Boolean getCleanup() {
         if (cleanup != null) {
             return cleanup;
         }
-        return true;
+        return false;
     }
 
     public Boolean getMakeredundant() {
         if (makeredundant != null) {
```
rafaelweingartner: Is there is a default value, why not use a primitive value with a default of `false`? Then, you do not need these logics in a POJO get method.

api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
```diff
@@ -66,14 +66,14 @@ public Boolean getCleanup() {
         if (cleanup != null) {
             return cleanup;
         }
-        return true;
+        return false;
```
rafaelweingartner: The same is valid here

rhtyd:
@blueorangutan test

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2849,89 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+        }
+
+        if (remainingRouters.size() < 2) {
+            return;
+        }
+
+        VirtualRouter backupRouter = null;
+        for (final VirtualRouter router : remainingRouters) {
+            if (router.getRedundantState() == VirtualRouter.RedundantState.BACKUP) {
+                backupRouter = router;
```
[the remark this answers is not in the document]
rafaelweingartner: Are these routers ordered some how? I mean, reading your next discussion with @DaanHoogland, it seems that the router we do not need will always be the last one (the third or fourth one) when this code is executed. Therefore, there must be some way to order this list before you reach here, right?

rhtyd:
The last test results (xenserver) shows no new errors caused by this PR, and other tests are incoming soon. Given we've two LGTMs on it from @DaanHoogland and @GabrielBrascher , we may merge this but this PR as soon as test pass BUT since this adds a new operational enhancement I would like to request - @rafaelweingartner @resmo @nvazquez @wido @ustcweizhou @mike-tutkowski @fmaximus @jayapalu @marcaurele @PaulAngus @sureshanaparti @terbolous @NuxRo for their feedback, and I'm happy to move this to next 4.11.2.0 milestone if we've consensus aorund this. Please respond by the end of this week, next week will focus only on blockers and start work towards pre-RC1 testing. Thanks.

(Please follow links for packages and systemvmtemplate in this PR if you're not looking for building it yourself)

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
rafaelweingartner: Javadocs are useful to keep the history behind these things. Keep in mind that the method is only public (java delimiter access); this means accessible to programmers to re-use its code. It is not something that end-users will have access to.

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2868,6 +2849,89 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
...
+        if (backupRouter == null) {
+            backupRouter = routers.get(routers.size() - 1);
+        }
+        if (backupRouter != null) {
+            _routerService.destroyRouter(backupRouter.getId(), context.getAccount(), context.getCaller().getId());
+        }
+    }
+
+    @Override
+    public boolean validateNewRouters(final List<? extends VirtualRouter> routers) {
```
rafaelweingartner: Instead of validate, what about `areNewRoutersrunning`?
It is not a validation method per se. It does not throw exceptions. This method is only checking if all routers of the giving list are in running state.

server/src/com/cloud/network/element/VirtualRouterElement.java
```diff
@@ -227,7 +227,7 @@ public boolean implement(final Network network, final NetworkOffering offering,
         final List<DomainRouterVO> routers = routerDeploymentDefinition.deployVirtualRouter();
 
         int routerCounts = 1;
```
rafaelweingartner: Instead of `routerCounts`, what about `expectedNumberRouter` or something similar. This is the number of VRs you expect to have in a network, right?

server/src/com/cloud/network/vpc/VpcManagerImpl.java
```diff
@@ -2435,4 +2444,43 @@ public boolean isSrcNatIpRequired(long vpcOfferingId) {
...
+
+        final DeployDestination dest = new DeployDestination(_dcDao.findById(vpc.getZoneId()), null, null, null);
+        final List<DomainRouterVO> oldRouters = _routerDao.listByVpcId(vpc.getId());
+
+        // Create a new router
+        if (oldRouters.size() > 0) {
+            vpc.setRollingRestart(true);
+        }
+        startVpc(vpc, dest, context);
+        if (oldRouters.size() > 0) {
```
rafaelweingartner: Can you explain this logic here? I mean, it is the same as the one at line 2456, but when it solves to true there, we set `vpc.setRollingRestart(true);` and then we set `vpc.setRollingRestart(false);` here. Is this something that was done only to satisfy some conditional at `startVpc`?

rafaelweingartner:
@rhtyd I have just reviewed the code. 

I tested it yesterday and everything was fine. I had only one problem, but that was an environment one. One of my hosts was not working properly.

api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
rhtyd: @rafaelweingartner when the option is not passed, the value should be false which is what the fix it.

api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
rhtyd: @rafaelweingartner when the option is not passed, the value should be false which is what the fix it.

server/src/com/cloud/network/element/VirtualRouterElement.java
rhtyd: @rafaelweingartner good idea, maybe for a next/different refactoring PR.

server/src/com/cloud/network/vpc/VpcManagerImpl.java
rhtyd: @rafaelweingartner if you look again I did refactor the code to `rollingRestartVpc` I'll stop at that. The lines of code in the `if` block are only 5.

api/src/org/apache/cloudstack/api/command/user/vpc/RestartVPCCmd.java
rafaelweingartner: This means that the default is `false`, right?
Can't you declare the variable as `private boolean cleanup = false`?

server/src/com/cloud/network/vpc/VpcManagerImpl.java
rafaelweingartner: I noticed it. This comment is the one I posted before you refactored the code. Github maintained the comment even though you have changed part of this code.

server/src/com/cloud/network/vpc/VpcManagerImpl.java
rhtyd: @rafaelweingartner if there are any existing old routers, we do rolling restart which is to deploy a new VR and then destroy old VRs. The `setRollingRestart` is a way to tell lower facade to treat deployment like for redundant routers, i.e. expect upto 2 VRs at a given point which is seen in changes in `VirtualRouterElement.java`. The startVpc will call the facade to deploy suitable number of routers. If there is only one old VPC VR, and if rolling restart is `true` this will cause the lower layers to deploy a new VPC VR. After that, we set the rolling restart to false which works great both in case of rVR (expected VRs is 2) and non-rVRs (expected VRs becomes 1), and finally we destroy old VR. If the VPC has rVRs, a new backup router is created when `startVpc` is called otherwise this will simply reconfigure existing VRs.

server/src/com/cloud/network/element/VirtualRouterElement.java
rhtyd: Fixed.

rhtyd:
Thanks @rafaelweingartner and @DaanHoogland  I've tried to address all the review comments, let me know if I've missed anything. Wrt to testing, existing lifecycle tests in VPC, network and routers related smoketests validate the restart operation.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
XenServer and Vmware tests are LGTM, 1-2 failure seen in intermittent/env related. I've rebase and fixed conflicts against latest 4.11, will kick test against KVM after packaging.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Tests LGTM, we've enough reviews to merge this. If there are no objections, this will be merged by end of the day tomorrow.

PR #2510, opened 2018-03-23
CLOUDSTACK-10334: Fix inadequate information for handling catch clauses

lzh3636:
## Description
The description of the problem:
https://issues.apache.org/jira/browse/CLOUDSTACK-10334
I just added stack traces information to those two logging statements, so that the exception type can be generated to the logs.
## Types of changes

## Checklist:

### DIFF
--- a/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
+++ b/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
@@ -258,13 +258,11 @@ public class ParamProcessWorker implements DispatchWorker {
                 }
 
             } catch (final IllegalArgumentException e) {
-                s_logger.error("Error initializing command " + cmd.getCommandName() + ", field " + field.getName() + " is not accessible.", e);
                 throw new CloudRuntimeException("Internal error initializing parameters for command " + cmd.getCommandName() + " [field " + field.getName() +
-                        " is not accessible]");
+                        " is not accessible]", e);
             } catch (final IllegalAccessException e) {
-                s_logger.error("Error initializing command " + cmd.getCommandName() + ", field " + field.getName() + " is not accessible.", e);
                 throw new CloudRuntimeException("Internal error initializing parameters for command " + cmd.getCommandName() + " [field " + field.getName() +
-                        " is not accessible]");
+                        " is not accessible]", e);
             }
 
         }
--- a/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
+++ b/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
@@ -258,11 +258,11 @@ public class ParamProcessWorker implements DispatchWorker {
                 }
 
             } catch (final IllegalArgumentException e) {
-                s_logger.error("Error initializing command " + cmd.getCommandName() + ", field " + field.getName() + " is not accessible.");
+                s_logger.error("Error initializing command " + cmd.getCommandName() + ", field " + field.getName() + " is not accessible.", e);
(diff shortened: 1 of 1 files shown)


rhtyd:
The change is simply in error logging, Travis is green so explicit trillian tests may not be necessary. +1 on tests based on Travis.

server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
```diff
@@ -258,11 +258,11 @@ public void processParameters(final BaseCmd cmd, final Map params) {
                 }
 
             } catch (final IllegalArgumentException e) {
-                s_logger.error("Error initializing command " + cmd.getCommandName() + ", field " + field.getName() + " is not accessible.");
+                s_logger.error("Error initializing command " + cmd.getCommandName() + ", field " + field.getName() + " is not accessible.", e);
```
rafaelweingartner: Why do you need to log the stack trace twice? I mean, the exception is re-thrown as a cloudruntime. Wherever it is catch, it will probably be logged.

In the Jira ticket you mention that the problem (confusion) is caused by the same message in the stack trace. Why not improve the messaging then, instead of logging printing the strack trace twice?

lzh3636: Thanks for your reply. The exception is thrown by a new exception cloudruntime, but the original exception variable "e" is lost when throwing the new exception, so the logs will not show if the exception type is IllegalArgumentException or IllegalAccessException.

It's true that it will be better to improve the log message, but I'm not so sure if I can change the message correctly. So it's a safe way to add original stack trace information here, at least we can know the true exception types from the logs generated without any harm.

nitin-maharana:
Code LGTM.

server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
DaanHoogland: @rafaelweingartner are you satisfied with this answer?

rafaelweingartner: Well, that was an answer. I would prefer to fix/improve the log messaging than to log twice the same stack trace. What do you think?

DaanHoogland: I do not see that point, sorry. An unexpected error is caught and logged (not rethrown) and than a more generic message/exception is thrown. Other than unifying the two catch clauses in a catch (IllegalArgumentException|IllegalAccessException e), I don't see the harm here. We could improve by not logging at all and adding the original exception as a root cause to the CloudRuntimeException. Is that what you mean?

rafaelweingartner: That is what I am trying to say :)

DaanHoogland: ok, so there is a misunderstanding here. The original exception is not rethrown, but a new one @lzh3636 . The question is can you add it to the throw and delete the error log instead?

lzh3636: Yes, thank you, it's done :)

rafaelweingartner:
@DaanHoogland sure.
No need to run the integration tests here?

DaanHoogland:
not sure @rafaelweingartner , should we?
@blueorangutan package

rhtyd:
@DaanHoogland @rafaelweingartner  this is simply a logging change, Travis is green. I'm okay merging based on 2+ lgtms and test run by Travis.

rafaelweingartner:
@DaanHoogland I am ok too. I will merge then 👍

PR #2519, opened 2018-04-02
CLOUDSTACK-10287 package require open jdk 1.8

DaanHoogland:
## Types of changes

## Checklist:

- [ ] All new and existing tests passed.

@blueorangutan package

### DIFF
--- a/packaging/centos7/cloud.spec
+++ b/packaging/centos7/cloud.spec
@@ -58,7 +58,7 @@ intelligent IaaS cloud implementation.
 
 %package management
 Summary:   CloudStack management server UI
-Requires: java => 1.8.0
+Requires: java-1.8.0-openjdk
 Requires: apache-commons-daemon-jsvc
 Requires: python
 Requires: bash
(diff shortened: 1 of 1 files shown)


packaging/centos7/cloud.spec
```diff
@@ -58,7 +58,7 @@ intelligent IaaS cloud implementation.
 
 %package management
 Summary:   CloudStack management server UI
-Requires: java => 1.8.0
```
rafaelweingartner: Sorry if this is a silly question, but the `requires` notation needs a full package name? Is that why you needed to change it?

DaanHoogland: no, the jdk is need and not just the jre

rafaelweingartner: Ah sure. 
Would it be possible to use `Requires: javac >= 1.8.0`?

DaanHoogland: will look into that

rhtyd: No, only jre is needed but on centos java-1.8.0-openjdk is the jre pkg and java-1.8.0-openjdk-devel has the jdk.

PR #2562, opened 2018-04-11
consoleproxy: use consoleproxy.domain for non-ssl enable env

rhtyd:
This allows CloudStack to use a console proxy domain instead of public
IP address even when ssl is not enabled but console proxy url/domain
is defined in global settings.

This allows use of domain url for console proxy endpoint, without needing to have
SSL certificates setup.

## Types of changes

## How Has This Been Tested?

I was trying to setup a lab for testing CloudStack, this lab is in my home. To make it
publicly accessible, I've a DO machine to which one of my machines does autossh
and port forwards to it. The DO machine has apache2 running that reverse proxies
to only certain paths on the forwarded port. The DO endpoint is secured by letsencrypt
certs. With this change, a user/admin will be able to run console proxy via a URL/domain
but on http instead of https internally. Externally, such as my setup, one can secure
urls/paths using a single frontend machine. The apache2 reverse proxy setup I'm using is
where the ip of the CPVM is 192.168.1.51 (home network/internal)
```
        # Console proxy
        ProxyPass /ajax http://192.168.1.51/ajax
        ProxyPassReverse /ajax http://192.168.1.51/ajax
        ProxyPass /ajaximg http://192.168.1.51/ajaximg
        ProxyPassReverse /ajaximg http://192.168.1.51/ajaximg
        ProxyPass /resource http://192.168.1.51/resource
        ProxyPassReverse /resource http://192.168.1.51/resource
```

screenshot from 2018-04-11 22-56-22

The following shows my home lab setup:
untitled diagram

## Checklist:

Testing
- [x] A full integration testsuite with all test that can run on my environment has passed.

@blueorangutan package

### DIFF
--- a/core/src/com/cloud/info/ConsoleProxyInfo.java
+++ b/core/src/com/cloud/info/ConsoleProxyInfo.java
@@ -19,6 +19,8 @@
 
 package com.cloud.info;
 
+import org.apache.commons.lang3.StringUtils;
+
 public class ConsoleProxyInfo {
 
     private boolean sslEnabled;
@@ -55,6 +57,9 @@ public class ConsoleProxyInfo {
                 proxyImageUrl += ":" + this.proxyUrlPort;
         } else {
             proxyAddress = proxyIpAddress;
+            if (StringUtils.isNotBlank(consoleProxyUrlDomain)) {
+                proxyAddress = consoleProxyUrlDomain;
+            }
             proxyPort = port;
             this.proxyUrlPort = proxyUrlPort;
 
--- a/server/src/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
+++ b/server/src/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
@@ -231,7 +231,7 @@ public class ConsoleProxyManagerImpl extends ManagerBase implements ConsoleProxy
     private String _instance;
 
     private int _proxySessionTimeoutValue = DEFAULT_PROXY_SESSION_TIMEOUT;
-    private boolean _sslEnabled = true;
+    private boolean _sslEnabled = false;
     private String _consoleProxyUrlDomain;
 
     // global load picture at zone basis
@@ -1246,8 +1246,7 @@ public class ConsoleProxyManagerImpl extends ManagerBase implements ConsoleProxy
 
         Map<String, String> configs = _configDao.getConfiguration("management-server", params);
 
-        String value = configs.get(Config.ConsoleProxyCmdPort.key());
-        value = configs.get("consoleproxy.sslEnabled");
+        String value = configs.get("consoleproxy.sslEnabled");
         if (value != null && value.equalsIgnoreCase("true")) {
             _sslEnabled = true;
         }


rhtyd:
@blueorangutan test

core/src/com/cloud/info/ConsoleProxyInfo.java
```diff
@@ -55,6 +57,9 @@ public ConsoleProxyInfo(boolean sslEnabled, String proxyIpAddress, int port, int
                 proxyImageUrl += ":" + this.proxyUrlPort;
         } else {
             proxyAddress = proxyIpAddress;
+            if (!Strings.isNullOrEmpty(consoleProxyUrlDomain)) {
```
rafaelweingartner: you can use directly `org.apache.commons.lang3.StringUtils.isNotBlank(CharSequence)`, then you do not need the `!`. It also catches "blanks" values, but I am not sure if they are possible here.

server/src/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
```diff
@@ -1246,8 +1246,7 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
 
         Map<String, String> configs = _configDao.getConfiguration("management-server", params);
 
-        String value = configs.get(Config.ConsoleProxyCmdPort.key());
-        value = configs.get("consoleproxy.sslEnabled");
+        String value = configs.get("consoleproxy.sslEnabled");
         if (value != null && value.equalsIgnoreCase("true")) {
```
rafaelweingartner: You can use `org.apache.commons.lang3.BooleanUtils.toBoolean(String)`.

core/src/com/cloud/info/ConsoleProxyInfo.java
rhtyd: @rafaelweingartner I think there is a divide in preference to use commons.lang or guava, we already have a lot of code that uses guava and things like `Strings.isNullOrEmpty`. I prefer using the guava utility/methods, also StringUtils.isNotBlank internally would do more or less the same kind of checks. I'm okay to change, but then we need a java style/doc on which methods to prefer in which case we can deprecate the use of one style over the other.

rafaelweingartner: No problem, but jut to be clear, that method (`Strings.isNullOrEmpty`) does not address the blank case.

rhtyd: @rafaelweingartner I'm not sure but last time I checked, the `isNullOrEmpty` did check for emptiness (blank) of string (https://google.github.io/guava/releases/snapshot-jre/api/docs/com/google/common/base/Strings.html#isNullOrEmpty-java.lang.String-)

rafaelweingartner: Well the java dos they just mention empty, and not blank. I checked their code, and they do not check for blank.

Test code:
```
    public static void main(String[] args) {
        System.out.println("Using Guava Strings suite");
        System.out.println("Result for empty: " + Strings.isNullOrEmpty(""));
        System.out.println("Result for null " + Strings.isNullOrEmpty(null));
        System.out.println("Result for blank: " + Strings.isNullOrEmpty("       "));
        System.out.println();
        System.out.println("Using Apache commns lang");
        System.out.println("Result for empty: " + StringUtils.isBlank(""));
        System.out.println("Result for null " + StringUtils.isBlank(null));
        System.out.println("Result for blank: " + StringUtils.isBlank("       "));
    }
```

Result:
```
Using Guava Strings suite
Result for empty: true
Result for null true
Result for blank: false

Using Apache commns lang
Result for empty: true
Result for null true
Result for blank: true
```

server/src/com/cloud/consoleproxy/ConsoleProxyManagerImpl.java
rhtyd: @rafaelweingartner I checked this is a string comparison, the 'true' has to do with the string stored in global setting's value from db. Looks like a massive cleanup may be needed in this class and others. I'll avoid changes in that case, keep the changes in PR minimal and maybe send a separate cleanup PR.

core/src/com/cloud/info/ConsoleProxyInfo.java
rhtyd: Thanks for testing and sharing @rafaelweingartner used `StringUtils` now!

rhtyd:
We can wait for Travis to go green and then this PR may be merged based on reviews and test results.

DennisKonrad:
Hi @rhtyd,

so how do you get the https in front of your consoleproxy.url.domain without setting consoleproxy.sslEnabled to true?

If I try what you do here I always get an http url from cloudstack management because I have not set sslEnabled. Therefor the browser warns me because it tries to load the http address of the consoleproxy after it got https from cloudstack management.

Thanks in advance

rhtyd:
@DennisKonrad yes the official way is to set the sslEnabled setting to true. In my setup, I wrote a url rewrite rule in apache2 to automatically redirect the consoleproxy.url.domain to an https:// endpoint of the same domain. Both my mgmt server and console proxy end points were behing an apache2 server that I use for ssl offloading.

PR #2572, opened 2018-04-13
Remove 'todb' in favor of 'encodeURIComponent'.

rafaelweingartner:
## Description

While executing the find/replace, I found some blocks of duplicated code. Therefore, I extracted the duplicated part to an utils file, and then removed the duplicated blocks.

## Types of changes

## How Has This Been Tested?
Locally

## Checklist:

Testing

@blueorangutan package

### DIFF
--- a/ui/scripts/accountsWizard.js
+++ b/ui/scripts/accountsWizard.js
@@ -209,9 +209,7 @@
                 array1.push("&firstname=" + args.data.firstname);
                 array1.push("&lastname=" + args.data.lastname);
 
-                var password = args.data.password;
-                password = todb(password);
-                array1.push("&password=" + password);
+                cloudStack.addPasswordToCommandUrlParameterArray(array1, args.data.password);
             }
 
             array1.push("&domainid=" + args.data.domainid);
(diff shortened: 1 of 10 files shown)


DaanHoogland:
Why remove todb for the longer encodeURIComponent, when the first is only a wrapper for the latter?

rafaelweingartner:
Good question.  Why not use directly a function that everybody knows (everybody that works with javascript)? 

When I first encountered it, I spent some time trying to find its documentation in the Javascript specs. Just then, I discovered that it was a wrapper created in CloudStack's javascript. This wrapper does not help us with anything. Moreover, it was not used consistently in the code base. There are quite a lot of places using the 'encodeURIComponent'.

If we want to write less, we should first stop copying and pasting code ;)

DaanHoogland:
right, but having a wrapper around an external piece of code is always good. Per our older discussions; let's reduce the surface layer of 3rd party software to ours.

DaanHoogland:
btw: look good in general it is just that I would have opted for giving a better name to todb and use the wrapper

rafaelweingartner:
The method `encodeURIComponent` is not a third party library. It is a native function in Javascript.

DaanHoogland:
ah, in that case I stand corrected, @rafaelweingartner. once again looks good

PR #2580, opened 2018-04-18
CLOUDSTACK-10357 Improve log messages that do not match with their method function

lzh3636:
## Description

There are some possible copy and paste errors in the log messages (The logging statement was copied from an old place to a new place, but the message wasn't changed to adapt to the function of new place).

Here's a list of this kind of problems I found, which contains the original place that the logging statement possibly be copied from, and the new place that the log be copied to (the log message is not suitable for the new place and should be changed):
https://issues.apache.org/jira/browse/CLOUDSTACK-10357

## Types of changes

## Checklist:

Testing

@blueorangutan package

### DIFF
--- a/plugins/dedicated-resources/src/main/java/org/apache/cloudstack/dedicated/DedicatedResourceManagerImpl.java
+++ b/plugins/dedicated-resources/src/main/java/org/apache/cloudstack/dedicated/DedicatedResourceManagerImpl.java
@@ -497,7 +497,7 @@ public class DedicatedResourceManagerImpl implements DedicatedService {
                     }
                     dedicatedResource = _dedicatedDao.persist(dedicatedResource);
                 } catch (Exception e) {
-                    s_logger.error("Unable to dedicate host due to " + e.getMessage(), e);
+                    s_logger.error("Unable to dedicate cluster due to " + e.getMessage(), e);
                     throw new CloudRuntimeException("Failed to dedicate cluster. Please contact Cloud Support.");
                 }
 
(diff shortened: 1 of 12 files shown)


plugins/dedicated-resources/src/main/java/org/apache/cloudstack/dedicated/DedicatedResourceManagerImpl.java
```diff
@@ -497,7 +497,7 @@ public boolean configure(final String name, final Map<String, Object> params) th
                     }
                     dedicatedResource = _dedicatedDao.persist(dedicatedResource);
                 } catch (Exception e) {
-                    s_logger.error("Unable to dedicate host due to " + e.getMessage(), e);
+                    s_logger.error("Unable to dedicate cluster due to " + e.getMessage(), e);
                     throw new CloudRuntimeException("Failed to dedicate cluster. Please contact Cloud Support.");
```
rafaelweingartner: What about passing the `e` to the `CloudRuntimeException`?

plugins/hypervisors/baremetal/src/main/java/org/apache/cloudstack/api/DeleteBaremetalRctCmd.java
```diff
@@ -64,7 +64,7 @@ public void execute() throws ResourceUnavailableException, InsufficientCapacityE
             SuccessResponse response = new SuccessResponse(getCommandName());
             setResponseObject(response);
         } catch (Exception e) {
-            s_logger.warn(String.format("unable to add baremetal RCT[%s]", getId()), e);
+            s_logger.warn(String.format("unable to delete baremetal RCT[%s]", getId()), e);
             throw new ServerApiException(ApiErrorCode.INTERNAL_ERROR, e.getMessage());
```
rafaelweingartner: The same thing here regarding the passing of exception to the new one that is thrown.

lzh3636: Sure, I'll do one more commit

PR #2582, opened 2018-04-19
cs-45to411-ugrade-fix: database errors during upgrade to 4.11

ernjvr:
## Description

This fixes a database error that occurs when upgrading from 4.5 to 4.11. A 'table already exists' SQL error is encountered during the upgrade.
If a table already exists in the database, it should only attempt to create a table if it does not exist.
The fix adds SQL syntax create a table only if not exists and also removes the drop table statements since there may be users that already have such a table with data in their database.

Expected behaviour: when upgrading from an older version of cloudstack, it should upgrade the database without producing any SQL errors. Specifically, the SQL upgrade scripts should not attempt to create a table if it already exists. It should also not attempt to drop existing tables unless it is in the cleanup scripts. 
Actual behaviour: a database error that occurs when upgrading from 4.5 to 4.11. A 'table already exists' SQL error is encountered during the upgrade.
## Types of changes

## How Has This Been Tested?

This was manually tested by installing a 4.5 version of cloudstack environment and then did the upgrade to version 4.11.1.0.
## Checklist:

Testing

@blueorangutan package

### DIFF
--- a/engine/schema/resources/META-INF/db/schema-41000to41100.sql
+++ b/engine/schema/resources/META-INF/db/schema-41000to41100.sql
@@ -392,7 +392,7 @@ CREATE VIEW `cloud`.`volume_view` AS
         `cloud`.`domain` resource_tag_domain ON resource_tag_domain.id = resource_tags.domain_id;
 
 -- Extra Dhcp Options
-CREATE TABLE  IF NOT EXISTS `cloud`.`nic_extra_dhcp_options` (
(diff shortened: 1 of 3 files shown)


rhtyd:
Routing regression test:
@blueorangutan test

ernjvr:
@rhtyd should I revert the removal of the DROP TABLE statement for usage_snapshot_on_primary?

ernjvr:
@DaanHoogland @rhtyd can this be merged?

PR #2584, opened 2018-04-19
Enhance and cleanup DatabaseUpgradeChecker

khos2ow:
## Description

Using a hierarchy of database version rather than a flat
list of them.

## Types of changes

## Checklist:

Testing

@blueorangutan package


khos2ow:
@rafaelweingartner Adding a new schema upgrade path was really cumbersome and error-prone, because we needed to maintain a flat map of versions and their corresponding list of upgrade paths (`DbUpgrade`). Instead I'm proposing to have a logical hierarchy structure of versions:

```
DatabaseVersionHierarchy.builder()
    .next("4.0.0"   , new Upgrade40to41())
    .next("4.0.1"   , new Upgrade40to41())
    .next("4.0.2"   , new Upgrade40to41())
    .next("4.1.0"   , new Upgrade410to420())
    .next("4.1.1"   , new Upgrade410to420())
    .next("4.2.0"   , new Upgrade420to421())
    ...
    .next("4.2.1"   , new Upgrade421to430())
    .next("4.9.3.0" , new Upgrade4930to41000())
    .next("4.10.0.0", new Upgrade41000to41100())
    .next("4.11.0.0", new Upgrade41100to41110())
    .build();
```

With this approach, when we need to add a new version upgrade path, rather than add that (e.g. `Upgrade41200to41300`) in **dozens** of places, we only need to add it in *correct* place in the hierarchy. It will be even backward compatible, it means that we can inject a version in the middle of the hierarchy as well (e.g. between existing `4.9.3.0` AND `4.10.0.0`).

As for your next question, originally I had it based on `master` but I figured it's good to have it on `4.11` forward, so I rebased it on 4.11 and waited for this exact discussion to mention what you mentioned. I even created another branch to target `master` with exact same fix. So we have two options, either apply this (with additional path) on `master` first, then apply this on `4.11` and when merge forward pretty much nothing will happen. Or apply this here and add extra path on master on another super-small PR.

BTW there's a small logical problem which will be fixed in couple minutes! (that's why Travis fails now)

rafaelweingartner:
Thanks @khos2ow. This is a great proposal to simplify the upgrade path code.
You already have another PR for master, that is what I call being proactive!

I see no problem in opening this PR to 4.11. I asked because it could cause more trouble for you during the merge forward. I mean, it could cause trouble for the committer, which would, in turn, ask for your help. 

I got your solution idea. However, I still have one doubt. Sorry, if this is clear in the code, but I normally prefer to understand what I am dealing with before coding/reviewing. There is one thing you mention that called my attention:
> it means that we can inject a version in the middle of the hierarchy as well (e.g. between existing 4.9.3.0 AND 4.10.0.0

Using this PR, what scripts would be executed in the following scenario?
* ACS most recent version is 4.11.0.0
* There are old, but still maintained version 4.10 and 4.9
* A user is in version 4.9.3.0
* Let’s say that we create a version 4.9.4.0 that executes some script in the database. This change is also pushed forward and is then released in version 4.11.1.0.
* If the user upgrades to 4.9.4.0, he/she will get the change, and then it is clear the path to 4.11.1.0
* If the user upgrades directly to 4.11.1.0, it is also clear the upgrade path
* What happens if the user goes to 4.10.0.0 (from 4.9.3.0)? Assuming that 4.10 was released before 4.9.4.0 (then, it does not have the new scripts). When the user goes from 4.10 to 4.11.1. Will the scripts from 4.9.4.0 be executed?

khos2ow:
I don't have the PR per se, I just have another branch on my fork to open a PR if need be :)

These changes will be required for the use case you mentioned:

- this snippet needs to be added in `4.9` branch (as previous):
```
_upgradeMap.put(CloudStackVersion.parse("4.9.3.0"), new DbUpgrade[] {new Upgrade4930to4940()});
```
`new Upgrade4930to4940()` must be added to **all** other versions in `_updateMap` as well.
- this snippet needs to be added in `4.10` branch (as previous):
```
_upgradeMap.put(CloudStackVersion.parse("4.9.3.0"), new DbUpgrade[] {new Upgrade4930to4940(), new Upgrade4940to41000()});
_upgradeMap.put(CloudStackVersion.parse("4.9.4.0"), new DbUpgrade[] {new Upgrade4940to41000()});
```
`new Upgrade4930to4940()` must be added to **all** other versions in `_updateMap` as well.

**Whereas** in `4.11` and `master` will be as follow:

- snippet on for new approach:
```
DatabaseVersionHierarchy.builder()
    .next("4.0.0"   , new Upgrade40to41())
    ...
    .next("4.9.3.0" , new Upgrade4930to4940()) <-- modified to include 4.9.4.0 (original value: Upgrade4930to41000)
    .next("4.9.4.0" , new Upgrade4940to41000()) <-- added to go include 4.9.4.0
    .next("4.10.0.0", new Upgrade41000to41100())
    .next("4.11.0.0", new Upgrade41100to41110())
    .build();
```

rafaelweingartner:
Interesting. So, you will track the scripts already executed, and if there is some "new" script that appears (even though it is in a previous version than the current one) you will execute it.

khos2ow:
It's not that I'm *tracking* the changes, rather than when upgrade procedure executed it will walk through the hierarchy and will generate `DbUpgrade[]`. So if a new item is added *anywhere* in the hierarchy (as long it doesn't have a duplicate `getUpgradedVersion()` value) it will be queued for execution.

rafaelweingartner:
Hmm. So, it will walk the hierarchy and build the array of DbUpgrade to be executed. How do you decide the version to start? I mean, if I am in version 4.10, what DbUpgrade will be the first one? 4.10.0.0 to 4.11.0.0?

khos2ow:
Correct. That part hasn't changed. It finds the `from` and `to` in `DatabaseUpgradeChecker#check()`. Basically from is the latest row in `version` table, and `to` is the current released version you're installing.

DaanHoogland:
great work @khos2ow , i have a minor concern, some paths have actually changed, i noticed. Doesn't have to be bad as it is unifying paths in the end.

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
```diff
@@ -0,0 +1,185 @@
...
+        final DbUpgrade upgrader;
+
+        private VersionNode(final CloudStackVersion version, final DbUpgrade upgrader) {
+            this.version = version;
+            this.upgrader = upgrader;
+        }
+    }
+
+    public static final class DatabaseVersionHierarchyBuilder {
+        private final List<VersionNode> _hierarchyBuilder = new LinkedList<>();
```
rafaelweingartner: You do not need to use these "_" anymore.

Is there a reason for you to use `LinkedList` here?

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
```diff
@@ -0,0 +1,185 @@
...
+     * @return true if hierarchy contains the version, false if not
+     */
+    public boolean contains(final CloudStackVersion version) {
+        return toList().contains(version);
+    }
+
+    /**
+     * Calculates an  upgrade path for the  passed <code>fromVersion</code>.  If the  <code>fromVersion</code>
+     * doesn't exist in list  of  available <code>VersionNode</code> hierarchy,  then calculation assumes that
+     * the <code>fromVersion</code> required no schema migrations or data conversions and no  upgrade path was
```
rafaelweingartner: Isn't these documentation duplicated?

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
```diff
@@ -0,0 +1,185 @@
...
+                .orElse(null);
+    }
+
+    /**
+     * Generate immutable list of available <code>CloudstackVersion</code> in the hierarchy
+     *
+     * @return list of available versions
+     */
+    public ImmutableList<CloudStackVersion> toList() {
+        List<CloudStackVersion> versions = hierarchy
```
rafaelweingartner: You for sure like Lambda and Streams :)

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
```diff
@@ -0,0 +1,185 @@
...
+                                                .collect(Collectors.toList());
+
+        return ImmutableList.copyOf(versions);
+    }
+
+    /**
+     * Find the distinct <code>VersionNode</code> based on the provided <code>getUpgradedVersion()</code>
+     */
+    private Predicate<VersionNode> distinct(Function<VersionNode, String> keyExtractor) {
+        Map<String, Boolean> seen = new ConcurrentHashMap<>();
```
rafaelweingartner: Do we need to use a synchronized hashmap here?

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
khos2ow: Good point. There are very similar, but one only accepts `fromVersion` the other one accepts both `fromVersion` and `toVersion`. I will update the second one to have more information about `toVersion` there.

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
khos2ow: They are really useful, specially that it shortens LoC a lot! :)

khos2ow:
@DaanHoogland which ones for example? I noticed a bunch (on `2.x.x` and `3.x.x` which I labeled them as legacy), but I'm not concerned about those because the schema creation starts from `4.0.0` anyway.

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
khos2ow: Technically no, but I used it because of the nature of stream which can be run in parallel, and to ensure it will work no matter what.

rafaelweingartner: That is what I thought, but i wanted to make sure you knew it.
Thanks for the explanations.

engine/schema/src/com/cloud/upgrade/DatabaseVersionHierarchy.java
rafaelweingartner: Yes, they are useful, but I still have not adopted them. I have some doubts regarding performance and readability with time. It is not a problem in this context that you are using them though.

khos2ow: It has a getting used to side on it. But once you got used to it, it's super fun. Specially `FunctionalInterface`, `Consumer`, `Supplier` and friends.
Using lambda doesn't have any performance impact at all, but stream and specially filters might, if misused. (they are essentially a loop with time and space complexity around them anyway)

engine/schema/src/com/cloud/upgrade/DatabaseUpgradeChecker.java
```diff
@@ -93,377 +92,90 @@
...
+                .next("4.5.3"   , new Upgrade453to460())
+                .next("4.6.0"   , new Upgrade460to461())
+                .next("4.6.1"   , new Upgrade461to470())
+                .next("4.6.2"   , new Upgrade461to470())
+                .next("4.7.0"   , new Upgrade470to471())
+                .next("4.7.1"   , new Upgrade471to480())
+                .next("4.7.2"   , new Upgrade471to480())
+                .next("4.8.0"   , new Upgrade480to481())
+                .next("4.8.1"   , new Upgrade481to490())
+                .next("4.8.2.0" , new Upgrade481to490())
```
nitin-maharana: @khos2ow, Thanks for the PR. I am wondering why both 4.8.1 and 4.8.2.0 are doing the same upgrade from 481to490.

engine/schema/src/com/cloud/upgrade/DatabaseUpgradeChecker.java
```diff
@@ -51,11 +51,11 @@
 import com.cloud.upgrade.dao.Upgrade440to441;
 import com.cloud.upgrade.dao.Upgrade441to442;
 import com.cloud.upgrade.dao.Upgrade442to450;
-import com.cloud.upgrade.dao.Upgrade443to450;
+import com.cloud.upgrade.dao.Upgrade443to444;
```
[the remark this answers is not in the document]
khos2ow: I understand the concern, but it's not an altered upgrade path per se, rather than the previous logical hierarchy was messed up and now it's in correct logical order.

`443to450` is replaced by `443to444` and the hierarchy is:
- `443to444` (Empty `schema-443to444.sql`. There's no `schema-443to444-cleanup.sql`) 
- `444to450` (Doesn't have anything in it and it extends `Upgrade442to450`)
- `443to450` itself is empty and only extends `Upgrade442to450` which hierarchy is already aware of.

And `452to460` is replaced by `452to453` and the hierarchy is:
- `452to453` (Empty `schema-452to453.sql` and `schema-452to453-cleanup.sql`) 
- `453to460` (Doesn't have anything in it and it extends `Upgrade452to460`)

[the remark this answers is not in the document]
khos2ow: I did two tests and start management server to see if normal stuff works: 
- Install on brand new database
- Upgrade to 4.12 from 4.10

khos2ow:
@rhtyd The way I see it, there won't be a risk on 4.11, because it essentially executes the same upgrade path. But if that's the general concerns, I'm ok to rebase it on master.

khos2ow:
@rhtyd @rafaelweingartner @DaanHoogland @marcaurele @nitin-maharana any suggestions? Leave it on `4.11` or move it on `4.12`?

rafaelweingartner:
For, it is indifferent. I have not seen a standard guiding us to decide what goes into an already released (minor version) version or not. If it passed the testes, and reviews. We can accept it in 4.11.

rhtyd:
@khos2ow in my opinion, you may also consider extending @ustcweizhou (Wei's) PR around fixing db upgrades using flyway? This can lead to a discussion around deprecating our homegrown db migration framework in favour of flyway or something else, so I would prefer 4.12 to get such changes.

khos2ow:
Fair enough, then I'm gonna rebase this on master and move forward with this PR until we have a good and working third-party based solution.

khos2ow:
Rebased on master.

khos2ow:
@rhtyd I finally had a chance to look at @ustcweizhou 's PR (#2479), it looks really promising, and the good thing  about that PR and this, is that they're completing each other, so we can move forward with both of them.
Any upgrade path since the beginning to one cut-off version (e.g. `4.11.1.0`) is going to be handled the *old* way (this PR), and from that cut-off version forward with *new* flyway.

rafaelweingartner:
@khos2ow can you rebase?

@rhtyd and @DaanHoogland I am inclined to merge this one. What do you guys think? Travis test should be enough to validate this one. Or do you guys think we need integration tests as well here?

khos2ow:
Rebased and applied code formatting on separate commit.
@rafaelweingartner @rhtyd Let me know if I should squash them or you will do it on merging.

rafaelweingartner:
It is not a matter of changing the technology only. Even if we change to flyway, we need to change the way we manage scripts.

There is a long discussion that we had about this issue some time ago: https://markmail.org/message/2dxtcnqx2euq5tl2

rhtyd:
@blueorangutan test

khos2ow:
Why do you think it's going to break @rhtyd ? This PR is a cleanup and to change a `Map` data structure to  essentially a `LinkedList`. On the other hand it should actually prevent breaking (because we wouldn't need to add to dozens of places when new upgrade path required).

@rafaelweingartner Thanks for providing the thread, I'll take a look later. And as I mentioned earlier the way I thought about this is, to agree on a cut-off version to migrate to Flyway. That's also the way implemented in #2479. Everything up to that version is going to be handled with the _old_ way, and afterward with Flyway.

I tested:
- fresh install of 4.12.0.0
- upgrade from 4.11.0.0
- upgrade from 4.10.0.0
- upgrade from 4.9.0

I added a unit test for _unknown_ version to upgrade from and to. But the rest of unit tests look good enough. May be we can add another one from older versio (e.g. 4.6-ish) to some new version (e.g. 4.11-ish), but I don't think it would add much value.

marcaurele:
@khos2ow @rhtyd @rafaelweingartner There was also this discussion about how the current state of DB migration was buggy, and to fix it (merged) https://github.com/apache/cloudstack/pull/1768 - I don't think it will break things. In my opinion it will only improve the current state by lifting the need to be very verbose, and error prone when adding a new upgrade path.

rhtyd:
@khos2ow fair enough, my concern was that the resultant linked-list and map (old way) should return the same list -- I was looking for such a test for all version keys (from the old map).

rhtyd:
Tests LGTM, some of the failures are intermitten, rest are known and recently fixed.

PR #2593, opened 2018-04-23
api: move default response from XML to JSON

marcaurele:
## Description
Moving the default response type (when not provided) from XML to JSON.

Java community used to love XML a lot, but let's face it, nowadays people are using JSON whenever they can.

## Types of changes

## Checklist:

- [x] I have updated the documentation accordingly. (https://github.com/apache/cloudstack-docs/pull/25)
Testing

### DIFF
--- a/server/src/main/java/com/cloud/api/ApiServlet.java
+++ b/server/src/main/java/com/cloud/api/ApiServlet.java
@@ -140,9 +140,9 @@ public class ApiServlet extends HttpServlet {
             s_logger.warn("UnknownHostException when trying to lookup remote IP-Address. This should never happen. Blocking request.", e);
             final String response = apiServer.getSerializedApiError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR,
                     "UnknownHostException when trying to lookup remote IP-Address", null,
-                    HttpUtils.RESPONSE_TYPE_XML);
+                    HttpUtils.RESPONSE_TYPE_JSON);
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/cloud/api/ApiServlet.java
```diff
@@ -170,12 +170,6 @@ void processRequestInContext(final HttpServletRequest req, final HttpServletResp
 
         try {
 
-            if (HttpUtils.RESPONSE_TYPE_JSON.equalsIgnoreCase(responseType)) {
```
rhtyd: I think we should keep this, otherwise people won't be able to get XML output if a response type was provided.

marcaurele: It's some dummy code. The content type is set when writing the String response to the HttpResponse later on. Doing it so early does not have much sense. See `HttpUtils.writeHttpResponse()`

wido:
I agree with @rafaelweingartner 

This would be a rather big change and I think we should discuss it properly. I understand the change, but we will break a lot of things.

Do I hear somebody saying CloudStack 5.0?

DaanHoogland:
let's do the 5.0 dance

rafaelweingartner:
As we discussed in CCC Montreal, this would fit our requirements for breaking changes.

rohityadavcloud:
It will be nearly 3+ years old PR and can't be accepted which would break our default API response compatibility. Pl re-open this PR if this can be reworked again latest main. Thank you for the PR.

PR #2595, opened 2018-04-23
CLOUDSTACK-10199: Support requesting a specific IPv4 address

GabrielBrascher:
This pull request allows deploying VMs with a specific IPv4 address.

DirectPodBasedNetworkGuru does not support requesting a custom IP-Address while creating a new NIC/Instance, throwing the following error:
> Error 530: Does not support custom ip allocation at this time: NicProfile[0-0-null-null-null
>     Unknown macro: { "cserrorcode"} 

Some use-cases prefer the ability to request the IPv4 address which the Instance will get.

This implementation adds unit test cases to cover and it was manually tested in Basic Networking. I can perform more tests if requested.

## Types of changes

## Checklist:

Testing

### DIFF
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -231,27 +231,27 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
     @Inject
     EntityManager _entityMgr;
     @Inject
-    DataCenterDao _dcDao = null;
+    DataCenterDao _dcDao;
     @Inject
-    VlanDao _vlanDao = null;
+    VlanDao _vlanDao;
     @Inject
-    IPAddressDao _ipAddressDao = null;
+    IPAddressDao _ipAddressDao;
     @Inject
-    AccountDao _accountDao = null;
+    AccountDao _accountDao;
     @Inject
     ConfigurationDao _configDao;
     @Inject
-    UserVmDao _userVmDao = null;
+    UserVmDao _userVmDao;
     @Inject
     AlertManager _alertMgr;
     @Inject
     ConfigurationManager _configMgr;
     @Inject
-    NetworkOfferingDao _networkOfferingDao = null;
+    NetworkOfferingDao _networkOfferingDao;
     @Inject
-    NetworkDao _networksDao = null;
+    NetworkDao _networksDao;
     @Inject
-    NicDao _nicDao = null;
+    NicDao _nicDao;
     @Inject
     RulesManager _rulesMgr;
     @Inject
@@ -860,6 +860,11 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
 
         NicVO vo = new NicVO(guru.getName(), vm.getId(), network.getId(), vm.getType());
 
+        DataCenterVO dcVo = _dcDao.findById(network.getDataCenterId());
+        if (dcVo.getNetworkType() == NetworkType.Basic) {
+            configureNicProfileBasedOnRequestedIp(requested, profile, network);
+        }
+
         deviceId = applyProfileToNic(vo, profile, deviceId);
 
         vo = _nicDao.persist(vo);
@@ -869,6 +874,85 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
                 _networkModel.getNetworkTag(vm.getHypervisorType(), network));
 
         return new Pair<NicProfile, Integer>(vmNic, Integer.valueOf(deviceId));
+    }
+
+    /**
+     * If the requested IPv4 address from the NicProfile was configured then it configures the IPv4 address, Netmask and Gateway to deploy the VM with the requested IP.
+     */
+    protected void configureNicProfileBasedOnRequestedIp(NicProfile requestedNicProfile, NicProfile nicProfile, Network network) {
+        if (requestedNicProfile == null) {
+            return;
+        }
+        String requestedIpv4Address = requestedNicProfile.getRequestedIPv4();
+        if (requestedIpv4Address == null) {
+            return;
+        }
+        if (!NetUtils.isValidIp4(requestedIpv4Address)) {
+            throw new InvalidParameterValueException(String.format("The requested [IPv4 address='%s'] is not a valid IP address", requestedIpv4Address));
+        }
+
+        VlanVO vlanVo = _vlanDao.findByNetworkIdAndIpv4(network.getId(), requestedIpv4Address);
+        if (vlanVo == null) {
+            throw new InvalidParameterValueException(String.format("Trying to configure a Nic with the requested [IPv4='%s'] but cannot find a Vlan for the [network id='%s']",
+                    requestedIpv4Address, network.getId()));
+        }
+
+        String ipv4Gateway = vlanVo.getVlanGateway();
+        String ipv4Netmask = vlanVo.getVlanNetmask();
+
+        if (!NetUtils.isValidIp4(ipv4Gateway)) {
+            throw new InvalidParameterValueException(String.format("The [IPv4Gateway='%s'] from [VlanId='%s'] is not valid", ipv4Gateway, vlanVo.getId()));
+        }
+        if (!NetUtils.isValidIp4Netmask(ipv4Netmask)) {
+            throw new InvalidParameterValueException(String.format("The [IPv4Netmask='%s'] from [VlanId='%s'] is not valid", ipv4Netmask, vlanVo.getId()));
+        }
+
+        acquireLockAndCheckIfIpv4IsFree(network, requestedIpv4Address);
+
+        nicProfile.setIPv4Address(requestedIpv4Address);
+        nicProfile.setIPv4Gateway(ipv4Gateway);
+        nicProfile.setIPv4Netmask(ipv4Netmask);
+
+        if (nicProfile.getMacAddress() == null) {
+            try {
+                String macAddress = _networkModel.getNextAvailableMacAddressInNetwork(network.getId());
+                nicProfile.setMacAddress(macAddress);
+            } catch (InsufficientAddressCapacityException e) {
+                throw new CloudRuntimeException(String.format("Cannot get next available mac address in [network id='%s']", network.getId()), e);
+            }
+        }
+    }
+
+    /**
+     *  Acquires lock in "user_ip_address" and checks if the requested IPv4 address is Free.
+     */
+    protected void acquireLockAndCheckIfIpv4IsFree(Network network, String requestedIpv4Address) {
+        IPAddressVO ipVO = _ipAddressDao.findByIpAndSourceNetworkId(network.getId(), requestedIpv4Address);
+        if (ipVO == null) {
+            throw new InvalidParameterValueException(
+                    String.format("Cannot find IPAddressVO for guest [IPv4 address='%s'] and [network id='%s']", requestedIpv4Address, network.getId()));
+        }
+        try {
+            IPAddressVO lockedIpVO = _ipAddressDao.acquireInLockTable(ipVO.getId());
+            validateLockedRequestedIp(ipVO, lockedIpVO);
+            lockedIpVO.setState(IPAddressVO.State.Allocated);
+            _ipAddressDao.update(lockedIpVO.getId(), lockedIpVO);
+        } finally {
+            _ipAddressDao.releaseFromLockTable(ipVO.getId());
+        }
+    }
+
+    /**
+     * Validates the locked IP, throwing an exeption if the locked IP is null or the locked IP is not in 'Free' state.
+     */
+    protected void validateLockedRequestedIp(IPAddressVO ipVO, IPAddressVO lockedIpVO) {
+        if (lockedIpVO == null) {
+            throw new InvalidParameterValueException(String.format("Cannot acquire guest [IPv4 address='%s'] as it was removed while acquiring lock", ipVO.getAddress()));
+        }
+        if (lockedIpVO.getState() != IPAddressVO.State.Free) {
+            throw new InvalidParameterValueException(
+                    String.format("Cannot acquire guest [IPv4 address='%s']; The Ip address is in [state='%s']", ipVO.getAddress(), lockedIpVO.getState().toString()));
+        }
     }
 
     protected Integer applyProfileToNic(final NicVO vo, final NicProfile profile, Integer deviceId) {
--- a/engine/orchestration/src/test/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestratorTest.java
+++ b/engine/orchestration/src/test/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestratorTest.java
@@ -23,31 +23,42 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Arrays;
 
-import com.cloud.exception.InvalidParameterValueException;
-import com.cloud.offerings.NetworkOfferingVO;
 import org.apache.log4j.Logger;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.JUnit4;
 import org.mockito.Matchers;
+import org.mockito.Mockito;
 
+import com.cloud.dc.Vlan;
+import com.cloud.dc.VlanVO;
+import com.cloud.dc.dao.VlanDao;
+import com.cloud.exception.InsufficientAddressCapacityException;
+import com.cloud.exception.InvalidParameterValueException;
 import com.cloud.network.Network;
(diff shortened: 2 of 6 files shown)


engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,73 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
...
+
+    /**
+     *  Acquires lock of in table "user_ip_address" and checks if the requested IPv4 address is Free.
+     */
+    protected void acquireLockAndCheckIfIpv4IsFree(Network network, String requestedIpv4Address) {
+        IPAddressVO ipVO = _ipAddressDao.findByIpAndSourceNetworkId(network.getId(), requestedIpv4Address);
+        if (ipVO == null) {
+            throw new CloudRuntimeException(String.format("Cannot find IPAddressVO for guest [IPv4 address='%s'] and [network id=%s]", requestedIpv4Address, network.getId()));
+        }
+        IPAddressVO lockedIpVO = _ipAddressDao.acquireInLockTable(ipVO.getId());
```
marcaurele: `lockedIpVO` can be `null` (https://github.com/apache/cloudstack/blob/master/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java#L1049). Moreover, it would be better IMO to put the lock/release operation with a `try/finally` block to ensure to release the lock all the time (when it could get it).

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,73 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
...
+        if (lockedIpVO.getState() != IPAddressVO.State.Free) {
+            String ipState = lockedIpVO.getState().toString();
```
marcaurele: You don't need to have a variable holding the state, you can use `lockedIpVo.getState()` directly in the `String.format()` function, its value won't change after releasing the lock on the VO.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,73 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
...
+        nicProfile.setIPv4Address(requestedIpv4Address);
+        nicProfile.setIPv4Gateway(ipv4Gateway);
+        nicProfile.setIPv4Netmask(ipv4Netmask);
+
+        if (nicProfile.getMacAddress() == null) {
+            try {
+                String macAddress = _networkModel.getNextAvailableMacAddressInNetwork(network.getId());
+                nicProfile.setMacAddress(macAddress);
+            } catch (InsufficientAddressCapacityException e) {
+                e.printStackTrace();
```
marcaurele: Require a logger here and to throw an exception

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,73 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
         return new Pair<NicProfile, Integer>(vmNic, Integer.valueOf(deviceId));
     }
 
+    /**
+     * If the requested IPv4 address from the NicProfile was configured then it configures the IPv4 address, Netmask and Gateway to deploy the VM with the requested IP. 
+     */
+    protected void configureNicProfileBasedOnRequestedIp(NicProfile requestedNicProfile, NicProfile nicProfile, Network network) {
+        String requestedIpv4Address = requestedNicProfile.getRequestedIPv4();
+        if (requestedIpv4Address == null) {
+            s_logger.debug(String.format("The requested [IPv4 address=%s] is null", requestedIpv4Address));
```
marcaurele: Could be simplified since `requestedIpv4Address` is always null to avoid parameter.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,73 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
         return new Pair<NicProfile, Integer>(vmNic, Integer.valueOf(deviceId));
     }
 
+    /**
+     * If the requested IPv4 address from the NicProfile was configured then it configures the IPv4 address, Netmask and Gateway to deploy the VM with the requested IP. 
+     */
+    protected void configureNicProfileBasedOnRequestedIp(NicProfile requestedNicProfile, NicProfile nicProfile, Network network) {
+        String requestedIpv4Address = requestedNicProfile.getRequestedIPv4();
+        if (requestedIpv4Address == null) {
```
rafaelweingartner: You can extract lines 893-902 to a specific method that validates `requestedIpv4Address `

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,73 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
...
+            throw new InvalidParameterValueException(String.format("The requested [IPv4 address='%s'] is not a valid IP address", requestedIpv4Address));
+        }
+
+        VlanVO vlanVo = _vlanDao.findByNetworkId(network.getId());
+        if (vlanVo == null) {
+            throw new CloudRuntimeException(String.format("Trying to configure a Nic with the requested [IPv4='%s'] but cannot find a Vlan for the [network id='%s']",
+                    requestedIpv4Address, network.getId()));
+        }
+
+        acquireLockAndCheckIfIpv4IsFree(network, requestedIpv4Address);
```
rafaelweingartner: Should not you validate the `vlanVo` before trying to acquire the IP?
I mean, you are executing the validation on `vlanVo` afterwards.

engine/schema/src/main/java/com/cloud/dc/dao/VlanDaoImpl.java
```diff
@@ -81,6 +81,22 @@ public VlanVO findByZoneAndVlanId(long zoneId, String vlanId) {
...
+    
+    /**
+     * Returns a vlan by the network id.
+     */
+    @Override
+    public VlanVO findByNetworkId(long networkId) {
+        List<VlanVO> vlanVoList = listVlansByNetworkId(networkId);
+        VlanVO vlanVo = null;
+        for (VlanVO vlan : vlanVoList) {
+            if (vlan.getRemoved() == null) {
```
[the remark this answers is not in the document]
rafaelweingartner: I have the same doubt. It looks like a logic that should be somewhere else. If the network can use multiple `Vlan`, then should not you look for the correct one (the one that is in the same CIDR) as the requested IPv4?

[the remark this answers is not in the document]
GabrielBrascher: Well noticed, thanks :)

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
nitin-maharana: @GabrielBrascher, @rafaelweingartner, I think the last if condition covers both null and blank check. (isValidIp4 method)

rhtyd:
In advanced zone, I can see the API allows acceptance of `ipaddress` or `ip6address` parameters. @GabrielBrascher how do I reproduce the issue (env details please)?

GabrielBrascher:
@rhtyd I am using basic network and KVM. It might be a problem only on the basic network then.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
GabrielBrascher: Thanks, you are right @nitin-maharana; however, the null check should exist to return without an exception and consider that the user did not request a specific IPv4.

engine/orchestration/src/test/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestratorTest.java
```diff
@@ -159,4 +174,255 @@ public void testDontRemoveDhcpServiceWhenNotProvided() {
...
+        Network network = mock(Network.class);
+        NicProfile requestedNicProfile = new NicProfile();
+        NicProfile nicProfile = Mockito.spy(new NicProfile());
+
+        configureTestConfigureNicProfileBasedOnRequestedIpTests(nicProfile, 0l, false, IPAddressVO.State.Free, "192.168.100.1", "255.255.255.0", "00-88-14-4D-4C-FB",
+                requestedNicProfile, null, requestedIpv4Address);
+        try {
+            testOrchastrator.configureNicProfileBasedOnRequestedIp(requestedNicProfile, nicProfile, network);
+        } catch (InvalidParameterValueException e) {
+            assertEquals(exceptionAssert, e.getMessage());
```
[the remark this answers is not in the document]
GabrielBrascher: I always use @Test(expected=Exception); however, it is not working here now, I will check again what might be causing it.

engine/schema/src/main/java/com/cloud/dc/dao/VlanDaoImpl.java
```diff
@@ -82,6 +83,24 @@ public VlanVO findByZoneAndVlanId(long zoneId, String vlanId) {
...
+     */
+    @Override
+    public VlanVO findByNetworkIdAndIpv4Range(long networkId, String ipv4Address) {
+        List<VlanVO> vlanVoList = listVlansByNetworkId(networkId);
+        for (VlanVO vlan : vlanVoList) {
+            String ipRange = vlan.getIpRange();
+            String[] ipRangeParts = ipRange.split("-");
+            String startIP = ipRangeParts[0];
+            String endIP = ipRangeParts[1];
+            if (NetUtils.isIpInRange(ipv4Address, startIP, endIP)) {
```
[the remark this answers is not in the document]
GabrielBrascher: The problem is that I do not have a CIDR. The network CIDR (at least in the basic network) is null. I like the idea of isIp6InNetwork, but I will need to ignore CIDR at this stage and use only ipv4 range.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,78 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
...
+        }
+        String requestedIpv4Address = requestedNicProfile.getRequestedIPv4();
+        if (requestedIpv4Address == null) {
+            return;
+        }
+        if (!NetUtils.isValidIp4(requestedIpv4Address)) {
+            throw new InvalidParameterValueException(String.format("The requested [IPv4 address='%s'] is not a valid IP address", requestedIpv4Address));
+        }
+
+        VlanVO vlanVo = _vlanDao.findByNetworkIdAndIpv4Range(network.getId(), requestedIpv4Address);
```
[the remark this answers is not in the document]
GabrielBrascher: Makes sense, I will change it. Thanks!

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -883,6 +885,78 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
...
+            } catch (InsufficientAddressCapacityException e) {
+                throw new CloudRuntimeException(String.format("Cannot get next available mac address in [network id='%s']", network.getId()), e);
+            }
+        }
+    }
+
+    /**
+     *  Acquires lock of in table "user_ip_address" and checks if the requested IPv4 address is Free.
+     */
+    protected void acquireLockAndCheckIfIpv4IsFree(Network network, String requestedIpv4Address) {
```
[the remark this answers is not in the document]
GabrielBrascher: Yes, that makes sense. I will update it.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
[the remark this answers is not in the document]
DaanHoogland: convention says to leave out the vlan (we are already in the vlandao. so findByNetworkIdAndIpv4Range actually makes sense.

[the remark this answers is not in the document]
GabrielBrascher: I can update it to `findByNetworkIdAndIpv4` then. What do you guys think of it?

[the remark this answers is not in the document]
rafaelweingartner: It makes sense. 👍

engine/orchestration/src/test/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestratorTest.java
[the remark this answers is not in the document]
GabrielBrascher: Fixed.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -869,6 +871,76 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
...
```
fmaximus: Will this check work for requested IP addresses in case of Advanced networking?
I believe if the network is of type Guest, then you won't be able to find an IPAddressVO.
So the method should only be called if the network is public or shared.

GabrielBrascher: Thanks, @fmaximus. I will check on that.

rafaelweingartner: @fmaximus cannot users define a guest IP? I mean, as long as the guest IP defined by the user is within the bounds of the guest network, it should be possible, right?

GabrielBrascher: @fmaximus thanks for pointing it out and sorry for the delay. I did some testing in advanced network and confirmed that a few modifications in this PR will be necessary. When I finish them I will ping you.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -860,6 +860,11 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
 
         NicVO vo = new NicVO(guru.getName(), vm.getId(), network.getId(), vm.getType());
 
+        DataCenterVO dcVo = _dcDao.findById(network.getDataCenterId());
+        if (dcVo.getNetworkType() == NetworkType.Basic) {
+            configureNicProfileBasedOnRequestedIp(requested, profile, network);
```
GabrielBrascher: @fmaximus I changed the execution flow; now this workflow will be executed just in case of a basic network.

After this change I could add IPv4 either in an advanced network, with a 'quest IP', and in a basic network with a 'public IP' (from the IPAddressVO in this case). Thanks again for pointing that issue.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rafaelweingartner:
@GabrielBrascher are the erros presented here caused by your changes?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

GabrielBrascher:
@rafaelweingartner I have checked and could not find any point indicating that the errors presented here are caused by my changes.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
GabrielBrascher: @marcaurele are you LGTM on this one?
Thanks for your reviews and hard work :+1:

DaanHoogland:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@DaanHoogland @nvazquez @rhtyd @borisstoyanov All ok after packaging :1st_place_medal: 
Can you please run Integration test on this one? Thanks in advance :+1:

nvazquez:
Sure @GabrielBrascher 
@blueorangutan test matrix

PR #2598, opened 2018-04-24
CLOUDSTACK-10360: Change the method name.

BruceKuiLiu:
The method is named as "scoped" that seems to whether the variable config is scoped in _scopedStorages or not.
Actually, the method tries to find a storage of which scope equals to the scope of config.
So that, the method name "findStorage" should be more clear than "scoped".

### DIFF
--- a/framework/config/src/main/java/org/apache/cloudstack/framework/config/impl/ConfigDepotImpl.java
+++ b/framework/config/src/main/java/org/apache/cloudstack/framework/config/impl/ConfigDepotImpl.java
@@ -166,7 +166,7 @@ public class ConfigDepotImpl implements ConfigDepot, ConfigDepotAdmin {
         return _configDao;
     }
 
-    public ScopedConfigStorage findStorage(ConfigKey<?> config) {
+    public ScopedConfigStorage findScopedConfigStorage(ConfigKey<?> config) {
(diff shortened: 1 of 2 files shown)


framework/config/src/main/java/org/apache/cloudstack/framework/config/impl/ConfigDepotImpl.java
```diff
@@ -166,7 +166,7 @@ public ConfigurationDao global() {
         return _configDao;
     }
 
-    public ScopedConfigStorage scoped(ConfigKey<?> config) {
+    public ScopedConfigStorage findStorage(ConfigKey<?> config) {
```
rafaelweingartner: I would use something like `findScopedConfigStorage`. This `findStorage` can be confused with find storage pool.

BruceKuiLiu: Thanks, "findScopedConfigStorage" is more clear.

DaanHoogland:
I like your take on radable code @BruceKuiLiu though I didn't agree with all your PRs so far, please keep it up

DaanHoogland:
@rhtyd @rafaelweingartner @wido @mike-tutkowski @anyone, do we need testing on this change?

PR #2599, opened 2018-04-24
CLOUDSTACK-10363: Change method names.

BruceKuiLiu:
…findXXX".

These three methods are not direct getter or list.
They try to find the target objects with the related arguments.
So that, renaming them as "findXXX" should be more intuitive.


nitin-maharana:
Hi @BruceKuiLiu, Thanks for the change, now the method names are more meaningful. Would be nice if you club all the similar PRs to one. (#2600, #2601, #2602)

BruceKuiLiu:
Hi @nitin-maharana ,
I think that #2600, #2061, #2602 can be classified into three different changes.
If it is necessary, I will merge them into one commit.

nitin-maharana:
I think it would be good to merge them into one. Let's wait for others view.

BruceKuiLiu:
@rafaelweingartner Thanks.

rhtyd:
This is a pure name refactoring with no effective overall logic/code changed. Merging this based on Travis pass and reviews by others.

PR #2601, opened 2018-04-24
CLOUDSTACK-10364: Change the "setXXX" method names to "getXXX".

BruceKuiLiu:
The two methods are named as "setXXX",
actually, they are not setter but getter.
Thus, they should be renamed as "getXXX".

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
@@ -298,11 +298,11 @@ public class BridgeVifDriver extends VifDriverBase {
         Script.runSimpleBashScript("test -d /sys/class/net/" + iface.getBrName() + "/brif/" + iface.getDevName() + " && brctl delif " + iface.getBrName() + " " + iface.getDevName());
     }
 
-    private String setVnetBrName(String pifName, String vnetId) {
+    private String generateVnetBrName(String pifName, String vnetId) {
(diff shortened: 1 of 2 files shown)


plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
```diff
@@ -298,11 +298,11 @@ public void detach(LibvirtVMDef.InterfaceDef iface) {
         Script.runSimpleBashScript("test -d /sys/class/net/" + iface.getBrName() + "/brif/" + iface.getDevName() + " && brctl delif " + iface.getBrName() + " " + iface.getDevName());
     }
 
-    private String setVnetBrName(String pifName, String vnetId) {
+    private String getVnetBrName(String pifName, String vnetId) {
```
DaanHoogland: i dont think getVnetBrName() is a good name, what it does is format/generate/create a vnetBrName! not get it from somewhere.

rhtyd:
@DaanHoogland you comment about `get` have been addressed, the author has used a `generate` method name prefix that sounds acceptable (generating a name from params).
This is a pure method name change, build and Travis have passed. I'll merge this based on these tests and reviews.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
ShangwenWang: Yes, that is to say, we may also name these methods as createXXX.
Right?

PR #2602, opened 2018-04-24
CLOUDSTACK-10365: Change the "getXXX" boolean-related method names to…

BruceKuiLiu:
… "isXXX".

These boolean-return methods are named as "getXXX".
Other boolean-return methods are named as "isXXX".
Considering there methods will return boolean values, it should be more clear and consistent to rename them as "isXXX".

### DIFF
--- a/api/src/main/java/com/cloud/agent/api/to/S3TO.java
+++ b/api/src/main/java/com/cloud/agent/api/to/S3TO.java
@@ -201,7 +201,7 @@ public final class S3TO implements ClientOptions, DataStoreTO {
         return DataStoreRole.Image;
     }
 
-    public boolean getEnableRRS() {
+    public boolean isEnableRRS() {
         return enableRRS;
     }
 
--- a/api/src/main/java/com/cloud/offering/ServiceOffering.java
+++ b/api/src/main/java/com/cloud/offering/ServiceOffering.java
@@ -76,7 +76,7 @@ public interface ServiceOffering extends DiskOffering, InfrastructureEntity, Int
     /**
      * @return Does this service plan offer HA?
      */
-    boolean getOfferHA();
+    boolean isOfferHA();
 
     /**
      * @return Does this service plan offer VM to use CPU resources beyond the service offering limits?
@@ -86,7 +86,7 @@ public interface ServiceOffering extends DiskOffering, InfrastructureEntity, Int
     /**
      * @return Does this service plan support Volatile VM that is, discard VM's root disk and create a new one on reboot?
      */
-    boolean getVolatileVm();
+    boolean isVolatileVm();
 
     /**
      * @return the rate in megabits per sec to which a VM's network interface is throttled to
--- a/api/src/main/java/com/cloud/offering/NetworkOffering.java
+++ b/api/src/main/java/com/cloud/offering/NetworkOffering.java
@@ -122,7 +122,7 @@ public interface NetworkOffering extends InfrastructureEntity, InternalIdentity,
 
     boolean getIsPersistent();
 
-    boolean getInternalLb();
+    boolean isInternalLb();
 
     boolean getPublicLb();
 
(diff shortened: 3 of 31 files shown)


api/src/main/java/com/cloud/agent/api/to/S3TO.java
```diff
@@ -201,7 +201,7 @@ public DataStoreRole getRole() {
         return DataStoreRole.Image;
     }
 
-    public boolean getEnableRRS() {
+    public boolean isEnableRRS() {
```
DaanHoogland: what is the is prefix for in this case? is the question enableRRS() ?

api/src/main/java/com/cloud/offering/ServiceOffering.java
```diff
@@ -76,7 +76,7 @@
     /**
      * @return Does this service plan offer HA?
      */
-    boolean getOfferHA();
+    boolean isOfferHA();
```
DaanHoogland: the java doc implies it should be called doesOfferHA()

api/src/main/java/com/cloud/offering/ServiceOffering.java
```diff
@@ -86,7 +86,7 @@
     /**
      * @return Does this service plan support Volatile VM that is, discard VM's root disk and create a new one on reboot?
      */
-    boolean getVolatileVm();
+    boolean isVolatileVm();
```
DaanHoogland: supportVolitileVMs() would be more pleasing on the human eye. Not that that should hamper this change but what is the objective? to adhere to another naming convention or to make better code for reading? and seriously both are valid motives under circumstances, but I am really wondering.

BruceKuiLiu: The intention of my pull requests is to make code more readable and understandable, thus further making developers to understand the code more easily with method names without more context information. 
But about method naming things, developers have two different opinions as you mentioned:
(1) keep the same naming convention as similar methods.
(2) make code more readable.

DaanHoogland: @BruceKuiLiu I am with 2 all the way, with the exception of the use of these methods in code generation tools. In this last method I would expect a name like isForVolatileVms() or the ealier sugested name. Of cause the accessor is named after the field implying that the field is not properly named either.
I am +0 on this one. Try to find more reviews.

api/src/main/java/com/cloud/offering/ServiceOffering.java
fmaximus: @DaanHoogland I would prefer to only have methods starting with either get or is.
Any other method name in a VO would not be usable for filtering,
according to the implementation of the VO MethodInterceptor

DaanHoogland: good point @fmaximus so the generic dao breaks if we go any other way is your message, is it?

rafaelweingartner: Do you know the name of the class that does this interception, so we can inspect it?
Is it using method names or POJO's properties directly via their annotation?

api/src/main/java/com/cloud/agent/api/to/S3TO.java
rafaelweingartner: Does not `enableRSS` give the idea that it is a method that "enables" RSS?
At least, that is what sounds to me.

DaanHoogland:
@BruceKuiLiu are you still with us? your branch is out off date, can you rebase?

BruceKuiLiu:
@DaanHoogland Yes. I removed all my branches. I need to create a new pull request.

DaanHoogland:
That's fine. at your own time ;)

BruceKuiLiu:
@DaanHoogland #2816

DaanHoogland:
closing in favour of #2816

PR #2606, opened 2018-04-25
When creating a new account (via domain admin) it is possible to select “root admin” as the role for the new user

rafaelweingartner:
## Description

Domain admins should not be able to assign the role of root admin to new users. Therefore, the role ‘root admin’ (or any other of the same type) should not be visible to domain admin accounts.

## Types of changes

## How Has This Been Tested?
I created unit test that cover all of the modified code. Moreover, I tested locally.

## Checklist:

Testing

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/acl/RoleService.java
+++ b/api/src/main/java/org/apache/cloudstack/acl/RoleService.java
@@ -25,30 +25,51 @@ import java.util.List;
 public interface RoleService {
 
     ConfigKey<Boolean> EnableDynamicApiChecker = new ConfigKey<>("Advanced", Boolean.class, "dynamic.apichecker.enabled", "false",
-            "If set to true, this enables the dynamic role-based api access checker and disables the default static role-based api access checker.",
-            true);
+            "If set to true, this enables the dynamic role-based api access checker and disables the default static role-based api access checker.", true);
 
     boolean isEnabled();
-    Role findRole(final Long id);
+
+    /**
+     *  Searches for a role with the given ID. If the ID is null or less than zero, this method will return null.
+     *  This method will also return null if no role is found with the provided ID. Moreover, we will check if the requested role is of 'Admin' type; roles with 'Admin' type should only be visible to 'root admins'.  Therefore, if a non-'root admin' user tries to search for an 'Admin' role, this method will return null.
+     */
+    Role findRole(Long id);
+
     Role createRole(final String name, final RoleType roleType, final String description);
+
     Role updateRole(final Role role, final String name, final RoleType roleType, final String description);
+
     boolean deleteRole(final Role role);
 
     RolePermission findRolePermission(final Long id);
+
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/apache/cloudstack/acl/RoleService.java
```diff
@@ -17,38 +17,62 @@
...
-    Role findRole(final Long id);
-    Role createRole(final String name, final RoleType roleType, final String description);
-    Role updateRole(final Role role, final String name, final RoleType roleType, final String description);
-    boolean deleteRole(final Role role);
 
-    RolePermission findRolePermission(final Long id);
-    RolePermission findRolePermissionByUuid(final String uuid);
+    /**
+     *  Searches for a role with the given ID. If the ID is null or less than zero, this method will return null.
+     *  This method will also return null if no role is found with the provided ID. Moreover, we will check if the requested role is of 'Admin' type; roles with 'Admin' type should only be visible to 'root admins'.  Therefore, if a non-'root admin' user tries to search for an 'Admin' role, this method will return null.
```
nitin-maharana: @rafaelweingartner, Thanks for the fix. Would be more appropriate if we break this line into two.

rafaelweingartner: Done.

DaanHoogland:
looks good, I will run the tests though.
@blueorangutan package

server/src/test/java/org/apache/cloudstack/acl/RoleManagerImplTest.java
```diff
@@ -0,0 +1,288 @@
...
+    public void beforeTest() {
+        Mockito.doReturn(accountMockId).when(accountMock).getId();
+        Mockito.doReturn(accountMock).when(roleManagerImpl).getCurrentAccount();
+
+        Mockito.doReturn(roleMockId).when(roleVoMock).getId();
+    }
+
+    @Test
+    public void findRoleTestIdNull() {
+        Role returedRole = roleManagerImpl.findRole(null);
```
nitin-maharana: TYPO. I think you mean to be returnedRole( r -> n ). Next seven tests from here.

rafaelweingartner: Fixed, thanks!

DaanHoogland:
@blueorangutan test

rafaelweingartner:
In the errors reported I found the following message `Exception: Host communication is not as expected: true. Instead, it's: false`. It looks like some action was taking longer than expected, and that is what causes the error. However, this PR does not change code that can have an effect on host status. Therefore, this is looking like an environment problem.

rhtyd:
@rafaelweingartner do you want to consider this security/bug to be in 4.11? If yes, please edit the base branch of the PR and rebase branch, otherwise we can squash merge this PR.

PR #2486, opened 2018-03-14
[CLOUDSTACK-10323] Allow changing disk offering during volume migration 

rafaelweingartner:
This is a continuation of work developed on PR #2425 (CLOUDSTACK-10240), which provided root admins an override mechanism to move volumes between storage systems types (local/shared) even when the disk offering would not allow such operation. To complete the work, we will now provide a way for administrators to enter a new disk offering that can reflect the new placement of the volume. We will add an extra parameter to allow the root admin inform a new disk offering for the volume. Therefore, when the volume is being migrated, it will be possible to replace the disk offering to reflect the new placement of the volume.

The API method will have the following parameters:

* storageid (required)
* volumeid (required)
* livemigrate(optional)
* newdiskofferingid (optional) – this is the new parameter

The expected behavior is the following:

* If “newdiskofferingid” is not provided the current behavior is maintained. Override mechanism will also keep working as we have seen so far.
* If the “newdiskofferingid” is provided by the admin, we will execute the following checks
    * new disk offering mode (local/shared) must match the target storage mode. If it does not match, an exception will be thrown and the operator will receive a message indicating the problem.
    * we will check if the new disk offering tags match the target storage tags. If it does not match, an exception will be thrown and the operator will receive a message indicating the problem.
    * check if the target storage has the capacity for the new volume. If it does not have enough space, then an exception is thrown and the operator will receive a message indicating the problem.
    * check if the size of the volume is the same as the size of the new disk offering. If it is not the same, we will ALLOW the change of the service offering, and a warning message will be logged.

We execute the change of the Disk offering as soon as the migration of the volume finishes. Therefore, if an error happens during the migration and the volume remains in the original storage system, the disk offering will keep reflecting this situation.

Attached is a screenshot to show how the migrate option will look like:
changeintheui

PR #2612, opened 2018-04-27
[migrateVolume API method] Filter disk offerings based on target storage pool selected

rafaelweingartner:
## Description

After using the feature introduced by #2486 in production, we felt the need for an improvement in the UI.  It is interesting to filter the displayed disk offerings according to the type of storage selected (local/shared) to migrate the volume to.

## Types of changes

## How Has This Been Tested?

Locally

## Checklist:

Testing

### DIFF
--- a/ui/scripts/sharedFunctions.js
+++ b/ui/scripts/sharedFunctions.js
@@ -2851,3 +2851,34 @@ cloudStack.createArrayOfParametersForCreatePodCommand = function (zoneId, data){
     cloudStack.addParameterToCommandUrlParameterArrayIfValueIsNotEmpty(array, "endIp", data.podEndIp);
     return array;
 }
+
+cloudStack.listDiskOfferings = function(options){
+    var defaultOptions = {
+            listAll: false,
+            isRecursive: false,
+            error: function(data) {
+                args.response.error(data);
+            }
+    };
+    var mergedOptions = $.extend({}, defaultOptions, options);
+    
+    var listDiskOfferingsUrl = "listDiskOfferings";
+    if(mergedOptions.listAll){
+        listDiskOfferingsUrl = listDiskOfferingsUrl + "&listall=true";
+    }
+    if(mergedOptions.isRecursive){
+        listDiskOfferingsUrl = listDiskOfferingsUrl + "&isrecursive=true";
+    }
+    var diskOfferings = undefined;
+    $.ajax({
+        url: createURL(listDiskOfferingsUrl),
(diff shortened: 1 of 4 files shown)


ui/scripts/sharedFunctions.js
```diff
@@ -2851,3 +2851,34 @@ cloudStack.createArrayOfParametersForCreatePodCommand = function (zoneId, data){
     cloudStack.addParameterToCommandUrlParameterArrayIfValueIsNotEmpty(array, "endIp", data.podEndIp);
     return array;
 }
+
+cloudStack.listDiskOfferings = function(options){
```
nitin-maharana: Hi @rafaelweingartner, This is a great improvement. I came across several places where the same AJAX call(Code) is repeated multiple places. We should put more of this to reduce redundancy. Thanks.

ui/scripts/sharedFunctions.js
```diff
@@ -2851,3 +2851,34 @@ cloudStack.createArrayOfParametersForCreatePodCommand = function (zoneId, data){
...
+    if(mergedOptions.listAll){
+        listDiskOfferingsUrl = listDiskOfferingsUrl + "&listall=true";
+    }
+    if(mergedOptions.isRecursive){
+        listDiskOfferingsUrl = listDiskOfferingsUrl + "&isrecursive=true";
+    }
+    var diskOfferings = undefined;
+    $.ajax({
+        url: createURL(listDiskOfferingsUrl),
+        data: mergedOptions.data,
```
nitin-maharana: For some calls, we don't pass 'data' in the parameter list. For e.g. [var diskOfferings = cloudStack.listDiskOfferings({listAll: true});] 

I think, it would be better if we check the existance of the parameter before assigning.

rafaelweingartner: I am not checking the existence of `mergedOptions.data` before ahe ssignment because the result would be the same.  I mean, if `mergedOptions.data` is `undefined`, than the object `data` was not provided. Therefore, the object `data` in the `ajax` function will have the value `undefined`. If I add an `IF` condition and decide to assign `data`  based on the existence of `mergedOptions.data`; then, when `mergedOptions.data == undefined` (I would not assign it), the `data` object of `ajax` would also be undefined. The result is the same, and thus there is no reason to add extra code.

Was my explanation clear?

nitin-maharana: @rafaelweingartner, Got it. Thanks.

PR #2627, opened 2018-05-04
Catch error in packagin script and fail the build

khos2ow:
## Description

This PR is going to:
- fail the build early if `--use-timestamp` is provided and working directory is *not* clean.
- catch any potential error and fail the build (fix the regression added in 1708838518 (#2433))

## Types of changes

## Checklist:

Testing

### DIFF
--- a/packaging/package.sh
+++ b/packaging/package.sh
@@ -162,11 +162,16 @@ function packaging() {
     cp "$PWD/$DISTRO/cloud.spec" "$RPMDIR/SPECS"
 
     (cd "$RPMDIR"; rpmbuild --define "_topdir ${RPMDIR}" "${DEFVER}" "${DEFFULLVER}" "${DEFREL}" ${DEFPRE+"$DEFPRE"} ${DEFOSSNOSS+"$DEFOSSNOSS"} ${DEFSIM+"$DEFSIM"} -bb SPECS/cloud.spec)
-    (cd $PWD/../; git reset --hard)
     if [ $? -ne 0 ]; then
+        if [ "$USE_TIMESTAMP" == "true" ]; then
+            (cd $PWD/../; git reset --hard)
+        fi
         echo "RPM Build Failed "
         exit 3
     else
+        if [ "$USE_TIMESTAMP" == "true" ]; then
+            (cd $PWD/../; git reset --hard)
+        fi
         echo "RPM Build Done"
     fi
     exit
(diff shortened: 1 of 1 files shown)


packaging/package.sh
```diff
@@ -162,11 +162,12 @@ function packaging() {
     cp "$PWD/$DISTRO/cloud.spec" "$RPMDIR/SPECS"
 
     (cd "$RPMDIR"; rpmbuild --define "_topdir ${RPMDIR}" "${DEFVER}" "${DEFFULLVER}" "${DEFREL}" ${DEFPRE+"$DEFPRE"} ${DEFOSSNOSS+"$DEFOSSNOSS"} ${DEFSIM+"$DEFSIM"} -bb SPECS/cloud.spec)
-    (cd $PWD/../; git reset --hard)
     if [ $? -ne 0 ]; then
+        (cd $PWD/../; git reset --hard)
         echo "RPM Build Failed "
         exit 3
     else
+        (cd $PWD/../; git reset --hard)
```
rhtyd: Should the `git reset --hard` be part of the packaging script at all?

khos2ow: Technically yes, because if `--use-timestamp` is used actual POM versions are changed and to leave the working directory in pristine state it's needed.

rhtyd: I would prefer to remove it, especially if someone is working on their feature/bugfix branch it risks removing changes (and losing code).

khos2ow: That's actually a good point, I'll add an extra if condition around it later.

packaging/package.sh
```diff
@@ -162,11 +162,16 @@ function packaging() {
     cp "$PWD/$DISTRO/cloud.spec" "$RPMDIR/SPECS"
 
     (cd "$RPMDIR"; rpmbuild --define "_topdir ${RPMDIR}" "${DEFVER}" "${DEFFULLVER}" "${DEFREL}" ${DEFPRE+"$DEFPRE"} ${DEFOSSNOSS+"$DEFOSSNOSS"} ${DEFSIM+"$DEFSIM"} -bb SPECS/cloud.spec)
-    (cd $PWD/../; git reset --hard)
     if [ $? -ne 0 ]; then
+        if [ "$USE_TIMESTAMP" == "true" ]; then
```
rafaelweingartner: What is the default of `USE_TIMESTAMP`? `false`?

khos2ow: Yes, defined on line 185.

borisstoyanov:
@blueorangutan package

PR #2636, opened 2018-05-10
Fix limitation on tag matching in 'migrateVolume' with disk offering replacement

rafaelweingartner:
## Description

When the feature to enable disk offering replacement during volume migration was created, we were forcing the tags of the new disk offering to exact the same as the tags of the target storage poll. However, that is not how ACS manages volumes allocation. This change modifies this validation to make it consistent with volume allocation.

The idea is to make the validation consistent with the mechanism used to select a storage pool to deploy a volume when a virtual machine is deployed or when a new data disk is allocated. The scenarios when this method returns true or false is presented in the following table.

<table border="1">
           <tr>
               <th>#</th><th>Disk offering tags</th><th>Storage tags</th><th>Does the storage support the disk offering?</th>
           </tr>
           <body>
           <tr>
               <td>1</td><td>A,B</td><td>A</td><td>NO</td>
          </tr>
           <tr>
               <td>2</td><td>A,B,C</td><td>A,B,C,D,X</td><td>YES</td>
           </tr>
           <tr>
               <td>3</td><td>A,B,C</td><td>X,Y,Z</td><td>NO</td>
           </tr>
           <tr>
               <td>4</td><td>null</td><td>A,S,D</td><td>YES</td>
           </tr>
           <tr>
               <td>5</td><td>A</td><td>null</td><td>NO</td>
           </tr>
           <tr>
               <td>6</td><td>null</td><td>null</td><td>YES</td>
           </tr>
           </body>
        </table>

## Types of changes

## How Has This Been Tested?
Locally and via unit tests that were added for the code that has been changed.

## Checklist:

Testing

### DIFF
--- a/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
@@ -19,6 +19,7 @@ package com.cloud.storage;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -2161,9 +2162,9 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
      * Performs the validations required for replacing the disk offering while migrating the volume of storage. If no new disk offering is provided, we do not execute any validation.
      * If a disk offering is informed, we then proceed with the following checks.
      * <ul>
-     *  <li>We check if the given volume is of ROOT type. We cannot change the disk offering of a ROOT volume. Therefore, we thrown an {@link InvalidParameterValueException}.
-     *  <li>We the disk is being migrated to shared storage and the new disk offering is for local storage (or vice versa), we throw an {@link InvalidParameterValueException}. Bear in mind that we are validating only the new disk offering. If none is provided we can override the current disk offering. This means, placing a volume with shared disk offering in local storage and vice versa.
-     *  <li>We then proceed checking if the tags of the new disk offerings match the tags of the target storage. If they do not match an {@link InvalidParameterValueException} is thrown.
+     *  <li>We check if the given volume is of ROOT type. We cannot change the disk offering of a ROOT volume. Therefore, we thrown an {@link InvalidParameterValueException};
+     *  <li>We the disk is being migrated to shared storage and the new disk offering is for local storage (or vice versa), we throw an {@link InvalidParameterValueException}. Bear in mind that we are validating only the new disk offering. If none is provided we can override the current disk offering. This means, placing a volume with shared disk offering in local storage and vice versa;
+     *  <li>We then proceed checking the target storage pool supports the new disk offering {@link #doesTargetStorageSupportNewDiskOffering(StoragePool, DiskOfferingVO)}.
      * </ul>
      *
      * If all of the above validations pass, we check if the size of the new disk offering is different from the volume. If it is, we log a warning message.
@@ -2175,10 +2176,9 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
         if ((destPool.isShared() && newDiskOffering.getUseLocalStorage()) || destPool.isLocal() && newDiskOffering.isShared()) {
             throw new InvalidParameterValueException("You cannot move the volume to a shared storage and assing a disk offering for local storage and vice versa.");
         }
-        String storageTags = getStoragePoolTags(destPool);
-        if (!StringUtils.areTagsEqual(storageTags, newDiskOffering.getTags())) {
-            throw new InvalidParameterValueException(String.format("Target Storage [id=%s] tags [%s] does not match new disk offering [id=%s] tags [%s].", destPool.getUuid(), storageTags,
-                    newDiskOffering.getUuid(), newDiskOffering.getTags()));
+        if (!doesTargetStorageSupportNewDiskOffering(destPool, newDiskOffering)) {
+            throw new InvalidParameterValueException(String.format("Target Storage [id=%s] tags [%s] does not match new disk offering [id=%s] tags [%s].", destPool.getUuid(),
+                    getStoragePoolTags(destPool), newDiskOffering.getUuid(), newDiskOffering.getTags()));
         }
         if (volume.getSize() != newDiskOffering.getDiskSize()) {
             DiskOfferingVO oldDiskOffering = this._diskOfferingDao.findById(volume.getDiskOfferingId());
@@ -2187,6 +2187,54 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
                     volume.getUuid(), oldDiskOffering.getUuid(), newDiskOffering.getUuid()));
         }
         s_logger.info(String.format("Changing disk offering to [uuid=%s] while migrating volume [uuid=%s, name=%s].", newDiskOffering.getUuid(), volume.getUuid(), volume.getName()));
+    }
+
+    /**
+     *  Checks if the target storage supports the new disk offering.
+     *  This validation is consistent with the mechanism used to select a storage pool to deploy a volume when a virtual machine is deployed or when a new data disk is allocated.
+     *
+     *  The scenarios when this method returns true or false is presented in the following table.
+     *
+     *   <table border="1">
+     *      <tr>
+     *          <th>#</th><th>Disk offering tags</th><th>Storage tags</th><th>Does the storage support the disk offering?</th>
+     *      </tr>
+     *      <body>
+     *      <tr>
+     *          <td>1</td><td>A,B</td><td>A</td><td>NO</td>
+     *      </tr>
+     *      <tr>
+     *          <td>2</td><td>A,B,C</td><td>A,B,C,D,X</td><td>YES</td>
+     *      </tr>
+     *      <tr>
+     *          <td>3</td><td>A,B,C</td><td>X,Y,Z</td><td>NO</td>
+     *      </tr>
+     *      <tr>
+     *          <td>4</td><td>null</td><td>A,S,D</td><td>YES</td>
+     *      </tr>
+     *      <tr>
+     *          <td>5</td><td>A</td><td>null</td><td>NO</td>
+     *      </tr>
+     *      <tr>
+     *          <td>6</td><td>null</td><td>null</td><td>YES</td>
+     *      </tr>
+     *      </body>
+     *   </table>
+     */
+    protected boolean doesTargetStorageSupportNewDiskOffering(StoragePool destPool, DiskOfferingVO newDiskOffering) {
+        String newDiskOfferingTags = newDiskOffering.getTags();
+        if (org.apache.commons.lang.StringUtils.isBlank(newDiskOfferingTags)) {
+            return true;
+        }
+        String storagePoolTags = getStoragePoolTags(destPool);
+        boolean isStorageTagsBlank = org.apache.commons.lang.StringUtils.isBlank(storagePoolTags);
+        if (isStorageTagsBlank) {
+            return false;
+        }
+        String[] storageTagsAsStringArray = org.apache.commons.lang.StringUtils.split(storagePoolTags, ",");
+        String[] newDiskOfferingTagsAsStringArray = org.apache.commons.lang.StringUtils.split(newDiskOfferingTags, ",");
+
+        return CollectionUtils.isSubCollection(Arrays.asList(newDiskOfferingTagsAsStringArray), Arrays.asList(storageTagsAsStringArray));
     }
 
     /**
--- a/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
@@ -2176,7 +2176,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
         if ((destPool.isShared() && newDiskOffering.getUseLocalStorage()) || destPool.isLocal() && newDiskOffering.isShared()) {
             throw new InvalidParameterValueException("You cannot move the volume to a shared storage and assing a disk offering for local storage and vice versa.");
         }
-        if (!doesTargetStorageSupportDiskOffering(destPool, newDiskOffering)) {
+        if (!doesTargetStorageSupportNewDiskOffering(destPool, newDiskOffering)) {
             throw new InvalidParameterValueException(String.format("Target Storage [id=%s] tags [%s] does not match new disk offering [id=%s] tags [%s].", destPool.getUuid(),
                     getStoragePoolTags(destPool), newDiskOffering.getUuid(), newDiskOffering.getTags()));
         }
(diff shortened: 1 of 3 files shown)


rhtyd:
@rafaelweingartner this is opened against master, if you do not want this in 4.11 can you change the milestone to 4.12.0.0?

server/src/test/java/com/cloud/storage/VolumeApiServiceImplTest.java
```diff
@@ -580,9 +580,86 @@ public void validateConditionsToReplaceDiskOfferingOfVolumeTestEverythingWorking
...
         inOrder.verify(newDiskOfferingMock, times(0)).isShared();
         inOrder.verify(volumeApiServiceImpl).getStoragePoolTags(storagePoolMock);
-        inOrder.verify(newDiskOfferingMock).getTags();
 
         inOrder.verify(volumeVOMock).getSize();
         inOrder.verify(newDiskOfferingMock).getDiskSize();
     }
+
+    @Test
+    public void doesTargetStorageSupportNewDiskOfferingTestDiskOfferingMoreTagsThanStorageTags() {
```
DaanHoogland: wow, how java of you @rafaelweingartner , can you translate this name?

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2142,6 +2142,55 @@ protected void validateConditionsToReplaceDiskOfferingOfVolume(VolumeVO volume,
...
+     *      <tr>
+     *          <td>5</td><td>A</td><td>null</td><td>NO</td>
+     *      </tr>
+     *      <tr>
+     *          <td>6</td><td>null</td><td>null</td><td>YES</td>
+     *      </tr>
+     *      </body>
+     *   </table>
+     */
+    protected boolean doesTargetStorageSupportNewDiskOffering(StoragePool destPool, DiskOfferingVO newDiskOffering) {
```
DaanHoogland: i see the parameter appearing in the method name. why not isSupported(pool, offering);?

server/src/test/java/com/cloud/storage/VolumeApiServiceImplTest.java
rafaelweingartner: Sure, this is something I do to control the unit tests. The pattern I normally use is the following:
<methodNamte>Test<parameterStateBeingTested>
Translating this case:
I am testing the method `doesTargetStorageSupportNewDiskOffering` with the following condition:
* the disk offering will have more tags than the target storage

There is also another format I use:
`<methodNamte>Test`
Happy day test. It can be the case when the method finishes successfully or a method that does not have different execution flows.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
rafaelweingartner: That is a good point. I normally try to be very descriptive with method names. When I see names such as ` isSupported`, even though we see the parameters there,  we have no clues what the method is checking. It could be that the method is checking if the storage has enough IOPs, or the method has enough space, or any other business logic. Maybe, a better name here would be ` doesTargetStorageSupportNewDiskOfferingTags`.

Of course, here I have documented the method in detail. However, this should not be an excuse not to try to work out a good and descriptive method name.

rafaelweingartner:
@rhtyd the feature I am enhancing is only in 4.12. That is why this PR is targeting 4.12.

rafaelweingartner:
Ops, now I see what you meant. I selected 4.11.2, instead of 4.12.
Problem fixed now.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2142,6 +2142,55 @@ protected void validateConditionsToReplaceDiskOfferingOfVolume(VolumeVO volume,
...
+        String newDiskOfferingTags = newDiskOffering.getTags();
+        boolean isnewDiskOfferingTagsBlank = org.apache.commons.lang.StringUtils.isBlank(newDiskOfferingTags);
```
nitin-maharana: Anyway, we use the variable only once, can we scrap this and put the statement directly in IF condition. Same for isStorageTagsBlank. Like if(org.apache.commons.lang.StringUtils.isBlank(newDiskOfferingTags)).

rafaelweingartner: Sure, I can do that.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2142,6 +2142,55 @@ protected void validateConditionsToReplaceDiskOfferingOfVolume(VolumeVO volume,
...
+        if (isnewDiskOfferingTagsBlank) {
+            return true;
+        }
+        String storageTags = getStoragePoolTags(destPool);
```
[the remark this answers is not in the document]
rafaelweingartner: Done

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2142,6 +2142,54 @@ protected void validateConditionsToReplaceDiskOfferingOfVolume(VolumeVO volume,
...
+     *      </body>
+     *   </table>
+     */
+    protected boolean doesTargetStorageSupportNewDiskOffering(StoragePool destPool, DiskOfferingVO newDiskOffering) {
+        String newDiskOfferingTags = newDiskOffering.getTags();
+        if (org.apache.commons.lang.StringUtils.isBlank(newDiskOfferingTags)) {
+            return true;
+        }
+        String storagePoolTags = getStoragePoolTags(destPool);
+        boolean isStorageTagsBlank = org.apache.commons.lang.StringUtils.isBlank(storagePoolTags);
```
nitin-maharana: Here as well.

server/src/test/java/com/cloud/storage/VolumeApiServiceImplTest.java
```diff
@@ -580,9 +580,86 @@ public void validateConditionsToReplaceDiskOfferingOfVolumeTestEverythingWorking
...
+        StoragePool storagePoolMock = Mockito.mock(StoragePool.class);
+        Mockito.doReturn("A").when(volumeApiServiceImpl).getStoragePoolTags(storagePoolMock);
+
+        boolean result = volumeApiServiceImpl.doesTargetStorageSupportNewDiskOffering(storagePoolMock, diskOfferingVoMock);
+
+        Assert.assertFalse(result);
+    }
+
+    @Test
+    public void doesTargetStorageSupportNewDiskOfferingTestDiskOfferingTagsIsSubSetOfStorageTags() {
```
nitin-maharana: I think, would be good if we add the equal case as well. Like DiskOfferingTag('A') and StoragePoolTag('A'). Should return true.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
rafaelweingartner: Done

server/src/test/java/com/cloud/storage/VolumeApiServiceImplTest.java
rafaelweingartner: Sure

rafaelweingartner:
@blueorangutan package

rafaelweingartner:
@DaanHoogland can you please run tests here?

DaanHoogland:
@blueorangutan test

borisstoyanov:
@blueorangutan test

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2189,6 +2189,53 @@ protected void validateConditionsToReplaceDiskOfferingOfVolume(VolumeVO volume,
...
+     *      <tr>
+     *          <td>5</td><td>A</td><td>null</td><td>NO</td>
+     *      </tr>
+     *      <tr>
+     *          <td>6</td><td>null</td><td>null</td><td>YES</td>
+     *      </tr>
+     *      </body>
+     *   </table>
+     */
+    protected boolean doesTargetStorageSupportNewDiskOffering(StoragePool destPool, DiskOfferingVO newDiskOffering) {
```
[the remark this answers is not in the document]
rafaelweingartner: Well, I am not sure if we have some other places that can use this method. I would normally only set a method to public and declare it in an interface if I need really need it. I can declare it if you think it is important though. 

P.S.: I find our interface usage sometimes quite silly with single implementations.  A good example is using interface for POJOs. It does not provide benefits, the only thing it does is adding dozens more lines for every object.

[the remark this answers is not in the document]
DaanHoogland: ok, nice subject: another valid usage is mutual dependent services. Or dependency injection.
I this case my concern is that the implementation of tag matching is doen in several places while it should be uniform and thus implemented centrally. doesTargetStorageASupportNewDiskOffering() is a specialisation of does_this_entities_tags_match_the_resource_it_is_going_to(entity, resource) which is applicable to volume/storage in case of new offering but also in case of migration, and is also applicable for vm-serviceoffering/host-cluster-pod etc. I cant' think of any others by head but these are surely there as well.

rafaelweingartner:
This new method that I implemented is the consolidation of the logic of tag matching in ACS, which is not extracted in a specific method and spread around the code. That is one of the reasons why I “made the mistake”  in one of the features I developed where the tag matching was being too strict (requiring all tags to match) and was not making sense to operation guys.

To answer your question, yes this method can be used in other places. I am not sure if simply declaring the method in the interface will help much though. The person that is willing to consolidate these execution flow could do both, declare this method in the interface, and then use it in all of those places that you mentioned.

What do you think?

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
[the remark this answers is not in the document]
rafaelweingartner: This new method that I implemented is the consolidation of the logic of tag matching in ACS, which is not extracted in a specific method and spread around the code. That is one of the reasons why I “made the mistake” in one of the features I developed where the tag matching was being too strict (requiring all tags to match) and was not making sense to operation guys.

To answer your question, yes this method can be used in other places. I am not sure if simply declaring the method in the interface will help much though. The person that is willing to consolidate these execution flow could do both, declare this method in the interface, and then use it in all of those places that you mentioned.

What do you think?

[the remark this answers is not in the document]
DaanHoogland: I think it is a nice future plan to extract the logic to a tagging utility and then make this method use it and be given a more general name like
```
 doesTargetStorageSupportDiskOffering() // not no 'New' in there
```
and have it call the utility
```
 matchOk(Set<String> tags, List<String> supported);
```
plan?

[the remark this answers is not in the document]
DaanHoogland: matchOk is just a bad method-name i came up with. we need a set of methods that do the matching for us for
Set<String> tags, List<String> supported or
List<List<String>> tags, List<String> supported or
String tag(s), List<String> supported etc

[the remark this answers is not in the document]
rafaelweingartner: Ok, but in this case I think we should create methods on demand. I mean, we do not know what we need yet. Therefore, there is no reason to start coding methods that we do not even know if we will use them.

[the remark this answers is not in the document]
DaanHoogland: ok, maybe postpone this discussion to a beer session ;)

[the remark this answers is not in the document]
DaanHoogland: of course I meant hackathon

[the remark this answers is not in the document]
rafaelweingartner: Ahaha.. I do not know why, but I would prefer a beer session...

I will do the change you asked later (making the method public in the interface).

rafaelweingartner:
@DaanHoogland I applied your suggestion here regarding "doesTargetStorageSupportDiskOffering" method in the interface.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2221,17 +2221,22 @@ protected void validateConditionsToReplaceDiskOfferingOfVolume(VolumeVO volume,
...
+        if (org.apache.commons.lang.StringUtils.isBlank(diskOfferingTags)) {
             return true;
         }
         String storagePoolTags = getStoragePoolTags(destPool);
         if (org.apache.commons.lang.StringUtils.isBlank(storagePoolTags)) {
             return false;
         }
         String[] storageTagsAsStringArray = org.apache.commons.lang.StringUtils.split(storagePoolTags, ",");
-        String[] newDiskOfferingTagsAsStringArray = org.apache.commons.lang.StringUtils.split(newDiskOfferingTags, ",");
+        String[] newDiskOfferingTagsAsStringArray = org.apache.commons.lang.StringUtils.split(diskOfferingTags, ",");
```
DaanHoogland: there is a StringUtils.listToCsvTags() method.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2189,6 +2189,58 @@ protected void validateConditionsToReplaceDiskOfferingOfVolume(VolumeVO volume,
...
+    @Override
+    public boolean doesTargetStorageSupportDiskOffering(StoragePool destPool, String diskOfferingTags) {
+        if (org.apache.commons.lang.StringUtils.isBlank(diskOfferingTags)) {
+            return true;
+        }
+        String storagePoolTags = getStoragePoolTags(destPool);
+        if (org.apache.commons.lang.StringUtils.isBlank(storagePoolTags)) {
+            return false;
+        }
+        String[] storageTagsAsStringArray = org.apache.commons.lang.StringUtils.split(storagePoolTags, ",");
```
DaanHoogland: there is a  StringUtils.csvTagsToList()

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
DaanHoogland: wrong advice here, see below

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
rafaelweingartner: Are you talking about `com.cloud.utils.StringUtils.csvTagsToList(String)`?
Didn't we talk about stop using that "utils" class, so we can then remove it in the future?

DaanHoogland: I think we agreed to move all string operation in it so it we have a small interface plain to external libraries.

rafaelweingartner: Did we?! That means we would create a layer over a library. I think we had a discussion, but I was under the impression that the idea was not to redo the wheel. Of course, my memory is not as good as it was before, so I am probably mistaken.

DaanHoogland: In my opinion it is best to isolate all external libraries in wrappers so porting from version to version or from vendor to vendor becomes easier/more isolated. In fact the hypervisor guru and resource and data motion strategies are examples of that as well. We didn't make an onmail list dicision of it but I thought aI had convinced you of it.

rafaelweingartner: Well, but those "gurus"/plugins/drivers and so on are different "things".
I mean, they are not library, they are part of our code base to deliver our business. On the other hand, creating a layer over third party utils libraries such as StringUtils, NumberUtils, Hibernate, Spring and so on sound like an overkill.

We should discuss that we beers in Montreal :)

DaanHoogland: one word: gson
and a lot of beers

DaanHoogland:
@rafaelweingartner , I would argue this is a bug fix and should go on the lts branch (4.11). agree?

rafaelweingartner:
Unless I am mistaken, this feature is not in 4.11

DaanHoogland:
ah, the offering replacement.

borisstoyanov:
Is there any outstanding comments? I think this looks good for merging.

PR #2639, opened 2018-05-11
[CLOUDSTACK-10276] listVolumes not working when storage UUID is not a UUID

rafaelweingartner:
## Description

When configuring a pre-setup primary storage we can enter the name-label of the storage that is going to be used by ACS and is already set up in the host. The problem is that we can use any String of characters there, and this String does not need to be a UUID. When listing volumes from a primary storage that has such conditions, the list will return all of the volumes in the cloud because the “API framework” will ignore that value as it is not a UUID type.

## Types of changes

## How Has This Been Tested?
Locally

## Checklist:

Testing

### DIFF
--- a/api/src/org/apache/cloudstack/api/command/admin/volume/ListVolumesCmdByAdmin.java
+++ b/api/src/org/apache/cloudstack/api/command/admin/volume/ListVolumesCmdByAdmin.java
@@ -34,32 +34,5 @@ import com.cloud.storage.Volume;
 @APICommand(name = "listVolumes", description = "Lists all volumes.", responseObject = VolumeResponse.class, responseView = ResponseView.Full, entityType = {Volume.class},
         requestHasSensitiveInfo = false, responseHasSensitiveInfo = false)
 public class ListVolumesCmdByAdmin extends ListVolumesCmd {
-    public static final Logger s_logger = Logger.getLogger(ListVolumesCmdByAdmin.class.getName());
-
-    @Parameter(name=ApiConstants.POD_ID, type=CommandType.UUID, entityType=PodResponse.class,
-            description="the pod id the disk volume belongs to")
-    private Long podId;
-
-
-    @Parameter(name=ApiConstants.STORAGE_ID, type=CommandType.UUID, entityType=StoragePoolResponse.class,
-            description="the ID of the storage pool, available to ROOT admin only", since="4.3", authorized = { RoleType.Admin })
-    private Long storageId;
-
-
-    /////////////////////////////////////////////////////
-    /////////////////// Accessors ///////////////////////
-    /////////////////////////////////////////////////////
-
-
-    @Override
-    public Long getPodId() {
-        return podId;
-    }
-
-
-    @Override
-    public Long getStorageId() {
-        return storageId;
-    }
 
 }
--- a/api/src/org/apache/cloudstack/api/command/admin/volume/ListVolumesCmdByAdmin.java
+++ b/api/src/org/apache/cloudstack/api/command/admin/volume/ListVolumesCmdByAdmin.java
@@ -16,23 +16,15 @@
 // under the License.
 package org.apache.cloudstack.api.command.admin.volume;
 
-import org.apache.log4j.Logger;
-
-import org.apache.cloudstack.acl.RoleType;
 import org.apache.cloudstack.api.APICommand;
-import org.apache.cloudstack.api.ApiConstants;
-import org.apache.cloudstack.api.Parameter;
 import org.apache.cloudstack.api.ResponseObject.ResponseView;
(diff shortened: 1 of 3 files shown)


api/src/org/apache/cloudstack/api/command/admin/volume/ListVolumesCmdByAdmin.java
```diff
@@ -34,32 +34,5 @@
 @APICommand(name = "listVolumes", description = "Lists all volumes.", responseObject = VolumeResponse.class, responseView = ResponseView.Full, entityType = {Volume.class},
         requestHasSensitiveInfo = false, responseHasSensitiveInfo = false)
 public class ListVolumesCmdByAdmin extends ListVolumesCmd {
-    public static final Logger s_logger = Logger.getLogger(ListVolumesCmdByAdmin.class.getName());
-
-    @Parameter(name=ApiConstants.POD_ID, type=CommandType.UUID, entityType=PodResponse.class,
-            description="the pod id the disk volume belongs to")
-    private Long podId;
```
rhtyd: @rafaelweingartner Is podId an unused parameter?

rafaelweingartner: It is in the superclass as well. These parameters were declared twice. One in the "admin" class, and other in the "normal" one.

rhtyd: make sense to remove if parent class has them already declared.

rafaelweingartner:
@rhtyd I thought about that. However, after evaluating the volume of changes, I decided to do a simple fix.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

DaanHoogland:
arguing with jenkins

DaanHoogland:
see if jenkins is convinced

rhtyd:
I've my reservations on the fix, instead perhaps it may be better and could be simpler to handle use of external id as uuid when adding a pre-setup storage pool. This may not require us to remove the uuid-id conversations/annotations in the list API, also this will break auto-completion of storage ids in various APIs for CLIs such as cloudmonkey.

rafaelweingartner:
@rhtyd how does it break cloudmonkey? 
Cloudmonkey continues to see the same thing. There is not change in the API, nothing will change for clients. I just stopped using the automagic conversion system via reflection that takes the UUID entered as ID (a UUID) and then converts to a database long ID.

rhtyd:
@rafaelweingartner yes you're right, I tested it and it works due to a heuristic I had implemented (and forgotten about it, which uses the api arg removing id/ids and tries to find a list API). It's eod for me here, let me get back to you tomorrow if there can be any simpler fix.

rhtyd:
I'm unable to find time to do the investigation on this. @DaanHoogland can you advise, and if it's all okay please go ahead with merging this.

rafaelweingartner:
I really do not see any other smaller/simpler solution for this issue. Moreover, it does not affect/break anything in the client side. The client is already seeing an ID, which is a UUID.

The changes are only internal to ACS. I changed the type of `storageId` from `long` to `String` to avoid using that method that converts the given UUID to an internal database ID. That method is validating the input, and ignoring anything that is not an UUID. Then, I am using the UUID as the filter at `QueryManagerImpl.java`.

The changed lines are only 2. The other lines are a result of a code cleanup that I did, and the code formatting I applied.

DaanHoogland:
@rafaelweingartner I am fine with merging this so far as we don't allow for manually entering a non-uuid storage id. In my understanding it doesn't. Is that correct?

borisstoyanov:
@rhtyd, @DaanHoogland, @PaulAngus Cool, this seems to be the last pending item in the 4.11.1 milestone, I think we can merge and cut the RC.

rafaelweingartner:
Actually, it does allow entering "non-UUID" parameters, and that is the problem I am addressing here. When adding a pre-setup storage, we use the field "name-label" from XenServer (e.g. "pre-setup storage 1"). If we set anything, but a UUID in that field, we can add a pre-setup storage with the name-label, and ACS will use the name-label as the UUID of the storage. 

This creates a problem, with the listVolumes method. Because we annotated the `storageId` as a UUID in the ACS code. There is a method, which intercepts the request and converts the UUIDs (String values) entered by users to database IDs, which are long values. The problem is that this method only handles UUIDs, anything else is ignored. Therefore, when listing volumes of a pre-setup storage, we get the volume listing without filtering by “storageId”.

By changing the `storageId` from `long` to `String`, and changing the annotation from `UUID` to `STRING`, we avoid using the method that converts String UUIDs into database long IDs, and everything starts working again.

Did you understand the problem and the solution I used?

rhtyd:
Alright merging this now.

PR #2656, opened 2018-05-18
Only perform certain actions here with managed storage if the VM is s…

mike-tutkowski:
…tarted up in a new cluster (not if it is just started up on a new host in the same cluster).

## Description
Example: A VM that uses managed storage is stopped. The VM is then started on a different host in the same cluster. The Start operation fails.

To get around this issue, you must either start the VM up on the same host or on a host in a different cluster.

The reason is due to a slightly erroneous check in VolumeOrchestrator.prepare.

To solve this issue, we should be checking if the cluster ID changes, not if the host ID changes.

## Types of changes

## How Has This Been Tested?

I noticed that I could not start a stopped VM that uses managed storage on a different host in the same cluster as the host it was originally running on.

After I changed the code (this PR), I re-tested the scenario. I can now start the VM up on a different host in the same cluster as the host it was originally running on. Also, I can still start it up on the same host as it was originally running on or on a host in a different cluster.

## Checklist:

Testing


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Merging this based on regression tests and code review.

PR #2668, opened 2018-05-23
Update vhd-util URL

imduffy15:
URL for vhd-util was broken, updated accordingly.


rafaelweingartner:
Are you that it is broken? I mean, I just tested `http://download.cloudstack.org/tools/vhd-util`, and it is working here.

rafaelweingartner:
Sure it has. However, the e-mail mentions `download.cloud.com` and not `download.cloudstack.org`. The last one is owned by ACS community, while the first one is still owned by Citrix and does not represents ACS anymore.

PR #2703, opened 2018-06-07
Fix concurrency problem when moving ACL rules with drag&drop

rafaelweingartner:
## Description

There was a concurrency problem with the “moveNetworkAclItem” API method. If two users were changing the ACL rules order at the same time, this could lead to inconsistent actions.

To solve the problem we added a “consistency check ” parameter, which is used to hold the consistency hash. This hash is created using an MD5 hash function on a String that is created with all ACL rules UUIDs concatenated in their order, which is defined via the ‘number’ field.

We also lock the editing of the ACL while executing the upgrade. This allows us to handle race conditions nicely, and present a good feedback for the user.

We are not forcing users to send the consistency hash variable. Therefore, they can use the API call without the hash parameter. Therefore, they assume the risk of problems that may arise due to race conditions scenarios.

## Types of changes

## How Has This Been Tested?
Via unit tests and locally.

## Checklist:

Testing

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/network/MoveNetworkAclItemCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/network/MoveNetworkAclItemCmd.java
@@ -43,6 +43,9 @@ public class MoveNetworkAclItemCmd extends BaseAsyncCustomIdCmd {
     @Parameter(name = ApiConstants.NEXT_ACL_RULE_ID, type = CommandType.STRING, description = "The ID of the rule that is right after the new position where the rule being moved is going to be placed. This value can be 'NULL' if the rule is being moved to the last position of the network ACL list.")
     private String nextAclRuleUuid;
 
+    @Parameter(name = ApiConstants.MOVE_ACL_CONSISTENCY_HASH, type = CommandType.STRING, description = "Md5 hash used to check the consistency of the ACL rule list before applying the ACL rule move. This check is useful to manage concurrency problems that may happen when multiple users are editing the same ACL rule listing. The parameter is not required. Therefore, if the user does not send it, he/she is assuming the risk of moving ACL rules without checking the consistency of the access control list before executing the move. We use MD5 hash function on a String that is composed of all UUIDs of the ACL rules in concatenated in their respective order (order defined via 'number' field).")
+    private String aclConsistencyHash;
+
     @Override
     public void execute() {
         CallContext.current().setEventDetails(getEventDescription());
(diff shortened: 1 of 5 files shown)


borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan package

borisstoyanov:
let me retry @rafaelweingartner 
@blueorangutan package

borisstoyanov:
@blueorangutan test

api/src/main/java/org/apache/cloudstack/api/command/user/network/MoveNetworkAclItemCmd.java
```diff
@@ -43,6 +43,9 @@
     @Parameter(name = ApiConstants.NEXT_ACL_RULE_ID, type = CommandType.STRING, description = "The ID of the rule that is right after the new position where the rule being moved is going to be placed. This value can be 'NULL' if the rule is being moved to the last position of the network ACL list.")
     private String nextAclRuleUuid;
 
+    @Parameter(name = ApiConstants.MOVE_ACL_CONSISTENCY_HASH, type = CommandType.STRING, description = "Md5 hash used to check the consistency of the ACL rule list before applying the ACL rule move. This check is useful to manage concurrency problems that may happen when multiple users are editing the same ACL rule listing. The parameter is not required. Therefore, if the user does not send it, he/she is assuming the risk of moving ACL rules without checking the consistency of the access control list before executing the move. We use MD5 hash function on a String that is composed of all UUIDs of the ACL rules in concatenated in their respective order (order defined via 'number' field).")
```
DaanHoogland: very solid method but very hard to use when calling the API, isn't it?
Would this work or what could I do with for instance cloudmonkey?

rafaelweingartner: Yes, you are right. I tried to think of a better solution, but I could not come up with anything better: (

I do not see much difficulty to use it programmatically. However, I do see the problem with cloudmonkey or other similar tools if they do not provide some out of the box enabler methods.  On the bright side, ACS UI works like a charm without any burden to operators or users.

BTW: I am always open to suggestions!

rafaelweingartner: Ah, there is something else that I forgot to mention. The hash is not mandatory. Therefore, it will work via CloudMonkey, but it will not be a safe operation (as it was not in the past).

PR #2706, opened 2018-06-11
packaging: use libuuid x86_64 package for cloudstack-common

rhtyd:
This forces the rpm cloudstack-common package to install libuuid x86_64 pkg instead of the i686 package.
screenshot from 2018-06-11 12-27-59

## Types of changes

## Checklist:

Testing

@blueorangutan package

### DIFF
--- a/tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
+++ b/tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
@@ -80,6 +80,13 @@ function install_packages() {
 
   ${apt_get} install links
 
+  #32 bit architecture support for vhd-util: not required for 32 bit template
+  if [ "${arch}" != "i386" ]; then
+    dpkg --add-architecture i386
+    apt-get update
+    ${apt_get} install libuuid1:i386 libc6:i386
+  fi
+
   # Install xenserver guest utilities as debian repos don't have it
   wget https://mirrors.kernel.org/ubuntu/pool/main/x/xe-guest-utilities/xe-guest-utilities_7.10.0-0ubuntu1_amd64.deb
   dpkg -i xe-guest-utilities_7.10.0-0ubuntu1_amd64.deb
--- a/tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
+++ b/tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
@@ -78,12 +78,7 @@ function install_packages() {
   apt-get clean
   apt-get autoclean
 
-  #32 bit architecture support:: not required for 32 bit template
-  if [ "${arch}" != "i386" ]; then
-    dpkg --add-architecture i386
-    apt-get update
-    ${apt_get} install links:i386 libuuid1:i386 libc6:i386
-  fi
+  ${apt_get} install links
 
   # Install xenserver guest utilities as debian repos don't have it
   wget https://mirrors.kernel.org/ubuntu/pool/main/x/xe-guest-utilities/xe-guest-utilities_7.10.0-0ubuntu1_amd64.deb
--- a/packaging/centos63/cloud.spec
+++ b/packaging/centos63/cloud.spec
@@ -109,6 +109,7 @@ Obsoletes: cloud-daemonize < 4.1.0
 Group:   System Environment/Libraries
 %description common
 The Apache CloudStack files shared between agent and management server
+%global __requires_exclude ^libuuid\\.so\\.1$
 
 %package agent
 Summary: CloudStack Agent for KVM hypervisors
--- a/packaging/centos63/cloud.spec
+++ b/packaging/centos63/cloud.spec
@@ -94,6 +94,7 @@ management, and intelligence in CloudStack.
 
 %package common
 Summary: Apache CloudStack common files and scripts
+Requires: libuuid
 Requires: python
 Requires: python-argparse
 Requires: python-netaddr
--- a/packaging/centos63/cloud.spec
+++ b/packaging/centos63/cloud.spec
@@ -94,7 +94,6 @@ management, and intelligence in CloudStack.
 
 %package common
 Summary: Apache CloudStack common files and scripts
-Requires: libuuid
 Requires: python
 Requires: python-argparse
 Requires: python-netaddr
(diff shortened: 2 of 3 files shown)


rhtyd:
Closing this, the built package still tries to install the following: /cc @DaanHoogland 
 libuuid                  i686         2.23.2-52.el7                             base                                                               81 k
I'm not sure how to fix that.

rafaelweingartner:
@rhtyd you should force ``glibc` to use the 64 bits package, then I guess it should fix this problem.

rhtyd:
@rafaelweingartner I'm not sure how to force that, if needed I'll relook at this tomorrow in case @DaanHoogland @PaulAngus requires my assistance. This certainly is not a huge issue, we can do in 4.11.2.0.

rafaelweingartner:
Something like `Requires: glibc.x86_64` should do the tricky.

DaanHoogland:
Rohit, i just did a test with the packages from this PR. the host has libuuid.x86_64 2.23.2-43.el7_4.2 installed, so that is fine
the MS has both installed but does start, so I wonder if this is a problem.
I'll try the 6x version

DaanHoogland:
just double checked: the host also has both versions installed (without conflict)

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

wido:
I'm not so much of a RPM expert and I don't know what the exclude here does.

Why are we playing with .so files in a spec file while a RPM should handle this for us?

DaanHoogland:
@blueorangutan test

wido:
Looks good to me removing those old 32-bit things from the SSVM template :)

tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
```diff
@@ -78,12 +78,7 @@ function install_packages() {
   apt-get clean
   apt-get autoclean
 
-  #32 bit architecture support:: not required for 32 bit template
-  if [ "${arch}" != "i386" ]; then
-    dpkg --add-architecture i386
-    apt-get update
-    ${apt_get} install links:i386 libuuid1:i386 libc6:i386
-  fi
+  ${apt_get} install links
```
rhtyd: The vhd-util we bundle with systemvm.iso/cloudstack is used by ssvm, and needs the i386 libs:
ldd ./scripts/vm/hypervisor/xenserver/vhd-util
	linux-gate.so.1 =>  (0xf7f23000)
	libuuid.so.1 => /lib/i386-linux-gnu/libuuid.so.1 (0xf7edb000)
	libc.so.6 => /lib/i386-linux-gnu/libc.so.6 (0xf7d05000)
	/lib/ld-linux.so.2 (0xf7f25000)

rafaelweingartner: Does it need the 32 bits library? Or, is it a matter of creating a symbolic link to the files of the 64 bits library?

rhtyd: Yes, see the ldd output. Either we'll need libuuid:i386 or rebuild/ship a 64-bit vhd-util binary.

rafaelweingartner: I mean, can't we simply do the following?
```
mkdir /lib/x86_64-linux-gnu/
ln -s /lib/x86_64-linux-gnu/libuuid.so.1 /lib/i386-linux-gnu/libuuid.so.1
ln -s /lib/x86_64-linux-gnu/libc.so.6 /lib/i386-linux-gnu/libc.so.6
```

The libraries are already in the 64bits system when we install the "links" for 64bits. The problem is that they are in a different path that `vhd-util` does not know.

If we create the symbolic links, we would have the following then:
```
root@debian:~# ls -lah /lib/i386-linux-gnu/
lrwxrwxrwx  1 root root   31 Jun 14 08:25 libc.so.6 -> /lib/x86_64-linux-gnu/libc.so.6
lrwxrwxrwx  1 root root   34 Jun 14 08:24 libuuid.so.1 -> /lib/x86_64-linux-gnu/libuuid.so.1
```

Would this work?

DaanHoogland: @rhtyd if we have the proper version to rebuild that would be best. In any way teh cloudstack-common package doesn't need it, only the systemvm template.
we can either
- ammend the commit that removes the 32 dependencies to allow for glibc and libuuid, or
- rebuild vhdutil
What ever is easiest will do for this release, i think

rhtyd: @DaanHoogland I can try to build vhd-util x86_64 binary for linux, but we can do that for the future 4.11.2.0 release. The easiest would be to continue to install the 32bit dependencies in systemvm like we've done in the current 4.11.1 systemvmtemplate.

rhtyd: @DaanHoogland Based on the last comment, can you remove the changes to the systemvm template build script? Once we have the 64-bit vhd-util we can remove the changes from this file.

DaanHoogland: will do, and then of course test if this fixes the centos7 dependencies at least.

DaanHoogland:
@blueorangutan package

rhtyd:
@blueorangutan package

tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
```diff
@@ -78,11 +78,13 @@ function install_packages() {
   apt-get clean
   apt-get autoclean
 
-  #32 bit architecture support:: not required for 32 bit template
+  ${apt_get} install links
```
rhtyd: For 64-bit binaries, this may be moved to lines up like line 60-73. I'm not sure where links is used at all (perhaps for debugging purposes).

DaanHoogland: change if you will

DaanHoogland:
@blueorangutan test

PR #2714, opened 2018-06-19
send unsupported answer only when applicable

DaanHoogland:
## Description

A pokemon catch was handling all errors in the agent with libvirt resource loaded. 

## Types of changes

## Checklist:

Testing

### DIFF
--- a/core/src/com/cloud/resource/RequestWrapper.java
+++ b/core/src/com/cloud/resource/RequestWrapper.java
@@ -30,8 +30,13 @@ import com.cloud.agent.api.Command;
 
 public abstract class RequestWrapper {
     static public class CommandNotSupported extends NullPointerException {
+        Throwable reason = null;
         public CommandNotSupported(String msg) {
             super(msg);
+        }
+        public CommandNotSupported(String msg, Throwable cause) {
+            super(msg);
+            reason = cause;
         }
     }
 
@@ -115,10 +120,8 @@ public abstract class RequestWrapper {
                 keepResourceClass = resourceClass2;
 
                 commandWrapper = retrieveCommands(command.getClass(), resourceCommands2);
-            } catch (final ClassCastException e) {
-                throw new CommandNotSupported("No key found for '" + command.getClass() + "' in the Map!");
-            } catch (final NullPointerException e) {
-                throw e;
+            } catch (final ClassCastException | NullPointerException e) {
+                throw new CommandNotSupported("No key found for '" + command.getClass() + "' in the Map!", e);
             }
         }
         return commandWrapper;
--- a/core/src/com/cloud/resource/RequestWrapper.java
+++ b/core/src/com/cloud/resource/RequestWrapper.java
@@ -30,13 +30,12 @@ import com.cloud.agent.api.Command;
 
 public abstract class RequestWrapper {
     static public class CommandNotSupported extends NullPointerException {
-        Throwable reason = null;
         public CommandNotSupported(String msg) {
             super(msg);
         }
         public CommandNotSupported(String msg, Throwable cause) {
             super(msg);
-            reason = cause;
+            initCause(cause);
         }
     }
 
--- a/core/src/com/cloud/resource/RequestWrapper.java
+++ b/core/src/com/cloud/resource/RequestWrapper.java
@@ -29,6 +29,11 @@ import com.cloud.agent.api.Answer;
 import com.cloud.agent.api.Command;
 
 public abstract class RequestWrapper {
+    static public class CommandNotSupported extends NullPointerException {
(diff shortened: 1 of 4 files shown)


plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -1432,13 +1434,21 @@ public boolean stop() {
         return true;
     }
 
+    /**
```
rafaelweingartner: Big 👍 . We really need more of these in our code base :)
Great job!

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -1432,13 +1434,21 @@ public boolean stop() {
...
+     * This finds a command wrapper to handle the command and executes it.
+     * If no wrapper is found an {@see UnsupportedAnswer} is sent back.
+     * Any other exceptions are to be caught and wrapped in an generic {@see Answer}, marked as failed.
+     *
+     * @param cmd the instance of a {@see Command} to execute.
+     * @return the for the {@see Command} appropriate {@see Answer} or {@see UnsupportedAnswer}
+     */
     @Override
     public Answer executeRequest(final Command cmd) {
```
rafaelweingartner: What about a unit test case? Then, you can have a test case that detects if we catch only `RequestWrapper.CommandNotSupported` exceptions.

DaanHoogland: that would require throwing everything else. I will add a subset to catch the most obvious cases

rafaelweingartner: Ah, yes. If you want to check all possible cases. However, you can test with the most obvious ones such as `Nullpointer, Exception, CloudRuntimeException, and RuntimeException`

DaanHoogland: I am not really convinced that any of these make a lot of difference to one and another. How about the known and unknown case I just added?

core/src/com/cloud/resource/RequestWrapper.java
```diff
@@ -30,9 +30,14 @@
 
 public abstract class RequestWrapper {
     static public class CommandNotSupported extends NullPointerException {
+        Throwable reason = null;
```
rafaelweingartner: What about overriding the method `getCause` to return this value here?

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
rafaelweingartner: It helps. I only think that you could go and test the method `retryWhenAllFails` instead of using `executeRequest`. If we want unit tests, we should write them for the units.

core/src/com/cloud/resource/RequestWrapper.java
DaanHoogland: sorry, didn't think of it. should do.

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
DaanHoogland: we could also unittest RequestWrapper, yes.
executeRequest is the unit here.
higher prios came up, i might revisit

rafaelweingartner: No problems ;)
Thanks anyways for writing at least some tests.

DaanHoogland:
@blueorangutan package

PaulAngus:
@blueorangutan test

rhtyd:
@blueorangutan test

PaulAngus:
@borisstoyanov what is your LGTM based on?

borisstoyanov:
@PaulAngus, I've did a code review, and noticed that the errors reported in the last run are repeating with other PRs.

DaanHoogland:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Test LGTM, failing tests are due to known issues or intermittent in nature, but not related to KVM agent/libvirt resource class changes.

PR #2720, opened 2018-06-21
Disable creation of new Basic Networking Zones

wido:
Advanced Networking with a Shared Network and VLAN isolation offers the same and even more functionality as a Basic Zone does.

Therefor Basic Networking can be deprecated in favor of Advanced Networking to reduce the amount of code and testing.

Existing Basic Networking zones will keep functioning, but by preventing new ones from beving created we can remove the code in the future when the time is right.

### DIFF
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -2114,41 +2114,16 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
         // Validate zone
         final DataCenterVO zone = _dcDao.findById(zoneId);
         if (zone.getNetworkType() == NetworkType.Basic) {
-            // In Basic zone the network should have aclType=Domain, domainId=1, subdomainAccess=true
-            if (aclType == null || aclType != ACLType.Domain) {
-                throw new InvalidParameterValueException("Only AclType=Domain can be specified for network creation in Basic zone");
-            }
+            /*
+                Basic Networking zones are deprecated are no longer supported for creation
 
-            // Only one guest network is supported in Basic zone
-            final List<NetworkVO> guestNetworks = _networksDao.listByZoneAndTrafficType(zone.getId(), TrafficType.Guest);
-            if (!guestNetworks.isEmpty()) {
-                throw new InvalidParameterValueException("Can't have more than one Guest network in zone with network type " + NetworkType.Basic);
-            }
+                Advanced Networking with a Shared Network with Security Grouping enabled offers the same and even
+                more functionality to end-users.
 
-            // if zone is basic, only Shared network offerings w/o source nat service are allowed
-            if (!(ntwkOff.getGuestType() == GuestType.Shared && !_networkModel.areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SourceNat))) {
-                throw new InvalidParameterValueException("For zone of type " + NetworkType.Basic + " only offerings of " + "guestType " + GuestType.Shared + " with disabled "
-                        + Service.SourceNat.getName() + " service are allowed");
-            }
-
-            if (domainId == null || domainId != Domain.ROOT_DOMAIN) {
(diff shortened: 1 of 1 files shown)


wido:
This is a WIP so we can have further discussions about this :)

wido:
Ping @DaanHoogland @GabrielBrascher @rafaelweingartner @mike-tutkowski @rhtyd

rhtyd:
As part of larger refactoring, first it needs to be determined how removal of basic type zone affects non-KVM environment especially VMware and XenServer. If in those non-KVM cases, they can too have a equivalent basic zone type zone (in adv zone setting) and a painless upgrade path can be created then this is a worthy effort otherwise it may not be worth exploring and for one's personal deployment preference they can simply deploy an adv zone (with KVM) with/without SG to have shared networks that gives what a basic zone would provide.

DaanHoogland:
I thought the proposal was to disable creation of new basic zones and not to remove supporting code in the first go @rhtyd. So we do not need to immediately create an upgrade patch either.
The code looks good out of context. Do I need to go through the creatGuestNetwork completely, @wido ? and did you test it, by ignoring the obvious?
One ommision I see is the wizard. You did disable it in that, yet.

marcaurele:
You should do the migration path directly otherwise you'll end up with a code base for basic networking which isn't tested anymore due to the fact that you cannot create a test scenario. Leaving the code behind in CS, which has already a lot of dead code, isn't a good option at all for me. IMO, you must do a full code removal for the basic networking with the migration path.

wido:
@DaanHoogland The idea was indeed to block the creation of new networks, not disable them.

To also answer @marcaurele , we have a lot of if-else statements in the code where we make *assumptions* based on Basic/Advanced networking and many of them are not true.

All these if-statements need checking and that stays a thing which we have to tag along.

By blocking the creation of new Basic Zones we at least prevent them from being created. That does not mean we are dropping support for them, but we do not want to encourage people in creating them.

Converting them from Basic to Advanced during a upgrade will be very difficult.

marcaurele:
I understand @wido , my point is that if you block the creation you won't be able to test it. Then some code changes might generate bugs in the code for basic networking which won't be tested anymore (except by the one running it in production and still upgrading), thus degrading the support for basic networking to a point where it might so buggy that people will drop it, stop upgrading (?) ....

Maybe, it should state that version X drops the support for basic networking, which is more truthful than simply disabling it. This is more the long term vision of this change I can see, and I don't think we should lie here.

I can understand that creating an upgrade from basic to advanced might be difficult. We should see on the user mailing list who still need support for it to decide if it's worth the effort.

wido:
@marcaurele I understand the testing part.

But how about a message then in the UI where we by default select Advanced and discourage Basic Networking?

Now that gets to the UI part where I don't have much knowledge of.

GabrielBrascher:
A message warning that Basic will not be supported after CloudStack X.Y (for instance 5.0) looks good as the first step. We also can warn that CloudStack will block the creation of the basic network in CloudStack 4.12.

What do you think of the following roadmap draft?
- 4.11.2 [warning]
Warn short life for the basic network in UI as well as detail it on the documentation and release announcement.
- 4.12: [limited support]
Block basic network creation in the UI; however, those using basic network can still upgrade their CloudStack to 4.12
- 5.0: [end of life]
Remove all code related to the basic network

marcaurele:
I agree with @GabrielBrascher proposal, with the add-on that version 5.0 should hopefully have a sort of upgrade path for the simple to advance networking.

Can the announcement say how to create a basic networking installation using the advanced (which values to set and not set): basic networking is configured now in advance mode with those values....

GabrielBrascher:
@marcaurele I agree with you, it would be great to have a documentation specifying how to configure a "basic" advanced zone.

wido:
Ok, but this is now probably a GUI thing, right? Because in the GUI we should encourage people to create a Advanced Zone with 'Basic functions'.

Now I am far, far from a GUI expert, so I wouldn't know where to start even. The initial wizard and such would need to be changed I assume?

DaanHoogland:
me too, @wido but i agree; as a first step it should be a gui thing.

wido:
I will close this PR then.

I'll open a new thread on dev where we can discuss this.

Issue #2657, opened 2018-05-21
KVM hosts are being reset by a script


melnik13:
##### ISSUE TYPE

Bug Report

##### COMPONENT NAME

KVM

##### CLOUDSTACK VERSION

All 4.x and 2.x versions seem to be affected.

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

This issue is related to KVM hosts. As far as I can see, XEN hosts are affected too.

##### SUMMARY

KVM hosts are being rebooted by `kvmheartbeat.sh` when the NFS-server that is being used as a primary storage becomes unreachable.

I'd like to propose change behavior of `kvmheartbeat.sh`, as it's not necessary to reboot a whole host when only one of VMs is having difficulties while other VMs are working fine with other storages.

##### STEPS TO REPRODUCE

It's easy to reproduce: just deploy a VM with a volume on an NFS-based storage, break NFS and wait.

##### EXPECTED RESULTS

The KVM host will be reset.

##### ACTUAL RESULTS

The `kvmscript.sh` will reboot the server.

serverchief:
As of 4.11 CloudStack has an enhancement for KVM HA. Instead of relying on shell script and specific primary store - we will check the write time stamp on the disks for all VMs that are known to reside on hypervisor.

All VM disks/files must not update the write timestamp for an extended amount of time(configurable) before a specifc action is taken (configurable). Even If a single vm disk will have a write timestamp and all other don’t  - we will not power off the host.

You can read more about KVM HA on cloudstack homepage, there is also a video talk by Rohit Yadav that explains it in great details.

melnik13:
Thank you! I didn't notice that 4.11.x have it fixed, in 4.9.2 it still can reset a host.

In any case, the script and `echo b > /proc/sysrq-trigger` is still present. Won't it be executed at some moment? Looks a bit dangerous, as for me.

wido:
The main problem is that it is hard to determine if the host isn't responding or that something else is going on.

You do not want the VM to run twice as that might (will!) cause data corruption. That's why a hard reboot is being issued if the storage is having trouble.

melnik13:
Thank you for notation.

That's exactly why I proposed (d82162565c99889731f508ac72b4a19922520adf) to find the VMs using the mountpoint (by `lsof` and `awk`) and then kill their `qemu-kvm` processes.

Slair1:
Please see PR #2722 and #2472 
I agree and created these PRs to address the built-in reboot

[...]

PR #2722, opened 2018-06-25
CLOUDSTACK-10310 Fix KVM reboot on storage issue

Slair1:
Rebase of #2472 onto 4.11

If the KVM heartbeat file can't be written to, the host is rebooted, and thus taking down all VMs running on it. The code does try 5x times before the reboot, but the there is not a delay between the retires, so they are 5 simultaneous retries, which doesn't help. Standard SAN storage HA operations or quick network blip could cause this reboot to occur.

Some discussions on the dev mailing list revealed that some people are just commenting out the reboot line in their version of the CloudStack source.

A better option would be have it sleep between tries so it isn't 5x almost simultaneous tries. Plus, instead of rebooting, the cloudstack-agent could just be stopped on the host instead. This will cause alerts to be issued and if the host is disconnected long-enough, depending on the HA code in use, VM HA could handle the host failure.

The built-in reboot of the host seemed drastic

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHABase.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHABase.java
@@ -34,7 +34,8 @@ public class KVMHABase {
     protected static String s_heartBeatPath;
     protected long _heartBeatUpdateTimeout = 60000;
     protected long _heartBeatUpdateFreq = 60000;
-    protected long _heartBeatUpdateMaxRetry = 3;
+    protected long _heartBeatUpdateMaxTries = 5;
+    protected long _heartBeatUpdateRetrySleep = 15000;
 
     public static enum PoolType {
         PrimaryStorage, SecondaryStorage
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
@@ -119,7 +119,8 @@ public class KVMHAMonitor extends KVMHABase implements Runnable {
                     }
 
                     String result = null;
-                    for (int i = 0; i < 5; i++) {
+                    // Try multiple times, but sleep in between tries to ensure it isn't a short lived transient error
+                    for (int i = 1; i <= _heartBeatUpdateMaxTries; i++) {
                         Script cmd = new Script(s_heartBeatPath, _heartBeatUpdateTimeout, s_logger);
                         cmd.add("-i", primaryStoragePool._poolIp);
                         cmd.add("-p", primaryStoragePool._poolMountSourcePath);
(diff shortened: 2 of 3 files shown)


Slair1:
Looks like there is an additional Issue open #2657

borisstoyanov:
@blueorangutan test

borisstoyanov:
@Slair1 why not master instead of 4.11? 
4.11.1 just went out.

Slair1:
@borisstoyanov I didn't know if there was going to be another 4.11.x release or not.  If we prefer, i can rebase to master?

DaanHoogland:
@Slair1 this is a bug fix. It can go into 4.11 and if we merge we also merge forward to master

csquire:
This PR doesn't seem to completely fix the reboot problem. We installed the RC release with this PR on a test system and are still able to get the KVM host to reboot by using iptables to drop outgoing requests to NFS. My investigation shows that the line [`storage = conn.storagePoolLookupByUUIDString(uuid);`](https://github.com/apache/cloudstack/blob/4.11/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java#L95) blocks indefinitely.  So, `kvmheartbeat.sh` is never executed, a host investigation is started, the host with blocked NFS is marked as `Down` and finally all VMs on that host are rescheduled and result in duplicate VMs.

I pulled a thread dump and found the KVMHAMonitor thread will hang here until NFS is unblocked, didn't dig any deeper yet though.

```"Thread-20" - Thread t@135
   java.lang.Thread.State: RUNNABLE
        at com.sun.jna.Native.invokePointer(Native Method)
        at com.sun.jna.Function.invokePointer(Function.java:470)
        at com.sun.jna.Function.invoke(Function.java:404)
        at com.sun.jna.Function.invoke(Function.java:315)
        at com.sun.jna.Library$Handler.invoke(Library.java:212)
        at com.sun.proxy.$Proxy3.virStoragePoolLookupByUUIDString(Unknown Source)
        at org.libvirt.Connect.storagePoolLookupByUUIDString(Unknown Source)
        at com.cloud.hypervisor.kvm.resource.KVMHAMonitor$Monitor.runInContext(KVMHAMonitor.java:95)
        - locked <1afb3370> (a java.util.concurrent.ConcurrentHashMap)

   Locked ownable synchronizers:
        - None```

rhtyd:
Good find @csquire, @Slair1 any comments? /cc @PaulAngus @borisstoyanov @DaanHoogland

csquire:
Sorry, I misspoke in my last comment (edited to make it correct). The blocked host doesn't reboot, it just gets marked as `Down` and the VMs are actually still running on it when duplicate VMs get provisioned. Maybe it's a completely separate issue, but will still prevent us from using 4.11 in production.

Slair1:
@csquire The VMs should eventually crash themselves since they won't have access to their disks.  In our environment when a host gets marked as down we get alerted and an engineer can take a look at what is going on.  Yes, we rely on VM HA to decide what to do with the problem host also.  We make sure lockd is enabled so we don't get a split-brain situation.

https://libvirt.org/locking-lockd.html

csquire:
@Slair1 thanks for the feedback, I will run more tests.  The behavior is still incorrect, so I went ahead and submitted issue #2890

somejfn:
@Slair1   Testing with virtlockd,  if the host that crashed can't come back I had to manually clear the locks for VM HA to fire and restart the VM.  At that point it's the same as disabling VM HA and manually restart VM's after a KVM host crash.    How is your setup different ?

Slair1:
@somejfn maybe because I have PR #2474 also installed?

rhtyd:
I should n't have taken word on this, I tested and could not reproduce the behaviour @csquire reported. I think we should revert the change and let the KVM host reset. VM disk corruption is worse than VM downtime.

PaulAngus:
IMHO i'd say if a VM on that storage is marked as ha-enabled,  it should be powered-off and restarted somewhere else, and if it isn't HA enabled, we shouldn't do anything with the running VM (as it's for the user of the VM to deal with it),
in either case we should probably set the host to 'alert' so then an admin can see it and do something about it.

PR #2765, opened 2018-07-24
cleanup: Remove unused variables

wido:
Found some variables throughout the code which are not being used.
They can be removed.


DaanHoogland:
i think we can merge this one if travis agrees (no smoke tests needed)

GabrielBrascher:
I agree. The Maven could build the project from the source without any problem :+1:

PR #2775, opened 2018-07-26
Fix typo in ISO url

khos2ow:
## Description

The correct ISO url is http://mirror.nbrc.ac.in/centos/7/isos/x86_64/CentOS-7-x86_64-Minimal-1804.iso and not ends with `1808`.

## Types of changes

## Checklist:

Testing

### DIFF
--- a/tools/appliance/builtin/template.json
+++ b/tools/appliance/builtin/template.json
@@ -23,9 +23,9 @@
       "disk_interface": "virtio",
       "net_device": "virtio-net",
 
-      "iso_url": "http://mirror.nbrc.ac.in/centos/7/isos/x86_64/CentOS-7-x86_64-Minimal-1808.iso",
+      "iso_url": "http://mirror.nbrc.ac.in/centos/7/isos/x86_64/CentOS-7-x86_64-Minimal-1804.iso",
       "iso_checksum": "13675c6f74880e7ff3481b91bdaf925ce81bda8f",
-      "iso_checksum_type": "md5",
+      "iso_checksum_type": "sha1",
 
       "vm_name": "builtin",
       "output_directory": "../dist",


tools/appliance/builtin/template.json
```diff
@@ -23,7 +23,7 @@
       "disk_interface": "virtio",
       "net_device": "virtio-net",
 
-      "iso_url": "http://mirror.nbrc.ac.in/centos/7/isos/x86_64/CentOS-7-x86_64-Minimal-1808.iso",
+      "iso_url": "http://mirror.nbrc.ac.in/centos/7/isos/x86_64/CentOS-7-x86_64-Minimal-1804.iso",
```
rafaelweingartner: Is this hash correct?

I took the hash of that file and it is something else:
```
$ wget http://mirror.nbrc.ac.in/centos/7/isos/x86_64/CentOS-7-x86_64-Minimal-1804.iso
$ md5sum CentOS-7-x86_64-Minimal-1804.iso
fabdc67ff3a1674a489953effa285dfd  CentOS-7-x86_64-Minimal-1804.iso
```

khos2ow: Actually I was just about to send the exact comment. The official has only `sha1` and `sha256` hashes (which is used in `systemvmtemplate/template.json`) but here the `iso_checksum_type` is set to `md5`. Either of those needs to be changed.

rafaelweingartner: if the type can be set to sha1 or sha256, please do so. Otherwise, we need to configure the right hash value there.

khos2ow:
I updated the checksum type to `sha1` and double checked that the has is actually valid, as well as the URL.

Issue #2777, opened 2018-07-27
Cannot add new host with bonded interface.


bennysp:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
cloudstack-agent
~~~

##### CLOUDSTACK VERSION

~~~
4.11
~~~

##### CONFIGURATION

1 server (using quicksetup)

##### OS / ENVIRONMENT

CentOS Linux release 7.5.1804 (Core)

##### SUMMARY

When I am going through the Setup Wizard, it gets to the host install/setup and fails.  When I check the agent setup log, this is what I see.

```
[root@csserv01 ~]# tail -1000f /var/log/cloudstack/agent/setup.log
DEBUG:root:execute:hostname -f
DEBUG:root:execute:selinuxenabled
DEBUG:root:execute:setenforce 0
DEBUG:root:cloudbr0 is not a network device, is it down?
DEBUG:root:execute:route -n|awk '/^0.0.0.0/ {print $2,$8}'
DEBUG:root:execute:ifconfig bond0
DEBUG:root:[Errno 2] No such file or directory
  File "/usr/lib64/python2.7/site-packages/cloudutils/serviceConfig.py", line 38, in configration
    result = self.config()
  File "/usr/lib64/python2.7/site-packages/cloudutils/serviceConfig.py", line 309, in config
    super(networkConfigRedhat, self).cfgNetwork()
  File "/usr/lib64/python2.7/site-packages/cloudutils/serviceConfig.py", line 108, in cfgNetwork
    device = self.netcfg.getDefaultNetwork()
  File "/usr/lib64/python2.7/site-packages/cloudutils/networkConfig.py", line 53, in getDefaultNetwork
    pdi = networkConfig.getDevInfo(dev)
  File "/usr/lib64/python2.7/site-packages/cloudutils/networkConfig.py", line 157, in getDevInfo
    elif networkConfig.isBridge(dev) or networkConfig.isOvsBridge(dev):

DEBUG:root:execute:setenforce 1
```

##### STEPS TO REPRODUCE

 

~~~
Setup a bond0 network between 2 or more NICs on the server.

Follow the quickstart guide.  When you get to the step of the wizard "Launch", everything will go through except the error on the host setup. 
~~~

##### EXPECTED RESULTS

~~~
That it finds the bond0 interface and continues to setup
~~~

##### ACTUAL RESULTS

~~~
Host fails and error below is produced...

[...]

PR #2779, opened 2018-07-29
Changed to use ifcfg library for cross compatibility of OS /version

bennysp:
## Description

Changed to use pip library [ifcfg](https://github.com/ftao/python-ifcfg) instead of OS ifconfig due to ifconfig format changes across OS and versions. 

Current bug request:
https://github.com/apache/cloudstack/issues/2777

Install CentOS 7;  Make sure `yum update -y` has been performed;  At the point of agent setup, review setup.log and you will notice there are no ip address, netmask or broadcast set.

## Types of changes

## GitHub Issue/PRs

Fixes:#2777
## How Has This Been Tested?

1. Installed fresh Centos 7 (1804)
2. Followed "quick start guide" on one server
3. Got to the "add a host" section and it would fail. 

*Note:* Separate issue, but I also had to install, at minimum, `yum install openvswitch.x86_64 -y` or else the code (line 109) will fail when it tries to use the command `ovs-vsctl` in this same file.

After swapping out these PR changes and installing openvswitch, I was finally able to add my host correctly.

This does not affect other areas of the code, but forces you to install an external library python-ifcfg.

## Checklist:

Testing

### DIFF
--- a/python/lib/cloudutils/networkConfig.py
+++ b/python/lib/cloudutils/networkConfig.py
@@ -20,6 +20,8 @@ import logging
 import os
 import re
 import subprocess
+import ifcfg
+import json
 
 class networkConfig:
     class devInfo:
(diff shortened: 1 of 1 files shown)


borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

bennysp:
@borisstoyanov - looks like you guys already tested on centos6 before I could answer!  :)

borisstoyanov:
actually thats just the build of packages, let me run the smoketests as well @bennysp 
@blueorangutan test centos6 kvm-centos6

rhtyd:
@bennysp have you tested this on ubuntu? /cc @wido

wido:
Where do we actually install ifcfg? And it's a third party library, do we want to use that in our code?

bennysp:
@borisstoyanov the “old way” doesn’t actually parse correctly because of “ifconfig” format changes happening. So, you actually don’t get an IP address, broadcast or MAC. 

If you don’t want to include a 3rd party, then most will tell you to use “ip” vs “ifconfig” for one reason being the format changes.

borisstoyanov:
Thanks @bennysp, so why don't we refactor the code to use 'ip' instead of introducing another 3rd party dependency?

bennysp:
Will do guys!  Give me a few. Should I update this PR when ready or create new?

borisstoyanov:
thanks @bennysp, don't get me wrong, but the reason I'm -1 on adding a new dependency is because I'm already having errors when using our automated deployments to test this and I need to change that to add the new lib. And I do believe I'm not the only one :)

borisstoyanov:
@bennysp I would prefer if we close this and do a separate one, thanks.

wido:
Thanks @bennysp ! Yes, you can update this one.

Awesome!

rhtyd:
Dependencies are missing, as @wido suggested consume output of `ip`, `ifconfig` is not available on most new distros now.

PR #2791, opened 2018-08-06
router: Fixes #2789 fix proper mark based packet routing across interfaces

### LINKED ISSUE #2789: Incorrect routing behaviour for link-local interface seen in XenServer/VPC VR
VPC VR on XenServer (7.0) fails to start after changing of system offering for the router. On investigation it was found that packets in the VR for eth0 (link-local interface) get routed to eth1 instead of eth0. This is due to incorrect eth1 routing table rules in `Table_eth0` in the VR, the current workaround is to: (1) use a system/router offering as part of the vpc/network offering so admin don't need to change it after deploying a VPC, (2) after VR is started remove eth1 related default routes in the `Table_eth0` table.

The more appropriate fix will go into `update_config.py` and other `cs` modules which configure routes for an interface. The routing table should be appropriately configured. `fwmark` should be setup for eth0 linking table 0 (Table_eth0) so packets from eth0, are routed back to eth0 for the 169.254.0.0/16 destination.

##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION
~~~
4.11.1.0
~~~

##### CONFIGURATION

Adv networking, XenServer 7.0, VPC.

##### STEPS TO REPRODUCE

~~~
- Deploy a VPC VR with default offering
- Create a new system offering for router
- Stop VPC VR, change to the new system offering for router
- Start the VR, it fails (ip_association related json application fails due to ssh timeout issue)
~~~

##### EXPECTED RESULTS

~~~
VR should start!
~~~

##### ACTUAL RESULTS

~~~
VR fails to start
~~~

rhtyd:
Previously, the ethernet device index was used as `rt_table` index and
packet marking id/integer. With eth0 that is sometimes used as link-local
interface, the rt_table index `0` would fail as `0` is already defined
as a catchall (unspecified). The fwmarking on packets on eth0 with 0x0
would also fail. This fixes the routing issues, by adding 100 to the
ethernet device index so the value is a non-zero, for example then the
relationship between rt_table index and ethernet would be like:

100 -> Table_eth0 -> eth0 -> fwmark 100 or 0x64
101 -> Table_eth1 -> eth1 -> fwmark 101 or 0x65
101 -> Table_eth2 -> eth2 -> fwmark 102 or 0x66

This would maintain the legacy design of routing based on packet mark
and appropriate routing table rules per table/ids. This also fixes a
minor NPE issue around listing of snapshots.

## Types of changes

## Checklist:

Testing

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsNetfilter.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsNetfilter.py
@@ -189,7 +189,8 @@ class CsNetfilters(object):
     def add_chain(self, rule):
         """ Add the given chain if it is not already present """
         if not self.has_chain(rule.get_table(), rule.get_chain()):
-            CsHelper.execute("iptables -t %s -N %s" % (rule.get_table(), rule.get_chain()))
+            if rule.get_chain() != "":
+                CsHelper.execute("iptables -t %s -N %s" % (rule.get_table(), rule.get_chain()))
             self.chain.add(rule.get_table(), rule.get_chain())
 
     def del_standard(self):
--- a/systemvm/debian/opt/cloud/bin/cs/CsNetfilter.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsNetfilter.py
@@ -189,7 +189,7 @@ class CsNetfilters(object):
     def add_chain(self, rule):
         """ Add the given chain if it is not already present """
         if not self.has_chain(rule.get_table(), rule.get_chain()):
-            if rule.get_chain() != "":
+            if rule.get_chain():
                 CsHelper.execute("iptables -t %s -N %s" % (rule.get_table(), rule.get_chain()))
             self.chain.add(rule.get_table(), rule.get_chain())
 
--- a/server/src/com/cloud/api/ApiResponseHelper.java
+++ b/server/src/com/cloud/api/ApiResponseHelper.java
@@ -501,7 +501,7 @@ public class ApiResponseHelper implements ResponseGenerator {
                 snapshotResponse.setZoneId(zone.getUuid());
             }
 
-            if (volume.getVolumeType() == Volume.Type.ROOT) {
+            if (volume.getVolumeType() == Volume.Type.ROOT && volume.getInstanceId() != null) {
                 //TODO combine lines and 489 into a join in the volume dao
                 VMInstanceVO instance = ApiDBUtils.findVMInstanceById(volume.getInstanceId());
                 if (instance != null) {
--- a/systemvm/debian/opt/cloud/bin/cloud-nic.sh
+++ b/systemvm/debian/opt/cloud/bin/cloud-nic.sh
@@ -69,7 +69,7 @@ unplug_nic() {
 
 action=$1
 dev=$2
-tableNo=${dev:3}
+tableNo=$((100+${dev:3}))
 tableName="Table_$dev"
 
 if [ $action == 'add' ]
--- a/systemvm/debian/opt/cloud/bin/cs/CsRoute.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsRoute.py
@@ -30,7 +30,7 @@ class CsRoute:
         return self.table_prefix + name
 
     def add_table(self, devicename):
-        tablenumber = devicename[3:]
+        tablenumber = 100 + int(devicename[3:])
         tablename = self.get_tablename(devicename)
         str = "%s %s" % (tablenumber, tablename)
         filename = "/etc/iproute2/rt_tables"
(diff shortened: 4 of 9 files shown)


rhtyd:
This is after the fix on a VPC VR:

Interface listing:
```
root@r-20-VM:~# ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 0e:00:a9:fe:03:e2 brd ff:ff:ff:ff:ff:ff
    inet 169.254.3.226/16 brd 169.254.255.255 scope global eth0
       valid_lft forever preferred_lft forever
3: eth1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether 1e:00:15:00:00:04 brd ff:ff:ff:ff:ff:ff
    inet 10.1.35.24/19 brd 10.1.63.255 scope global eth1
       valid_lft forever preferred_lft forever
```

`rt_tables` entry:
```
root@r-20-VM:~# cat /etc/iproute2/rt_tables
#
# reserved values
#
255	local
254	main
253	default
0	unspec
#
# local
#
#1	inr.ruhep
100 Table_eth0
252 static_route
251 static_route_back
101 Table_eth1
```
ip rules:
```
root@r-20-VM:~# ip rule
0:	from all lookup local 
32762:	from all fwmark 0x65 lookup Table_eth1 
32763:	from 10.20.0.0/16 lookup static_route_back 
32764:	from 10.20.0.0/16 lookup static_route 
32765:	from all fwmark 0x64 lookup Table_eth0 
32766:	from all lookup main 
32767:	from all lookup default 
root@r-20-VM:~# ip rule show table Table_eth0
32765:	from all fwmark 0x64 lookup Table_eth0 
root@r-20-VM:~# ip rule show table Table_eth1
32762:	from all fwmark 0x65 lookup Table_eth1 
```

Routing rules:
```
root@r-20-VM:~# ip route
default via 10.1.63.254 dev eth1 
10.1.32.0/19 dev eth1 proto kernel scope link src 10.1.35.24 
169.254.0.0/16 dev eth0 proto kernel scope link src 169.254.3.226 
root@r-20-VM:~# ip route show table Table_eth0
root@r-20-VM:~# ip route show table Table_eth1
default via 10.1.63.254 dev eth1 proto static 
throw 10.1.32.0/19 proto static 
```

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
Tested on KVM and XenServer, works alright.
Steps to reproduce issue:
- Deploy a VPC VR
- Stop the VPC VR (optionally change the router offering)
- Start the VPC VR (it should fail with SSH timeout fails without the patch)
- Ping/ssh the link-local interface from hypervisor/host (it should fail after stop+start of VR without this fix)

In case of VMware, eth0 is used as private/mgmt IP (not link-local) so it works in my env but could fail other environments. The same VR fix would apply as the fix in rt_tables/routing is agnostic of hypervisor.
/cc @borisstoyanov

systemvm/debian/opt/cloud/bin/cs/CsNetfilter.py
```diff
@@ -189,7 +189,8 @@ def compare(self, list):
     def add_chain(self, rule):
         """ Add the given chain if it is not already present """
         if not self.has_chain(rule.get_table(), rule.get_chain()):
-            CsHelper.execute("iptables -t %s -N %s" % (rule.get_table(), rule.get_chain()))
+            if rule.get_chain() != "":
```
DaanHoogland: so when rule.get_chain() != self.chain && rule.get_chain() != "", we still add it to self. chain?

rhtyd: I found an error in logs where it was trying to add an empty chain: "iptables -t filter -N"
I did not debug further as to why an empty chain gets added to `self.chain`. However, removing the case completely (the self.chain add) created several other failures/side-effects so I kept the change minimal.

rhtyd:
@blueorangutan test matrix

systemvm/debian/opt/cloud/bin/cs/CsNetfilter.py
wido: I would use:

<pre>
if rule.get_chain():
</pre>

This would check for None and empty strings in Python

rhtyd:
Vmware, XenServer tests LGTM. I've backported some fixes we did in smoke test for KVM from master, let me rekick tests for KVM.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Thanks for verification @borisstoyanov, will merge based on test results and reviews.

Issue #2763, opened 2018-07-24
Host is still unsecured when provisionCertificate api is called


borisstoyanov:
##### SUMMARY

When provisionCertificate api is executed the host ends up in unsecured state, instead should be secured

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API, agent
~~~

##### CLOUDSTACK VERSION

~~~
4.12
~~~

##### CONFIGURATION

ca.framework.cert.automatic.renewal = true
ca.plugin.root.allow.expired.cert = true
ca.plugin.root.auth.strictness = false

##### OS / ENVIRONMENT

NA

##### STEPS TO REPRODUCE

~~~
execute provisionCertificate api with a random kvm host with/out recconect=true, the api call is successful but connection is unsecured. 
~~~

##### EXPECTED RESULTS

~~~
Connection is secured 
~~~

##### ACTUAL RESULTS

~~~
Connection is unsecured 

~~~

[agent_unsecured.txt](https://github.com/apache/cloudstack/files/2222393/agent_unsecured.txt)
[unsecured_connection.txt](https://github.com/apache/cloudstack/files/2222394/unsecured_connection.txt)

borisstoyanov:
ping @rhtyd and @dhlaluku. Could this be related to these changes: #2756

dhlaluku:
@borisstoyanov I do not think so, that small fix that I did was for the NPE issue. The VM lifecycle tests have already been failing before and I think it might be due to regressions in provisionCertificate API itself that need looking into. Perhaps @rhtyd can help us here since he authored the Secure live migration feature.

PR #2756, opened 2018-07-19
fix provisionCertificate api returns NPE when 'reconnect' parameter is true

dhlaluku:
## Description

This PR fixes NPE with the provisionCertificateCmd when reconnect is set to True.

Also fixes the following Marvin test failures:
- test_certauthority_root.py

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## GitHub Issue/PRs

Fixes: #2753 

## Screenshots (if appropriate):
image

## How Has This Been Tested?
With cloudmonkey, see screenshot

Dev environment components:

- Platform: ACS-4.12
- Management OS: Ubuntu 18.04 LTS
- Hypervisor: 1 Host running KVM with CentOS 7

## Checklist:

- [x] I have read the [CONTRIBUTING](https://github.com/apache/cloudstack/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
Testing
- [ ] I have added tests to cover my changes.
- [x] All relevant new and existing integration tests have passed.
- [ ] A full integration testsuite with all test that can run on my environment has passed.

PR #2754, opened 2018-07-19
Changing to the right api call for securing hosts

borisstoyanov:
## Description

This fixes the tests for secure live migration, looks like they were pointing to the wrong api call. However during testing of this I've hit a new issue #2753 which is blocking the tests. 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## GitHub Issue/PRs
The tests will pass when #2753 is fixed, since they are using a parameter in the provisionCertificate api call which makes a NPE

## Screenshots (if appropriate):
NA
## How Has This Been Tested?
Locally with a debugger, api call seem right but it hits the mentioned issue

## Checklist:

- [x] I have read the [CONTRIBUTING](https://github.com/apache/cloudstack/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
Testing
- [x] I have added tests to cover my changes.
- [ ] All relevant new and existing integration tests have passed.
- [ ] A full integration testsuite with all test that can run on my environment has passed.

PR #2795, opened 2018-08-08
kvm: Properly report available memory to Management Server

wido:
## Description
The KVM Agent had two mechanisms for reporting its capabilities
and memory to the Management Server.

On startup it would ask libvirt the amount of Memory the Host has
and subtract and add the reserved and overcommit memory.

When the HostStats were however reported to the Management Server
these two configured values on the Agent were no longer reported
in the statistics thus showing all the available memory in the
Agent/Host to the Management Server.

This commit unifies this by using the same logic on Agent Startup
and during statistics reporting.

  memory=3069636608, reservedMemory=1073741824

This was reported by a 4GB Hypervisor with this setting:

  host.reserved.mem.mb=1024

The GUI (thus API) would then show:

  Memory Total	2.86 GB

This way the Agent properly 'lies' to the Management Server about its
capabilities in terms of Memory.

This is very helpful if you want to overprovision or undercommit machines
for various reasons.

Overcommitting can be done when KSM or ZSwap or a fast SWAP device is
installed in the machine.

Underprovisioning is done when the Host might run other tasks then a KVM
hypervisor, for example when it runs in a hyperconverged setup with Ceph.

In addition internally many values have been changed from a Double to a Long
and also store the amount of bytes instead of Kilobytes.

## Types of changes

## How Has This Been Tested?

I tested this on a CloudStack 4.12 test setup running on my local system.

The KVM Agent will report the amount of memory it has minus any reservations. This way the Mgmt server will never know the exact amount of memory the Agent has.

## Checklist:
Testing

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -19,9 +19,7 @@ package com.cloud.hypervisor.kvm.resource;
 import java.io.BufferedReader;
 import java.io.File;
 import java.io.FileNotFoundException;
-import java.io.FileReader;
 import java.io.IOException;
-import java.io.Reader;
 import java.io.StringReader;
 import java.net.InetAddress;
 import java.net.URI;
@@ -55,12 +53,12 @@ import org.apache.cloudstack.storage.to.TemplateObjectTO;
 import org.apache.cloudstack.storage.to.VolumeObjectTO;
 import org.apache.cloudstack.utils.hypervisor.HypervisorUtils;
 import org.apache.cloudstack.utils.linux.CPUStat;
+import org.apache.cloudstack.utils.linux.KVMHostInfo;
 import org.apache.cloudstack.utils.linux.MemStat;
 import org.apache.cloudstack.utils.qemu.QemuImg.PhysicalDiskFormat;
 import org.apache.cloudstack.utils.security.KeyStoreUtils;
 import org.apache.commons.collections.MapUtils;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.math.NumberUtils;
 import org.apache.log4j.Logger;
@@ -311,8 +309,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     protected int _cmdsTimeout;
     protected int _stopTimeout;
     protected CPUStat _cpuStat = new CPUStat();
-    protected MemStat _memStat = new MemStat();
-
+    protected MemStat _memStat = new MemStat(_dom0MinMem, _dom0OvercommitMem);
     private final LibvirtUtilitiesHelper libvirtUtilitiesHelper = new LibvirtUtilitiesHelper();
 
     @Override
@@ -871,7 +868,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
 
         value = (String)params.get("host.reserved.mem.mb");
         // Reserve 1GB unless admin overrides
-        _dom0MinMem = NumbersUtil.parseInt(value, 1024) * 1024 * 1024L;
+        _dom0MinMem = NumbersUtil.parseInt(value, 1024) * 1024* 1024L;
 
         value = (String)params.get("host.overcommit.mem.mb");
         // Support overcommit memory for host if host uses ZSWAP, KSM and other memory
@@ -2661,12 +2658,14 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     @Override
     public StartupCommand[] initialize() {
 
-        final List<Object> info = getHostInfo();
+        final KVMHostInfo info = new KVMHostInfo(_dom0MinMem, _dom0OvercommitMem);
+
+        final String capabilities = String.join(",", info.getCapabilities());
 
         final StartupRoutingCommand cmd =
-                new StartupRoutingCommand((Integer)info.get(0), (Long)info.get(1), (Long)info.get(2), (Long)info.get(4), (String)info.get(3), _hypervisorType,
+                new StartupRoutingCommand(info.getCpus(), info.getCpuSpeed(), info.getTotalMemory(), info.getReservedMemory(), capabilities, _hypervisorType,
                         RouterPrivateIpStrategy.HostLocal);
-        cmd.setCpuSockets((Integer)info.get(5));
+        cmd.setCpuSockets(info.getCpuSockets());
         fillNetworkInformation(cmd);
         _privateIp = cmd.getPrivateIpAddress();
         cmd.getHostDetails().putAll(getVersionStrings());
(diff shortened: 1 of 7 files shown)


dhlaluku:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/linux/HostInfo.java
```diff
@@ -0,0 +1,132 @@
...
+import org.libvirt.LibvirtException;
+import org.libvirt.NodeInfo;
+
+import java.io.FileReader;
+import java.io.IOException;
+import java.io.Reader;
+import java.util.ArrayList;
+import java.util.List;
+
+public class HostInfo {
```
rhtyd: Can you rename this class to KVMHostInfo, that way it will be more obvious what this class is expected of?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/linux/HostInfo.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static final Logger s_logger = Logger.getLogger(HostInfo.class);
```
rhtyd: Remove the s_, instead using `LOG` or `LOGGER`?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

wido:
@rhtyd Done, I've renamed the class and the logger variable. How does this look like?

wido:
Ping @RPDiep

dhlaluku:
@blueorangutan package

dhlaluku:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/linux/KVMHostInfo.java
```diff
@@ -0,0 +1,132 @@
...
+            this.cpusockets = hosts.sockets;
+            if (hosts.nodes > 0) {
+                this.cpusockets = hosts.sockets * hosts.nodes;
+            }
+            this.cpus = hosts.cpus;
+
+            final LibvirtCapXMLParser parser = new LibvirtCapXMLParser();
+            parser.parseCapabilitiesXML(conn.getCapabilities());
+            final ArrayList<String> oss = parser.getGuestOsType();
+            for (final String s : oss) {
```
rafaelweingartner: Why doing all of this process if you only report the host supporting HVM? I mean, KVM does not run if the CPU does not support HVM. Therefore, all KVM hosts will always have HVM support, right?

wido: Yes, they will. But what I did is take the original code and use it. I didn't modify the logic behind the capabilities.

I would make that a different PR if we want to do that.

GabrielBrascher:
@rhtyd do you know if the failing tests are false positives? I have been testing some actions such as VM migrations and adding/removing hosts. I had no problem when using primary storage with nfs and a cluster with KVM servers.

dhlaluku:
@GabrielBrascher most of the failing tests have already been addressed and merged into master with this PRs #2763 #2756 #2754 

@wido have this changes been rebased against the latest master?

wido:
@dhlaluku I've rebased against master

dhlaluku:
@blueorangutan test

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

wido:
The tests which were failing here do not seem related to the change. Is this one ready to merge?

dhlaluku:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

plugins/hypervisors/kvm/src/test/java/org/apache/cloudstack/utils/linux/KVMHostInfoTest.java
```diff
@@ -0,0 +1,34 @@
...
+
+import org.junit.Test;
+import org.junit.Assume;
+import org.mockito.Mockito;
+
+import org.libvirt.NodeInfo;
+
+public class KVMHostInfoTest {
+    @Test
+    public void getCpuSpeed() {
```
rafaelweingartner: A test case without an assert?

Also, why not mock things, and then run the test always? I mean, if somebody is using MacOS (not sure here) or Windows the test will be ignored.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/linux/MemStat.java
```diff
@@ -22,30 +22,47 @@
 import java.util.Map;
 import java.util.Scanner;
 
+
 public class MemStat {
+    /*
```
rafaelweingartner: Is this a documentation or a code comment?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/linux/KVMHostInfo.java
rafaelweingartner: I understand that you are only moving the old code around, but why not remove useless code now?

I mean, why would we keep a code that does not aggregate value, and can be misleading to newcomers?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -383,11 +378,11 @@ public LibvirtUtilitiesHelper getLibvirtUtilitiesHelper() {
     }
 
     public CPUStat getCPUStat() {
-        return _cpuStat;
+        return new CPUStat();
```
rafaelweingartner: Is this coded on purpose? I mean, can we use it as a stateless call? At first sight this `getCPUStat` seem to return an object property, but instead it will always return a different object instance.

rhtyd:
@blueorangutan package

wido:
It seems we are ready to merge this one?

rafaelweingartner:
I would say that I have unanswered questions...

wido:
I see @rafaelweingartner ! Missed those, I will look at them.

rhtyd:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -383,11 +378,11 @@ public LibvirtUtilitiesHelper getLibvirtUtilitiesHelper() {
...
     }
 
     public MemStat getMemStat() {
-        return _memStat;
+        return new MemStat(_dom0MinMem, _dom0OvercommitMem);
```
dhlaluku: Same comment as above, the returned object instance is always different for every method call

rhtyd:
@wido this is failing build, can you check?

wido:
I will @rhtyd 

Currently in the middle of VXLAN testing, so postponing this one for a moment. Need to pick it up soon and go through the comments.

wido:
@rafaelweingartner I think I addressed your comments.

I left the 'hvm' code in there as it does serve a purpose. If the node the Agent is started on does not have VT enabled in the BIOS we throw an error and also report to the mgmt server we are not HVM capable. So that's good.

Otherwise I also fixed the Unit Test, so I think we are good :)

rafaelweingartner:
forcing Jenkins

rafaelweingartner:
@wido Jenkins is not being able to build this PR. Can this be a dependency problem?
```
[ERROR] Errors: 
[ERROR]   KVMHostInfoTest.getCpu:39 » UnsatisfiedLink Unable to load library 'virt': Nat...
[ERROR]   KVMHostInfoTest.getMemory:47 » NoClassDefFound Could not initialize class org....
```

wido:
Jenkins is happy now. How are we with this PR? Can we merge it?

All comments seem to be adressed

rhtyd:
@wido let's test one last time
@blueorangutan package

rhtyd:
@wido can you check and fix build failures, see job 1 of travis

wido:
I just force pushed. This seems to be a thing (Unit Test) which fails on Travis, but works on my local system.

Could be a problem where tests are running inside a Container?

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

rafaelweingartner:
@wido we need to check the integration test results before being able to merge it.
Thanks for the PR.

rhtyd:
@blueorangutan test

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

Issue #2760, opened 2018-07-23
SystemVMs running in Xen HVM mode are not configured


dpassante:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
systemVM image 4.11 for XenServer
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1
~~~

##### CONFIGURATION

* Cloudstack 4.11.1
* XenServer 7.1.1
* SystemVM 4.11.1 Fri Jun 22 07:52:17 UTC 2018 running in HVM mode

##### OS / ENVIRONMENT

N/A

##### SUMMARY

Newly created system VMs are not configured.

Related to #2561 

virt-what cpuid helper detects Xen HVM guests with viridian extension enabled as hyperV guests.

~~~bash
root@v-2158-VM:~# /usr/lib/x86_64-linux-gnu/virt-what-cpuid-helper
Microsoft Hv
~~~

So, virt-what returns the following on Xen HVM SystemVMs:

~~~bash
root@v-2158-VM:~# virt-what
hyperv
xen
xen-domU
~~~

In this case, the ```/opt/cloud/bin/setup/cloud-early-config``` script assumes that we're running on a paravirtualized guest and fails to get boot args.

##### EXPECTED RESULTS

hypervisor: xen-hvm
systemvm type: should be detected.

~~~
Sun Jul 22 03:18:28 UTC 2018 Executing cloud-early-config
Sun Jul 22 03:18:28 UTC 2018 Detected that we are running inside xen-hvm
Sun Jul 22 03:18:28 UTC 2018 Scripts checksum detected: oldmd5=76c6bbddfc1d55d1b212ad8ca07190d4 newmd5=76c6bbddfc1d55d1b212ad8ca07190d4
Sun Jul 22 03:18:28 UTC 2018 Configuring systemvm type=consoleproxy
Sun Jul 22 03:18:29 UTC 2018 Setting up console proxy system vm
...
~~~

##### ACTUAL RESULTS

hypervisor: xen-domU
systemvm type : empty

~~~
root@systemvm:~# cat /var/log/cloud.log
Fri Jul 20 12:55:19 UTC 2018 Executing cloud-early-config
Fri Jul 20 12:55:19 UTC 2018 Detected that we are running inside xen-domU
[...]

PR #2824, opened 2018-08-28
Fix SystemVMs running in Xen HVM mode are not configured (#2760)

dpassante:
Set hypervisor to xen-hvm when virt-what detects both HyperV cpuid and xen-domU

## Types of changes

## GitHub Issue/PRs

Fixes: #2760 

## How Has This Been Tested?

Tested on Cloudstack 4.11.1 + XenServer 7.1.1 with a patched systemvm template.

## Checklist:

Testing

### DIFF
--- a/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
+++ b/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
@@ -35,6 +35,16 @@ log_it() {
 }
 
 hypervisor() {
+  if [ -d /proc/xen ]; then
+    mount -t xenfs none /proc/xen
+    $(dmesg | grep -q "Xen HVM")
+    if [ $? -eq 0 ]; then  # 1=PV,0=HVM
+      echo "xen-hvm" && return 0
+    else
+      echo "xen-pv" && return 0
+    fi
+  fi
+
   [ -x /usr/sbin/virt-what ] && local facts=( $(virt-what) )
   if [ "$facts" != "" ]; then
     # Xen HVM is recognized as Hyperv when Viridian extensions are enabled
@@ -47,16 +57,6 @@ hypervisor() {
 
   grep -q QEMU /proc/cpuinfo  && echo "kvm" && return 0
   grep -q QEMU /var/log/messages && echo "kvm" && return 0
-
-  [ -d /proc/xen ] && mount -t xenfs none /proc/xen
-  if [ -d /proc/xen ]; then
-    $(dmesg | grep -q "Xen HVM")
-    if [ $? -eq 0 ]; then  # 1=PV,0=HVM
-      echo "xen-hvm" && return 0
-    else
-      echo "xen-pv" && return 0
-    fi
-  fi
 
   vmware-checkvm &> /dev/null && echo "vmware" && return 0
 
--- a/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
+++ b/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
@@ -35,8 +35,15 @@ log_it() {
 }
 
 hypervisor() {
-  local try=$([ -x /usr/sbin/virt-what ] && virt-what | tail -1)
-  [ "$try" != "" ] && echo $try && return 0
+  [ -x /usr/sbin/virt-what ] && local facts=( $(virt-what) )
+  if [ "$facts" != "" ]; then
+    # Xen HVM is recognized as Hyperv when Viridian extensions are enabled
+    if [ "${facts[-1]}" == "xen-domU" ] && [ "${facts[0]}" == "hyperv" ]; then
+      echo "xen-hvm" && return 0
+    else
+      echo ${facts[-1]} && return 0
+    fi
+  fi
 
   grep -q QEMU /proc/cpuinfo  && echo "kvm" && return 0
   grep -q QEMU /var/log/messages && echo "kvm" && return 0


systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -35,8 +35,15 @@ log_it() {
 }
 
 hypervisor() {
-  local try=$([ -x /usr/sbin/virt-what ] && virt-what | tail -1)
-  [ "$try" != "" ] && echo $try && return 0
+  [ -x /usr/sbin/virt-what ] && local facts=( $(virt-what) )
```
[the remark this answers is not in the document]
dpassante: Good catch, according to the virt-what script, the returns should be:
Xen PV:
* ```xen``` + ```xen-domU``` 
* or just ```xen``` 

Xen HVM:
* ```xen``` + ```xen-hvm```

virt-what never returns ```xen-pv```.
I think the ```xen-pv``` case in the get_boot_params() function should be ```xen-pv|xen-domU|xen)``` to ensure that any PV VMs read their configuration from /proc/cmdline.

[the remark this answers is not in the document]
rafaelweingartner: Also, when I use virt-what in an HVM machine I get:
```
xen
xen-hvm
```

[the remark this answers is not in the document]
dpassante: virt-what always returns xen + xen-hvm on Xen HVM guests when it's cpuid helper detects ```XenVMMXenVMM``` signature.
 You can have a look [here](http://git.annexia.org/?p=virt-what.git;a=blob;f=virt-what.in;h=d037a992f57a4154d89007f0b93538f41425639e;hb=HEAD#l231) to see in details how virt-what detects Xen guests.

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -35,8 +35,15 @@ log_it() {
...
+  if [ "$facts" != "" ]; then
+    # Xen HVM is recognized as Hyperv when Viridian extensions are enabled
+    if [ "${facts[-1]}" == "xen-domU" ] && [ "${facts[0]}" == "hyperv" ]; then
```
rhtyd: We have support for `hyperv`, it may be wrong to return 'xen-hvm' when fact is hyperv? Can you fix this?

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -35,8 +35,15 @@ log_it() {
 }
 
 hypervisor() {
-  local try=$([ -x /usr/sbin/virt-what ] && virt-what | tail -1)
-  [ "$try" != "" ] && echo $try && return 0
+  [ -x /usr/sbin/virt-what ] && local facts=( $(virt-what) )
+  if [ "$facts" != "" ]; then
+    # Xen HVM is recognized as Hyperv when Viridian extensions are enabled
```
rhtyd: What is a Viridian extension?

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -71,7 +78,7 @@ config_guest() {
 
 get_boot_params() {
   case $HYPERVISOR in
-     xen-pv|xen-domU)
```
rhtyd: What if xen-domU is returned, removing the case would mean we no longer could process hypervisor where `xen-domU` would be returned.

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
dpassante: If I'm not mistaken, we will get ```xen-hvm``` only if the latest fact returned by virt-what is "xen-domU" when the first fact is ```hyperv```. On all other cases, we will get the latest fact.

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
dpassante: I'm not an expert about it but, as I understand, it's a Xen feature which speeds up Windows guests by providing some hyperV extensions. Viridian was the codename of HyperV.

dpassante:
I also ran virt-what (original and 1.15 versions) in an older PV SSVM (ACS 4.3) running on XenServer 6.2 and I get:
    xen
    xen-hvm

I'm sure the VM is a PV guest since the following lines are present in the dmesg:

    Booting paravirtualized kernel on Xen
    Xen: using vcpuop timer interface
    Initialising Xen virtual ethernet driver.

and there is no reference to ```HVM```.

Perhaps we should use the ```dmesg``` method that seems to be more accurate than virt-what for Xen guests?

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
rhtyd: In that case, please remove/revert the change @dpassante

rhtyd:
@dpassante sounds good to me if you want to grep `dmesg` for accuracy

rhtyd:
@DaanHoogland @nvazquez @PaulAngus @GabrielBrascher @wido @khos2ow can you please review this? I won't be able to test it without building a new systemvmtemplate, if we agree we can merge this based on review and run testing after #2829 is merged, revert if necessary.

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -51,6 +45,19 @@ hypervisor() {
...
+  if [ "$facts" != "" ]; then
+    # Xen HVM is recognized as Hyperv when Viridian extensions are enabled
+    if [ "${facts[-1]}" == "xen-domU" ] && [ "${facts[0]}" == "hyperv" ]; then
+      echo "xen-hvm" && return 0
+    else
+      echo ${facts[-1]} && return 0
+    fi
+  fi
+
+  grep -q QEMU /proc/cpuinfo  && echo "kvm" && return 0
```
DaanHoogland: I realize this is old code moved but it does not seem reliable or future proof. How about a call to virsh (--version or so)?

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -51,6 +45,19 @@ hypervisor() {
     fi
   fi
 
+  [ -x /usr/sbin/virt-what ] && local facts=( $(virt-what) )
+  if [ "$facts" != "" ]; then
+    # Xen HVM is recognized as Hyperv when Viridian extensions are enabled
+    if [ "${facts[-1]}" == "xen-domU" ] && [ "${facts[0]}" == "hyperv" ]; then
```
khos2ow: From XenServer only point of view, this block of code is redundant. Because result of `dmesg | grep -q "Xen HVM"` on HVM instance on XenServer, line 41, will work 100% of the time. The gotcha point of this function is to look for hvm on xen the first thing and if not a match let the rest of the function try to detect the _type_.

A while back I had the same issue and did a non-conclusive investigation about how `virt-what` works, and was actually getting all sorts of results/order on HVM and PV instance on Xen Server. Essentially the workaround I came up with the same as mentioned above.

rhtyd:
@khos2ow the changes are in cloud-early-config which patches using systemvm.iso, so the first time a systemvm boots it won't be able to patch and use the new cloud-early-config from a (new) `systemvm.iso`. On after 2+ boots, it will be able to use any new changes in cloud-early-config. Therefore a systemvmtemplate change is needed.

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
rhtyd: virsh is only the hypervisor host, not in the guest VM.

rhtyd:
Merging based on commentary, we'll proceed with testing a new systemvmtemplate from latest 4.11 branch and conclude next week if this needs to be reverted at all.

PR #2825, opened 2018-08-28
expunge if flag is set

DaanHoogland:
## Description

In integration work for CCS I found that the service call UserVmService.destroyVm(long uuid, boolean expunge) does not honour the expunge flag. I traced it down to the implementation VirtualMachineManagerImpl.destroy(String vmUuid, boolean expunge)

## Types of changes

## How Has This Been Tested?
manual testing so far, testing will pose some crosscutting challanges as the behaviour and implementation are seperated by about five layers of abstraction.

## Checklist:

Testing

### DIFF
--- a/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -1890,6 +1890,13 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
                     if (!stateTransitTo(vm, VirtualMachine.Event.DestroyRequested, vm.getHostId())) {
                         s_logger.debug("Unable to destroy the vm because it is not in the correct state: " + vm);
                         throw new CloudRuntimeException("Unable to destroy " + vm);
+                    } else {
+                        if (expunge) {
+                            if (!stateTransitTo(vm, VirtualMachine.Event.ExpungeOperation, vm.getHostId())) {
+                                s_logger.debug("Unable to expunge the vm because it is not in the correct state: " + vm);
+                                throw new CloudRuntimeException("Unable to expunge " + vm);
+                            }
+                        }
                     }
                 } catch (final NoTransitionException e) {
                     s_logger.debug(e.getMessage());


DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@rhtyd , isn't that the designed / envisioned way of changing? It didn't feel like forcing to me. it will refuse if it doesn't have a transition defined for that state/event combination.

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -1890,6 +1890,13 @@ public void doInTransactionWithoutResult(final TransactionStatus status) throws
                     if (!stateTransitTo(vm, VirtualMachine.Event.DestroyRequested, vm.getHostId())) {
```
rafaelweingartner: Is this change only about the change in the state in the DB? Or is that `stateTransitTo` going to execute the commands to actually remove the volume?

DaanHoogland: It is going to signal listeners through a publish-subscribe. I suppose one of them will somehow automagically call the hypervisor to destroy the VM. The new code below does the same for expunge, if expunge flag is set which is only so in one of the three places the method is called. The other two have a hardcoded false in place.

rafaelweingartner: Ok, thanks for the feedback. I still have some doubts though.

Why did you use the `else`?
I mean, instead of `ELSE/IF (expunge)`, you can simply use `if (expunge)`.

DaanHoogland: because I don't want to do the expunge when the destroy fails or is refused.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@rafaelweingartner @DaanHoogland do we have any outstanding issue on this?

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
rafaelweingartner: if it fails, or if it is refused (I have no idea what it means), an exception is thrown (I am assuming that for these cases the method will return false). Therefore, there is no need for the `else` block.

rhtyd:
@blueorangutan test

engine/orchestration/src/com/cloud/vm/VirtualMachineManagerImpl.java
DaanHoogland: the method returns a boolean and throws an exception when it returns false. I take action only on the true case. I guess it's a matter of style.

rafaelweingartner: but in the true case, we do not execute the IF body, and we proceed, which is the same as not using the ELSE block, and adding the IF(expunge) directly because the FALSE case will always throw an exception.

For it is not a matter of style, it is a matter of removing a nested condition IF/ELSE/IF, but sure we can leave it as is.

DaanHoogland: relying on a mid method exit to occur is not my style. I'd like to see structure and I do not think it is readable like that, either. sorry.

PR #2829, opened 2018-08-30
CLOUDSTACK-9473: storage pool capacity check when volume is resized or migrated

rhtyd:
This was previously opened as PR but prematurely closed:
https://github.com/apache/cloudstack/pull/1931
https://github.com/apache/cloudstack/pull/1932

Issue: https://issues.apache.org/jira/browse/CLOUDSTACK-9473

Storage pool checker is not being called on resize and migrate volume.
This may lead to allocated percentage of storage above 100%.

Setup:
1 VMware cluster with 2 Hosts.

Executed Steps:

Applied the following global settings:
storage.overprovisioning.factor = 1
pool.storage.allocated.capacity.disablethreshold = 1
pool.storage.capacity.disablethreshold = 1
Restarted management server
Executed Resize and migrate pool and Observed that Storage pool checker is not performed on resizeVolume and migrateVolume.
Result:
Root cause analysis shows storage pool checker is not called when doing migration and resizing.

Signed-off-by: Rohit Yadav <rohit.yadav@shapeblue.com>

## Types of changes

## Checklist:

Testing

### DIFF
--- a/server/src/com/cloud/storage/StorageManagerImpl.java
+++ b/server/src/com/cloud/storage/StorageManagerImpl.java
@@ -1788,8 +1788,8 @@ public class StorageManagerImpl extends ManagerBase implements StorageManager, C
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Destination pool id: " + pool.getId());
         }
-
-        StoragePoolVO poolVO = _storagePoolDao.findById(pool.getId());
+        // allocated space includes templates
+        final StoragePoolVO poolVO = _storagePoolDao.findById(pool.getId());
         long allocatedSizeWithTemplate = _capacityMgr.getAllocatedPoolCapacity(poolVO, null);
         long totalAskingSize = 0;
 
@@ -1830,12 +1830,37 @@ public class StorageManagerImpl extends ManagerBase implements StorageManager, C
             }
         }
 
+        return checkPoolforSpace(pool, allocatedSizeWithTemplate, totalAskingSize);
+    }
+
+    @Override
+    public boolean storagePoolHasEnoughSpaceForResize(StoragePool pool, long currentSize, long newSiz) {
+        if (!checkUsagedSpace(pool)) {
+            return false;
+        }
+        if (s_logger.isDebugEnabled()) {
+            s_logger.debug("Destination pool id: " + pool.getId());
+        }
+        long totalAskingSize = newSiz - currentSize;
+
+        if (totalAskingSize <= 0) {
+            return true;
+        } else {
+            final StoragePoolVO poolVO = _storagePoolDao.findById(pool.getId());
+            final long allocatedSizeWithTemplate = _capacityMgr.getAllocatedPoolCapacity(poolVO, null);
+            return checkPoolforSpace(pool, allocatedSizeWithTemplate, totalAskingSize);
+        }
+    }
+
+    private boolean checkPoolforSpace(StoragePool pool, long allocatedSizeWithTemplate, long totalAskingSize) {
+        // allocated space includes templates
+        StoragePoolVO poolVO = _storagePoolDao.findById(pool.getId());
+
         long totalOverProvCapacity;
 
         if (pool.getPoolType().supportsOverProvisioning()) {
             BigDecimal overProvFactor = getStorageOverProvisioningFactor(pool.getId());
             totalOverProvCapacity = overProvFactor.multiply(new BigDecimal(pool.getCapacityBytes())).longValue();
-
             s_logger.debug("Found storage pool " + poolVO.getName() + " of type " + pool.getPoolType().toString() + " with overprovisioning factor " + overProvFactor.toString());
             s_logger.debug("Total over provisioned capacity calculated is " + overProvFactor + " * " + pool.getCapacityBytes());
         } else {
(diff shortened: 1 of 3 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

server/src/com/cloud/storage/StorageManagerImpl.java
```diff
@@ -1830,12 +1830,37 @@ public boolean storagePoolHasEnoughSpace(List<Volume> volumes, StoragePool pool,
...
 
+        return checkPoolforSpace(pool, allocatedSizeWithTemplate, totalAskingSize);
+    }
+
+    @Override
+    public boolean storagePoolHasEnoughSpaceForResize(StoragePool pool, long currentSize, long newSiz) {
+        if (!checkUsagedSpace(pool)) {
+            return false;
+        }
+        if (s_logger.isDebugEnabled()) {
```
wido: Is this if really needed? Can't you just log to debug and let the filter of log4j handle it?

rhtyd: If log4j does not have debug enabled, then the string concat operation will still be performed by the JVM only when the string is passed to the `debug()` method it will be ignored. So, it depends if we can afford the CPU/time spent on the operation.

rafaelweingartner: I think we can afford this tiny bit of extra CPU.

server/src/com/cloud/storage/StorageManagerImpl.java
```diff
@@ -1830,12 +1830,37 @@ public boolean storagePoolHasEnoughSpace(List<Volume> volumes, StoragePool pool,
...
+            s_logger.debug("Destination pool id: " + pool.getId());
+        }
+        long totalAskingSize = newSiz - currentSize;
+
+        if (totalAskingSize <= 0) {
```
rafaelweingartner: Is this possible?

server/src/com/cloud/storage/StorageManagerImpl.java
```diff
@@ -1847,21 +1872,26 @@ public boolean storagePoolHasEnoughSpace(List<Volume> volumes, StoragePool pool,
         s_logger.debug("Total capacity of the pool " + poolVO.getName() + " with ID " + pool.getId() + " is " + totalOverProvCapacity);
 
         double storageAllocatedThreshold = CapacityManager.StorageAllocatedCapacityDisableThreshold.valueIn(pool.getDataCenterId());
-
-        s_logger.debug("Checking pool: " + pool.getId() + " for volume allocation " + volumes.toString() + ", maxSize : " + totalOverProvCapacity + ", totalAllocatedSize : "
-                + allocatedSizeWithTemplate + ", askingSize : " + totalAskingSize + ", allocated disable threshold: " + storageAllocatedThreshold);
+        if (s_logger.isDebugEnabled()) {
```
rafaelweingartner: Why checking before logging?

We are removing this from other places. Let's remove conditionals from our code base.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1137,6 +1137,14 @@ private VolumeVO orchestrateResizeVolume(long volumeId, long currentSize, long n
         UserVmVO userVm = _userVmDao.findById(volume.getInstanceId());
         StoragePoolVO storagePool = _storagePoolDao.findById(volume.getPoolId());
         boolean isManaged = storagePool.isManaged();
+
+        List<Volume> volumes = new ArrayList<Volume>();
```
rafaelweingartner: Are you using this variable?

server/src/com/cloud/storage/StorageManagerImpl.java
DaanHoogland: I think as a matter of principle we should alwas add the check. You never no for what reason someone will alter the message. I like using String.format to construct the message, btw, and then pass the single resulting string to the logger

server/src/com/cloud/storage/StorageManagerImpl.java
rafaelweingartner: Got it. Thanks

server/src/com/cloud/storage/StorageManagerImpl.java
rafaelweingartner: What principle?

I also use String.format, and I do not see the problem.

DaanHoogland: "bla" + "blob" is a costly operation, basically to memcpy()s. String.format() as well in comparison. We don't want those operations performed if we don't end up using the result in a log statement. so the boolean check potentially saves us a lot of cpu cycles, especially in the debug() and trace() cases.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
DaanHoogland: don't think so and also these methods here are to huge for my puny brain

server/src/com/cloud/storage/StorageManagerImpl.java
rafaelweingartner: Yes, I do know that. However, we should ask ourselves, if the addition of a if conditions (it will clutter the code) is worth the CPU cycles. For me, I would not trade readability for this small CPU cost in a web system (ACS is a at the end of the day a Java web application).

server/src/com/cloud/storage/StorageManagerImpl.java
rafaelweingartner: I still see no real benefits of adding this log level checks in our context. On the other hand, I do see problems with readability of the code. 

I cannot object though, if you want to proceed with this ;)

server/src/com/cloud/storage/VolumeApiServiceImpl.java
rafaelweingartner: My brain only works with up to 10 lines of code methods. Anything else is just a blur.

I think the code is ok; we just need to check if this variable is used; if it is not, please remove it.

rhtyd: Removed, unused code.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Test lgtm

Issue #2838, opened 2018-09-07
BLOCKER: 4.11.1: VMware: VM start fails with ArrayIndexOutOfBoundsException


resmo:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VM
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1
~~~

##### OS / ENVIRONMENT

VMware 5.5

##### SUMMARY

Upgraded ACS from 4.5.x to 4.11.1. Some existing VMs which we could stop/start under 4.5 can no longer be started with cloudstack after stopping. After rollback to 4.5, the VM can be started again with cloudstack. The VM is quite "old", created 01 Apr 2015 09:09:40. The VM _can_ be started on vCenter also also stopped by cloudstack, but not started with cloudstack.

Another newer VM (2018), but created also before the upgrade 4.11.1, can be stopped and started with cloudstack. So it is not a general problem.

~We didn't find the reason why this particular VM can not be started.~

##### STEPS TO REPRODUCE

~~~
* set configuration vmware.create.full.clone = true
* create VM
* set configuration vmware.create.full.clone = false
* stop/start the same VM 
~~~

##### EXPECTED RESULTS

~~~
VM is started
~~~

##### ACTUAL RESULTS

~~~
2018-09-07 14:18:35,880 DEBUG [cloud.vm.VmWorkJobHandlerProxy] (Work-Job-Executor-73:ctx-f0153f83 job-31342/job-31343 ctx-749d330b) Execute VM work job: com.cloud.vm.VmWorkStart{"dcId":1,"podId":1,"clusterId":1,"hostId":1,"userId":2,"accountId":2,"vmId":172,"handlerName":"VirtualMachineManagerImpl"}
2018-09-07 14:18:35,899 DEBUG [cloud.capacity.CapacityManagerImpl] (Work-Job-Executor-73:ctx-f0153f83 job-31342/job-31343 ctx-749d330b) VM state transitted from :Stopped to Starting with event: StartRequestedvm's original host id: 1 new host id: null host id before state transition: null
2018-09-07 14:18:35,900 DEBUG [cloud.vm.VirtualMachineManagerImpl] (Work-Job-Executor-73:ctx-f0153f83 job-31342/job-31343 ctx-749d330b) Successfully transitioned to start state for VM[User|i-28-172-VM] reservation id = f0d1fbbe-6e71-43f5-bd2b-9559ca06e1cf
[...]

PR #2840, opened 2018-09-11
Fix for Vmware full clones update

nvazquez:
## Description
The use of full/linked clones on Vmware caused some VMs not starting due to ArrayIndexOutOfBoundsException

## Types of changes

## GitHub Issue/PRs
#2838 BLOCKER: 4.11.1: VMware: VM start fails with ArrayIndexOutOfBoundsException

## How Has This Been Tested?
On Vmware environment:
- Set configuration vmware.create.full.clone = true
- Create VM
- Set configuration vmware.create.full.clone = false
- Stop/start the same VM 

## Checklist:
Testing

### DIFF
--- a/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
@@ -1399,7 +1399,7 @@ public class VolumeOrchestrator extends ManagerBase implements VolumeOrchestrati
                 if (cloneSettingVO != null) {
                     if (!cloneSettingVO.getCloneType().equals(cloneType.toString())) {
                         cloneSettingVO.setCloneType(cloneType.toString());
-                        _vmCloneSettingDao.update(cloneSettingVO.getVmId(), cloneSettingVO);
+                        _vmCloneSettingDao.update(cloneSettingVO.getId(), cloneSettingVO);
                     }
                 } else {
                     UserVmCloneSettingVO vmCloneSettingVO = new UserVmCloneSettingVO(vm.getId(), cloneType.toString());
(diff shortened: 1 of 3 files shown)


resmo:
I wonder why does cloudstack make any assumption based on a database record and not of the vmware vsphere API if the volume is a linked or full clone. 

Let's assume we have a `vmware.create.full.clone = false` and create a VM. After that we storage migrate the VM. The linked clone becomes a full clone, however, cloudstack still handles it as linked clone because the database says so. 

Does this make any sense?

nvazquez:
@rhtyd thanks, let me try that approach which will be simpler

Thanks @resmo, I partially agree with you. In my point of view it is not bad to make the assumption based on the DB record as long as that record is updated (which is not the case as you mentioned). I would suggest updating DB entries after migrations with the information provided by the vSphere API. What do you think?

resmo:
I must confess, I may not get every use case why this is implemented

But AFAICS the record gets updated by source of whatever cloudstack assumes it is based on, not what it actually is. The thing is that you can perfectly fine storage migrate using vsphere only without cloudstack and cloudstack does not get informed by this action. 

IMHO in this case cloudstack manages something of none if its business. Every code not written is bug free code.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

engine/schema/resources/META-INF/db/schema-41100to41110.sql
```diff
@@ -91,3 +91,9 @@ INSERT IGNORE INTO `cloud`.`guest_os_hypervisor` (uuid,hypervisor_type, hypervis
 -- XCP-NG 7.4
 INSERT IGNORE INTO `cloud`.`hypervisor_capabilities`(uuid, hypervisor_type, hypervisor_version, max_guests_limit, max_data_volumes_limit, storage_motion_supported) values (UUID(), 'XenServer', 'XCP-ng 7.4.0', 500, 13, 1);
 INSERT IGNORE INTO `cloud`.`guest_os_hypervisor` (uuid,hypervisor_type, hypervisor_version, guest_os_name, guest_os_id, created, is_user_defined) SELECT UUID(),'Xenserver', 'XCP-ng 7.4.0', guest_os_name, guest_os_id, utc_timestamp(), 0  FROM `cloud`.`guest_os_hypervisor` WHERE hypervisor_type='Xenserver' AND hypervisor_version='7.4.0';
+
+-- Fix Vmware full clones issue
+ALTER TABLE `cloud`.`user_vm_clone_setting`
```
rhtyd: @nvazquez can you move the sql changes from this file to the 4.11.1.0->4.11.2.0 sql path? Also rebase this against latest 4.11 branch?

nvazquez:
@resmo I think the aim of the feature was to control the use of linked clones or full clones at storage pool level at VM deployment time. If the configuration value is not set for a storage pool then the global value should be used. In case of stop/start a VM then only DB is updated to keep consistency between the storage pool configuration value for existing VMs in that storage pool.

engine/schema/resources/META-INF/db/schema-41100to41110.sql
nvazquez: Sure, done

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

PR #2816, opened 2018-08-20
CLOUDSTACK-10365: Change the "getXXX" boolean-related method names to…

BruceKuiLiu:
… "isXXX".

These boolean-return methods are named as "getXXX".
Other boolean-return methods are named as "isXXX".
Considering there methods will return boolean values, it should be more clear and consistent to rename them as "isXXX".
(rebase #2602)

PR #2847, opened 2018-09-17
CLOUDSTACK-10365: Change the "getXXX" boolean-related method names to…

BruceKuiLiu:
… "isXXX".

These boolean-return methods are named as "getXXX".
Other boolean-return methods are named as "isXXX".
Considering there methods will return boolean values, it should be more clear and consistent to rename them as "isXXX".
(rebase #2602 and #2816)

### DIFF
--- a/engine/schema/src/main/java/com/cloud/offerings/NetworkOfferingVO.java
+++ b/engine/schema/src/main/java/com/cloud/offerings/NetworkOfferingVO.java
@@ -172,7 +172,7 @@ public class NetworkOfferingVO implements NetworkOffering {
     }
 
     @Override
-    public boolean getForVpc() {
+    public boolean isForVpc() {
         return forVpc;
     }
 
@@ -249,7 +249,7 @@ public class NetworkOfferingVO implements NetworkOffering {
     }
 
     @Override
-    public boolean getSpecifyVlan() {
+    public boolean isSpecifyVlan() {
         return specifyVlan;
     }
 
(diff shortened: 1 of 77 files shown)


borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

DaanHoogland:
@BruceKuiLiu I agree with @rafaelweingartner as I noted in your prior PR, please have some patience with us, we are not the fasted mergers in the world of github.

engine/schema/src/main/java/com/cloud/offerings/NetworkOfferingVO.java
```diff
@@ -485,17 +485,17 @@ public void setIsPersistent(Boolean isPersistent) {
     }
 
     @Override
-    public boolean getIsPersistent() {
+    public boolean isPersistent() {
         return isPersistent;
```
rhtyd: This cause the GC bug, our DB framework assumes that the attributes (map) is created by a method interceptor so the name of the variable should not have the 'is', the name of the variable that related to a db table's column should be something after removing get/is string prefix: https://github.com/apache/cloudstack/blob/master/framework/db/src/main/java/com/cloud/utils/db/SearchBase.java#L480
/cc @nvazquez @anuragaw @sh0wstopper @dhlaluku https://github.com/apache/cloudstack/issues/2935

PR #2869, opened 2018-10-01
Fix some Marvin smoke tests

PaulAngus:
## Description

Fixes for test_deploy_virtio_scsi_vm.py which throws an exception if the hypervisor does not support the tests.
And test_vm_life_cycle.py and test_vm_life_cycle.py which fail when the environment hasn't fully recovered from the previous test.

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Tested by patching tests and then running them against a virtualized environment

## Checklist:

Testing
- [ x ] All relevant new and existing integration tests have passed.

### DIFF
--- a/test/integration/smoke/test_deploy_virtio_scsi_vm.py
+++ b/test/integration/smoke/test_deploy_virtio_scsi_vm.py
@@ -96,6 +96,7 @@ class TestDeployVirtioSCSIVM(cloudstackTestCase):
         cls.zone = get_zone(cls.apiclient, testClient.getZoneForTests())
         cls.pod = get_pod(cls.apiclient, cls.zone.id)
         cls.services['mode'] = cls.zone.networktype
+        cls.cleanup = []
         if cls.hypervisor.lower() not in ['kvm']:
             cls.hypervisorNotSupported = True
             return
--- a/test/integration/smoke/test_hostha_kvm.py
+++ b/test/integration/smoke/test_hostha_kvm.py
@@ -274,9 +274,12 @@ class TestHAKVM(cloudstackTestCase):
             Tests Enable HA when host is in Maintenance mode, should be Ineligible
         """
         self.logger.debug("Starting test_hostha_enable_ha_when_host_in_maintenance")
-
+        self.logger.debug("Pausing to wait for VMs to have finished starting")
+        time.sleep(300)
+        
         # Enable HA
         self.configureAndEnableHostHa()
+        
 
         # Prepare for maintenance Host
         self.setHostToMaintanance(self.host.id)
--- a/test/integration/smoke/test_vm_life_cycle.py
+++ b/test/integration/smoke/test_vm_life_cycle.py
@@ -900,6 +900,23 @@ class TestSecuredVmMigration(cloudstackTestCase):
         vm_response = VirtualMachine.list(self.apiclient, id=vm.id)[0]
         self.assertEqual(vm_response.hostid, dest_host.id, "Check destination host ID of migrated VM")
 
+    def waitUntilHostInState(self, hostId, state="Up", interval=5, retries=20):
+        while retries > -1:
+            print("Waiting for host: %s to be %s. %s retries left." % (hostId, state, retries))
+            time.sleep(interval)
+            host = Host.list(
+                self.apiclient,
+                hostid=hostId,
+                type='Routing'
+            )[0]
+            if host.state != state:
+                if retries >= 0:
+                    retries = retries - 1
+                    continue
+            else:
+                print("Host %s now showing as %s" % (hostId, state))
+                return
+
     def unsecure_host(self, host):
         SshClient(host.ipaddress, port=22, user=self.hostConfig["username"], passwd=self.hostConfig["password"])\
             .execute("rm -f /etc/cloudstack/agent/cloud* && \
@@ -909,8 +926,8 @@ class TestSecuredVmMigration(cloudstackTestCase):
                       service libvirtd restart && \
                       sleep 30 && \
                       service cloudstack-agent restart")
- 
-        time.sleep(30)
+        print("Unsecuring Host: %s" % (host.name))
+        self.waitUntilHostInState(hostId=host.id, state="Up") 
         self.check_connection(host=host, secured='false')
         return host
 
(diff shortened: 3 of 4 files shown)


PaulAngus:
@blueorangutan package

test/integration/smoke/test_hostha_kvm.py
```diff
@@ -274,9 +274,12 @@ def test_hostha_enable_ha_when_host_in_maintenance(self):
             Tests Enable HA when host is in Maintenance mode, should be Ineligible
         """
         self.logger.debug("Starting test_hostha_enable_ha_when_host_in_maintenance")
-
+        self.logger.debug("Pausing to wait for VMs to have finished starting")
+        time.sleep(300)
```
rafaelweingartner: Can't we execute this "sleep" differently? I mean, wouldn't it be better to check the "startVm" job to see if it has finished properly?

PaulAngus: ideally yes.  BUT a. that's beyond my ability (this is the 2nd bit of python i've ever written) and b. Across the test infrastructure, there may be many VMs/VRs and there may be many hosts.  In our Trillian environment there are only 2 hosts by default, so there is a fifty/fifty chance that there was a VR on the host that was fenced and then recovered.
In the end, even trying to cover all of the possibilities seemed excessive for the use case.

test/integration/smoke/test_deploy_virtio_scsi_vm.py
```diff
@@ -96,6 +96,7 @@ def setUpClass(cls):
         cls.zone = get_zone(cls.apiclient, testClient.getZoneForTests())
         cls.pod = get_pod(cls.apiclient, cls.zone.id)
         cls.services['mode'] = cls.zone.networktype
+        cls.cleanup = []
```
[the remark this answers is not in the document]
PaulAngus: cls.cleanup is defined AFTER the if statement which checks if the hypervisor is supported.  
if the hypervisor ISN't supported, the teardown still occurs but 
```cleanup_resources(cls.apiclient, cls.cleanup)```
throws an exception because cls.cleanup hadn't been defined. @borisstoyanov

File: test/integration/smoke/test_hostha_kvm.py (hunk 1 above)
Comment (borisstoyanov) [reply test/integration/smoke/test_hostha_kvm.py]: I though CloudStack moves all VMs/VRs out of a host when preparing for maintenance? Let's work this out in other more elegant way, not a fan of setting things on hold for 5mins just like that..

File: test/integration/smoke/test_deploy_virtio_scsi_vm.py (hunk 2 above)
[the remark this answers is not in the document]
Comment (borisstoyanov) [reply test/integration/smoke/test_deploy_virtio_scsi_vm.py]: had another look at the code, I think the problem is in that 'if' hypervisor check, see at the end of it there's a return, which will end the execution of this method and that's why we don't get the cleanup defined. See the issue is that instead of return we should have skip(), which will 'cancel' the test execution as it is. let me help with that

File: test/integration/smoke/test_hostha_kvm.py (hunk 1 above)
Comment (PaulAngus) [reply test/integration/smoke/test_hostha_kvm.py]: I'm happy for you to work that out, I'm just trying to get a set of tests that aren't constantly throwing false positives, which make them useless.

**Comment by borisstoyanov** [comment]
No, you can't bypass the tearDown if you do a skip somewhere in code, it'll still enter in it. But you're right you're registering resources which you'll end up destroying and not using... let me restructure..

File: test/integration/smoke/test_vm_life_cycle.py [hunk 3]
```diff
@@ -839,6 +839,7 @@ def setUpClass(cls):
 
     @classmethod
     def tearDownClass(cls):
+        time.sleep(120)
```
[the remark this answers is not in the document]
PaulAngus: I'm happy for you to work that out, I'm just trying to get a set of tests that aren't constantly throwing false positives, which make them useless. I don't have time to figure out all cases.

borisstoyanov:
I've just reverted the changes, it ain't stupid if it works :)

test/integration/smoke/test_hostha_kvm.py
rafaelweingartner: @borisstoyanov indeed CloudStack does that. There is something we could use, the host status. It will only go to maintenance mode when everything is migrated out of the host. Also, if there is a problem and something is not migrated, the host will go to `ErrorInMaitenance` mode.

borisstoyanov:
@blueorangutan test

rhtyd:
Let's also change base branch to 4.11?

PaulAngus:
how do i change the base branch @rhtyd ?  close and reopen against 4.11. ?

rhtyd:
@Paul in github you'll need to edit the base branch to 4.11, you'll also need to rebase your branch against 4.11 branch and force push.

rhtyd:
@PaulAngus fixed and rebased the PR branch against 4.11, will help with testing/reviewing tomorrow.

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

tools/marvin/marvin/lib/base.py
```diff
@@ -605,22 +605,34 @@ def create(cls, apiclient, services, templateid=None, accountid=None,
...
-                # EIP/ELB (netscaler) enabled zone
-                vm_ssh_ip = virtual_machine.publicip
-            else:
-                # regular basic zone with security group
-                vm_ssh_ip = virtual_machine.nic[0].ipaddress
-            virtual_machine.ssh_ip = vm_ssh_ip
-            virtual_machine.public_ip = vm_ssh_ip
+        retries = 5
+        interval = 30
+        while retries > 0:
```
rhtyd: Rewrite this using `wait_until`.

test/integration/smoke/test_vm_life_cycle.py
```diff
@@ -907,9 +907,10 @@ def unsecure_host(self, host):
                       sed -i 's/listen_tcp.*/listen_tcp=1/g' /etc/libvirt/libvirtd.conf && \
                       sed -i '/.*_file=.*/d' /etc/libvirt/libvirtd.conf && \
                       service libvirtd restart && \
+                      sleep 30 && \
                       service cloudstack-agent restart")
-
-        time.sleep(10)
+ 
+        time.sleep(30)
```
rhtyd: Why do we need to sleep? Could be do a wait_until check here instead?

PaulAngus:
@blueorangutan package

rhtyd:
are we good to merge this? @PaulAngus @borisstoyanov @rafaelweingartner ?

borisstoyanov:
yes, I believe so @rhtyd

PaulAngus:
we've been running with them under trillian - and they have been reliable

PR #2879, opened 2018-10-05
ca: Fixes #2877 mgmt server cert should have all addrs of default nic

### LINKED ISSUE #2877: RootCAProvider certificate validation fails if default interface has multiple IP addresses
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
CA
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1.0
~~~

##### CONFIGURATION

Two management servers using a VIP and UCARP for failover.

##### OS / ENVIRONMENT

N/A

##### SUMMARY

Secure connections between two management servers using the default root CA provider don't pass validation if the default interface has multiple IP addresses.

For example, using a Virtual IP with UCARP for management server failover, the default interface on the primary management server will have two IP addresses assigned. The Cloudstack RootCAProvider generates certificates by using one IP address from the default interface and including it in the certificate as a SAN (from my experience, the Virtual IP is put in the certificate). When the primary management server attempts connection to the secondary management server, the connection is rejected because the secondary server determines the connection to be from the 'real' IP address of the primary server and cannot validate the certificate because only the Virtual IP is listed as a SAN.

##### STEPS TO REPRODUCE

~~~
Example:
Let's say we have the following:
primary server's IP is 192.168.1.10 and Virtual IP is 192.168.1.20

The certificate is generated with 192.168.1.20 as a SAN.
The secondary server attempts to validate a connection from the primary. The connection is from 192.168.1.10, 
so it can't validate the certificate presented by the primary controller.

In the logs you can find the following:
Certificate ownership verification failed for client: 192.168.1.10

A client/agent attempting connection from address=192.168.1.10 has presented these certificate(s):
Certificate [1] :
Serial: xxx
Not Before:Wed Oct 03 04:29:46 UTC 2018
Not After:Fri Sep 25 16:29:46 UTC 2048
Signature Algorithm:SHA256withRSA
Version:3
Subject DN:CN=cloudstack1.example.com
Issuer DN:CN=ca.cloudstack.apache.org
Alternative Names:[[7, 192.168.1.20], [2, cloudstack1.example.com]]
Certificate [2] :
Serial: xxx
Not Before:Mon Aug 20 05:27:08 UTC 2018
Not After:Wed Aug 12 17:27:08 UTC 2048
Signature Algorithm:SHA256withRSA
Version:3
Subject DN:CN=ca.cloudstack.apache.org
Issuer DN:CN=ca.cloudstack.apache.org
Alternative Names:null
~~~

##### EXPECTED RESULTS

~~~
I would expect all IP addresses for the default interface to be included as Subject Alternative Names in the certificate.
~~~

##### ACTUAL RESULTS

~~~
Only the Virtual IP is included in the certificate as a Subject Alternative Name on the primary controller and connection to the secondary controller fails validation.
~~~

rhtyd:
This fixes the default RootCA provider implementation to initiate
and issue certificate for mgmt server on startup for all the IP addresses
on the default nic of that host.

Fixes #2877 

## Types of changes

### DIFF
--- a/plugins/ca/root-ca/src/org/apache/cloudstack/ca/provider/RootCAProvider.java
+++ b/plugins/ca/root-ca/src/org/apache/cloudstack/ca/provider/RootCAProvider.java
@@ -359,7 +359,7 @@ public final class RootCAProvider extends AdapterBase implements CAProvider, Con
             return true;
         }
         final Certificate serverCertificate = issueCertificate(Collections.singletonList(NetUtils.getHostName()),
-                Collections.singletonList(NetUtils.getDefaultHostIp()), getCaValidityDays());
+                NetUtils.getAllDefaultNicIps(), getCaValidityDays());
         if (serverCertificate == null || serverCertificate.getPrivateKey() == null) {
             throw new CloudRuntimeException("Failed to generate management server certificate and load management server keystore");
         }
--- a/utils/src/main/java/com/cloud/utils/net/NetUtils.java
+++ b/utils/src/main/java/com/cloud/utils/net/NetUtils.java
@@ -225,6 +225,27 @@ public class NetUtils {
         }
     }
 
+    public static List<String> getAllDefaultNicIps() {
+        final List<String> addrs = new ArrayList<>();
+        final String pubNic = getDefaultEthDevice();
+
+        if (pubNic == null) {
+            return addrs;
+        }
+
+        NetworkInterface nic = null;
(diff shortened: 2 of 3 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

Issue #2873, opened 2018-10-03
VR seen to have memory leaks


rhtyd:
In one of the 4.11.x environments, it is seen that VR memory consumption increases the maximum leading to swapping. 

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1.0/4.11.2.0-rc2
~~~

##### CONFIGURATION

Advanced zone + adv networking

##### OS / ENVIRONMENT

Seen in case of VMware 5.5.x

##### STEPS TO REPRODUCE

~~~
Deploy VPC VR
Wait for 3 days
See the VR has consumed all the RAM, swapping could be seen
~~~

##### EXPECTED RESULTS

~~~
VR should run without causing memory leaks
~~~

##### ACTUAL RESULTS

~~~
VR found thrashing and swapping
~~~

PR #2892, opened 2018-10-10
vr: memory and swap optimizations

rhtyd:
This tries to provide a threshold based fix for #2873 where swappinness of VR is not used until last resort. By limiting swappiness unless actually needed, the VR system degradation can be avoided for most cases. The other change is around not starting `baremetal-vr` by default on all VRs, according to the spec https://cwiki.apache.org/confluence/display/CLOUDSTACK/Baremetal+Advanced+Networking+Support only vmware VRs need to run it and that too only as the last step of the setup/completion, so we don't need to run it all the time.

## Types of changes

## How Has This Been Tested?

The default swapiness on VRs is 60, by reducing the value swapiness it is anticipated that swappiness will not be normally seen. I could not reproduce the swapping so could not test it.

### DIFF
--- a/systemvm/debian/etc/sysctl.conf
+++ b/systemvm/debian/etc/sysctl.conf
@@ -62,4 +62,4 @@ net.ipv6.conf.all.accept_redirects = 0
 net.ipv6.conf.all.autoconf = 0
 
 # Minimum swappiness without disabling it
-vm.swappiness=5
+vm.swappiness=1
--- a/systemvm/debian/etc/sysctl.conf
+++ b/systemvm/debian/etc/sysctl.conf
@@ -60,3 +60,6 @@ net.ipv6.conf.all.forwarding = 0
 net.ipv6.conf.all.accept_ra = 0
 net.ipv6.conf.all.accept_redirects = 0
 net.ipv6.conf.all.autoconf = 0
+
+# Minimum swappiness without disabling it
+vm.swappiness=5
--- a/plugins/hypervisors/baremetal/src/com/cloud/baremetal/networkservice/BaremetalKickStartServiceImpl.java
+++ b/plugins/hypervisors/baremetal/src/com/cloud/baremetal/networkservice/BaremetalKickStartServiceImpl.java
@@ -226,11 +226,16 @@ public class BaremetalKickStartServiceImpl extends BareMetalPxeServiceBase imple
             throw new CloudRuntimeException(String.format("please specify 'baremetal.internal.storage.server.ip', which is the http server/nfs server storing kickstart files and ISO files, in global setting"));
         }
 
+        Pair<Boolean, String> ret = SshHelper.sshExecute(mgmtNic.getIPv4Address(), 3922, "root", getSystemVMKeyFile(), null, "systemctl start baremetal-vr");
+        if (!ret.first()) {
+            throw new CloudRuntimeException(String.format("failed to start baremetal agent in virtual router[id:%s]", vr.getId()));
+        }
+
         List<String> tuple =  parseKickstartUrl(profile);
         String cmd =  String.format("/opt/cloud/bin/prepare_pxe.sh %s %s %s %s %s %s", tuple.get(1), tuple.get(2), profile.getTemplate().getUuid(),
                 String.format("01-%s", nic.getMacAddress().replaceAll(":", "-")).toLowerCase(), tuple.get(0), nic.getMacAddress().toLowerCase());
         s_logger.debug(String.format("prepare pxe on virtual router[ip:%s], cmd: %s", mgmtNic.getIPv4Address(), cmd));
-        Pair<Boolean, String> ret = SshHelper.sshExecute(mgmtNic.getIPv4Address(), 3922, "root", getSystemVMKeyFile(), null, cmd);
+        ret = SshHelper.sshExecute(mgmtNic.getIPv4Address(), 3922, "root", getSystemVMKeyFile(), null, cmd);
         if (!ret.first()) {
             throw new CloudRuntimeException(String.format("failed preparing PXE in virtual router[id:%s], because %s", vr.getId(), ret.second()));
         }
--- a/systemvm/debian/etc/systemd/journald.conf
+++ b/systemvm/debian/etc/systemd/journald.conf
@@ -1,5 +1,5 @@
 [Journal]
 Compress=yes
-SystemMaxUse=50M
+SystemMaxUse=40M
 SystemMaxFileSize=10M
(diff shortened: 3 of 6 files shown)


rhtyd:
There is another workaround we can employ to have a cronjob that flushes the buffers/cache to reclaim memory every few 10s of minutes like: `sync && echo 3 > /proc/sys/vm/drop_caches`.
@blueorangutan package

systemvm/debian/etc/sysctl.conf
```diff
@@ -60,3 +60,6 @@ net.ipv6.conf.all.forwarding = 0
 net.ipv6.conf.all.accept_ra = 0
 net.ipv6.conf.all.accept_redirects = 0
 net.ipv6.conf.all.autoconf = 0
+
+# Minimum swappiness without disabling it
+vm.swappiness=5
```
wido: Fine with me, 5% works, 1% works or even 0%.

If you would set it to 0% it would only swap in the worst-case scenarios, but this is fine as well.

rhtyd: @wido thanks, 0% on newer kernels may disable swapping. I'm okay to go as low as 1% which means swapping is only used as a last resort. Will change the value to 1.

rhtyd: Fixed to 1 now.

systemvm/debian/etc/sysctl.conf
```diff
@@ -60,3 +60,6 @@ net.ipv6.conf.all.forwarding = 0
 net.ipv6.conf.all.accept_ra = 0
 net.ipv6.conf.all.accept_redirects = 0
 net.ipv6.conf.all.autoconf = 0
+
+# Minimum swappiness without disabling it
+vm.swappiness=1
```
rafaelweingartner: Shouldn't this be 5? I mean, to do what you say in the title of the PR.

rhtyd:
@blueorangutan package

resmo:
LGTM

I would also consider to change vm.panic_on_oom to reboot the VR on oom regardless.

~~~diff
- vm.panic_on_oom = 1
+ vm.panic_on_oom = 2
~~~


~~~
panic_on_oom

This enables or disables panic on out-of-memory feature.

If this is set to 0, the kernel will kill some rogue process,
called oom_killer.  Usually, oom_killer can kill rogue processes and
system will survive.

If this is set to 1, the kernel panics when out-of-memory happens.
However, if a process limits using nodes by mempolicy/cpusets,
and those nodes become memory exhaustion status, one process
may be killed by oom-killer. No panic occurs in this case.
Because other nodes' memory may be free. This means system total status
may be not fatal yet.

If this is set to 2, the kernel panics compulsorily even on the
above-mentioned. Even oom happens under memory cgroup, the whole
system panics.

The default value is 0.
1 and 2 are for failover of clustering. Please select either
according to your policy of failover.
panic_on_oom=2+kdump gives you very strong tool to investigate
why oom happens. You can get snapshot.
~~~

wido:
@resmo But then probably also set:

<pre>kernel.panic = 60</pre>

In that case the kernel also reboots the system after 60 seconds in case of a kernel panic, otherwise the VM will just hang.

rhtyd:
@wido the `kernel.panic = 10` is already set to reboot VR faster on panic (10s).
@resmo we already have VR process monitoring service in CloudStack that can check if a process is running or not and send alerts (spec https://cwiki.apache.org/confluence/display/CLOUDSTACK/Monitoring+VR+services), so ideally if any of the core services die an alert will be issued. With the value set at 2, VRs will be at a higher risk to get frequent OOM if the memory is not sufficient, for example the default system offering for routers is bare minimum (1cpu and 256mb ram) actually too small to be used in production (I believer most people change/configure the router/system offering based on their needs) and may hit oom frequently. Thoughts?

rhtyd:
I checked https://cwiki.apache.org/confluence/display/CLOUDSTACK/Baremetal+Advanced+Networking+Support, we don't need to run baremetal-vr agent on non-vmware VRs, also we don't need to run it unless a baremetal host is added. I'll add a commit to make it not run by default, and only when baremetal pxe needs to run in VR. This will save some 20-30MB ram in the VR for all hypervisors.

rhtyd:
@wido @resmo please see the additional commit around baremetal-vr 
@blueorangutan package

systemvm/debian/etc/sysctl.conf
rafaelweingartner: Ok, now it makes sense.

rhtyd:
Update: made another change to limit systemd-journald usage. By default, the journald can consume upto 4GBs depending on size of partition etc. This should limit the amount of disk/buffer cache usage. I'm testing this now.

resmo:
> Update: made another change to limit systemd-journald usage. By default, the journald can consume upto 4GBs depending on size of partition etc. This should limit the amount of disk/buffer cache usage. I'm testing this now.

Well, the docs says:

> The first pair defaults to 10% and the second to 15% of the size of the respective file system, but each value is capped to 4G.

and this default is fine, and works as expected 

Compress is also enabled by default, journald have sane defaults. I don't see any need to change them.

rhtyd:
@resmo the max size and runtime size values have been changed. The default is 100M.

resmo:
btw we found a typo in post-init systemd service..
https://github.com/apache/cloudstack/blob/0e6a06cd5f305fc4989e388cdc7ab68cd0e341b9/systemvm/debian/etc/systemd/system/cloud-postinit.service#L5

~~~diff
- Requires=network.service
+ Requires=networking.service
~~~

see
~~~
root@r-1604-VM:~# systemctl status network.service
Unit network.service could not be found.
~~~

vs
~~~
root@r-1604-VM:~# systemctl status networking.service
● networking.service - Raise network interfaces
   Loaded: loaded (/lib/systemd/system/networking.service; enabled; vendor preset: enabled)
   Active: active (exited) since Mon 2018-10-08 15:53:36 UTC; 3 days ago
     Docs: man:interfaces(5)
 Main PID: 908 (code=exited, status=0/SUCCESS)
    Tasks: 0 (limit: 9830)
   CGroup: /system.slice/networking.service
~~~

rhtyd:
Thanks @resmo for the find, I'll fix that as well.

rhtyd:
Rebased against latest 4.11 and fixed review issues.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Based on several days of testing, and reviews and tests I'll merge this. I found that the VR no longer swaps until the last resort and the overall memory is saved by few 20-40MBs.

Issue #2842, opened 2018-09-12
listApis command returns inconsistent results


csquire:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API
~~~

##### CLOUDSTACK VERSION

~~~
Tested on 4.9.3.0 and 4.11.1.0
~~~

##### CONFIGURATION

API throttling is enabled with default throttling of 25 requests per second

##### SUMMARY

Users reported that they weren't getting all apis listed in cloudmonkey when running a `sync`. After some debugging, I found that the problem is that the `ApiDiscoveryService` is calling `ApiRateLimitServiceImpl.checkAccess()`, so the results of the listApis command are being truncated because Cloudstack believes the user has exceeded their API throttling rate.

##### STEPS TO REPRODUCE

I enabled throttling with a 25 request per second limit. I then created a test role with only `list*` permissions and assigned it to a test user. When this user calls `listApis`, they will typically receive anywhere from 15-18 results. Checking the logs, you see `The given user has reached his/her account api limit, please retry after 218 ms.`. 

I raised the limit to 200 requests per second, restarted the management server and tried again. This time I got 143 results and no log messages about the user being throttled.

~~~

~~~

##### EXPECTED RESULTS

~~~
All of the apis the user has permission to access are listed
~~~

##### ACTUAL RESULTS

~~~
The user only gets a small number of apis listed in their results
~~~

kioie:
Seems like this issue should be closed.

rhtyd:
Closing per https://github.com/apache/cloudstack/pull/2894

PR #2894, opened 2018-10-10
api: don't throttle api discovery for listApis command

csquire:
## Description

Users reported that they weren't getting all apis listed in cloudmonkey when running a `sync`. After some debugging, I found that the problem is that the `ApiDiscoveryService` is calling `ApiRateLimitServiceImpl.checkAccess()`, so the results of the listApis command are being truncated because Cloudstack believes the user has exceeded their API throttling rate.

I enabled throttling with a 25 request per second limit. I then created a test role with only `list*` permissions and assigned it to a test user. When this user calls `listApis`, they will typically receive anywhere from 15-18 results. Checking the logs, you see `The given user has reached his/her account api limit, please retry after 218 ms.`. 

I raised the limit to 200 requests per second, restarted the management server and tried again. This time I got 143 results and no log messages about the user being throttled.

## Types of changes

## GitHub Issue/PRs

Fixes: #2842 

## How Has This Been Tested?

Updated the appropriate unit tests to verify the ApiRateLimitService is not longer called when discovering APIs.

## Checklist:

Testing


plugins/api/discovery/src/org/apache/cloudstack/discovery/ApiDiscoveryServiceImpl.java
```diff
@@ -236,7 +238,9 @@ private ApiDiscoveryResponse getCmdRequestMap(Class<?> cmdClass, APICommand apiC
                 boolean isAllowed = true;
                 for (APIChecker apiChecker : _apiAccessCheckers) {
                     try {
-                        apiChecker.checkAccess(user, apiName);
+                        if (!"ApiRateLimitServiceImpl".equals(apiChecker.getName())) {
```
rhtyd: you can use something like this.class.getName() or the <class>.getName()... instead of the hard coded string comparison.

plugins/api/discovery/src/org/apache/cloudstack/discovery/ApiDiscoveryServiceImpl.java
```diff
@@ -223,7 +223,9 @@ private ApiDiscoveryResponse getCmdRequestMap(Class<?> cmdClass, APICommand apiC
 
             for (APIChecker apiChecker : _apiAccessCheckers) {
                 try {
-                    apiChecker.checkAccess(user, name);
+                    if (!"ApiRateLimitServiceImpl".equals(apiChecker.getName())) {
```
rhtyd: you can use something like this.class.getName() or the .getName()... instead of the hard coded string comparison.

plugins/api/discovery/src/org/apache/cloudstack/discovery/ApiDiscoveryServiceImpl.java
csquire: Using the 'ApiRateLimitServiceImpl' class instead of a hard coded string introduces the rate limit plugin modules as a dependency to the api discovery module.  I didn't think that was a good idea.

plugins/api/discovery/src/org/apache/cloudstack/discovery/ApiDiscoveryServiceImpl.java
csquire: Using the 'ApiRateLimitServiceImpl' class instead of a hard coded string introduces the rate limit plugin module as a dependency to the api discovery module. I didn't think that was a good idea.

rafaelweingartner: What if you used the command class instead? I mean, you are using the "service class", but at this point here you have the command that was used to list the APIs, right?  You can use it, and then you do not need to use this hardcoded String.

rafaelweingartner: Ok, I am now seeing the big picture. I looking at the code here. I will provide some feedback in another comment

rafaelweingartner:
@csquire, from your PR description I understood that we should not need/use the API rate limit service. Therefore, we only need to check if the user has access to the API method. Is that it?

If that is the case, why don’t we do something different?

At `org.apache.cloudstack.discovery.ApiDiscoveryServiceImpl.listApis(User, String)` method, instead of injecting all "APIChecker” implementations, we can only inject the `StaticRoleBasedAPIAccessChecker` and `DynamicRoleBasedAPIAccessChecker`. Then, you solve all of your problems with little or no coding at all.
To change the beans that are injected you can change the XML file `spring-server-core-misc-context.xml`. There is a bean called `apiDiscoveryServiceImpl`; then, you need to check the property being configured `apiAccessCheckers`. I guess you can change the property injection from:
`<property name="apiAccessCheckers" value="#{apiCheckersRegistry.registered}" />`

To something like:
```
        <property name="apiAccessCheckers">
            <util:list list-class="org.apache.cloudstack.acl.APIChecker">
                <ref bean="DynamicRoleBasedAPIAccessChecker" />
                <ref bean="StaticRoleBasedAPIAccessChecker" />
            </util:list>
        </property>

```
Of course, you need to import the “util” namespace in `spring-server-core-misc-context.xml`, but that is quite easy to do.

csquire:
yeah, great idea, that looks like a good solution.  I'll update the PR.

csquire:
I took a look at your suggestion @rafaelweingartner but those beans cannot be referenced directly in this context.

rafaelweingartner:
ow really?
So, what if you change the "set" method of that property to remove the "apiRateService" from the list of API checkers being configured?

csquire:
With the idea of not injecting the rate limit checker, I tried a different approach by creating a new marker interface and new registry for ACL checkers.  Thoughts?

rafaelweingartner:
I confess that I really dislike that registry solution that people implemented in ACS. Instead of using Spring life-cycle properly, people decided to create something else.

Your solution now works, and it is better than changing the code like it was done in the first attempt. However, I would still prefer something more integrated with Spring life-cycle.

Right now I am not able to think in something to enable us to work around this problem. Therefore, I am inclined to say that this last proposal is fine.

api/src/org/apache/cloudstack/acl/APIAclChecker.java
```diff
@@ -0,0 +1,7 @@
+package org.apache.cloudstack.acl;
```
rafaelweingartner: You missed the license header here

csquire: added!

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland:
@rhtyd is this ready?

rhtyd:
Let me test this tomorrow @DaanHoogland

Issue #2880, opened 2018-10-05
[4.11.2]: rolling upgrade of VR results in duplicate VRs


resmo:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### SUMMARY

~~~
2018-10-05 11:24:43,634 INFO  [jobs.impl.AsyncJobMonitor] (API-Job-Executor-1:ctx-2f1d201c job-32445) Add job-32445 into job monitoring
2018-10-05 11:24:43,637 DEBUG [jobs.impl.AsyncJobManagerImpl] (API-Job-Executor-1:ctx-2f1d201c job-32445) Executing AsyncJobVO {id:32445, userId: 2, accountId: 2, instanceType: None, instanceId: null, cmd: org.apache.cloudstack.api.command.user.network.RestartNetworkCmd, cmdInfo: {"makeredundant":"false","cleanup":"true","response":"json","ctxUserId":"2","httpmethod":"GET","ctxStartEventId":"1145358","id":"48df88b1-c2e2-432d-811a-681a2be74dde","ctxDetails":"{\"interface com.cloud.network.Network\":\"48df88b1-c2e2-432d-811a-681a2be74dde\"}","ctxAccountId":"2","uuid":"48df88b1-c2e2-432d-811a-681a2be74dde","cmdEventType":"NETWORK.RESTART","_":"1538731483586"}, cmdVersion: 0, status: IN_PROGRESS, processStatus: 0, resultCode: 0, result: null, initMsid: 345050582394, completeMsid: null, lastUpdated: null, lastPolled: null, created: null}
2018-10-05 11:24:43,664 DEBUG [engine.orchestration.NetworkOrchestrator] (API-Job-Executor-1:ctx-2f1d201c job-32445 ctx-dd2ddecb) Restarting network 207...
2018-10-05 11:24:43,665 DEBUG [engine.orchestration.NetworkOrchestrator] (API-Job-Executor-1:ctx-2f1d201c job-32445 ctx-dd2ddecb) Performing rolling restart of routers of network Ntwk[207|Guest|8]
2018-10-05 11:24:43,677 DEBUG [engine.orchestration.NetworkOrchestrator] (API-Job-Executor-1:ctx-2f1d201c job-32445 ctx-dd2ddecb) Asking VirtualRouter to implemenet Ntwk[207|Guest|8]
[...]

PR #2900, opened 2018-10-15
network: Allow ability to disable rolling restart feature

rhtyd:
This adds a global setting for admins who may want the lebacy behaviour of network restart with cleanup=true. In future, this setting may be removed.

The examples of error are in #2880, by having this toggle switch a temporary relief may be possible for such environments.

## Types of changes

## How Has This Been Tested?

The global settings ships as 'true', so rolling restart feature is shipped as the default behaviour. If somebody wants legacy behaviour they can set the value to false and try to restart an isolated network or redundant network with cleanup enabled.

### DIFF
--- a/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -2860,7 +2860,7 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
         s_logger.debug("Implementing the network " + network + " elements and resources as a part of network restart without cleanup");
         try {
             implementNetworkElementsAndResources(dest, context, network, offering);
-            setRestartRequired(network, true);
+            setRestartRequired(network, false);
             return true;
         } catch (final Exception ex) {
             s_logger.warn("Failed to implement network " + network + " elements and resources as a part of network restart due to ", ex);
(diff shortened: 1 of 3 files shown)


engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2860,7 +2860,7 @@ public boolean restartNetwork(final Long networkId, final Account callerAccount,
         s_logger.debug("Implementing the network " + network + " elements and resources as a part of network restart without cleanup");
         try {
             implementNetworkElementsAndResources(dest, context, network, offering);
-            setRestartRequired(network, true);
+            setRestartRequired(network, false);
```
rhtyd: This fixes an issue where restarting a network without cleanup always put the network's restart required flag to true.

rhtyd:
@borisstoyanov can you help test this one (manually maybe, both cases for restart with/without cleanup with/without redundant isolated network and VPCs) /cc @PaulAngus @DagSonsteboSB
@blueorangutan package

rhtyd:
@blueorangutan test

engine/orchestration/src/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
rafaelweingartner: this will the opposite? I mean, if I restart with cleanup, will not this cause the flag to always be false? Or, is the flag from the command captured and used somewhere else?

rhtyd: @rafaelweingartner the `restart required` flag is a presentational issue and used to communicate to users/admins that a network requires a restart because likely a previous restart attempt failed..

If you restart a network current without the cleanup=true flag, in the API response and UI the network's restart required state will change to true which is wrong. The state/flag should only be true if something went wrong in its last restart. You may read a recent thread here: https://markmail.org/message/bn57hymlvohf54b7

rhtyd:
Ping @resmo can you review and/or test?

rhtyd:
Ping @rafaelweingartner @nvasquez @PaulAngus can you review? Thanks.

rhtyd:
Merging this to not further block 4.11 rc3.

PR #2904, opened 2018-10-16
agent: Fixes #2899 on shutdown don't allow server reconnection

### LINKED ISSUE #2899: Initial agent connect dropped by management server
It is observed in 4.11.1/2-rc2 that in some environments when KVM agent connects to the management server, the channel is closed just after initial cmd. After that the ModifyStoragePool fails to be sent to the kvm host as the agent is disconnected putting the host in Alert state. The agent, however, will reconnect and finally, the state will be Up. In a rare setup, where certificates were already provisioned the agent failed to connect to the mgmt server at all on setup (adding host out of band) and the auth strictness was false.

##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME

~~~
Agent
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1
~~~

rhtyd:
When agent is stopped, don't allow reconnection. Previously this would
send a shutdown command to the management server which would put the
host state to Disconnected but then agent's reconnection logic may kick
in sometimes which would connect the agent to the management server
but then the agent process would terminate causing the host to be
put in Alert state (due to ping timeout or it waiting too long).

This fixes the issue by ensuring that when the agent is stopped, it
does not reconnect to the management server.

## Types of changes

## How Has This Been Tested?

With a KVM env, just stop the kvm agent the host previously was put in Alert state sometimes. Now, it is put in disconnected state.

### DIFF
--- a/agent/src/com/cloud/agent/Agent.java
+++ b/agent/src/com/cloud/agent/Agent.java
@@ -303,6 +303,7 @@ public class Agent implements HandlerFactory, IAgentControl {
 
     public void stop(final String reason, final String detail) {
         s_logger.info("Stopping the agent: Reason = " + reason + (detail != null ? ": Detail = " + detail : ""));
+        _reconnectAllowed = false;
         if (_connection != null) {
             final ShutdownCommand cmd = new ShutdownCommand(reason, detail);
             try {


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #2905, opened 2018-10-17
Fix delete domain button in the UI to update only after deleteDomain job finishes

rafaelweingartner:
## Description

When the “deleteDomain” command is sent via UI, the HTML is already updated assuming that the “deleteDomain” command will actually delete it. However, there are cases (domain with allocated resources) that will not allow ACS to remove the command; as a consequence, the UI will not be in sync with ACS state, which can lead to misunderstandings.

## Types of changes

## How Has This Been Tested?

Manually

### DIFF
--- a/ui/scripts/domains.js
+++ b/ui/scripts/domains.js
@@ -69,11 +69,23 @@
                                 async: false,
                                 success: function(json) {
                                     var jid = json.deletedomainresponse.jobid;
+                                    var wasDomainDeletedWithSuccess = false;
+                                    $.ajax({
+                                        url: createURL("queryAsyncJobResult&jobId=" + jid),
+                                        dataType: "json",
+                                        async: false,
+                                        success: function(json) {
+                                            wasDomainDeletedWithSuccess = json.queryasyncjobresultresponse.jobresultcode ==  0;
+                                        }
+                                    });
                                     args.response.success({
                                         _custom: {
                                             jobId: jid
                                         }
                                     });
+                                    if(!wasDomainDeletedWithSuccess){
+                                        return;
+                                    }
 
                                     // Quick fix for proper UI reaction to delete domain
                                     var $item = $('.name.selected').closest('li');


ui/scripts/domains.js
```diff
@@ -69,11 +69,23 @@
                                 async: false,
                                 success: function(json) {
                                     var jid = json.deletedomainresponse.jobid;
+                                    var wasDomainDeletedWithSuccess = false;
+                                    $.ajax({
+                                        url: createURL("queryAsyncJobResult&jobId=" + jid),
+                                        dataType: "json",
+                                        async: false,
```
rhtyd: async `false` will block the UI... is there a way to have this but without blocking the UI?

rafaelweingartner: that is a good question. If you take a look, the upper method "deleteDomain" (at line 69), it is already an "async" call. I could for instance, move the code from lines 81-93 to the success function (line 78), but that can cause an odd behavior if the user clicks somewhere in the UI before the method finishing. That is why I decided to roll with the "standard" and use a "sync" call as well.

rafaelweingartner: @rhtyd do you think this can be a problem?

rafaelweingartner:
@DaanHoogland and @nvazquez can you guys provide some feedback regarding this PR as well?

ui/scripts/domains.js
rhtyd: @rafaelweingartner sorry for a late reply, you don't need to query job results manually but use `args.response.success` and pass the `jobId`. Ideally an async notification callback could have been explored and used.

rafaelweingartner: That is actually what is being used right now. I tried though to use the " args.response.success" with a custom function to render the HTML in case of success. However, I was not able to find a way to send a function there. The only way I have seen people using is the following:

```
                                    args.response.success({
                                        _custom: {
                                            jobId: jid
                                        }
                                    });
```
Which takes only the jobid into account, and does not provide me a callback. 

Would you like me to revert this one then?

Issue #2909, opened 2018-10-18
debian does not have libvirt-bin it has libvirt


querplis:
https://github.com/apache/cloudstack/blob/7667846bfa70704cfe785ffe29fc13f0d066a765/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtPostCertificateRenewalCommandWrapper.java#L41-L45

https://github.com/apache/cloudstack/blob/7313a5ead2eb6db570647225f1ec9cf9e703455b/debian/rules#L47-L48

https://github.com/apache/cloudstack/blob/7313a5ead2eb6db570647225f1ec9cf9e703455b/debian/control#L25

https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/python/lib/cloud_utils.py#L706-L770

Debian does not have libvirt-bin it has libvirtd

https://packages.debian.org/search?suite=stable&section=all&arch=any&searchon=names&keywords=libvirt

rhtyd:
@querplis officially supported KVM hosts are RHEL/CentOS and Ubuntu (http://docs.cloudstack.apache.org/en/4.11.1.0/releasenotes/compat.html#supported-hypervisor-versions). Please submit a pull request for debian.
@wido can you comment if you've ever tested on Debian+KVM?

rhtyd:
The fix is simply that only in case of Ubuntu the process name is libvirt-bin, otherwise is should be libvirtd.

wido:
I think we can completely switch to the name *libvirtd*, I just tested on Ubuntu 16.04:

<pre>systemctl status libvirtd</pre>

<pre>cat /lib/systemd/system/libvirt-bin.service</pre>

<pre>[Install]
Alias=libvirtd.service
</pre>

As you can see, on Ubuntu it creates an alias.

rhtyd:
I think on 14.04 it's still libvirt-bin @wido.

wido:
We dropped Ubuntu 14.04 support in master already, so we can ignore Ubuntu 14.04 on master.

rhtyd:
Alright, in that case let's switch to libvirtd service name. /cc @sh0wstopper

wido:
Seems good to me! Looking forward to the PR

querplis:
it looks like this still mentions libvirt-bin
[...]

PR #2912, opened 2018-10-21
kvm: Use libvirtd as service name for libvirt for Debian (#2909)

sh0wstopper:
Debian does not have libvirt-bin package. Therefore, only for Ubuntu host distro process name libvirt-bin is used, otherwise libvirtd will be used

## Description
As discussed in #2909 libvirt-bin is not resent in Debian but it is in Ubuntu. Host check should only be for Ubuntu and not for Debian.
cloudstack/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtPostCertificateRenewalCommandWrapper.java is changed to fix this.


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

Issue #2910, opened 2018-10-18
mvn needs -B switch


querplis:
https://github.com/apache/cloudstack/blob/233f46c94b0b792d3f674cda120066a97e04c8ea/tools/build/setnextversion.sh#L127

without -B ( batch mode ), if maven has to download anything it will end in variable.

worth to change here too, not critical tho
https://github.com/apache/cloudstack/blob/233f46c94b0b792d3f674cda120066a97e04c8ea/tools/build/setnextversion.sh#L131

rhtyd:
LGTM, @querplis can you send a pull request to master?

rhtyd:
Fixed in PR. Closing.

PR #2914, opened 2018-10-22
tools: Added '-B' flag for mvn current version cmd (#2910)

sh0wstopper:
maven command for finding current version might need to download packages and without batch mode it will end in variable. Added '-B' to enable batch mode.

## Description
Added '-B' to enable batch mode in maven command for find current version maven command in cloudstack/tools/build/setnextversion.sh shell script.


rhtyd:
Since the change is in the release script and Travis passed, we don't need smoketests for this. Merging now.

PR #2979, opened 2018-10-29
vr: defer was broken in VR because of json name change

rhtyd:
After upgrade from CS 4.10 to CS 4.11, multiple VRs did not start through.
It did not properly defer the finalize config in update_config.py.
Apparently, the json files are now called differently: where it used to
be vm_dhcp_entry.json it now has a uuid added, for example
vm_metadata.json.4d727b6e-2b48-49df-81c3-b8532f3d6745.
The if statement that checks if the finalize can be safely deferred
therefore no longer matches. This PR contains a fix so finalize is
defered again.

Rohit: I made a boo-boo by force pushing and trying to rebash on @RPDiep 's repo/branch. Submitting his commit on his behalf against 4.11 branch. Sorry @RPDiep /cc @wido.

## Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/update_config.py
+++ b/systemvm/debian/opt/cloud/bin/update_config.py
@@ -52,7 +52,7 @@ def process_file():
     qf.setFile(sys.argv[1])
     qf.load(None)
     # These can be safely deferred, dramatically speeding up loading times
-    if not (os.environ.get('DEFER_CONFIG', False) and sys.argv[1] in ('vm_dhcp_entry.json', 'vm_metadata.json')):
+    if not (os.environ.get('DEFER_CONFIG', False) and ('vm_dhcp_entry.json' in sys.argv[1] or 'vm_metadata.json' in sys.argv[1])):
         # Converge
         finish_config()
 


rhtyd:
For reference, this was authored by @RPDiep and he will not lose his authorship on the commit. We can use my review/lgtm from previous PR:
screenshot from 2018-10-29 15-18-38

rhtyd:
Our lab is temporary under maintenance, I'll tested this locally in a [monkeybox](https://github.com/rhtyd/monkeybox)+KVM based env and LGTM. Will merge this as soon as travis is OK.

RPDiep:
No problem on the new PR.
I just read that @rhtyd  preferred that the line would use `startswith`?
I think that's indeed better but this PR is now already merged.
Shall I prepare a PR to incorporate that?

wido:
I doubt that *startswith* is the best method. We are now just providing a filename, but if somebody suddenly decides to change this to a absolute path that breaks again.

rhtyd:
Alright, no need of a new PR. In some cases the path may be absolute, but I think we've moved to relative paths now in most cases

Issue #2890, opened 2018-10-08
KVMHAMonitor thread blocks indefinitely while NFS not available


csquire:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
KVM Agent
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2.0-41120rc2
~~~

##### SUMMARY

Also see comment thread on PR #2722

We installed an RC release which includes PR #2722 on a test system expecting the host to get marked as `Disconnected` after using iptables to drop NFS requests, but instead the host gets marked as  `Down`. My investigation shows that the line [`storage = conn.storagePoolLookupByUUIDString(uuid);`](https://github.com/apache/cloudstack/blob/4.11/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java#L95) blocks indefinitely. So, kvmheartbeat.sh is never executed, a host investigation is started, the host with blocked NFS is marked as Down and finally all VMs on that host are rescheduled and result in duplicate VMs.

I pulled a thread dump and found the KVMHAMonitor thread will hang here until NFS is unblocked.

  ```
 java.lang.Thread.State: RUNNABLE
        at com.sun.jna.Native.invokePointer(Native Method)
        at com.sun.jna.Function.invokePointer(Function.java:470)
        at com.sun.jna.Function.invoke(Function.java:404)
        at com.sun.jna.Function.invoke(Function.java:315)
        at com.sun.jna.Library$Handler.invoke(Library.java:212)
        at com.sun.proxy.$Proxy3.virStoragePoolLookupByUUIDString(Unknown Source)
        at org.libvirt.Connect.storagePoolLookupByUUIDString(Unknown Source)
        at com.cloud.hypervisor.kvm.resource.KVMHAMonitor$Monitor.runInContext(KVMHAMonitor.java:95)
        - locked <1afb3370> (a java.util.concurrent.ConcurrentHashMap)
        at org.apache.cloudstack.managed.context.ManagedContextRunnable$1.run(ManagedContextRunnable.java:49)
        at org.apache.cloudstack.managed.context.impl.DefaultManagedContext$1.call(DefaultManagedContext.java:56)
        at org.apache.cloudstack.managed.context.impl.DefaultManagedContext.callWithContext(DefaultManagedContext.java:103)
        at org.apache.cloudstack.managed.context.impl.DefaultManagedContext.runWithContext(DefaultManagedContext.java:53)
[...]

PR #2984, opened 2018-10-30
kvm: reset KVM host on heartbeat failure

rhtyd:
On actual testing, I could see that kvmheartbeat.sh script fails on NFS
server failure and stops the agent only. Any HA VMs could be launched
in different hosts, and recovery of NFS server could lead to a state
where a HA enabled VM runs on two hosts and can potentially cause
disk corruptions. In most cases, VM disk corruption will be worse than
VM downtime. I've kept the sleep interval between check/rounds but
reduced it to 10s. The change in behaviour was introduced in #2722.

## Types of changes

## How Has This Been Tested?

See comment: https://github.com/apache/cloudstack/issues/2890#issuecomment-434207830

### DIFF
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHABase.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/KVMHABase.java
@@ -35,7 +35,7 @@ public class KVMHABase {
     protected long _heartBeatUpdateTimeout = 60000;
     protected long _heartBeatUpdateFreq = 60000;
     protected long _heartBeatUpdateMaxTries = 5;
-    protected long _heartBeatUpdateRetrySleep = 15000;
+    protected long _heartBeatUpdateRetrySleep = 10000;
 
     public static enum PoolType {
         PrimaryStorage, SecondaryStorage
--- a/scripts/vm/hypervisor/kvm/kvmheartbeat.sh
+++ b/scripts/vm/hypervisor/kvm/kvmheartbeat.sh
@@ -155,10 +155,10 @@ then
   exit 0
 elif [ "$cflag" == "1" ]
 then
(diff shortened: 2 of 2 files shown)


rhtyd:
@blueorangutan package

wido:
LGTM

#2722 should have never been merged. If the host has a problem, it has to kill itself to prevent these weird issues and disk corruptions.

There is no very fancy way of doing this otherwise.

rhtyd:
Testing may not be needed as the changes are not covered by smoketests. Manually tested. Travis is green, however I'll kick test as an obligated process and merge this so RC4 can be cut soon.
@blueorangutan test

somejfn:
@wido Allow me to disagree.   If a transient NFS issue happens and result in mass reset of KVM hosts after 5 minutes, this creates way more damage/outages than waiting for the NFS server to come back and VM just resume IO to disk.    Last time we saw this it was a failed switch update that caused a network partition for 22 minutes.  On 4.9.3 we were lucky to survive the event while on 4.11 Im afraid I'd have to explain why 2000 VMs got rebooted.

rhtyd:
@somejfn most people would choose VM downtime over potential disk corruption

wido:
I understand @somejfn , but 22 minutes is a very long time. In this case you can't blame CloudStack for the fact that it would reboot,

Merely stopping the Agent is no guarantee that the VMs are really down.

You can argue on how long the timeout should be, 5, 10, 15 minutes, but at some point a HV would need to kill itself so that we can be sure that VMs are dead.

NFS does not provide proper locking of images, so two VMs could run from the same disk.

Downtime is not nice, but data corruption is horrible.

somejfn:
I guess VM-HA is handled differently on managed storage right ?   I mean
Ceph has image locking builtin and I hope Solidfire can do the same to
guarantee exclusive access.

On NFS the risk to keep HA is greater than its benefits so sadly I'll have
to disable it and humans handle host failures

PR #2985, opened 2018-10-30
kvm/ceph: When create a new RBD image use modern flags/features for t…

wido:
…he image

These additional RBD features allow for faster lookups of how much space a RBD
image is using, but with the exclusive locking we prevent two VMs from writing
to the same RBD image at the same time.

These are the default features used by Ceph for any new RBD image.

## Types of changes

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
@@ -72,7 +72,12 @@ public class LibvirtStorageAdaptor implements StorageAdaptor {
     private String _manageSnapshotPath;
 
     private String rbdTemplateSnapName = "cloudstack-base-snap";
-    private int rbdFeatures = (1 << 0); /* Feature 1<<0 means layering in RBD format 2 */
+    private static final int RBD_FEATURE_LAYERING = 1;
+    private static final int RBD_FEATURE_EXCLUSIVE_LOCK = 4;
+    private static final int RBD_FEATURE_OBJECT_MAP = 8;
+    private static final int RBD_FEATURE_FAST_DIFF = 16;
+    private static final int RBD_FEATURE_DEEP_FLATTEN = 32;
+    private int rbdFeatures = RBD_FEATURE_LAYERING + RBD_FEATURE_EXCLUSIVE_LOCK + RBD_FEATURE_OBJECT_MAP + RBD_FEATURE_FAST_DIFF + RBD_FEATURE_DEEP_FLATTEN;
     private int rbdOrder = 0; /* Order 0 means 4MB blocks (the default) */
 
     public LibvirtStorageAdaptor(StorageLayer storage) {


plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
```diff
@@ -72,7 +72,15 @@
...
-    private int rbdFeatures = (1 << 0); /* Feature 1<<0 means layering in RBD format 2 */
+    /*
+        Ceph RBD features to use when creating a new RBD image
+        (1 << 0): RBD_FEATURE_LAYERING
+        (1 << 2): RBD_FEATURE_EXCLUSIVE_LOCK
+        (1 << 3): RBD_FEATURE_OBJECT_MAP
+        (1 << 4): RBD_FEATURE_FAST_DIFF
+        (1 << 5): RBD_FEATURE_DEEP_FLATTEN
+     */
+    private int rbdFeatures = (1 << 0) + (1 << 2) + (1 << 3) + (1 << 4) + (1 << 5);
```
rhtyd: How about define the feature as various final int hex values and `rbdFeatures` can be the or-ed value of those features. Should there also be some kind of version vs feature check?

rhtyd:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
wido: Yeah, I could also just write **61** as that's the default.

I took these values from the C-header file from Ceph so that people see what we actually use.

That's mainly a code style thing.

GabrielBrascher: Agree that it is mainly a code style thing. But here are some suggestions as well.
```
private final int RBD_FEATURE_LAYERING = 1;
private final int RBD_FEATURE_EXCLUSIVE_LOCK = 4;
private final int RBD_FEATURE_OBJECT_MAP = 8;
private final int RBD_FEATURE_FAST_DIFF = 16;
private final int RBD_FEATURE_DEEP_FLATTEN = 32;
private int rbdFeatures = RBD_FEATURE_LAYERING + RBD_FEATURE_EXCLUSIVE_LOCK + RBD_FEATURE_OBJECT_MAP + RBD_FEATURE_FAST_DIFF + RBD_FEATURE_DEEP_FLATTEN;
```
or
```
/**
	Ceph RBD features to use when creating a new RBD image
        RBD_FEATURE_LAYERING: 1
        RBD_FEATURE_EXCLUSIVE_LOCK: 4
        RBD_FEATURE_OBJECT_MAP: 8
        RBD_FEATURE_FAST_DIFF: 16
        RBD_FEATURE_DEEP_FLATTEN: 32
	Total:61
*/
private final int RBF_FEATURES = 61;
```

DaanHoogland: I like the first suggestion @GabrielBrascher makes, @wido . it comes down to "self documenting code by proper naming"

wido:
Thanks for the feedback! See the updated code

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
```diff
@@ -72,7 +72,12 @@
     private String _manageSnapshotPath;
 
     private String rbdTemplateSnapName = "cloudstack-base-snap";
-    private int rbdFeatures = (1 << 0); /* Feature 1<<0 means layering in RBD format 2 */
+    private final int RBD_FEATURE_LAYERING = 1;
```
DaanHoogland: the errors with checkstuyle are due to the vars not being static (final). full capitalised is reserved for those constants only.

wido:
Yes, I trusted my IDE to help with this, but fixed it now! Compiles :)

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

GabrielBrascher:
@wido @DaanHoogland @rhtyd merging it to the master branch, based on code review (3 LGTMs) and that all checks have passed.

Issue #2991, opened 2018-11-01
library surface reduction


DaanHoogland:
library interface surfaces should be reduced to improve portability and upgradability. examples of lockins that cloudstack suffers from are gson, stringutils and even the completely Apache internal log4j1.2 to log4j2 upgrade is a problem because of this.
jira issue CLOUDSTACK-10089 asks for a log4j update.
other issues have been encountered in teh past (will add references if i encounter them)
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
cross cutting concern
~~~

##### CLOUDSTACK VERSION

~~~
all
~~~

##### SUMMARY

I think a number of PR need to be created for the mentioned utilities and maybe others to smoothly get to a maintainable situation.
Each of them must have all functionality used in CloudStack proxied in them.

##### STEPS TO REPRODUCE

~~~
when upgrading or migrating from a library to a library of similar funvctionality, changes should only have to be made in one area of the cloudstack code base. This is the ideal situation and may never be true for 3rd party plugins for instance. Big improvements can be made however on the mentioned libraries
* StringUtils
* Gson (json utilities)
* logging
~~~

##### EXPECTED RESULTS

~~~
one class or package at most has to be changed when upgrading or migrating
~~~

##### ACTUAL RESULTS

~~~
a plethora of hundreds of files need to be changed for each upgrade, risking numerous regressions due to oversights or misinterpretations of coding styles.
~~~

#####  PR list (will be trying to include dependencies in a tree like structure, here)
[ ] #2992

rohityadavcloud:
@DaanHoogland can we close this if we move to log4j 2.x in near future or is this still applicable?

DaanHoogland:
I will have to find new ways to get the idea into the heads of the community. This remains an issue we'll keep facing.

PR #2276, opened 2017-09-27
WIP: CLOUDSTACK-10089 log4j update to version 2

DaanHoogland:

PR #987, opened 2015-10-27
[wip] CLOUDSTACK-8677: use wrapper classes for gson serialisation

DaanHoogland:
non functional change with great risk: replace all basic type fields of json serializable objects to wrapper class type fields. It is needed for gson upgrade which is needed for phone home functionality.

PR #2992, opened 2018-11-01
PoC for log library surface reduction (2991)

DaanHoogland:
## Description

part of #2991, this is the initial step to unifying the log libraries and put the surface in a single location. Note that even for this one functionality about half the files in the entire project will be touched in the end. This one and the StringUtils one are probably the biggest. This and the gson one will be the more complicated than StringUtils. The DAO framework will have even more impact but the may be out of scope for #2991

## Types of changes


DaanHoogland:
@blueorangutan package

utils/src/main/java/org/apache/cloudstack/utils/log/Logger.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public void warn(String s, Exception e) {
+        logger.warn(s,e);
+    }
+
+    public void info(String s) {
+        logger.info(s);
+    }
+
+    public void error(String s) {
```
marcaurele: Can you add a`public void error (String msg, Throwable e)` and change the `public void warn(String s, Exception e)` to `public void warn(String s, Throwable e)`

marcaurele:
Good change to move all the logs under a single class to ease any future changes or choice of library.

utils/src/main/java/org/apache/cloudstack/utils/log/Logger.java
DaanHoogland: I will as we progress change the interface to what we really use. I am also still considering to use an actual java-interface. Unfortunately java requires an explicit 'implements', otherwise the proxy mechs would be some easier.

DaanHoogland:
@blueorangutan test

csquire:
Why not use SLF4J instead of reinventing your own facade?
As a developer of an internal plugin, I find it frustrating to be forced into using limited one-off implementations. In particular, I am sending managment server logs to ELK in a JSON format. With SLF4J, Logback with the [Logstash Logback Encoder](https://github.com/logstash/logstash-logback-encoder), I can use Markers to add event-specific custom fields wherever I want. This is very useful for searching and filtering logs in ELK, not to mention the potential that exists for visualization of log events based on custom JSON fields. I assume a similar thing can be done with SLF4J and log4j2, but I have not reasearched it much. But, if Cloudstack uses a homegrown facade, it will not have a full logging feature set available.

DaanHoogland:
@csquire hearsay admittedly but apache logging folk say that log4j2 is better faster neater than slf4j and when we consolidated the util package we can decide to wrap slf4j in it as well, right now slf4j is used at some places in the system, java logging at others and log4j version 1 mostly. I want to consolidate all logging after merging this PoC first and than upgrade to whatever we decide. The idea here is that we have only one place where the version of the 3rd party lib matters.

csquire:
Understood. I'd just like plugin development to be taken into consideration and by using at thin wrapper around logging that doesn't implement a full feature set, plugin developers would have to either bypass it to do what I want or accept that I cannot do certain things. I also understand that I'm probably in the minority (as far as Cloudstack goes) of wanting to do things like add event-specific custom fields to log events.

DaanHoogland:
happy to work with you on what you need, @csquire . We do want to implement a large enough subset so all are served. ATM we have a maintenance problem, removal of which is our first objective. If you add an issue for your requirements I will add it if my puny abilities allow. Do not see the implementation of this PoC as the final stage or full interface.

csquire:
Will do, thanks!

chiradeep:
+1

rhtyd:
@DaanHoogland I'll evaluate this soon and get back to you, thanks for the PR!

rhtyd:
@blueorangutan package

DaanHoogland:
hm, there where packages for the latest commit already. test had been run as well.

utils/src/main/java/org/apache/cloudstack/utils/log/LogFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cloudstack.utils.log;
+
+public class LogFactory {
+//  for log4j 1.2.x the actual LogFactory is org.apache.log4j.LogManager
```
rhtyd: Indent comment, may get checkstyle issue.

utils/src/main/java/com/cloud/utils/AutoCloseableUtil.java
```diff
@@ -16,10 +16,11 @@
 // under the License.
 package com.cloud.utils;
 
-import org.apache.log4j.Logger;
+import org.apache.cloudstack.utils.log.LogFactory;
+import org.apache.cloudstack.utils.log.Logger;
 
 public class AutoCloseableUtil {
-    private final static Logger s_logger = Logger.getLogger(AutoCloseableUtil.class);
+    private final static Logger s_logger = LogFactory.getLogger(AutoCloseableUtil.class);
```
rhtyd: I guess we'll need to refactor codebase wide to use LogFactory.

DaanHoogland: I would regard that out of scope for this PoC, but part of a series of follow-up PRs.

utils/src/main/java/org/apache/cloudstack/utils/log/LogFactory.java
DaanHoogland: will make sure to indent

DaanHoogland:
This is not a feature, @rafaelweingartner . It is a way to make the system a bit more maintainable and deal with technical debt.

rafaelweingartner:
I understand that is not a feature, but I do not agree that this will help. I will try to write something until tomorrow to explain why.

DaanHoogland:
ok, you will fail, but I will be patient for another day.

DaanHoogland:
@rafaelweingartner, I have no idea why someone would not want to do this. It is a best practice and in my opinion a blocker to upgrading log4j and any other logging frameworks. Please give me any arguments not to do this. I am pretty sure I can counter them.

DaanHoogland:
the travis build is successful but still reports an error here. I will extend this PoC somewhat to see what happened there.

rafaelweingartner:
Hey @DaanHoogland, sorry the delay. I have not had much time lately to work with CloudStack.

I am not seeing how this PR will reduce the surface of changes to upgrade to log4j2. You will still need to replace all of the points/places in the code where log4j-1 is currently used by your brand new eggshell (that will sit on top of a known logging framework such as log4j). Therefore, you are not reducing the surface of change right now. You would achieve some surface reduction if people adopted more Object orientation approach, and used less static methods/variables. I also do not see the value in adding this abstraction layer on top of a framework (in this case a logging framework). Why stop at the logging framework then? Why not create a layer of abstraction on top of Spring, JPA, and Web frameworks as well? At least that was our discussion in Montreal. We should move on towards standard implementations, and stop reinventing the well. For instance, your proposal could be applied to JPA (as it has different implementations too) as well, will we do that (when we move towards Spring-data and a real JPA implementation)? I hope not…

From log4j-1 to log4j-2, the methods  (logging ones such as debug, info, warn, and so on) did not change the signature. Only the classes and some other things moved to new places. Of course, new “methods/implementations/integrations” were added, but that is not the case we are dealing with here. Therefore, to execute the upgrade, you would only need to run a script such as the following:

```
#!/bin/bash
CLOUDSTACK_FOLDER="./"

ALL_CLOUDSTACK_CLASSES=$(find $CLOUDSTACK_FOLDER -name "*.java")

for class in $ALL_CLOUDSTACK_CLASSES
do
  count=$((count+1))
  echo "Processing ($count) class: $class"

  echo "Replacing org.apache.log4j.Logger with org.apache.logging.log4j.Logger"
  sed -i 's/org.apache.log4j.Logger/org.apache.logging.log4j.Logger/g' $class
  
  echo "Adding import to org.apache.logging.log4j.LogManager"
  sed -i '/org.apache.logging.log4j.Logger/a import org.apache.logging.log4j.LogManager;' $class
 
  echo "Replacing Logger.getLogger with LogManager.getLogger"
  sed -i 's/Logger.getLogger/LogManager.getLogger/g' $class

  echo "Replacing org.slf4j.Logger with org.apache.logging.log4j.Logger"
  sed -i 's/org.slf4j.Logger/org.apache.logging.log4j.Logger/g' $class


  echo "Replacing org.slf4j.LoggerFactory with org.apache.logging.log4j.Logger"
  sed -i 's/org.slf4j.LoggerFactory/org.apache.logging.log4j.LogManager/g' $class

  echo "Replacing LoggerFactory.getLogger with /LogManager.getLogger"
  sed -i 's/LoggerFactory.getLogger/LogManager.getLogger/g' $class
done
```

Moreover, as you can see by the script I created, someone else already did something similar what you are proposing with SLF4j. I mean, creating a POC, and hoping/wanting to do things on demand. The result…Nothing changed, and then we were using 2 different logging frameworks. With your proposal, we would have 3 different points/classes used for logging simultaneously.

I already took the liberty and did some work towards the upgrade and removing SLF4J. I only need now to solve the implementation of a custom SNMP logging mechanisms that was connected to log4j, and after that, the job is done. No need to create this abstraction layer.

DaanHoogland:
tl;dr you are not giving any argument to not use this strategy, you are only saying why this PoC is just a PoC.

What you are describing is very elaborate as well, @rafaelweingartner. The script you describe is error prone. Not all calls and declaration are formatted the same way and the expressions look a bit to simple.

I have seen the attempt to use slf4j in the console proxy code and it needs to be addressed as well. It is in the wrong place and incomplete.

you say "For instance, your proposal could be applied to JPA ... as well, will we do that (when we move towards Spring-data and a real JPA implementation)? I hope not…". I do hope so, and I think you should as well, but this will be very challenging in the DAO case. I think it would be great if we could.

I think we should definitely do this for a lot of external libraries, not reinventing the wheel but abstracting out and isolating dependencies on external eccentricities of those libraries, like call syntax/sematics. As you can see in my description of the issue (#2991) I am not being religious. I exclude (not definitely but as example) the DAO framework, as indeed spring data may make it more convenient to go about it differently. But NOTE: this is a temptation, luring us into a tie into the specifics of a framework we decide to commit ourself into in such a way.

As you can see in #987 and #2276 and, I am sure numerous upgrade attempts by others as well, what you describe is not as simple as you say without isolating the use of an external library.

I will follow up on this PR given mandate. we need to reduce the internal exposure to external libraries, not reinventing the wheel but isolating there use in proxy libraries. Your solution to upgrading log4j is not unifying logging framework use but only upgrading log4j as a one off solution essentially perpetuating the problem.

In short, this PoC in itself will not solve the problem. it needs follow up and I dedicate myself to that. I do not want a PR like #2276, with 1710 files changed if a dependency decides to change a signature or a exposed structure. This will disable merging forward and hinder regular maintenance.

rafaelweingartner:
> Also, @rafaelweingartner, I will -1 a PR that 'just' upgrades a structurally changed dependency, if it fails to makes efforts to isolate the dependency as described.

What do you mean with "isolate the dependency"? It is a dependency, we choose it (wisely), we evaluate it, and we move on.

As I said, there are different ways to tackle the issues. One is for sure doing what you are proposing and creating a layer on top of everything. Other would be a better Class hierarchy and using properly object orientation.

Anyways, to upgrade the log4j is not as a Dantesc task as you are portraiting. I am almost done, and I have almost had no time to work on it. I can finish, and then show you for us to discuss if you have an open mind.

DaanHoogland:
> Anyways, to upgrade the log4j is not as a Dantesc task as you are portraiting. I am almost done, and I have almost had no time to work on it. I can finish, and then show you for us to discuss if you have an open mind.
If you unify the log framework usage as well, I am open to your solution.

rafaelweingartner:
> If you unify the log framework usage as well, I am open to your solution.
That is actually what I am doing. That is why it is taking some time (time that I have not had much lately). Also, as you noticed the simple script I created does not address some very specific case in our code base. Therefore, I need to visit them manually, and to evaluate if they are really needed, and if we can move on to the use of standard variables.

DaanHoogland:
so basically you are re-doing my work from #2276 but without reducing the surface area. I do not like it much so far.

DaanHoogland:
@rafaelweingartner , can you enter something of a PoC PR so we can discuss further?
Or sugest changes to this one?

rafaelweingartner:
Sure I will

DaanHoogland:
@rafaelweingartner I would like to merge this and build on it. I have not seen any follow up. can you expand or comply? /cc @GabrielBrascher

rafaelweingartner:
@DaanHoogland I have not had free time to work on these things lately. I also do not see the benefits of this PR.

But, besides that, master is on freeze, and we are working to release 4.12. Therefore, this PR has to wait in one way or the other.

DaanHoogland:
> But, besides that, master is on freeze, and we are working to release 4.12. Therefore, this PR has to wait in one way or the other.

It was in way before the freeze so I don't see why it has to wait.

rafaelweingartner:
I do know you are not asking my help ;)

The part I disagree with is the method used to do the so called "reduction of the surface of change". There are other alternatives, which in my opinion work better. I tried to explain, but I do not think you understood what I meant. Therefore, I tried to create an example for you, but I have also not been able to finish it.

GabrielBrascher:
@DaanHoogland thanks for the initiative!

There are still open concerns to address and to be fair, now that I had some time to review it, I am sharing the same thoughts of Rafael.

I will be happy to assist in reviewing deep further and discussing different approaches; however,  considering our tight schedule and the lack of reviewers, I am afraid that this one will have to wait for the freeze.

DaanHoogland:
ok, then let me urge to give this priority. We are using antiquated libraries, as in out of service -, out of support libraries and moving forward will not get easier over time.

If you method is the scripting search and replace of all methods used, I am afraid this is just a one of solution and not an architectural pattern for future stability. Please enlighten me because indeed I do not understand the alternative to be viable.

GabrielBrascher:
I totally agree with you @DaanHoogland, this is an issue to give priority and as soon as we have 4.12.0.0 released this should be one of the priorities among all the CloudStack refactoring efforts.
Unfortunately, I think that for this release it is still a bit raw.

I would like to add that our next LTS should be delivered this year to cover the 4.11 end of life (1 July 2019) [1]. I was absent on this discussions so far but will address some effort as well to help on getting rid of all the antiquated libraries on time to merge into the next LTS.

[1] https://cwiki.apache.org/confluence/display/CLOUDSTACK/LTS

rafaelweingartner:
@DaanHoogland on a second thought. Feel free to do what you would like here ;)

I think I gave a -1, but I will not be able to finish what I started to show you. Therefore, I hereby remove it.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

DaanHoogland:
@rafaelweingartner I saw you IM comment and will look into what a spring suit would mean for logging. more later

DaanHoogland:
used a addepted version of @rafaelweingartner's script above : 
```
#!/bin/bash
CLOUDSTACK_FOLDER="."

ALL_CLOUDSTACK_CLASSES=$(find $CLOUDSTACK_FOLDER -name "*.java")

for class in $ALL_CLOUDSTACK_CLASSES
do
  count=$((count+1))
  echo "Processing ($count) class: $class"

  echo "Replacing org.apache.log4j.Level with org.apache.cloudstack.utils.log.Level"
  sed -i .loch 's/org.apache.log4j.Level/org.apache.cloudstack.utils.log.Level/g' "$class"

  echo "Replacing org.apache.log4j.Logger with org.apache.cloudstack.utils.log.Logger"
  sed -i .loch 's/org.apache.log4j.Logger/org.apache.cloudstack.utils.log.Logger/g' "$class"

  echo "Replacing java.util.logging.Logger with org.apache.cloudstack.utils.log.Logger"
  sed -i .loch 's/java.util.logging.Logger/org.apache.cloudstack.utils.log.Logger/g' "$class"

  echo "Adding import to org.apache.cloudstack.utils.log.LogFactory"
  sed -i .loch '/import org.apache.cloudstack.utils.log.Logger/a \
import org.apache.cloudstack.utils.log.LogFactory;
' "$class"

  echo "Replacing Logger.getLogger with LogFactory.getLogger"
  sed -i .loch 's/Logger.getLogger/LogFactory.getLogger/g' "$class"

  echo "Replacing org.slf4j.Logger with org.apache.cloudstack.utils.log.Logger"
  sed -i .loch 's/org.slf4j.Logger/org.apache.cloudstack.utils.log.Logger/g' "$class"

  echo "Replacing org.slf4j.LoggerFactory with org.apache.cloudstack.utils.log.LogFactory"
  sed -i .loch 's/org.slf4j.LoggerFactory/org.apache.cloudstack.utils.log.LogFactory/g' "$class"

  echo "Replacing s_logger with LOG"
  sed -i .loch 's/s_logger/LOG/g' "$class"
done
```
the above scripts leaves some to be desired like:
# I don't have a constructor taking a String, yet. there where only a handful of those and all used a Class.getname, so I converted the calls instead.
# on occasion the script led to a double insertion of the factory import. fixed those by hand as well

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
framework/managed-context uses slf4j and is not dependend on utils. This needs to be solved by
1. isolating logging in its own project and make it second to checkstyle
1. exclude managed context from the standard use of logging

rhtyd:
Trillian env failed to deploy twice, I've rekicked the job again. cc @DaanHoogland

DaanHoogland:
no public support

PR #2993, opened 2018-11-01
Remove api rate limiting plugin

marcaurele:
## Description
Remove the API rate limiting plugin. This will ease the removal of ehcache in CS completely, and this feature can be configured on a HTTP frontend like Nginx to perform the same job.

[Discussion](https://lists.apache.org/thread.html/5699105e2153408fc35556409ef63499fc2c973e2d5c784130b5f617@%3Cdev.cloudstack.apache.org%3E) in the dev mailing list.

Link to initial discussion about ehcache: #2913.

The PR can be closed if people still want this feature in the code.

## Types of changes


rhtyd:
Looks like this may be in use by users, we can deprecate this over time but that will be a long process than simply remove it.

marcaurele:
I would have changed the documentation to state the alternative way with Nginx and its rate-limiting plugin. It would be equivalent.

PR #2983, opened 2018-10-29
KVM live storage migration intra cluster from NFS source and destination

nvazquez:
## Description

Feature Specification: https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=95653548

Live storage migration on KVM under these conditions:
- From source and destination hosts within the same cluster
- From NFS primary storage to NFS cluster-wide primary storage
- Source NFS and destination NFS storage mounted on hosts

In order to enable this functionality, database should be updated in order to enable live storage capacibilty for KVM, if previous conditions are met. This is due to existing conflicts between qemu and libvirt versions. This has been tested on CentOS 6 hosts.

Additional notes:

1. To use this feature set the `storage_motion_supported=1` in the hypervisor_capability table for KVM. This is done by default as the feature may not work in some environments, read below.
2. This feature of online storage+VM migration for KVM will only work with CentOS6 and possible Ubuntu as KVM hosts but not with CentOS7 due to:
- https://bugs.centos.org/view.php?id=14026
- https://bugzilla.redhat.com/show_bug.cgi?id=1219541
On CentOS7 the error we see is: `" error: unable to execute QEMU command 'migrate': this feature or command is not currently supported"` (reference https://ask.openstack.org/en/question/94186/live-migration-unable-to-execute-qemu-command-migrate/). Reading through various lists looks like the migrate feature with qemu may be available with paid versions of RHEL-EV but not centos7 however this works with CentOS6.

Fix for CentOS 7:
- Create repo file on /etc/yum.repos.d/:
````
[qemu-kvm-rhev]
name=oVirt rebuilds of qemu-kvm-rhev
baseurl=http://resources.ovirt.org/pub/ovirt-3.5/rpm/el7Server/
mirrorlist=http://resources.ovirt.org/pub/yum-repo/mirrorlist-ovirt-3.5-el7Server
enabled=1
skip_if_unavailable=1
gpgcheck=0
````
- `yum install qemu-kvm-common-ev-2.3.0-29.1.el7.x86_64 qemu-kvm-ev-2.3.0-29.1.el7.x86_64 qemu-img-ev-2.3.0-29.1.el7.x86_64`
- Reboot host

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
[...]

PR #2997, opened 2018-11-03
Allow KVM VM live migration with ROOT volume on file storage type

GabrielBrascher:
This PR proposes a new `DataMotionStrategy` `KvmNonManagedStorageDataMotionStrategy` that extends the `StorageSystemDataMotionStrategy`. This strategy allows migrating KVM VMs with the ROOT volume in  a primary local file system storage (volumes on path /var/lib/libvirt/images/).

This design also aims to ease the development of any non-managed KVM volume migration, such as the implementation of @nvazquez in PR #2983 (NFS to NFS).

### DIFF
--- a/core/src/main/java/com/cloud/agent/api/MigrateCommand.java
+++ b/core/src/main/java/com/cloud/agent/api/MigrateCommand.java
@@ -19,7 +19,9 @@
 
 package com.cloud.agent.api;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import com.cloud.agent.api.to.VirtualMachineTO;
@@ -33,6 +35,7 @@ public class MigrateCommand extends Command {
     private boolean isWindows;
     private VirtualMachineTO vmTO;
     private boolean executeInSequence = false;
+    private List<MigrateDiskInfo> migrateDiskInfoList = new ArrayList<>();
 
     protected MigrateCommand() {
     }
@@ -90,6 +93,14 @@ public class MigrateCommand extends Command {
         return executeInSequence;
     }
 
+    public List<MigrateDiskInfo> getMigrateDiskInfoList() {
+        return migrateDiskInfoList;
+    }
+
+    public void setMigrateDiskInfoList(List<MigrateDiskInfo> migrateDiskInfoList) {
+        this.migrateDiskInfoList = migrateDiskInfoList;
+    }
+
     public static class MigrateDiskInfo {
         public enum DiskType {
             FILE, BLOCK;
@@ -123,6 +134,7 @@ public class MigrateCommand extends Command {
         private final DriverType driverType;
         private final Source source;
         private final String sourceText;
+        private boolean isSourceDiskOnStorageFileSystem;
 
         public MigrateDiskInfo(final String serialNumber, final DiskType diskType, final DriverType driverType, final Source source, final String sourceText) {
             this.serialNumber = serialNumber;
@@ -150,6 +162,14 @@ public class MigrateCommand extends Command {
 
         public String getSourceText() {
             return sourceText;
+        }
+
+        public boolean isSourceDiskOnStorageFileSystem() {
+            return isSourceDiskOnStorageFileSystem;
+        }
+
+        public void setSourceDiskOnStorageFileSystem(boolean isDiskOnFileSystemStorage) {
+            this.isSourceDiskOnStorageFileSystem = isDiskOnFileSystemStorage;
         }
     }
 }
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -30,7 +30,6 @@ import org.apache.cloudstack.engine.subsystem.api.storage.TemplateDataFactory;
 import org.apache.cloudstack.engine.subsystem.api.storage.TemplateInfo;
 import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
-import org.springframework.stereotype.Component;
 
 import com.cloud.agent.api.Answer;
 import com.cloud.agent.api.MigrateCommand;
@@ -51,7 +50,6 @@ import com.cloud.vm.DiskProfile;
  * Extends {@link StorageSystemDataMotionStrategy}, allowing KVM hosts to migrate VMs with the ROOT volume on a non managed local storage pool.
  * As {@link StorageSystemDataMotionStrategy} is considering KVM, this implementation also migrates only from/to KVM hosts.
  */
-@Component
 public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMotionStrategy {
 
     @Inject
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.cloudstack.storage.motion;
+
+import java.io.File;
+import java.util.Map;
+import java.util.Set;
+
+import javax.inject.Inject;
+
+import org.apache.cloudstack.engine.subsystem.api.storage.DataStore;
+import org.apache.cloudstack.engine.subsystem.api.storage.StrategyPriority;
+import org.apache.cloudstack.engine.subsystem.api.storage.TemplateDataFactory;
+import org.apache.cloudstack.engine.subsystem.api.storage.TemplateInfo;
+import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
+import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
+import org.apache.commons.lang.StringUtils;
+import org.springframework.stereotype.Component;
+
+import com.cloud.agent.api.Answer;
+import com.cloud.agent.api.MigrateCommand;
+import com.cloud.agent.api.MigrateCommand.MigrateDiskInfo;
+import com.cloud.agent.api.storage.CreateAnswer;
+import com.cloud.agent.api.storage.CreateCommand;
+import com.cloud.agent.api.to.VirtualMachineTO;
+import com.cloud.host.Host;
+import com.cloud.hypervisor.Hypervisor.HypervisorType;
+import com.cloud.storage.DataStoreRole;
+import com.cloud.storage.DiskOfferingVO;
+import com.cloud.storage.Storage.StoragePoolType;
+import com.cloud.storage.VolumeVO;
+import com.cloud.utils.exception.CloudRuntimeException;
+import com.cloud.vm.DiskProfile;
+
+/**
+ * Extends {@link StorageSystemDataMotionStrategy}, allowing KVM hosts to migrate VMs with the ROOT volume on a non managed local storage pool.
+ * As {@link StorageSystemDataMotionStrategy} is considering KVM, this implementation also migrates only from/to KVM hosts.
+ */
+@Component
+public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMotionStrategy {
+
+    @Inject
+    private TemplateDataFactory tmplFactory;
+
+    /**
+     * Uses the canHandle from the Super class {@link StorageSystemDataMotionStrategy}. If the storage pool is of file and the internalCanHandle from {@link StorageSystemDataMotionStrategy} CANT_HANDLE, returns the HIGHEST strategy priority. otherwise returns CANT_HANDLE.
+     * Note that the super implementation (override) is called by {@link #canHandle(Map, Host, Host)} which ensures that {@link #internalCanHandle(Map)} will be executed only if the source host is KVM.
+     */
+    @Override
+    protected StrategyPriority internalCanHandle(Map<VolumeInfo, DataStore> volumeMap) {
+        if (super.internalCanHandle(volumeMap) == StrategyPriority.CANT_HANDLE) {
+            Set<VolumeInfo> volumeInfoSet = volumeMap.keySet();
+
+            for (VolumeInfo volumeInfo : volumeInfoSet) {
+                StoragePoolVO storagePoolVO = _storagePoolDao.findById(volumeInfo.getPoolId());
+                if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem) {
+                    return StrategyPriority.CANT_HANDLE;
+                }
+            }
+            return StrategyPriority.HYPERVISOR;
+        }
+        return StrategyPriority.CANT_HANDLE;
+    }
+
+    /**
+     * Configures a {@link MigrateDiskInfo} object configured for migrating a File System volume and calls rootImageProvisioning.
+     */
+    @Override
+    protected MigrateCommand.MigrateDiskInfo configureMigrateDiskInfo(VolumeInfo srcVolumeInfo, String destPath) {
+        return new MigrateCommand.MigrateDiskInfo(srcVolumeInfo.getPath(), MigrateCommand.MigrateDiskInfo.DiskType.FILE, MigrateCommand.MigrateDiskInfo.DriverType.QCOW2,
+                MigrateCommand.MigrateDiskInfo.Source.FILE, destPath);
+    }
+
+    /**
+     * Generates the volume path by appending the Volume UUID to the Libvirt destiny images path.</br>
+     * Example: /var/lib/libvirt/images/f3d49ecc-870c-475a-89fa-fd0124420a9b
+     */
+    @Override
+    protected String generateDestPath(VirtualMachineTO vmTO, VolumeVO srcVolume, Host destHost, StoragePoolVO destStoragePool, VolumeInfo destVolumeInfo) {
+        DiskOfferingVO diskOffering = _diskOfferingDao.findById(srcVolume.getDiskOfferingId());
+        DiskProfile diskProfile = new DiskProfile(destVolumeInfo, diskOffering, HypervisorType.KVM);
+        String templateUuid = getTemplateUuid(destVolumeInfo.getTemplateId());
+        CreateCommand rootImageProvisioningCommand = new CreateCommand(diskProfile, templateUuid, destStoragePool, true);
+
+        Answer rootImageProvisioningAnswer = _agentMgr.easySend(destHost.getId(), rootImageProvisioningCommand);
+
+        if (rootImageProvisioningAnswer == null) {
+            throw new CloudRuntimeException(String.format("Migration with storage of vm [%s] failed while provisioning root image", vmTO.getName()));
+        }
+
+        if (!rootImageProvisioningAnswer.getResult()) {
+            throw new CloudRuntimeException(String.format("Unable to modify target volume on the host [host id:%s, name:%s]", destHost.getId(), destHost.getName()));
+        }
+
+        String libvirtDestImgsPath = StringUtils.EMPTY;
+        if (rootImageProvisioningAnswer instanceof CreateAnswer) {
+            libvirtDestImgsPath = ((CreateAnswer)rootImageProvisioningAnswer).getVolume().getName() + File.separator;
+        }
+        return libvirtDestImgsPath + destVolumeInfo.getUuid();
+    }
+
+    /**
+     * Returns the template UUID of the given {@link VolumeVO}.
+     */
+    protected String getTemplateUuid(Long templateId) {
+        if (templateId == null) {
+            return null;
+        }
+        TemplateInfo templateImage = tmplFactory.getTemplate(templateId, DataStoreRole.Image);
+        return templateImage.getUuid();
+    }
+
+    /**
+     * Sets the volume path as the volume UUID.
+     */
+    @Override
+    protected void setVolumePath(VolumeVO volume) {
+        volume.setPath(volume.getUuid());
+    }
+}
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -30,7 +30,6 @@ import org.apache.cloudstack.engine.subsystem.api.storage.TemplateDataFactory;
 import org.apache.cloudstack.engine.subsystem.api.storage.TemplateInfo;
 import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
-import org.apache.commons.lang.StringUtils;
 import org.springframework.stereotype.Component;
 
 import com.cloud.agent.api.Answer;
@@ -56,10 +55,10 @@ import com.cloud.vm.DiskProfile;
 public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMotionStrategy {
 
     @Inject
-    private TemplateDataFactory tmplFactory;
+    private TemplateDataFactory templateDataFactory;
 
     /**
-     * Uses the canHandle from the Super class {@link StorageSystemDataMotionStrategy}. If the storage pool is of file and the internalCanHandle from {@link StorageSystemDataMotionStrategy} CANT_HANDLE, returns the HIGHEST strategy priority. otherwise returns CANT_HANDLE.
+     * Uses the canHandle from the Super class {@link StorageSystemDataMotionStrategy}. If the storage pool is of file and the internalCanHandle from {@link StorageSystemDataMotionStrategy} CANT_HANDLE, returns the StrategyPriority.HYPERVISOR strategy priority. otherwise returns CANT_HANDLE.
      * Note that the super implementation (override) is called by {@link #canHandle(Map, Host, Host)} which ensures that {@link #internalCanHandle(Map)} will be executed only if the source host is KVM.
      */
     @Override
@@ -69,7 +68,7 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
 
             for (VolumeInfo volumeInfo : volumeInfoSet) {
                 StoragePoolVO storagePoolVO = _storagePoolDao.findById(volumeInfo.getPoolId());
-                if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem) {
+                if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem && storagePoolVO.getPoolType() != StoragePoolType.NetworkFilesystem) {
                     return StrategyPriority.CANT_HANDLE;
                 }
             }
@@ -108,21 +107,22 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
             throw new CloudRuntimeException(String.format("Unable to modify target volume on the host [host id:%s, name:%s]", destHost.getId(), destHost.getName()));
         }
 
-        String libvirtDestImgsPath = StringUtils.EMPTY;
+        String libvirtDestImgsPath = null;
         if (rootImageProvisioningAnswer instanceof CreateAnswer) {
-            libvirtDestImgsPath = ((CreateAnswer)rootImageProvisioningAnswer).getVolume().getName() + File.separator;
+            libvirtDestImgsPath = ((CreateAnswer)rootImageProvisioningAnswer).getVolume().getName();
         }
-        return libvirtDestImgsPath + destVolumeInfo.getUuid();
+        // File.getAbsolutePath is used to keep the file separator as it should be and eliminate a verification to check if exists a file separator in the last character of libvirtDestImgsPath.
+        return new File(libvirtDestImgsPath, destVolumeInfo.getUuid()).getAbsolutePath();
     }
 
     /**
-     * Returns the template UUID of the given {@link VolumeVO}.
+     * Returns the template UUID with the given id. If the template ID is null, it returns null.
      */
     protected String getTemplateUuid(Long templateId) {
         if (templateId == null) {
             return null;
         }
-        TemplateInfo templateImage = tmplFactory.getTemplate(templateId, DataStoreRole.Image);
+        TemplateInfo templateImage = templateDataFactory.getTemplate(templateId, DataStoreRole.Image);
         return templateImage.getUuid();
     }
 
@@ -132,5 +132,14 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
     @Override
     protected void setVolumePath(VolumeVO volume) {
         volume.setPath(volume.getUuid());
+    }
+
+    /**
+     * Return true if the volume should be migrated. Currently only supports migrating volumes on storage pool of the type StoragePoolType.Filesystem.
+     * This ensures that volumes on shared storage are not migrated and those on local storage pools are migrated.
+     */
+    @Override
+    protected boolean shouldMigrateVolume(StoragePoolVO sourceStoragePool, Host destHost, StoragePoolVO destStoragePool) {
+        return sourceStoragePool.getPoolType() == StoragePoolType.Filesystem;
     }
 }
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
@@ -1687,10 +1687,12 @@ public class StorageSystemDataMotionStrategy implements DataMotionStrategy {
 
     /**
      * For each disk to migrate:
-     *   Create a volume on the target storage system.
-     *   Make the newly created volume accessible to the target KVM host.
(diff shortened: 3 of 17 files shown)


engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,142 @@
...
+     * Note that the super implementation (overrided) is called by {@link #canHandle(Map, Host, Host)} which ensures that {@link #internalCanHandle(Map)} will be executed only if the source host is KVM.
+     */
+    @Override
+    protected StrategyPriority internalCanHandle(Map<VolumeInfo, DataStore> volumeMap) {
+        if (super.internalCanHandle(volumeMap) == StrategyPriority.CANT_HANDLE) {
+            Set<VolumeInfo> volumeInfoSet = volumeMap.keySet();
+
+            for (VolumeInfo volumeInfo : volumeInfoSet) {
+                StoragePoolVO storagePoolVO = _storagePoolDao.findById(volumeInfo.getPoolId());
+                if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem) {
```
wido: What would happen if the VM has a ROOT volume on a local disk (FileSystem) and has one or more data disks on NFS or Ceph for example?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,142 @@
...
+            throw new CloudRuntimeException(String.format("Migration with storage of vm [%s] failed while provisioning root image", vmTO.getName()));
+        }
+
+        if (!rootImageProvisioningAnswer.getResult()) {
+            throw new CloudRuntimeException(String.format("Unable to modify target volume on the host [host id:%s, name:%s]", destHost.getId(), destHost.getName()));
+        }
+
+        String libvirtDestImgsPath = StringUtils.EMPTY;
+        if (rootImageProvisioningAnswer instanceof CreateAnswer) {
+            libvirtDestImgsPath = ((CreateAnswer)rootImageProvisioningAnswer).getVolume().getName() + "/";
```
wido: Would it not be better to use:

<pre>File.separator</pre>

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
```diff
@@ -203,8 +207,13 @@ Use VIR_DOMAIN_XML_SECURE (value = 1) prior to v1.0.0.
...
 
             if (destDomain != null) {
-                for (final DiskDef disk : disks) {
-                    libvirtComputingResource.cleanupDisk(disk);
+                for (DiskDef disk : disks) {
+                    MigrateDiskInfo migrateDiskInfo = searchDiskDefOnMigrateDiskInfoList(migrateDiskInfoList, disk);
+                    if (migrateDiskInfo != null && migrateDiskInfo.isSourceDiskOnStorageFileSystem()) {
+                        deleteLocalVolume(disk.getDiskPath());
+                    } else {
+                        libvirtComputingResource.cleanupDisk(disk);
```
wido: Would this clean up a disk which is non-local? Isn't that a potential danger that you remove a data disk?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -1991,9 +2034,16 @@ private void verifyLiveMigrationMapForKVM(Map<VolumeInfo, DataStore> volumeDataS
...
-                throw new CloudRuntimeException("Migrating a volume online with KVM can currently only be done when moving to managed storage.");
-            }
+            throwExeptionIfCannotLiveMigrateOnANonManagedStorage(destStoragePoolVO);
+        }
+    }
+
+    /**
+     * It throws a {@link CloudRuntimeException} if the destiny storage pool is not a managed storage pool.
+     */
+    protected void throwExeptionIfCannotLiveMigrateOnANonManagedStorage(StoragePoolVO destStoragePoolVO) {
```
nvazquez: I only find one usage of this method, is it planned to be invoked from elsewhere? Can we also rename it to something like 'checkDestinationStoragePool' ?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
```diff
@@ -51,19 +76,19 @@ public MigrateKVMAsync(final LibvirtComputingResource libvirtComputingResource,
...
 
         // set compression flag for migration, if libvirt version supports it
         if (dconn.getLibVirVersion() >= 1000003) {
-            flags |= 1 << 11;
+            flags += VIR_MIGRATE_COMPRESSED;
         }
 
         if (migrateStorage) {
-            flags |= 1 << 6;
+            flags += VIR_MIGRATE_NON_SHARED_DISK;
```
nvazquez: Really liked these changes :thumbsup:

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStoragePoolManager.java
```diff
@@ -138,20 +139,26 @@ public boolean connectPhysicalDisksViaVmSpec(VirtualMachineTO vmSpec) {
         List<DiskTO> disks = Arrays.asList(vmSpec.getDisks());
 
         for (DiskTO disk : disks) {
-            if (disk.getType() != Volume.Type.ISO) {
-                VolumeObjectTO vol = (VolumeObjectTO)disk.getData();
-                PrimaryDataStoreTO store = (PrimaryDataStoreTO)vol.getDataStore();
-                KVMStoragePool pool = getStoragePool(store.getPoolType(), store.getUuid());
+            if (disk.getType() == Volume.Type.ISO) {
+                result = true;
```
nvazquez: Is it ok to assume this in every case?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,142 @@
...
+        if (super.internalCanHandle(volumeMap) == StrategyPriority.CANT_HANDLE) {
+            Set<VolumeInfo> volumeInfoSet = volumeMap.keySet();
+
+            for (VolumeInfo volumeInfo : volumeInfoSet) {
+                StoragePoolVO storagePoolVO = _storagePoolDao.findById(volumeInfo.getPoolId());
+                if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem) {
+                    return StrategyPriority.CANT_HANDLE;
+                }
+            }
+            return StrategyPriority.HIGHEST;
```
DaanHoogland: if this is KVM specific, the priority shgould not be more than StrategyPriority.HYPERVISOR, should it?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
DaanHoogland: isn't this the old behaviour anyway?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: Unfortunately, this implementation is not working in such an example. The user would need to unplug the attached volume first and then migrate. I will check alternatives and ping in case I find a better way. If I am not able to solve it, I will make sure to document it and send the proper log/message for the user.

wido: Yes, if it doesn't work we should throw a proper error message saying that we can't migrate the Instance as long as data disks are attached.

Check if numDataVolumes > 0, then throw an Exception.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
GabrielBrascher: @wido the `libvirtComputingResource.cleanupDisk(disk)` was already being executed. The cleanupDisk just disconnects from the source host, after it has been migrated to another host; it works that way with iScsi, for instance. I can also work on renaming/documenting the cleanupDisk method, if you think that it would good.

The `deleteLocalVolume` method was added to handle in case of the disk is on a file; in such case it deletes the file from the source host (local primary storage).

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
GabrielBrascher: The plan was to extract it to a method so I could change in the NonManaged to not throw an exception, but to be honest, I could just remove the exception as if it is not a managed storage it should not execute it anyway.

core/src/main/java/com/cloud/agent/api/MigrateCommand.java
```diff
@@ -151,5 +163,13 @@ public Source getSource() {
         public String getSourceText() {
             return sourceText;
         }
+
+        public boolean isSourceDiskOnStorageFileSystem() {
```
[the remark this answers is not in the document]
DaanHoogland: old gson does, but when we try to upgrade, we'll have issues, so let's start to use wrapper classes

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: Solved by using `File(libvirtDestImgsPath, destVolumeInfo.getUuid()).getAbsolutePath()`. It works either way, with the libvirtDestImgsPath containing or not the File.separator at the end (`/var/lib/libvirt/images` or `/var/lib/libvirt/images/`).

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStoragePoolManager.java
GabrielBrascher: Yes @nvazquez, we can always assume result as true in case of an ISO; as the ISO connection does not have any error, because it is not executed. Then we continue the iteration over the VM volumes List.

If an issue happens while connecting a Disk, the result will be false as you can check at lines 157 - 161.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
GabrielBrascher: Removed that method @nvazquez. It does not make sense to have that `throwExeptionIfCannotLiveMigrateOnANonManagedStorage`. Thanks for pointing that!

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
```diff
@@ -203,8 +207,13 @@ Use VIR_DOMAIN_XML_SECURE (value = 1) prior to v1.0.0.
             destDomain = migrateThread.get(10, TimeUnit.SECONDS);
 
             if (destDomain != null) {
-                for (final DiskDef disk : disks) {
-                    libvirtComputingResource.cleanupDisk(disk);
+                for (DiskDef disk : disks) {
+                    MigrateDiskInfo migrateDiskInfo = searchDiskDefOnMigrateDiskInfoList(migrateDiskInfoList, disk);
```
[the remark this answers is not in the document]
GabrielBrascher: Extracted to a method, documented it, and created a JUnit test.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
```diff
@@ -203,9 +207,7 @@ Use VIR_DOMAIN_XML_SECURE (value = 1) prior to v1.0.0.
             destDomain = migrateThread.get(10, TimeUnit.SECONDS);
 
             if (destDomain != null) {
-                for (final DiskDef disk : disks) {
-                    libvirtComputingResource.cleanupDisk(disk);
-                }
+                deleteOrDisconnectDisksOnSourcePool(libvirtComputingResource, migrateDiskInfoList, disks);
```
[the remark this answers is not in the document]
GabrielBrascher: I am trying to be generic and treating it as a storage pool. However, I can change the name and/or documentation if necessary.

nvazquez:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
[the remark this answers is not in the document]
rafaelweingartner: No need. I understand your point (after the talk with had off Github).

Thanks for the PR.

nvazquez:
@blueorangutan package

GabrielBrascher:
Thanks for running blueorangutan, @nvazquez! Is there any issue with the Debian packaging?

nvazquez:
Yes but not related to the PR, let me kick tests for centos7 and KVM
@blueorangutan test

nvazquez:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

nvazquez:
@blueorangutan test

GabrielBrascher:
@blueorangutan package

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: @wido @rafaelweingartner updated the code and tested. Now it migrates VMs with attached data disk on shared storage.
The added code can be found in `StorageSystemDataMotionStrategy.copyAsync(Map, VirtualMachineTO, Host, Host, AsyncCompletionCallback)`:
```
                if (!shouldMigrateVolume(sourceStoragePool, destHost, destStoragePool)) {
                    continue;
                }
```

rafaelweingartner: Awesome. Thanks!

nvazquez:
@blueorangutan test

GabrielBrascher:
@mike-tutkowski, @rhtyd, @DaanHoogland, @nvazquez, and **others**
Are you ok with this PR?
@mike-tutkowski can you please check if I did not break any piece of Managed Storage implementation?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+        if (super.internalCanHandle(volumeMap) == StrategyPriority.CANT_HANDLE) {
+            Set<VolumeInfo> volumeInfoSet = volumeMap.keySet();
+
+            for (VolumeInfo volumeInfo : volumeInfoSet) {
+                StoragePoolVO storagePoolVO = _storagePoolDao.findById(volumeInfo.getPoolId());
+                if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem && storagePoolVO.getPoolType() != StoragePoolType.NetworkFilesystem) {
+                    return StrategyPriority.CANT_HANDLE;
+                }
+            }
+            return StrategyPriority.HYPERVISOR;
```
mike-tutkowski: Just a quick question here: The comment says we return the HIGHEST priority here, but I see we are returning the HYPERVISOR priority here. Can you clarify which one is true? Thanks!

GabrielBrascher: My mistake here, now the documentation is updated and set as HYPERVISOR.
Thanks for pointing that, @mike-tutkowski!

rhtyd:
@blueorangutan package

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -131,21 +132,23 @@
 
 @Component
 public class StorageSystemDataMotionStrategy implements DataMotionStrategy {
-    private static final Logger LOGGER = Logger.getLogger(StorageSystemDataMotionStrategy.class);
+    protected static final Logger LOGGER = Logger.getLogger(StorageSystemDataMotionStrategy.class);
```
rhtyd: Why make this protected?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
```diff
@@ -51,19 +76,19 @@ public MigrateKVMAsync(final LibvirtComputingResource libvirtComputingResource,
 
     @Override
     public Domain call() throws LibvirtException {
-        long flags = 1 << 0;
+        long flags = VIR_MIGRATE_LIVE;
 
         // set compression flag for migration, if libvirt version supports it
         if (dconn.getLibVirVersion() >= 1000003) {
-            flags |= 1 << 11;
```
rhtyd: Do you also want to make the version a final Integer?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
```diff
@@ -51,19 +76,19 @@ public MigrateKVMAsync(final LibvirtComputingResource libvirtComputingResource,
...
+            flags += VIR_MIGRATE_COMPRESSED;
```
rhtyd: Understanding how flags work, the or was a more readable syntax for me than the +=.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
GabrielBrascher: @rhtyd I am using this logger on the `KvmNonManagedStorageDataMotionStrategy` class, thus have set as protected.
However, I shall fix it for `protected Logger logger = Logger.getLogger(getClass());`

Thanks for pointing that!

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
GabrielBrascher: Agree, will do that. Thanks!

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test matrix

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
GabrielBrascher: My mistake here, @rhtyd. The class that extends this one does not need to log for now; thus, I kept this as private as it was before.

GabrielBrascher:
@rhtyd thanks for running the integration tests! I have addressed your comments.

rhtyd:
Great, let me review and test this tomorrow and then merge it.

nvazquez:
@rhtyd are we ok to merge it?

rhtyd:
Sorry for the delay, let me get back by tomorrow on this.

GabrielBrascher:
@rohit reviews were done, tests passed and all checks are green. I will wait for your review until tomorrow, afterwards I will merge it if I get no objections.

rhtyd:
Hi @GabrielBrascher I'm on holidays so couldn't review sooner. I'll back at work tomorrow, will try to review otherwise go ahead. I'll kick a test round as I see a new commit since last test run. 
@blueorangutan package

GabrielBrascher:
Thanks for the update @rhtyd!

rhtyd:
@GabrielBrascher can you look at Travis failures
@blueorangutan test

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+import com.cloud.storage.Storage.StoragePoolType;
+import com.cloud.storage.VolumeVO;
+import com.cloud.utils.exception.CloudRuntimeException;
+import com.cloud.vm.DiskProfile;
+
+/**
+ * Extends {@link StorageSystemDataMotionStrategy}, allowing KVM hosts to migrate VMs with the ROOT volume on a non managed local storage pool.
+ * As {@link StorageSystemDataMotionStrategy} is considering KVM, this implementation also migrates only from/to KVM hosts.
+ */
+@Component
```
rhtyd: `@Component` can be removed as the bean `KvmNonManagedStorageDataMotionStrategy` is defined.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+    @Inject
+    private TemplateDataFactory templateDataFactory;
+
+    /**
+     * Uses the canHandle from the Super class {@link StorageSystemDataMotionStrategy}. If the storage pool is of file and the internalCanHandle from {@link StorageSystemDataMotionStrategy} CANT_HANDLE, returns the StrategyPriority.HYPERVISOR strategy priority. otherwise returns CANT_HANDLE.
+     * Note that the super implementation (override) is called by {@link #canHandle(Map, Host, Host)} which ensures that {@link #internalCanHandle(Map)} will be executed only if the source host is KVM.
+     */
+    @Override
+    protected StrategyPriority internalCanHandle(Map<VolumeInfo, DataStore> volumeMap) {
+        if (super.internalCanHandle(volumeMap) == StrategyPriority.CANT_HANDLE) {
```
rhtyd: Simply if-else. Check if conditional mismatch return `CANT_HANDLE`, rest of code don't need to be in `if{}`

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+        TemplateInfo templateImage = templateDataFactory.getTemplate(templateId, DataStoreRole.Image);
+        return templateImage.getUuid();
+    }
+
+    /**
+     * Sets the volume path as the volume UUID.
+     */
+    @Override
+    protected void setVolumePath(VolumeVO volume) {
+        volume.setPath(volume.getUuid());
```
rhtyd: perform null check here?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -0,0 +1,145 @@
...
+import com.cloud.storage.VolumeVO;
+import com.cloud.utils.exception.CloudRuntimeException;
+import com.cloud.vm.DiskProfile;
+
+/**
+ * Extends {@link StorageSystemDataMotionStrategy}, allowing KVM hosts to migrate VMs with the ROOT volume on a non managed local storage pool.
+ * As {@link StorageSystemDataMotionStrategy} is considering KVM, this implementation also migrates only from/to KVM hosts.
+ */
+@Component
+public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMotionStrategy {
```
rhtyd: Should this be called `KVMNonManagedLocalStorage...` as `shouldMigrateVolume` method only allows when source volume  is filesystem type.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
rhtyd: For example, write as ``flags |= constant`

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
rhtyd: Write as `flags |= VIR_MIGRATE_NON_SHARED_DISK`. In general, enabling is flag is more readable as performing an `OR, |` operation than an addition operation.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
```diff
@@ -51,19 +81,18 @@ public MigrateKVMAsync(final LibvirtComputingResource libvirtComputingResource,
...
         }
 
         if (migrateStorage) {
-            flags |= 1 << 6;
+            flags += VIR_MIGRATE_NON_SHARED_DISK;
         }
 
         if (autoConvergence && dconn.getLibVirVersion() >= 1002003) {
-            flags |= 1 << 13;
+            flags += VIR_MIGRATE_AUTO_CONVERGE;
```
rhtyd: same as above^^

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
```diff
@@ -338,7 +382,7 @@ private String replaceStorage(String xmlDesc, Map<String, MigrateCommand.Migrate
                         String path = getPathFromSourceText(migrateStorage.keySet(), sourceText);
 
                         if (path != null) {
-                            MigrateCommand.MigrateDiskInfo migrateDiskInfo = migrateStorage.remove(path);
+                            MigrateCommand.MigrateDiskInfo migrateDiskInfo = migrateStorage.get(path);
```
rhtyd: Was this a bug, to get and not remove a path. Can you send any bugfix PR towards 4.11?

rhtyd:
Test run lgtm.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
GabrielBrascher: This `remove` affects file migration (local file or NFS); however, before this, there was no KVM storage migration considering file path. Therefore, it does not impact on previous releases.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: Thanks, @rhtyd. That was my first idea as well.
```
if (super.internalCanHandle(volumeMap) == StrategyPriority.HIGHEST) {
} else {
    Set<VolumeInfo> volumeInfoSet = volumeMap.keySet();

    for (VolumeInfo volumeInfo : volumeInfoSet) {
        StoragePoolVO storagePoolVO = _storagePoolDao.findById(volumeInfo.getPoolId());
        if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem && storagePoolVO.getPoolType() != StoragePoolType.NetworkFilesystem) {
        }
    }
    return StrategyPriority.HYPERVISOR;
}
```
However, this approach would ignore cases where the `super.internalCanHandle(volumeMap)` returns the following `StrategyPriority`: `DEFAULT`, `HYPERVISOR`, or `PLUGIN`. Thus, I decided to be safe and use the `super.internalCanHandle(volumeMap) == StrategyPriority.CANT_HANDLE` conditional.

One second approach could be considering all non desired alternatives with an `OR`. However, it makes the code ugly and yet error prone, in case of a new StrategyPriority is adopted.

```
if (super.internalCanHandle(volumeMap) == StrategyPriority.HIGHEST || 
    super.internalCanHandle(volumeMap) == StrategyPriority.DEFAULT ||
    super.internalCanHandle(volumeMap) == StrategyPriority.PLUGIN) {
} else {
    ...
}
```

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: Good catch, @rhtyd! Thanks!
Fixed, and updated it. Indeed the bean is initialized without `@Component`, a leftover from the beginning of the implementation.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: The idea of this class is to add a Strategy that separates the managed and non-managed KVM data motion. For now, it handles local filesystem, but it can be used in the future also for NFS (with the @nvazquez PR) for instance. In the future, we also can start separating into non-managed Files strategy (NFS, or local file), non-managed Object Storage, non-managed Block Storage, etc.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
```diff
@@ -36,6 +36,36 @@
     private boolean migrateStorage;
     private boolean autoConvergence;
 
+    /**
+     * Do not pause the domain during migration. The domain's memory will be transferred to the destination host while the domain is running. The migration may never converge if the domain is changing its memory faster then it can be transferred. The domain can be manually paused anytime during migration using virDomainSuspend.
+     * @value 1
+     * @see Libvirt <a href="https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainMigrateFlags">virDomainMigrateFlags</a> documentation
+     */
+    private static final long VIR_MIGRATE_LIVE = 1l;
```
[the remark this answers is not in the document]
GabrielBrascher: The variable (`VIR_MIGRATE_LIVE`) is a primitive long; the JVM would make the Long into a primitive long via the auto unboxing process. I don't think that making it Long would make much a difference.
Are you talking about this recomendation?
> The suffix L is preferred, because the letter l (ell) is often hard to distinguish from the digit 1 (one).

I don't think that it would improve much the readability of the code here.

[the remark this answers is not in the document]
rhtyd: @GabrielBrascher to people with weak eyes and misleading fonts configured `l` might look like a `1`. Please use `L`, even the jdk8 docs advise to use `L` than `l`.

GabrielBrascher:
Thanks for the feedback, @rhtyd, @DaanHoogland, @nvazquez, @wido, @rafaelweingartner, @mike-tutkowski! Trillian test is good, and we have 4 LGTMs. This one looks ready to merge.

The Travis failures are not related to this PR. I have been digging on what is causing Travis to fail on recent PRs.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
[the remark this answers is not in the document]
GabrielBrascher: Will update it soon, thanks!

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: This method is used to set a VolumeVO that was created using the method `StorageSystemDataMotionStrategy.duplicateVolumeOnAnotherStorage(Volume, StoragePoolVO)` on `StorageSystemDataMotionStrategy.copyAsync` (line 1727 on `StorageSystemDataMotionStrategy`), which writes on the database a VolumeVO to be then used on the `setVolumePath(VolumeVO volume)`. Thus, this method will not be called with a null VolumeVO.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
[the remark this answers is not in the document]
GabrielBrascher: Done!

rafaelweingartner:
Integration tests passed, reviews were made and adjusts executed. Travis problems are not related to this PR. Therefore, I am going to merge it.

rhtyd:
Few of my comments were not addressed or replied to, but I'll live with the changes. @GabrielBrascher @rafaelweingartner

rafaelweingartner:
@rhtyd which ones?

rhtyd:
@rafaelweingartner go to the files changed tab, there are a few minor comments but nothing I cannot live with.

rafaelweingartner:
Well, I indeed checked everything, and everything seems to be already answered/addressed. There was only one comment that you did regarding the flag building that @GabrielBrascher used a more human readable method instead of playing/shifting bits around. That seemed to be covered by the feedback of the community, which was very positive.

Issue #2998, opened 2018-11-04
Support dummy ethernet devices as bridge slaves


bwsw:
##### ISSUE TYPE

 * Enhancement Request

##### COMPONENT NAME

~~~
KVM Agent
~~~

##### CLOUDSTACK VERSION

~~~
4.12
~~~

##### CONFIGURATION

CloudStack with KVM agent

##### SUMMARY

CloudStack currently checks for specific device [names](https://github.com/apache/cloudstack/blob/323d381767c9f5c6edf180b4b9ed3d515e4e3c20/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java#L1294) when network is configured. In the list

```java
    String [] _ifNamePatterns = {
            "^eth",
            "^bond",
            "^vlan",
            "^vx",
            "^em",
            "^ens",
            "^eno",
            "^enp",
            "^team",
            "^enx",
            "^p\\d+p\\d+"
    };
```
Only real interfaces specified, but no interfaces which can be used as "fake" devices. For testing purposes it's beneficial to deploy CS inside a single VM with nested KVM, in such a deployment, no external NICS must be used for bridges, thus some fake devices must be used. 

Those fake devices can be created in the form of dummy interfaces with names supported by CloudStack, like `eth255` but it's a workaround. 

The proposal introduces additional valid device names "^dummy" and "^lo", which can be used in such testing configurations.

##### STEPS TO REPRODUCE

The feature will be used to deploy completely private development clusters with nested KVM inside single VM without real NICS.

~~~

~~~

##### EXPECTED RESULTS

~~~

~~~

##### ACTUAL RESULTS

~~~

~~~

PR #2893, opened 2018-10-10
ApiServer: signature v3 to accept more formats

greut:
## Description

The format expected by `signatureVersion=3&expires=....` is [quite limited](http://docs.cloudstack.apache.org/en/4.11.1.0/developersguide/dev.html?#enabling-api-call-expiration
).

It should accepts the following formats that are containing a timezone and/or milliseconds.

- 2018-10-01T08:12:14Z
- 2018-10-01T08:12:14+01:00
- 2018-10-01T08:12:14+0100
- 2018-10-01T08:12:14.000Z
- 2018-10-01T08:12:14.000+01:00
- 2018-10-01T08:12:14.000+0100

afaik only `2018-10-01T08:12:14+0100` is accepted by the current codebase.

This PR echoes another pull requests I made earlier this year. https://github.com/apache/cloudstack/pull/2392 and https://github.com/apache/cloudstack/pull/2867

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## GitHub Issue/PRs

n/a

## Screenshots (if appropriate):

n/a

## How Has This Been Tested?

the Java unit tests. If anyone want to help with the marvin work... [cs](https://github.com/exoscale/cs/pull/70/files) does it already.

## Checklist:

- [x] I have read the [CONTRIBUTING](https://github.com/apache/cloudstack/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
Testing
- [x] I have added tests to cover my changes.
- [ ] All relevant new and existing integration tests have passed.
- [ ] A full integration testsuite with all test that can run on my environment has passed.

PR #3000, opened 2018-11-04
Added two more device names (lo* and dummy*).

bwsw:
Added dummy and lo devices to be treated as a normal bridge slave devs

## Description

Changed device names list for bridge slave devices which are valid for networking.

https://github.com/apache/cloudstack/issues/2998

## Types of changes

## How Has This Been Tested?

The unit test was improved to handle new device names.


bwsw:
Bug listed in the build
https://travis-ci.org/apache/cloudstack/jobs/450628551#L1047
Doesn't connected with changes and is related to datetime comparisons, happens from time to time. Probably introduced in another PR recently.

bwsw:
```java
    @Test
    public void zonedTimeFormatIsoNoColonZMs() throws ParseException {
        Date time = new Date();
        DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ");
        String str = OffsetDateTime.now().format(formatter);

        Date dtParsed = DateUtil.parseTZDateString(str);

        assertEquals(str, time.toString(), dtParsed.toString());
    }
```

This unittest is obviously buggy. Because there are two times objects are used to get current DT, as a result, sometimes it fails.

bwsw:
Bug introduced here: https://github.com/apache/cloudstack/pull/2893

rhtyd:
Can you check Travis failure @bwsw, also drop the old styled s_ variable name.

bwsw:
@rhtyd I have checked, please take a look at
PR: https://github.com/apache/cloudstack/pull/2893
it's merged and the bug is introduced there.

bwsw:
@rhtyd varname is fixed.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2149,7 +2152,7 @@ So if getMinSpeed() returns null we fall back to getSpeed().
         /* Add a VirtIO channel for SystemVMs for communication and provisioning */
         if (vmTO.getType() != VirtualMachine.Type.User) {
             devices.addDevice(new ChannelDef(vmTO.getName() + ".vport", ChannelDef.ChannelType.UNIX,
-                              new File(_qemuSocketsPath + "/" + vmTO.getName() + ".agent")));
+                    new File(_qemuSocketsPath + "/" + vmTO.getName() + ".agent")));
```
wido: While you are busy with this, why not use File.seperator?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2159,7 +2162,7 @@ So if getMinSpeed() returns null we fall back to getSpeed().
 
         /* Add a VirtIO channel for the Qemu Guest Agent tools */
         devices.addDevice(new ChannelDef(_qemuGuestAgentSocketName, ChannelDef.ChannelType.UNIX,
-                          new File(_qemuSocketsPath + "/" + vmTO.getName() + "." + _qemuGuestAgentSocketName)));
+                new File(_qemuSocketsPath + "/" + vmTO.getName() + "." + _qemuGuestAgentSocketName)));
```
wido: Same goes here

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
bwsw: @wido it's not me. It's IDEA did that occasionally.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
bwsw: @wido same here

bwsw:
@wido I addressed your comments by refactoring the code even further. Changed "+" to Paths.get(...) which does it safely. Please, review.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -1291,7 +1292,7 @@ private String matchPifFileInDirectory(final String bridgeName) {
         return "";
     }
 
-    String [] _ifNamePatterns = {
+    static String [] ifNamePatterns = {
```
rafaelweingartner: @bwsw, what if we invert the logic?

Let me explain myself. We already have 13! (that is too much, but it is indeed a nice magic number) patterns to tell which ones are the valid physical interfaces names that we should look for. However, we can do the opposite. So, we can say that the physical interface is anything that is not “^vnet*”. Then, if I decide to use a physical interface named “XYZ”, I will not need to change anything else in the code.

What do you think?

bwsw: @rafaelweingartner I think it's a good extension, but suppose it's a conceptual change which requires voting from the project team. Don't you think so?
Personally, I like your approach, but people "potentially" may meet incompatibilities in their deployments if they have weird bridging configurations. 

@rhtyd @PaulAngus @DaanHoogland @wido @GabrielBrascher can give your opinions on such a change, that @rafaelweingartner suggests?

rafaelweingartner: Is it possible for people to configure/customize the "vnet" name of the interfaces? I thought that was something hard coded in KVM. That is why I thought doing the opposite would be easier.

bwsw: @rafaelweingartner they can, but not sure that they can with CloudStack. Actually, your idea that only "vnet" devices must be excluded is good but just would like to get the opinion of other persons before changing that. In theory, it can break someone's deployment even if our tests work well.

bwsw: @rafaelweingartner what I personally like is that people will no longer be limited by using "bond" or "team" names for physical interfaces, but can use other semantic names, that they like, the same is for single NICs like "public", "private" instead of eth0, eth1.

rhtyd:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
GabrielBrascher: @bwsw I think that the proposal from @rafaelweingartner  of inverting the logic is valid and gives us flexibility. I do not see any potential issue that such change would bring to us.
+1 on following that path.

Do you guys share the same thoughts @rhtyd @PaulAngus @DaanHoogland @wido @nvazquez?

DaanHoogland: I agree with @rafaelweingartner , @GabrielBrascher , but it would be a separate issue. I am not sure of the consequences so I would leave it out of scope for this PR.

rhtyd:
@bwsw can you fix the conflict, thanks.

bwsw:
@rhtyd Conflicts fixed. Local build and tests passed.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3004, opened 2018-11-05
[Placeholder] Add new data motion strategy for KVM non managed storage live migrations

nvazquez:
## Description
Add a new data motion strategy as a placeholder for new KVM live storage migration features: #2997 and #2983 

## Types of changes

### DIFF
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageSystemDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageSystemDataMotionStrategy.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


nvazquez:
@blueorangutan package

nvazquez:
@DaanHoogland yes, we are looking forward to extend this new class on both PRs: #2997 and #2983

rafaelweingartner:
@nvazquez, I am with @DaanHoogland here. Can you explain further the need for this PR? I mean, this class that you introduce here seems to be something that is already introduced in one of those PRs that you mention. Therefore, if someone else needs it, or if the PRs are somehow dependent, I would say, let’s work together to get the complete PR in, and not just some part of it. Then, after getting the first one, we can work to get the second one.

nvazquez:
Thanks @rafaelweingartner, the aim of this PR was just introducing the new strategy class as a placeholder so we could both extend its functionality on separate PRs. We can also close this PR and create the new file on both PRs and then rebase and resolve the conflicts when one of them gets merged as you mentioned, but thought this way would be easier to track changes and at the same time PRs won't depend on each other, they'll just add logic into existing new strategy.

nvazquez:
Thanks for the input @rafaelweingartner, that was what we were looking for with this PR. Sure I agree with that approach too

Issue #3002, opened 2018-11-05
api: listApis command missing 'tags' field for many responses


csquire:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API
~~~

##### CLOUDSTACK VERSION

~~~
4.9.3.0, 4.11.2.0
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

At some point around 4.9, many API command stopped reporting the `tags` field in their API response fields. For example UserVMResponse and TemplateResponse. I believe this was caused by [this commit](https://github.com/apache/cloudstack/commit/db54b26fed51d55879f2f704bcfbc25e52e15461#diff-9a2b0c645d3b186d9f2fdda9f56b9946). After this change ApiDiscoveryService is not getting the `tags` field from the superclass of the response classes.

The [xanzy go-cloudstack](https://github.com/xanzy/go-cloudstack) client depends on the output of `listApis` to generate it's Cloudstack client SDK.  In later versions of the xanzy cloudstack client, the tags fields are missing from many of the responses. This in turn means that the Terraform Cloudstack Provider (which uses the xanzy client) cannot use a newer version of xanzy go-cloudstack without a breaking change. 

I also believe the current version of xanzy go-cloudstack used by the Terraform Cloudstack Provider  may be incompatible with Cloudstack 4.11 because of a [type change](https://github.com/apache/cloudstack/commit/4534cef#diff-0c5e3890cfb1d9051d383fd2b41d6132) in the Cloudstack API, changing ProjectResponse.secondaryStorageTotal from an int to a float. Following semver, breaking changes like this type change shouldn't be made in minor version updates, but that's a different conversation.

##### STEPS TO REPRODUCE

~~~
Open the following and notice that the listVirtualMachines API does not advertise the `tags` field in the response: 
http://localhost:8080/client/api?command=listApis&response=json&name=listVirtualMachines
~~~

##### EXPECTED RESULTS

~~~
The `tags` field should be present.
~~~

##### ACTUAL RESULTS

The `tags` field is not present.

~~~

~~~

PR #3005, opened 2018-11-05
Discover tags field on superclass of API responses

csquire:
## Description

Updated `ApiServiceDiscoveryImpl` to check superclasses of API responses for fields.

Fixes: #3002 
## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

Additional unit test and tested locally to verify the tags field was present in the output of listApis response for listVirtualMachines API.

### DIFF
--- a/plugins/api/discovery/src/org/apache/cloudstack/discovery/ApiDiscoveryServiceImpl.java
+++ b/plugins/api/discovery/src/org/apache/cloudstack/discovery/ApiDiscoveryServiceImpl.java
@@ -16,21 +16,13 @@
 // under the License.
 package org.apache.cloudstack.discovery;
 
-import java.lang.reflect.Field;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import javax.inject.Inject;
-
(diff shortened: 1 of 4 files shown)


plugins/api/discovery/src/org/apache/cloudstack/discovery/ApiDiscoveryServiceImpl.java
```diff
@@ -109,7 +107,12 @@ public boolean start() {
             }
             response.setRelated(responseName);
 
-            Field[] responseFields = apiCmdAnnotation.responseObject().getDeclaredFields();
+            List<Field> responseFields = new ArrayList<>(Arrays.asList(responseClass.getDeclaredFields()));
```
rafaelweingartner: Instead of doing this whole loop thing, what about using `org.reflections.ReflectionUtils.getAllFields(Class<?>)`? It will produce the same result with much less code.

csquire: Sure, done

rhtyd:
@blueorangutan test

rhtyd:
Lgtm, merged.

PR #3013, opened 2018-11-08
Print log on INFO if Host reached Max Guests Limit

wido:
This should not be in DEBUG as people would want to know that the host was skipped
because it didn't have enough slots available to run the VM.


rhtyd:
This is only logging change, I'll go with Travis test results and merge this.

PR #3022, opened 2018-11-12
systemvmtemplate: update debian 9.6 iso url and checksum

rhtyd:
This fixes the failing systemvmtemplate build due to 404 on old ISO url.

Signed-off-by: Rohit Yadav <rohit.yadav@shapeblue.com>

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

PR #3023, opened 2018-11-12
Changed SystemVM template to use debian-9.6.0-amd64-netinst.iso

DennisKonrad:
Hi,

this are the changes to use debian 9.6 instead of 9.5. I tested the build but NOT if the
systemvm is fully working with the changes introduced by 9.6. Also i'm not a
debian guy so I don't really know what's changing there from 9.5. to 9.6.

If there are problems with this I would very much appreciate the feedback.

##  Description

Changed build script to use debian 9.6 because 9.5 is not available at the link provided in the script.
## Types of changes

## How Has This Been Tested?
Build new systemvm with it.

### DIFF
--- a/tools/appliance/systemvmtemplate/template.json
+++ b/tools/appliance/systemvmtemplate/template.json
@@ -38,8 +38,8 @@
       "disk_interface": "virtio",
       "net_device": "virtio-net",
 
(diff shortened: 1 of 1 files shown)


rafaelweingartner:
Isn't this one already covered here [1]? 
[1] https://github.com/apache/cloudstack/pull/3022

DennisKonrad:
@rafaelweingartner Oh, I think it is. Thanks for pointing out.  I began to do this when the request wasnt open :)

PR #3024, opened 2018-11-12
travis: fail fast if --with-marvin fails with nose

rhtyd:
This fixes issue with recent Travis runs which gave incorrect results
around smoketests with simulator where each test run failed with an
error like "nosetests: error: no such option: --with-marvin".

## Types of changes

We can merge this is Travis actually runs the tests and passes. See all recent travis test runs, they all silently fail.

### DIFF
--- a/tools/travis/before_install.sh
+++ b/tools/travis/before_install.sh
@@ -98,8 +98,6 @@ echo "<settings>
 
 echo -e "\nInstalling some python packages: "
 
-pip install --user --upgrade pip
-
 for ((i=0;i<$RETRY_COUNT;i++))
 do
   pip install --user --upgrade lxml paramiko nose texttable ipmisim pyopenssl mock flask netaddr pylint pycodestyle six astroid > /tmp/piplog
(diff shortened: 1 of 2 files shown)


DaanHoogland:
@rhtyd I don't see what you are trying to do here. The description doesn't seem related to the changes in code. Is the pip upgrade the cause of the marvin failure?

rhtyd:
Yes @DaanHoogland looks like latest pip does not install marvin well. Distro provided pip v9.0 works okay. There is one failure that I need to look at.

rhtyd:
Travis passes now, issue was a missing dependency and use of upgraded pip. Please review (no other tests needed as changes are only in travis scripts) - @wido @GabrielBrascher @rafaelweingartner and others

tools/travis/before_install.sh
```diff
@@ -98,11 +98,9 @@ echo "<settings>
 
 echo -e "\nInstalling some python packages: "
 
-pip install --user --upgrade pip
```
nvazquez: So we are not upgrading pip, do we need to document which version is expected?

rhtyd: The default one that is v9.0 is okay.

rhtyd:
Thanks all, now on PRs travis can actually run smoke tests. Previously they all fail to run, but travis would still stay green.

Issue #3049, opened 2018-11-22
Account change dynamic role


bwsw:
##### ISSUE TYPE

 * Feature Idea
 
##### COMPONENT NAME

~~~
API, UI
~~~

##### CLOUDSTACK VERSION

~~~
4.12
~~~

##### SUMMARY

Allow account to change dynamic role.

##### STEPS TO REPRODUCE

Administrators may want to create restricted roles which users get first and next to switch them to extended roles. This feature addresses that.

##### EXPECTED RESULTS

The administrator can change a role for an account.

bwsw:
PR is merged. Close it.

PR #3058, opened 2018-11-26
CLOUDSTACK-3049: update dynamic role for an account

bwsw:
## Description

Existing 'updateAccount' API call now can update role as well as account name and network domain. During the implementation 'newname' attribute was marked as optional, previously in the API was annotated as required, which is not a right behavior.

New Dockerfile was added Dockerfile.smokedev which is convenient if you develop smoke tests.

Helper script was added to deploy datacenter and run tests from the same docker image which guarantees a predictable environment.

```bash
docker run -v ~/dev/tmp:/tmp \
      -v ~/IdeaProjects/cloudstack/test/integration/smoke:/root/test/integration/smoke \
      -it \
      --name simulator -p 8080:8080 -p8096:8096 simulator:4.12

# second console 
docker exec -it simulator bash
# deploy datacenter
bash /root/docker_run_tests.sh advanced smoke
# follow the instructions
```

Implements #3049 

## Types of changes

## How Has This Been Tested?

- The code was tested with cloudmonkey
- Smoketest for Marvin was implemented

### DIFF
--- a/server/src/main/java/com/cloud/user/AccountManagerImpl.java
+++ b/server/src/main/java/com/cloud/user/AccountManagerImpl.java
@@ -37,8 +37,10 @@ import javax.crypto.spec.SecretKeySpec;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
+
 import org.apache.cloudstack.acl.ControlledEntity;
 import org.apache.cloudstack.acl.QuerySelector;
+import org.apache.cloudstack.acl.Role;
 import org.apache.cloudstack.acl.RoleType;
 import org.apache.cloudstack.acl.SecurityChecker;
 import org.apache.cloudstack.acl.SecurityChecker.AccessType;
@@ -1019,8 +1021,11 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
     @DB
     @ActionEvents({@ActionEvent(eventType = EventTypes.EVENT_ACCOUNT_CREATE, eventDescription = "creating Account"),
         @ActionEvent(eventType = EventTypes.EVENT_USER_CREATE, eventDescription = "creating User")})
-    public UserAccount createUserAccount(final String userName, final String password, final String firstName, final String lastName, final String email, final String timezone, String accountName,
-            final short accountType, final Long roleId, Long domainId, final String networkDomain, final Map<String, String> details, String accountUUID, final String userUUID,
+    public UserAccount createUserAccount(final String userName, final String password, final String firstName,
+                                         final String lastName, final String email, final String timezone,
+                                         String accountName, final short accountType, final Long roleId, Long domainId,
+                                         final String networkDomain, final Map<String, String> details,
+                                         String accountUUID, final String userUUID,
             final User.Source source) {
 
         if (accountName == null) {
@@ -1155,7 +1160,7 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
         UserVO user = retrieveAndValidateUser(updateUserCmd);
         s_logger.debug("Updating user with Id: " + user.getUuid());
 
-        validateAndUpdatApiAndSecretKeyIfNeeded(updateUserCmd, user);
+        validateAndUpdateApiAndSecretKeyIfNeeded(updateUserCmd, user);
         Account account = retrieveAndValidateAccount(user);
 
         validateAndUpdateFirstNameIfNeeded(updateUserCmd, user);
@@ -1344,7 +1349,7 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
      * <li>If a pair of keys is provided, we validate to see if there is an user already using the provided API key. If there is someone else using, we throw an {@link InvalidParameterValueException} because two users cannot have the same API key.
      * </ul>
      */
-    protected void validateAndUpdatApiAndSecretKeyIfNeeded(UpdateUserCmd updateUserCmd, UserVO user) {
+    protected void validateAndUpdateApiAndSecretKeyIfNeeded(UpdateUserCmd updateUserCmd, UserVO user) {
         String apiKey = updateUserCmd.getApiKey();
         String secretKey = updateUserCmd.getSecretKey();
 
@@ -1687,6 +1692,7 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
     public AccountVO updateAccount(UpdateAccountCmd cmd) {
         Long accountId = cmd.getId();
         Long domainId = cmd.getDomainId();
+        Long roleId = cmd.getRoleId();
         String accountName = cmd.getAccountName();
         String newAccountName = cmd.getNewName();
         String networkDomain = cmd.getNetworkDomain();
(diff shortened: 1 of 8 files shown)


rhtyd:
Can you instead implement a cleaner approach, say create a new API?

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -1714,23 +1720,37 @@ public AccountVO updateAccount(UpdateAccountCmd cmd) {
         // Check if user performing the action is allowed to modify this account
         checkAccess(getCurrentCallingAccount(), _domainMgr.getDomain(account.getDomainId()));
 
-        // check if the given account name is unique in this domain for updating
-        Account duplicateAcccount = _accountDao.findActiveAccount(newAccountName, domainId);
-        if (duplicateAcccount != null && duplicateAcccount.getId() != account.getId()) {
-            throw new InvalidParameterValueException(
-                    "There already exists an account with the name:" + newAccountName + " in the domain:" + domainId + " with existing account id:" + duplicateAcccount.getId());
+        if(newAccountName != null) {
```
rafaelweingartner: You can use `StringUtils.isEmpty` here. Then, you do not need all of these netsting

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

server/src/main/java/com/cloud/user/AccountManagerImpl.java
bwsw: @rafaelweingartner 
Hm, I don't think so or don't get the idea. I would like to distinguish three situations:
1. newAccountName is not specified
2. newAccountName is specified but empty string -> exception 1
3. newAccountName is specified and not empty string -> action or exception

So, don't get how `StringUtils.isEmpty()` helps.

bwsw:
> Can you instead implement a cleaner approach, say create a new API?

@rhtyd I considered both alternatives. My arguments for the current implementation:
1. it adds smoke tests for updateAccount and as you can see before the implementation there wasn't testing for the method at all.
2. updateAccount method was strangely implemented, e.g. one has to pass newname mandatory, this enhancement fixes the behavior as well.
3. It's natural, I agree with Rafael, otherwise, it's true to implement every `update<Entity><Attribute>` separately and it will be hell without scoping.

What do you think?

rhtyd:
It's end of my day, let me get back to you soon. In short, my hesitation stems from the fact that the operation in question has some security and privileges implications. As a general rule such an update param/operation must be privileged and checked. It's better to have this as separate api. Take example of other apis which change ownership, transfer relationships etc all cases which could too be solved by an update api as well.

rhtyd:
Another note that there is a translation implication of accounttype that maybe implemented as well. Due to backward compatibility, we support both account type and roleid for the create api for example.

bwsw:
> It's end of my day, let me get back to you soon. In short, my hesitation stems from the fact that the operation in question has some security and privileges implications. As a general rule such an update param/operation must be privileged and checked. It's better to have this as separate api. Take example of other apis which change ownership, transfer relationships etc all cases which could too be solved by an update api as well.

@rhtyd Well, if we decide that only global admin (not domain admin) can change that, then you are right, it's better to implement as a separate API call.

server/src/main/java/com/cloud/user/AccountManagerImpl.java
bwsw: @rafaelweingartner But how it will behave if `newAccountName` is null which is normal if API is not intending to change that. It is a normal case and whole code block must be skipped.

rafaelweingartner: good catch. I was not paying attention to this detail. Then, I would simply do `newAccountName != null && StringUtils.isBlank(newAccountName)`. 

That is only a suggestion. You can leave the code as is if you think it is better this way.

bwsw: @rafaelweingartner ok. I'll leave as is.

rafaelweingartner:
@bwsw if you use the `authorize` field in the `@Parameter` definition for the new role (the role being changed/updated) I think that then you can achieve what @rhtyd wanted. I mean, then we would be able to allow the use of such parameter only for admins or root admins.

bwsw:
@rhtyd @rafaelweingartner 
Well, currently the implementation allows changing the account details only for admin or domain admin. Admin can set any role while domain admin can set only DomainAdmin, ResourceAdmin, User types. 
I suppose that basically, it's what users need.

Please, approve the basic design and then I write the tests for security assurance. Just don't want to spend the time if the design initially doesn't fit.

rafaelweingartner:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -1687,6 +1692,7 @@ public AccountVO disableAccount(String accountName, Long domainId, Long accountI
     public AccountVO updateAccount(UpdateAccountCmd cmd) {
         Long accountId = cmd.getId();
         Long domainId = cmd.getDomainId();
+        Long roleId = cmd.getRoleId();
         String accountName = cmd.getAccountName();
         String newAccountName = cmd.getNewName();
         String networkDomain = cmd.getNetworkDomain();
```
dhlaluku: @bwsw while we are here, I have just realized that 'cmd.getDetails' might throw an NPE. Could you please add an NPE check for this call?

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -1712,25 +1720,48 @@ public AccountVO updateAccount(UpdateAccountCmd cmd) {
...
+        checkAccess(caller, _domainMgr.getDomain(account.getDomainId()));
 
-        // check if the given account name is unique in this domain for updating
-        Account duplicateAcccount = _accountDao.findActiveAccount(newAccountName, domainId);
-        if (duplicateAcccount != null && duplicateAcccount.getId() != account.getId()) {
-            throw new InvalidParameterValueException(
-                    "There already exists an account with the name:" + newAccountName + " in the domain:" + domainId + " with existing account id:" + duplicateAcccount.getId());
+        if(newAccountName != null) {
+
+            if (newAccountName.isEmpty()) {
```
dhlaluku: @bwsw could you please add a comment here to specify that you are checking that the user did not pass an empty string for an optional parameter. Otherwise I think it is a bit confusing why you check if a not null string is empty. At least I found it confusing until I saw the conversation between you and @rafaelweingartner. Overall LGTM

bwsw:
@GabrielBrascher Hi, Gabriel, you have mentioned this PR as a problematic. It has got two LGTM. Please, comment on what do you want me to improve to be merged?

GabrielBrascher:
@bwsw this one is looking good indeed, I must have dropped this PR on the wrong list.

Just to mention, independently of the "list" that the issue/PR was placed I will be getting in touch and help reviewing, testing, and pinging contributors/reviewers.

Thanks for pinging me, keep the good work :+1:

GabrielBrascher:
@blueorangutan package

server/src/main/java/com/cloud/user/AccountManagerImpl.java
GabrielBrascher: @bwsw can you please address @dhlaluku comment? Thanks!

bwsw:
@GabrielBrascher Sounds great.

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
Thanks for the work guys, this one looks ready to merge (LGTMs + tests are good).

PR #3063, opened 2018-11-27
Remove unused stats.timeout configuration variable

GabrielBrascher:
## Description

While extending and doing some tests with the StatsCollector implementation, I have noticed that the `stats.timeout` configuration is not used in the stats command execution flow. 

Thus, this PR proposes the deletion of the `stats.timeout` configuration variable.

## Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -1902,13 +1902,10 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
     }
 
     @Override
-    public HashMap<String, VolumeStatsEntry> getVolumeStatistics(long clusterId, String poolUuid, StoragePoolType poolType, List<String> volumeLocator, int timeout) {
+    public HashMap<String, VolumeStatsEntry> getVolumeStatistics(long clusterId, String poolUuid, StoragePoolType poolType, List<String> volumeLocator) {
         List<HostVO> neighbors = _resourceMgr.listHostsInClusterByStatus(clusterId, Status.Up);
         for (HostVO neighbor : neighbors) {
             GetVolumeStatsCommand cmd = new GetVolumeStatsCommand(poolType, poolUuid, volumeLocator);
-            if (timeout > 0) {
-                cmd.setWait(timeout/1000);
-            }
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -1902,13 +1902,10 @@ public void removeCustomOfferingDetails(long vmId) {
     }
 
     @Override
-    public HashMap<String, VolumeStatsEntry> getVolumeStatistics(long clusterId, String poolUuid, StoragePoolType poolType, List<String> volumeLocator, int timeout) {
+    public HashMap<String, VolumeStatsEntry> getVolumeStatistics(long clusterId, String poolUuid, StoragePoolType poolType, List<String> volumeLocator) {
         List<HostVO> neighbors = _resourceMgr.listHostsInClusterByStatus(clusterId, Status.Up);
         for (HostVO neighbor : neighbors) {
             GetVolumeStatsCommand cmd = new GetVolumeStatsCommand(poolType, poolUuid, volumeLocator);
-            if (timeout > 0) {
```
rhtyd: Well looks like it indeed is used but you removed it @GabrielBrascher ?

GabrielBrascher: @rhtyd the timeout is used to set the `GetVolumeStatsCommand` `wait` variable. However, I was not able to find any piece of code using the `wait` value, neither on `LibvirtGetVolumeStatsCommandWrapper` and `CitrixGetVolumeStatsCommandWrapper`.

GabrielBrascher:
Thanks for the feedback, @DaanHoogland!
What do you consider a major upgrade? Would that be a "CloudStack 5.0"?

DaanHoogland:
5.0 indeed. A user might have scripts setting the setting. If we ignore it, no harm, if we remove it the script will fail and we break contract.

rhtyd:
@GabrielBrascher can you rebase and fix the conflicts, fix the db upgrade path to be in 4.12->4.13.

GabrielBrascher:
@rhtyd I will do that, thanks!

GabrielBrascher:
Closing this one for now. Once we raise a 5.0 release, this one will be able to get merged into (due to breaking API compatibility).

PR #3076, opened 2018-12-03
security: increase keystore setup/import timeout

rhtyd:
This increases and uses a default 15mins timeout for VR scripts and for
KVM agent increases timeout from 60s to 5mins. The timeout can
specifically occur when keystore does not get enough entropy from CPU
and script gets killed due to timeout. This is a very specific corner
case and generally should not happen on baremetal/prod environment, but
sometimes seen in nested/test environments.

/cc @PaulAngus @DagSonstebo

## Types of changes


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@nvazquez yes they can be made configurable, however it may make it slightly complex on how we propagate the settings in case of KVM agent and in case of systemvm (cpvm/ssvm). I've used a very high number to be on safe side, generally they should complete within 60s.

PR #3078, opened 2018-12-05
Add influxdb to statscollector

GabrielBrascher:
## Description

Add Support for InfluxDB on StatsCollector.

The collector is defined on the URI scheme. The _Global Settings_ configuration with the name `stats.output.uri` is used to set the URI to send StatsCollector statistics to.

Example: `graphite://graphite-hostaddress:port` or `influxdb://influxdb-hostaddress:port/dbname`. Note that the port is optional, if not added the default port for the respective collector (graphite or influxdb) will be used. Additionally, the database name `dbname` is also optional; default db name is `cloudstack`. Note that one must create and configure the database if using influxdb.

- Host stats
For the host metrics it creates an InfluxDB point with the measure `host_stats` and containing the following tags and fields:
Tags:
-- host_id
-- uuid
-- pod_id
-- data_center_id
Fields:
-- total_memory_kbs
-- free_memory_kbs
-- cpu_utilization
-- cpus
-- cpu_sockets
-- network_read_kbs
-- network_write_kbs

- VM stats
It creates an InfluxDB point with the measure `vm_stats` and containing the following tags and fields:
Tags:
-- vm_id
-- uuid
-- pod_id
-- data_center_id
-- host_id
Fields:
-- total_memory_kbs
-- free_memory_kbs
-- memory_target_kbs
-- cpu_utilization
-- cpus
-- network_read_kbs
-- network_write_kbs
-- disk_read_kbs
-- disk_read_iops
-- disk_write_kbs
-- disk_write_iops

**Note:** As the current stats collector supports only KVM, this extension supports only KVM. I am planning on enhance and refactor the StatsCollector architecture on the future. Due to the inner class structure, test case for some methods will not be implemented. On the future it will be necessary to refactor the whole StatsCOllector architecture and extract inner classes.

## Types of changes

## How Has This Been Tested?

Apart from the JUnit tests, the following tests have been manually done, asserting the expected behavior.
Test 1: correct scheme and address, database "cloudstack" exists, default port 8086 is open and listening  (influxdb://172.16.0.10, influxdb://172.16.0.10/cloudstack, or influxdb://172.16.0.10:8086/cloudstack)
- Result: CloudStack starts correctly, data are stored in the database cloudstack (the "cloudstack" is already configured on the influxdb before testing) and no error/warning logs.

Test 2: `stats.output.uri` has wrong URI scheme (http://172.16.0.10 or influxd://172.16.0.10)
 - Result: Cannot store data on the InfluxDB host, logs the following Errors
```
ERROR  [c.c.s.StatsCollector] (main:null) (logid:) influxd is not a valid protocol for external statistics. No statistics will be send.
(...)
ERROR [c.c.s.StatsCollector] (main:null) (logid:) Failed to parse external statistics URI: Cannot define a port for the Stats Collector service. The configured URI in stats.output.uri is not supported. Please configure as the following examples: graphite://graphite-hostaddress:port, or influxdb://influxdb-hostaddress:port. Note that the port is optional, if not added the default port for the respective collector (graphite or influxdb) will be used.: influxd://172.16.0.10
```

Test 3: correct URI scheme but wrong address (influxdb://172.16.0.11)
- Result: Cannot store data on the InfluxDB host, logs the following error
```
ERROR [c.c.s.StatsCollector] (StatsCollector-6:ctx-bf6d551c) (logid:d5c92b19) Error trying to retrieve host stats
org.influxdb.InfluxDBIOException: java.net.NoRouteToHostException: No route to host (Host unreachable)
```

Test 4: correct URI scheme and address, database does not exist (influxdb://172.16.0.10/nonexistingdb)
- Result: Cannot store data on the InfluxDB host, logs the following Errors
```
ERROR [c.c.s.StatsCollector] (StatsCollector-1:ctx-5921793c) (logid:d4bab167) Error trying to retrieve host stats
com.cloud.utils.exception.CloudRuntimeException: Database with name nonexistingdb does not exist in influxdb host 172.16.0.10:8086

```
Test 5: correct URI scheme and address, database exists, customized port is not correctily configured (influxdb://172.16.0.10:8888/)
- Result: Cannot store data on the InfluxDB host, logs the following Errors
```
ERROR [c.c.s.StatsCollector] (StatsCollector-1:ctx-b2371dde) (logid:209a4b10) Error trying to retrieve host stats
org.influxdb.InfluxDBIOException: java.net.ConnectException: Failed to connect to /172.16.0.10:8888

```
Test 6: set stats.output.uri to null/empty
- Result: it does not write in the external stats collector and does not log exceptions, as expected.

### DIFF
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -158,10 +158,10 @@
             <artifactId>opensaml</artifactId>
         </dependency>
         <dependency>
-			<groupId>org.influxdb</groupId>
-			<artifactId>influxdb-java</artifactId>
-			<version>2.8</version>
-		</dependency>
+            <groupId>org.influxdb</groupId>
+            <artifactId>influxdb-java</artifactId>
+            <version>2.8</version>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
--- a/server/pom.xml
+++ b/server/pom.xml
@@ -157,6 +157,11 @@
             <groupId>org.opensaml</groupId>
             <artifactId>opensaml</artifactId>
         </dependency>
+        <dependency>
+			<groupId>org.influxdb</groupId>
+			<artifactId>influxdb-java</artifactId>
+			<version>2.8</version>
+		</dependency>
     </dependencies>
     <build>
         <plugins>
--- a/server/src/main/java/com/cloud/server/StatsCollector.java
+++ b/server/src/main/java/com/cloud/server/StatsCollector.java
@@ -47,6 +47,8 @@ import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
 import org.apache.cloudstack.utils.graphite.GraphiteClient;
 import org.apache.cloudstack.utils.graphite.GraphiteException;
 import org.apache.cloudstack.utils.usage.UsageUtils;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 import org.influxdb.InfluxDB;
@@ -289,7 +291,6 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
     long volumeStatsInterval = -1L;
     long autoScaleStatsInterval = -1L;
 
-    List<Long> hostIds = null;
     private double _imageStoreCapacityThreshold = 0.90;
 
     String externalStatsPrefix = "";
@@ -348,11 +349,11 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
                     s_logger.error(externalStatsScheme + " is not a valid protocol for external statistics. No statistics will be send.");
                 }
 
-                if (!StringUtils.isEmpty(uri.getHost())) {
+                if (StringUtils.isNotEmpty(uri.getHost())) {
                     externalStatsHost = uri.getHost();
                 }
 
-                externalStatsPort = configureExternalStatsPort(uri);
+                externalStatsPort = retrieveExternalStatsPortFromUri(uri);
 
                 databaseName = configureDatabaseName(uri);
 
@@ -478,7 +479,7 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
      * Default values are 8086 for influx DB and 2003 for GraphiteDB.
      * Throws URISyntaxException in case of non configured port and external StatsType
      */
-    protected int configureExternalStatsPort(URI uri) throws URISyntaxException {
+    protected int retrieveExternalStatsPortFromUri(URI uri) throws URISyntaxException {
         int port = uri.getPort();
         if (externalStatsType != ExternalStatsProtocol.NONE) {
             if (port != UNDEFINED_PORT_VALUE) {
@@ -502,12 +503,8 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
             try {
                 s_logger.debug("HostStatsCollector is running...");
 
-                SearchCriteria<HostVO> sc = _hostDao.createSearchCriteria();
-                sc.addAnd("status", SearchCriteria.Op.EQ, Status.Up.toString());
-                sc.addAnd("resourceState", SearchCriteria.Op.NIN, ResourceState.Maintenance, ResourceState.PrepareForMaintenance, ResourceState.ErrorInMaintenance);
-                sc.addAnd("type", SearchCriteria.Op.EQ, Host.Type.Routing.toString());
+                SearchCriteria<HostVO> sc = createSearchCriteriaForHostTypeRoutingStateUpAndNotInMaintenance();
 
-                ConcurrentHashMap<Long, HostStats> hostStats = new ConcurrentHashMap<Long, HostStats>();
                 Map<Object, Object> metrics = new HashMap<>();
                 List<HostVO> hosts = _hostDao.search(sc, null);
 
@@ -518,7 +515,7 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
                         metrics.put(hostStatsEntry.getHostId(), hostStatsEntry);
                         _hostStats.put(host.getId(), hostStatsEntry);
                     } else {
-                        s_logger.warn("Received invalid host stats for host: " + host.getId());
+                        s_logger.warn("The Host stats is null for host: " + host.getId());
                     }
                 }
 
@@ -536,25 +533,22 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
          * Updates GPU details on hosts supporting GPU.
          */
         private void updateGpuEnabledHostsDetails(List<HostVO> hosts) {
-            // Get a subset of hosts with GPU support from the list of "hosts"
             List<HostVO> gpuEnabledHosts = new ArrayList<HostVO>();
-            if (hostIds != null) {
-                for (HostVO host : hosts) {
-                    if (hostIds.contains(host.getId())) {
-                        gpuEnabledHosts.add(host);
-                    }
+            List<Long> hostIds = _hostGpuGroupsDao.listHostIds();
+            if (CollectionUtils.isEmpty(hostIds)) {
+                return;
+            }
+            for (HostVO host : hosts) {
+                if (hostIds.contains(host.getId())) {
+                    gpuEnabledHosts.add(host);
                 }
-            } else {
-                // Check for all the hosts managed by CloudStack.
-                gpuEnabledHosts = hosts;
             }
             for (HostVO host : gpuEnabledHosts) {
                 HashMap<String, HashMap<String, VgpuTypesInfo>> groupDetails = _resourceMgr.getGPUStatistics(host);
-                if (groupDetails != null) {
+                if (MapUtils.isEmpty(groupDetails)) {
                     _resourceMgr.updateGPUDetails(host.getId(), groupDetails);
                 }
             }
-            hostIds = _hostGpuGroupsDao.listHostIds();
         }
 
         @Override
@@ -569,10 +563,7 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
             try {
                 s_logger.trace("VmStatsCollector is running...");
 
-                SearchCriteria<HostVO> sc = _hostDao.createSearchCriteria();
-                sc.addAnd("status", SearchCriteria.Op.EQ, Status.Up.toString());
-                sc.addAnd("resourceState", SearchCriteria.Op.NIN, ResourceState.Maintenance, ResourceState.PrepareForMaintenance, ResourceState.ErrorInMaintenance);
-                sc.addAnd("type", SearchCriteria.Op.EQ, Host.Type.Routing.toString());
+                SearchCriteria<HostVO> sc = createSearchCriteriaForHostTypeRoutingStateUpAndNotInMaintenance();
                 List<HostVO> hosts = _hostDao.search(sc, null);
 
                 Map<Object, Object> metrics = new HashMap<>();
@@ -701,10 +692,7 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
                     public void doInTransactionWithoutResult(TransactionStatus status) {
                         s_logger.debug("VmDiskStatsTask is running...");
 
-                        SearchCriteria<HostVO> sc = _hostDao.createSearchCriteria();
-                        sc.addAnd("status", SearchCriteria.Op.EQ, Status.Up.toString());
-                        sc.addAnd("resourceState", SearchCriteria.Op.NIN, ResourceState.Maintenance, ResourceState.PrepareForMaintenance, ResourceState.ErrorInMaintenance);
-                        sc.addAnd("type", SearchCriteria.Op.EQ, Host.Type.Routing.toString());
+                        SearchCriteria<HostVO> sc = createSearchCriteriaForHostTypeRoutingStateUpAndNotInMaintenance();
                         sc.addAnd("hypervisorType", SearchCriteria.Op.EQ, HypervisorType.KVM); // support KVM only util 2013.06.25
                         List<HostVO> hosts = _hostDao.search(sc, null);
 
@@ -748,10 +736,7 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
                                         continue;
                                     }
 
-                                    if (previousVmDiskStats != null && ((previousVmDiskStats.getCurrentBytesRead() != vmDiskStat_lock.getCurrentBytesRead())
-                                            || (previousVmDiskStats.getCurrentBytesWrite() != vmDiskStat_lock.getCurrentBytesWrite())
-                                            || (previousVmDiskStats.getCurrentIORead() != vmDiskStat_lock.getCurrentIORead())
-                                            || (previousVmDiskStats.getCurrentIOWrite() != vmDiskStat_lock.getCurrentIOWrite()))) {
+                                    if (isCurrentVmDiskStatsDifferentFromPrevious(previousVmDiskStats, vmDiskStat_lock)) {
                                         s_logger.debug("vm disk stats changed from the time GetVmDiskStatsCommand was sent. " + "Ignoring current answer. Host: " + host.getName()
                                                 + " . VM: " + vmDiskStat.getVmName() + " Read(Bytes): " + vmDiskStat.getBytesRead() + " write(Bytes): " + vmDiskStat.getBytesWrite()
                                                 + " Read(IO): " + vmDiskStat.getIORead() + " write(IO): " + vmDiskStat.getIOWrite());
@@ -832,10 +817,7 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
                     public void doInTransactionWithoutResult(TransactionStatus status) {
                         s_logger.debug("VmNetworkStatsTask is running...");
 
-                        SearchCriteria<HostVO> sc = _hostDao.createSearchCriteria();
-                        sc.addAnd("status", SearchCriteria.Op.EQ, Status.Up.toString());
-                        sc.addAnd("resourceState", SearchCriteria.Op.NIN, ResourceState.Maintenance, ResourceState.PrepareForMaintenance, ResourceState.ErrorInMaintenance);
-                        sc.addAnd("type", SearchCriteria.Op.EQ, Host.Type.Routing.toString());
+                        SearchCriteria<HostVO> sc = createSearchCriteriaForHostTypeRoutingStateUpAndNotInMaintenance();
                         List<HostVO> hosts = _hostDao.search(sc, null);
 
                         for (HostVO host : hosts) {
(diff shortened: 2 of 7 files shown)


bwsw:
@GabrielBrascher Hi, Gabriel, what is the rationale to integrate it into CloudStack, why don't use EventBus to publish Metrics Events. When I start thinking about how to test that I see it's pretty difficult because a 3rd-party system is involved. 

We use
https://github.com/bwsw/cs-pulse-sensor-kafka
to publish events to Kafka Event Bus (and it would be great if CS could do it by itself).

And we use
https://github.com/bwsw/python-kafka-influxdb-importer
To export to InfluxDB or another system if you have proper exporter.

I mean, Look InfuxDB Open Source Edition doesn't provide HA-features, you need to deploy commercial clustered solution to support clustered and now CS publishes in that non-HA system which is not reliable.

Why not implement all those stuff through Even Bus and establish events protocol?

GabrielBrascher:
Hi @bwsw,
Thanks, I appreciate your suggestions. To avoid misunderstandings, I would like to start by saying that this implementation focus on adding support to InfluxDB (**as a driver**) in the StatsCollector that we already have in ACS. Your proposal to add a driver in StatsColletor to send data to the EventBus is for sure interesting, but it is out of scope here.

Now, addressing your points:
This PR enables ACS to send StatsCollector's data to InfluxDB (we are just adding a driver to support the integration of InfluxDB and ACS). Additionally, as you can see, StatsCollector works fine but needs a huge refactoring, extracting code, adding tests, documenting, and so on; this implementation is a starting point for all of that. Even in such a "simple" InfluxDB implementation, it took hundreds of lines to fit in the StatsCollector codebase right now.

Regarding the 3rd party issue; as you mentioned, we are a small community and cannot afford to waste time, resources and have pieces of code that nobody will know how to test, use, and extend. That is indeed why I do not see any problem on using the InfluxDB 3rd-party component; with a few well-documented lines, it allowed me to create an InfluxDB connection and write the stats (again, I am only introducing a driver here to enable ACS to communicate/send data to InfluxDB).
Regarding the high-availability, not all use cases require such feature. I do see the value of your suggestion and agree that it is an interesting feature to add to CloudStack; However, I do have a problem with the idea of forcing all users to prepare a redundant environment to support such HA (multiple nodes running InfluxDB + the Kafka environment). Both approaches can coexist and I encourage you to create a PR with such feature, I will gladly help to review it.

Moreover, are you aware of https://github.com/influxdata/influxdb-relay? It might be interesting to check it. Another option (later!) is to implement a Telegraph driver; it is a stats proxy which can send stats to InfluxDB, Graphite, ElasticSearch, etc. It would work with Kafka as well.

server/pom.xml
```diff
@@ -157,6 +157,11 @@
             <groupId>org.opensaml</groupId>
             <artifactId>opensaml</artifactId>
         </dependency>
+        <dependency>
```
rafaelweingartner: can you fix the formatting here?

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -288,24 +336,25 @@ private void init(Map<String, String> configs) {
...
 
                 try {
-                    externalStatsType = ExternalStatsProtocol.valueOf(scheme.toUpperCase());
+                    externalStatsType = ExternalStatsProtocol.valueOf(externalStatsScheme.toUpperCase());
                 } catch (IllegalArgumentException e) {
-                    s_logger.info(scheme + " is not a valid protocol for external statistics. No statistics will be send.");
+                    s_logger.error(externalStatsScheme + " is not a valid protocol for external statistics. No statistics will be send.");
                 }
 
                 if (!StringUtils.isEmpty(uri.getHost())) {
```
rafaelweingartner: what about using `StringUtils.isNotEmpty` instead?
Also, if it is empty, should not you throw an exception then?

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -411,7 +461,42 @@ private void init(Map<String, String> configs) {
...
+            return dbname;
+        }
+    }
+
+    /**
+     * Configures the port to be used when connecting with the stats collector service.
+     * Default values are 8086 for influx DB and 2003 for GraphiteDB.
+     * Throws URISyntaxException in case of non configured port and external StatsType
+     */
+    protected int configureExternalStatsPort(URI uri) throws URISyntaxException {
```
rafaelweingartner: Are you configuring a port? Or, are you retrieving a port? I mean, when you configure, I would expect a set something somewhere. Here, I see you checking a few things, returning some data according to some conditions, and using some defaults.

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -420,55 +505,65 @@ protected void runInContext() {
...
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.LocalSecondaryStorage.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.TrafficMonitor.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.SecondaryStorageVM.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.ExternalFirewall.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.ExternalLoadBalancer.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.NetScalerControlCenter.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.L2Networking.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.BaremetalDhcp.toString());
-                sc.addAnd("type", SearchCriteria.Op.NEQ, Host.Type.BaremetalPxe.toString());
+                sc.addAnd("type", SearchCriteria.Op.EQ, Host.Type.Routing.toString());
```
rafaelweingartner: Awesome!!
Instead of defining what you do not want, you tell what you want.

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -420,55 +505,65 @@ protected void runInContext() {
...
-                    HostStatsEntry stats = (HostStatsEntry)_resourceMgr.getHostStatistics(host.getId());
-                    if (stats != null) {
-                        hostStats.put(host.getId(), stats);
+                    HostStatsEntry hostStatsEntry = (HostStatsEntry)_resourceMgr.getHostStatistics(host.getId());
+                    if (hostStatsEntry != null) {
+                        hostStatsEntry.setHostVo(host);
+                        metrics.put(hostStatsEntry.getHostId(), hostStatsEntry);
+                        _hostStats.put(host.getId(), hostStatsEntry);
                     } else {
                         s_logger.warn("Received invalid host stats for host: " + host.getId());
```
rafaelweingartner: Instead of saying `invalid`, can you write in the log entry that the host stat is null? Then, the message is clearer.

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -420,55 +505,65 @@ protected void runInContext() {
...
-                    gpuEnabledHosts = hosts;
-                }
-                for (HostVO host : gpuEnabledHosts) {
-                    HashMap<String, HashMap<String, VgpuTypesInfo>> groupDetails = _resourceMgr.getGPUStatistics(host);
-                    if (groupDetails != null) {
-                        _resourceMgr.updateGPUDetails(host.getId(), groupDetails);
-                    }
+
+                if (externalStatsType == ExternalStatsProtocol.INFLUXDB) {
+                    sendMetricsToInfluxdb(metrics);
```
rafaelweingartner: So, this data is only sent to Influx?

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -420,55 +505,65 @@ protected void runInContext() {
...
+                updateGpuEnabledHostsDetails(hosts);
             } catch (Throwable t) {
                 s_logger.error("Error trying to retrieve host stats", t);
             }
         }
+
+        /**
+         * Updates GPU details on hosts supporting GPU.
+         */
+        private void updateGpuEnabledHostsDetails(List<HostVO> hosts) {
```
rafaelweingartner: This method is very odd. Why not load all host ids that have GPUs as the first thing? And then proceed with the execution? It is weird to execute once, then load some state data, and then in the second execution, you use that data (in this context, it is not making much sense).

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -477,16 +572,10 @@ protected void runInContext() {
                 SearchCriteria<HostVO> sc = _hostDao.createSearchCriteria();
```
rafaelweingartner: I have seen this before. Can you reuse it?

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -692,64 +734,62 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
...
-                                            ((previousVmDiskStats.getCurrentBytesRead() != vmDiskStat_lock.getCurrentBytesRead()) ||
-                                                    (previousVmDiskStats.getCurrentBytesWrite() != vmDiskStat_lock.getCurrentBytesWrite()) ||
-                                                    (previousVmDiskStats.getCurrentIORead() != vmDiskStat_lock.getCurrentIORead()) || (previousVmDiskStats.getCurrentIOWrite() != vmDiskStat_lock.getCurrentIOWrite()))) {
-                                        s_logger.debug("vm disk stats changed from the time GetVmDiskStatsCommand was sent. " + "Ignoring current answer. Host: " +
-                                                host.getName() + " . VM: " + vmDiskStat.getVmName() + " Read(Bytes): " + vmDiskStat.getBytesRead() + " write(Bytes): " +
-                                                vmDiskStat.getBytesWrite() + " Read(IO): " + vmDiskStat.getIORead() + " write(IO): " + vmDiskStat.getIOWrite());
+                                    if (previousVmDiskStats != null && ((previousVmDiskStats.getCurrentBytesRead() != vmDiskStat_lock.getCurrentBytesRead())
```
rafaelweingartner: Can you extract the content of this conditional (not the IF body) to a method?

server/src/main/java/com/cloud/server/StatsCollector.java
GabrielBrascher: Good catch, the name might mislead. I will update to `retrieveExternalStatsPortFromUri`

server/src/main/java/com/cloud/server/StatsCollector.java
GabrielBrascher: Sure, good point. Thanks!

server/src/main/java/com/cloud/server/StatsCollector.java
GabrielBrascher: Yes, indeed. Graphite was only receiving data from the VmStatsCollector.
I could have added Graphite on the HostCollector as well, but I think that is better to leave such addition to another PR.

server/src/main/java/com/cloud/server/StatsCollector.java
GabrielBrascher: Fixed, thanks.

server/src/main/java/com/cloud/server/StatsCollector.java
GabrielBrascher: Good catch! Done

server/src/main/java/com/cloud/server/StatsCollector.java
GabrielBrascher: Thanks, @rafaelweingartner. Updated!

server/pom.xml
GabrielBrascher: fixed!

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -411,82 +462,111 @@ private void init(Map<String, String> configs) {
...
+            if (externalStatsType == ExternalStatsProtocol.INFLUXDB) {
+                return INFLUXDB_DEFAULT_PORT;
+            }
+        }
+        throw new URISyntaxException(uri.toString(), String.format(
+                "Cannot define a port for the Stats Collector host %s://%s:%s or URI scheme is incorrect. The configured URI in stats.output.uri is not supported. Please configure as the following examples: graphite://graphite-hostaddress:port, or influxdb://influxdb-hostaddress:port. Note that the port is optional, if not added the default port for the respective collector (graphite or influxdb) will be used.",
+                externalStatsPrefix, externalStatsHost, externalStatsPort));
+    }
+
+    class HostCollector extends AbstractStatsCollector {
```
DaanHoogland: As we now have a hierarchy of collectors and this file is already over  a 1000 lines it makes sense to factor these three classes out into their own package. Also a helper class for all the query methods seem prudent, at this time.

utils/src/main/java/org/apache/cloudstack/utils/graphite/GraphiteClient.java
```diff
@@ -67,7 +67,7 @@ protected long getCurrentSystemTime() {
      *
      * @param metrics the metrics as key-value-pairs
      */
-    public void sendMetrics(Map<String, Integer> metrics) {
+    public void sendMetrics(Map<Object, Object> metrics) {
```
DaanHoogland: I think I missed the functional need for this type change. Can you expand @GabrielBrascher

DaanHoogland:
I pressed a UI widget to early, @GabrielBrascher . My comment is that the extracting of the inner classes is probably better as part of this PR. I don't think it will be a large job, do you?

utils/src/main/java/org/apache/cloudstack/utils/graphite/GraphiteClient.java
GabrielBrascher: @DaanHoogland this method was restricted to the Graphite workflow; I needed to send another type of Objects for InfluxDB. I could create two different metrics maps or make a generic map of metrics. You can check  lines 600 and 603 of `StatsCollector.java` to see the points where I need to use the map for both cases.

GabrielBrascher:
@DaanHoogland thanks for pointing the inner class issue; I totally agree with you. I tried to go a bit further on my PR and fix this design; however the PR would be too large.

The problem here is that such change will require a major refactoring on the stats collector architecture. It is not just about extracting the inner classes, we need to change their implementation to use object orientation in order to reduce the number of lines. My idea is to get back on this class and enhance it on +1 or +2 PRs. This way it will be easier for reviewers and for me to work on it.

GabrielBrascher:
Updated PR pushing the following change.

In order to avoid redudancy, removed the following tags:
```
private static final String HOST_ID_TAG = "host_id";
private static final String VM_ID_TAG = "vm_id";
private static final String INSTANCE_NAME_TAG = "instance_name";
private static final String POD_ID_TAG = "pod_id";
private static final String DATA_CENTER_ID_TAG = "data_center_id";
```

Leaving VMs and hosts tagged only by their UUID, which is sufficient.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
I have been testing this for a while, it is all working fine. JUnit tests have passed.
I would really appreciate some extra reviews @DaanHoogland @kiwiflyer @rhtyd @nvazquez @nathanejohnson @dhlaluku @borisstoyanov @bwsw and others.
Thank you, guys!

Additionally, Travis has been failing for PRs, I can take a look into it as well. I see that PR #3115 is already tackling some failures.

utils/src/main/java/org/apache/cloudstack/utils/graphite/GraphiteClient.java
DaanHoogland: i see

                            metrics.clear();
                        }
as lines 602 and 603 but no change before it. Are these still the lines you mean?

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -497,86 +572,35 @@ protected void runInContext() {
...
-                                        g.sendMetrics(metrics);
-                                    } catch (GraphiteException e) {
-                                        s_logger.debug("Failed sending VmStats to Graphite host " + externalStatsHost + ":" + externalStatsPort + ": " + e.getMessage());
-                                    }
-
-                                    metrics.clear();
+                                if (externalStatsType == ExternalStatsProtocol.GRAPHITE) {
+                                    sendVmMetricsToGraphiteHost(metrics, host);
+                                } else if (externalStatsType == ExternalStatsProtocol.INFLUXDB) {
+                                    sendMetricsToInfluxdb(metrics);
```
GabrielBrascher: @DaanHoogland sorry, the lines 600 - 603 have changed, now are 596 and 598. 
I was referring to the following mthods `sendVmMetricsToGraphiteHost(metrics, host);` and `sendMetricsToInfluxdb(metrics);`

I had two options: (i) keep the Graphite metrics Map as it was, and create a new map for the influxdb (for instance metricsInfluxDB):
```
if (externalStatsType == ExternalStatsProtocol.GRAPHITE) {
    sendVmMetricsToGraphiteHost(metrics, host);
} else if (externalStatsType == ExternalStatsProtocol.INFLUXDB) {
    sendMetricsToInfluxdb(metricsInfluxDB);
```
Or (ii) change the metrics map from `Map<String, Integer>` to a generic `Map<Object, Object>`, which implicates on changing the GraphiteClient.java sendMetrics method, and reuse the `metrics` for InfluxDB and Graphite.

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -155,16 +164,51 @@ public String toString() {
...
+    private static final String DEFAULT_DATABASE_NAME = "cloudstack";
+    private static final String INFLUXDB_HOST_MEASUREMENT = "host_stats";
+    private static final String INFLUXDB_VM_MEASUREMENT = "vm_stats";
+
     static final ConfigKey<Integer> vmDiskStatsInterval = new ConfigKey<Integer>("Advanced", Integer.class, "vm.disk.stats.interval", "0",
             "Interval (in seconds) to report vm disk statistics. Vm disk statistics will be disabled if this is set to 0 or less than 0.", false);
     static final ConfigKey<Integer> vmDiskStatsIntervalMin = new ConfigKey<Integer>("Advanced", Integer.class, "vm.disk.stats.interval.min", "300",
             "Minimal interval (in seconds) to report vm disk statistics. If vm.disk.stats.interval is smaller than this, use this to report vm disk statistics.", false);
     static final ConfigKey<Integer> vmNetworkStatsInterval = new ConfigKey<Integer>("Advanced", Integer.class, "vm.network.stats.interval", "0",
             "Interval (in seconds) to report vm network statistics (for Shared networks). Vm network statistics will be disabled if this is set to 0 or less than 0.", false);
```
dhlaluku: @GabrielBrascher since your new Config has private access, does it make sense to make all the rest private? I think that will promote/enforce a certain style to the next person who adds another setting here.

GabrielBrascher: Updated, thanks.

GabrielBrascher:
@blueorangutan package

rhtyd:
@GabrielBrascher thanks for replying, I saw the email about the PR being merged and I spent most of the time reviewing the diff, skipping the commentary. Glad that it's on your radar to refactor the sub-system as the next step.

PR #3085, opened 2018-12-07
server/test: Move test files to right location

wido:
## Description
This PR only moves some files for Unit Testing purposes.

These files were not in the right directory and thus not being executed by Maven.

By moving the files we make sure these tests are run again.

## Types of changes

## How Has This Been Tested?
Run the tests locally afterwards and they succeed.


rhtyd:
Changes are only file/path move of test classes, no need of regression test as long as travis job#1 passes that builds cloudstack and runs unit tests.

PR #2493, opened 2018-03-16
CLOUDSTACK-10326: Prevent hosts fall into Maintenance when there are running VMs on it

nvazquez:
JIRA Ticket: https://issues.apache.org/jira/browse/CLOUDSTACK-10326

This issue was discovered, fixed and tested on KVM, but applies for every hypervisor.

### Background
When enabling maintenance mode in a host, host state is put into 'PrepareForMaintenance' and running VMs are migrated into another host. After every VM is migrated, host goes to 'Maintenance' state.

Checks are performed on `ResourceManagerImpl.checkAndMaintan()` method:

- List VMs with host_id = HOST_ID
- List VMs with last_host_id = HOST_ID and state=Migrating

When both queries are empty, then the host can be put into Maintenance.

When a VM is being migrated to DEST_HOST, its host_id column is set to DEST_HOST, last_host_id = ORIGIN_HOST and state = Migrating. If then migration fails, host_id = last_host_id = ORIGIN_HOST 

### Issue
This sequence:

- Enable maintenance mode on ORIGIN_HOST
- VMs start being migrated to a host, say DEST_HOST
- checkAndMaintain() starts:
   - First check passes (no VM with host_id = ORIGIN_HOST_ID as those are being migrated)
   - Before the second check, one or more migrations fail
   - Second check passes, however there are VMs running on the host as migrations have failed.
- Host goes into Maintenance state.

Screenshots attached, query executed on each case:
`select id, name, instance_name, state, host_id, last_host_id from vm_instance;`

Before enabling maintenance mode on host 4:
cloudstack-10326-initialstate

While host = 'PrepareForMaintenance' and VM is being migrated to host 1:
cloudstack-10326-migrating

At this point the first check is performed:
cloudstack-10326-debug1

Made migrations fail adding these rules on host 4:
````
iptables -I OUTPUT -j REJECT -m state --state NEW -m tcp -p tcp --dport 49152:49215 -m comment --comment 'test block migrations'
iptables -I OUTPUT -j REJECT -m state --state NEW -m tcp -p tcp --dport 16509 -m comment --comment 'test block migrations'
````
Migration failed and VM goes Running into host 4:
cloudstack-10326-migrationfailed

Second check passes and host goes into Maintenance
cloudstack-10326-debug

PR #3095, opened 2018-12-13
Prevent corner case for infinite PrepareForMaintenance

nvazquez:
## Description
A corner case was found on 4.11.2 for #2493 leading to an infinite loop in state `PrepareForMaintenance`

To prevent such cases, in which failed migrations are detected but still running on the host, this feature adds a new cluster setting `host.maintenance.retries` which is the number of retries before marking the host as `ErrorInMaintenance` if migration errors persist.

## Types of changes

## How Has This Been Tested?
- 2 KVM hosts, pick one which has running VMs as H
- Block migrations ports on H to simulate failures on migrations:
`
iptables -I OUTPUT -j REJECT -m state --state NEW -m tcp -p tcp --dport 49152:49215 -m comment --comment 'test block migrations'
iptables -I OUTPUT -j REJECT -m state --state NEW -m tcp -p tcp --dport 16509 -m comment --comment 'test block migrations
`
- Put host H in Maintenance
- Observe that host is indefinitely in `PrepareForMaintenance` state (after this fix it goes into `ErrorInMaintenance` after retrying `host.maintenance.retries` times)

### DIFF
--- a/server/src/com/cloud/resource/ResourceManagerImpl.java
+++ b/server/src/com/cloud/resource/ResourceManagerImpl.java
@@ -26,6 +26,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
+import java.util.concurrent.ConcurrentHashMap;
 
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
@@ -272,7 +273,7 @@ public class ResourceManagerImpl extends ManagerBase implements ResourceManager,
 
     private SearchBuilder<HostGpuGroupsVO> _gpuAvailability;
 
-    private Map<Long,Integer> retryHostMaintenance = new HashMap<>();
+    private Map<Long,Integer> retryHostMaintenance = new ConcurrentHashMap<>();
 
     private void insertListener(final Integer event, final ResourceListener listener) {
         List<ResourceListener> lst = _lifeCycleListeners.get(event);
--- a/server/src/com/cloud/resource/ResourceManagerImpl.java
+++ b/server/src/com/cloud/resource/ResourceManagerImpl.java
@@ -31,6 +31,7 @@ import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
 import com.cloud.vm.dao.UserVmDetailsDao;
+import org.apache.cloudstack.framework.config.ConfigKey;
 import org.apache.commons.lang.ObjectUtils;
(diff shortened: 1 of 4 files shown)


nvazquez:
@blueorangutan package

server/src/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -1350,7 +1364,23 @@ protected boolean isHostInMaintenance(HostVO host, List<VMInstanceVO> runningVms
             return CollectionUtils.isEmpty(failedMigrations) ?
                     setHostIntoMaintenance(host) :
                     setHostIntoErrorInMaintenance(host, failedMigrations);
+        } else if (retryHostMaintenance.containsKey(host.getId())) {
```
rafaelweingartner: you do not need the else here. I mean, the other if body will always return. Therefore, you can simply follow with an IF. This will change a bit the formatting.

server/src/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -1350,7 +1364,23 @@ protected boolean isHostInMaintenance(HostVO host, List<VMInstanceVO> runningVms
...
+            Integer retriesLeft = retryHostMaintenance.get(host.getId());
+            if (retriesLeft != null) {
```
rafaelweingartner: If the `retriesLeft ` is null, does it mean that ACS will retry forever?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

server/src/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -271,6 +272,8 @@ public void setDiscoverers(final List<? extends Discoverer> discoverers) {
 
     private SearchBuilder<HostGpuGroupsVO> _gpuAvailability;
 
+    private Map<Long,Integer> retryHostMaintenance = new HashMap<>();
```
rhtyd: This may be called by concurrent threads of execution, @nvazquez can you use a concurrent hashmap and/or instead of integer perhaps one of the atomic ints?

nvazquez: Done, thanks

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

PR #3100, opened 2018-12-18
ipv6: Send userdata to Virtual Router if IPv6 is enabled

wido:
## Description
There is no reason to not send userdata+password to the VR as all Instances in CloudStack are Dual-Stacked. They have IPv4 and IPv6 so they can query their metadata over IPv4 at the VR.

## Types of changes

## How Has This Been Tested?
Tested on a local testing environment with IPv6 enabled.

### DIFF
--- a/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
+++ b/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
@@ -1022,11 +1022,6 @@ NetworkMigrationResponder, AggregatedCommandExecutor, RedundantResource, DnsServ
                 return false;
             }
 
-            if (network.getIp6Gateway() != null) {
-                s_logger.info("Skip password and userdata service setup for IPv6 VM");
-                return true;
-            }
-
             final VirtualMachineProfile uservm = vm;
 
             final List<DomainRouterVO> routers = getRouters(network, dest);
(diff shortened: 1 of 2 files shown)


rhtyd:
@wido have you tested both userdata and password features, I think the password server may not listen on ipv6 interface (needs verification) and also the cloud-init compatibility to consume userdata in guests. Can you advise your test setup and how we can test/review?

wido:
@rhtyd The password server does not listen on IPv6, that is correct.

However, all Instances deployed on CloudStack have IPv4 and IPv6. Cloud-init will use IPv4 for fetching Userdata and Passwords from the VR.

IPv6 is not used in this case.

The original code would not send the data to the VR if IPv6 was enabled for that Instance, but that is not correct.

The Instance (cloud-init) will use IPv4 to fetch the information.

rhtyd:
@wido thanks for the explaination. So, basically it removes a check and allows userdata+password for ipv4+ipv6 enabled vms and networks, that's all.
@blueorangutan package

rhtyd:
Looks like some travis smoketest failures could be related, can you check @wido -- also we need to fix the failing travis+smoketests soon (I'm on holidays this week so couldn't try to fix them).

wido:
Indeed @rhtyd , it removes a condition. That if a Instance has IPv6 enabled we still send the Userdata to the VR.

In Basic Networking this already works, but for Advanced Networking there was this limitation.

bwsw:
@rhtyd @wido Shouldn't the ticket addressed in 4.11 branch. Looks like it's a bug, no?

wido:
@bwsw I highly doubt if anybody is using IPv6 with Advanced Networking with 4.11

anuragaw:
Code LGTM

PR #3108, opened 2018-12-21
Throw InvalidParameterValueException for failing ApiArgValidator

anuragaw:
This is important because it helps in communicating back the exact
error to the API callee.

Current behavior is that ParamProcessWorker#processParameters catches
the exception and returns an incorrect type exception without the
proper message.

## Types of changes

## How Has This Been Tested?

Use Cloudmonkey to call an API which has ApiArgValidator set for params. This does not return the proper exception message but sends an internal error occured with no information.


anuragaw:
@blueorangutan package

anuragaw:
@blueorangutan test

PR #3112, opened 2018-12-28
network: Offerings do not have to have Security Grouping enabled

wido:
## Description
Offerings can co-exist where on does provide Security Grouping in the
network, but other guest Networks have no Security Grouping.

In V(X)LAN isolation environments the L2 separation is handled by V(X)LAN
and protection between Instances is handled by Security Grouping.

There are multiple scenarios possible where one network has Security Grouping
enabled because that is required in that network.

In the other network, but in the same zone it could be a choice to have
Security Grouping disabled and allow all traffic to flow.

## Types of changes

## How Has This Been Tested?
Tested on a Local KVM cloud where VXLAN isolation was being used.

Observed that VMs deployed in the first network have SG enabled and in a other network have it disabled.

### DIFF
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -2246,9 +2246,6 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
                 if (_networkModel.areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SourceNat)) {
                     throw new InvalidParameterValueException("Service SourceNat is not allowed in security group enabled zone");
                 }
-                if (!_networkModel.areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SecurityGroup)) {
-                    throw new InvalidParameterValueException("network must have SecurityGroup provider in security group enabled zone");
-                }
             }
 
             //don't allow eip/elb networks in Advance zone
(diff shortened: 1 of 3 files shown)


DaanHoogland:
@wido , without disapreciation of your desired scenario, as an admin would I not be able to enforce security groups on a zone level? I think I'd want to. Then if I don't, your scenario makes sense.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -2246,9 +2246,6 @@ public Network createGuestNetwork(final long networkOfferingId, final String nam
                 if (_networkModel.areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SourceNat)) {
                     throw new InvalidParameterValueException("Service SourceNat is not allowed in security group enabled zone");
                 }
-                if (!_networkModel.areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SecurityGroup)) {
```
rafaelweingartner: Instead of removing the check, shouldn't we fix the `ntwkOff` object and related objects/attributes/configurations that are used to identify if the network supports security groups?

wido: I'm not sure I get what you are hinting at. I'm just removing the check that SG isn't mandatory.

Could you elaborate a bit more on what you mean?

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
rhtyd: lgtm

rafaelweingartner: @wido it was my mistake. After checking this code with @GabrielBrascher, I now understood what you did.

PR #3113, opened 2018-12-28
kvm: Security Group enhancements and refactor old code

wido:
## Description
This PR enhances the security group Python script. It refactors old code, removes deprecated code and fixes many bugs.

- PEP-8 styling
- Remove old LXC code from Python script
- Remove deprecated optparse from Python
- Use proper calls to iptables

## Types of changes

## How Has This Been Tested?
I have tested this on a Local Cloud we have running with the master branch.

Tested both IPv4 and IPv6 in a Advanced Zone using VXLAN L2 isolation.

### DIFF
--- a/scripts/vm/network/security_group.py
+++ b/scripts/vm/network/security_group.py
@@ -148,6 +148,11 @@ def split_ips_by_family(ips):
     return ip4s, ip6s
 
 
+def get_bridge_physdev(brname):
+    physdev = execute("bridge -o link show | awk '/master %s / && !/^[0-9]+: vnet/ {print $2}'" % brname)
+    return physdev.strip()
+
+
 def destroy_network_rules_for_vm(vm_name, vif=None):
     vmchain = iptables_chain_name(vm_name)
     vmchain_egress = egress_chain_name(vm_name)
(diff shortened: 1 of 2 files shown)


dhlaluku:
@blueorangutan package

scripts/vm/network/security_group.py
```diff
@@ -35,11 +33,6 @@
 logpath = "/var/run/cloud/"        # FIXME: Logs should reside in /var/log/cloud
```
dhlaluku: Since you are refactoring, should we perhaps fix the logpath as well?

dhlaluku:
@blueorangutan test

scripts/vm/network/security_group.py
wido: I am not sure what the implications might be.

People might be parsing the files at that location, but also, if they are in /var/run/cloud they will be gone after a reboot (which is good!)

rafaelweingartner:
@wido would you mind rebasing on top of current master?
Travis is presenting some odd results, which I think are not caused by your changes.

scripts/vm/network/security_group.py
GabrielBrascher: 5.0.0 sounds like a good plan.

wido:
@rafaelweingartner I have just rebased against master and force pushed again

GabrielBrascher:
@blueorangutan package

dhlaluku:
@blueorangutan package

dhlaluku:
@blueorangutan test

GabrielBrascher:
Merging this one as the tests are looking good and we have 3 LGTMs.
Thanks for the PR @wido, and for the reviews @rafaelweingartner and @dhlaluku.

PR #3120, opened 2019-01-07
Rework for destroy VM with volumes UI

dhlaluku:
## Description

This PR removes the redundant checkbox to tick if a user wants to delete volumes when destroying VMs

## Types of changes

image

### DIFF
--- a/ui/l10n/en.js
+++ b/ui/l10n/en.js
@@ -1816,7 +1816,7 @@ var dictionary = {
 "label.volume":"Volume",
 "label.volume.details":"Volume details",
 "label.volume.empty":"No volumes attached to this VM",
-"label.volume.ids":"Volume ID's",
+"label.volume.ids":"Delete Volumes",
 "label.volume.limits":"Volume Limits",
 "label.volume.migrated":"Volume migrated",
 "label.volume.name":"Volume Name",
--- a/ui/scripts/instances.js
+++ b/ui/scripts/instances.js
@@ -116,13 +116,14 @@
                     volumes: {
                         label: 'label.delete.volumes',
                         isBoolean: true,
-                        isChecked: false
+                        isChecked: true,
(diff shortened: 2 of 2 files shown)


ui/l10n/en.js
```diff
@@ -1816,7 +1816,7 @@ var dictionary = {
 "label.volume":"Volume",
 "label.volume.details":"Volume details",
 "label.volume.empty":"No volumes attached to this VM",
-"label.volume.ids":"Volume ID's",
+"label.volume.ids":"Delete Volumes",
```
GabrielBrascher: @dhlaluku I would recommend keeping `label.volume.ids` as it is ("Volume ID's"). Instead, you can use the existing `"label.delete.volumes":"Volumes to be deleted"` when listing the volumes IDs to be deleted.

ui/scripts/instances.js
```diff
@@ -116,13 +116,14 @@
                     volumes: {
                         label: 'label.delete.volumes',
                         isBoolean: true,
-                        isChecked: false
+                        isChecked: true,
+                        isHidden: true,
                     },
                     volumeids: {
                         label: 'label.volume.ids',
```
GabrielBrascher: What do you think of using `label.delete.volumes` here?

ui/l10n/en.js
dhlaluku: Fixed and reverted back the Volume ID's label

GabrielBrascher: Thanks!

dhlaluku:
@rafaelweingartner @wido any suggestions with this? Otherwise, I think we can merge it since it is rather a very small rework. /CC @DaanHoogland @andrijapanic @rhtyd

borisstoyanov:
@blueorangutan package

Issue #3039, opened 2018-11-20
VMware 4.11.2.0 system VMs memory consumption grows overtime until heavy swapping occurs


rhtyd:
Memory Consumption on vSphere
vRouter are starting to swap with low memory available.
Increasing memory only causes a longer time period before heavy swapping occurs.

@resmo shared a related issue on dev list: https://github.com/systemd/systemd/issues/8015

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
ACS 4.11.x
~~~

##### OS / ENVIRONMENT

Seen only in VMware environments

##### STEPS TO REPRODUCE

~~~
- Deploy VR in vmware
- SSH into it from management server
- See memory growth over time
~~~

##### EXPECTED RESULTS

~~~
No additional memory consumption.
~~~

onitake:
Hi @rhtyd @PaulAngus - thanks for taking on this issue.

If you have any suggestions on more tests we could run, please share.

onitake:
Is the fix going to be something like the one mentioned here https://lists.apache.org/thread.html/cb36b5efe8327c080919086688707650a3ff036c884a2dd87dd6e66d@%3Cdev.cloudstack.apache.org%3E , or is this an unrelated issue?

onitake:
We were able to drive memory usage through the roof with SSH connections, because each session would leak memory. But we also had success with lots of network reconfiguration requests, so I guess both are valid test cases.

onitake:
@PaulAngus Is it also possible to get just the templates in lieu of full 4.11.2.1 RPMs?
If I try to access your package server through a web browser, I just get a blank page.

andrijapanicsb:
Any progress on this one - anyone else to share experience or a proper fix ?

PaulAngus:
This is addressed by https://github.com/apache/cloudstack/pull/3127

andrijapanicsb:
Pardon my blind eyes ... ... ...

rhtyd:
Interesting notes from https://github.com/systemd/systemd/issues/8015#issuecomment-475324192
```
On Debian I fix the memory consumption commenting the following line on /etc/pam.d/common-session:

#session optional pam_systemd.so

But, with that, the ssh sessions doesn't terminate correctly. The fix is already done:

cp /usr/share/doc/openssh-client/examples/ssh-session-cleanup.service /etc/systemd/system/
[...]

PR #3126, opened 2019-01-11
Improve System VM startup and memory usage

PaulAngus:
## Description

In order to reduce memory footprint and improve boot speed/predictability; The following changes have been made:

- Add vm.min_free_kbytes to sysctl
- periodically clear disk cache (depending on memory size)
- only start guest services specific to hypervisor
- use systemvm code to determine hypervisor type (not systemd)
- start cloud service at end of post init rather than through systemd
- reduce initial threads started for httpd
- fix vmtools config file
- disable all required services (do not start on boot)

Some changes will require a new systemvmtemplate that can be done in 4.14:
- start only required services during post init.
- allow one-shot ACS configuration services to terminate after running.

Fixes: #3039
## Types of changes

## How Has This Been Tested?

Changes have been subjected to regression testing and a burn-in test where portforwarding rules were constantly updated over a period of 3 days.  no swapping occured in a 256MB RAM VR.


PaulAngus:
@blueorangutan package

PaulAngus:
@blueorangutan test matrix

wido:
See my comments here: #3127 

I would not touch those sysctl settings

PaulAngus:
Annotation 2019-03-28 134017

Continuing to monitor the above VR while running:
`while true; do ssh -p3922 -i /var/cloudstack/management/.ssh/id_rsa root@10.2.7.101 "echo hi"; done`
memory usage growth in negligible

PaulAngus:
RESULTS of tests
The following script was left running from the management server:
while true; do ssh -p3922 -i /var/cloudstack/management/.ssh/id_rsa root@10.2.7.101 "echo hi"; done

The memory usage was then monitored by periodically running htop from the VR console.

| Uptime  | Memory Usage |
| ------------- | ------------- |
| 08:50  | 82.3MB |
| 1d 00:43 | 94.3 MB |
| 1d 04:51 | 94.8 MB |
| 1d 22:57 | 97.4 MB |
| 2d 07:42 | 91.9 MB |
| 5d 05:35 | 92.0 MB |

I believe that the variance is due to the timing of the check relative to the disk cache flushing.

IMO - I think that we can consider this a successful test

andrijapanicsb:
Based on the original issues reported, this looks LGTM.

I'm also OK with those systemctl settings, since kernel *should* do it's job, but it doesn't in some cases (had silly experience with same fix on systems with 64GB of ram, running GFS2)

onitake:
I wonder... If swapping is such a big concern, why not simply turn off the swap completely?
If the VM is so badly under pressure that it needs to swap, it's probably better to just let it OOM kill and reboot itself right away, than suffer under terrible performance due to heavy I/O load.

rhtyd:
Rebased to latest master and fixed conflicts.
I think @PaulAngus intends to get rid of the postinit service but any changes to the cloud-early-config systemd service would require creating a new systemvmtemplate. So, I propose such changes requiring systemd/cloud-early-config changes be addressed in 4.14 where we'll need a new Debian 10 based systemvmtemplate.

I'll run some tests, this is still in WIP.

systemvm/debian/lib/systemd/system/open-vm-tools.service
```diff
@@ -0,0 +1,12 @@
+﻿[Unit]
+Description=Service for virtual machines hosted on VMware
+Documentation=http://open-vm-tools.sourceforge.net/about.php
+DefaultDependencies=no
+Before=cloud-init-local.service
```
rhtyd: @PaulAngus I see references to cloud-init-local.service and cloud-init.service but they are not in codebase or the template. Is there any file to be added, or did you mean cloud-early-config.service?

rhtyd:
Pinging for review - @nvazquez @shwstppr @anuragaw @borisstoyanov @andrijapanic 
@blueorangutan package

borisstoyanov:
@blueorangutan test matrix

rhtyd:
Created this issue to track further future changes: https://github.com/apache/cloudstack/issues/3426

PR #3127, opened 2019-01-11
Improve System VM startup and memory usage [4.11]

PaulAngus:
## Description

In order to reduce memory footprint and improve boot speed/predictability; The following changes have been made:

- Add vm.min_free_kbytes to sysctl
- periodically clear disk cache (depending on memory size)
- only start guest services specific to hypervisor
- use systemvm code to determine hypervisor type (not systemd)
- start cloud service at end of post init rather than through systemd
- reduce initial threads started for httpd
- fix vmtools config file
- disable all required services (do not start on boot)
- start only required services during post init.
- allow one-shot ACS configuration services to terminate after running.

Fixes: #3039
## Types of changes

## How Has This Been Tested?

Changes have been subjected to regression testing and a burn-in test where portforwarding rules were constantly updated over a period of 3 days.  no swapping occured in a 256MB RAM VR.

### DIFF
--- a/systemvm/debian/etc/cron.daily/clear_cache
+++ b/systemvm/debian/etc/cron.daily/clear_cache
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+# clear memory cache to ultimately reduce swapping
+
+sync && echo 1 > /proc/sys/vm/drop_caches
--- a/systemvm/debian/etc/cron.hourly/clear_cache
+++ b/systemvm/debian/etc/cron.hourly/clear_cache
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# clear memory cache to ultimately reduce swapping
+
+phymem=$(free|awk '/^Mem:/{print $2}')
+if [ $phymem -lt 513000 ]; then
+  sync && echo 1 > /proc/sys/vm/drop_caches
+fi
--- a/systemvm/debian/etc/vmware-tools/tools.conf
+++ b/systemvm/debian/etc/vmware-tools/tools.conf
@@ -1,0 +1,17 @@
+[logging]
+# Turns on logging globally. It can still be disabled for each domain.
+# log = true
+
+# Disables core dumps on fatal errors; they're enabled by default.
+enableCoreDump = false
+
+# Defines the "vmsvc" domain, logging to file
+# vmsvc.level = message
+vmsvc.handler = file
+# Setup file rotation - keep 3 files
+vmsvc.maxOldLogFiles = 2
+# Max log file size kept: 1 MB
+vmsvc.maxLogSize = 1
+
+# Defines the "vmtoolsd" domain, and disable logging for it.
+# vmtoolsd.level = none
--- a/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
+++ b/systemvm/debian/opt/cloud/bin/setup/cloud-early-config
@@ -23,7 +23,8 @@ CMDLINE=/var/cache/cloud/cmdline
 
 # Clear boot up flag, it would be created by rc.local after boot up done
 mkdir -p /var/cache/cloud
-rm -f /var/cache/cloud/boot_up_done
+rm -f /var/cache/cloud/enabled_svcs
+rm -f /var/cache/cloud/disabled_svcs
 
 [ -x /sbin/ifup ] || exit 0
 
@@ -205,6 +206,50 @@ patch() {
   return 0
 }
 
+
+enable_guest_services() {
+
+HYPERVISOR=$(</etc/cloudstack-agent_detected_hypervisor)
+  case $HYPERVISOR in
+     xen-pv|xen-domU)
+        systemctl stop ntpd
+	systemctl disable ntpd
+        systemctl start xe-daemon
+        ;;
+     xen-hvm)
+        systemctl stop ntpd
+	systemctl disable ntpd
+        systemctl start xe-daemon
+        ;;
+     kvm)
+        systemctl start qemu-guest-agent
+        VPORT=$(find /dev/virtio-ports -type l -name '*.vport' 2>/dev/null|head -1)
+        ;;
+     vmware)
+        # system time sync'd with host via vmware tools
+        systemctl stop ntpd
+	systemctl disable ntpd
+        systemctl start open-vm-tools
+        ;;
+     virtualpc|hyperv)
+        # Hyper-V is recognized as virtualpc hypervisor type. Boot args are passed using KVP Daemon
+        systemctl start hyperv-daemons.hv-fcopy-daemon.service hyperv-daemons.hv-kvp-daemon.service hyperv-daemons.hv-vss-daemon.service
+        ;;
+     virtualbox)
+        # Virtualbox is used to test the virtual router
+        ;;
+   esac
+}
+
+add_cache_backpressure() {
+  # set cache back pressure based on amount of physical memory 100 is default
+  physmem=$(free|awk '/^Mem:/{print $2}')
+  if [ $((physmem)) -lt 409600 ]; then
+      sed  -i "/^vm.vfs_cache_pressure/ c\vm.vfs_cache_pressure = 200" /etc/sysctl.conf
+  else
+      sed  -i "/^vm.vfs_cache_pressure/ c\vm.vfs_cache_pressure = 100" /etc/sysctl.conf
+  fi
+}
 start() {
   log_it "Executing cloud-early-config"
 
@@ -217,13 +262,19 @@ start() {
   export HYPERVISOR=$(hypervisor)
   [ $? -ne 0 ] && log_it "Failed to detect hypervisor type, bailing out of early init" && exit 10
   log_it "Detected that we are running inside $HYPERVISOR"
+  echo "$HYPERVISOR" > /etc/cloudstack-agent_detected_hypervisor
 
   config_guest
   get_boot_params
   get_systemvm_type
   patch
   sync
+  add_cache_backpressure
   sysctl -p
+  
+  log_it "Starting Virtual Guest Services for $HYPERVISOR"
+
+  enable_guest_services
 
   log_it "Configuring systemvm type=$TYPE"
 
@@ -232,6 +283,10 @@ start() {
   else
       /opt/cloud/bin/setup/default.sh
   fi
+ 
+  log_it "Running postinit to start services"
+
+  /opt/cloud/bin/setup/postinit.sh
 
   log_it "Finished setting up systemvm"
 
--- a/systemvm/debian/opt/cloud/bin/setup/common.sh
+++ b/systemvm/debian/opt/cloud/bin/setup/common.sh
@@ -583,8 +583,8 @@ setup_ntp() {
 }
 
 routing_svcs() {
-   echo "cloud haproxy apache2" > /var/cache/cloud/enabled_svcs
-   echo "nfs-common portmap" > /var/cache/cloud/disabled_svcs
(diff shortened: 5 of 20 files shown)


PaulAngus:
@blueorangutan package

PaulAngus:
@blueorangutan matrix

PaulAngus:
@blueorangutan test matrix

systemvm/debian/etc/vmware-tools/tools.conf
```diff
@@ -0,0 +1,17 @@
+[logging]
+# Turns on logging globally. It can still be disabled for each domain.
+# log = true
+
+# Disables core dumps on fatal errors; they're enabled by default.
+enableCoreDump = false
```
rhtyd: Does this work with openvmtools @PaulAngus as we don't install vmware tools, unless you've changed that.

systemvm/debian/opt/cloud/bin/setup/cloud-early-config
```diff
@@ -205,6 +206,50 @@ patch() {
   return 0
 }
 
+
+enable_guest_services() {
+
+HYPERVISOR=$(</etc/cloudstack-agent_detected_hypervisor)
+  case $HYPERVISOR in
+     xen-pv|xen-domU)
+        systemctl stop ntpd
```
rhtyd: We can have all three xen cases in this block.

systemvm/debian/opt/cloud/bin/setup/common.sh
```diff
@@ -583,14 +583,14 @@ setup_ntp() {
 }
 
 routing_svcs() {
-   echo "haproxy apache2" > /var/cache/cloud/enabled_svcs
-   echo "cloud nfs-common portmap" > /var/cache/cloud/disabled_svcs
+   echo "cloud haproxy apache2" > /var/cache/cloud/enabled_svcs
```
rhtyd: On VRs we don't need to run any agent that is run by the cloud service.

tools/appliance/systemvmtemplate/scripts/configure_systemvm_services.sh
```diff
@@ -108,6 +108,19 @@ function configure_services() {
   systemctl disable strongswan
   systemctl disable x11-common
   systemctl disable xl2tpd
+  systemctl disable vgauth
+  systemctl disable sshd
```
rhtyd: This may cause error in somecases, ideally we want sshd enabled in all cases.

tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
```diff
@@ -68,9 +68,10 @@ function install_packages() {
     python-flask \
     haproxy \
     radvd \
-    sharutils genisoimage aria2 \
+    sharutils genisoimage \
```
rhtyd: This may break one of the register template features that uses aria2, @nvazquez can you comment?

tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
```diff
@@ -68,9 +68,10 @@ function install_packages() {
...
     strongswan libcharon-extra-plugins libstrongswan-extra-plugins \
-    virt-what open-vm-tools qemu-guest-agent hyperv-daemons
+    virt-what open-vm-tools qemu-guest-agent hyperv-daemons \
+    haveged
```
rhtyd: +1

systemvm/debian/etc/vmware-tools/tools.conf
PaulAngus: Yes, this a bug in the Debian distribution of open-vm-tools.
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=889884

tools/appliance/systemvmtemplate/scripts/configure_systemvm_services.sh
PaulAngus: My recollection is that sshd is actually only started after the firewall rules are established, otherwise there will be a period of vulnerability during startup.

tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
PaulAngus: I believe the use of aria2 was scrapped.

systemvm/debian/opt/cloud/bin/setup/common.sh
PaulAngus: yes we could, i kept the separation in case it was useful in future, (this is why i left the virtualbox case in despite it not required at this time).
I can combine the xen cases if that is preferred.

rhtyd:
Note: Debian 9.7 has been released, build will fail due to incorrect iso url. 
https://www.debian.org/News/2019/20190123

systemvm/debian/etc/cron.daily/clear_cache
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# clear memory cache to ultimately reduce swapping
+
+sync && echo 1 > /proc/sys/vm/drop_caches
```
wido: I understand what you are doing there, but it is the Linux kernel which will remove files from it's case if it needs this memory for something else.

I wouldn't mess with this.

systemvm/debian/etc/cron.hourly/clear_cache
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# clear memory cache to ultimately reduce swapping
+
+phymem=$(free|awk '/^Mem:/{print $2}')
+if [ $phymem -lt 513000 ]; then
+  sync && echo 1 > /proc/sys/vm/drop_caches
```
wido: Same applies here. The kernel is the one who clears the page cache if it needs the memory for something else

systemvm/debian/opt/cloud/bin/setup/secstorage.sh
```diff
@@ -19,14 +19,14 @@
...
-  echo "apache2 cloud nfs-common portmap" > /var/cache/cloud/enabled_svcs
+  echo "cloud apache2 nfs-common portmap" > /var/cache/cloud/enabled_svcs
   echo "conntrackd keepalived haproxy dnsmasq" > /var/cache/cloud/disabled_svcs
   mkdir -p /var/log/cloud
 }
 
 setup_secstorage() {
   log_it "Setting up secondary storage system vm"
-  sysctl vm.min_free_kbytes=8192
+  #sysctl vm.min_free_kbytes=8192
```
wido: Why comment it and not remove it?

systemvm/debian/etc/cron.daily/clear_cache
PaulAngus: As recommended amount of memory for a Debian host is 4GB as we are working **so** far below that, I would say that we are operating well outside 'normal' practices.  Personally, I have long questioned the use of Debian, and would have thought FreeBSD was a better choice.  wrt to memory, the minimum memory for Debian is 96MB.

at 256MB RAM we can't really afford almost ANY caching of disk to memory.  If we started at 4GB then i wouldn't dream of touching this.

The specific, highly replicable issue that is designed to fix, is that the cached disk access is not given back, instead the host starts swapping memory to disk, quickly making the VR unusably due to the CPU usage that this causes (see  https://github.com/apache/cloudstack/issues/3039)

I hope you can see why this sub-optimal fix is proposed to remedy the massively sub-optimal host spec that the VR is trying to operate in.

systemvm/debian/etc/cron.hourly/clear_cache
PaulAngus: As recommended amount of memory for a Debian host is 4GB as we are working **so** far below that, I would say that we are operating well outside 'normal' practices.  Personally, I have long questioned the use of Debian, and would have thought FreeBSD was a better choice.  wrt to memory, the minimum memory for Debian is 96MB.

at 256MB RAM we can't really afford almost ANY caching of disk to memory.  If we started at 4GB then i wouldn't dream of touching this.

The specific, highly replicable issue that is designed to fix, is that the cached disk access is not given back, instead the host starts swapping memory to disk, quickly making the VR unusably due to the CPU usage that this causes (see  https://github.com/apache/cloudstack/issues/3039)

I hope you can see why this sub-optimal fix is proposed to remedy the massively sub-optimal host spec that the VR is trying to operate in.

This *ridiculous* level of outside intervention is only being applied to VRs with 512MB or less available to it, which itself is a ridiculously low amount of RAM to give a Debian host.

I should add that this was 'discovered' and tested through manually monitoring the VR memory usage over time and determining that the disk cache was the 'culprit' for the growing memory footprint.

wido:
I understand @PaulAngus 

I'd really opt that the default VR then gets more memory. The changes you want to make are fine with me, but memory is cheap nowadays, so why don't we bump the default to 1GB.

PaulAngus:
Thanks @wido.  I could (if i was allowed) name at least 3 organisations with 3000+ VRs.  That adds up to a lot of memory and hosts to put them in, plus cooling yadda, yadda, yadda.... 
Maybe we should have a health warning somewhere about running with extremely low memory.

systemvm/debian/etc/cron.daily/clear_cache
andrijapanicsb: This is safe in general, AND we are should not be concerned with disk latency, since we are doing network IO 99% of time.

tools/appliance/systemvmtemplate/scripts/install_systemvm_packages.sh
andrijapanicsb: aria2 Used by KVM hosts with "bypass Sec.Stor." during template download.

Just tested Add template (from remote http) - agent/java download the template... 
"Upload from local" uses POST, so should not be relevant.

rhtyd:
@blueorangutan package

systemvm/debian/opt/cloud/bin/setup/common.sh
```diff
@@ -587,10 +587,10 @@ routing_svcs() {
    echo "cloud nfs-common portmap" > /var/cache/cloud/disabled_svcs
    if [ "$RROUTER" -eq "1" ]
    then
-       echo "keepalived conntrackd" >> /var/cache/cloud/enabled_svcs
+       echo "cloud keepalived conntrackd" >> /var/cache/cloud/enabled_svcs
        echo "dnsmasq" >> /var/cache/cloud/disabled_svcs
    else
-       echo "dnsmasq" >> /var/cache/cloud/enabled_svcs
+       echo "cloud dnsmasq" >> /var/cache/cloud/enabled_svcs
```
rhtyd: cloud service should only run on cpvm, ssvm.

tools/appliance/systemvmtemplate/scripts/configure_systemvm_services.sh
```diff
@@ -108,11 +108,27 @@ function configure_services() {
...
+  systemctl disable hyperv-daemons.hv-vss-daemon.service
+  systemctl disable qemu-guest-agent
 
   configure_apache2
   configure_strongswan
   configure_issue
   configure_cacerts
+
+  # patch known systemd/sshd memory leak - https://github.com/systemd/systemd/issues/8015#issuecomment-476160981
+  echo '@include null' >> /etc/pam.d/systemd-user
```
rhtyd: :+1:

rhtyd:
Per Paul's advise, closing this one in favour of the PR on master branch that brings in enhancements and memory optimisation changes https://github.com/apache/cloudstack/pull/3126

PR #2773, opened 2018-07-26
CLOUDSTACK-10328: Add Secondary IPv6 address through API

GabrielBrascher:
This PR allows the addition of IPv6 address to the secondary NIC through API. It was tested with the basic and the advanced network.

Some use cases are illustrated with the following screenshots:
- the given IP address is checked in the UI :
acquirenewipstringcheck2

- it acquires multiple IP addresses:
acquirenewips
acquirenewipsvmview

- fails to acquire an allocated IP:
acquirenewipsfailduplicated

- releases the IP addresses:
releaseipv6

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## Checklist:

- [x] I have read the [CONTRIBUTING](https://github.com/apache/cloudstack/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
Testing
- [x] I have added tests to cover my changes.
- [ ] All relevant new and existing integration tests have passed.
- [ ] A full integration testsuite with all test that can run on my environment has passed.

PR #3129, opened 2019-01-12
Fix “test_nic_secondaryip_add_remove” test

rafaelweingartner:
## Description

This regression was introduced with PR #2773 (Support IPv6 address in addIpToNic). The contributor did not take into consideration that the method “addIpToNic” was designed to add/allocate other IPs to a NIC. If users did not specify an IP, ACS should generate one for the network where the NIC is plugged into.

Even though I am fixing this regression here, it is still important to highlight that for IPV6, the user is not able to allocate an IP without specifying it.

## Types of changes

## How Has This Been Tested?
Locally

### DIFF
--- a/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
+++ b/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
@@ -198,7 +198,7 @@ import com.cloud.vm.dao.VMInstanceDao;
 /**
  * NetworkServiceImpl implements NetworkService.
  */
-public class NetworkServiceImpl extends ManagerBase implements  NetworkService {
+public class NetworkServiceImpl extends ManagerBase implements NetworkService {
     private static final Logger s_logger = Logger.getLogger(NetworkServiceImpl.class);
 
     private static final long MIN_VLAN_ID = 0L;
@@ -429,8 +429,7 @@ public class NetworkServiceImpl extends ManagerBase implements  NetworkService {
             throw new InvalidParameterException("There are multiple services used ip " + ip.getAddress() + ".");
         }
         if (service != null && !((Service)services.toArray()[0] == service || service.equals(Service.Firewall))) {
-            throw new InvalidParameterException("The IP " + ip.getAddress() + " is already used as " + ((Service)services.toArray()[0]).getName() + " rather than "
-                    + service.getName());
+            throw new InvalidParameterException("The IP " + ip.getAddress() + " is already used as " + ((Service)services.toArray()[0]).getName() + " rather than " + service.getName());
         }
         return true;
     }
@@ -519,8 +518,8 @@ public class NetworkServiceImpl extends ManagerBase implements  NetworkService {
 
     @Override
     @ActionEvent(eventType = EventTypes.EVENT_NET_IP_ASSIGN, eventDescription = "allocating Ip", create = true)
-    public IpAddress allocateIP(Account ipOwner, long zoneId, Long networkId, Boolean displayIp) throws ResourceAllocationException, InsufficientAddressCapacityException,
-            ConcurrentOperationException {
+    public IpAddress allocateIP(Account ipOwner, long zoneId, Long networkId, Boolean displayIp)
+            throws ResourceAllocationException, InsufficientAddressCapacityException, ConcurrentOperationException {
 
         Account caller = CallContext.current().getCallingAccount();
         long callerUserId = CallContext.current().getCallingUserId();
@@ -559,8 +558,8 @@ public class NetworkServiceImpl extends ManagerBase implements  NetworkService {
 
     @Override
     @ActionEvent(eventType = EventTypes.EVENT_PORTABLE_IP_ASSIGN, eventDescription = "allocating portable public Ip", create = true)
-    public IpAddress allocatePortableIP(Account ipOwner, int regionId, Long zoneId, Long networkId, Long vpcId) throws ResourceAllocationException,
-            InsufficientAddressCapacityException, ConcurrentOperationException {
+    public IpAddress allocatePortableIP(Account ipOwner, int regionId, Long zoneId, Long networkId, Long vpcId)
+            throws ResourceAllocationException, InsufficientAddressCapacityException, ConcurrentOperationException {
         Account caller = CallContext.current().getCallingAccount();
         long callerUserId = CallContext.current().getCallingUserId();
         DataCenter zone = _entityMgr.findById(DataCenter.class, zoneId);
@@ -613,7 +612,7 @@ public class NetworkServiceImpl extends ManagerBase implements  NetworkService {
         try {
             return releaseIpAddressInternal(ipAddressId);
         } catch (Exception e) {
-           return false;
+            return false;
         }
     }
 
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

rhtyd:
Rekick travis

rhtyd:
@blueorangutan test

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -711,15 +710,11 @@ public NicSecondaryIp allocateSecondaryGuestIP(final long nicId, IpAddresses req
...
-            try {
-                if (ipv4Address != null) {
-                    ipaddr = _ipAddrMgr.allocateGuestIP(network, ipv4Address);
-                }
-                if (ipv6Address != null) {
-                    ip6addr = ipv6AddrMgr.allocateGuestIpv6(network, ipv6Address);
-                }
-            } catch (InsufficientAddressCapacityException e) {
-                throw new InvalidParameterValueException("Allocating guest ip for nic failed");
+            if ((ipv4Address != null || NetUtils.isIpv4(network.getGateway()) && org.apache.commons.lang3.StringUtils.isBlank(ipv6Address))) {
```
rhtyd: This would imply that ipv4 address is allocated only when ipv6 is not passed. Do you think this may cause issue for ipv6 users /cc @wido ?

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -784,8 +779,7 @@ public Long doInTransaction(TransactionStatus status) {
 
     @Override
     @DB
-    @ActionEvent(eventType = EventTypes.EVENT_NIC_SECONDARY_IP_UNASSIGN, eventDescription = "Removing secondary ip " +
-            "from nic", async = true)
+    @ActionEvent(eventType = EventTypes.EVENT_NIC_SECONDARY_IP_UNASSIGN, eventDescription = "Removing secondary ip " + "from nic", async = true)
```
rhtyd: Remove concatenation of static strings

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -1247,11 +1240,10 @@ public Network createGuestNetwork(CreateNetworkCmd cmd) throws InsufficientCapac
...
 
         // Regular user can create Guest Isolated Source Nat enabled network only
-        if (_accountMgr.isNormalUser(caller.getId())
-                && (ntwkOff.getTrafficType() != TrafficType.Guest || ntwkOff.getGuestType() != Network.GuestType.Isolated
-                        && areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SourceNat))) {
-            throw new InvalidParameterValueException("Regular user can create a network only from the network" + " offering having traffic type " + TrafficType.Guest
-                    + " and network type " + Network.GuestType.Isolated + " with a service " + Service.SourceNat.getName() + " enabled");
+        if (_accountMgr.isNormalUser(caller.getId()) && (ntwkOff.getTrafficType() != TrafficType.Guest
+                || ntwkOff.getGuestType() != Network.GuestType.Isolated && areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SourceNat))) {
+            throw new InvalidParameterValueException("Regular user can create a network only from the network" + " offering having traffic type " + TrafficType.Guest + " and network type "
```
rhtyd: Remove concatenation of static strings.

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -1289,17 +1281,16 @@ public Network createGuestNetwork(CreateNetworkCmd cmd) throws InsufficientCapac
...
             }
         }
 
         // Vlan is created in 1 cases - works in Advance zone only:
         // 1) GuestType is Shared
-        boolean createVlan = (startIP != null && endIP != null && zone.getNetworkType() == NetworkType.Advanced
-                && ((ntwkOff.getGuestType() == Network.GuestType.Shared)
-                || (ntwkOff.getGuestType() == GuestType.Isolated &&
-                !areServicesSupportedByNetworkOffering(ntwkOff.getId(), Service.SourceNat))));
+        boolean createVlan = (startIP != null && endIP != null && zone.getNetworkType() == NetworkType.Advanced && ((ntwkOff.getGuestType() == Network.GuestType.Shared)
```
rhtyd: @rafaelweingartner What is the guideline of breaking lines? In many places line breaks were made when the length of the line was long, and here, for example, the final result makes the line length from short to long.

rhtyd:
@rafaelweingartner I've left few comments, will try to complete my review later this week (sorry got busy had to pull back in between the review just now).

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
rafaelweingartner: No, it would not. The user can call the method to ask an IPv4 (is the network supports). And then, in a second call, he/she can ask for an IPv6 as well. The only difference now is that for IPv6 an IP must be defined by the user. ACS will not generate one.

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
rafaelweingartner: Fixed.

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
rafaelweingartner: Fixed.

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
rafaelweingartner: That is a good question. I am using this code formatter here [1], which is available in our Wiki.
I did not take much attention on these details. I was expecting the formatter there to be the right one for us to use. The formatter is presented in this page here [2].

Would you like me to change some of the changes that were automatically applied by our formatter [1]?

[1] https://cwiki.apache.org/confluence/download/attachments/29687985/ApacheCloudStack.xml?version=1&modificationDate=1382960469000&api=v2
[2] https://cwiki.apache.org/confluence/display/CLOUDSTACK/Coding+conventions

rafaelweingartner:
Thanks for the review. I have addressed the String concatenations that you noticed, and also the questions regarding IPv4 and IPv6, and the code formatter.

thanks for the feedback here ;)

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
GabrielBrascher: That is the idea indeed @rafaelweingartner @rhtyd.
It was designed to allow allocating ipv4 or ipv6 (does not accept null ipv6). If a user wants to allocate both, IPv4 and IPv6, he/she needs to call the API one time per IP address.
My mistake here was adding a null verification for ipv4 addresses as well.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
@blueorangutan test

PR #3130, opened 2019-01-14
Description field option is needed for networking NAT, ACL and Security Group rules.

raiciks:
## Description
Hello!
Many of our customers asking for description field option in networking NAT, ACL rules. Some of our customers have huge list of rules, and they need strictly to know who create a rule, for what service rule is created and so on. This option is required not only for system administrators, but it is requested by security managers and GDPR. This issue is affected by: Guest Networks, VPC networks - ACL list of rules and Security Groups. I prepared screen shots where this option is necessary.  

## Types of changes

I prepared screen shots where this option is necessary.  
guest_networks


rafaelweingartner:
Something like this https://github.com/apache/cloudstack/pull/2475?

raiciks:
Yes! But this feature we need in Guest Networks, ACL and Security Groups.

rafaelweingartner:
I guess they all use the same ACL endpoint. Therefore, it is a matter of extending the Javascript that is used by these functions.

rafaelweingartner:
I will close this one in favor of #3131

PR #3131, opened 2019-01-14
Working with ACL List Rules in VPC networks

raiciks:
## Description
Hello!
From our customers we receive questions about some inconvenience working with ACL List of Rules, for example:
1) It is necessary that the names of fields could be active for sorting - Ascending / Descending functionality;
2) It is necessary for customers to offer filtering of Traffic Type. Customer would be very happy if they can choose to view only Ingress or Egress rules.
3) If customer have huge list of ACL rules, than bottom scrollbar is not visible if you are in top or middle of ACL rules list. This is troublesome in daily administrating tasks.


rafaelweingartner:
Something like this #2475?

I guess the Javascripts for these different functions all use the same API method. Therefore, it is a matter of extending the Javascript that is used by these functions.

rafaelweingartner:
@raiciks this PR is quite odd. The commits are from 2013!
Moreover, these features were already addressed. I suspect this is coming from CloudPlatform from Accelerite. 


>  It is necessary that the names of fields could be active for sorting - Ascending / Descending functionality;
Rules are ordered by `#Rule`. Therefore, order method for sorting might confuse users in this screen. 

> It is necessary for customers to offer filtering of Traffic Type. Customer would be very happy if they can choose to view only Ingress or Egress rules.
That is a great idea, but I am not seeing any Javascript files being changed here. We only need to add some Javascript code to add this feature.

> If customer have huge list of ACL rules, than bottom scrollbar is not visible if you are in top or middle of ACL rules list. This is troublesome in daily administrating tasks.
This has already been fixed with PR https://github.com/apache/cloudstack/pull/2456

I will close this PR, as it does not seem to bring new feature or values. It does provide an interesting idea regarding ACL items being listed, but I am not seeing any code to enable such feature.

PR #3134, opened 2019-01-16
keystore: don't restart systemvm cloud.service post cert import

rhtyd:
This ensures that the systemvm agent (cloud.service) is not restarted
on certificate import. The agent has an inbuilt logic to attempt reconnection.
If the old certificates/keystore is invalid agent will attempt reconnection.

## Types of changes

## How Has This Been Tested?

Deployed a CPU-bound environment (hypervisor host with limited CPU and 1core), saw that previously irrespective of keystore setup the restart logic would cause agent to restart (sig.kill is the reason seen in logs) and on management server side it would seem agent connects, then disconnects, then connects and is finally in `Up` state. With this fix, the reconnection attempt won't be observed on the management server side. The agent will keep attempting reconnection when certificates are invalid, the cert import script does not need to restart cloud service manually.

### DIFF
--- a/scripts/util/keystore-cert-import
+++ b/scripts/util/keystore-cert-import
@@ -41,6 +41,12 @@ fi
 # Use a new keystore file
 NEW_KS_FILE="$KS_FILE.new"
 
+# Check/store old KS state
+OLD_KS_FILE_EXISTS=false
+if [ -f $KS_FILE ]; then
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

rhtyd:
Tests LGTM.

PR #3136, opened 2019-01-16
ipv6: Do not allow Secondary IPv6 addresses to be EUI-64

wido:
## Description
When requesting a secondary IPv6 address the address specified by the requester should not be a EUI-64 (SLAAC) address.

This can be seen by looking at *ff:fe* which is always in the same position in a IPv6 address when using SLAAC/EUI-64.

<pre>2001:db8:13f:0:1c00:4aff:fe00:fe</pre>

That is an IPv6 EUI-64 address for example.

See: https://community.cisco.com/t5/networking-documents/understanding-ipv6-eui-64-bit-address/ta-p/3116953

  Extended Unique Identifier (EUI), as per RFC2373, allows a host to assign iteslf a unique 64-Bit IP 
  Version 6 interface identifier (EUI-64). This feature is a key benefit over IPv4 as it eliminates the need of manual configuration or DHCP as in the world of IPv4. The IPv6 EUI-64 format address is obtained 
  through the 48-bit MAC address. The MAC address is first separated into two 24-bits, with one being 
  OUI (Organizationally Unique Identifier) and the other being NIC specific. The 16-bit 0xFFFE is then 
  inserted between these two 24-bits for the 64-bit EUI address. IEEE has chosen FFFE as a reserved 
  value which can only appear in EUI-64 generated from the an EUI-48 MAC address.

image

## Types of changes

## How Has This Been Tested?
Tested this locally

### DIFF
--- a/server/src/main/java/com/cloud/network/Ipv6AddressManagerImpl.java
+++ b/server/src/main/java/com/cloud/network/Ipv6AddressManagerImpl.java
@@ -104,6 +104,11 @@ public class Ipv6AddressManagerImpl extends ManagerBase implements Ipv6AddressMa
                     network.getDataCenterId());
         }
 
+        if (NetUtils.isIPv6EUI64(requestedIpv6)) {
+            throw new InsufficientAddressCapacityException(String.format("Requested IPv6 address [%s] may not be a EUI-64 address", requestedIpv6), DataCenter.class,
+                    network.getDataCenterId());
+        }
+
         checkIfCanAllocateIpv6Address(network, requestedIpv6);
 
         IpAddresses requestedIpPair = new IpAddresses(null, requestedIpv6);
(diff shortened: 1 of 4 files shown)


wido:
@rafaelweingartner This is related to #3077

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

rafaelweingartner:
> @rafaelweingartner This is related to #3077

Thanks @wido. I did not reply in #3077 because I did not have time before, but I talked with @GabrielBrascher, and afterwards we proceeded with the merge there.

wido:
I have added some documentation to the code. How does this look?

rafaelweingartner:
@blueorangutan package

PR #3154, opened 2019-01-28
Copy template to target KVM host if needed when migrating local <> local storage

GabrielBrascher:
PR #2997 allowed KVM VM live migration with ROOT volume on local storage (filesystem), if the template is already on the target host, the migration will be successful; however, if the VM template is not on the target host the migration fails.

This PR fixes the bug introduced by #2997, copying the VM template to the target host if needed.

## Types of changes

## How Has This Been Tested?

- [x] Tested migrating a VM to a host that did not download the template to its primary storage. The added code is then executed, copying the template to the target host/primary storage; after migrating the template, the execution flows keep as it was previously and the migration is succesfull.
Additionally, table `template_spool_ref` is correctly updated with the template reference matching with the target storage pool.
- [x] Successfully migrated a VM in the case where both storage pools (source and destiny host) have the VM template.

### DIFF
--- a/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
+++ b/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
@@ -101,6 +101,9 @@ public interface StorageManager extends StorageService {
             ConfigKey.Scope.Cluster,
             null);
 
+    ConfigKey<Integer> PRIMARY_STORAGE_DOWNLOAD_WAIT = new ConfigKey<Integer>("Storage", Integer.class, "primary.storage.download.wait", "10800",
+            "In second, timeout for download template to primary storage", false);
+
     /**
      * Returns a comma separated list of tags for the specified storage pool
      * @param poolId
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -29,7 +29,12 @@ import org.apache.cloudstack.engine.subsystem.api.storage.StrategyPriority;
 import org.apache.cloudstack.engine.subsystem.api.storage.TemplateDataFactory;
 import org.apache.cloudstack.engine.subsystem.api.storage.TemplateInfo;
 import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
+import org.apache.cloudstack.storage.command.CopyCommand;
+import org.apache.cloudstack.storage.datastore.DataStoreManagerImpl;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
+import org.apache.cloudstack.storage.to.TemplateObjectTO;
+import org.apache.commons.lang.StringUtils;
+import org.apache.log4j.Logger;
 
 import com.cloud.agent.api.Answer;
 import com.cloud.agent.api.MigrateCommand;
@@ -37,14 +42,21 @@ import com.cloud.agent.api.MigrateCommand.MigrateDiskInfo;
 import com.cloud.agent.api.storage.CreateAnswer;
 import com.cloud.agent.api.storage.CreateCommand;
 import com.cloud.agent.api.to.VirtualMachineTO;
+import com.cloud.exception.AgentUnavailableException;
+import com.cloud.exception.OperationTimedoutException;
 import com.cloud.host.Host;
 import com.cloud.hypervisor.Hypervisor.HypervisorType;
 import com.cloud.storage.DataStoreRole;
 import com.cloud.storage.DiskOfferingVO;
 import com.cloud.storage.Storage.StoragePoolType;
+import com.cloud.storage.StorageManager;
+import com.cloud.storage.StoragePool;
+import com.cloud.storage.VMTemplateStoragePoolVO;
 import com.cloud.storage.VolumeVO;
+import com.cloud.storage.dao.VMTemplatePoolDao;
 import com.cloud.utils.exception.CloudRuntimeException;
 import com.cloud.vm.DiskProfile;
+import com.cloud.vm.VirtualMachineManager;
 
 /**
  * Extends {@link StorageSystemDataMotionStrategy}, allowing KVM hosts to migrate VMs with the ROOT volume on a non managed local storage pool.
@@ -54,6 +66,14 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
 
     @Inject
     private TemplateDataFactory templateDataFactory;
+    @Inject
+    private VMTemplatePoolDao vmTemplatePoolDao;
+    @Inject
+    private DataStoreManagerImpl dataStoreManagerImpl;
+    @Inject
+    private VirtualMachineManager virtualMachineManager;
+
+    private static final Logger LOGGER = Logger.getLogger(KvmNonManagedStorageDataMotionStrategy.class);
 
     /**
      * Uses the canHandle from the Super class {@link StorageSystemDataMotionStrategy}. If the storage pool is of file and the internalCanHandle from {@link StorageSystemDataMotionStrategy} CANT_HANDLE, returns the StrategyPriority.HYPERVISOR strategy priority. otherwise returns CANT_HANDLE.
@@ -95,7 +115,7 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
         String templateUuid = getTemplateUuid(destVolumeInfo.getTemplateId());
         CreateCommand rootImageProvisioningCommand = new CreateCommand(diskProfile, templateUuid, destStoragePool, true);
 
-        Answer rootImageProvisioningAnswer = _agentMgr.easySend(destHost.getId(), rootImageProvisioningCommand);
+        Answer rootImageProvisioningAnswer = agentManager.easySend(destHost.getId(), rootImageProvisioningCommand);
 
         if (rootImageProvisioningAnswer == null) {
             throw new CloudRuntimeException(String.format("Migration with storage of vm [%s] failed while provisioning root image", vmTO.getName()));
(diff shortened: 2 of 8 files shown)


GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@dhlaluku I could not add you on the requested reviewers list then I am pinging here ;)

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -140,4 +160,56 @@ protected void setVolumePath(VolumeVO volume) {
     protected boolean shouldMigrateVolume(StoragePoolVO sourceStoragePool, Host destHost, StoragePoolVO destStoragePool) {
         return sourceStoragePool.getPoolType() == StoragePoolType.Filesystem;
     }
+
+    /**
+     * If the template is not on the target primary storage then it migrates the template.
+     * Otherwise the execution flow fails during migration and creates a null file created on the target storage pool.
```
DaanHoogland: this comments confuses me. you wrote in the description that the template would be copied and this problem would no longer be there. if that is true this "Otherwise ..." is not true or badly formulated. Can you explain?
As it is written now it seems that the condition after "Otherwise" can still occur.
If I understand this correctly you should start with "If we don't do this ...", or "This is needed to prevent ...". Am I correct?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -140,4 +160,56 @@ protected void setVolumePath(VolumeVO volume) {
...
+    /**
+     * Logs in debug mode the copy command failure if the CopyCommand Answer has result as false.
+     */
+    protected void logInCaseOfTemplateCopyFailure(Answer copyCommandAnswer, TemplateObjectTO sourceTemplate, DataStore destDataStore) {
+        if (copyCommandAnswer != null && !copyCommandAnswer.getResult()) {
+            String failureDetails = StringUtils.EMPTY;
+            if (copyCommandAnswer.getDetails() != null) {
+                failureDetails = "; details: " + copyCommandAnswer.getDetails();
+            }
+            LOGGER.debug(String.format("Failed to copy template [id=%s, name=%s] to the primary storage pool [id=%s]%s", sourceTemplate.getId(), sourceTemplate.getName(),
```
DaanHoogland: seems like an 'ERROR' to me

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -140,4 +160,56 @@ protected void setVolumePath(VolumeVO volume) {
...
+     * Sends the CopyCommand to migrate the template to the dest host.
+     */
+    protected void sendCopyCommand(Host destHost, TemplateObjectTO sourceTemplate, TemplateObjectTO destTemplate, DataStore destDataStore) {
+        boolean executeInSequence = virtualMachineManager.getExecuteInSequence(HypervisorType.KVM);
+        CopyCommand copyCommand = new CopyCommand(sourceTemplate, destTemplate, StorageManager.PRIMARY_STORAGE_DOWNLOAD_WAIT.value(), executeInSequence);
+        try {
+            Answer copyCommandAnswer = agentManager.send(destHost.getId(), copyCommand);
+            logInCaseOfTemplateCopyFailure(copyCommandAnswer, sourceTemplate, destDataStore);
+        } catch (AgentUnavailableException | OperationTimedoutException e) {
+            throw new CloudRuntimeException(String.format("Failed to copy template [id=%s, name=%s] to the primary storage pool [id=%s].", sourceTemplate.getId(),
```
DaanHoogland: why is this string constructed twice, here and in below logInCaseOfTemplateCopyFailure()?
they seem to describe the exact same thing

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -1859,14 +1861,21 @@ protected void setVolumePath(VolumeVO volume) {
         volume.setPath(volume.get_iScsiName());
     }
 
+    /**
+     * For this strategy it is not necessary to check and migrate the template; however, classes that extend this one may need to check if the template is on the target storage pool (and if not then migrate) before migrating the VM.
```
DaanHoogland: can you try to wrap this to fit in a github window?

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -1859,14 +1861,21 @@ protected void setVolumePath(VolumeVO volume) {
...
+     */
+    protected void migrateTemplateToTargetFilesystemStorageIfNeeded(VolumeInfo srcVolumeInfo, DataStore destDataStore, StoragePool destStoragePool, Host destHost) {
+        // This method is used by classes that extend this one
```
DaanHoogland: I am wary of this pattern. Can you explain why you implemented it this way? It seems you are implementing a pattern for genericity while already knowing it is only applicable for localstorage.

engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
```diff
@@ -101,6 +101,9 @@
             ConfigKey.Scope.Cluster,
             null);
 
+    ConfigKey<Integer> PRIMARY_STORAGE_DOWNLOAD_WAIT = new ConfigKey<Integer>("Storage", Integer.class, "primary.storage.download.wait", "10800",
```
DaanHoogland: isn't such a variable already in place somewhere? there are other places where it would have use. If not, I think it should be applied to the initial deploy equally.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
GabrielBrascher: @DaanHoogland The class KvmNonManagedStorageDataMotionStrategy was implemented extending the StorageSystemDataMotionStrategy and adding methods such as this one in the superclass execution flow. This was the best that I could think regarding using the current architecture and easying future maintenance and extensions. Ideally, we should refactor this class structure and use proper O.O. However, that was not a simple task, and as such I think it is best executed separated from any new feature introduction.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: You are right @DaanHoogland. I will remove the `Otherwise the execution flow fails during migration and creates a null file created on the target storage pool.` from the Javadoc and leave just the `If the template is not on the target primary storage then it migrates the template.`. Thanks for pointing it!

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: I will update this to 'ERROR' I will also extract it to a method that returns the message, avoiding duplicated code.

engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
GabrielBrascher: `Config.PrimaryStorageDownloadWait` has this configuration indeed.

I can pick from there then and removed this one. However, on the other hand, `Config.PrimaryStorageDownloadWait` is deprecated. Should I use `Config.PrimaryStorageDownloadWait`, or remove the `Config.PrimaryStorageDownloadWait` and keep with the `new ConfigKey<Integer>...`?

GabrielBrascher:
@DaanHoogland I will ping when updating the code. Thanks for the review!

engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
DaanHoogland: I'd say keep your new one and find usages of the old one to unify them. If need be move the key to a more centralised place.

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: Extracted to a method :+1:

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: Fixed!

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
GabrielBrascher: Done!

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
Thanks for the review, DaanHoogland! Are you ok with the updated code addressing your requests?

GabrielBrascher:
@blueorangutan test

PR #3170, opened 2019-02-08
NotImplemented as a local exception

DaanHoogland:
## Description

NotImplementedException from sun.reflect.generics.reflectiveObjects is not standard/guarateed by the jvm so a custom implementation is in order.

## Types of changes

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
+++ b/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
@@ -20,11 +20,10 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
 
-import sun.reflect.generics.reflectiveObjects.NotImplementedException;
-
 import com.cloud.storage.Storage;
 import com.cloud.utils.script.OutputInterpreter;
 import com.cloud.utils.script.Script;
+import org.apache.commons.lang.NotImplementedException;
 
 public class QemuImg {
 
@@ -72,7 +71,7 @@ public class QemuImg {
             case FAT:
                 return PreallocationType.Full;
             default:
-                throw new NotImplementedException();
+                throw new NotImplementedException(String.format("type %s not defined as member-value of PreallocationType", provisioningType));
             }
         }
     }


DaanHoogland:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
```diff
@@ -72,7 +70,7 @@ public static PreallocationType getPreallocationType(final Storage.ProvisioningT
             case FAT:
                 return PreallocationType.Full;
             default:
-                throw new NotImplementedException();
+                throw new NotImplementedProvisioningTypeException();
```
rafaelweingartner: What if instead, you throw a cloud runtime with a message describing the problem/error here?

DaanHoogland: That is certainly an option, @rafaelweingartner . I had discussed with @nvazquez and he suggested the commons-lang NotImplementedException as it is loaded through the storage module anyhow. Thoughts?

rafaelweingartner: It also works. I only think that we need to add something in the exception. I mean, a message saying what is not implemented (the preallocation type selected/sent by the user).

borisstoyanov:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
DaanHoogland: check

nvazquez: Agree :+1:

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
```diff
@@ -399,4 +397,9 @@ public void resize(final QemuImgFile file, final long size, final boolean delta)
     public void resize(final QemuImgFile file, final long size) throws QemuImgException {
         this.resize(file, size, false);
     }
+
+    static class NotImplementedProvisioningTypeException extends RuntimeException {
+    }
 }
+
+;
```
nvazquez: Can this be removed please?

DaanHoogland: of course, sorry

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
```diff
@@ -72,7 +71,7 @@ public static PreallocationType getPreallocationType(final Storage.ProvisioningT
             case FAT:
                 return PreallocationType.Full;
             default:
-                throw new NotImplementedException();
+                throw new NotImplementedProvisioningTypeException(String.format("type {} not defined as member-value of PreallocationType", provisioningType));
```
rafaelweingartner: Is this a new type of `String.format` use? I am missing a `%s` there.

DaanHoogland: hm, i might be mixing python and java, will check.

DaanHoogland:
@blueorangutan package

rhtyd:
Let's use org.apache.commons.lang.NotImplementedException which is already used by some storage component?

nathanejohnson:
LGTM

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

Issue #3041, opened 2018-11-21
Decision making logic of which Secondary storage to be used


dhlaluku:
Related to this unresolved issue https://issues.apache.org/jira/browse/CLOUDSTACK-2919
Add proper check for disk space on secondary storage to be used for volume migration across clusters. The current logic does not validate if volume will fit in the destination storage pool, but only checks if whether secondary storage disk utilization has not reached 90%. 

##### ISSUE TYPE

 * Improvement Request
 * Enhancement Request

##### COMPONENT NAME

~~~
Storage
~~~

##### CLOUDSTACK VERSION

~~~
4.9+
~~~

##### CONFIGURATION

2+ Secondary Storage (ImageStore) pools with Zone scope
2+ Clusters 

##### OS / ENVIRONMENT

N/A

##### SUMMARY

When performing cross-cluster volume migration in multi ImageStore environments within the same Zone. The volume is first copied the secondary storage, attached to a worker VM and then migrated to the destination storage pool. This is procedure is only successful when the following 2 conditions are satisfied
1. The ImageStore disk usage should not exceed a hardcoded value of 0.9 (90 %)
2. The size of the volume the volume to be migrated should exist the amount of free space available in the ImageStore.
![image](https://user-images.githubusercontent.com/19572501/48846435-d09be880-eda7-11e8-8c5d-f30bb558d38b.png)

##### STEPS TO REPRODUCE

~~~
- Fill up secondary storage disk up to 90%
- Perform cross-cluster volume migration through UI or API
~~~

##### EXPECTED RESULTS

~~~
An exception should be thrown explaining that no suitable secondary storage pool was found for volume migration.
~~~

##### ACTUAL RESULTS

~~~
CloudStack selects a random secondary storage pool and attempts to migrate the volume anyways and fails when the size of the volume exceeds the amount of free space within that particular secondary storage.
~~~

dhlaluku:
Throw 'InsufficientCapacityException' when migration fails due to insufficient space in secondary storage

rafaelweingartner:
Any chance on getting this into 4.12.0? Otherwise, I would move it to 5.0.0

GabrielBrascher:
@dhlaluku any update on this one?

dhlaluku:
@GabrielBrascher @rafaelweingartner I will pick this up some time next week.

kioie:
@dhlaluku this one?

dhlaluku:
@kioe yes

[...]

PR #3184, opened 2019-02-16
Fixed Decision Making Logic when selecting which secondary storage to use from a selection

kioie:
## Description

This is in reference to https://github.com/apache/cloudstack/issues/3041 . I removed the `Collection.shuffle()` and replaced it with `Collection.sort()`, the idea being that instead of introducing the size of volume as a query parameter, we can just iterate through an already sorted list, based on available size per image store. We iterate through the sorted list, querying if disk usage per image store has exceeded 90%. This eliminates the possibility of there being another location in the zone with more disk space and not being selected as referenced in this issue ( https://issues.apache.org/jira/browse/CLOUDSTACK-2919 ). Also included in this commit is the method `imageStoreCurrentCapacity` that checks remaining capacity per imagestore.

Signed-off-by: @kioie <kioieddy@google.com>

## Types of changes

## How Has This Been Tested?

Tested on this environment: 
3 Secondary Storages (ImageStore) pools with Zone scope
2 Clusters

### DIFF
--- a/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
+++ b/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
@@ -18,18 +18,16 @@
  */
 package org.apache.cloudstack.storage.image.manager;
 
+import javax.annotation.PostConstruct;
+import javax.inject.Inject;
+
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-
-import javax.annotation.PostConstruct;
-import javax.inject.Inject;
-
-import org.apache.log4j.Logger;
-import org.springframework.stereotype.Component;
 
 import org.apache.cloudstack.engine.subsystem.api.storage.DataStore;
 import org.apache.cloudstack.engine.subsystem.api.storage.DataStoreProviderManager;
@@ -42,6 +40,8 @@ import org.apache.cloudstack.storage.image.ImageStoreDriver;
 import org.apache.cloudstack.storage.image.datastore.ImageStoreEntity;
 import org.apache.cloudstack.storage.image.datastore.ImageStoreProviderManager;
 import org.apache.cloudstack.storage.image.store.ImageStoreImpl;
+import org.apache.log4j.Logger;
+import org.springframework.stereotype.Component;
 
 import com.cloud.server.StatsCollector;
 import com.cloud.storage.ScopeType;
@@ -146,7 +146,13 @@ public class ImageStoreProviderManagerImpl implements ImageStoreProviderManager 
     @Override
     public DataStore getImageStore(List<DataStore> imageStores) {
         if (imageStores.size() > 1) {
-            Collections.shuffle(imageStores); // Randomize image store list.
+            Collections.sort(imageStores, new Comparator<DataStore>() {
+                @Override
+                public int compare(DataStore o1, DataStore o2) {
+                    return Long.compare(_statsCollector.imageStoreCurrentCapacity(o1),
+                            _statsCollector.imageStoreCurrentCapacity(o2));
+                }
+            });
             Iterator<DataStore> i = imageStores.iterator();
             DataStore imageStore = null;
             while(i.hasNext()) {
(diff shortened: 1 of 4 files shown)


engine/storage/image/src/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
```diff
@@ -146,7 +144,14 @@ public ImageStoreEntity getImageStore(String uuid) {
     @Override
     public DataStore getImageStore(List<DataStore> imageStores) {
         if (imageStores.size() > 1) {
-            Collections.shuffle(imageStores); // Randomize image store list.
+            //Collections.shuffle(imageStores); // Randomize image store list.
```
sateesh-chodapuneedi: Please remove commented code.

engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
```diff
@@ -146,7 +146,13 @@ public ImageStoreEntity getImageStore(String uuid) {
     @Override
     public DataStore getImageStore(List<DataStore> imageStores) {
         if (imageStores.size() > 1) {
-            Collections.shuffle(imageStores); // Randomize image store list.
+            Collections.sort(imageStores, new Comparator<DataStore>() {
+                @Override
```
rhtyd: This still does not cleanly solve the design issue. The `getImageStore` may be used in several places and simply randomized return (whether shuffled or sorted by capacity) does not make much difference. Instead, a check may be performed preferably via a different method that checks if image store has enough capacity, that check method may be used either by this method or the consumers of `getImageStore`.

engine/storage/image/src/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
```diff
@@ -0,0 +1,168 @@
...
+import org.apache.cloudstack.storage.image.store.ImageStoreImpl;
+import org.apache.log4j.Logger;
+import org.springframework.stereotype.Component;
+
+import com.cloud.server.StatsCollector;
+import com.cloud.storage.ScopeType;
+import com.cloud.storage.dao.VMTemplateDao;
+
+@Component
+public class ImageStoreProviderManagerImpl implements ImageStoreProviderManager {
```
rhtyd: This won't compile/be used because the class is not under the standard maven structure (i.e. under `src/main/java`). Moreover, `@Component` based scanning is not used in CloudStack, beans need to be declared.

server/src/com/cloud/server/StatsCollector.java
```diff
@@ -0,0 +1,1342 @@
...
+import com.cloud.vm.dao.NicDao;
+import com.cloud.vm.dao.UserVmDao;
+import com.cloud.vm.dao.VMInstanceDao;
+
+/**
+ * Provides real time stats for various agent resources up to x seconds
+ *
+ */
+@Component
+public class StatsCollector extends ManagerBase implements ComponentMethodInterceptable, Configurable {
```
rhtyd: Remove this, and use existing class under `src/main/java...` path for any changes.

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -1364,6 +1364,12 @@ public boolean imageStoreHasEnoughCapacity(DataStore imageStore) {
         return false;
     }
 
+    public Long imageStoreCurrentCapacity(DataStore imageStore) {
```
rhtyd: The name is misleading, moreover `imageStoreHasEnoughCapacity` can be used instead.

engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
kioie: The idea behind Sorting by capacity was that the imagestore with the most available capacity will always be on the top of the list hence selected first.
I thought about checking if imagestore has enough capacity but that would mean that I compare the image size vs remaining capacity of the image store, that of course would mean that I introduce another parameter to this `getImageStore` method, perhaps Image id or something, to check for size of image. Not sure how this will go down with the consumers of `getImageStore`.

rhtyd:
Thank you for your PR. But it's incomplete and given we're closer to the freeze, I'll take over and submit issues around secondary storage selection affecting:
https://github.com/apache/cloudstack/issues/3041
https://github.com/apache/cloudstack/issues/3287
https://github.com/apache/cloudstack/issues/3478

kioie:
No worries

PR #3186, opened 2019-02-19
Add possibility to set MTU size for NIC

gmueller-ewerk:
## Description
This Pull Request enables the user to edit the MTU size for each NIC. the given value will be added to customize the MTU interface (see libvirt XML - https://libvirt.org/formatnetwork.html#elementsConnect)
The MTU of each individual NIC is editable in the interface by

- adding a network to the VM
bildschirmfoto 2019-02-19 um 14 12 34

- updating a network to the VM
bildschirmfoto 2019-02-19 um 14 13 41

## Types of changes

- [ x] Enhancement (improves an existing feature and functionality)


svenvogel:
@gmueller-ewerk great!

svenvogel:
@GabrielBrascher @rafaelweingartner @rhtyd can you review it? thanks

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -1191,6 +1196,9 @@ public String toString() {
             if (_networkName != null) {
                 netBuilder.append("<target dev='" + _networkName + "'/>\n");
             }
+            if (_mtu > 0 && _mtu <= 9000) {
```
wido: I see you silently skip the MTU if it's equal to 9000, but some networks also allow 9216 as that's what the switch allows.

I'm also not seeing why we should allow an MTU of 10 or something. Shouldn't 1280 be the minimum?

And shouldn't the API do this check already?

DennisKonrad:
@wido @gmueller-ewerk @svenvogel 
I also think we should check the min/max values in the frontend (WebUI) and backend (java/API).

Also the idea is to manually set the MTU to activate the feature for the specific NIC. This is to not break existing functionality. At the moment the MTU is implicitely set through the nic type and one doesn't know what value is.

So the feature here allows to override the active configuration without forcing an default value onto other vms. This is to allow a soft migration of existing CS installations.

wido:
@DennisKonrad Good and valid point. The MTU should only be set if specified. Otherwise we should not define it in the XML and leave it up to libvirt to decide.

So if MTU !=0 or NULL we should add the element to the XML, otherwise just leave it out and let libvirt decide what to do.

@gmueller-ewerk Can you address these comments?

gmueller-ewerk:
Yes, I will have a look at the issues

svenvogel:
@wido can you check again?
@rhtyd @borisstoyanov @rafaelweingartner can you check this?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
```diff
@@ -203,6 +202,12 @@ public boolean parseDomainXML(String domXML) {
                 String model = getAttrValue("model", "type", nic);
                 String slot = StringUtils.removeStart(getAttrValue("address", "slot", nic), "0x");
 
+                String mtuAttrValue = getAttrValue("mtu", "size", nic);
+                int mtu = 1500;
+                if (mtuAttrValue != null) {
+                    mtu = Integer.parseInt(mtuAttrValue);
+                }
```
wido: So we always set the MTU to 1500 if it's not set. Shouldn't we just ignore the MTU if it's not set?

Leave it to the libvirt default unless somebody specifies it

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -1191,6 +1197,9 @@ public String toString() {
             if (_networkName != null) {
                 netBuilder.append("<target dev='" + _networkName + "'/>\n");
             }
+            if (NetUtils.isValidMtu(_mtu)) {
```
wido: As we check it here we can just pass a null when the MTU is not set. No need to hardcode 1500 in the code.

utils/src/main/java/com/cloud/utils/net/NetUtils.java
```diff
@@ -487,6 +490,10 @@ public static boolean validIpRange(final String startIP, final String endIP) {
         return startIPLong <= endIPLong;
     }
 
+    public static boolean isValidMtu(final int mtu) {
```
wido: Shouldn't we check for null here as well? If null, return false

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
gmueller-ewerk: Correct. I change it the an **Integer** with default null value.

utils/src/main/java/com/cloud/utils/net/NetUtils.java
```diff
@@ -487,6 +490,14 @@ public static boolean validIpRange(final String startIP, final String endIP) {
         return startIPLong <= endIPLong;
     }
 
+    public static boolean isValidMtu(final Integer mtu) {
```
wido: Looks good!

One thing though, I don't see a UnitTest for this method.

Could you just add a UnitTest to test it? Then it looks good to go!

gmueller-ewerk: Good hint, I added a test and hope you are fully satisfied :)

rhtyd:
@blueorangutan package

svenvogel:
@DaanHoogland why thumbs down?

DaanHoogland:
@svenvogel because the packaging failed on all platforms. The thumb down is on the blueorangutan output.

DaanHoogland:
restarting
@blueorangutan package

api/src/main/java/org/apache/cloudstack/api/command/user/vm/UpdateVmNicIpMtuCmd.java
```diff
@@ -46,9 +28,27 @@
...
+import org.apache.cloudstack.api.response.UserVmResponse;
+import org.apache.cloudstack.context.CallContext;
+import org.apache.log4j.Logger;
 
-@APICommand(name = "updateVmNicIp", description = "Update the default Ip of a VM Nic", responseObject = UserVmResponse.class)
-public class UpdateVmNicIpCmd extends BaseAsyncCmd {
+import java.util.ArrayList;
+import java.util.EnumSet;
+
+@APICommand(name = "updateVmNicIpMtu", description = "Update the default Ip and MTU of a VM Nic", responseObject = UserVmResponse.class)
```
DaanHoogland: rename of a public API is a big no-no. it breaks backwards compatibility. if the extra parameter is optional, you can just add it, else create an extra API, please.

api/src/main/java/org/apache/cloudstack/api/command/user/vm/UpdateVmNicIpMtuCmd.java
```diff
@@ -101,11 +105,18 @@ public Long getNicId() {
...
-        if (ipAddr != null) {
-            return ipAddr;
-        } else {
+        if (Strings.isNullOrEmpty(ipAddr)) {
             return null;
         }
+        return ipAddr;
+    }
+
+    public int getMtu () {
```
DaanHoogland: just retrun mtu will do, unless you are meaning to check for mtu <= 0 or something else.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -1459,9 +1459,10 @@ public UserVm updateDefaultNicForVirtualMachine(UpdateDefaultNicForVMCmd cmd) th
     }
 
     @Override
-    public UserVm updateNicIpForVirtualMachine(UpdateVmNicIpCmd cmd) {
+    public UserVm updateNicIpMtuForVirtualMachine(UpdateVmNicIpMtuCmd cmd) {
```
DaanHoogland: service level call renaming is not as bad as API but migyht break third party plugins as well!
Also I don't think it is needed, a small refactor and two methods, one calling the other will do the trick.

holgerstolzenberg:
During out latests internal tests we discovered some smaller issues with this feature. We are going to fix that. In the meantime, please DO NOT merge the PR. 

@gmueller-ewerk: Can anyone with admin permission please reflag the PR with `[DO NOT MERGE WIP]`.

ui/scripts/instances.js
```diff
@@ -2948,7 +2969,7 @@
                                     }
 
                                     $.ajax({
-                                        url: createURL('updateVmNicIp'),
+                                        url: createURL('updateVmNicIpMtu'),
```
holgerstolzenberg: @DaanHoogland 

As for reverting the API changes, I guess this needs to be reverted also?

engine/schema/src/main/resources/META-INF/db/create-default-role-api-mappings.sql
```diff
@@ -332,7 +332,7 @@ INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`,
 INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`, `sort_order`) values (UUID(), 2, 'updateVMAffinityGroup', 'ALLOW', 309) ON DUPLICATE KEY UPDATE rule=rule;
 INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`, `sort_order`) values (UUID(), 2, 'updateVPC', 'ALLOW', 310) ON DUPLICATE KEY UPDATE rule=rule;
 INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`, `sort_order`) values (UUID(), 2, 'updateVirtualMachine', 'ALLOW', 311) ON DUPLICATE KEY UPDATE rule=rule;
-INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`, `sort_order`) values (UUID(), 2, 'updateVmNicIp', 'ALLOW', 312) ON DUPLICATE KEY UPDATE rule=rule;
+INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`, `sort_order`) values (UUID(), 2, 'updateVmNicIpMtu', 'ALLOW', 312) ON DUPLICATE KEY UPDATE rule=rule;
```
holgerstolzenberg: @DaanHoogland: As for reverting the API naming changes, I guess this needs to be reverted also?

ui/scripts/instances.js
DaanHoogland: yes @holgerstolzenberg

engine/schema/src/main/resources/META-INF/db/create-default-role-api-mappings.sql
DaanHoogland: yes

DaanHoogland:
@holgerstolzenberg can you resolve the conflicts and force push your branch?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -1018,72 +1019,71 @@ public String toString() {
...
             _model = model;
             _networkRateKBps = networkRateKBps;
             _dpdkExtraLines = extra;
+            _mtu = mtu;
         }
 
-        public void defDirectNet(String sourceName, String targetName, String macAddr, NicModel model, String sourceMode) {
-            defDirectNet(sourceName, targetName, macAddr, model, sourceMode, 0);
-        }
+        /* ----- DIRECT-NET ----- */
```
GabrielBrascher: Comments are fine, but Javadoc would be better.

DaanHoogland:
@gmueller-ewerk there is a small conflict:
```
.detail-view .details.group-multiple {
  float: left;
<<<<<<< nic_set_mtu_size
  height: 700px;
=======
>>>>>>> master
  width: 100%;
  height: 600px;
  margin-bottom: 30px;
}
```
Can you quickly decide to use 600px or 700px?

rhtyd:
@gmueller-ewerk can you fix the conflicts and address review comments.
Thanks.

gmueller-ewerk:
@rhtyd I updated the PR. Accidentally I made a rebase before the merge from master but resolved all conflicts. Could you please have a look? Thank you.

DaanHoogland:
we've got a drunk rat ;) the file services/console-proxy-rdp/rdpconsole/pom.xml is unchanged so it shouldn't be in the changes, @gmueller-ewerk . I don't see why it would give a license error, but can you please remove it from the change?

gmueller-ewerk:
@DaanHoogland Thank you for figuring out. I removed the xml file.

svenvogel:
> If a variable of type Boolean is equal to null, this tends to indicate the value is false (but here we seem to have it indicate the value is true). Is this really what we want to do?

@mike-tutkowski 
do you mean this should result in the MTU change on the bridges?

DaanHoogland:
@gmueller-ewerk new conflicts, sorry. can you please resolve them?

gmueller-ewerk:
@DaanHoogland I'm already working on it.

rhtyd:
@gmueller-ewerk can you merge/rebase against latest master and fix the conflicts?

rhtyd:
@gmueller-ewerk can you merge/rebase against latest master and fix the conflicts? cc @svenvogel

svenvogel:
@gmueller-ewerk can you rebase it?

@rhtyd @ustcweizhou @wido there are anybody has an ideas why we get these error here with OVS.
https://github.com/apache/cloudstack/pull/3186#pullrequestreview-242716403

rhtyd:
@svenvogel I'm not sure about OVS issues.
Fixed db upgrade path conflict

svenvogel:
@mdominka can you please fix the conflicts?

svenvogel:
@rhtyd can you kick tests again? @wido can you check again. i think we are now ready to merge.

rhtyd:
@blueorangutan package

mdominka:
@rhtyd can you please check this failed test:
TESTS="component/find_hosts_for_migration component/test_acl_isolatednetwork component/test_acl_isolatednetwork_delete component/test_acl_listsnapshot

svenvogel:
@rhtyd what do you think looks good?

svenvogel:
LGTM now i think.

rhtyd:
@blueorangutan package

svenvogel:
@wido @GabrielBrascher could you find time to look at it?

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -563,6 +569,11 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
                 }
 
                 _networkOfferingDao.persistDefaultL2NetworkOfferings();
+
+                final String mtu_size =_configDao.getValue(KVM_MTU_SIZE);
+                if (nonNull(mtu_size)) {
```
wido: Can we use our utils for this or otherwise use Apache Commons notNull?

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -865,6 +876,10 @@ public void saveExtraDhcpOptions(final String networkUuid, final Long nicId, fin
             configureNicProfileBasedOnRequestedIp(requested, profile, network);
         }
 
+        if(Objects.isNull(profile.getMtu())) {
```
wido: Same applies here. Please use the Apache Commons validator

engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
```diff
@@ -540,3 +664,7 @@ CREATE TABLE `cloud`.`template_ovf_properties` (
 -- Add VM snapshot ID on usage helper tables
 ALTER TABLE `cloud_usage`.`usage_vmsnapshot` ADD COLUMN `vm_snapshot_id` BIGINT(20) NULL DEFAULT NULL AFTER `processed`;
 ALTER TABLE `cloud_usage`.`usage_snapshot_on_primary` ADD COLUMN `vm_snapshot_id` BIGINT(20) NULL DEFAULT NULL AFTER `deleted`;
+
+-- Add configuration to set the MTU for OVS and KVM
+INSERT IGNORE INTO `cloud`.`configuration` VALUES ('Network', 'DEFAULT', 'management-server', 'kvm.mtu.size', null,
```
wido: This isn't needed if you use the configkey storage of CloudStack

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -563,6 +569,11 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
                 }
 
                 _networkOfferingDao.persistDefaultL2NetworkOfferings();
+
+                final String mtu_size =_configDao.getValue(KVM_MTU_SIZE);
```
wido: Please use ConfigKey for this

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
mdominka: Hi, sure, but what's wrong with Java 1.7 java.util.Objects?

wido: As we mainly use those tools in CloudStack for validating

mdominka:
@wido please review the changes, thanks.

svenvogel:
@andrijapanicsb can you take  a look again?

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -230,7 +227,18 @@
 public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestrationService, Listener, Configurable {
     static final Logger s_logger = Logger.getLogger(NetworkOrchestrator.class);
 
-    private static final String KVM_MTU_SIZE = "kvm.mtu.size";
+    public static final ConfigKey<Integer> NetworkGcWait = new ConfigKey<Integer>(Integer.class,
+        "network.gc.wait", "Advanced", "600",
```
andrijapanicsb: Why are these here? I mean, they already exist in 4.13?

andrijapanicsb: network.gc.interval and network.gc.wait

api/src/main/java/org/apache/cloudstack/api/command/user/vm/UpdateVmNicIpCmd.java
```diff
@@ -46,8 +28,25 @@
...
+import org.apache.cloudstack.api.response.NicResponse;
+import org.apache.cloudstack.api.response.UserVmResponse;
+import org.apache.cloudstack.context.CallContext;
+import org.apache.log4j.Logger;
 
-@APICommand(name = "updateVmNicIp", description = "Update the default Ip of a VM Nic", responseObject = UserVmResponse.class)
+import java.util.ArrayList;
+import java.util.EnumSet;
+
+@APICommand(name = "updateVmNicIp", description = "Update the default Ip and MTU of a VM Nic", responseObject = UserVmResponse.class)
```
andrijapanicsb: "IP" not "Ip".

engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
```diff
@@ -170,6 +170,130 @@ INSERT IGNORE INTO `cloud`.`guest_os_hypervisor` (uuid, hypervisor_type, hypervi
 ALTER TABLE `cloud`.`service_offering_details` CHANGE COLUMN `value` `value` TEXT NOT NULL;
 
 ALTER TABLE `cloud`.`vpc_offerings` ADD COLUMN `sort_key` int(32) NOT NULL default 0 COMMENT 'sort key used for customising sort method';
```
andrijapanicsb: You are still adding these to schema-41200to41300.sql instead of schema-41300to41400.sql !

utils/src/main/java/com/cloud/utils/net/NetUtils.java
```diff
@@ -19,6 +19,22 @@
```
andrijapanicsb: Just wondering, why all these rearrangements of the position of _import_ lines?
This is done on couple of files if I see it correctly?

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
mdominka: I did not add these keys, just moved them from the bottom up.

utils/src/main/java/com/cloud/utils/net/NetUtils.java
mdominka: I uncheckd the "optimize import" feature of IDEA, hope that helps.

mdominka:
@andrijapanicsb can you take a look again? Thanks.

svenvogel:
@andrijapanicsb @wido can you please take a look again?

engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
mdominka: I think you mean this: 
-- PR#3186 Add possibility to set MTU size for NIC
ALTER TABLE `cloud`.`nics` ADD COLUMN `mtu` smallint (6) COMMENT 'MTU size for the interface';

DROP VIEW IF EXISTS `cloud`.`user_vm_view`;
CREATE VIEW `cloud`.`user_vm_view` AS 
...
and I moved that to the current schema.

engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
```diff
@@ -539,4 +539,4 @@ CREATE TABLE `cloud`.`template_ovf_properties` (
 
 -- Add VM snapshot ID on usage helper tables
 ALTER TABLE `cloud_usage`.`usage_vmsnapshot` ADD COLUMN `vm_snapshot_id` BIGINT(20) NULL DEFAULT NULL AFTER `processed`;
-ALTER TABLE `cloud_usage`.`usage_snapshot_on_primary` ADD COLUMN `vm_snapshot_id` BIGINT(20) NULL DEFAULT NULL AFTER `deleted`;
+ALTER TABLE `cloud_usage`.`usage_snapshot_on_primary` ADD COLUMN `vm_snapshot_id` BIGINT(20) NULL DEFAULT NULL AFTER `deleted`;
```
andrijapanicsb: I believe you should not be touching an old upgrade schema whatsoever?

mdominka: I changed/add nothing to this schema.. Just moved the added lines from Gerd Müller (see annoted ..) to the actual schema.

mdominka:
Now the changes are only in the current SQL schema.

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
mdominka: changed, please check again

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
mdominka: changed, please check again

engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
mdominka: changed, please check again

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
mdominka: changed, please check again

api/src/main/java/org/apache/cloudstack/api/command/user/vm/UpdateVmNicIpCmd.java
mdominka: changed, please check again

andrijapanicsb:
@svenvogel I kindly suggest that you guys install clean 4.13 version and then upgrade to packages built from this PR, to see if the config key defined global setting will be added or not during upgrades. I've seen cases where it is NOT added during upgrades

GabrielBrascher:
@svenvogel I am testing it right now; should have a feedback soon (based on code and manual tests).

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -578,7 +583,17 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
 
                 _networkOfferingDao.persistDefaultL2NetworkOfferings();
 
-                _kvmMtuSize = NumbersUtil.parseInt(_configDao.getValue(KvmMtuSize.key()), 0);
+                final List<ClusterVO> clusterVOs = _clusterDao.listByHypervisorAndCluster(HypervisorType.KVM, Cluster.ClusterType.CloudManaged);
+
+                if ((clusterVOs != null) && !clusterVOs.isEmpty()) {
```
GabrielBrascher: Can you please update this if conditional to `if (!CollectionUtils.isEmpty(clusterVOs)) { ... }`?
The "isEmpty" is a Null-safe check; thus, it is a great option for checking if a collection is empty (returning true if null or empty).

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -1054,7 +1061,7 @@ public void validateCustomParameters(ServiceOfferingVO serviceOffering, Map<Stri
 
             if (serviceOffering.getSpeed() == null) {
                 String cpuSpeed = customParameters.get(UsageEventVO.DynamicParameters.cpuSpeed.name());
-                if ((cpuSpeed == null) || (NumbersUtil.parseInt(cpuSpeed, -1) <= 0)) {
+                if ((cpuSpeed == null) || (parseInt(cpuSpeed, -1) <= 0)) {
```
GabrielBrascher: Could you please update the `cpuSpeed == null` check to `StringUtils.isNotBlank(cpuSpeed)`?

The `lang3.StringUtils.isNotBlank(CharSequence cs)` checks if a CharSequence is not empty (""), not null and not whitespace only.
```
 StringUtils.isNotBlank(null)      = false
 StringUtils.isNotBlank("")        = false
 StringUtils.isNotBlank(" ")       = false
 StringUtils.isNotBlank("bob")     = true
 StringUtils.isNotBlank("  bob  ") = true
```

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -2083,31 +2090,41 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
 
         Map<String, String> configs = _configDao.getConfiguration("AgentManager", params);
 
-        _kvmMtuSize = NumbersUtil.parseInt(_configDao.getValue(KVM_MTU_KEY), 0);
+        final List<ClusterVO> clusterVOs = _clusterDao.listByHypervisorAndCluster(HypervisorType.KVM, Cluster.ClusterType.CloudManaged);
+
+        if ((clusterVOs != null) && !clusterVOs.isEmpty()) {
```
GabrielBrascher: CollectionUtils.isEmpty(clusterVOs) would be nice here :slightly_smiling_face:  :+1:

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -1019,21 +1021,26 @@ public String toString() {
...
             _macAddr = macAddr;
             _model = model;
             _networkRateKBps = networkRateKBps;
+            _mtu = mtu;
         }
-
-        public void defDpdkNet(String dpdkSourcePath, String dpdkPort, String macAddress, NicModel model,
-                               Integer networkRateKBps, String extra, String interfaceMode) {
+       /**
+       * Dpdk Net
```
GabrielBrascher: Same here, a bit more details would be nice. Otherwise, I think that the Javadoc block could be removed.
PS: thumbs up for the Javadoc initiative it is always a plus :+1:

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
mdominka: I did not change this code (only static import), but I adapt that to StringUtils.isBlank(). Thanks for the hint.

GabrielBrascher:
@mdominka @svenvogel @gmueller-ewerk thanks for the changes!

Quick update on tests. I have been testing it and overall looks fine; however, it would be nice to have VRs automatically configured with the MTU.

When a VR is deployed, the MTU is being correctly configured on the KVM (e.g. _/run/libvirt/qemu/r-2293-VM.xml_) but the VR OS is still configured with default MTU (1500). It still would be needed to pass this in 'cmdline' to the VR on boot to have VRs booted with correct MTU.

svenvogel:
>When a VR is deployed, the MTU is being correctly configured on the KVM (e.g. /run/libvirt/qemu/r-2293-VM.xml) but the VR OS is still configured with default MTU (1500). It still would be needed to pass this in 'cmdline' to the VR on boot to have VRs booted with correct MTU.

@GabrielBrascher thanks for testing! we dont know about the router scripts in VPC or virtual router an how we can implement it. i think its not so easy. maybe you can implement it?

@andrijapanicsb yes we tested it on a new 4.13 install. seems to work fine.

wido:
@svenvogel Without sending the MTU parameter to the VR things like password resets and other metadata >1500 bytes would become a problem.

During startup the 'cmdline' is send to the VR/SSVM which it uses to configure it's IPs and other network related.

There we would need to add a parameter which also sets the MTU on those NIC(s).

andrijapanicsb:
> @andrijapanicsb yes we tested it on a new 4.13 install. seems to work fine.

I was speaking about upgrades, where the config key (older style of code) will not be executed during the upgrades and the setting will be missing completely.

svenvogel:
@wido you are completely right. we will test it tomorrow. the main problem is we dont have the knowledge about the router internal scripts. you know? there is nobody who know where "cmdline" will be used. a grep inside a virtual router gave ~20 script where this variable will be used. 

i will repeat. we dont have the skills to change this. we can try to add the mtu value to the "cmdline" but we need help by the change of the router scripts? can you or @GabrielBrascher help?

Cheers

Sven

svenvogel:
@andrijapanicsb we test both thing but we dont encounter any problem. can you check this too?

GabrielBrascher:
@svenvogel parsing data to systemvms requires a bit of work indeed; here follows a bit of explanation.

The [common.sh](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/setup/common.sh) script (_cloudstack/systemvm/debian/opt/cloud/bin/setup/common.sh_) parses the 'cmdline' values (as mentioned above by @wido ). That script is the base for multipe scripts such as:

- [consoleproxy.sh](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/setup/consoleproxy.sh)
- [router.sh](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/setup/router.sh)
- [secstorage.sh](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/setup/secstorage.sh)
- [vpcrouter.sh](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/setup/vpcrouter.sh)

On VRs we can see the **cmdline** values at _/var/cache/cloud/cmdline_:

`root@r-2293-VM:~# cat /var/cache/cloud/cmdline`

`template=domP name=r-2293-VM eth0ip=109.72.91.50 eth0mask=255.255.255.192 gateway=109.72.91.1 domain=cs1cloud.internal cidrsize=26 dhcprange=109.72.91.1 eth1ip=169.254.0.52 eth1mask=255.255.0.0 type=dhcpsrvr disable_rp_filter=true dns1=93.180.70.22 dns2=93.180.70.30 (etc....)`

For VRs, there is [router.sh](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/setup/router.sh), used for configuring virtual router (e.g. network interfaces).

On the management server side, there is the preparation code that sets up a StringBuilder that will contain all these configuration data. For (non-VPC) VRs we can find a bit of details at method finalizeVirtualMachineProfile in [VirtualNetworkApplianceManagerImpl.java](https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/network/router/VirtualNetworkApplianceManagerImpl.java#L1311)

On the execution flow, there are also the python scripts [merge.py](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/merge.py) & [CsAddress.py](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/cs/CsAddress.py).

The question now is how to proceed with this. Does it feet as a new PR? Would it best fit in this PR?

svenvogel:
Hi @GabrielBrascher @mdominka yes we should start another PR after this one. Make this sense? we can approve this one so this PR seems ready.

svenvogel:
@DaanHoogland is https://github.com/apache/cloudstack/pull/3186/commits/1a91c457c1f16ce6d7e2750c85aacc7e30f91e2c correctly?

possibly the wrong PR? 👍

DaanHoogland:
@svenvogel @gmueller-ewerk it is better practice to rebase the branch upon master, but as we squash-merge on merge a new commit will be created anyway, so it could just turn out fine. The cosmetic and administrative disadvantages will disappear.

andrijapanicsb:
I agree with @GabrielBrascher, I've seen some nasty issues with MTU not being in place correctly.

I do want an opinion from @rhtyd as well on this one.

That being said, I do need to back out from this PR, due to lack of time (instead of silently ignoring it) as it would mean considerable testing effort (I can't help much with code obviously).

svenvogel:
@GabrielBrascher @andrijapanicsb sure we can put it on-hold and work in progress. at the moment there are no resources to get the things done with the router scripts.

andrijapanicsb:
@svenvogel do you think you could work on this, or should we move it to the 4.15 milestone?

rohityadavcloud:
Hi @gmueller-ewerk, given inactivity on this PR for last 3+ years, I'm closing this. If you happen to work on this PR again, pl re-open this with conflicts fixed or re-raise a fresh PR. I've meanwhile logged this as an issue for tracking purposes https://github.com/apache/cloudstack/issues/6014

Issue #3192, opened 2019-02-22
Security Group rules not applied at all for XenServer 6.5 / Advanced Zone


andrijapanicsb:
##### ISSUE TYPE

 * Bug Report
 

##### COMPONENT NAME

~~~
Management Server most probably
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2, 4.12-RC2/RC3
~~~

##### CONFIGURATION

Advanced Zone with Security Groups enabled, VM deployed inside Shared Network

##### OS / ENVIRONMENT

Hypervisor: XenServer 6.5(sp1)

##### SUMMARY

Security Group rules (iptable rules) not applied at all on XenServer 6.5 / 6.5SP1, in Advanced Zone with SG enabled on 4.11+ (FYI: works in 4.9.3 release)

Also migrating a VM to another host, does NOT remove rules from source host and does NOT create rules (just an empty iptable chains) on the destination host

##### STEPS TO REPRODUCE

- Deploy a VM in a Shared Network with Security Groups (Advanced Zone with Security Groups enabled) will result in no rules being applied on the Hypervisor side (XenServer 6.5 / 6.5SP1)
- Edit default Security Group and add a few rules for incoming/outgoing traffic
- On XenServer 6.5 do "iptables -L -n | grep i-2-3-VM" and observe no rules for specific ports are present (i-2-3-VM being VM in test; specific ports being ports you added to default SG previously) 

For migration issue - simply migrate VM to another host and run "iptables -L -n | grep i-2-3-VM" and observe no rules on destination host are created.

##### EXPECTED RESULTS

~~~
Security Group rules (iptable rules) applied properly on XenServer 6.5
~~~

##### ACTUAL RESULTS

~~~
no rules applied. cloud.log from XenServer on 4.9.3.1 vs 4.11.2 below (confirmed issue also for 4.12-RC2/RC3)
~~~
[4112-logs.txt](https://github.com/apache/cloudstack/files/2893090/4112-logs.txt)
[4391-logs.txt](https://github.com/apache/cloudstack/files/2893091/4391-logs.txt)

GabrielBrascher:
@andrijapanicsb please review PR #3197. Thanks!

andrijapanicsb:
Give me 5min @GabrielBrascher ...

GabrielBrascher:
No problem @andrijapanicsb

GabrielBrascher:
Are you ok with PR #3197 @andrijapanic?

[...]

PR #3197, opened 2019-02-28
Fix XenServer Security Groups 'vmops' script

GabrielBrascher:
Fixing issue: #3192 Security Group rules not applied at all for
XenServer 6.5 / Advanced Zone

- Fix `tokens = line.split(':')` to `tokens = line.split(';')`
- Fix expected tokens size from 5 to 4
```
if len(tokens) != 5:
   continue
```
- Add logs
- remove unused vmops script. The XCP patch points to the vmops script on the parent folder [1]. Thus, all XenServer versions are considering the vmops script located at [2].

[1] https://github.com/apache/cloudstack/blob/master/scripts/vm/hypervisor/xenserver/xcposs/patch#L30
[2] https://github.com/apache/cloudstack/blob/master/scripts/vm/hypervisor/xenserver/vmops
## Types of changes

## How Has This Been Tested?

- SSH fail as expected
- Add Ingress and Egress rules with start port=22, end port=22, and cidr=0.0.0.0/0
- SSH succeeds as expected

**iptables -L**
```
Chain i-2-125-VM (1 references)
target     prot opt source               destination         
ACCEPT     tcp  --  anywhere             anywhere             tcp dpt:ssh ctstate NEW
DROP       all  --  anywhere             anywhere            

Chain i-2-125-VM-def (2 references)
target     prot opt source               destination         
RETURN     udp  --  anywhere             anywhere             PHYSDEV match --physdev-in vif22.0 --physdev-is-bridged match-set i-2-125-VM src udp dpt:domain
DROP       all  --  anywhere             anywhere             PHYSDEV match --physdev-in vif22.0 --physdev-is-bridged ! match-set i-2-125-VM src
DROP       all  --  anywhere             anywhere             PHYSDEV match --physdev-out vif22.0 --physdev-is-bridged ! match-set i-2-125-VM dst
i-2-125-VM-eg  all  --  anywhere             anywhere             PHYSDEV match --physdev-in vif22.0 --physdev-is-bridged match-set i-2-125-VM src
i-2-125-VM  all  --  anywhere             anywhere             PHYSDEV match --physdev-out vif22.0 --physdev-is-bridged

Chain i-2-125-VM-eg (1 references)
target     prot opt source               destination         
RETURN     tcp  --  anywhere             anywhere             tcp dpt:ssh ctstate NEW
DROP       all  --  anywhere             anywhere            

```
**cloud.log output**
```
DEBUG [root] Network rules compressed in base64 [eJzztCpJLrA2MgIhAz0w1DfQ8XONCFFwxS0FAG5GDOs=].
 INFO [root] Network rules to be processed [I:tcp;22;22;0.0.0.0/0,NEXT E:tcp;22;22;0.0.0.0/0,NEXT ].
 INFO [root] Programming network rules for vm  i-2-125-VM seqno=37 numrules=3 signature=5173d229b469dcfb63575ae93fde95ec guestIp=192.168.100.152, update iptables, reason=seqno_change_or_sig_change
DEBUG [root] Processing rule [I:tcp;22;22;0.0.0.0/0,NEXT].
DEBUG [root] Tokens ['I:tcp', '22', '22', '0.0.0.0/0,NEXT'].
 INFO [root] Ipset chaing [i-2-125-VM], VM [i-2-125-VM], Action [ACCEPT], Direction [src], egress rules [0]
DEBUG [root] port range [22:22]
DEBUG [root] ['iptables', '-I', 'i-2-125-VM', '-p', 'tcp', '-m', 'tcp', '--dport', '22:22', '-m', 'state', '--state', 'NEW', '-j', 'ACCEPT']
DEBUG [root] Processing rule [E:tcp;22;22;0.0.0.0/0,NEXT].
DEBUG [root] Tokens ['E:tcp', '22', '22', '0.0.0.0/0,NEXT'].
 INFO [root] Ipset chaing [i-2-125-VM-e], VM chain [i-2-125-VM-eg], VM name [i-2-125-VM], Action [RETURN], Direction [dst], egress rules [1]
DEBUG [root] port range [22:22]
DEBUG [root] ['iptables', '-I', 'i-2-125-VM-eg', '-p', 'tcp', '-m', 'tcp', '--dport', '22:22', '-m', 'state', '--state', 'NEW', '-j', 'RETURN']
DEBUG [root] Processing rule [].
DEBUG [root] Tokens [''].
WARNING [root] Network rule tokens size [1] is different from the expected size [4], ignoring rule ['']
DEBUG [root] Executing command: [['iptables', '-I', 'i-2-125-VM', '-p', 'tcp', '-m', 'tcp', '--dport', '22:22', '-m', 'state', '--state', 'NEW', '-j', 'ACCEPT']].
DEBUG [root] Executing command: [['iptables', '-I', 'i-2-125-VM-eg', '-p', 'tcp', '-m', 'tcp', '--dport', '22:22', '-m', 'state', '--state', 'NEW', '-j', 'RETURN']].
DEBUG [root] Writing log to /var/run/cloud/i-2-125-VM.log
```


GabrielBrascher:
Ping @andrijapanic @wido @rafaelweingartner @rhtyd @DaanHoogland @kiwiflyer @nathanejohnson @dhlaluku.

rhtyd:
@blueorangutan package

scripts/vm/hypervisor/xenserver/vmops
```diff
@@ -1517,7 +1550,8 @@ def network_rules(session, args):
 
     return 'true'
   except:
-    logging.debug("Failed to network rule !")
+    e = sys.exc_info()[0]
+    logging.error("Failed to network rule. Error: %s" % e)
```
wido: Might want to use logging.exception here as you catch an Exception.

scripts/vm/hypervisor/xenserver/vmops
```diff
@@ -1438,14 +1448,31 @@ def network_rules(session, args):
...
+        # token_type: I
+        # protocol: tcp
+        token_type, protocol = tokens[0].split(':')
+        #tokens[1] = '12'
+        start = tokens[1]
+        #tokens[2] = '34'
+        end = tokens[2]
+        #cidrs = ['1.2.3.4/24', 'NEXT']
         cidrs = tokens.pop().split(",")
+        #cidrs = ['1.2.3.4/24']
```
wido: All the above rules seem debugging lines, don't they?

andrijapanicsb:
Tested with XenServer 6.5sp1 - works fine, rules are added fine - all good.

LGTM

scripts/vm/hypervisor/xenserver/vmops
```diff
@@ -1438,14 +1448,31 @@ def network_rules(session, args):
...
+        #tokens[0] = I:tcp
+        # token_type: I
+        # protocol: tcp
+        token_type, protocol = tokens[0].split(':')
+        #tokens[1] = '12'
+        start = tokens[1]
+        #tokens[2] = '34'
+        end = tokens[2]
+        #cidrs = ['1.2.3.4/24', 'NEXT']
         cidrs = tokens.pop().split(",")
```
rafaelweingartner: Why use pop? I mean, that is the same as `cidrs = tokens[3].split(",")`, which would follow the method you are using to access the array items.

scripts/vm/hypervisor/xenserver/vmops
```diff
@@ -1438,14 +1448,31 @@ def network_rules(session, args):
...
+        #cidrs = ['1.2.3.4/24']
         cidrs.pop()
```
rafaelweingartner: This last pop needs some explanations. This is only here to remove the "NEXT " element in the array of CIDRs

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
Thanks for the reviews @rafaelweingartner @wido @andrijapanicsb.
@wido and @rafaelweingartner I updated the code considering your comments.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
All checks have passed, 2 LGTMs, manual tests passed and the Trillian tests that are failing are not relative to this change. I will be merging this PR.

This commit addresses the blocker #3192 raised for 4.12 RC3; thus, I will also continue with the VOTE process for 4.12 release candidate 4.

Thanks for the feedback guys!

andrijapanicsb:
Just FTR - in https://github.com/apache/cloudstack/issues/3192 there is another issue of SG rules not applied during VM live migration to new host - probably should be fixed here if possible.

core/src/main/java/com/cloud/agent/api/SecurityGroupRulesCmd.java
```diff
@@ -173,7 +173,7 @@ public String getSecIpsString() {
         final StringBuilder sb = new StringBuilder();
         final List<String> ips = getSecIps();
         if (ips == null) {
-            sb.append("0:");
+            sb.append("0;");
```
rafaelweingartner: Shouldn't you use here `RULE_COMMAND_SEPARATOR`?

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/wrapper/xenbase/CitrixMigrateCommandWrapper.java
```diff
@@ -81,6 +81,13 @@ public Answer execute(final MigrateCommand command, final CitrixResourceBase cit
                 }
                 citrixResourceBase.migrateVM(conn, dsthost, vm, vmName);
                 vm.setAffinity(conn, dsthost);
+
+                final String result = citrixResourceBase.callHostPlugin(conn, "vmops", "destroy_network_rules_for_vm", "vmName", command.getVmName());
```
rafaelweingartner: So, the cleanup was never implemented...?

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/wrapper/xenbase/CitrixMigrateCommandWrapper.java
```diff
@@ -81,6 +81,13 @@ public Answer execute(final MigrateCommand command, final CitrixResourceBase cit
...
+                if (result == null || result.isEmpty() || !Boolean.parseBoolean(result)) {
```
rafaelweingartner: Have you tried using `BooleanUtils`?

GabrielBrascher:
@wido @rafaelweingartner @andrijapanicsb and others. I have updated this PR addressing the issue reported on #3192.

Now the `CitrixMigrateCommandWrapper.java` calls the script `vmops` removing the SG network rules on the source host. Additionally, fixed a bug when secondary IP was null (was expecting `"0;"` instead of `"0:"`), adding the rules on the target host when migrating a VM.

andrijapanicsb:
thx @GabrielBrascher - will test today and let you know.

andrijapanicsb:
@blueorangutan package

GabrielBrascher:
Going to extract the added code to a method (executed if `citrixResourceBase.canBridgeFirewall()`), and add some JUnit tests.

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/wrapper/xenbase/CitrixMigrateCommandWrapper.java
GabrielBrascher: Good point @rafaelweingartner. I will update to

```
if (BooleanUtils.toBoolean(result)) {
    s_logger.debug(String.format("Removed network rules from source host [%s] for migrated vm [%s]", dsthost.getHostname(conn), command.getVmName()));
} else {
    s_logger.warn(String.format("Failed to remove network rules from source host [%s] for migrated vm [%s]", dsthost.getHostname(conn), command.getVmName()));                
}
```

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/wrapper/xenbase/CitrixMigrateCommandWrapper.java
GabrielBrascher: Yes, it looks like @rafaelweingartner.

I am not aware if it was or not removing the network rules before 4.11/4.12, but I don't see how it would be removed as the mgmt was not calling `destroy_network_rules_for_vm` from the vmops.

rafaelweingartner: Thanks for the explanation

core/src/main/java/com/cloud/agent/api/SecurityGroupRulesCmd.java
GabrielBrascher: Yes, that would avoid future issues in case of an update on the RULE_COMMAND_SEPARATOR. I will update it soon when having the other raised points addressed as well. Thanks for pointing that!

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
Updated PR with unit tests and addressing reviewers.

andrijapanicsb:
LGTM - adding rules work fine, also rules are migrated (and removed from source host) after VM live migration

^^^ on XenServer 6.5sp1

Thanks a lot @GabrielBrascher 

Any idea why/when things broke so badly ? (between 4.10 and 4.11) ?

GabrielBrascher:
@andrijapanic happy to be usefull :)

The problem is that we have some different workflows for XenServer and KVM. Thus, we struggled to cover all the possible cases.

The best approach, in my opinion, would be to have only one script handling SG network rules, regardless of the hypervisor. However, such change would require a deep refactoring on our codebase.

GabrielBrascher:
All checks (Jenkins + Travis) have passed and we have enough LGTMs.
I am on hold to check the CI results. If there are no objections I will be then merging this PR.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

PR #3198, opened 2019-02-28
[WIP DO NOT MERGE] Set service offering unique name over api and gui

pbankonier:
## Description
This adds the possibility to add/change the defaultUse/unique name of system service offerings over the ui and api.

The create and update command of the system service offerings now have an optional parameter for the default use.

The default use parameter is a boolean value.

When setting a a new service offering or when editing an existing one to be a default one, the unique name and default use value of the offering which currently had it gets set to NULL.

The unique name depends on the system vm type, localstorage and defaultuse values.

## Types of changes

## Screenshots

In the list it gets shown if the service offering is a default one.

Bildschirmfoto 2019-04-25 um 11 25 52

You can now set the offering to a default one when creating a new system service offering.

Bildschirmfoto 2019-04-24 um 13 38 20

You can set/unset the service offering as a default one in the edit view.

Bildschirmfoto 2019-04-24 um 13 38 46

Bildschirmfoto 2019-04-24 um 13 39 14

I also added the possibility to create a elastic lb here because it was missing.

Bildschirmfoto 2019-04-24 um 13 39 40

## How Has This Been Tested?


svenvogel:
@rhtyd @GabrielBrascher @rafaelweingartner @borisroman can you add a ui icon and check these?

svenvogel:
@DaanHoogland @borisstoyanov @GabrielBrascher can you approve the pull request?

bwsw:
@svenvogel your implementation must respect uniqueness inside the domain but not across all the system. Sven, what is the rationale behind that?

bwsw:
Why this feature is necessary? I don't think it is enough info without the motivation for that even if it does break nothing...

svenvogel:
@bwsw what do you speak about? i dont understand your question "Why this feature is necessary?"  its a feature to set the the offering for the vpc and change it easy with the gui? 

next one to your "Why this feature is necessary? I don't think it is enough info without the motivation for that even if it does break nothing..." -> it doesnt break anything and motivation is clear... change it into the gui.

a the moment you have to do it in the database directly and there is a unique name. 

https://cloudstack-documentation.readthedocs.io/en/latest/adminguide/service_offerings.html

bwsw:
@svenvogel Never mind, I haven't checked it's for system service offerings. Sorry for the mess.

engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
```diff
@@ -0,0 +1,53 @@
+ALTER VIEW `cloud`.`service_offering_view` AS
```
DaanHoogland: compilation fails because of missing license header. @pbankonier please add it

DennisKonrad: That's right. We're going to add the header

DaanHoogland:
@blueorangutan package

DaanHoogland:
@pbankonier as this is marked WIP, what is still to be done for it?

DaanHoogland:
@blueorangutan test

pbankonier:
@DaanHoogland I'm currently working on a optimization here. I found that it is not needed to pass the uniquename via a parameter.It is only needed to pass a default flag and the unique name can get determined by the backend logic cause all information is passed via other parameters. This change will be up soon.

PR #3211, opened 2019-03-12
ui: remove CA certificate button from UI

rhtyd:
This removes CA certificate button from UI as this is not used and
confuses both admin and users how to use it. This may be re-introduced
in future when a use-case can be justified.

## Types of changes


svenvogel:
@rhtyd what was the use case before?

rhtyd:
@svenvogel in 4.11+ if you notice there is a button on right/top side that downloads the CA certificate. This is not used at all and confuses users/admins. This removes it. The original use-case was something to do with cert setup, but that never got implemented. We may put it back if we do that in future.

PaulAngus:
@blueorangutan package

PaulAngus:
@blueorangutan test

svenvogel:
@rhtyd okay i see. no problem. sounds good.

PR #3213, opened 2019-03-12
server: allow admins to blacklist vm details that users should not see

rhtyd:
**Problem**: The listVirtualMachines API response displays sensitive information which should not be visible to any other user which is not Root Admin.
**Root Cause**: There are currently no restrictions in place to control what
instance settings should be visible to Root Admin only.
**Solution**: The issue has been fixed by refactoring the response to filter details based on a global setting and also honour the cloud.user_vm_details table’s display field. The global setting used to hide details that Root admin can set is called `user.vm.blacklisted.details` with the following defaults;
• memoryOvercommitRatio
• cpuOvercommitRatio
• rootdisksize
• Message.ReservedCapacityFreed.Flag

## Types of changes


borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

DagSonsteboSB:
LGTM

PR #3216, opened 2019-03-12
api: include tags in listvmsnapshots response

rhtyd:
**Problem**: Tags of VM snapshots are not returned as part of the listVMSnapshots API and therefore was not shown in the UI.
**Root Cause**: The API response creator was not setting the tags in the API response.
**Solution**: The issue was fixed to return tags of a VM snapshot as part of the listVMSnapshots API response and now it is possible to list/add/remove tags using the UI. The tags are also included for VMSnapshot usage records in the listUsageRecords API response when `includetags=true` parameter is passed.

## Types of changes

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/response/VMSnapshotResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/VMSnapshotResponse.java
@@ -18,18 +18,19 @@
 package org.apache.cloudstack.api.response;
 
 import java.util.Date;
-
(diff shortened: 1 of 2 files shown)


borisstoyanov:
@blueorangtan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DagSonsteboSB:
LGTM

PR #3218, opened 2019-03-12
vmware: don't use redundant worker VM to extract volume

rhtyd:
**Problem**: A VM fails to start, which has a VM snapshot after performing the extractVolume operation on the stopped VM.
**Root Cause**: While downloading a volume of a stopped VM, it creates a worker VM with the volume attached and it creates a temporary snapshot, and then it creates another worker VM cloned from that snapshot and exports an OVA file and deletes the cloned worker VM. Afterward, it deletes the initial worker VM and the temporary snapshot which merges any previous snapshot(s) to the main disk. So, while starting the VM, VMware is not able to find the snapshot disk.
**Solution**: Export the volume using a single worker VM (when VM is stopped) without creating a temporary snapshot.
**Upgrade Notes**: After the upgrade, the admin should replace the live systemvm ISO file in all secondary storage pools with the new systemvm.iso on the management server (keeping the original name which the original systemvm.iso had on secondary storage). Finally, admin should destroy all the old SSVMs.

## Types of changes

### DIFF
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
@@ -930,12 +930,12 @@ public class VmwareStorageManagerImpl implements VmwareStorageManager {
             String prevSnapshotUuid, String prevBackupUuid, String workerVmName, Integer nfsVersion) throws Exception {
 
         String backupUuid = UUID.randomUUID().toString();
-        exportVolumeToSecondaryStroage(vmMo, volumePath, secStorageUrl, getSnapshotRelativeDirInSecStorage(accountId, volumeId), backupUuid, workerVmName, nfsVersion);
+        exportVolumeToSecondaryStorage(vmMo, volumePath, secStorageUrl, getSnapshotRelativeDirInSecStorage(accountId, volumeId), backupUuid, workerVmName, nfsVersion, true);
         return backupUuid + "/" + backupUuid;
     }
 
-    private void exportVolumeToSecondaryStroage(VirtualMachineMO vmMo, String volumePath, String secStorageUrl, String secStorageDir, String exportName, String workerVmName,
-            Integer nfsVersion) throws Exception {
+    private void exportVolumeToSecondaryStorage(VirtualMachineMO vmMo, String volumePath, String secStorageUrl, String secStorageDir, String exportName, String workerVmName,
+                                                Integer nfsVersion, boolean clonedWorkerVMNeeded) throws Exception {
 
         String secondaryMountPoint = _mountService.getMountPoint(secStorageUrl, nfsVersion);
         String exportPath = secondaryMountPoint + "/" + secStorageDir + "/" + exportName;
@@ -961,16 +961,19 @@ public class VmwareStorageManagerImpl implements VmwareStorageManager {
                 throw new Exception(msg);
             }
 
-            // 4 MB is the minimum requirement for VM memory in VMware
-            vmMo.cloneFromCurrentSnapshot(workerVmName, 0, 4, volumeDeviceInfo.second(), VmwareHelper.getDiskDeviceDatastore(volumeDeviceInfo.first()));
-            clonedVm = vmMo.getRunningHost().findVmOnHyperHost(workerVmName);
-            if (clonedVm == null) {
-                String msg = "Unable to create dummy VM to export volume. volume path: " + volumePath;
-                s_logger.error(msg);
-                throw new Exception(msg);
+            if (clonedWorkerVMNeeded) {
+                // 4 MB is the minimum requirement for VM memory in VMware
+                vmMo.cloneFromCurrentSnapshot(workerVmName, 0, 4, volumeDeviceInfo.second(), VmwareHelper.getDiskDeviceDatastore(volumeDeviceInfo.first()));
+                clonedVm = vmMo.getRunningHost().findVmOnHyperHost(workerVmName);
+                if (clonedVm == null) {
+                    String msg = "Unable to create dummy VM to export volume. volume path: " + volumePath;
+                    s_logger.error(msg);
+                    throw new Exception(msg);
+                }
+                clonedVm.exportVm(exportPath, exportName, false, false);  //Note: volss: not to create ova.
+            } else {
+                vmMo.exportVm(exportPath, exportName, false, false);
             }
-
-            clonedVm.exportVm(exportPath, exportName, false, false);  //Note: volss: not to create ova.
         } finally {
             if (clonedVm != null) {
                 clonedVm.detachAllDisks();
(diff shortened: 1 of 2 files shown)


borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

borisstoyanov:
@blueorangutan test

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
```diff
@@ -961,16 +961,19 @@ private void exportVolumeToSecondaryStroage(VirtualMachineMO vmMo, String volume
...
+            if (clonedWorkerVMNeeded) {
+                // 4 MB is the minimum requirement for VM memory in VMware
+                vmMo.cloneFromCurrentSnapshot(workerVmName, 0, 4, volumeDeviceInfo.second(), VmwareHelper.getDiskDeviceDatastore(volumeDeviceInfo.first()));
+                clonedVm = vmMo.getRunningHost().findVmOnHyperHost(workerVmName);
+                if (clonedVm == null) {
+                    String msg = "Unable to create dummy VM to export volume. volume path: " + volumePath;
+                    s_logger.error(msg);
+                    throw new Exception(msg);
+                }
+                clonedVm.exportVm(exportPath, exportName, false, false);  //Note: volss: not to create ova.
```
DaanHoogland: how about unifying both calls by assigning the clonedVm or vmMo to a local var, so it is obvious where the real work is done in both cases?

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
```diff
@@ -1014,16 +1018,19 @@ private void exportVolumeToSecondaryStroage(VirtualMachineMO vmMo, String volume
                 String datastoreVolumePath = getVolumePathInDatastore(dsMo, volumePath + ".vmdk", searchExcludedFolders);
                 workerVm.attachDisk(new String[] {datastoreVolumePath}, morDs);
                 vmMo = workerVm;
```
DaanHoogland: if we move clonedWorkerVMNeeded outside the try workerVm is not needed.

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -999,15 +1000,18 @@ private String getVolumePathInDatastore(DatastoreMO dsMo, String volumeFileName,
...
-            exportVolumeToSecondaryStroage(vmMo, volumePath, secStorageUrl, destVolumePath, exportName, hostService.getWorkerName(hyperHost.getContext(), cmd, 1), _nfsVersion);
+            exportVolumeToSecondaryStorage(vmMo, volumePath, secStorageUrl, destVolumePath, exportName, hostService.getWorkerName(hyperHost.getContext(), cmd, 1), _nfsVersion, clonedWorkerVMNeeded);
             return new Pair<>(destVolumePath, exportName);
 
         } finally {
-            vmMo.removeSnapshot(exportName, false);
+            if (vmMo != null && vmMo.getSnapshotMor(exportName) != null) {
+                vmMo.removeSnapshot(exportName, false);
+            }
             if (workerVm != null) {
```
DaanHoogland: same here as in VmwareStorageManagerImpl.java#1020; if we move clonedWorkerVMNeeded outside the try block we don't need this extra workerVm

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
```diff
@@ -1684,14 +1688,18 @@ public Answer createTemplateFromSnapshot(CopyCommand cmd) {
...
-            clonedVm.exportVm(exportPath, exportName, false, false);
-            return new Pair<>(volumeDeviceInfo.second(), disks);
+            String diskDevice = volumeDeviceInfo.second();
+            String disks[] = vmMo.getCurrentSnapshotDiskChainDatastorePaths(diskDevice);
+            if (clonedWorkerVMNeeded) {
+                // 4 MB is the minimum requirement for VM memory in VMware
+                Pair<VirtualMachineMO, String[]> cloneResult =
+                        vmMo.cloneFromCurrentSnapshot(workerVmName, 0, 4, diskDevice, VmwareHelper.getDiskDeviceDatastore(volumeDeviceInfo.first()));
+                clonedVm = cloneResult.first();
+                clonedVm.exportVm(exportPath, exportName, false, false);
```
DaanHoogland: also here, it would be nice to see just one call for doing the 'real' work.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
rhtyd: It's because, if we need to clone i.e. require a worker VM `clonedVm`, then in finally we need to detach and destroy that worker VM. This is why we cannot refactor them to use a single variable.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
rhtyd: Same reason as before, using different instances/variable, it's easier to track is we need to destroy the worker vm after we're done. If we re-use `vmMo`, we might accidently destroy the user-vm.

plugins/hypervisors/vmware/src/main/java/com/cloud/storage/resource/VmwareStorageProcessor.java
rhtyd: In the current mechanism, we need a worker VM to take snapshot and export the disk; without at least one worker VM I'm not sure how we can export template/disk?

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
DaanHoogland: ok, as said it is a matter of style but we have that boolean to decide: clonedWorkerVMNeeded

rhtyd:
@DaanHoogland I'm not exactly sure how you want us to refactor the changes, we'll need at least one worker VM to extract/snapshot the disks to a single vmdk file because the disk in question may be linked-one (multiple files). If we do figure out a clever way out, I'll raise another PR without blocking this, if that's okay.

DaanHoogland:
I am not 👎 and was about to merge, @rhtyd .

DaanHoogland:
actually ask if merge is ok (tag [WIP] still in place)

borisstoyanov:
yes I think so, if @rhtyd doesn't have something in mind

rhtyd:
Thanks @DaanHoogland

borisstoyanov:
@blueorangutan package

PR #3183, opened 2019-02-15
Improvements after jquery update

NicoWohlfarth:
## Description
Fixed issues occurring with [this PR](https://github.com/apache/cloudstack/pull/3069)
Added jQuery ui CSS to the project and adapt existing stylings
Removed unnecessary styling definitions from cloudstack.css
Swap no longer recommended jQuery method

## Types of changes
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [x] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?
We deployed the changes to our test environment and tested the UI functionality by clicking through the affected views

PR #3225, opened 2019-03-13
ui: fix computer diagram css margin that blocks down arrow

rhtyd:
**Problem**: Down arrow key of the vertical scroll bar in the VM deployment wizard does not work.
**Root Cause**: The computer diagram on the right side in the VM deployment wizard spills its margin on the left causing the arrow key to be blocked.
**Solution**: Fix the `background-position` to get the correct position than `margin-left` for the `zone-panel`. With the fix, the left size arrow keys and vertical scrollbars are no longer under the diagram image and press-able now.

## Types of changes

Before the fix, as seen from dev tools:
Screenshot from 2019-03-13 14-51-00

The layout as seen in dev tools after the fix:
Screenshot from 2019-03-13 15-00-33

### DIFF
--- a/ui/css/cloudstack3.css
+++ b/ui/css/cloudstack3.css
@@ -6878,10 +6878,9 @@ label.error {
 .multi-wizard.instance-wizard .diagram .part.zone-plane {
   width: 354px;
   height: 117px;
-  background-position: 0px -55px;
+  background-position: -38px -55px;
   /*+placement:displace -38px 259px;*/
   position: absolute;
-  margin-left: -38px;
   margin-top: 259px;
 }
 


DennisKonrad:
Hi @rhtyd,

I think this is already fixed in #3183 
We also added some code in response to questions/suggestions from others. Can you look over #3183 and merge?

DennisKonrad:
I double checked this, and the down arrow is also not clickable with #3183. So this is no duplicate  @rhtyd

rhtyd:
Thanks @DennisKonrad

borisstoyanov:
@blueorangutan package

DagSonsteboSB:
LGTM

andrijapanicsb:
Whohoo - finally :)

PR #3227, opened 2019-03-14
ubuntu16:  fix three issues with ubuntu 16.04 hosts

ustcweizhou:
## Description

This fixes three issues with ubuntu 16.04 hosts

1. fix unable to add host if cloudbrX is not configured
2. Stop service libvirt-bin.socket while add a host
3. Diable libvirt default network

Details can be found in each commit.

## Types of changes

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -72,6 +72,7 @@ import org.libvirt.DomainInterfaceStats;
 import org.libvirt.DomainSnapshot;
 import org.libvirt.LibvirtException;
 import org.libvirt.MemoryStatistic;
+import org.libvirt.Network;
 import org.libvirt.NodeInfo;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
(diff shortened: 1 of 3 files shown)


rhtyd:
@blueorangutan package

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -914,6 +915,20 @@ public boolean configure(final String name, final Map<String, Object> params) th
...
         }
 
+        // destroy default network, see https://libvirt.org/sources/java/javadoc/org/libvirt/Network.html
+        try {
+            Network network = conn.networkLookupByName("default");
+            s_logger.debug("Found libvirt default network, destroying it and setting autostart to false");
+            if (network.isActive() == 1) {
+                network.destroy();
+            }
+            if (network.getAutostart()) {
```
rhtyd: If we destroy the network and set autostart to false, would that cause an issue?

rhtyd:
@blueorangutan test

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
ustcweizhou: @rhtyd We have destroyed the default network and did not see any issue in our production.
it seems to break packaging/debian/cloudstack-agent.init, however, as /etc/init.d/cloudstack-agent is not in use any more after systemd changes, it should not be a problem.

wido:
I have never seen these issues, but I can understand why they happen.

LGTM

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
rhtyd: We still do support initd with centos6 packages.

ustcweizhou: @rhtyd 
I checked the code again, the packaging/debian/cloudstack-agent.init will be packaged in ubuntu/debs and deployed on ubuntu servers as /etc/init.d/cloudstack-agent. 
Since we do not support ubuntu 12.04 any more , and systemd is used in ubuntu 14.04 and later, I think this change will not break other functions.

rhtyd: LGTM

rhtyd:
@ustcweizhou can you rebase to latest master? We've only deprecated older Ubuntu versions recently, but not in 4.11.

ustcweizhou:
@rhtyd thanks for review. rebased with latest master.

rhtyd:
Thanks Wei
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3233, opened 2019-03-25
ui: don't ignore ''mine" when listing "all" templates in projects

rhtyd:
Problem: When uploading template or iso in a project, the progress is not seen
Root Cause: When users upload a template/iso in project view, the template/iso is not visible in the all filter. This creates confusion that template/iso has not been uploaded.
Solution: Since the api listtemplates with a projectid acts is a superset of list templates without a project id, we should not ignore the project id when listing templates in all filter.

## Types of changes


borisstoyanov:
@blueorangutan package

PR #3238, opened 2019-03-26
client: don't disable TLSv1, TLSv1.1 by default that breaks VMware env

rhtyd:
This fixes the issue that TLSv1 and TLSv1.1 are still used by CloudStack
management server to communicate with VMware vCenter server. With the
current defaults, the setup/deployment on VMware fails. Users/admins
can however setup the security file according to their env needs to
disable TLSv1 and TLSv1.1 for server sockets (8250/agent service for
example).

## Types of changes

### DIFF
--- a/client/conf/java.security.ciphers.in
+++ b/client/conf/java.security.ciphers.in
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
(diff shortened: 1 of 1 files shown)


rhtyd:
Without this fix, by default the env deployment fails for older versions of VMware such as 5.5 and 6.0 (less than up3).
@blueorangutan package

rhtyd:

PR #3240, opened 2019-03-28
api: instance and template details are free text

rhtyd:
Problem: Users don't know what keys/values to enter for template and VM details.
Root Cause: The feature does not exist that can list possible details and options.
Solution: Based on the possible VM and template details handled by the
codebase, those details were refactored and a list API is introduced
that can return users those details along with possible values. When
users add details now, they will be presented with a list of key details
and their possible options if any.

Introduces a new API:
```
listDetailOptions: Lists all possible details and their options for a resource type such as a VM or a template
Required params: resourcetype, 
API Params               Type     Description
==========               ====     ===========
resourceid               string   the UUID of the resource (optional)
resourcetype             string   the resource type such as UserVm, Template etc.
```

## Types of changes

Screenshot from 2019-03-27 18-12-18

### DIFF
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -3434,7 +3434,7 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
         options.put(VmDetailConstants.ROOT_DISK_SIZE, Collections.emptyList());
 
         if (HypervisorType.KVM.equals(hypervisorType)) {
-            options.put(VmDetailConstants.ROOT_DISK_CONTROLLER, Arrays.asList("ide", "scsi", "virtio"));
+            options.put(VmDetailConstants.ROOT_DISK_CONTROLLER, Arrays.asList("osdefault", "ide", "scsi", "virtio"));
         }
 
         if (HypervisorType.VMware.equals(hypervisorType)) {
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -25,6 +25,8 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import javax.inject.Inject;
 
@@ -3426,6 +3428,16 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
                 break;
             default:
                 throw new CloudRuntimeException("Resource type not supported.");
+        }
+        if (CallContext.current().getCallingAccount().getType() != Account.ACCOUNT_TYPE_ADMIN) {
+            final List<String> userBlacklistedSettings = Stream.of(QueryService.UserVMBlacklistedDetails.value().split(","))
+                    .map(item -> (item).trim())
+                    .collect(Collectors.toList());
+            for (final String detail: options.keySet()) {
+                if (userBlacklistedSettings.contains(detail)) {
+                    options.remove(detail);
+                }
+            }
         }
         return new DetailOptionsResponse(options);
     }
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -3433,8 +3433,8 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
             final List<String> userBlacklistedSettings = Stream.of(QueryService.UserVMBlacklistedDetails.value().split(","))
                     .map(item -> (item).trim())
                     .collect(Collectors.toList());
-            for (final String detail: options.keySet()) {
-                if (userBlacklistedSettings.contains(detail)) {
+            for (final String detail : userBlacklistedSettings) {
+                if (options.containsKey(detail)) {
                     options.remove(detail);
                 }
             }
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -3431,6 +3431,7 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
 
         options.put(VmDetailConstants.KEYBOARD, Arrays.asList("uk", "us", "jp", "fr"));
         options.put(VmDetailConstants.CPU_CORE_PER_SOCKET, Collections.emptyList());
+        options.put(VmDetailConstants.ROOT_DISK_SIZE, Collections.emptyList());
 
         if (HypervisorType.KVM.equals(hypervisorType)) {
             options.put(VmDetailConstants.ROOT_DISK_CONTROLLER, Arrays.asList("ide", "scsi", "virtio"));
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -17,7 +17,10 @@
 package com.cloud.api.query;
 
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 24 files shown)


borisstoyanov:
@blueorangutan package

borisstoyanov:
@rhtyd there's some build failures with this one, can you have a look? 
```
[INFO] 
[INFO] --- maven-clean-plugin:2.5:clean (default-clean) @ cloud-plugin-network-vcs ---
[INFO] Deleting /jenkins/workspace/acs-pr-centos6-pkg-builder/dist/rpmbuild/BUILD/cloudstack-4.13.0.0-SNAPSHOT/plugins/network-elements/brocade-vcs (includes = [target, dist], excludes = [])
[INFO] 
[INFO] --- maven-checkstyle-plugin:3.0.0:check (cloudstack-checkstyle) @ cloud-plugin-network-vcs ---
/opt/rh/maven30/root/usr/bin/mvn: line 9:  2041 Killed                  $M2_HOME/bin/mvn "$@"
error: Bad exit status from /var/tmp/rpm-tmp.ozlF7H (%build)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.ozlF7H (%build)
+ '[' 1 -ne 0 ']'
+ '[' false == true ']'
+ echo 'RPM Build Failed '
RPM Build Failed 
+ exit 3
Build step 'Execute shell' marked build as failure
Stopping all containers
Finished: FAILURE
```

rhtyd:
@borisstoyanov looks like an env issue, let's try again
@blueorangutan package

borisstoyanov:
@blueorangutan test

rhtyd:
Thanks for reporting the issue @borisstoyanov - the issue was missing role permission/authorization for the new API. I've fixed it.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
With https://github.com/apache/cloudstack/pull/3213, users won't be able to see the details admin decides to block. However, it's a presentation limitation only. With this FR, the user is only allowed a set of details that I thought can be allowed in most/all cases. (we can discuss/add/remove the details we want to export or not)

WIth https://github.com/apache/cloudstack/pull/3244, users will alway see the tab and that some details may always be shown to the user but not allowed to edit. This again is a presentation limitation, a smart use can still use the API/cmk.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -3399,6 +3407,50 @@ else if (!template.isPublicTemplate() && caller.getType() != Account.ACCOUNT_TYP
...
+    @Override
+    public DetailOptionsResponse listDetailOptions(final ListDetailOptionsCmd cmd) {
+        final ResourceObjectType type = cmd.getResourceType();
+        final String resourceUuid = cmd.getResourceId();
+        final Map<String, List<String>> options = new HashMap<>();
+        switch (type) {
+            case Template:
+            case UserVm:
+                HypervisorType hypervisorType = HypervisorType.None;
+                if (!Strings.isNullOrEmpty(resourceUuid) && ResourceObjectType.Template.equals(type)) {
```
DaanHoogland: shouldn't resourceUuid be validated in the cmd already?

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -3399,6 +3407,50 @@ else if (!template.isPublicTemplate() && caller.getType() != Account.ACCOUNT_TYP
...
+        }
+        return new DetailOptionsResponse(options);
+    }
+
+    private void fillVMOrTemplateDetailOptions(final Map<String, List<String>> options, final HypervisorType hypervisorType) {
+        if (options == null) {
+            throw new CloudRuntimeException("Invalid/null detail-options response object passed");
+        }
+
+        options.put(VmDetailConstants.KEYBOARD, Arrays.asList("uk", "us", "jp", "fr"));
```
DaanHoogland: I don't think we should maintain the possible values in the queryService. Rather ask them from the VO?

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -3399,6 +3407,50 @@ else if (!template.isPublicTemplate() && caller.getType() != Account.ACCOUNT_TYP
...
+        options.put(VmDetailConstants.CPU_CORE_PER_SOCKET, Collections.emptyList());
+
+        if (HypervisorType.KVM.equals(hypervisorType)) {
+            options.put(VmDetailConstants.ROOT_DISK_CONTROLLER, Arrays.asList("ide", "scsi", "virtio"));
+        }
+
+        if (HypervisorType.VMware.equals(hypervisorType)) {
+            options.put(VmDetailConstants.NIC_ADAPTER, Arrays.asList("E1000", "PCNet32", "Vmxnet2", "Vmxnet3"));
```
nvazquez: Similar as @DaanHoogland points out, what if we move these to a place in which we can read them and also perform validation?

borisstoyanov:
@rhtyd do you plan to address the comments?

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
rhtyd: Defensive programming, we shouldn't assume; even if upper layers may have validated it.

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
rhtyd: Actually all the options are wrt what's supported in this version of CloudStack, than something that can be accessed via a global setting etc. Which is why they are all hard-coded. Currently, we don't have any mechanism for the java code to query what the UI supports etc.

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
rhtyd: I'll try to address if this is possible.

rhtyd:
@borisstoyanov the comments are asking for refactoring, no changes in functionality. Yes, I'll try to address them.

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
rhtyd: I looked the code, I think I decided to hard-code because the enums are defined in VirtualEthernetCardType (vmware-base) and using the enum will add dependency hell as `api` should not depend on `vmware-base`. I  looked at other options as well, they all have this same problem, there is no easy way (without code-base wide refactoring) to find out what options are available for a detail/key.

rhtyd:
@borisstoyanov due to codebase limitations that will require a rather new set of refactorings, I'm not going to address the comments.

There is no easy way to centralise all the details and their options without hardcoding them in current design. Even if I write an interface that the sub-systems will implement, those options will partly spread in the sub-systems where they would be hard-coded. We can go with the hardcoded option values for now, and maybe address them in future.

borisstoyanov:
Thanks @rhtyd, proceeding with testing this as is.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

borisstoyanov:
@blueorangutan package

rhtyd:
Last change is UI only change to show all possible drop down options while editing an existing detail. No smoketests needed.
Screenshot from 2019-06-20 14-24-30

PaulAngus:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
Fixed few minor issues cc @borisstoyanov @andrijapanic - Kindly kick packaging and build new Trillian env to review. Thanks.

rhtyd:
@blueorangutan package

borisstoyanov:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Lgtm received from @andrijapanic - can you confirm on the PR. I'll merge this based on testing and review.

andrijapanicsb:
LGTM

PR #3247, opened 2019-03-28
Make the API documentation version not *hardcoded* to v4.9.0

GabrielBrascher:
Currently, the API documentation is hardcoded to _v4.9.0_. This commit makes the API documentation automatically updated to the current project version. If one wants to set another version, it is possible by just editing the `ACS_RELEASE` variable in _build-apidoc.sh_.

## Types of changes

Prior to the changes proposed in this PR the generated API documentaion was labeled as '4.9.0':
image

## How Has This Been Tested?

Project API documentation page:
image

Example of an API command documentation page:
image

### DIFF
--- a/tools/apidoc/generatecommands.xsl
+++ b/tools/apidoc/generatecommands.xsl
@@ -55,13 +55,12 @@ version="1.0">
 			<div class="api_titlebox">
 				<div class="api_titlebox_left">
 				<xsl:for-each select="command/command">
-					<!-- Modify this line for the release version -->
 					<span>
-									Apache CloudStack %ACS_RELEASE% Root Admin API Reference
-								</span>
-								<p></p>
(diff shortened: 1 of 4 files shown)


tools/apidoc/generatecommands.xsl
```diff
@@ -57,7 +57,7 @@ version="1.0">
 				<xsl:for-each select="command/command">
 					<!-- Modify this line for the release version -->
 					<span>
-									Apache CloudStack v4.9.0 Root Admin API Reference
+									Apache CloudStack v%ACS_RELEASE% Root Admin API Reference
```
rafaelweingartner: do we need/want the `v`? It is always `v` anyways.

GabrielBrascher: Good point. I agree with you @rafaelweingartner. Updated it!

PR #3275, opened 2019-04-14
[CLOUDSTACK-10408] Fix String.replaceAll() to replace() for bet…

bd2019us:
## Description

Fixes [#CLOUDSTACK-10408](https://issues.apache.org/jira/browse/CLOUDSTACK-10408)

Change String.replaceAll() to replace() when the replaced string is not a regular expression for better performance.

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)


DaanHoogland:
@bd2019us code lgtm. Do you have data on the performance improvements?

bd2019us:
Hi @DaanHoogland , I did not test the performance but analyzed theoretically. The code is further optimized by passing a char for replacement instead of string. The improvement sometimes can be small (only once) and sometimes can be huge (intensively invoked). Anyway, it will benefit the performance. Thanks.

framework/ipc/src/main/java/org/apache/cloudstack/framework/serializer/OnwireClassRegistry.java
```diff
@@ -109,7 +109,7 @@ public void scan() {
                     if (filePath != null) {
                         // WINDOWS HACK
                         if (filePath.indexOf("%20") > 0)
-                            filePath = filePath.replaceAll("%20", " ");
+                            filePath = filePath.replace("%20", " ");
```
DaanHoogland: @bd2019us did you intent to make this one a char as well, instead of " ", ' '?

bd2019us: We cannot since there does not exist an API to replace an `String` with a `char`. Therefore, I think it should be fine like it is. Thanks.

DennisKonrad:
@bd2019us Can you resolve the conflicts?
After that I would check and merge this PR.

DennisKonrad:
Just out of personal interest: You sure have some use case that was too slow without this. Would you mind sharing the problem this solves for you?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

andrijapanicsb:
@blueorangutan  test

andrijapanicsb:
@blueorangutan test

rhtyd:
I need to read the difference between the two methods, according to this https://stackoverflow.com/a/10827900; such a change can introduce subtle bugs cc @DaanHoogland @andrijapanicsb  - what do you think? Should we keep it for 4.14 or move to 4.15?

DaanHoogland:
@rhtyd the characters %20'-/" and the names of bridges are used in this PR, I don't see the regexp risk spoken about in that link. I have no objections to this

rhtyd:
Smoketest on vmware needed, I'll kick one
@blueorangutan package

DaanHoogland:
@blueorangutan test

framework/ipc/src/main/java/org/apache/cloudstack/framework/serializer/OnwireClassRegistry.java
```diff
@@ -109,7 +109,7 @@ public void scan() {
                     if (filePath != null) {
                         // WINDOWS HACK
                         if (filePath.indexOf("%20") > 0)
-                            filePath = filePath.replaceAll("%20", " ");
```
weizhouapache: we do not have a setup on windows. is it possible there are more than 1 "%20" ?

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -2326,9 +2326,9 @@ public UserAccount authenticateUser(final String username, final String password
 
                 // Block when is not in the list of allowed IPs
                 if (!NetUtils.isIpInCidrList(loginIpAddress, accessAllowedCidrs.split(","))) {
-                    s_logger.warn("Request by account '" + account.toString() + "' was denied since " + loginIpAddress.toString().replaceAll("/", "") + " does not match " + accessAllowedCidrs);
+                    s_logger.warn("Request by account '" + account.toString() + "' was denied since " + loginIpAddress.toString().replace("/", "") + " does not match " + accessAllowedCidrs);
```
weizhouapache: looks good. there is only 1 "/" in the string.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2951,7 +2951,7 @@ public String rebootVM(final Connect conn, final String vmName) {
                         final int vnetId = Integer.parseInt(nic.getBrName().replaceFirst("cloudVirBr", ""));
                         final String pifName = getPif(_guestBridgeName);
                         final String newBrName = "br" + pifName + "-" + vnetId;
-                        vmDef = vmDef.replaceAll("'" + nic.getBrName() + "'", "'" + newBrName + "'");
+                        vmDef = vmDef.replace("'" + nic.getBrName() + "'", "'" + newBrName + "'");
```
weizhouapache: looks good.
maybe nobody is still using old style bridge name "cloudVirBrXXX".

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
```diff
@@ -523,7 +523,7 @@ public Answer execute(VmwareHostService hostService, CreateVolumeFromSnapshotCom
 
         String details = null;
         boolean success = false;
-        String newVolumeName = UUID.randomUUID().toString().replaceAll("-", "");
+        String newVolumeName = UUID.randomUUID().toString().replace("-", "");
```
weizhouapache: we need to use replaceAll here.
The UUID contains 4 "-".

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
```diff
@@ -1051,7 +1051,7 @@ private String getVolumePathInDatastore(DatastoreMO dsMo, String volumeFileName,
             Integer nfsVersion) throws Exception {
 
         String volumeFolder = String.valueOf(volumeId) + "/";
-        String newVolume = UUID.randomUUID().toString().replaceAll("-", "");
+        String newVolume = UUID.randomUUID().toString().replace("-", "");
```
weizhouapache: same as above

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareStorageManagerImpl.java
DaanHoogland: @weizhouapache are you sure? according to https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#replace-java.lang.CharSequence-java.lang.CharSequence- the replace method already replaces each accurance. I might be overlooking versions, though.

weizhouapache: @DaanHoogland a. yes, you are right.

PR #3276, opened 2019-04-15
Allow VM that has never started to have volumes attached

anuragaw:
With this patch https://github.com/apache/cloudstack/commit/b766bf7fc9787b2516eee2eac6f00dd03a327b8b
we started tracking disks in attaching state so that other attach request can fail gracefully. However this missed the case where disks were in allocated state but attach was requested.

For the use case where users want to attach disk in allocated state but not ready, we need to have allocated-attaching transition as well. We must take care of returning to the original state - allocated or ready - when attach request has completed. 

For the use case of unstarted vm's the disk must proceed as follows - "Allocated" -> Attaching -> Allocated. When VM is started, the disk is "created" and pool is assigned. For the use case of started VMs it's more trivial and disk proceeds as follows - Ready -> Attaching -> Ready. 

Test this by creating a VM with "startvm=false", create a disk and try attaching it in allocated state. It would give an exception on latest 4.11 but will be fixed on this patch.

## Types of changes

## How Has This Been Tested?

Manually (using CloudMonkey, dashboard) and through Marvin Test

Allocated_Attached_Disk

### DIFF
--- a/api/src/com/cloud/storage/Volume.java
+++ b/api/src/com/cloud/storage/Volume.java
@@ -52,7 +52,7 @@ public interface Volume extends ControlledEntity, Identity, InternalIdentity, Ba
         UploadInProgress("Volume upload is in progress"),
         UploadError("Volume upload encountered some error"),
         UploadAbandoned("Volume upload is abandoned since the upload was never initiated within a specificed time"),
-        Attaching("The volume is attaching to a VM from Ready State");
+        Attaching("The volume is attaching to a VM");
 
         String _description;
 
@@ -119,10 +119,14 @@ public interface Volume extends ControlledEntity, Identity, InternalIdentity, Ba
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(UploadInProgress, Event.OperationTimeout, UploadError, null));
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(UploadError, Event.DestroyRequested, Destroy, null));
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(UploadAbandoned, Event.DestroyRequested, Destroy, null));
+
+            // Attaching can start from Allocated or Ready state. We need to track attaching and return to the original state.
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Allocated, Event.AttachRequested, Attaching, null));
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Ready, Event.AttachRequested, Attaching, null));
-            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.OperationSucceeded, Ready, null));
-            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.OperationFailed, Ready, null));
+            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.AttachFromAllocatedSucceeded, Allocated, null));
+            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.AttachFromAllocatedFailed, Allocated, null));
+            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.AttachFromReadySucceeded, Ready, null));
+            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.AttachFromReadyFailed, Ready, null));
         }
     }
 
@@ -145,7 +149,11 @@ public interface Volume extends ControlledEntity, Identity, InternalIdentity, Ba
         ExpungingRequested,
         ResizeRequested,
         AttachRequested,
-        OperationTimeout;
+        OperationTimeout,
+        AttachFromReadySucceeded,
+        AttachFromReadyFailed,
+        AttachFromAllocatedSucceeded,
+        AttachFromAllocatedFailed;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


anuragaw:
@blueorangutan package

anuragaw:
@blueorangutan test

rhtyd:
@anuragaw can you implement a marvin test for the test-case.

anuragaw:
@blueorangutan package

anuragaw:
> @anuragaw can you implement a marvin test for the test-case.

Done @rhtyd , let me know if this covers the case . Thanks!

api/src/com/cloud/storage/Volume.java
```diff
@@ -52,7 +52,8 @@
         UploadInProgress("Volume upload is in progress"),
         UploadError("Volume upload encountered some error"),
         UploadAbandoned("Volume upload is abandoned since the upload was never initiated within a specificed time"),
-        Attaching("The volume is attaching to a VM");
+        AttachingFromReady("The volume is attaching to a VM from Ready State"),
```
rhtyd: @anuragaw don't change the original state `Attaching` to honour db volume states that may already be in `Attaching` state in db. This may cause usage issues post an upgrade.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1920,12 +1920,13 @@ private Volume orchestrateDetachVolumeFromVM(long vmId, long volumeId) {
...
 
-            // volume.getPoolId() should be null if the VM we are detaching the disk from has never been started before
-            DataStore dataStore = volume.getPoolId() != null ? dataStoreMgr.getDataStore(volume.getPoolId(), DataStoreRole.Primary) : null;
-
-            volService.revokeAccess(volFactory.getVolume(volume.getId()), host, dataStore);
-
-            handleTargetsForVMware(hostId, volumePool.getHostAddress(), volumePool.getPort(), volume.get_iScsiName());
+            // volumePool() should be null if the VM we are detaching the disk from has never been started before
+            // only revoke access on volumes that are actually on a datastore
+            if (volumePool != null) {
```
rhtyd: Should the original logic be retained? What happens if volumePool is null?

api/src/com/cloud/storage/Volume.java
```diff
@@ -52,7 +52,8 @@
...
+        AttachingFromAllocated("The volume is attaching to a VM from Allocated State");
```
rhtyd: I think we can still fix the issue without introducing a new state. All we're saying is to allow volumes's attaching state whether their previous state is Ready or Allocated.

api/src/com/cloud/storage/Volume.java
```diff
@@ -119,9 +120,12 @@ public String getDescription() {
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(UploadInProgress, Event.OperationTimeout, UploadError, null));
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(UploadError, Event.DestroyRequested, Destroy, null));
             s_fsm.addTransition(new StateMachine2.Transition<State, Event>(UploadAbandoned, Event.DestroyRequested, Destroy, null));
-            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Ready, Event.AttachRequested, Attaching, null));
-            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.OperationSucceeded, Ready, null));
-            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Attaching, Event.OperationFailed, Ready, null));
+            s_fsm.addTransition(new StateMachine2.Transition<State, Event>(Allocated, Event.AttachRequested, AttachingFromAllocated, null));
```
rhtyd: Let's keep this only and rest can follow the same, i.e. allow both Ready and Allocated state volumes to go into Attaching state on the event.

test/integration/smoke/test_volumes.py
```diff
@@ -857,6 +857,60 @@ def test_10_list_volumes(self):
         self.assertTrue(hasattr(root_volume, "podname"))
         self.assertEqual(root_volume.podname, list_pods.name)
 
+    @attr(tags = ["advanced", "advancedns", "smoke", "basic"], required_hardware="true")
+    def test_11_attach_volume_with_unstarted_vm(self):
```
rhtyd: Test case LGTM. Also check Travis if this passes OK.

api/src/com/cloud/storage/Volume.java
anuragaw: The reason we need a new state is that if an attach fails we want the disk to return to the original state it was in - Ready -> Attaching -> FAIL -> Ready, or Allocated -> Attaching -> FAIL -> Allocated.

Perhaps I am misunderstanding something?

api/src/com/cloud/storage/Volume.java
anuragaw: Ah, my bad. Fixing.

server/src/com/cloud/storage/VolumeApiServiceImpl.java
anuragaw: If the volumePool was null initially the dataStore was null and revokeAccess did nothing. Also I don't think it should be null as the ```handleTargetsForVMWare(hostId,volumePool.getHostAddress()...``` would crash.

api/src/com/cloud/storage/Volume.java
anuragaw: Looking back at the original patch, it seems that simple Allocated->Attaching would suffice. I'll change.

api/src/com/cloud/storage/Volume.java
anuragaw: Agreed.

anuragaw:
@blueorangutan package

anuragaw:
@blueorangutan test

anuragaw:
@blueorangutan package

anuragaw:
@blueorangutan test

anuragaw:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

PR #3279, opened 2019-04-17
WIP: Skip getting used bytes for volumes that are not in Ready state

skattoju4:
## Description
Slow VM creation times are experienced when using solid fire managed storage. It is suspected that this is caused by calls to getUsedBytes in the solid fire storage driver that result in calls to the solid fire appliance when volume details are not available in the database. Volume details should be available in the database for volumes in ready state. We expect the error to be resolved by only attempting to get bytes used by active volumes by skipping volumes that are not in ready state.

Steps to reproduce
1. Create a large number of volumes in the database that are not in ready state.
2. Attempt to create a VM

Expected behaviour:
VM creation should happen in a less than a few minutes

Actual behaviour:
VM creation takes around 5 minutes even with small images

## Types of changes


svenvogel:
@skattju3 thanks for that! We will test it tomorrow.

skattoju4:
@syed could you take a look ?

skattoju4:
Further investigation revealed that the issue was caused by getUsedBytes taking a long time to compute the size of snapshots. This PR was closed automatically after the branch of renamed. This change is now being tracked by #3282

PR #3280, opened 2019-04-18
Remove code that generated /var/lib/libvirt/images/null on target host

GabrielBrascher:
## Description

This commit simplifies the `generateDestPath` method and fixes an issue where an extra file, named as '_null_', was created on the target storage pool during VM local storage volume migration. Without this fix, the VM is migrated and there is no data loss; however, 193 KB is allocated for the unused file named as '_null_' and the file stays on the target storage.

## Types of changes

### DIFF
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -40,15 +40,11 @@ import org.apache.log4j.Logger;
 import com.cloud.agent.api.Answer;
 import com.cloud.agent.api.MigrateCommand;
 import com.cloud.agent.api.MigrateCommand.MigrateDiskInfo;
-import com.cloud.agent.api.storage.CreateAnswer;
-import com.cloud.agent.api.storage.CreateCommand;
-import com.cloud.agent.api.to.VirtualMachineTO;
 import com.cloud.exception.AgentUnavailableException;
 import com.cloud.exception.OperationTimedoutException;
 import com.cloud.host.Host;
 import com.cloud.hypervisor.Hypervisor.HypervisorType;
 import com.cloud.storage.DataStoreRole;
-import com.cloud.storage.DiskOfferingVO;
 import com.cloud.storage.Storage.StoragePoolType;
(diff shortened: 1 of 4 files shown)


engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -111,28 +106,8 @@ protected StrategyPriority internalCanHandle(Map<VolumeInfo, DataStore> volumeMa
...
-        }
-
-        String libvirtDestImgsPath = null;
-        if (rootImageProvisioningAnswer instanceof CreateAnswer) {
-            libvirtDestImgsPath = ((CreateAnswer)rootImageProvisioningAnswer).getVolume().getName();
-        }
-        // File.getAbsolutePath is used to keep the file separator as it should be and eliminate a verification to check if exists a file separator in the last character of libvirtDestImgsPath.
-        return new File(libvirtDestImgsPath, destVolumeInfo.getUuid()).getAbsolutePath();
+    protected String generateDestPath(Host destHost, StoragePoolVO destStoragePool, VolumeInfo destVolumeInfo) {
+        return new File(destStoragePool.getPath(), destVolumeInfo.getUuid()).getAbsolutePath();
```
wido: You assume that all KVM storage is always a file. Although that is true for NFS and iSCSI, in Ceph's case this isn't.

I haven't tested it, but this might become a problem with Ceph storage. You might want to look at this and maybe implement a getPath() function per type of storage.

GabrielBrascher: @wido thanks for pointing that, you are right. We will need to do some changes to support live migrating VMs between Ceph storage pools.

So far, the internalCanHandle method handles only storage of the type 'Filesystem' or 'NetworkFilesystem'.

```
if (storagePoolVO.getPoolType() != StoragePoolType.Filesystem && storagePoolVO.getPoolType() != StoragePoolType.NetworkFilesystem) {
}
```

rhtyd:
@blueorangutan package

GabrielBrascher:
@blueorangutan test

GabrielBrascher:
@rhtyd could you please run @blueorangutan test? Thanks!

rhtyd:
@blueorangutan test

PR #3282, opened 2019-04-19
Fix slow vm creation when large sf snapshot count

skattoju4:
## Description
Slow VM creation times are experienced when using solid fire managed storage. It is suspected that this is caused by calls to getUsedBytes in the solid fire storage driver. Storage used by snapshots is not being computed efficiently resulting in a large number of DB queries.

getUsedBytes is called a number of times during the vm creation process. This PR proposes to refactor it so that a constant number of queries are used to compute it. We do one query each to the volumes table, the snapshot_store_ref table and the template_spool_ref table instead of querying the respective *_details tables by object id to get the size of each type of object.

Steps to reproduce:
Attempt to create a VM in a Cloudstack instance that has a large number of snapshots
Expected behaviour:
VM creation should happen in a less than a few minutes

Actual behaviour:
VM creation takes around 5 minutes even with small images

## Types of changes

## How Has This Been Tested?

The change was tested in a cloudstack environment which had a large number of snapshots. Vm creation time was significantly improved since counts are being done in memory after getting all volumes, snapshots and templates in single queries.

## Further work
VM creation is still not as fast as with NFS when creating VMs from ISO templates. Further investigation is needed but this is out side the scope of this PR.


skattoju4:
@svenvogel @syed could you please take a look ?

syed:
@skattoju4 Can you also add the check for template as well?

svenvogel:
@skattoju4 template?

skattoju4:
Hi Sven, it was a holiday yesterday. Will add that today.

DennisKonrad:
Hi @skattoju4,

I just tested and the startup time was 7 minutes -.-
Seems this is not solving the problem.

engine/schema/src/main/java/org/apache/cloudstack/storage/datastore/db/SnapshotDataStoreDao.java
```diff
@@ -30,6 +30,8 @@
 
     List<SnapshotDataStoreVO> listByStoreId(long id, DataStoreRole role);
 
+    List<SnapshotDataStoreVO> listByStoreIdInReadyState(long id);
```
syed: Can you make this a generic function `listByStoreIdAndState`

skattoju4: yup fixed. wasn't sure why state was mapped to ObjectInDataStoreStateMachine.State and not Snapshot.State ..

syed: Both track different things. `SnapshotState` is more for the end snapshot object whereas `ObjectInDataStore` tracks the status of the snapshot on a particular storage pool

svenvogel:
@skattoju3 @syed we will test it now and come back!

svenvogel:
@syed @skattoju3 we test it! its seems faster but now 1 minute. we dont know why but thats long too.

skattoju4:
@svenvogel @syed as discussed / tested getUsedBytes is no longer contributing significantly to slowness in VM creation. Further investigation is needed to determine why VM creation from ISOs is sometimes slow. Further fixes can be addressed in separate PRs. If you guys agree this one is ready be merged.

svenvogel:
@skattoju4 lets take the other changes to a new PR. LGTM

svenvogel:
@syed can you approve, then we can merge it?

syed:
@skattoju4  LGTM. Can you remote the WIP in the title. This is good to merge :+1:

skattoju4:
removed, its just in the email thread ...

DennisKonrad:
This LGTM. After the recent changes this works like a charm

svenvogel:
@rhtyd can you merge it?

svenvogel:
Fix tested in production with managed storage (netapp solidfire). there are no dependencies to other bugs.

Issue #3287, opened 2019-04-23
Secondary storage pool 90pct limit ignored


DagSonsteboSB:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
* Secondary storage pool / SSVM

##### CLOUDSTACK VERSION
* Any

##### CONFIGURATION
* Any

##### OS / ENVIRONMENT
* N/A

##### SUMMARY
* Secondary storage pools should in theory not be considered for template / volume / ISO / snapshot upload if percentage utilisation is > 90%
* However if no other pool exist < 90% then the first available secondary pool will be chosen and the 90% utilisation limit ignored.
* This results in secondary storage pools potentially filling up until copy process fails, potentially leading to data loss
* In the following line secondary pools below 90% utilisation are returned:
https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java#L155
* However if no secondary storage pool exists <90% then the first secondary storage pool is still picked: https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java#L160

##### STEPS TO REPRODUCE
* Create zone with single secondary pool and fill this to above 90%
* Try to upload template / ISO / volume and observe this succeeding

##### EXPECTED RESULTS
* Upload process should gracefully fail.

##### ACTUAL RESULTS
* Upload process allows for upload to happen above 90% utilisation

rhtyd:
Dupes https://github.com/apache/cloudstack/issues/3041 closing

PR #3293, opened 2019-04-24
Limit selecting of secondary storage with 90% utilization

kioie:
## Description

Upload should gracefully fail when the secondary storage selected has 90% utilization, even if its the only available storage.

Fixes: #3287

## Types of changes

## How Has This Been Tested?

Create zone with single secondary pool and fill this to above 90%
Try to upload template / ISO / volume and observe this should fail

### DIFF
--- a/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
+++ b/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
@@ -157,6 +157,7 @@ public class ImageStoreProviderManagerImpl implements ImageStoreProviderManager 
                 }
             }
         }
-        return imageStores.get(0);
+        s_logger.debug("Can't find staging storage in zone with less than 90% usage");
+        return null;
     }
 }


engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
```diff
@@ -157,6 +157,7 @@ public DataStore getImageStore(List<DataStore> imageStores) {
                 }
             }
         }
-        return imageStores.get(0);
+        s_logger.debug("Can't find staging storage in zone with less than 90% usage");
```
DaanHoogland: isn't this threshold value meant to be configurable? change makes sense if not!
/cc @DagSonsteboSB @rhtyd

kioie: The idea is that it shouldn't be.

anuragaw:
Simply returning ```null``` could possibly result in newer bugs? ( On a quick inspection it seems that the method consumers would not always check for null values ).

I wonder if more testing scenarios than mentioned on the PR should be covered. Thoughts?

kioie:
> Simply returning `null` could possibly result in newer bugs. ( On a quick inspection it seems that the method consumers would not always check for null values ).
> 
> I wonder if more testing scenarios than mentioned on the PR should be covered. Thoughts?

I debated with the thought of editing the method consumers to handle null returns intrinsically, but I noticed in most cases (at least to the best of my knowledge) null returns are eventually handled by level 2 consumers. I'd be curious to try out other related test cases

anuragaw:
I can suggest doing some more tests with storage to max capacity -
1. Creating snapshots (VM and Storage)
2. Launching VMs
3. Migrating Volumes

The only place I saw where the null check was not carried was in `SecondaryStorageManagerImpl`-
```    @Override
    public boolean finalizeVirtualMachineProfile(VirtualMachineProfile profile, DeployDestination dest, ReservationContext context) {
        SecondaryStorageVmVO vm = _secStorageVmDao.findById(profile.getId());
        Map<String, String> details = _vmDetailsDao.listDetailsKeyPairs(vm.getId());
        vm.setDetails(details);

        DataStore secStore = _dataStoreMgr.getImageStore(dest.getDataCenter().getId());
        assert (secStore != null);```

I hope it helps.

engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
shwstppr: @kioie 90% cutoff value is defined here https://github.com/apache/cloudstack/blob/7e4a300a216a31348d2adbcc641d26e8339da205/server/src/main/java/com/cloud/server/StatsCollector.java#L290 and it can change in the future. Would it be better to not use constant in logs?

svenvogel:
@kioie some news here?

kioie:
@svenvogel let me work on it tonight

rhtyd:
Thank you for your PR. But it's incomplete and given we're closer to the freeze, I'll take over and submit issues around secondary storage selection affecting:
https://github.com/apache/cloudstack/issues/3041
https://github.com/apache/cloudstack/issues/3287
https://github.com/apache/cloudstack/issues/3478

kioie:
No problem

PR #3294, opened 2019-04-25
Set default service offering over api and ui

pbankonier:
## Description
This adds the possibility to add/change the defaultUse/unique name of system service offerings over the ui and api.

The create and update command of the system service offerings now have an optional parameter for the default use.

The default use parameter is a boolean value.

When setting a a new service offering or when editing an existing one to be a default one, the unique name and default use value of the offering which currently had it gets set to NULL.

The unique name depends on the system vm type, localstorage and defaultuse values.

## Types of changes

## Screenshots

In the list it gets shown if the service offering is a default one.

Bildschirmfoto 2019-04-25 um 11 25 52

You can now set the offering to a default one when creating a new system service offering.

Bildschirmfoto 2019-04-24 um 13 38 20

You can set/unset the service offering as a default one in the edit view.

Bildschirmfoto 2019-04-24 um 13 38 46

Bildschirmfoto 2019-04-24 um 13 39 14

I also added the possibility to create a elastic lb here because it was missing.

Bildschirmfoto 2019-04-24 um 13 39 40

## How Has This Been Tested?


pbankonier:
I opened a new PR because I had some conflicts when trying to push my changes to the old one.
https://github.com/apache/cloudstack/pull/3198

engine/schema/src/main/java/com/cloud/service/dao/ServiceOfferingDaoImpl.java
```diff
@@ -289,4 +294,33 @@ public ServiceOfferingVO findDefaultSystemOffering(String offeringName, Boolean
...
+
+    @Override
+    public void resetUniqueName(long id, String uniqueName){
+      SearchCriteria<ServiceOfferingVO> sc = IdSearch.create();
+      sc.setParameters("id", id);
+      List<ServiceOfferingVO> vos = search(sc, null, null, false);
+      if (vos.size() != 0) {
+        ServiceOfferingVO seVo = vos.get(0);
+        seVo.setUniqueName(uniqueName);
+        seVo.setDefaultUse(true);
```
DaanHoogland: does a reset of uniquename always mean this will be the default offering?

DaanHoogland:
@svenvogel code looks alright but please explain how uniquename and defaultuse are connected.

engine/schema/src/main/java/com/cloud/service/dao/ServiceOfferingDaoImpl.java
pbankonier: I implemented this function so the uniquename and default status of the previously default service offering gets reset when persisting the new one fails.

pbankonier:
> @svenvogel code looks alright but please explain how uniquename and defaultuse are connected.

@DaanHoogland The unique name is a combination of the "default use", "use local storage" and system vm type of the offering.

svenvogel:
@DaanHoogland can you approve?

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

ustcweizhou:
I would prefer to add a global setting to specify the default service offering for routers.
The service offering will be determined in findServiceOfferingId() in RouterDeploymentDefinition.java and VpcRouterDeploymentDefinition.java

If you do not mind, I can create a PR for it.

DaanHoogland:
@pbankonier @svenvogel , I agree with @ustcweizhou . It would be better to reduce this PR to only enabling setting the uniquename and to let a global setting point to the default offering. The functionality like this seems clunky. Also we may be changing names against the convention that someone might want to maintain. What are your thoughts?

svenvogel:
@DaanHoogland we dont agree. we would prefer this functionality. we dont think its somehow like "clunky". "clunky" was it implemented before. we think it should be easier so set it for all types of system vms. this feature allow to set different type of system vm images. to get this a graphical menu is more user friendly than one value for each type of system vm in global settings. at the moment its a 

> I would prefer to add a global setting to specify the default service offering for routers.
> The service offering will be determined in findServiceOfferingId() in RouterDeploymentDefinition.java and VpcRouterDeploymentDefinition.java
> 
> If you do not mind, I can create a PR for it.

what this means its determined in this functions? is there any functionality they breaks this? why it should set this in global setting when all other things for "offering" are set there?

svenvogel:
> Also we may be changing names against the convention that someone might want to maintain.

@DaanHoogland which function? which convention? please declare it exactly what you mean. at the moment we dont think we break anything.

pbankonier:
@DaanHoogland @ustcweizhou The uniquename gets set seperate per every system service offering. I don't understand how you want to realize this with a global setting maybe you can explain this in more detail.

Here are some screens of the affected db tables. In the first one you see the service_offering table and in the second one the disk offerings. I marked the affected columns/entries.

Bildschirmfoto 2019-05-03 um 16 49 09

Bildschirmfoto 2019-05-03 um 16 49 46

ustcweizhou:
> @DaanHoogland we dont agree. we would prefer this functionality. we dont think its somehow like "clunky". "clunky" was it implemented before. we think it should be easier so set it for all types of system vms. this feature allow to set different type of system vm images. to get this a graphical menu is more user friendly than one value for each type of system vm in global settings. at the moment its a
> 
> > I would prefer to add a global setting to specify the default service offering for routers.
> > The service offering will be determined in findServiceOfferingId() in RouterDeploymentDefinition.java and VpcRouterDeploymentDefinition.java
> > If you do not mind, I can create a PR for it.
> 
> what this means its determined in this functions? is there any functionality they breaks this? why it should set this in global setting when all other things for "offering" are set there?

@svenvogel 
Currently we have global setting consoleproxy.service.offering for CPVM and secstorage.service.offering for SSVM, no code change is needed. 
Of course it would be better to change it from global setting to zone setting, so zones can have different offering for systemvms f there are multiple zones created.

We do not have setting for router vms. I prefer to add an account setting (eg router.service.offering" so we can set offering of router globally and also for individual account. it is a simple change similar as consoleproxy.service.offering and secstorage.service.offering

I understand that you have spent some time on this PR and it might be running on your production already. I agree that it is user friendly. Considering It will not break any functionality, this PR is ok for me.

svenvogel:
@ustcweizhou can you approve it?

pbankonier:
@ustcweizhou @svenvogel @DaanHoogland The motivation behind this was to make it possible to set a default system offering over the ui instead of the way described at the bottom of this page http://docs.cloudstack.apache.org/projects/cloudstack-administration/en/4.8/service_offerings.html.

PaulAngus:
@svenvogel @pbankonier - I can see that you've put a lot of work into this, but unless I've missed something, couldn't this be done a bit simpler by populating the service_offering_id for all of the default network offerings as: 9 or 10 (shared/local storage offerings for domainrouter).

Then the network offering would be linked to an actual service offering rather than the weird global settings business.

Migrating from the global settings if people have updated them wouldn't be difficult, the service offerings 9 and 10 would just have to be updated with the values in the global settings.

Editing the material properties of a offering has always been a no-no (with very good reasons) but maybe we can make an exception for the system vm offerings to allow change CPU and RAM values.
// @ustcweizhou

svenvogel:
@PaulAngus thanks for info i will discuss this with @pbankonier and come back

svenvogel:
@PaulAngus can you make it more clear what you mean with "service_offering_id" 9 and 10?

PaulAngus:
image

In a base installation there are domainrouter offerings (which i don't think are actually used)

i believe that they always have IDs 8 and 9

svenvogel:
> @svenvogel @pbankonier - I can see that you've put a lot of work into this, but unless I've missed something, couldn't this be done a bit simpler by populating the service_offering_id for all of the default network offerings as: 9 or 10 (shared/local storage offerings for domainrouter).

i try to understand this... maybe we dont have them anymore. you mean populating the service_offering_id. what do you mean practically with this? after some changes you will have different ids because you have different ram and cpu sizes each linked to a domainrouter. the only difference is the default use flag.


image


> Migrating from the global settings if people have updated them wouldn't be difficult, the service offerings 9 and 10 would just have to be updated with the values in the global settings.

you mean that these values are always fix? whats with different sizes? whats the difference between globalsettings vs http://docs.cloudstack.apache.org/projects/archived-cloudstack-administration/en/latest/service_offerings.html database hack?

PaulAngus:
If we _actually_ use the system service offerings 'properly' the global settings 'default' becomes redundant.  We should probably make specifying a service offering for a network offering mandatory.

I guess my idea is to do one 'hack' to end them all and connect network offerings to service offerings properly and then update the API to allow operators to change the service offering of a network offering.
 
To be honest, i'm also struggling to see why we separate out the offerings wrt vm_type - CPVM, SSVM, VR etc.  Sure maybe _name_ them differently, there should probably be VPC independently too,  but 2G RAM 2vCPU is the same in a CPVM, SSVM or VR.

svenvogel:
Yes you are right at the moment we have two way to set default values

1. global settings
2. this PR

First of all we introduced this to make it possible to use the UI to change the service offering for the routers not to use this ugly database hack. before you could set it in the global settings and the hack. now you can set in in the UI. we focused on not to break the functionality. there are different tables. 

"disk offering" and "service offering" they are connected. there are a "unique_name" and "default_use" ... they seems to be connected. 

first we thought to rewrite them but there is a lot of work to do and it would break the existing functionality.

okay here is the plan,

1. we implement this PR
2. we and our team will work on all our issues and we will improve and change them step by step accordingly. we remove in a second PR the global settings 

_it can be under discussion to change/adapt the "network offering" "but" i have to admit  "network offering" the present network offerings meet all recent requirements. if we change them we would create a huge and complex beast. these mean, change the database and the UI. its a lot of work._

at the moment i think the System Offering is the best location to set up values.
image
the main problem here is the awful connection between the "disk offering" "service offering" because in "Service Offering" there will be saved CPU, RAM and so one...

lets try to work out best solution...

PaulAngus:
well, since you ask....

My view on _that_ is that a service offering should specify and disk offering. And that disk size should be decoupled from disk offering.  Disk offering performance will probably have to offer performance as either 'per GB' or absolute values ie: 1000 IOPS or 100 IOPS/GB

Maybe an operator would want to specify some preset sizes, which could then optionally be included in a service offering.

svenvogel:
@PaulAngus just so I understand it correctly...

1. add the disk offering from for domain router, CPVM, SSVM ... to the service offering table. right?
2. remove them, after unused from disk offering?

is that correct?

i need to add that if you have a managed storage the VPC, SSVM, CPVM will be created with...
image

i dont know if this is hard coded...

rafaelweingartner:
@pbankonier and @svenvogel thanks for the PR and your valuable contributions. I have read the discussions, but still, have a question here.

What does this change mean for the current method we have to override/change the default service offerings for system VMs? I mean, currently we have the following configurations available:
* consoleproxy.service.offering
* internallbvm.service.offering
* secstorage.service.offering

I do understand that you are introduced a method to change these default for other types of VMs, such as VRs, but still, how will this play out for people that are already used to use the current global settings parameters to configure the service offering for console proxy, ssvm, and internal LB VM?

I do like the UI change you are proposing, but I see that it can be implemented with the same backend we have. Of course, you are now supporting the override for other types of system VMs as well, but I guess that we could just use the "standard" we have currently in place.

Now, if the proposal is to change this "standard", then we need to create/propose a migration path for people.

svenvogel:
@rafaelweingartner thanks for the comment.

Yes, the proposal is to change this "standard" and add a to create/propose a migration path for people who use the default settings.

rafaelweingartner:
@svenvogel can you describe how this migration path would work for people that are currently using the "configurations" table when they start to use the code introduced by this PR?

pbankonier:
@svenvogel @DennisKonrad @rafaelweingartner @PaulAngus looked into the code for the global default offering settings and found that the only place where this global config variables are used is in the ManagerImpls for the different offering types.
The global settings are just used to check if there is already an offering with the given id existing in the database. There is no code which sets an offering to a default one with this configs. So maybe this was implemented to prevent that a new default offering was created at the start of the application before the uniquenames were introduced.

There is no need for a migration path with this PR. There will be no changes which will break existing functionality.
After my analysis the global default settings do nothing in the current state of the software and we can remove them.

svenvogel:
@rafaelweingartner ?

rhtyd:
Ping @rafaelweingartner @ustcweizhou @PaulAngus - any comments?

rafaelweingartner:
@pbankonier what happens if I changed the value of `ConsoleProxyServiceOffering` configuration? How will the new service offering I can set there work with your proposal? Will it be used? Will this configuration still be available in the configuration view?

rhtyd:
Ping?

borisstoyanov:
@blueorangutan package

rhtyd:
Looks like there are still on-going discussions and disagreement around the approach. I've moved this to the next major milestone 4.14.0.0 now.

pbankonier:
@rafaelweingartner As far as I saw the configuration in the global settings can and should be removed because it does not affect anything. Currently the uniquename gets used to determine which offering should be used as the default one.

svenvogel:
@rhtyd closed in favor of cloudstack primate development https://github.com/apache/cloudstack-primate/

Issue #3178, opened 2019-02-13
GUI project pulldown list slowdown on large number of projects


DagSonsteboSB:
##### ISSUE TYPE
 * Improvement Request

##### COMPONENT NAME
* ACS GUI

##### CLOUDSTACK VERSION
* Reproduced on ACS 4.11.2

##### CONFIGURATION
* ACS 4.11.2
* Problem is environment agnostic

##### OS / ENVIRONMENT
* ACS management on CentOS7

##### SUMMARY
* When number of projects is large (thousands) and there's a lot of tags in use (hundreds) then the ACS GUI pulldown menu slows down drastically to the point it sometimes fails to show the project list
* Initial code investigation shows this could be down to:
** Code looks up all projects - active and inactive
** Then runs multiple loops against tags
** As a result processing is carried out with inactive projects included
* Code investigation shows this procedure could be made more efficient by only processing active project thereby cutting down on tag lookup processing time.

##### STEPS TO REPRODUCE
* Create thousands of projects, leave a handful active, the rest inactive
* Tag the projects with 10-20 tags each
* Use project pulldown menu in CloudStack GUI

##### EXPECTED RESULTS
* Relatively quick pulldown menu population

##### ACTUAL RESULTS
* Pulldown menu population can take 10-20 seconds to populate and will sometimes not populate at all (assuming due to timeout somewhere)

JnSchl:
Hi, we encounter the same behavior but with less projects (in the small three-digit range). Also non root users are affected by the longer loading times if only their projects are listed.

voteforpresident:
We are experiencing the same issue on the Cloudstack Version 4.11.2.

ustcweizhou:
We had similar issue on listing vms if vms have several tags.
We fixed it by recreating view without resource tag, and get resource tags by a query in server/src/com/cloud/api/query/dao/UserVmJoinDaoImpl.java.
```
         // update tag information
-        long tag_id = userVm.getTagId();
-        if (tag_id > 0 && !userVmResponse.containTag(tag_id)) {
-            ResourceTagJoinVO vtag = ApiDBUtils.findResourceTagViewById(tag_id);
-            if (vtag != null) {
-                userVmResponse.addTag(ApiDBUtils.newResourceTagResponse(vtag, false));
-            }
+        List<ResourceTagJoinVO> tags = ApiDBUtils.listResourceTagViewByResourceUUID(userVm.getUuid(), ResourceObjectT
+ype.UserVm);
[...]

PR #3306, opened 2019-04-30
server: reduce execution time while listing project if projects have many resource tags

ustcweizhou:
## Description

If projects have many resource tags, it will take long time to list projects.
Remove resource tags information from project_view will fix it the issue.

Fixes: #3178

## Types of changes

## How Has This Been Tested?

create project,
add resource tags to project,
listprojects  lists all resouce tags.

### DIFF
--- a/server/src/main/java/com/cloud/api/query/dao/ProjectJoinDaoImpl.java
+++ b/server/src/main/java/com/cloud/api/query/dao/ProjectJoinDaoImpl.java
@@ -32,6 +32,7 @@ import com.cloud.api.query.vo.AccountJoinVO;
 import com.cloud.api.query.vo.ProjectJoinVO;
 import com.cloud.api.query.vo.ResourceTagJoinVO;
 import com.cloud.projects.Project;
+import com.cloud.server.ResourceTag.ResourceObjectType;
 import com.cloud.user.Account;
 import com.cloud.user.dao.AccountDao;
 import com.cloud.utils.db.GenericDaoBase;
@@ -81,12 +82,9 @@ public class ProjectJoinDaoImpl extends GenericDaoBase<ProjectJoinVO, Long> impl
         response.setOwner(proj.getOwner());
 
         // update tag information
-        Long tag_id = proj.getTagId();
-        if (tag_id != null && tag_id.longValue() > 0) {
-            ResourceTagJoinVO vtag = ApiDBUtils.findResourceTagViewById(tag_id);
-            if (vtag != null) {
-                response.addTag(ApiDBUtils.newResourceTagResponse(vtag, false));
-            }
+        List<ResourceTagJoinVO> tags = ApiDBUtils.listResourceTagViewByResourceUUID(proj.getUuid(), ResourceObjectType.Project);
+        for (ResourceTagJoinVO vtag : tags) {
+            response.addTag(ApiDBUtils.newResourceTagResponse(vtag, false));
         }
 
         //set resource limit/count information for the project (by getting the info of the project's account)
@@ -101,14 +99,6 @@ public class ProjectJoinDaoImpl extends GenericDaoBase<ProjectJoinVO, Long> impl
 
     @Override
     public ProjectResponse setProjectResponse(ProjectResponse rsp, ProjectJoinVO proj) {
-        // update tag information
-        Long tag_id = proj.getTagId();
-        if (tag_id != null && tag_id.longValue() > 0) {
-            ResourceTagJoinVO vtag = ApiDBUtils.findResourceTagViewById(tag_id);
-            if (vtag != null) {
-                rsp.addTag(ApiDBUtils.newResourceTagResponse(vtag, false));
-            }
-        }
         return rsp;
     }
 
--- a/server/src/main/java/com/cloud/api/query/dao/ProjectJoinDaoImpl.java
+++ b/server/src/main/java/com/cloud/api/query/dao/ProjectJoinDaoImpl.java
@@ -98,11 +98,6 @@ public class ProjectJoinDaoImpl extends GenericDaoBase<ProjectJoinVO, Long> impl
     }
 
     @Override
-    public ProjectResponse setProjectResponse(ProjectResponse rsp, ProjectJoinVO proj) {
-        return rsp;
-    }
-
-    @Override
     public List<ProjectJoinVO> newProjectView(Project proj) {
         SearchCriteria<ProjectJoinVO> sc = prjIdSearch.create();
         sc.setParameters("id", proj.getId());
--- a/engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
@@ -74,3 +74,33 @@ CREATE VIEW `cloud`.`data_center_view` AS
         `cloud`.`dedicated_resources` ON data_center.id = dedicated_resources.data_center_id
             left join
         `cloud`.`affinity_group` ON dedicated_resources.affinity_group_id = affinity_group.id;
+
+-- Remove key/value tags from project_view
+DROP VIEW IF EXISTS `cloud`.`project_view`;
+CREATE VIEW `cloud`.`project_view` AS
+    select
+        projects.id,
+        projects.uuid,
+        projects.name,
+        projects.display_text,
+        projects.state,
+        projects.removed,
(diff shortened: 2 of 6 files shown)


anuragaw:
If the issue is mostly prevalent in UI, I would like to propose an alternate - 
1) Add a parameter ```showTags``` to ```list projects``` API
2) Set the flag to false to list project call in GUI.

Although if not many users need the tags as part of response this is perfectly reasonable too.

svenvogel:
@ustcweizhou we will test it and come back :)

ustcweizhou:
> If the issue is mostly prevalent in UI, I would like to propose an alternate -
> 
> 1. Add a parameter `showTags` to `list projects` API
> 2. Set the flag to false to list project call in GUI.
> 
> Although if not many users need the tags as part of response this is perfectly reasonable too.

@anuragaw In my point of view,  it would be preferable to fix an issue without api change.

DaanHoogland:
not sure if the 'standard' travis smoke tests cover enough of this functionality.
@blueorangutan package

DaanHoogland:
@blueorangutan test

svenvogel:
@ustcweizhou we are tested it and it seems and feels faster but it seems there is still a small delay. is there anything to make it more faster?

ustcweizhou:
@svenvogel we have similar issue while listing accounts. for project, it can be improved by
```
diff --git a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
index 91e0466..2b15113 100644
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -1298,7 +1298,7 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
         Long accountId = null;
         String path = null;

-        Filter searchFilter = new Filter(ProjectJoinVO.class, "id", false, startIndex, pageSize);
+        Filter searchFilter = new Filter(ProjectJoinVO.class, null, false, startIndex, pageSize);
         SearchBuilder<ProjectJoinVO> sb = _projectJoinDao.createSearchBuilder();
         sb.select(null, Func.DISTINCT, sb.entity().getId()); // select distinct
         // ids
```

Do not order project_view will reduce execution time of db query.

svenvogel:
@ustcweizhou oh i see. hmm should we add this to this pr?

svenvogel:
@ustcweizhou we will pick up your change ne try to fix the sort of the project list with jQuery.

DennisKonrad:
@ustcweizhou @svenvogel  I built and tested both the PR and the speed improvemennt for the db query (not ordering project_view). Although there seems to be a slight improvement in time for us it still takes way too long. It always feels like you miss-clicked and users then tend to click two or three times (which ends in misery because the lists closes with the second click)

ustcweizhou:
@DennisKonrad did you recreate project_view in your testing ?

DennisKonrad:
@ustcweizhou Yes, i executed the schema update. I think there is an issue about the project listing that's not about tags used. We have about ~30 projects now and this seems to be too much

ustcweizhou:
@DennisKonrad can you login database to see how many projects do you use and how many records in project_view ?
If you do not use tag , the root cause should be different .

DennisKonrad:
@ustcweizhou We have ~40 entries and 30 projects that are selectable in the list. Our database is very fast so the queries should not be a problem.

Do you see any improvments that can be done to the project listing code that would make it faster?

ustcweizhou:
@DennisKonrad 
I think it might be caused by the queries to get resource limit of project account.
I will have a look.

ustcweizhou:
@DennisKonrad I will raise a PR to fix the issue in listProjects and listAccounts.
What cloudstack version do you use ?

DennisKonrad:
@ustcweizhou We more or less use the master branch. So you can base the PR on that if you want. Thanks

ustcweizhou:
@DennisKonrad PR #3331 to apache/master

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

PaulAngus:
Guys,
personally I suspect this is just hiding the problem, and doesn't help anyone who does want to see tags.

I created the followingissue a few days ago https://github.com/apache/cloudstack/issues/3375
I suspect that this looping through accounts in a project is causing the delay.

??

ustcweizhou:
@PaulAngus I think the issue #3375 can be fixed.

in https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/api/query/ViewResponseHelper.java#L201-L211

with this PR, fillProjectDetails will do nothing, so the tags will be set only once.
```
    public static ProjectResponse fillProjectDetails(ProjectResponse rsp, ProjectJoinVO proj) {
        return s_projectJoinDao.setProjectResponse(rsp, proj);
    }
```

rhtyd:
Can you review and test this @anuragaw

rhtyd:
Just to confirm this does not break returning of tags in the API response, just the mechanics of how they are found/used @ustcweizhou ? cc @PaulAngus

rhtyd:
@blueorangutan package

ustcweizhou:
> Just to confirm this does not break returning of tags in the API response, just the mechanics of how they are found/used @ustcweizhou ? cc @PaulAngus

in my testing , tags are still in the api response and same as before.

rhtyd:
Thanks @ustcweizhou I'll kick some test and review/lgtm it soon.

rhtyd:
@blueorangutan test

server/src/main/java/com/cloud/api/query/dao/ProjectJoinDaoImpl.java
```diff
@@ -101,14 +99,6 @@ public ProjectResponse newProjectResponse(ProjectJoinVO proj) {
 
     @Override
     public ProjectResponse setProjectResponse(ProjectResponse rsp, ProjectJoinVO proj) {
```
anuragaw: This method seems to be called in the following path -

`ViewResponseHelper#createProjectResponse() > ApiDBUtils#fillProjectDetails() > ProjectJoinDaoImpl#setProjectResponse` 

I am a little confused by the necessity of this method call at all. Read the below method definitions-


```

ApiDBUtils {
    public static ProjectResponse fillProjectDetails(ProjectResponse rsp, ProjectJoinVO proj) {
        return s_projectJoinDao.setProjectResponse(rsp, proj);
    }
}

ViewResponseHelper {
    public static List<ProjectResponse> createProjectResponse(ProjectJoinVO... projects) {
        Hashtable<Long, ProjectResponse> prjDataList = new Hashtable<Long, ProjectResponse>();
        // Initialise the prjdatalist with the input data
        for (ProjectJoinVO p : projects) {
            ProjectResponse pData = prjDataList.get(p.getId());
            if (pData == null) {
                // first time encountering this vm
                pData = ApiDBUtils.newProjectResponse(p);
            } else {
                // update those  1 to many mapping fields
                pData = ApiDBUtils.fillProjectDetails(pData, p);
            }
            prjDataList.put(p.getId(), pData);
        }
        return new ArrayList<ProjectResponse>(prjDataList.values());
    }
}
```

Why not simply delete this path and modify the ViewResponseHelper to generate the projects response list at one go?

Apologies if I am missing something obvious?

ustcweizhou: @anuragaw 
you are right.
Actually there is similar comment before https://github.com/apache/cloudstack/pull/3306#issuecomment-502142160

anuragaw: Sounds good. I think we should make this simple change then to make code cleaner and easy to comprehend in this PR.

rhtyd:
I'll wait for your final change @ustcweizhou, then it it ready for merge. The latest tests LGTM.

ustcweizhou:
@rhtyd @anuragaw 
rebased with latest master, and added a separate commit to remove the methods not in use.
If we need to fill more information to project response in the future, we can revert the commit.

rhtyd:
Thanks @ustcweizhou
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

PR #3308, opened 2019-05-01
Console Proxy: Ignore META key mask if control was pressed

### LINKED ISSUE #3229: Console Proxy sends redundant CTRL and ALT keys to VMware
##### ISSUE TYPE

 * Bug Report
 
##### COMPONENT NAME

~~~
Console Proxy (or possibly VMware SDK...)
~~~

##### CLOUDSTACK VERSION

~~~
4.11.X
~~~

##### CONFIGURATION

VMware zone

##### OS / ENVIRONMENT

VMware zone

##### SUMMARY

on VMware Zone, hitting CTRL over Console Proxy will send CTRL+ALT + CTRL keys or possible CTRL + ALT + right CTRL + right ALT (as can be seen on Windows with on-screen keyboard)

##### STEPS TO REPRODUCE

use "showkey" tool on Linux - and hit just CTRL key - you will see keycode 29,56,29 pressed and then keycode 29,56 released ?!) 
On Windows guest VM, start on-screen keyboard, and again press CTRL  - it will show both CTRL and both ALT keys being pressed - ctrl + c doesn't work and will produce copyright sign instead of copying text, etc.

##### EXPECTED RESULTS

~~~
When pressing CTRL, send only "CTRL" to VM console.
~~~

##### ACTUAL RESULTS

~~~
Read above from "steps to reproduce" section please
~~~

anuragaw:
On VMware Zone, hitting CTRL over Console Proxy will send a mask of Meta key as well. This makes Ctrl+A, Ctrl+E and many functionalities to not work in console.

Read https://github.com/apache/cloudstack/issues/3229 for details

For fixing ignore Meta key flag passed by SDK if Control was pressed. The Jquery implementation sets the meta key to control key to support IE browser.

## Description

Fixes #3229 
## Types of changes

## How Has This Been Tested?

Tested the change on KVM, VMWare on Ubuntu based Chrome and MacBook based Chrome. Fix solves the problem.

### DIFF
--- a/systemvm/agent/js/ajaxviewer.js
+++ b/systemvm/agent/js/ajaxviewer.js
@@ -1431,7 +1431,9 @@ AjaxViewer.prototype = {
 		if(e.shiftLeft)
 			modifiers |= AjaxViewer.LEFT_SHIFT_MASK;
 
-		if(e.metaKey)
+		// Don't pass meta key modifier filter if control key is pressed.
+		// For more details see https://github.com/apache/cloudstack/issues/3229
+		if(e.metaKey && !e.ctrlKey)
 			modifiers |= AjaxViewer.META_KEY_MASK;
 
 		return modifiers;


anuragaw:
@blueorangutan package

rhtyd:
I found this old jQuery bug that expects metakey to be sent when ctrl key is hit: https://bugs.jquery.com/ticket/11328

This is a Javascript only change, only manual testing is needed wrt VMware on all three Linux, Mac and Windows wrt Firefox and Chrome browsers. 

I tested it on Chrome and Firefox on Linux and fix passes testing. We want some testing to be done on Windows and Mac to qualify merging the bugfix.

anuragaw:
More details - Meta key for Mac (Cmd) has no effect on the Console proxy. This is true for with and without the fix in the PR. The bug is rooted in the jquery implementation to support both Windows and Cmd Meta key and it's desire to support IE Browser. Whether we want to fix Cmd key is not considered as part of this PR.

rhtyd:
@blueorangutan package

rhtyd:
Pure JS change, backend testing not necessary.

DaanHoogland:
One question @anuragaw : what happens if both keys are pressed? are we now ignoring the meta key?

rhtyd:
I'm not sure @DaanHoogland 
@anuragaw can you advise?

anuragaw:
> More details - Meta key for Mac (Cmd) has no effect on the Console proxy. This is true for with and without the fix in the PR. The bug is rooted in the jquery implementation to support both Windows and Cmd Meta key and it's desire to support IE Browser. Whether we want to fix Cmd key is not considered as part of this PR.

@DaanHoogland - If both are pressed we ignore META key. Also note META key functionality for MAC was already broken with the existing code. Although I am not sure which cases we need to use both. Perhaps we can test @borisstoyanov to see if major keyboard shortcuts work as expected on all platforms (UI browsers). I tested on mentioned scenarios and found all shortcuts working as expected.

PR #3312, opened 2019-05-02
'sed -i' need postfix on mac

DaanHoogland:
## Description

build fails on mac due to #3247 . adding extension that will only be created in '/target/' anyway

## Types of changes

### DIFF
--- a/tools/apidoc/build-apidoc.sh
+++ b/tools/apidoc/build-apidoc.sh
@@ -18,7 +18,7 @@
  
 
 # cloud-build-api-doc.sh -- builds api documentation.
-#set -x
+set -x
 #set -u
 ACS_RELEASE="$1"
 shift
@@ -61,8 +61,8 @@ set -e
  cp "$thisdir"/*.java .
  cp "$thisdir"/*.xsl .
  sed -e 's,%API_HEADER%,All APIs,g' "$thisdir/generatetoc_header.xsl" >generatetoc.xsl
- sed -i "s/%ACS_RELEASE%/${ACS_RELEASE}/g" generatetoc.xsl
- sed -i "s/%ACS_RELEASE%/${ACS_RELEASE}/g" generatecommands.xsl
+ sed -i .bak "s/%ACS_RELEASE%/${ACS_RELEASE}/g" generatetoc.xsl
+ sed -i .bak "s/%ACS_RELEASE%/${ACS_RELEASE}/g" generatecommands.xsl
 
  PLATFORM=`uname -s`
  if [[ "$PLATFORM" =~ .*WIN.* ]]
--- a/tools/apidoc/build-apidoc.sh
+++ b/tools/apidoc/build-apidoc.sh
@@ -18,8 +18,7 @@
  
 
 # cloud-build-api-doc.sh -- builds api documentation.
-set -x
-#set -u
+
 ACS_RELEASE="$1"
 shift
 TARGETJARDIR="$1"


DaanHoogland:
ping @GabrielBrascher @nathanejohnson , this slipped in end of march. didn't notice, too much on fedora :(, not enough on a mac :)

tools/apidoc/build-apidoc.sh
```diff
@@ -61,8 +60,8 @@ set -e
  cp "$thisdir"/*.java .
  cp "$thisdir"/*.xsl .
  sed -e 's,%API_HEADER%,All APIs,g' "$thisdir/generatetoc_header.xsl" >generatetoc.xsl
- sed -i "s/%ACS_RELEASE%/${ACS_RELEASE}/g" generatetoc.xsl
- sed -i "s/%ACS_RELEASE%/${ACS_RELEASE}/g" generatecommands.xsl
+ sed -i .bak "s/%ACS_RELEASE%/${ACS_RELEASE}/g" generatetoc.xsl
```
rafaelweingartner: What if we execute the commands at line 63 and 64 like the one in 62? Or, maybe the other way around. 

I mean, they do the same thing, but with different approaches. Therefore, it might be interesting to maintain some consistency here.

DaanHoogland: @rafaelweingartner, not entirely sure what you mean but 62 an 63 can certainly be combined. not sure how 64 can, it is touching a different file.

rafaelweingartner: I meant the following:
All of the commands are replacing "place holders" in the same file. However, two of them are doing in line (via -i flag), whereas line 62 is replacing the entire file (not using the -i flag and redirecting the output of the `sed` to a file with the same name). I do not mind if we do in different commands the replace, but we could at least try to use the same approach in all of them. 

By approach here I mean the method, either inline or not.

DaanHoogland: I think you are misreading the code,
 line 62 redirects to a new file, from "$thisdir/generatetoc_header.xsl" to generatetoc.xsl
 then 63 operates on the resulting generatetoc.xsl
 but 64 is operating on generatecommands.xsl

I don't think shell operations would allow redecting to input, it would empty the file before openening it for read if you tried that.

rhtyd:
@DaanHoogland env issue, the sed on OSX is freebsd based you can try gnu-sed like:
```
$ brew install gnu-sed
```

DaanHoogland:
> @DaanHoogland env issue, the sed on OSX is freebsd based

I realize that, but it complicates build and thus inhibits people entering the playing field. I would prefer adding the extension, it is in the target dir and doesn't hurt anybody.

nathanejohnson:
@rhtyd you can easily add gnu sed to mac via homebrew as you point out, however the binary is called 'gsed' to avoid conflicting with the main sed installed in /usr/bin .

tools/apidoc/build-apidoc.sh
GabrielBrascher: @DaanHoogland @rafaelweingartner A (ugly) solution is running the command with the -e and move the new file to the old one:
`sed -e "s/%ACS_RELEASE%/${ACS_RELEASE}/g" newGeneratetoc.xsl && mv newGeneratetoc.xsl generatetoc.xsl;`

GabrielBrascher: @DaanHoogland on linux the `-i` must be appended with the `.bak` (`-i.bak`)

```
-i[SUFFIX], --in-place[=SUFFIX]
              edit files in place (makes backup if SUFFIX supplied)
```

rhtyd:
@blueorangutan package

GabrielBrascher:
@DaanHoogland I don't think that this would work on linux. The `-i` must be appended with the .bak (-i.bak)

Documentation:
```
-i[SUFFIX], --in-place[=SUFFIX]
              edit files in place (makes backup if SUFFIX supplied)
```

rhtyd:
closed-open PR to rekick Travis
@blueorangutan package

DaanHoogland:
> @DaanHoogland I don't think that this would work on linux. The `-i` must be appended with the .bak (-i.bak)
@GaborApatiNagy, I'll try and find time to test this on fedora and if so I'll checnge it to the long option. 
I hope there is some compatible option/argument set between the two.

DaanHoogland:
travis is quite clear in this case: needs tlc

rhtyd:
@DaanHoogland can you debug via the backend jenkins what failed?

DaanHoogland:
@rhtyd if you mean travis, yes. but kind of high prio -3.

rhtyd:
@DaanHoogland fails on linux with:
```
Warning, API Cmd class com.cloud.api.commands.SimulatorAddSecondaryAgent has no APICommand annotation 
Scanned and found 607 APIs
sed: -e expression #1, char 1: unknown command: `.'
[ERROR] Failed to execute goal org.codehaus.mojo:exec-maven-plugin:1.2.1:exec (compile) on project cloud-apidoc: Command execution failed. Process exited with an error: 1 (Exit value: 1) -> [Help 1]
```

rhtyd:
This is `onhold` until Travis passes.

rhtyd:
Closing as https://github.com/apache/cloudstack/pull/4253 fixed it

Issue #3089, opened 2018-12-10
VMs with invalid IDE configuration when using Solidfire as primary storage


DennisKonrad:
##### _ISSUE TYPE_
 * Bug Report

##### _COMPONENT NAME_
Domain XML with multiple disks attached using Solidfire primary storage

##### _CLOUDSTACK VERSION_
Master branch

##### _OS / ENVIRONMENT_
Solidfire primary storage

##### _SUMMARY_
Example:
When we select "Windows Server 2016 (64-bit)" as OS Type for an W2K16 cd image the resulting
VMs will not start with multiple disks attached. This is due to the usage of _**multiple IDE interfaces in the VM description XML**_.

It seems this is not just a problem with this OS Type but with the way the domain XMLs are generated when using solidfire storage.

This _**only happens when the disks are on Solidfire primary storage**_. We confirmed that using NFS primary storage this is not the case.

##### STEPS TO REPRODUCE
Try to use boot a VM with at least two disks on solidfire primary storage. Depending on the OS-Type
that is selected for the image that was used to create the VM, the domain xml generated may be invalid.

In our case we tested this with a W2K16 vm which will not start when using OS Type "Windows Server 2016 (64-bit)". **_Libvirt does not boot the machine_** and raise the following error message:

`org.libvirt.LibvirtException: unsupported configuration: Only a single IDE controller is supported for this machine type`

### Logs & Attachments
Here you can see two excerpts from the generated domain xmls. First the working one (NFS) and below that the one not working (Solidfire).

##### NFS Primary Storage

    <disk type='file' device='disk'>
      <driver name='qemu' type='qcow2' cache='none'/>
      <source file='/mnt/e0e5ba2b-15bc-3fd8-a010-40a404c3cdfc/2f4ddade-b93e-47c5-b97f-b738f395d9e1'/>
      <backingStore/>
      <target dev='hda' bus='ide'/>
      <alias name='ide0-0-0'/>
      <address type='drive' controller='0' bus='0' target='0' unit='0'/>
    </disk>
    <disk type='file' device='cdrom'>
      <driver name='qemu' type='raw'/>
      <source file='/mnt/43c36bb7-1d8d-3405-b9b8-ddd402ef7913/215-34-015cecb6-7c65-3a37-80da-fde928458db4.iso'/>
      <backingStore/>
      <target dev='hdc' bus='ide'/>
      <readonly/>
      <alias name='ide0-1-0'/>
[...]

PR #3319, opened 2019-05-08
Use IDE as the bus type for root disks and VIRTIO for data disks on platforms without support for para virtualization when using managed storage

skattoju4:
## Description
This change addresses #3089. There was an issue when disks were being added with bus type IDE when creating windows VMs from ISOs. It is not possible to select bus type when creating a VM with an ISO. The bus type is inferred based on the platform emulator string provided to the KVM agent. Currently when creating a VM with managed storage (ex: Solidfire) and OS type string Windows*, all disks are added as IDE. Qemu currently does not support multiple IDE controllers and this configuration results in VMs that cannot be started. This issue does not occur when using NFS as the storage provider due to logic in that KVM agent that makes all data volumes (non root) use a virtio controller for file based disk. Similar logic was added for raw physical disks so that managed storage has the same behavior as NFS. In addition specific versions were removed from the code that guesses the disk controller to be used based on the platform emulator string since most modern operating systems support virtio.

steps to reproduce are described in #3089

## Types of changes

## How Has This Been Tested?

In a Cloudstack system with Solidfire managed primary storage and a KVM host the following steps were performed:
1. Register to an ISO template with OS set to Windows 2016
2. Create a VM using this template with Solidfire Storage
3. Verify that the VM starts successfully
4. Verify that disks are attached on the virtio bus and only the cdrom device is ide by executing `virsh dumpxml <id>` on the KVM host

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2330,9 +2330,13 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
 
             // if params contains a rootDiskController key, use its value (this is what other HVs are doing)
             DiskDef.DiskBus diskBusType = getDiskModelFromVMDetail(vmSpec);
-
             if (diskBusType == null) {
-                diskBusType = getGuestDiskModel(vmSpec.getPlatformEmulator());
+                // always use bus type ide for windows root volumes unless specified in vmspec
+                if (volume.getType() == Volume.Type.ROOT && vmSpec.getPlatformEmulator().startsWith("Windows")){
+                    diskBusType = DiskDef.DiskBus.IDE;
+                } else {
+                    diskBusType = getGuestDiskModel(vmSpec.getPlatformEmulator());
+                }
             }
 
             // I'm not sure why previously certain DATADISKs were hard-coded VIRTIO and others not, however this
@@ -3205,7 +3209,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
                 platformEmulator.startsWith("Debian GNU/Linux") ||
                 platformEmulator.startsWith("FreeBSD") ||
                 platformEmulator.startsWith("Oracle") ||
-                platformEmulator.startsWith("Windows") ||
+                platformEmulator.startsWith("Windows PV") ||
                 platformEmulator.startsWith("Other PV")) {
             return DiskDef.DiskBus.VIRTIO;
         } else {
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2331,8 +2331,8 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             // if params contains a rootDiskController key, use its value (this is what other HVs are doing)
             DiskDef.DiskBus diskBusType = getDiskModelFromVMDetail(vmSpec);
             if (diskBusType == null) {
-                // always use bus type ide for windows root volumes unless specified in vmspec
-                if (volume.getType() == Volume.Type.ROOT && vmSpec.getPlatformEmulator().startsWith("Windows")){
+                // always use bus type ide for windows root volumes unless specified in vm details or windows pv is selected
+                if (volume.getType() == Volume.Type.ROOT && vmSpec.getPlatformEmulator().startsWith("Windows") && !vmSpec.getPlatformEmulator().startsWith("Windows PV")){
                     diskBusType = DiskDef.DiskBus.IDE;
                 } else {
                     diskBusType = getGuestDiskModel(vmSpec.getPlatformEmulator());
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -3202,14 +3202,13 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             return DiskDef.DiskBus.IDE;
         } else if (platformEmulator.startsWith("Other PV Virtio-SCSI")) {
             return DiskDef.DiskBus.SCSI;
-        } else if (platformEmulator.startsWith("Ubuntu") ||
+        } else if (platformEmulator.contains("Ubuntu") ||
                 platformEmulator.startsWith("Fedora") ||
                 platformEmulator.startsWith("CentOS") ||
                 platformEmulator.startsWith("Red Hat Enterprise Linux") ||
                 platformEmulator.startsWith("Debian GNU/Linux") ||
                 platformEmulator.startsWith("FreeBSD") ||
                 platformEmulator.startsWith("Oracle") ||
-                platformEmulator.startsWith("Windows PV") ||
                 platformEmulator.startsWith("Other PV")) {
             return DiskDef.DiskBus.VIRTIO;
         } else {
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2331,14 +2331,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             // if params contains a rootDiskController key, use its value (this is what other HVs are doing)
             DiskDef.DiskBus diskBusType = getDiskModelFromVMDetail(vmSpec);
             if (diskBusType == null) {
-                // always use bus type ide for windows root volumes and virtio for data volumes unless specified in vm details or windows pv is selected
-                if (volume.getType() == Volume.Type.ROOT && vmSpec.getPlatformEmulator().startsWith("Windows") && !vmSpec.getPlatformEmulator().startsWith("Windows PV")){
-                    diskBusType = DiskDef.DiskBus.IDE;
-                } else if (volume.getType() == Volume.Type.DATADISK && vmSpec.getPlatformEmulator().startsWith("Windows")) {
-                    diskBusType = DiskDef.DiskBus.VIRTIO;
-                } else {
-                    diskBusType = getGuestDiskModel(vmSpec.getPlatformEmulator());
-                }
+                diskBusType = getGuestDiskModel(vmSpec.getPlatformEmulator());
             }
 
             // I'm not sure why previously certain DATADISKs were hard-coded VIRTIO and others not, however this
@@ -2375,7 +2368,12 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
                     disk.defNetworkBasedDisk(glusterVolume + path.replace(mountpoint, ""), pool.getSourceHost(), pool.getSourcePort(), null,
                             null, devId, diskBusType, DiskProtocol.GLUSTER, DiskDef.DiskFmtType.QCOW2);
                 } else if (pool.getType() == StoragePoolType.CLVM || physicalDisk.getFormat() == PhysicalDiskFormat.RAW) {
-                    disk.defBlockBasedDisk(physicalDisk.getPath(), devId, diskBusType);
+                    if (volume.getType() == Volume.Type.DATADISK) {
+                        disk.defBlockBasedDisk(physicalDisk.getPath(), devId, diskBusTypeData);
+                    }
+                    else {
+                        disk.defBlockBasedDisk(physicalDisk.getPath(), devId, diskBusType);
+                    }
                 } else {
                     if (volume.getType() == Volume.Type.DATADISK) {
                         disk.defFileBasedDisk(physicalDisk.getPath(), devId, diskBusTypeData, DiskDef.DiskFmtType.QCOW2);
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2331,9 +2331,11 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             // if params contains a rootDiskController key, use its value (this is what other HVs are doing)
             DiskDef.DiskBus diskBusType = getDiskModelFromVMDetail(vmSpec);
             if (diskBusType == null) {
-                // always use bus type ide for windows root volumes unless specified in vm details or windows pv is selected
+                // always use bus type ide for windows root volumes and virtio for data volumes unless specified in vm details or windows pv is selected
                 if (volume.getType() == Volume.Type.ROOT && vmSpec.getPlatformEmulator().startsWith("Windows") && !vmSpec.getPlatformEmulator().startsWith("Windows PV")){
                     diskBusType = DiskDef.DiskBus.IDE;
+                } else if (volume.getType() == Volume.Type.DATADISK && vmSpec.getPlatformEmulator().startsWith("Windows")) {
+                    diskBusType = DiskDef.DiskBus.VIRTIO;
                 } else {
                     diskBusType = getGuestDiskModel(vmSpec.getPlatformEmulator());
                 }
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -3198,12 +3198,15 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             return DiskDef.DiskBus.IDE;
         } else if (platformEmulator.startsWith("Other PV Virtio-SCSI")) {
             return DiskDef.DiskBus.SCSI;
-        } else if (platformEmulator.startsWith("Ubuntu") || platformEmulator.startsWith("Fedora 13") || platformEmulator.startsWith("Fedora 12") || platformEmulator.startsWith("Fedora 11") ||
-                platformEmulator.startsWith("Fedora 10") || platformEmulator.startsWith("Fedora 9") || platformEmulator.startsWith("CentOS 5.3") || platformEmulator.startsWith("CentOS 5.4") ||
-                platformEmulator.startsWith("CentOS 5.5") || platformEmulator.startsWith("CentOS") || platformEmulator.startsWith("Fedora") ||
-                platformEmulator.startsWith("Red Hat Enterprise Linux 5.3") || platformEmulator.startsWith("Red Hat Enterprise Linux 5.4") ||
-                platformEmulator.startsWith("Red Hat Enterprise Linux 5.5") || platformEmulator.startsWith("Red Hat Enterprise Linux 6") || platformEmulator.startsWith("Debian GNU/Linux") ||
-                platformEmulator.startsWith("FreeBSD 10") || platformEmulator.startsWith("Oracle") || platformEmulator.startsWith("Other PV")) {
+        } else if (platformEmulator.startsWith("Ubuntu") ||
+                platformEmulator.startsWith("Fedora") ||
+                platformEmulator.startsWith("CentOS") ||
+                platformEmulator.startsWith("Red Hat Enterprise Linux") ||
+                platformEmulator.startsWith("Debian GNU/Linux") ||
+                platformEmulator.startsWith("FreeBSD") ||
+                platformEmulator.startsWith("Oracle") ||
+                platformEmulator.startsWith("Windows") ||
+                platformEmulator.startsWith("Other PV")) {
             return DiskDef.DiskBus.VIRTIO;
         } else {
             return DiskDef.DiskBus.IDE;


svenvogel:
@skattoju3 thanks we will test it and come back tomorrow.

ustcweizhou:
@skattoju4 
for windows vm, we need to install virtio drivers.
for now, we can register template without virtio drivers as Windows XX, and template with virtio drivers as "Windows PV".
This PR will break actions above.

so -1 for me.

svenvogel:
@ustcweizhou

Maybe we talk past each other. It should only set the first disk to ide device and all other to virtio like in the bug Report #3089 described. 

We don’t want to break things and we don’t want only virtio disks. 

Simple it should be 
Cdrom ide
First disk ide
Second virtio
Third virtio 
And so one ...

The bug with solidfire / iscsi we got is 

Cdrom ide
First ide
Second ide
...that’s but really supported in kvm. With nfs there is no problem. 

@skattoju4 is that what I understand it Correctly?

@syed what do you think?

ustcweizhou:
@svenvogel 
the main problem for this PR is, all Windows vm will be running with virtio (no ide any more !!)
If vm does not have virtio drivers, they will not boot up

svenvogel:
@ustcweizhou

You are right. That should not be the solution. I understand. Let’s investigate and fix that. What I wrote above the goal was another.

syed:
I agree with @svenvogel. The first disk should always be IDE. The subsequent disks can be virtio.

ustcweizhou:
@skattoju4 @svenvogel @syed 
I have no idea why the first disk should be always IDE.
If I have installed virtio driver in windows vm, and OS type (vm or template) is set to 'Windows PV', I hope all disk are virtio, not IDE.

skattoju4:
@ustcweizhou the thinking behind always setting the root disk to ide was to be able to boot a windows vm regardless of whether or not virtio drivers were installed. That said setting all disks to virtio when selecting "Windows PV" when registering the template is one way to move forward. @svenvogel what do you think ?

svenvogel:
@skattoju4 @ustcweizhou @syed i will try to explain it.

> @ustcweizhou the thinking behind always setting the root disk to ide was to be able to boot a windows vm regardless of whether or not virtio drivers were installed. That said setting all disks to virtio when selecting "Windows PV" when registering the template is one way to move forward. 

first i need to say is yes! lets me explain. how written in #3089 we have "NOT" only the problem with template! 
@DennisKonrad wrote:

When we select "Windows Server 2016 (64-bit)" as OS Type for an **W2K16 cd image** the resulting
VMs will not start with multiple disks attached. This is due to the usage of multiple IDE interfaces in the VM description XML.

1. we create a machine with "Windows Server 2016 (64-bit)" we have **one** disk with type root "ide". 
Bildschirmfoto 2019-05-10 um 21 50 40
2. after that we add **two more data disks** and they should be "NOT" ide because QEMU dont like this! When we use NFS the second and third disk will be a "virtio" disk automatically.

this pr should "NOT" remove/break or add functionality. 

it should be correct the normal functionality for managed storage like NFS works and qemu it wants if you choose "Windows Server 201X..." as OS Type. its a bug they only will affect (at least what i know) managed storage. If any user wants "ONLY" virtio driver he can choose "other pv ..." OS Type.
Bildschirmfoto 2019-05-10 um 21 58 11

there is no discussion about remove perfomance of ide or move all Microsoft Windows OS Types to virtio. it should clear for everybody here that virtio is faster than ide and yes we dont need for the root disk a driver. this is just as little the content of the discussion.

i hope it is more clear now?

ustcweizhou:
@svenvogel thanks for you explanation.
Currently we use 'Windows PV' for windows vm. All disk are virtio.
We hope this will not be broken. 
Change OS type from 'Windows PV' to 'Other PV' will lead to changes not only in cloudstack itself, but also in some maintenance scripts.

so please change the code to following if possible.
```
                if (volume.getType() == Volume.Type.ROOT &&
                    vmSpec.getPlatformEmulator().startsWith("Windows") &&
                    ! vmSpec.getPlatformEmulator().equals("Windows PV")){
```

svenvogel:
@ustcweizhou no we will check this before and have a look on the PR that we dont break the virtio.

@syed @skattoju4 what do think, can we add this?

syed:
Sounds good to me. @skattoju4 can you take care of it?

kiwiflyer:
Also note that as of 4.12,  kvm enlightenment is also enabled via Windows
PV. I'm working on a PR to also allow it to be enabled by the vmdetails
feature.

DennisKonrad:
@kiwiflyer Such a feature would a way in the direction of further untangling this os-type problem. I am looking forward to your pr.

svenvogel:
> @svenvogel thanks for you explanation.
> Currently we use 'Windows PV' for windows vm. All disk are virtio.
> We hope this will not be broken.
> Change OS type from 'Windows PV' to 'Other PV' will lead to changes not only in cloudstack itself, but also in some maintenance scripts.
> 
> so please change the code to following if possible.
> 
> ```
>                 if (volume.getType() == Volume.Type.ROOT &&
>                     vmSpec.getPlatformEmulator().startsWith("Windows") &&
>                     ! vmSpec.getPlatformEmulator().equals("Windows PV")){
> ```

@skattoju4 @syed do you think the fix from @ustcweizhou will fix this? 

@kiwiflyer better configuration options for the os type sounds good.

skattoju4:
@sven , @ustcweizhou 's suggestion sounds reasonable. I think this is ok for now but ideally this should be handle in a better way. @kiwiflyer could you link your PR please ?

svenvogel:
@skattoju4 do you mean we should test it? i think @kiwiflyer 's PR is out of scope of this one. right?

skattoju4:
@svenvogel now we should have the same behavior as nfs when selecting os type Windows* (except Windows PV) I think it might be better to select Windows PV and have all disk be virtio when para-virtualization is supported. With the current code vm created from iso templates with os type Windows* (Except Windows PV) will have an ide root disk but there is no way to attach a data disk as IDE. (one additional ide disk is possible before the error is triggered if i am not mistaken)

I'm guessing @kiwiflyer 's PR will enable setting the disk controller when registering an iso similar to how this can be specified when registering a template.  I think this would be a cleaner solution than trying to guess the controller / bus type based on the os type string.

svenvogel:
@skattoju4 we will test it 
> (one additional ide disk is possible before the error is triggered if i am not mistaken)
unfortunately no. only one ide disk so what i know.

okay there should not be a second ide disk right?

skattoju4:
yes, just the root disk will be ide now.

DennisKonrad:
@skattoju4 @svenvogel I will build this today and testing can happen the following days.

kiwiflyer:
To be clear, the feature that allows selection of root disk controller for a template using vmdetails has been in cloudstack for a couple of major versions now - 

image

syed:
Yes. But it doesn't work for ISOs

kiwiflyer:
Since the vmdetails is exposed per vm, we should just allow the selection at VM creation.

skattoju4:
I think since we always create vms from templates this should be specified when registering templates to be consistent with the current workflow.

skattoju4:
@kiwiflyer after discussing this with @swill i have a better understanding. Essentially registering ISO is supposed to be hypervisor agnostic. We can only know valid selections for disk controller in the context of a selected hypervisor which in the case of ISOs is during VM creation. Given this it makes sense to have the option of specifying hypervisor specific options during VM creation. The deployVirtualMachine api has a details map that can be used to pass this info.

svenvogel:
@skattoju4 should we test it now?

skattoju4:
> @skattoju4 should we test it now?

Hi @svenvogel yes please go ahead

rhtyd:
@kiwiflyer @svenvogel @syed @DennisKonrad and others - what's the consensus on this PR? I don't see any lgtm.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

rhtyd:
What's the consensus on the PR, is it ready for review/testing or there are outstanding concerns - @ustcweizhou @syed @kiwiflyer @svenvogel @DennisKonrad @skattoju4 ?

rhtyd:
I've added `onhold` label unless we hear from the previous reviewers and the author.

rhtyd:
cc @swill

DennisKonrad:
I tested this PR manually some time ago and it worked just fine. Codewise it LGTM

rhtyd:
Thanks for sharing @DennisKonrad 
Smoketests LGTM, we can merge as soon as we have another lgtm - ping @ustcweizhou @wido @kiwiflyer @svenvogel @GabrielBrascher?

rhtyd:
@ustcweizhou you initial shared -1 on this, are you still -1 on this PR? Can you review? Thanks.

kiwiflyer:
Just catching up on this issue. Just so I understand the use case here...in what situations would you want ide for the boot disk only and not virtio for all?

ustcweizhou:
@rhtyd I am still -1 on this PR.

@skattoju4 @svenvogel 
Does the issue happen other OSes than Windows ? 
I have tested os types windows/windows pv/other linux/ubuntu , all seem ok (the data disk is virtio), vms are running on NFS. So I think it is an issue with the disk definition in kvm plugin.

look at the following lines in LibvirtComputingResource.java
for NFS, the DATADISK has different disk bus type (diskBusTypeData = VIRTIO/SCSI) than ROOT disk (diskBusType = IDE/ VIRTIO/SCSI)
```
                } else if (pool.getType() == StoragePoolType.CLVM || physicalDisk.getFormat() == PhysicalDiskFormat.RAW) {
                    disk.defBlockBasedDisk(physicalDisk.getPath(), devId, diskBusType);
                } else {
                    if (volume.getType() == Volume.Type.DATADISK) {
                        disk.defFileBasedDisk(physicalDisk.getPath(), devId, diskBusTypeData, DiskDef.DiskFmtType.QCOW2);
                    } else {
                        disk.defFileBasedDisk(physicalDisk.getPath(), devId, diskBusType, DiskDef.DiskFmtType.QCOW2);
                    }
                }
```

ustcweizhou:
I tested on NFS and Ceph, both are working without this PR.

Ubuntu 16.04
Cloudstack 4.11.3

* NFS:
disks on vm with PV: all are virtio
disks on vm without PV: root disk is ide, others are virtio

* Ceph
disks on vm with PV: all are virtio
disks on vm without PV: all are ide.

svenvogel:
wow many answers. okay lets bring some light into the issue.

first of all i think all here understand the problem. 
https://github.com/apache/cloudstack/pull/3319#issuecomment-491415223

if this so then lets talk about test and if we found the right place for the code.

We use Netapp Solidfire managed storage and KVM.

@kiwiflyer thats for our customer and devs they dont want to install virtio driver in the image before or with boot time ;). so ide is "slower" than virto but its easier for them. we only want the normal configuration like the other systems. if we want complete virtio we can choose it.

@ustcweizhou yes. so what we found out is that it was only with windows os and yes we use KVM. 
KVM (libvirt/qemu) wants only one ide disk root disk and the other one should be virtio. 

ping We should ask @syed and @skattoju4 for that. maybe you are right. what do you think to what @ustcweizhou says?

---
regardless of this fix ... the code below should included in any case because it looks ugly at the moment. (code below)

```java
        } else if (platformEmulator.startsWith("Ubuntu") ||
                platformEmulator.startsWith("Fedora") ||
                platformEmulator.startsWith("CentOS") ||
                platformEmulator.startsWith("Red Hat Enterprise Linux") ||
                platformEmulator.startsWith("Debian GNU/Linux") ||
                platformEmulator.startsWith("FreeBSD") ||
                platformEmulator.startsWith("Oracle") ||
                platformEmulator.startsWith("Windows PV") ||
                platformEmulator.startsWith("Other PV")) {

**Comment by ustcweizhou** [comment]
@svenvogel thanks for your reply.
@skattoju4 @syed could you please consider the following change ?
```
diff --git a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
index 6f62b4a..411ba70 100644
--- a/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2388,7 +2388,11 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
                     disk.defNetworkBasedDisk(glusterVolume + path.replace(mountpoint, ""), pool.getSourceHost(), pool.getSourcePort(), null,
                             null, devId, diskBusType, DiskProtocol.GLUSTER, DiskDef.DiskFmtType.QCOW2);
                 } else if (pool.getType() == StoragePoolType.CLVM || physicalDisk.getFormat() == PhysicalDiskFormat.RAW) {
-                    disk.defBlockBasedDisk(physicalDisk.getPath(), devId, diskBusType);
+                    if (volume.getType() == Volume.Type.DATADISK) {
+                        disk.defBlockBasedDisk(physicalDisk.getPath(), devId, diskBusTypeData);
+                    } else {
+                        disk.defBlockBasedDisk(physicalDisk.getPath(), devId, diskBusType);
+                    }
                 } else {
                     if (volume.getType() == Volume.Type.DATADISK) {
                         disk.defFileBasedDisk(physicalDisk.getPath(), devId, diskBusTypeData, DiskDef.DiskFmtType.QCOW2);
```

**Comment by ustcweizhou** [comment]
Agree with you.
Furthermore, I think it is better to use 
```
platformEmulator.contains("Ubuntu")
```
because some users use "Other Ubuntu" for the Ubuntu versions which are not in the list, for example Ubuntu 17.04, Ubuntu 18.04.


> Regardless of this fix ... the code below should included in any case because it looks ugly at the moment. (code below)
> 
> ```java
>         } else if (platformEmulator.startsWith("Ubuntu") ||
>                 platformEmulator.startsWith("Fedora") ||
>                 platformEmulator.startsWith("CentOS") ||
>                 platformEmulator.startsWith("Red Hat Enterprise Linux") ||
>                 platformEmulator.startsWith("Debian GNU/Linux") ||
>                 platformEmulator.startsWith("FreeBSD") ||
>                 platformEmulator.startsWith("Oracle") ||
>                 platformEmulator.startsWith("Windows PV") ||
>                 platformEmulator.startsWith("Other PV")) {
> ```

**Comment by skattoju4** [comment]
@kiwiflyer we need ide for the root disk in case the guest OS is missing para virtualization support
@ustcweizhou this only happens with windows and managed storage (not nfs) the datadisk logic you pointed out is not reached

**Comment by svenvogel** [comment]
@ustcweizhou 

> @svenvogel Agree with you.
> Furthermore, I think it is better to use
> 
> ```
> platformEmulator.contains("Ubuntu")
> ```
yes sounds good!

@skattoju4 sounds logical. so we debugged that right?

**Comment by ustcweizhou** [comment]
> @kiwiflyer we need ide for the root disk in case the guest OS is missing para virtualization support
> @ustcweizhou this only happens with windows and managed storage (not nfs) the datadisk logic you pointed out is not reached

@skattoju4 Can you test with a Linux server (without PV, for example OS type is "Other Linux") on Solidfire Storage ?

**Comment by skattoju4** [comment]
@svenvogel yeah @ustcweizhou suggestion is cleaner / more generic the windows specific logic can be removed. As pointed out the issue is guest OSs with pv support using managed storage. (sorry missed previous messages)

**Comment by svenvogel** [comment]
@skattoju4 should we change it like the suggestion of @ustcweizhou? i mean the ide managed storage...

**Comment by skattoju4** [comment]
@svenvogel it sounds logical will test and update the pr

**Comment by svenvogel** [comment]
@skattoju4 sound good. thx!

**Comment by ustcweizhou** [comment]
@skattoju4 good, looking forward to your update.
@svenvogel @skattoju4 thanks for nice discussion tonight. have a good evening.

**Comment by skattoju4** [comment]
> @svenvogel Agree with you.
> Furthermore, I think it is better to use
> 
> ```
> platformEmulator.contains("Ubuntu")
> ```
> 
> because some users use "Other Ubuntu" for the Ubuntu versions which are not in the list, for example Ubuntu 17.04, Ubuntu 18.04.
> 
> > Regardless of this fix ... the code below should included in any case because it looks ugly at the moment. (code below)
> > ```java
> >         } else if (platformEmulator.startsWith("Ubuntu") ||
> >                 platformEmulator.startsWith("Fedora") ||
> >                 platformEmulator.startsWith("CentOS") ||
> >                 platformEmulator.startsWith("Red Hat Enterprise Linux") ||
> >                 platformEmulator.startsWith("Debian GNU/Linux") ||
> >                 platformEmulator.startsWith("FreeBSD") ||
> >                 platformEmulator.startsWith("Oracle") ||
> >                 platformEmulator.startsWith("Windows PV") ||
> >                 platformEmulator.startsWith("Other PV")) {
> > ```

Are there any Ubuntu versions which don't support PV ?

**Comment by ustcweizhou** [comment]
> > @svenvogel Agree with you.
> > Furthermore, I think it is better to use
> > ```
> > platformEmulator.contains("Ubuntu")
> > ```
> > 
> > 
> > because some users use "Other Ubuntu" for the Ubuntu versions which are not in the list, for example Ubuntu 17.04, Ubuntu 18.04.
> > > Regardless of this fix ... the code below should included in any case because it looks ugly at the moment. (code below)
> > > ```java
> > >         } else if (platformEmulator.startsWith("Ubuntu") ||
> > >                 platformEmulator.startsWith("Fedora") ||
> > >                 platformEmulator.startsWith("CentOS") ||
> > >                 platformEmulator.startsWith("Red Hat Enterprise Linux") ||
> > >                 platformEmulator.startsWith("Debian GNU/Linux") ||
> > >                 platformEmulator.startsWith("FreeBSD") ||
> > >                 platformEmulator.startsWith("Oracle") ||
> > >                 platformEmulator.startsWith("Windows PV") ||
> > >                 platformEmulator.startsWith("Other PV")) {
> > > ```
> 
> Are there any Ubuntu versions which don't support PV ?

as far as  I know, pv is supported by all ubuntu versions in past 10 years.

**Comment by rhtyd** [comment]
Looks like there are still on-going discussions with no clear consensus. Because of this I've removed the milestone on this PR, kindly ping me when a consensus is reached. If this could make into 4.13.0.0 before the freeze I'll be happy to help run tests, otherwise we can revisit this after 4.13.0.0, either in 4.13.1.0 or 4.14.0.0.

File: plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java [hunk 1]
```diff
@@ -3198,12 +3202,15 @@ public boolean isCentosHost() {
...
-                platformEmulator.startsWith("Red Hat Enterprise Linux 5.5") || platformEmulator.startsWith("Red Hat Enterprise Linux 6") || platformEmulator.startsWith("Debian GNU/Linux") ||
-                platformEmulator.startsWith("FreeBSD 10") || platformEmulator.startsWith("Oracle") || platformEmulator.startsWith("Other PV")) {
+        } else if (platformEmulator.startsWith("Ubuntu") ||
+                platformEmulator.startsWith("Fedora") ||
+                platformEmulator.startsWith("CentOS") ||
+                platformEmulator.startsWith("Red Hat Enterprise Linux") ||
+                platformEmulator.startsWith("Debian GNU/Linux") ||
+                platformEmulator.startsWith("FreeBSD") ||
+                platformEmulator.startsWith("Oracle") ||
+                platformEmulator.startsWith("Windows PV") ||
```
ustcweizhou: actually the platformEmulator is "Other PV" if guest os type is "Windows PV".
so line 3212 is not needed

```
mysql> select * from guest_os where display_name='Windows PV';
+-----+-------------+------+--------------------------------------+--------------+---------------------+---------+-----------------+---------+
| id  | category_id | name | uuid                                 | display_name | created             | removed | is_user_defined | display |
+-----+-------------+------+--------------------------------------+--------------+---------------------+---------+-----------------+---------+
| 160 |           6 | NULL | e1990eb3-4b9d-11e3-8346-2c44fd7a3378 | Windows PV   | 2016-07-27 10:48:33 | NULL    |               0 |       0 |
+-----+-------------+------+--------------------------------------+--------------+---------------------+---------+-----------------+---------+
1 row in set (0.00 sec)

mysql> select * from guest_os_hypervisor where guest_os_id=160;
+------+-----------------+---------------+-------------+--------------------+--------------------------------------+---------------------+---------+-----------------+
| id   | hypervisor_type | guest_os_name | guest_os_id | hypervisor_version | uuid                                 | created             | removed | is_user_defined |
+------+-----------------+---------------+-------------+--------------------+--------------------------------------+---------------------+---------+-----------------+
| 1264 | KVM             | Other PV      |         160 | default            | e768e05d-53d6-11e6-8d7a-2c44fd7a3454 | 2016-07-27 10:48:35 | NULL    |               0 |
| 2114 | LXC             | Other PV      |         160 | default            | e965179c-53d6-11e6-8d7a-2c44fd7a3454 | 2016-07-27 10:48:38 | NULL    |               0 |
+------+-----------------+---------------+-------------+--------------------+--------------------------------------+---------------------+---------+-----------------+
2 rows in set (0.00 sec)
```

svenvogel: @ustcweizhou does it select "display_name" or  "guest_os_name"? you mean we dont need "Windows PV" in the list right?

ustcweizhou: @svenvogel  yes, we do not need "Windows PV" in the line. 
the platformEmulator is "Other PV" when guest os is "Windows PV".

ustcweizhou:
> Looks like there are still on-going discussions with no clear consensus. Because of this I've removed the milestone on this PR, kindly ping me when a consensus is reached. If this could make into 4.13.0.0 before the freeze I'll be happy to help run tests, otherwise we can revisit this after 4.13.0.0, either in 4.13.1.0 or 4.14.0.0.

@rhtyd 
the last commit looks good to me.

svenvogel:
@skattoju4 is it ready for merge?

rhtyd:
Let me also help review and test it @svenvogel 
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

PR #3331, opened 2019-05-15
api/server: Add option 'details' to listProjects and listAccounts

ustcweizhou:
## Description

If there are many projects and accounts, listing projects/accounts will take long time getting the resource limitation and resource count in the process. However resource count/limitation are not needed sometimes. 

Add an option 'details' to listProjects and listAccounts. If you do not need the resource count/limitation, please add details=min to api call. The api execution time will be reduced significantly.

## Types of changes

## How Has This Been Tested?

create projects
list projects with details=all
list projects with details=min

create accounts
list accounts with details=all
list accounts with details=min

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/account/ListAccountsCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/account/ListAccountsCmd.java
@@ -16,16 +16,22 @@
 // under the License.
 package org.apache.cloudstack.api.command.user.account;
 
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
+
 import org.apache.log4j.Logger;
 
 import org.apache.cloudstack.api.APICommand;
 import org.apache.cloudstack.api.ApiConstants;
+import org.apache.cloudstack.api.ApiConstants.DomainDetails;
 import org.apache.cloudstack.api.BaseListDomainResourcesCmd;
 import org.apache.cloudstack.api.Parameter;
 import org.apache.cloudstack.api.ResponseObject.ResponseView;
 import org.apache.cloudstack.api.response.AccountResponse;
 import org.apache.cloudstack.api.response.ListResponse;
 
+import com.cloud.exception.InvalidParameterValueException;
 import com.cloud.user.Account;
 
 @APICommand(name = "listAccounts", description = "Lists accounts and provides detailed account information for listed accounts", responseObject = AccountResponse.class, responseView = ResponseView.Restricted, entityType = {Account.class},
@@ -55,6 +61,12 @@ public class ListAccountsCmd extends BaseListDomainResourcesCmd {
     @Parameter(name = ApiConstants.STATE, type = CommandType.STRING, description = "list accounts by state. Valid states are enabled, disabled, and locked.")
     private String state;
 
+    @Parameter(name = ApiConstants.DETAILS,
+               type = CommandType.LIST,
+               collectionType = CommandType.STRING,
+               description = "comma separated list of account details requested, value can be a list of [ all, resource, min]")
+    private List<String> viewDetails;
+
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
     /////////////////////////////////////////////////////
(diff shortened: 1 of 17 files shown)


server/src/main/java/com/cloud/api/query/dao/ProjectJoinDaoImpl.java
```diff
@@ -91,9 +94,11 @@ public ProjectResponse newProjectResponse(ProjectJoinVO proj) {
 
         //set resource limit/count information for the project (by getting the info of the project's account)
         Account account = _accountDao.findByIdIncludingRemoved(proj.getProjectAccountId());
-        AccountJoinVO accountJn = ApiDBUtils.newAccountView(account);
-        _accountJoinDao.setResourceLimits(accountJn, false, response);
-        response.setProjectAccountName(accountJn.getAccountName());
+        if (details.contains(DomainDetails.all) || details.contains(DomainDetails.resource)) {
+            AccountJoinVO accountJn = ApiDBUtils.newAccountView(account);
```
ustcweizhou: line 98 takes around 0.2 seconds for each project in my testing environment.
The actual time depends on how many accounts in your system.
```
AccountJoinVO accountJn = ApiDBUtils.newAccountView(account);
```

svenvogel:
@ustcweizhou what is used in the UI - all or min?

ustcweizhou:
@svenvogel if list projects with id, then details is not set so it will list all resource limits.
if list projects without id, then details is set to min, only project information and tags will be added

svenvogel:
@ustcweizhou i mean ... its a extension from #3306 ? so we need to add this in the ui to get them faster?

ustcweizhou:
@svenvogel both #3306 and #3331 will reduce the execution time while list projects.
#3306 aims to fix the issue caused by multiple tags created for projects.
#3331 aims to fix the issue when there are many accounts existing (similar issue while list domains fixed by #1901 for CLOUDSTACK-9405).

ustcweizhou:
@anuragaw added integration test as you suggested.

svenvogel:
> @svenvogel both #3306 and #3331 will reduce the execution time while list projects.
> #3306 aims to fix the issue caused by multiple tags created for projects.
> #3331 aims to fix the issue when there are many accounts existing (similar issue while list domains fixed by #1901 for CLOUDSTACK-9405). UI changes are included in this pull request.

@ustcweizhou all questions answered. now i understand it. okay i hope it will be faster 👍

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan  test

rhtyd:
@blueorangutan package

api/src/main/java/org/apache/cloudstack/api/command/user/account/ListAccountsCmd.java
```diff
@@ -55,6 +61,12 @@
     @Parameter(name = ApiConstants.STATE, type = CommandType.STRING, description = "list accounts by state. Valid states are enabled, disabled, and locked.")
     private String state;
 
+    @Parameter(name = ApiConstants.DETAILS,
+               type = CommandType.LIST,
+               collectionType = CommandType.STRING,
+               description = "comma separated list of account details requested, value can be a list of [ all, resource, min]")
+    private List<String> viewDetails;
```
rhtyd: details throughout CloudStack relates to an API arg that is of key:value map. Why not call something meaningful like `view` or verbosity etc?

DaanHoogland: listDetails seems appropriate, as well

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

api/src/main/java/org/apache/cloudstack/api/command/user/account/ListAccountsCmd.java
ustcweizhou: @rhtyd @DaanHoogland 
this is same as viewDetails in ListVMsCmd and ListHostsCmd.java
In some other API, details are in key/value pairs.

I agree 'view' or 'listdetails' make more sense.
I am afraid of backward compatibility.

DaanHoogland: You are right, @ustcweizhou . I think we can deprecate thiose and add better parameter-names there as well. It is not so much the internal name, 'viewDetails' as the API parameter name, ApiConstants.DETAILS (.equals("details")) which might be confusing as an external name, 
The proposal is to change the bit ```@Parameter(name = ApiConstants.DETAILS, ...```. As it is new I don't think we can have a compatibility issue there.
For my part this can not block this PR btw.

svenvogel:
@ustcweizhou @DaanHoogland is there any progress in this pr that we can get it to merge like #3306 and #3331. we test both and it seems faster. its not perfect but a good step forward. 👍

rhtyd:
@ustcweizhou can you fix the merge conflict? Thanks.

api/src/main/java/org/apache/cloudstack/api/command/user/account/ListAccountsCmd.java
ustcweizhou: @DaanHoogland good

> @ustcweizhou can you fix the merge conflict? Thanks.

@rhtyd rebased with latest master

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@ustcweizhou can you fix the merge conflict? Thanks.

ustcweizhou:
@rhtyd rebased with latest master

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3333, opened 2019-05-17
server: ssh-keygen in PEM format and reduce main systemvm patching script

rhtyd:
On first startup, the management server creates and saves a random
ssh keypair using `ssh-keygen` in the database. The command does
not specify keys in PEM format which is not the default as generated
by latest ssh-keygen tool.

The systemvmtemplate always needs re-building whenever there is a change
in the cloud-early-config file. This also tries to fix that by introducing a
`stage 2` bootstrap.sh where the changes specific to hypervisor detection
etc are refactored/moved. The initial cloud-early-config only patches
before the other scripts are called.

Additional commentary:
On latest CentOS7 release (CentOS Linux release 7.6.1810 (Core)) I hit a
bug where router_proxy (ssh) failed due to public key issue and kept
prompting for a passphrase (the default is blank/empty). By adding a
newline to the id_rsa.cloud private key it did not prompt the same and
ssh worked. By performing a visual diff on bytes of the file I found
that the DB version of the keypair and the version that gets put on the
KVM host had a missing `line feed (0xA0)` character. This patch fixes
the issue by ensuring we save the ssh keypair in the database exactly
as it is generated by ssh-keygen. Furthermore, I found that sometimes
the VR startup could fail if the VR took more time to patch, so I
increased the number of retries it previously used to do.

Screenshot showing the binary difference between ssh private key files (as on management server vs on the kvm host):

Screenshot from 2019-05-17 13-43-46

After the fix the checksums match on both the kvm host and on the management server:

Screenshot from 2019-05-17 13-59-30

Screenshot from 2019-05-17 13-59-41

SSH packages details on my CentOS7 env:
libssh2.x86_64                             1.4.3-12.el7_6.2            @updates
openssh.x86_64                             7.4p1-16.el7                @anaconda
openssh-askpass.x86_64                     7.4p1-16.el7                @base
openssh-clients.x86_64                     7.4p1-16.el7                @anaconda
openssh-server.x86_64                      7.4p1-16.el7                @anaconda

## Types of changes


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
The main issue with failing VR came out to be due to the ssh private key having 'OPENSS... PRIVATE KEY' instead of 'RSA PRIVATE KEY' that ssh trilead-ssh2 library expects. This happens when you update to latest openssh-client

@blueorangutan package

rhtyd:
@blueorangutan test matrix

rhtyd:
@blueorangutan package

rhtyd:
Kicked test with centos6 out of band.

rhtyd:
@ustcweizhou This is ready for review now, the issue is affecting CloudStack with latest versions of ssh-keygen primarily.

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtStartCommandWrapper.java
```diff
@@ -125,9 +125,14 @@ public Answer execute(final StartCommand command, final LibvirtComputingResource
                 // try to patch and SSH into the systemvm for up to 5 minutes
                 for (int count = 0; count < 10; count++) {
                     // wait and try passCmdLine for 30 seconds at most for CLOUDSTACK-2823
-                    libvirtComputingResource.passCmdLine(vmName, vmSpec.getBootArgs());
-                    // check router is up?
-                    final VirtualRoutingResource virtRouterResource = libvirtComputingResource.getVirtRouterResource();
+                    if (libvirtComputingResource.passCmdLine(vmName, vmSpec.getBootArgs())) {
+                        break;
+                    }
```
ustcweizhou: @rhtyd 
we pass cmdline to systemvm multiple times here, is because the operation passCmdLine will always succeed  but actually systemvm cannot get the data via /dev/vport01p1 inside it, so we pass the cmdline to systemvm around every 15 seconds (virtRouterResource.connect) until control ip is reachable.

rhtyd: @ustcweizhou the blind waiting and passing cmdline is not necessary now, in the new patch.sh script it waits until it's able to communicate with the qemu-guest-agent running inside the systemvm appliance. See: https://github.com/apache/cloudstack/blob/master/scripts/vm/hypervisor/kvm/patch.sh#L61

rhtyd: @ustcweizhou I can reduce that to 3 maybe.
I could consistently reproduce the issue with latest Ubuntu 19.10 with latest Centos7.6 used as kvm host.

rhtyd:
Tests LGTM, the one usage related tests failing on xenserver-71 env failed due to env issues:
```
  File "/usr/lib64/python2.7/site-packages/mysql/connector/connection_cext.py", line 182, in _open_connection
    sqlstate=exc.sqlstate)
OperationalError: 1043 (08S01): Bad handshake
```
However, the usage related tests are agnostic of hypervisor and have passed in other environments.

PaulAngus:
@rhtyd   https://github.com/apache/cloudstack/pull/3126 &3127 has been waiting to be merged in 4.11 and master for some time.  Can we merge that and then rebase this?

rhtyd:
@PaulAngus I can help review/merge #3126 after this PR.

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -1362,7 +1362,7 @@ private boolean checkOvsNetwork(final String networkName) {
     }
 
     public boolean passCmdLine(final String vmName, final String cmdLine) throws InternalErrorException {
-        final Script command = new Script(_patchScriptPath, 30 * 1000, s_logger);
+        final Script command = new Script(_patchScriptPath, 300 * 1000, s_logger);
```
anuragaw: Curious question - Is that a desired timeout value? Seems too high.

rhtyd: In environment with slow cpu and storage network, the timeout may be even higher. The logic of 5minute timeout comes from the libvirt wrapper command handler:
```
                // try to patch and SSH into the systemvm for up to 5 minutes
                for (int count = 0; count < 10; count++) {
```
For historic reasons, you can see CLOUDSTACK-2823

rhtyd:
Merging based on code reviews from Bobby, Abhishek and Anurag and smoke test results and Travis.

plugins/hypervisors/kvm/src/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
anuragaw: Thanks!

Issue #3334, opened 2019-05-17
Cannot enable static NAT on VM with NICs on multiple VPC tiers in UI


nvazquez:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
Network, VPC
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### CONFIGURATION
Advanced networking, VPC with multiple tier networks

##### OS / ENVIRONMENT
N/A

##### SUMMARY
When trying to enable static NAT on VM with NICs on multiple VPC tier networks on the 'Public IP addresses' router section, it fails with error: `VM ip <IP> address not belongs to the vm` where <IP> is the IP of the NIC tried to map for static NAT.

Note: The same process works under the tier network static NATs section, acquiring a new IP, but not from the VPC router section.

##### STEPS TO REPRODUCE

~~~
- Create VPC
- Add at least 2 tier networks
- Deploy VM on at least 2 of the VPC tier networks
- Network -> VPC -> Router -> Public IP addresses [1]
- Acquire new IP
- Enable static NAT: [2]
   - Select one of the tier networks
   - Select VM
   - Apply
~~~

[1] ![image](https://user-images.githubusercontent.com/5295080/58020060-73c28600-7add-11e9-91bc-5b4a322227ec.png)

[2] 
![image](https://user-images.githubusercontent.com/5295080/58020205-dddb2b00-7add-11e9-859f-c32d0188da9a.png)

##### EXPECTED RESULTS

~~~
Static NAT mapped correctly
~~~

##### ACTUAL RESULTS

~~~
VM ip <IP> address not belongs to the vm
~~~

PaulAngus:
@nvazquez please can you detail where in the UI you are trying to add this IP.

ustcweizhou:
patch

```
diff --git a/ui/scripts/vpc.js b/ui/scripts/vpc.js
index 8058f07..b7fac5b 100644
--- a/ui/scripts/vpc.js
+++ b/ui/scripts/vpc.js
@@ -39,7 +39,7 @@
                             url: createURL('listNics'),
                             data: {
                                 virtualmachineid: instance.id,
-                                nicId: instance.nic[0].id
+                                networkId: network.id
                             },
                             success: function(json) {
                                 var nic = json.listnicsresponse.nic[0];
```

nvazquez:
Done @PaulAngus thanks

ustcweizhou:
[...]

PR #3338, opened 2019-05-20
ui: fix enable static nat only towards first nic and not on any other interface

ustcweizhou:
## Description
When enable static nat in a vpc on UI, it only lists the primary and secondary ips of first nic of a vm, no matter which vpc tier is selected. 
The same issue happens when add a vm to load balancer.

Fixes: #3334

## Types of changes

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -91,7 +91,7 @@
                 url: createURL('listNics'),
                 data: {
                     virtualmachineid: instance.id,
-                    networkId: network.id
+                    networkid: (args.context.networkid != undefined) ? args.context.networkid : network.id
                 },
                 success: function(json) {
                     var nic = json.listnicsresponse.nic[0];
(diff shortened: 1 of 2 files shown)


ui/scripts/network.js
```diff
@@ -87,12 +87,21 @@
...
+            };
+            if (args.context.networkid != undefined) {
+                $.extend(data, {
+                    networkid: args.context.networkid
+                });
+            } else {
+                $.extend(data, {
+                    networkid: network.id
+                });
+            }
```
anuragaw: very minor suggestion- Alternate way of writing could be (just to make it smaller)-
```
var data = {
         virtualmachineid: instance.id,
         networkid: (args.context.networkid != undefined) ? args.context.networkid : network.id
}
```

ustcweizhou: @anuragaw 
Changed the code as you suggested, tested ok.
thanks for good suggestion.

nvazquez:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Purely UI changes, merged based on reviews.

PR #3344, opened 2019-05-21
server: return usage description with resource names and UUIDs

rhtyd:
The usage record descriptions have CloudStack's internal integer IDs
which makes it difficult for users to read their usages. This PRs
introduces a new API boolean flag `oldformat` which when set to true
would return the older description format, otherwise by default
listUsageRecords will process and return description with names and
UUIDs of resources.

## Types of changes


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

api/src/main/java/org/apache/cloudstack/api/command/admin/usage/ListUsageRecordsCmd.java
```diff
@@ -81,6 +82,10 @@
     @Parameter(name = ApiConstants.INCLUDE_TAGS, type = CommandType.BOOLEAN, description = "Flag to enable display of Tags for a resource")
     private Boolean includeTags;
 
+    @Parameter(name = ApiConstants.OLD_FORMAT, type = CommandType.BOOLEAN, description = "Flag to enable description rendered in old format with no processing")
```
DaanHoogland: should this have a default (of true)?

api/src/main/java/org/apache/cloudstack/api/command/admin/usage/ListUsageRecordsCmd.java
```diff
@@ -145,6 +150,9 @@ public void setUsageId(String usageId) {
         this.usageId = usageId;
     }
 
+    public boolean getOldFormat() {
+        return oldFormat != null && oldFormat;
```
DaanHoogland: this is a way to make the default be false. It should be in the description of the Parameter-annotation.

api/src/main/java/org/apache/cloudstack/api/command/admin/usage/ListUsageRecordsCmd.java
andrijapanicsb: 1. old behaviour should be the default, unless the documentation is (somewhere...) updated.
2. word "processing" - does this mean that there will be additional CPU usage and DB hammering on top of existing (I know some people's usage_jobs taking many hours to complete - would like to understand performance consequences)

api/src/main/java/org/apache/cloudstack/api/command/admin/usage/ListUsageRecordsCmd.java
rhtyd: Will fix description.

api/src/main/java/org/apache/cloudstack/api/command/admin/usage/ListUsageRecordsCmd.java
rhtyd: I'll update the docs. Will fix description.

rhtyd:
@DaanHoogland @andrijapanic I've sent the docs PR https://github.com/apache/cloudstack-documentation/pulls
And updated the API doc description

api/src/main/java/org/apache/cloudstack/api/command/admin/usage/ListUsageRecordsCmd.java
DaanHoogland: I agree that this default is sensible and I don't think the backwards incompatibility of not using internal ids anymore is going to be very hurtful as the processing of such ids is more intensive than processing uuids using the api. maybe a point of discussion for some, so let's be extra cautious.

andrijapanicsb:
LGTM

Possible consideration is the performance penalty due to API having to query the main "cloud" DB.
If no crazy JOINs, than it should be fine.

@rhtyd @DaanHoogland ^^^

rhtyd:
@andrijapanic thanks, sounds good to me. This is a given that using UUIDs in the description will incur some processing. In large environments this will make the list API slightly slower than before, however all the UUID translation logic makes individual listing (select * query) so no crazy joins but multiple select queries may be executed depending on the type of the record. In few cases I've optimised not querying for a VM/volume/template resource which is already queried in the beginning on the block.

andrijapanicsb:
Sounds good @rhtyd

LGTM.

PR #3358, opened 2019-05-25
Update vmware reservations description

andrijapanicsb:
we do NOT always reserve VMware CPU/RAM resources - only when "vmware.reserve.cpu" or "vmware.reserve.mem" setting is set to TRUE - AND we do so, irelevant if overprovisioning is active or not. Verified for both system VMs and user VMs.

## Types of changes

## How Has This Been Tested?

CPU and RAM overprovisioning was active (set to 2.0), while vmware.reserve.cpu and vmware.reserve.mem was set to TRUE or FALSE and the end result was observed in vCenter (6.5).
The same test repeated, with CPU and RAM overprovisioning set to 1.0 (no overprovisioning).


rhtyd:
LGTM @andrijapanicsb should we say in the descripotion that when the global settings are true, the reservation value is considered based on the overprovisioning factors.

rhtyd:
LGTM, merging based on Travis job#1 because the change is strictly in static string description and does not require full regression test run.

Issue #3353, opened 2019-05-24
4.11.2 HA VRs with Second Public NIC get different routes between Startup and Failover


richardlawley:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### CONFIGURATION

Advanced Networking, HA VRs

##### OS / ENVIRONMENT

XenServer 6.5

##### SUMMARY

When a network has public IPs from multiple ranges and a VM has a Static NAT rule from the second IP range (different range from the Source NAT IP), the outbound IP changes when a VR fails over.  This is because default routes for each interface are not restored when a failover occurs, but are added correctly when the VR is built.

##### STEPS TO REPRODUCE

1. Create zone with 2 or more public IP ranges
2. Create an advanced network, allocate IPs until you get IPs from both networks (or use the database to assign a specific IP)
3. Create a VM and assign a Static NAT IP to the IP from the second range
4. Reboot both VRs at the same time to get a "clean" situation
5. Log into Master VR
6. Run `ip route table list all > /root/route.1`
7. Run `systemctl restart keepalived` to force a failover without rebuilding VR
8. Log into the new VR and fail back using `systemctl restart keepalived`
9. Run `ip route table list all > /root/route.2`
10. Compare routing tables with diff /root/route.1 /root/route.2

##### EXPECTED RESULTS

~~~
Routing table should be identical between first build and failover
~~~

##### ACTUAL RESULTS

Default routes for each interface were missing, resulting in the outgoing traffic from the VM with static NAT originating from the Source NAT IP

~~~
# diff /root/route.1 /root/route.2
2d1
< default via X.X.X.X dev eth2 table Table_eth2 proto static 
6d4
< default via X.X.X.X dev eth3 table Table_eth3 proto static
~~~

We are able to fix this manually by re-adding the route:
~~~
ip route add default via X.X.X.X table Table_eth3 proto static
~~~
NB - the eth2 route doesn't seem to be necessary, but adding the eth3 route fixes the originating IPs.

ustcweizhou:
This is one of the issues that happen on network/vpc with public IPs in mutiple subnets.
We have seen many issues in this scenario from 4.7.  Some have already been fixed. Some issue still exist

richardlawley:
These routes also get restored when a new static NAT association is made.

richardlawley:
[...]

Issue #3168, opened 2019-02-06
Static NAT with multiple public interfaces uses wrong outgoing IP


richardlawley:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### CONFIGURATION

Advanced Networking, XenServer 6.5, Non-VPC

##### SUMMARY

This is a regression, caused by f60f3ce I believe.  When a VR has public IPs on multiple interfaces, a static NAT rule which is not on the first public interface will have outgoing traffic originating from the Source NAT IP.  I believe this is because the logic of marking packets was changed from interface number to 100+interface number everywhere *except* where the static NAT rules are configured in `configure.py`.

I've been able to fix this temporarily in my own VR by changing [configure.py L930](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/configure.py#L930) from `hex(int(device[len("eth"):]))` to `hex(100 + int(device[len("eth"):]))`

##### STEPS TO REPRODUCE

* Assign public IP from different subnet to VR (so that VR gets eth3 created).  
* Deploy a VM on this network
* Create static NAT rule for this new public IP to the new VM
* Run `curl -s http://ipecho.net/plain` inside the VM

##### EXPECTED RESULTS

Output should be the Static NAT IP

Created firewall rule in mangle / PREROUTING should be: `state NEW MARK set 0x67`

##### ACTUAL RESULTS

Output is the Source NAT IP

Created firewall rule in mangle / PREROUTING is be: `state NEW MARK set 0x3`

richardlawley:
There were a couple of other places I could find the mark being used:
* configure.py L961 - `hex(int(public_fwinterface[3:]))`
* CsAddress.py L261 - `self.dnum = hex(int(dev[3:]))`

I'm not entirely sure what these are each for and whether they should all be updated.

Issue #3352, opened 2019-05-24
4.11.2 HA VRs with Second Public NIC do not get constant MAC


richardlawley:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VRs
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### CONFIGURATION

XenServer 6.5, Advanced Networking

##### OS / ENVIRONMENT

N/A

##### SUMMARY

When an advanced network has multiple public interfaces, the second (and subsequent) interfaces do not have the same MAC address on both VRs.  This means that after a failover, connectivity with the IPs on the second+ ranges is interrupted until the ARP entry times out.

##### STEPS TO REPRODUCE

1.  Create zone with 2 or more public IP ranges
2.  Create an advanced network, allocate IPs until you get IPs from both networks (or use the database to assign a specific IP)
3.  Create a VM and assign a Static NAT IP to the IP from the second range
4.  Log into both VRs, compare the MAC addresses of eth2 and eth3 on both VRs

##### EXPECTED RESULTS

~~~
MAC addresses for eth3 should be the same on both VRs
~~~

##### ACTUAL RESULTS

~~~
MAC addresses for eth3 are different.
~~~

richardlawley:
The additional public NICs do not seem to have a database entry - I was hoping to be able to adjust it as a short-term fix.

richardlawley:
My workaround for now is to send garp messages when transitioning to master (037b1e9) but it would be better if the MAC address was constant, like for eth2.

rhtyd:
Future note: explore a way to make the mac address constant.

PR #3361, opened 2019-05-29
Fix 4.11 VR Issues with Multiple Public Subnets

richardlawley:
## Description

This PR resolves 2 issues related to Virtual Routers with multiple public interfaces, and works around a third.

1. Fixes: #3353 - Adds missing throw routes for eth0/eth1 to eth3+ when there are >1 public IPs 
2. Fixes: #3168 - Incorrect marks set on some static NAT rules (some code references were changed from `hex(int(interfacenum))` to `hex(100 + int(interfacenum))` - this change just adds the remaining ones
3. Works around #3352 - Sends Gratuitous ARP messages when a HA VR becomes master to work around the problem of the MAC address being different between HA VRs.  If that issue is fixed properly (i.e. a database entry for the subsequent interfaces so they can be static) then this is unnecessary, though should not cause any problems.

## Types of changes

## How Has This Been Tested?

These changes form part of the ISO we are deploying in production.  Upon upgrading from 4.8 to 4.11.2, we ran into the issues described above and fixed them in a scripts ISO we have customised for our deployment.  

Our deployments are all advanced networking/non-VPC.  These changes are mostly within conditionals which should stop them making any difference to a "standard" advanced VR (i.e. single public interface).

### DIFF
--- a/systemvm/debian/opt/cloud/bin/configure.py
+++ b/systemvm/debian/opt/cloud/bin/configure.py
@@ -858,7 +858,7 @@ class CsForwardingRules(CsDataBag):
                 rule['protocol'],
                 rule['protocol'],
                 public_fwports,
-                hex(int(public_fwinterface[3:]))
+                hex(100 + int(public_fwinterface[3:]))
               )
         fw6 = "-A PREROUTING -d %s/32 -i %s -p %s -m %s --dport %s -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" % \
               (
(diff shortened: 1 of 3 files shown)


rhtyd:
@b

rhtyd:
@blueorangutan package

richardlawley:
@rhtyd The third issue I would prefer to be solved by making the MAC address constant somehow, but I don't really know enough about why it's been done like it is to know what problems it may cause.  Hopefully it can be solved properly still, but I figured it wasn't a trivial task and might not make it into the next release!

rhtyd:
@richardlawley fair enough, I've fixed few pylint errors caught by travis. Will kick regression tests now.
@blueorangutan test

rhtyd:
Test lgtm

Issue #3177, opened 2019-02-13
Static NAT causes duplicate firewall rules to be added


richardlawley:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### CONFIGURATION

Advanced Networking

##### OS / ENVIRONMENT

n/a

##### SUMMARY

When two static NAT IPs exist on a network, every time a firewall rule is added or removed, duplicate CONNMARK rules are added to mangle/PREROUTING.  Over time and on a busy network, this causes a significant number of duplicate rules and an increase in processing time.

I believe this is caused by the rules being defined with `-I PREROUTING` instead of `-A PREROUTING`, as the code appears to be trying to match them to the output of iptables-save.  This may have been introduced in 40d77460386342126fb5533a9139bf6d08137d1d.

I've fixed this locally by changing in `configure.py`:
```
self.fw.append(["mangle", "",
    "-I PREROUTING -s %s/32 -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" %
    rule["internal_ip"]])
```
to
```
self.fw.append(["mangle", "front",
    "-A PREROUTING -s %s/32 -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" %
    rule["internal_ip"]])
```
However, there are more instances of adding rules with `-I` in the file related to VPC and VPNs, which I don't have a way of testing at the moment.

##### STEPS TO REPRODUCE

* Create advanced network
* Deploy two VMs
* Enable static NAT for both IPs
* Create one Firewall rule
* Log onto VR and run `iptables -t mangle -nL PREROUTING --line-numbers`
* Add a rule and repeat

For a simpler repeat repro, log onto the VR, then:
* Find a firewall rules file in /var/cache/cloud/processed/firewall_rules,json.XXXXX.gz and gunzip it
* `iptables -t mangle -F PREROUTING --line-numbers`
* `/opt/cloud/bin/configure /var/cache/cloud/processed/firewall_rules,json.XXXXX`
* `iptables -t mangle -F PREROUTING --line-numbers`

##### EXPECTED RESULTS

No increase in the number of rules in PREROUTING

~~~
# iptables -t mangle -nL PREROUTING --line-numbers
Chain PREROUTING (policy ACCEPT)
num  target     prot opt source               destination
[...]

PR #3366, opened 2019-05-31
Fix rule duplication with static NAT rules

richardlawley:
## Description

The VR code has provision for inserting rules at the top or bottom by specifying "front" as the second parameter to `self.fw.append`.  However, there are a number of cases where someone has been unaware of this and added a rule with the pattern `self.fw.append(["mangle", "", "-I PREROUTING"...`.  This causes the code to check for the rule already being present to fail, and duplicate rules end up being added.

This PR fixes two of these cases which apply to adding static NAT rules.  I am aware of more of these cases, but I don't have the ability to easily test the outcome of fixing them.  I'm happy to add these in if you're confident that the automated tests will be sufficient.  Searching for "-I (case sensitive) finds these.

The [code for dealing with "front"](https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/cs/CsNetfilter.py#L166) is included below to show that this shouldn't have any ill effects:

~~~
if fw[1] == "front":
    cpy = cpy.replace('-A', '-I')
~~~

Fixes: #3177 

## Types of changes

## How Has This Been Tested?

Local tests were performed (see #3177 for details of symptoms), and we have had this fix deployed in production on advanced networks for a couple of months.

### DIFF
--- a/systemvm/debian/opt/cloud/bin/configure.py
+++ b/systemvm/debian/opt/cloud/bin/configure.py
@@ -922,11 +922,11 @@ class CsForwardingRules(CsDataBag):
         if device is None:
             raise Exception("Ip address %s has no device in the ips databag" % rule["public_ip"])
 
-        self.fw.append(["mangle", "",
-                        "-I PREROUTING -s %s/32 -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" %
+        self.fw.append(["mangle", "front",
+                        "-A PREROUTING -s %s/32 -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" %
                         rule["internal_ip"]])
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

rhtyd:
Can you review this - @ustcweizhou @anuragaw @shwstppr ?
@blueorangutan package

rhtyd:
Thanks for the PR @richardlawley - do you think it can cause regression when the VR has multiple public IPs (i.e multiple public nics) as this reverts changes from https://github.com/apache/cloudstack/commit/40d77460386342126fb5533a9139bf6d08137d1d

rhtyd:
Nevermind I compared your changes with what Jayapal was trying to do, LGTM.

rhtyd:
@blueorangutan test

ustcweizhou:
I was just testing it. it works fine.

PR #3368, opened 2019-05-31
[WIP DO NOT MERGE] server: fix public IP addresses filtering

### LINKED ISSUE #3295: listPublicIPAddresses return not filtered for current network
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
* CloudStack GUI

##### CLOUDSTACK VERSION
* ACS 4.11.2

##### CONFIGURATION
* Advanced zone
* 2 x shared networks in "setup" state

##### OS / ENVIRONMENT
* N/A

##### SUMMARY
* GUI calls listPublicIPAddresses as follows and does not filter for current network when using the "change IP address" function:
command=listPublicIpAddresses&response=json&allocatedonly=false&networkid=01891a96-1a86-4401-948a-ed434c5be1ba&forvirtualnetwork=false
* As a result all IP addresses from both shared networks are returned, allowing the end user to select an IP address which can not be assigned to the NIC in question
* Checking http://cloudstack.apache.org/api/apidocs-4.11/apis/listPublicIpAddresses.html  it seems the parameter "networkid" is not actually implemented, however this plays out as follows:
** Documentation specifies this as parameter "associatednetworkid"
** CloudMonkey etc. will return "associatednetworkid" when querying input parameters
** However the API will not actually filter for "networkid" and using "associatednetworkid" returns 0 nil records, hence matching does not work

##### STEPS TO REPRODUCE
* Create two shared networks and make sure they are in "setup" state
* Create VM with NIC assigned to shared net 1
* Try to use the "change IP" function in the GUI 
* Observe this returns a list of ALL shared IP addresses for both shared networks
* For completeness sake try to assign an IP address on the second shared network and observe this failing whether VM is running or not

##### EXPECTED RESULTS

* GUI should return only IP addresses on current shared network

##### ACTUAL RESULTS
* GUI returns list of IP addresses from both shared networks

shwstppr:
## Description
Fixes #3295 

listPublicIPAddresses API was missing **networkid** parameter and it has been used in UI to list public IP addresses while changing IP address for a VM.

_networkid_ parameter is added to listPublicIPAddresses API which would allow filtering public IP addresses based on the source network. This will help to show relevant IP addresses when users try to change assigned IP address for a VM NIC.

## Types of changes

Screenshot from 2019-05-31 16-43-10

## How Has This Been Tested?
UI and cmk

### DIFF
--- a/server/src/main/java/com/cloud/server/ManagementServerImpl.java
+++ b/server/src/main/java/com/cloud/server/ManagementServerImpl.java
@@ -37,7 +37,6 @@ import javax.crypto.spec.SecretKeySpec;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
-import com.cloud.storage.ScopeType;
 import org.apache.cloudstack.acl.ControlledEntity;
 import org.apache.cloudstack.affinity.AffinityGroupProcessor;
 import org.apache.cloudstack.affinity.dao.AffinityGroupVMMapDao;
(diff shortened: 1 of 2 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -1893,7 +1894,8 @@ private boolean hasSuitablePoolsForVolume(final VolumeVO volume, final Host host
         sb.and("vlanDbId", sb.entity().getVlanId(), SearchCriteria.Op.EQ);
         sb.and("id", sb.entity().getId(), SearchCriteria.Op.EQ);
         sb.and("physicalNetworkId", sb.entity().getPhysicalNetworkId(), SearchCriteria.Op.EQ);
-        sb.and("associatedNetworkIdEq", sb.entity().getAssociatedWithNetworkId(), SearchCriteria.Op.EQ);
+        sb.and("associatedNetworkId", sb.entity().getAssociatedWithNetworkId(), SearchCriteria.Op.EQ);
+        sb.and("sourceNetworkId", sb.entity().getSourceNetworkId(), SearchCriteria.Op.EQ);
```
ravening, line 1898: @shwstppr any reason why `sourceNetworkId` is chosen instead of `networkId` parameter?

PR #2246, opened 2017-08-23
CLOUDSTACK-10046 checksum validation for any java supported Digests-type

DaanHoogland:
This fixes a very old regression where the checksum wouldn't be checked on downloaded templates.

It checks again now but also allows to specify an algorithm like "{SHA-1}98765....". No algorithm assumes md5 ("MD5") for now

PR #3371, opened 2019-06-03
Fix virtual template size for managed storage for KVM / refactor cloud-install-sys-tmplt

svenvogel:
## Description

* cloud-install-sys-tmplt: add function to calculate the virtual size of template for managed storage like netapp solidfire so that the copy process is working from NFS to iSCSI correctly
* add qemu-img (CentOS) / qemu-utils (Debian) packages to management build dependency
* refactor cloud-install-sys-tmplt to be more verbose and user friendly help
* createtmplt.sh: remove unused chksum flag

image

## Types of changes

## How Has This Been Tested?

tested manually in our production environment with different files like vhd, qcow2 and ova

### DIFF
--- a/scripts/storage/secondary/cloud-install-sys-tmplt
+++ b/scripts/storage/secondary/cloud-install-sys-tmplt
@@ -1,5 +1,4 @@
 #!/bin/bash
-# $Id: installrtng.sh 11251 2010-07-23 23:40:44Z abhishek $ $HeadURL: svn://svn.lab.vmops.com/repos/vmdev/java/scripts/storage/secondary/installrtng.sh $
 
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
@@ -20,15 +19,30 @@
 
 
 usage() {
-  printf "Usage: %s: -m <secondary storage mount point> -f <system vm template file> [-h <hypervisor name: kvm|vmware|xenserver|hyperv|ovm3> ] [ -s <mgmt server secret key, if you specified any when running cloudstack-setup-database, default is password>][-u <Url to system vm template>] [-F <clean up system templates of specified hypervisor>] [-e <Template suffix, e.g vhd, ova, qcow2>] [-o <Database server hostname or ip, e.g localhost>] [-r <Database user name, e.g root>] [-p <mysql database port>] [-d <Database password. Fllowed by nothing if the password is empty>]\n" $(basename $0) >&2
-  printf "or\n" >&2
-  printf "%s: -m <secondary storage mount point> -u <http url for system vm template> [-h <hypervisor name: kvm|vmware|xenserver|hyperv|ovm3> ] [ -s <mgmt server secret key>]\n" $(basename $0) >&2
+  printf "\nUsage: %s:\n\t-m secondary storage mount point\n\t-f system vm template file\n\t-h hypervisor name: kvm|vmware|xenserver|hyperv|ovm3\n\t-s mgmt server secret key, if you specified any when running cloudstack-setup-database, default is password\n\t-u Url to system vm template\n\t-F clean up system templates of specified hypervisor\n\t-e Template suffix, e.g vhd, ova, qcow2\n\t-o Database server hostname or ip, e.g localhost\n\t-r Database user name, e.g root\n\t-p mysql database port\n\t-d Database password. Followed by nothing if the password is empty\n\n" $(basename $0) >&2
+  printf "\tor\n"
+  printf "\nUsage: %s:\n\t-m secondary storage mount point\n\t-u http url for system vm template\n\t-h hypervisor name: kvm|vmware|xenserver|hyperv|ovm3\n\t-s mgmt server secret key\n\n" $(basename $0) >&2
 }
 
+# Usage: e.g. failed $? "this is an error"
 failed() {
-  echo "Installation failed"
-  exit $1
+  local returnval=$1
+  local returnmsg=$2
+  
+  # check for an message, if there is no one dont print anything
+  if [[ -z $returnmsg ]]; then
+    :
+  else
+    echo -e $returnmsg
+  fi
+  if [[ $returnval -eq 0 ]]; then
+    return 0
+  else
+    echo "Installation failed"
+    exit $returnval
+  fi
 }
+
 #set -x
 mflag=
 fflag=
@@ -42,7 +56,15 @@ dbUser="root"
 dbPassword=
 dbPort=3306
 jasypt='/usr/share/cloudstack-common/lib/jasypt-1.9.2.jar'
-while getopts 'm:h:f:u:Ft:e:s:o:r:d:p:'# OPTION
+
+# check if first parameter is not a dash (-) then print the usage block
+if [[ ! $@ =~ ^\-.+ ]]; then
+	usage
+	exit 0
+fi
+
+OPTERR=0
+while getopts 'm:h:f:u:Ft:e:Ms:o:r:d:p:'# OPTION
 do
   case $OPTION in
   m)    mflag=1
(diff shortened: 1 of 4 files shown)


svenvogel:
@rhtyd please run your check again so that we can find the problem.

rhtyd:
Thanks
@blueorangutan package

DennisKonrad:
@rhtyd From my point of view the code from sven is a cleanup/improvement. How can one see now what causes the problem?

For me the problem is now, how can I decide if there's a problem with svens code or is there something that should change in Trillian?

DennisKonrad:
https://github.com/shapeblue/Trillian

Is the code here up to date? There's no way to set something like this up in a few days but if we would choose to do it: Would an installation using the code below behave (more or less) the same as blueorangutan?

rhtyd:
I'll have to review if the flags have changed. From a users point of view, the command should be backward compatible. Trillian is working OK with other PRs so likely the issue is in the PR, or the script was changed such that the usage is not backward compatible. 
@blueorangutan package

rhtyd:
@DennisKonrad you can test against a local kvm based nested env in few hours: https://github.com/rhtyd/monkeybox
@blueorangutan package

scripts/storage/secondary/createtmplt.sh
```diff
@@ -131,92 +123,66 @@ do
...
-                let "size>>=10"
-		ulimit -f $size
-		;;
-  h)	hflag=1
-		hvm="true"
-		;;
-  u)	cleanup="true"
-		;;
-  v)	verbose="true"
-		;;
```
rhtyd: @svenvogel @DennisKonrad several flags/opts are removed from the script, do you think it will cause regression in case the script is evoked via mgmt server or ssvm agent?

DennisKonrad:
> @DennisKonrad you can test against a local kvm based nested env in few hours: https://github.com/rhtyd/monkeybox
> @blueorangutan package

Isn't the whole use case of Trillian to spin up an cloudstack env on the fly to run the smoke tests? Also you pointed out the merge guideline states the smoketests should pass. So it seems to me there's not much use in manually setting up monkey box to test. 

Maybe I don't understand the difference between Monkeybox and Trillian. Could you elaborate on this?

Also the question would be of importance to me: Can one use the trillian repo to get (more or less) the same results (as blueorangutan)?

rhtyd:
@DennisKonrad yes, but Trillian requires a lot of effort to setup and manage, and can span multiple hosts/machines. Monkeybox is aimed for developers who can setup a nested kvm based dev-test env on a single machine. Marvin based smoketests can run on both env.
The merge guideline require that smoketests should pass and the author/CI-runner to prove that no regressions were seen, irrespective of whether the tests were run manually or via a script/job.

BO is a github bot account that triggers jobs in our internal Jenkins which uses Trillian to deploy and run tests against an env.

DennisKonrad:
@rhtyd Ok, that clears up things a lot. But what about those two. How can one make sense of the test result there?

https://builds.apache.org/job/cloudstack-pr-analysis
https://travis-ci.org/apache/cloudstack

The jenkins fails with some code that doesnt even belong to the PR and the Travis tests mostly seem to pass this time even though they didn't with the same code before??

rhtyd:
@blueorangutan test

rhtyd:
@DennisKonrad master needs stabilisation, which is why you're seeing intermittent errors. The jenkins job is not reliable, I'm not sure how to manage that, I usually ignore that like most people. I check Travis test results and smoketests via BO.

4.11 for example is most stable branch where Travis mostly passes along with smoketests. We'll need to put some effort on master to ensure Travis test runs are reliable, this would require either fixing the failing tests or the code that caused that regression.

scripts/storage/secondary/createtmplt.sh
svenvogel: @rhtyd yes maybe. can you see how/where the createtmplt.sh will called? can you see wich options are important?

svenvogel: @rhtyd if i could know which are important for check i can put them back.

DennisKonrad:
@rhtyd So to resolve this issue can you provide the calls to those scripts to @svenvogel so he can maybe adapt this PR to work with the BO that's currently in place?

I see no other way to resolve this issue.

scripts/storage/secondary/createtmplt.sh
rhtyd: Let me review the results, do code investigation and get back to you tomorrow.

svenvogel: @rhtyd thank you for help. you investigation will help to find a good solution.

svenvogel: @rhtyd could you find something?

rhtyd:
@svenvogel @DennisKonrad here are bunch of issues with this PR:
- It breaks the cloud-install-sys-tmplt usage which is not backward compatible anymore, to simply setup/seed the initial systemvmtemplate before deploying a zone fails (it should not need to check if qemu-img etc is available). Trillian logs for reference:
```
20:08:11 TASK [cloudstack-config : Install System VM templates] *************************
20:08:12 failed: [pr3371-t3639-kvm-centos7-mgmt1] (item=-u http://staging.yadav.xyz/systemvmtemplate/custom/4113/systemvmtemplate-4.11.3-kvm.qcow2.bz2 -h kvm -F) => {"changed": false, "cmd": ["/usr/share/cloudstack-common/scripts/storage/secondary/cloud-install-sys-tmplt", "-m", "/mnt/secstoragetmp", "-u", "http://staging.yadav.xyz/systemvmtemplate/custom/4113/systemvmtemplate-4.11.3-kvm.qcow2.bz2", "-h", "kvm", "-F"], "delta": "0:00:00.195024", "end": "2019-06-03 14:38:29.618615", "failed": true, "item": "-u http://staging.yadav.xyz/systemvmtemplate/custom/4113/systemvmtemplate-4.11.3-kvm.qcow2.bz2 -h kvm -F", "rc": 2, "start": "2019-06-03 14:38:29.423591", "stderr": "which: no qemu-img in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin)", "stderr_lines": ["which: no qemu-img in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin)"], "stdout": "Please install qemu-img: command\non CentOS run \"yum install qemu-img\"\non Ubuntu/Debian run \"apt-get install qemu-utils\"\nInstallation failed", "stdout_lines": ["Please install qemu-img: command", "on CentOS run \"yum install qemu-img\"", "on Ubuntu/Debian run \"apt-get install qemu-utils\"", "Installation failed"]}
```
- You can find usage of the second script using `git grep createtmplt.sh`, I found that the script is used both the ssvm agent and kvm to process templates. By removing flags that once were passed, it breaks the usage of the script:
https://github.com/apache/cloudstack/blob/master/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java#L723

Given this is a refactoring/cleanup PR, this ought to be backward compatible wrt script execution usages, i.e. don't remove flags and don't replace old flags, without proper documentation don't change how script is used.

svenvogel:
@rhtyd @syed thanks for help. i will introduce a flag for managed storage and revert the necessary changes back for createtemplt.sh script. after that there should be not a problem. after that we can run the next test.

svenvogel:
@rhtyd how can i test it by self with "@blue...."?

rhtyd:
@svenvogel to setup a dev-test env;
If you're on Mac/Windows, the basic process is to setup a VM with its own private/host network with a NAT and setup the host, storage on the system.
If you're on Linux you can use https://github.com/rhtyd/monkeybox

I would encourage that you validate, iterate and test your changes on a local env, as the changes in the file still remove flags from the script and can potentially fail the env deployment again.

svenvogel:
@rhtyd can you rekick a test for me?

svenvogel:
@blueorangutan test

svenvogel:
> @blueorangutan test

rhtyd:
@blueorangutan package

ustcweizhou:
@rhtyd 
qemu-img is needed to get the virtual size of systemvm template in QCOW2 format.
In the past systemvm template can still be installed even qemu-img is not installed on management server. With this PR it will not continue.
It breaks some deployment indeed, However I think this change makes sense.

Is it possible to install qemu-img (centos6/centos7) or qemu-utils (ubuntu) when set up a Trillian testing environment ?



> @svenvogel @DennisKonrad here are bunch of issues with this PR:
> 
> * It breaks the cloud-install-sys-tmplt usage which is not backward compatible anymore, to simply setup/seed the initial systemvmtemplate before deploying a zone fails (it should not need to check if qemu-img etc is available). Trillian logs for reference:
> 
> ```
> 20:08:11 TASK [cloudstack-config : Install System VM templates] *************************
> 20:08:12 failed: [pr3371-t3639-kvm-centos7-mgmt1] (item=-u http://staging.yadav.xyz/systemvmtemplate/custom/4113/systemvmtemplate-4.11.3-kvm.qcow2.bz2 -h kvm -F) => {"changed": false, "cmd": ["/usr/share/cloudstack-common/scripts/storage/secondary/cloud-install-sys-tmplt", "-m", "/mnt/secstoragetmp", "-u", "http://staging.yadav.xyz/systemvmtemplate/custom/4113/systemvmtemplate-4.11.3-kvm.qcow2.bz2", "-h", "kvm", "-F"], "delta": "0:00:00.195024", "end": "2019-06-03 14:38:29.618615", "failed": true, "item": "-u http://staging.yadav.xyz/systemvmtemplate/custom/4113/systemvmtemplate-4.11.3-kvm.qcow2.bz2 -h kvm -F", "rc": 2, "start": "2019-06-03 14:38:29.423591", "stderr": "which: no qemu-img in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin)", "stderr_lines": ["which: no qemu-img in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin)"], "stdout": "Please install qemu-img: command\non CentOS run \"yum install qemu-img\"\non Ubuntu/Debian run \"apt-get install qemu-utils\"\nInstallation failed", "stdout_lines": ["Please install qemu-img: command", "on CentOS run \"yum install qemu-img\"", "on Ubuntu/Debian run \"apt-get install qemu-utils\"", "Installation failed"]}
> ```
> 
> * You can find usage of the second script using `git grep createtmplt.sh`, I found that the script is used both the ssvm agent and kvm to process templates. By removing flags that once were passed, it breaks the usage of the script:
>   https://github.com/apache/cloudstack/blob/master/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java#L723
> 
> Given this is a refactoring/cleanup PR, this ought to be backward compatible wrt script execution usages, i.e. don't remove flags and don't replace old flags, without proper documentation don't change how script is used.

rhtyd:
Okay @ustcweizhou, I'll try again: https://github.com/shapeblue/Trillian/commit/2f53242aeb9c8e2926b8ef129b6754e5348edc97
@blueorangutan package

rhtyd:
@blueorangutan test

svenvogel:
@rhtyd are the checks ok now?

rhtyd:
@svenvogel have patience please, a typical test takes 6-10hours to complete. What I can confirm is that initial deployment seems to be working.

ustcweizhou:
@rhtyd it still failed. might qemu-img not be installed ?

svenvogel:
@ustcweizhou thanks for help. can you take a look to how we can fix this. 

@rhtyd can you see where trillian has a problem?

scripts/storage/secondary/createtmplt.sh
rhtyd: @svenvogel cc @ustcweizhou  - can you put back all the flags back? The issue is likely that your changes are breaking usage of this script. You can use `git grep <script name>` and use an IDE to find usages of the script.

svenvogel:
@ustcweizhou what do you think? for me take the option back as dummy " // not used for now " looks like the best way?

@rhtyd can you add qemu-img in the trillian test?

ustcweizhou:
@svenvogel it looks the issue is not related to "-s", but one of "-n", "-t" and "-f" is not set

svenvogel:
@rhtyd can you run your check again that i can see which arg will be called and failed. i changed temporary the output of the script. thanks! 👍

svenvogel:
@rhtyd can you run the blueorangutan/trillian again?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

ustcweizhou:
@svenvogel 
I tested in my testing environment. it can be fixed by
```
root@s-158-VM:~# diff -u /usr/local/cloud/systemvm/scripts/storage/secondary/createtmplt.sh /usr/local/cloud/systemvm/scripts/storage/secondary/createtmplt.sh.new
--- /usr/local/cloud/systemvm/scripts/storage/secondary/createtmplt.sh  2019-06-13 11:34:00.119649777 +0000
+++ /usr/local/cloud/systemvm/scripts/storage/secondary/createtmplt.sh.new      2019-06-13 13:15:11.475953400 +0000
@@ -112,7 +112,7 @@
 fi

 OPTERR=0
-while getopts 't:n:f:' OPTION
+while getopts 'vuht:n:f:s:c:d:S:' OPTION
 do
   case $OPTION in
   t)   tflag=1
@@ -124,9 +124,6 @@
   f)   fflag=1
                tmpltimg="$OPTARG"
                ;;
-  ?)   usage
-               exit 2
-               ;;
   esac
 done
```

svenvogel:
@ustcweizhou great thanks you! can you see is there any specific flag which is lost or do we need all flags? maybe we dont need a (v and u and S)

```
-while getopts 't:n:f:' OPTION
+while getopts 'vuht:n:f:s:c:d:S:' OPTION
```

svenvogel:
@ustcweizhou i found the needed flags. i need to think about it. hmm
https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/template/DownloadManagerImpl.java#L396-L427

ustcweizhou:
@svenvogel yes, you found it
here is an example 
```
/usr/local/cloud/systemvm/scripts/storage/secondary/createtmplt.sh -s 3 -S 53687091200 -d systemvm-test -h -n 016960d4-00f7-38a4-86b2-5d7258be7586.qcow2 -t /mnt/SecStorage/485ccf0d-3724-3575-a65e-318913593f8e/template/tmpl/2/211 -f /mnt/SecStorage/485ccf0d-3724-3575-a65e-318913593f8e/template/tmpl/2/211/dnld8772464283378614122tmp_ -u
```
It should be executed with result
```
template file /mnt/SecStorage/485ccf0d-3724-3575-a65e-318913593f8e/template/tmpl/2/211/dnld8772464283378614122tmp_ doesn't exist
```

svenvogel:
@ustcweizhou thanks. thats good. can you explain me whats the main task of the "DownloadManagerImpl.java" and why we create a template and after that we clean it up?

svenvogel:
@ustcweizhou why this script will be invoked? is this only for test? maybe you can help to clarify it for me.

ustcweizhou:
@svenvogel this script is used to generate template.properties in secondary storage vm when we register a template and it is downloaded successfully.

rhtyd:
@svenvogel looks like the refactorings add more regressions. Can you maybe instead start again with fixing the core issue and then submit refactoring changes (as a separate PR). Regression testing has failed few time on this PR already. Thanks.

rhtyd:
I've removed milestone from this PR, in case the work is resumed and can make it in time before 4.13.0.0 freeze kindly ping me. Thanks.

svenvogel:
@rhtyd can you kick the tests again. i think i am now ready :)

svenvogel:
@ustcweizhou @syed @skattoju4 as discussed with Syed and Sid. leave the "createtmplt.sh" as it is. there is only a small fix in this file. i changed only "cloud-install-sys-tmplt" file to get the correct virtual size for managed storages.  its an ugly situation that we call  "createtmplt.sh" in the java code and we should change this in the future to use complete java and remove the dependency of "createtmplt.sh".

can you check again?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test centos7 vmware-65

svenvogel:
@rhtyd thanks :). i test VMware too. it works. I think there is nothing changed for Xen or VMware. 👍

rhtyd:
Thanks @svenvogel, in that case I'll ask one of my colleagues for a review as I've not myself tested your changes /cc @PaulAngus @andrijapanicsb @borisstoyanov 

svenvogel:
@syed added comment for QCOW2/KVM only.

PaulAngus:
Hi @svenvogel - what does the cloud-install-sys-tmplt have to do with managed storage?  The script is there to get the system VM template into secondary storage when there isn't an SSVM up yet?

svenvogel:
Hi @PaulAngus thanks for question. i will try to explain it. if you use a managed storage like "netapp solidfire" you need a systemvm template with virtualsize parameter in the template.properties file and not with the physical size. in java code there is xen and vmware already implemented to create the virtualsize automatically. but not for kvm. this virtualsize is set in the "template.properties" file. if there is no correctly set in the template.properties file it cant copied from the SSVM to Solidfire System. the systemvm is always be copyied from the qcow2 systemvm template to the solidfire system if there is not present. if there is an update we will delete it and it will copied again. At the moment there was the only workaround to set the values in the database. this is not very useful and dangerous.

i added a flag "-M Managed Storage (QCOW2/KVM only)" for QCOW2/KVM with managed storage to get the virtualsize created in the template.properties file. every time if an "admin" will update the systemvm template with that script he can get the virtualsize with this parameter and dont set it manually. 

you can see that here in the video from @mike-tutkowski 

https://youtu.be/-tExEvkMJYs?list=PLqOXKM0Bt13DFnQnwUx8ZtJzoyDV0Uuye&t=550

@syed @mike-tutkowski is this the correct explanation?

mike-tutkowski:
That sounds correct to me, @svenvogel.

andrijapanicsb:
hm... I've personally used KVM with SolidFire, including hosting all system VMs on SF.

I read the explanation but still don't get it why it is now required @svenvogel

svenvogel:
> hm... I've personally used KVM with SolidFire, including hosting all system VMs on SF.
> 
> I read the explanation but still don't get it why it is now required @svenvogel

@andrijapanicsb we use Netapp Solidfire since end of 2018 and work with them in KVM. you will see from testing to production the timeframe is not so large. The term "now" is not entirely correct. its 6 month of tests and wip. the PR is open since June 2019 with some changes and to find out how it works. i dont think its good to edit the database for new install and/or change values all the time i need an upgrade. except an emergency, I do not think it makes sense to edit in the database. do you change always the database? we want to min the change of the database.

I think the video of @mike-tutkowski is a good explanation.

> i.e. what EXACTLY doesn't work atm in 4.11+?

i dont know before or after this release. i know that since we use the solidfire we have described problem above. the red rectangle with virtualsize is set without this fix to the disk disk size from the qcow2 and not virtualsize. if you upgrade the systemvm and change and the copy process will start it uses the disk size. SF system refuses the disk size. it needs an virtualsize and refuse the copy process.

with fix.
image

without fix. solidfire copy process will not work.
image

make that sense?

maybe @syed can add more historical information in this case.

andrijapanicsb:
qemuimgcmd=$(which qemu-img)
...
vrtmpltsize=$($qemuimgcmd info $tmpltfile | grep -i 'virtual size' |
...

This suggests that in order for it to work, you need to have qemu-img installed on the management server - do you plan to add some documentation on this?

@blueorangutan package

scripts/storage/secondary/createtmplt.sh
```diff
@@ -210,7 +210,6 @@ echo "filename=$tmpltname" > /$tmpltfs/template.properties
 echo "description=$descr" >> /$tmpltfs/template.properties
 # we need to rethink this property as it might get changed after download due to decompression
 # option is to recalcutate it here
-echo "checksum=$cksum" >> /$tmpltfs/template.properties
```
andrijapanicsb: Why is this line being deleted from the unrelated file?

scripts/storage/secondary/cloud-install-sys-tmplt
```diff
@@ -20,17 +19,33 @@
...
 }
 
+# Usage: e.g. failed $? "this is an error"
 failed() {
-  echo "Installation failed"
-  exit $1
+  local returnval=$1
+  local returnmsg=$2
+  
+  # check for an message, if there is no one dont print anything
```
andrijapanicsb: gramar...

andrijapanicsb:
@svenvogel @rhtyd @PaulAngus - I've just checked/tested the thing and the decision to store the VIRTUAL size (vs. the physical size that we do currently) in the "qcow2.virtualsize" and "virtualsize""  is the correct thing to do. This is exactly what is done when we download a template from i.e. internet into ACS - otherwise, trying to spin a system VM on SF will fail.

Now, a few points from my side:
- In ideal world, we should make qemu-img a mandatory/dependency binary while installing the "cloudstack-management" binaries and remove that -M flag in this PR, and **ALWAYS** store the VIRTUAL size in those 2 values, irrelevant of what storage is used.
- we are refactoring something that has been working ever since (some needless refactoring like putting ELSE in the same line, vs. a new line previously - I really don't see a need to do so)
- If you stick to the current PR, then you are relying on having qemu-img installed, so this needs documentation update (I see error messages raised etc) - or do a better solution as I suggested above and make it a dependency and remove the -M flag conditions completely.

scripts/storage/secondary/createtmplt.sh
svenvogel: in my first PR i tried to remove the "createtmplt.sh" file completely from creation but it is "ugly" in java code used. by this way i found an not used variable. i think checksum calculation was removed from Dan  Hoogland #2246. it was forgotten. i can make a seperate PR if this is better?

scripts/storage/secondary/cloud-install-sys-tmplt
svenvogel: i will change it like -> "check for a message, if there is none dont print anything" . better?

svenvogel:
@andrijapanicsb 

* for me its absolute okay to make it mandatory and remove the flag. if other storages dont have a problem with virtual size i would prefer your way. i think too its a good solution. if its necessary i will upgrade the documentation. 

should we do it?

svenvogel:
> In ideal world, we should make qemu-img a mandatory/dependency binary while installing the "cloudstack-management" binaries and remove that -M flag in this PR, and ALWAYS store the VIRTUAL size in those 2 values, irrelevant of what storage is used.

@andrijapanicsb @rhtyd should we do it? i remove the flag and we use always virtualsize? after that we can move forward.

svenvogel:
@syed can you take a look. i thinks its ready now.

svenvogel:
@ustcweizhou what do you think about the flag or remove it completely?

andrijapanicsb:
@svenvogel , if others agree @weizhouapache @rhtyd  - see this PR from Rohit on how package dependencies are added, and then:
- simply add depedency on qemu-img for both Debian/CentOS mgmt server packages
- remove any checks whether qemu-img is present or not (since it has to be present in order even install the mgmt server)
- remove the flag and unconditionally store the right value in the virtual size properties (only for KVM, as per the current implementation)

PaulAngus:
@svenvogel - My apologies, I've been swamped and this topic got buried...   What happens when the template is registered ie in preparation for an upgrade?  are the correct parameters added then?
We have a work in progress which you should have a look at  

https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=134745959

cc @Darrin.Husselmann@shapeblue.com

weizhouapache:
@svenvogel I totally agree with what @andrijapanicsb said in https://github.com/apache/cloudstack/pull/3371#issuecomment-559590637

svenvogel:
@weizhouapache @andrijapanicsb i will change it and then we should ready to go.

@PaulAngus no problem. yes i looked at the document. it looks like a new solution to upgrade the image. i speak about initial prepare and we mostly use command line script for upgrade the SystemVM

svenvogel:
@rhtyd can you kick the the tests for me?

@weizhouapache @andrijapanicsb can you take a look again?

weizhouapache:
@svenvogel I will test it. it looks ok.
by the way, you need to "git rebase" with latest master,or "Resolve conflicts" above.

svenvogel:
> @svenvogel I will test it. it looks ok.
> by the way, you need to "git rebase" with latest master,or "Resolve conflicts" above.

@weizhouapache thanks for checking. yes you are sure. now it should be rebased 👍

syed:
LGTM :+1:  Thanks @svenvogel

rhtyd:
I'll get back on this soon to review, let me kick tests in the meanwhile
@blueorangutan package

rhtyd:
@blueorangutan test

svenvogel:
4 LGTM, checks passed tested in different environments CentOS and Ubuntu.

andrijapanicsb:
Apologies @svenvogel , I have reverted this PR/merge - since tests have NOT complete (there is no output) - let me kick the tests again, and if all fine, you can merge it after that, please.

andrijapanicsb:
@blueorangutan test

Issue #3377, opened 2019-06-05
project tags test fails in all travis runs


DaanHoogland:
test_15_project_tag in component/test_tags fails consistently in travis jobv 17

##### ISSUE TYPE

 * Bug Report
 * Other - test failure

##### COMPONENT NAME

~~~

~~~

##### CLOUDSTACK VERSION

~~~

~~~

##### STEPS TO REPRODUCE

~~~
go to the travis test job page and then to job 17. restart it if needed.
~~~

##### EXPECTED RESULTS

~~~
job passes
~~~

##### ACTUAL RESULTS

~~~
the mentioned test lists tags which return NONE insted of the epected list of one.
~~~

ustcweizhou:
@DaanHoogland 
in test_tags.py, projectid needs to be added.

```
diff --git a/test/integration/component/test_tags.py b/test/integration/component/test_tags.py
index 9180e7b..be4c6e2 100644
--- a/test/integration/component/test_tags.py
+++ b/test/integration/component/test_tags.py
@@ -1756,6 +1756,7 @@ class TestResourceTags(cloudstackTestCase):
             listall=True,
             resourceType='project',
             resourceIds=project.id,
+            projectid=project.id,
             key=tag_key,
         )
         self.debug("tags = %s" % tags)
```

rhtyd:
Fixed in https://github.com/apache/cloudstack/pull/3476: 
```
==== Marvin Init Successful ====
=== TestName: test_15_project_tag | Status : SUCCESS ===
```

Closing as this is part of general stabilising effort on master, not an explicit bug.

PR #3323, opened 2019-05-09
User allowed to tag project created by him

Spaceman1984:
## Description

Fixed issue where a user cannot tag a project.

To reproduce:

in the UI:

log in as a user type
create a project
tag the new created project

Fixes: #2823
## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

Tested from UI.

Created a new project, and tagged the project, no errors.

Tested with cloudmonkey

Tested under Ubuntu server 18.04.2 LTS using the simulator

PR #3378, opened 2019-06-05
project id added in test

DaanHoogland:
as per @ustcweizhou's comment: https://github.com/apache/cloudstack/issues/3377#issuecomment-499056300

## Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -646,7 +646,10 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
 
         Ternary<Long, Boolean, ListProjectResourcesCriteria> domainIdRecursiveListProject = new Ternary<Long, Boolean, ListProjectResourcesCriteria>(cmd.getDomainId(), cmd.isRecursive(), null);
 
-        Long projectId = cmd.getProjectId() == null ? cmd.getResourceId() : cmd.getProjectId();
+        Long projectId = cmd.getProjectId();
+        if (resourceType.equalsIgnoreCase("project") && projectId == null) {
+            projectId = Long.parseLong(resourceId);
+        }
         _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), projectId, permittedAccounts, domainIdRecursiveListProject, listAll, false);
 
         Long domainId = domainIdRecursiveListProject.first();
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -648,7 +648,11 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
 
         Long projectId = cmd.getProjectId();
         if (resourceType.equalsIgnoreCase("project") && projectId == null) {
-            projectId = Long.parseLong(resourceId);
+            try {
+                projectId = Long.parseLong(resourceId);
+            } catch (NumberFormatException e) {
+                projectId = _projectDao.findByUuid(resourceId).getId();
+            }
         }
         _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), projectId, permittedAccounts, domainIdRecursiveListProject, listAll, false);
 
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -646,7 +646,9 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
 
         Ternary<Long, Boolean, ListProjectResourcesCriteria> domainIdRecursiveListProject = new Ternary<Long, Boolean, ListProjectResourcesCriteria>(cmd.getDomainId(), cmd.isRecursive(), null);
 
-        _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), cmd.getProjectId(), permittedAccounts, domainIdRecursiveListProject, listAll, false);
+        Long projectId = cmd.getProjectId() == null ? cmd.getResourceId() : cmd.getProjectId();
+        
+        _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), projectId, permittedAccounts, domainIdRecursiveListProject, listAll, false);
         Long domainId = domainIdRecursiveListProject.first();
         Boolean isRecursive = domainIdRecursiveListProject.second();
         ListProjectResourcesCriteria listProjectResourcesCriteria = domainIdRecursiveListProject.third();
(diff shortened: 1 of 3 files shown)


ustcweizhou:
I saved test_15_project_tag in a seperated python file, and run it in marvin test 
it failed without this change, and succeed with this change.

so this is ok for me. @DaanHoogland  @rhtyd

anuragaw (approved):

test/integration/component/test_tags.py
```diff
@@ -1756,6 +1756,7 @@ def test_15_project_tag(self):
             listall=True,
             resourceType='project',
             resourceIds=project.id,
+            projectid=project.id,
```
rhtyd: I would have expected this to work without specifying project ID. Do you think we need to fix mgmt server instead?

rhtyd: @DaanHoogland Read this https://github.com/apache/cloudstack/pull/3323/files, how about we add a check to read the resource type and use the provided resource ID as account ID when type is project.

ustcweizhou: @rhtyd that should be good.

DaanHoogland: yes, that would work, @rhtyd. I don't think it matters much, as the id is already in the API, but that solution would have less breakage of ecosystem scripting so let's go for it.

rhtyd: yes @DaanHoogland, I'm more concerned with existing systems and integrations that may break. And moreover, it breaks the API semantics (we're already passing projectid in the resource ID so why not use it).

DaanHoogland: on second thought @rhtyd , it won't work. There is no resource ID unless the project ID is passed as such. in the call there is no information that we can use
```
            listall=True,
            resourceType='project',
            key=tag_key,
```
We could get the current account from the call context but that would only satisfy some use-cases. if an admin wants tags for some project they are not in we still need the parameter.

DaanHoogland: never mind, i think I was stupid just there ^^. I'll revisit.

DaanHoogland:
I think I just overkilled it. didn't want to spend more time on it.

DaanHoogland:
@ustcweizhou I realize that now. I am confused by @rhtyd 's remark to use the resource ID it would not be the account ID of the owner of the project but the project itself. I don't think we can use that either. We could use it to find the account ID.

test/integration/component/test_tags.py
DaanHoogland: I've been mostly busy on other things today but gave it a second look. I think it can be fixed as you suggest in the query service. I will give it a swing soon. I'll remove the java change for now but leave the test in.

DaanHoogland:
note to self; need to do something like:
```
        Long projectId = cmd.getProjectId() == null ? cmd.getResourceId() : cmd.getProjectId();
        
        _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), projectId, permittedAccounts, domainIdRecursiveListProject, listAll, false);
```
 but translate ```cmd.getResourceId()``` to a Long id (from uuid?)

ustcweizhou:
> getProjectId

@DaanHoogland 
looked into the code, resourceId can be id or uuid, both are supported.

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -646,7 +646,9 @@
 
         Ternary<Long, Boolean, ListProjectResourcesCriteria> domainIdRecursiveListProject = new Ternary<Long, Boolean, ListProjectResourcesCriteria>(cmd.getDomainId(), cmd.isRecursive(), null);
 
-        _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), cmd.getProjectId(), permittedAccounts, domainIdRecursiveListProject, listAll, false);
+        Long projectId = cmd.getProjectId() == null ? cmd.getResourceId() : cmd.getProjectId();
```
ustcweizhou:
```
    public String getResourceId() {
        return resourceId;
    }
```

anuragaw (approved):
Previous approval was based on 1 line change. Need to review again after mutiple changes.

anuragaw (changes requested):
Previous approval was based on 1 line change. Need to review again after new changes.

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -646,7 +646,12 @@
 
         Ternary<Long, Boolean, ListProjectResourcesCriteria> domainIdRecursiveListProject = new Ternary<Long, Boolean, ListProjectResourcesCriteria>(cmd.getDomainId(), cmd.isRecursive(), null);
 
-        _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), cmd.getProjectId(), permittedAccounts, domainIdRecursiveListProject, listAll, false);
+        Long projectId = cmd.getProjectId();
+        if (resourceType.equalsIgnoreCase("project") && projectId == null) {
+            projectId = Long.parseLong(resourceId);
```
ustcweizhou: @DaanHoogland 
resourceId might be a uuid
it is better to have check like 
https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java#L681

DaanHoogland: yes @ustcweizhou but the search parameter is always Long, so we'll have to either refuse it if it is a uuid or add a different search. right now an exception is thrown.
I am thinkin g of abandoning this and just fix the test by setting the projectId in marvin.

DaanHoogland: implemented a fetch of the id based on assumed uuid

rhtyd:
@DaanHoogland failed following marvin test: (not sure if it's related to the PR)
```
test_15_project_tag     | marvin.cloudstackExcept | 1.249   | test_tags      |
|                         | ion.CloudstackAPIExcept |         |                |
|                         | ion                     |         |            
```

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -646,7 +646,16 @@
 
         Ternary<Long, Boolean, ListProjectResourcesCriteria> domainIdRecursiveListProject = new Ternary<Long, Boolean, ListProjectResourcesCriteria>(cmd.getDomainId(), cmd.isRecursive(), null);
 
-        _accountMgr.buildACLSearchParameters(caller, null, cmd.getAccountName(), cmd.getProjectId(), permittedAccounts, domainIdRecursiveListProject, listAll, false);
+        Long projectId = cmd.getProjectId();
+        if (resourceType.equalsIgnoreCase("project") && projectId == null) {
+            try {
+                projectId = Long.parseLong(resourceId);
+            } catch (NumberFormatException e) {
+                projectId = _projectDao.findByUuid(resourceId).getId();
```
ustcweizhou: What if the uuid is invalid ?

rhtyd:
Ping @DaanHoogland

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
DaanHoogland: it throws an exception: garbage in, garbage out. should we handle that? I think it is the exception @rhtyd is pointing at. How do you suggest this should be handled?

rhtyd:
I'm taking over the failing smoketests/Travis issues here: https://github.com/apache/cloudstack/pull/3476
Closing on this remark.

PR #3382, opened 2019-06-07
ui: fix instance and functionality

NicoWohlfarth:
## Description

* Fixed a bug, where after sorting a column of the instance list view the multi select action buttons wouldn't show up after selecting one ore more entries of the table.
* Added a behavior, so that an already starting/running/stopped/stopping instance of the instance view will (when multi selected alone or together with other instances) not create an API request if it already has the desired state.

## Types of changes

## How Has This Been Tested?

Deployed to local environment and tested functionality.

### DIFF
--- a/ui/scripts/instances.js
+++ b/ui/scripts/instances.js
@@ -79,39 +79,47 @@
         },
         action: function(args) {
           var instances = args.context.instances;
-          $(instances).map(function(index, instance) {
-            var data = {
-              id: instance.id
-            };
-            if (args.$form.find('.form-item[rel=hostId]').css("display") != "none" && args.data.hostId != -1) {
-              $.extend(data, {
-                hostid: args.data.hostId
+          var skippedInstances = 0;
+          $(instances).each(function(index, instance) {
+            if (instance.state === 'Running' || instance.state === "Starting") {
+              skippedInstances++;
+            } else {
+              var data = {
+                id: instance.id
+              };
+              if (args.$form.find('.form-item[rel=hostId]').css("display") != "none" && args.data.hostId != -1) {
+                $.extend(data, {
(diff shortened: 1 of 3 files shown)


svenvogel:
@rhtyd can you run checks please.

ping @DaanHoogland

svenvogel:
ui behaviour before the change: (bug)
1. nothing sorted or selected
Bildschirmfoto 2019-06-10 um 21 52 08
2. select any virtual machine (red - buttons come up)
Bildschirmfoto 2019-06-10 um 21 52 22
3. sort any column (blue) and select again any virtual machine (red), you dont see the buttons (red)
Bildschirmfoto 2019-06-10 um 21 52 41

svenvogel:
This is a UI only change, manual testing/screenshot is only needed. Smoketests not necessary.

svenvogel:
@shwstppr can you also take a look here?

shwstppr:
@NicoWohlfarth Is it possible to change the behavior of showing stop confirmation/host selection dialog when selected instance(s) are already in stopped/running state.
Although the start/stop VM is not called in this case but UI shows dialog unecessarily

Screenshot from 2019-06-12 11-19-34

Screenshot from 2019-06-12 11-19-39

Otherwise LGTM!

NicoWohlfarth:
> @NicoWohlfarth Is it possible to change the behavior of showing stop confirmation/host selection dialog when selected instance(s) are already in stopped/running state.
> Although the start/stop VM is not called in this case but UI shows dialog unecessarily

@shwstppr Yes, one could certainly do this, but i though it would then also be strange to have a button you can click and nothing would happen.
I think we should assume that no one would necessarily try to start/stop an already started/stopped machine and then be surprised that nothing happens. 😄

PR #3383, opened 2019-06-07
template: copy md5 mismatch

### LINKED ISSUE #3191: Copy template between zones fails with md5 mismatch
##### ISSUE TYPE

 * Bug Report
 
##### COMPONENT NAME

~~~
SSVM
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### CONFIGURATION

1 region, 2 zones, template registered/exists in zone1

##### OS / ENVIRONMENT

Issue confirmed on both KVM and VMware as Hypervisors

##### SUMMARY

When copying existing template from Zone1 to Zone2 (copyTemplate API call), md5 checksum between source and destination (copied/downloaded to Zone2) template fails

##### STEPS TO REPRODUCE

~~~
Register a template in Zone1, wait until Ready, thy to copy to Zone2 via GUI/API
~~~

##### EXPECTED RESULTS

~~~
Template copy works (did work in some of 4.5 or 4.6)
~~~

##### ACTUAL RESULTS

Copy operation is shown as success  (in GUI at least), but actually fails with errors (visible in GUI), from SSVM cloud.log "Failed post download script: checksum "{MD5}f916328a9273c511b36e7062a1b12262" didn't match the given value, "{MD5}a92d0b7882f9c3660bb6536206f1ee2a""

Further analysis, show that we are comparing destination md5 with a wrong source md5 - check the image https://pasteboard.co/I2eJYPb4.png or in other words:
In case of VMware env,  when we register the template with Zone1, there are 2 OVA files created on Secondary Storage in Zone1  i.e.
 f960f054-2be3-3431-b876-72b2814fdeb3.ova
and 
202-2-cac71084-dacc-35e8-9404-f72660bec1c3.ova
(both identical sizes)

Checking cloud.log in SSVM in Zone2 during copyTemplate API call, we can see we are copying over 202-2-cac71084-dacc-35e8-9404-f72660bec1c3.ova file to Zone2 – BUT we are then comparing md5sum (of downloaded file) to the wrong source file f960f054-2be3-3431-b876-72b2814fdeb3.ova.

I did not capture logs from KVM env, but as far as I recall, it failed with same error message.

shwstppr:
## Description
Fixes #3191 

When a template is registered, code stores md5sum of the downloaded file in the `vm_template` table. However, this downloaded file could be deleted after template installation if it is not an actual (`.qcow2`, `.ova`, etc.) file. When the user copies a template using copyTemplate API, the actual template file will be copied across the image stores. Matching checksum for the copied templated file and the stored value from the `vm_template` table will result in a mismatch.
Changes will set an empty checksum value for the copied template while passing to download service which allows skipping wrong checksum check for the copied while install.
However, this results in a change in checksum value for concerned template entry in `vm_template` table post template install.

## Types of changes

template-md5-mismatch-1

template-md5-mismatch-2

## How Has This Been Tested?

cmk

### DIFF
--- a/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
+++ b/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
@@ -1045,8 +1045,6 @@ public class TemplateServiceImpl implements TemplateService {
         }
         tmplForCopy.setUrl(url);
 
-        tmplForCopy.getImage().setChecksum("");
-
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Mark template_store_ref entry as Creating");
         }
--- a/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
+++ b/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
@@ -1045,6 +1045,8 @@ public class TemplateServiceImpl implements TemplateService {
         }
         tmplForCopy.setUrl(url);
 
+        tmplForCopy.getImage().setChecksum("");
+
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Mark template_store_ref entry as Creating");
         }
--- a/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
+++ b/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
@@ -1052,6 +1052,9 @@ public class TemplateServiceImpl implements TemplateService {
         DataObject templateOnStore = destStore.create(tmplForCopy);
         templateOnStore.processEvent(Event.CreateOnlyRequested);
 
+        // I am the big assumer^Wpretender
+        ((TemplateObject)templateOnStore).getImage().setChecksum(null);
+
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Invoke datastore driver createAsync to create template on destination store");
         }
--- a/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
+++ b/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
@@ -1052,8 +1052,9 @@ public class TemplateServiceImpl implements TemplateService {
         DataObject templateOnStore = destStore.create(tmplForCopy);
         templateOnStore.processEvent(Event.CreateOnlyRequested);
 
-        // I am the big assumer^Wpretender
-        ((TemplateObject)templateOnStore).getImage().setChecksum(null);
+        if (templateOnStore instanceof TemplateObject) {
+            ((TemplateObject)templateOnStore).getImage().setChecksum(null);
+        } // else we don't know what to do.
 
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Invoke datastore driver createAsync to create template on destination store");
--- a/server/src/main/java/com/cloud/storage/download/DownloadMonitorImpl.java
+++ b/server/src/main/java/com/cloud/storage/download/DownloadMonitorImpl.java
@@ -53,28 +53,20 @@ import com.cloud.configuration.Config;
 import com.cloud.storage.RegisterVolumePayload;
 import com.cloud.storage.Storage.ImageFormat;
 import com.cloud.storage.VMTemplateStorageResourceAssoc.Status;
-import com.cloud.storage.Volume;
-import com.cloud.storage.dao.VMTemplateDao;
-import com.cloud.storage.dao.VolumeDao;
 import com.cloud.storage.template.TemplateConstants;
 import com.cloud.storage.upload.UploadListener;
-import com.cloud.template.VirtualMachineTemplate;
 import com.cloud.utils.component.ComponentContext;
 import com.cloud.utils.component.ManagerBase;
 import com.cloud.utils.exception.CloudRuntimeException;
 
 @Component
 public class DownloadMonitorImpl extends ManagerBase implements DownloadMonitor {
-    static final Logger s_logger = Logger.getLogger(DownloadMonitorImpl.class);
+    static final Logger LOGGER = Logger.getLogger(DownloadMonitorImpl.class);
 
     @Inject
     private TemplateDataStoreDao _vmTemplateStoreDao;
     @Inject
-    private VolumeDao _volumeDao;
-    @Inject
     private VolumeDataStoreDao _volumeStoreDao;
-    @Inject
-    private final VMTemplateDao _templateDao = null;
     @Inject
     private AgentManager _agentMgr;
     @Inject
@@ -94,7 +86,7 @@ public class DownloadMonitorImpl extends ManagerBase implements DownloadMonitor 
 
         String cert = configs.get("secstorage.ssl.cert.domain");
         if (!"realhostip.com".equalsIgnoreCase(cert)) {
-            s_logger.warn("Only realhostip.com ssl cert is supported, ignoring self-signed and other certs");
+            LOGGER.warn("Only realhostip.com ssl cert is supported, ignoring self-signed and other certs");
         }
 
         _copyAuthPasswd = configs.get("secstorage.copy.password");
@@ -125,7 +117,7 @@ public class DownloadMonitorImpl extends ManagerBase implements DownloadMonitor 
 
     private void initiateTemplateDownload(DataObject template, AsyncCompletionCallback<DownloadAnswer> callback) {
         boolean downloadJobExists = false;
-        TemplateDataStoreVO vmTemplateStore = null;
+        TemplateDataStoreVO vmTemplateStore;
         DataStore store = template.getDataStore();
 
         vmTemplateStore = _vmTemplateStoreDao.findByStoreTemplate(store.getId(), template.getId());
@@ -141,7 +133,6 @@ public class DownloadMonitorImpl extends ManagerBase implements DownloadMonitor 
         Long maxTemplateSizeInBytes = getMaxTemplateSizeInBytes();
         if (vmTemplateStore != null) {
             start();
-            VirtualMachineTemplate tmpl = _templateDao.findById(template.getId());
             DownloadCommand dcmd = new DownloadCommand((TemplateObjectTO)(template.getTO()), maxTemplateSizeInBytes);
             dcmd.setProxy(getHttpProxy());
             if (downloadJobExists) {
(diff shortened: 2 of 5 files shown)


rhtyd:
Is this ready for testing @shwstppr or implementation in progress?

shwstppr:
@rhtyd I don't think the change/implementation is correct. I'm thinking of a better way. Will appreciate help here.
As mentioned in summary, when copying template code should not compare the checksum of the original download template file with the copied file. But if I make any changes here, https://github.com/apache/cloudstack/blob/master/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/template/DownloadManagerImpl.java#L383 it would affect normal template registration/download as well.

andrijapanicsb:
ping if any updates here.

shwstppr:
@andrijapanicsb will look into it next week

DaanHoogland:
@shwstppr there are two different checksums indeed.
- One for the file to be downloaded to compare against its origin
- One for checking internal copying
the original design has not taken this into account.

DaanHoogland:
@andrijapanicsb, I discussed with @shwstppr and this simple fix seems to work but does not take into account any special requirements on the design flaws as described above. any opinions? /cc @PaulAngus

andrijapanicsb:
As discussed (internally) this fix simply set the source image md5sum to NULL (instead of reading a wrong md5sum from DB, from the compressed file) and thus when SSVM receives the files in the new zone, since no source MD5 sum was defined, it will not check and will consider an image as downloaded successfully - which should be fine in 99% of cases.

I would like to test a silly thing - if the download of template (via the SSVM in the destination zone) is somehow interrupted - do we ever reach the MD5 "to check or not to check" part of code - ideally the download process should break and no false positive in sense of "successful" download.

This can be tested while copying a template from zone1 to zone2, by logging in to zone1 SSVM and simply stopping the Apache webserver - simulating connection drop - and checking/confirm that on the destination zone the template download has broken as exected.

@weizhouapache ping for opinion pls CC @rhtyd

rhtyd:
Is this ready or needs further update - @shwstppr @andrijapanicsb @PaulAngus @borisstoyanov ?

andrijapanicsb:
needs testing as I suggested, unless someone has other opinions as well

shwstppr:
I'm not sure if adding checksum parameter would add any functionality to the user or solve the problem here. As mentioned above copy and create template same code for downloading and checking checksum. For correctly checking checksum while copying a template code should compare the checksum of the copied file and the installed template file and not the checksum of copied file with the original checksum of the template.
Adding a checksum parameter, I'm not sure if user will ever input the checksum of template's installed file. And if user inputs original checksum of template copy will fail. Making it irrelevant IMO.
We should rather refactor underlying code (https://github.com/apache/cloudstack/blob/master/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java#L1026-L1075) to pass checksum string for the installed template file, adding a flag in DownloadCmd and DownloadAnswer classes to identify if we are copying so it doesn't change checksum value in DB later(https://github.com/apache/cloudstack/blob/master/engine/storage/src/main/java/org/apache/cloudstack/storage/image/BaseImageStoreDriverImpl.java#L247), after successful completion.
I'll try to do it over the weekend. Will update if I'm not able to.
cc @DaanHoogland

DaanHoogland:
I agree with @borisstoyanov that having a checking mechanism is useful functionality, but not that we are cutting that with this PR. It is not there and would be new functionality. The proposal is sensible, but the problem is that the user would not know the checksum to input, unless they are an admin with access to the source store and manually create it, @shwstppr. Making your solution an automation of the process that would happen in @borisstoyanov 's solution. If you can make it happen, please. Else, we are half way there with the extra parameter.

andrijapanicsb:
@shwstppr @borisstoyanov  /cc @DaanHoogland  where are we with this one?
i.e. we don't even expose MD5 option to users when downloading the template from the internet - sharing this just for the sake of the feature comparison

andrijapanicsb:
I'm also OK if we move forward withOUT MD5 sum now for 4.14, and create an additional issue for 4.15 to add MD5 checksum ?

DaanHoogland:
@andrijapanicsb It is a bug, i think we want it on 4.13

andrijapanicsb:
Agree @DaanHoogland - let's just enable copying the stuff between zones

DaanHoogland:
@shwstppr can you cherrypick to 4.14 and force push to your branch please?

DaanHoogland:
@shwstppr forgive, i mean cherry pick to 4.13 and force push, please

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
I agree with the objections that @borisstoyanov has but think it is a folowup. this fixes the issue described. I'll work on a follow up PR for the internal source to destination check. I don't think we must wait with this fix and risk releasing broken functionality.
cc @andrijapanicsb, @PaulAngus

andrijapanicsb:
+1, I'm OK to merge it as long as it's tested manually! (not sure smoke tests cover the template copy across zones)

DaanHoogland:
just tested this with favourable result:
image

we need more work here.

andrijapanicsb:
The fix is not working? Was it ever tested?

DaanHoogland:
@andrijapanicsb I have an implementation for this which is somewhat of a discussion point; the problem was/is that if you download an archived template (i.e. bz2 or so) the checksum you give will be that of the archive and not of the template therein. On download complete the checksum is checked (if given) and if not it is generated and added to the template.
I have changed the code so that the added checksum is the generated from the template and not from the archive, which makes a little more sense as that is what we are going to copy around later on. However in the old case, whether you gave a checksum or not, and in the new case only if you give a checksum, that will be entered in the db and upon registration completely useless. if such an archive-checksum exists any copy will fail. If it doesn't or if the template wasn't an archive to begin with, the cross copy zone will succeed.

what do you think,
cc @rhtyd @PaulAngus @borisstoyanov @othersoranybody

DaanHoogland:
@blueorangutan test matrix

andrijapanicsb:
@DaanHoogland not sure that one fixes anything for existing users - existing checksums in DB will be the one which we are reading and the problem will not be solved, right? (it does fix indeed new templates)

1. I liked the idea from @shwstppr as the temp solution - to ignore the MD5 sum given from the DB, since it's the archive md5sim, not the template itself (which somehow doesn't work, can we make it work?)

2. I also like the idea that we do ad-hoc md5sum calculation on the template BEFORE we start copying it to another zone (I.e. don't read from DB, since it useless in the current implementation) - but do it ad-hoc (which does have it's issues of stressing CPVM CPU more frequently than needed)

Either way, this needs more serious change (4.15?) to implement what you have already done, but to also forbid users to enter md5sum of the archive - well, not forbid, perhaps that is used for the validation of the download... but then when extracted, we can overwrite that in DB with the md5sum of the extracted file (how about multi-disk-ova???).

Another approach is to copy (to another zone) the original archive itself (assuming we don't wipe it after the extraction), and do no more changes. Here, keep in mind for VMware, I saw 2 x OVA files (the original one and the generated one), both archives by definition, and the new OVA (generated) is being copied over and unpacked again (well this doesnt work, bug...).
For KVM not sure what is being copied over.

I can rant until tomorrow - but I believe simply always keeping the archive/original file and have that one's md5sum in DB AND COPYING that one always (and unpacking as needed) is the way to go. Forget ever copying the extracted file (why would we do that - more space, slower, etc?)

DaanHoogland:
@andrijapanicsb my "hack in the chksm" seems to have done it. source checksum will be ignored for cross  zone cp. Not a final sturdy design but it deals with old urls and archive checksums.

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@borisstoyanov @andrijapanicsb @shwstppr @rhtyd @PaulAngus ,

I managed to avoid the rabbit-hole kind of. We are (for now) not giving a checksum when copying from zone to zone. The reason for not creating one before starting the copy is that the copy is done from a constructed URL and the original file is never in play. In the future I think we want a command-answer implementation for getChecksumForFile.

I have done some refactoring and implemented sha-512 instead of md5 along the way. I am testing the download on a added zone for a all-zones template and copy a zone specific template to the new zone as well. I'll be firing regressions as well, but none of those will touch this change.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan test centos7 vmware65

DaanHoogland:

DaanHoogland:

andrijapanicsb:
what's the penalty with SHA-512, on 500Mhz SSVM? Otherwise sounds fine to me.

andrijapanicsb:
And what about backward "compatibility".
If adding i.e. a new storage, SSVNM will try to download template from the original URL - does md5 sum gets checked and if so, will it fail for older templates created using md5sum if you use sha512?
i. e. will it break anything?

DaanHoogland:
no, in theory ;) if the checksum stored is an md5 a check will be done against that.

DaanHoogland:
`The job exceeded the maximum log length, and has been terminated.` not related to the code. we might want to lower the travis log levels? merging

DaanHoogland:
ping @shwstppr @rhtyd @andrijapanicsb @nvazquez @PaulAngus , alright like this? I need a second ok, after @borisstoyanov 's testing

engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
```diff
@@ -1052,6 +1052,9 @@ public void syncTemplateToRegionStore(long templateId, DataStore store) {
         DataObject templateOnStore = destStore.create(tmplForCopy);
         templateOnStore.processEvent(Event.CreateOnlyRequested);
 
+        // I am the big assumer^Wpretender
```
rhtyd: @shwstppr use instanceof to check; avoid exceptions

DaanHoogland: Ok, i checked the code path and it will be an `TemplateObject` always, but you are right.

DaanHoogland: I changed the code but the net effect stays the same. If it isn't a `TemplateObject` an exception is thrown and should be thrown. we'll be in the wrong spot in that case.
@rhtyd ?

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan package

borisstoyanov:
@blueorangutan test

DaanHoogland:
@rhtyd good to go?

rhtyd:
LGTM, checking travis. Will merge if it's ok.

Issue #3385, opened 2019-06-07
UI Translate code removes same-level content


richardlawley:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

The UI code replaces `<translate key="blah"/>` with the localised content.  However, the translate code incorrectly removes content at the same level, for example the Zone Names on the capacity list of the dashboard:

```
<div class="label">
    <translate key="label.zone"/>: <span data-list-item="zoneName"></span>
</div>
```

This results in the dashboard not showing the Zone name.  This also affects the "Optional" label in VM creation from a VPC on the Network step.

This is because the `$.find('translate[key]')` code is not treating translate as a valid self-closing tag.  This problem doesn't affect cases where there is no non-tag content at the same level, which is why it hasn't been picked up yet.

Other known cases:
* Instance Wizard - VPC Select Network tier - optional label not shown
* Instance Wizard - Step 8 (Review) - Name and Group do not show Optional Label

##### STEPS TO REPRODUCE

View the Dashboard
Observe that the Capacities do not list zone names

##### EXPECTED RESULTS

Zone Names should be displayed in Capacity list

##### ACTUAL RESULTS

![image](https://user-images.githubusercontent.com/1608083/59103002-8f26f100-8925-11e9-87d3-9a4344ca7523.png)

anuragaw:
This should be fixed on latest master @richardlawley https://github.com/apache/cloudstack/pull/3326

CC/ @rhtyd , @rafaelweingartner , @PaulAngus

richardlawley:
@anuragaw #3326 only fixes the Dashboard instance - there were about 10 occurrences of the same problem which I fixed in #3386.  However, the method you used to fix (wrapping in a label) is better than mine.  Will try and alter mine to fix the other occurrences.

PR #3326, opened 2019-05-10
Bug fix for zone names not appearing in dashboard

anuragaw:
3 years ago when we moved from JSP this translate was not wrapped in label tags and consequently the following span wasn't captured when dashboard.js mapped the data.

Fixes: #3314

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## Screenshots (if appropriate):
Screenshot from 2019-05-10 17-20-06

## How Has This Been Tested?

Through UI

PR #3386, opened 2019-06-07
Fix labels broken by translation code

richardlawley:
## Description

As discussed in #3385, the translation code does not work with self closing translate tags when there is other content at the same level, e.g. `<div><translate key="foo"/>: Something</div>` - it will remove the "`: Something`".  This PR changes the affected tags to non-self closing, e.g. `<translate key="foo"></translate>: Something`

This fixes the Zone names not being displayed on the capacity graphs on the dashboard, plus some minor missing labels elsewhere.

Fixes: #3385 

## Types of changes

Before:
image

After:
image

## How Has This Been Tested?

Verified that missing labels appear in UI.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -311,9 +311,9 @@
                                 <p id="from_vpc_tier">
                                     <div class="specify-ip">
                                         <label>
-                                            <translate key="label.ip.address"/>
-                                            (<translate key="label.optional"/>):
+                                            <translate key="label.ip.address" />
                                         </label>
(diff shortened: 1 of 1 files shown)


richardlawley:
It would be better to fix the translate code so it doesn't affect this, but I don't have time at the moment.

richardlawley:
Updated this to remove the fix to the dashboard already fixed in #3326, and changed to keep translate tag self-closing, just wrapped in <label>.

New example of what's fixed - (Optional) labels were affected and don't show without the fix.
image

PR #3389, opened 2019-06-08
when destroy the vms, delete the tags from virtual router

matheusdaluz:
## Description
When tags are placed in the vms and after we delete them, they remain in the virtual router generating problems for other vms that gain the same ip in the virtual router.

## Types of changes

## How Has This Been Tested?
I put the tags in the vms and after that I deleted them and accessed the virtual router to verify that they were deleted correctly.

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -39,6 +39,8 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
+import com.cloud.server.ResourceTag;
+import com.cloud.server.TaggedResourceService;
 import org.apache.cloudstack.acl.ControlledEntity.ACLType;
 import org.apache.cloudstack.acl.SecurityChecker.AccessType;
 import org.apache.cloudstack.affinity.AffinityGroupService;
@@ -472,6 +474,8 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
     private TemplateApiService _tmplService;
     @Inject
     private ConfigurationDao _configDao;
+    @Inject
+    private TaggedResourceService _taggedResourceService;
 
     private ScheduledExecutorService _executor = null;
     private ScheduledExecutorService _vmIpFetchExecutor = null;
@@ -2818,6 +2822,16 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         return destroyedVm;
     }
 
+    protected void removeTagsFromVm(long vmId) {
+        UserVmVO vm = _vmDao.findById(vmId);
+        Long resourceId = _taggedResourceService.getResourceId(vm.getUuid(), ResourceTag.ResourceObjectType.UserVm);
+        List<? extends ResourceTag> resourceTags = _taggedResourceService.listByResourceTypeAndId(ResourceTag.ResourceObjectType.UserVm, resourceId);
+        if (resourceTags.size() > 0) {
+            List<String> resourceIds = Arrays.asList(vm.getUuid());
+            _taggedResourceService.deleteTags(resourceIds, ResourceTag.ResourceObjectType.UserVm, null);
+        }
+    }
+
     private List<VolumeVO> getVolumesFromIds(DestroyVMCmd cmd) {
         List<VolumeVO> volumes = new ArrayList<>();
         if (cmd.getVolumeIds() != null) {
@@ -4606,6 +4620,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         try {
             VirtualMachineEntity vmEntity = _orchSrvc.getVirtualMachine(vm.getUuid());
             status = vmEntity.destroy(Long.toString(userId), expunge);
+            removeTagsFromVm(vmId);
         } catch (CloudException e) {
             CloudRuntimeException ex = new CloudRuntimeException("Unable to destroy with specified vmId", e);
             ex.addProxyObject(vm.getUuid(), "vmId");
--- a/server/src/test/java/com/cloud/vm/UserVmManagerImplTest.java
+++ b/server/src/test/java/com/cloud/vm/UserVmManagerImplTest.java
@@ -17,8 +17,14 @@
 package com.cloud.vm;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.List;
 
+import com.cloud.server.ResourceTag;
+import com.cloud.tags.ResourceTagVO;
+import com.cloud.tags.TaggedResourceManagerImpl;
+import com.cloud.tags.dao.ResourceTagDao;
 import org.apache.cloudstack.api.BaseCmd.HTTPMethod;
 import org.apache.cloudstack.api.command.user.vm.UpdateVMCmd;
 import org.apache.cloudstack.context.CallContext;
@@ -26,6 +32,7 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+
 import org.mockito.BDDMockito;
 import org.mockito.InjectMocks;
 import org.mockito.Mock;
@@ -75,6 +82,13 @@ public class UserVmManagerImplTest {
 
     @Mock
     private NetworkModel networkModel;
+
+    @Mock
+    private TaggedResourceManagerImpl taggedResourceManagerMock;
+
+    @Mock
+    private ResourceTagDao resourceTagDaoMock;
+
 
     private long vmId = 1l;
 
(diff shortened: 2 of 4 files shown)


rcicm:
+1

ustcweizhou:
@matheusdaluz 
Could you tell us how to verify it ?
When I add/delete a tag to a vm, I did not see any change in the virtual router.

matheusdaluz:
You should add a tag and after that destroy the vm.

svenvogel:
@matheusdaluz is it for VPC or virtual router?

ustcweizhou:
> @ustcweizhou
> First you ought to create a network and after that attach the vm, next step is create tags to vm, after you destroy the vm and you'll see the tags in virtual router yet.

@matheusdaluz thanks. could you tell me how to see the tags in the virtual router ?

matheusdaluz:
@svenvogel virtual router.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4601,17 +4627,23 @@ public UserVm destroyVm(long vmId, boolean expunge) throws ResourceUnavailableEx
...
             VirtualMachineEntity vmEntity = _orchSrvc.getVirtualMachine(vm.getUuid());
             status = vmEntity.destroy(Long.toString(userId), expunge);
+            resultRemovedTags = removeTagsFromVm(vmId);
         } catch (CloudException e) {
             CloudRuntimeException ex = new CloudRuntimeException("Unable to destroy with specified vmId", e);
             ex.addProxyObject(vm.getUuid(), "vmId");
             throw ex;
         }
 
+        if (!resultRemovedTags) {
```
rafaelweingartner: If there are no tags, you will log an error message that is misleading.

matheusdaluz: should be a warn ?

rafaelweingartner: I do agree with the error message if there is a problem when deleting them. However, the problem here is the design used to "identify" when there is an error. If there are no tags, you return "false" in that delete tags method. Personally, we are using Java; and therefore, we should avoid these designs (commonly used in C) where we return true/false to indicate if the method was executed properly.

matheusdaluz:
> @matheusdaluz thanks. could you tell me how to see the tags in the virtual router ?

in the folder /var/www/html/metadata and ip from network.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
matheusdaluz: ok, i'll fix this.

matheusdaluz: i fixed

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -2818,6 +2822,26 @@ public UserVm destroyVm(DestroyVMCmd cmd) throws ResourceUnavailableException, C
...
 
+    protected void removeTagsFromVm(long vmId) {
+        UserVmVO vm = _vmDao.findById(vmId);
+        Long resourceId = _taggedResourceService.getResourceId(vm.getUuid(), ResourceTag.ResourceObjectType.UserVm);
+        List<? extends ResourceTag> resourceTags = _taggedResourceService.listByResourceTypeAndId(ResourceTag.ResourceObjectType.UserVm, resourceId);
+        if (resourceTags.size() > 0) {
+            List<String> resourceIds = Arrays.asList(vm.getUuid());
+            Map<String, String> tags = new HashMap<>();
+            _taggedResourceService.deleteTags(resourceIds, ResourceTag.ResourceObjectType.UserVm, tags);
+            s_logger.debug("Tags from vm id: " + vmId + " has been removed.");
```
rafaelweingartner: what about adding the deleted tags to the debug message as well?

rafaelweingartner: Also, are you sending an empty list of tags for the deleteTags method?

matheusdaluz: Yes, if someone would like debug the method and i need send a void map to method work.

rafaelweingartner: That last parameter can be null

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -2818,6 +2822,26 @@ public UserVm destroyVm(DestroyVMCmd cmd) throws ResourceUnavailableException, C
...
+        }
+    }
+
+    public void setTaggedResourceService(TaggedResourceService taggedResourceServiceMock) {
+        _taggedResourceService = taggedResourceServiceMock;
+    }
+
+    public void setVmDao(UserVmDao userVmMock) {
```
rafaelweingartner: why did you create this set?

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -2818,6 +2822,26 @@ public UserVm destroyVm(DestroyVMCmd cmd) throws ResourceUnavailableException, C
...
+        List<? extends ResourceTag> resourceTags = _taggedResourceService.listByResourceTypeAndId(ResourceTag.ResourceObjectType.UserVm, resourceId);
+        if (resourceTags.size() > 0) {
+            List<String> resourceIds = Arrays.asList(vm.getUuid());
+            Map<String, String> tags = new HashMap<>();
+            _taggedResourceService.deleteTags(resourceIds, ResourceTag.ResourceObjectType.UserVm, tags);
+            s_logger.debug("Tags from vm id: " + vmId + " has been removed.");
+        }
+    }
+
+    public void setTaggedResourceService(TaggedResourceService taggedResourceServiceMock) {
```
rafaelweingartner: And this one?

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
rafaelweingartner: Can you add the deleted tags to the message?

matheusdaluz: My bad, i'll change.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
matheusdaluz: to mock tests

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
matheusdaluz: to mock tests

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
matheusdaluz: i fixed.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
matheusdaluz: fixed.

server/src/test/java/com/cloud/vm/UserVmManagerImplTest.java
```diff
@@ -276,4 +285,36 @@ private void configureValidateOrReplaceMacAddressTest(int times, String macAddre
         Mockito.verify(networkModel, Mockito.times(times)).getNextAvailableMacAddressInNetwork(Mockito.anyLong());
         Assert.assertEquals(expectedMacAddress, returnedMacAddress);
     }
+
+    @Test
+    public void validateRemoveTagsWhenExists() {
```
rafaelweingartner: You have test cases without any assert. How do you check if the method is doing what you want it to do?

matheusdaluz: unfortunately is not easy to test.

ustcweizhou:
> > @matheusdaluz thanks. could you tell me how to see the tags in the virtual router ?
> 
> @ustcweizhou in the folder /var/www/html/metadata and ip from network.

thanks @matheusdaluz 
Which cloudstack version are you using ? I did not find information related to resource tags in the folders you mentioned.
It would be nice if you can give some examples
(1) file/content before destroying a vm
(2) changes after destroying a vm

Anyway, the changes in server/src/main/java/com/cloud/vm/UserVmManagerImpl.java looks good to me.

matheusdaluz:
> > > @matheusdaluz thanks. could you tell me how to see the tags in the virtual router ?
> > 
> > 
> > @ustcweizhou in the folder /var/www/html/metadata and ip from network.
> 
> thanks @matheusdaluz
> Which cloudstack version are you using ? I did not find information related to resource tags in the folders you mentioned.
> It would be nice if you can give some examples
> (1) file/content before destroying a vm
> (2) changes after destroying a vm
> 
> Anyway, the changes in server/src/main/java/com/cloud/vm/UserVmManagerImpl.java looks good to me.

@ustcweizhou some examples
1- Vm with tags in virtual router and before i destroying
image

2 - Vm destroy and tags gone.
Screen Shot 2019-06-11 at 11 46 28

ustcweizhou:
@matheusdaluz thanks.
are you using official apache cloudstack ?

I am also running with 4.11.2. However, I do not see tags in the folders in /var/www/html, weird
```
root@r-6-VM:~# ls -l /var/www/html/metadata/192.168.10.88/
total 11
-rw-r--r-- 1 www-data www-data   9 Jun 11 18:14 availability-zone
-rw-r--r-- 1 www-data www-data  49 Jun 11 18:14 cloud-identifier
-rw-r--r-- 1 www-data www-data  36 Jun 11 18:14 instance-id
-rw-r--r-- 1 www-data www-data  19 Jun 11 18:14 local-hostname
-rw-r--r-- 1 www-data www-data  13 Jun 11 18:14 local-ipv4
-rw-r--r-- 1 www-data www-data 136 May 14 09:35 meta-data
-rw-r--r-- 1 www-data www-data  13 Jun 11 18:14 public-hostname
-rw-r--r-- 1 www-data www-data  13 Jun 11 18:14 public-ipv4
-rw-r--r-- 1 www-data www-data 380 Jun 11 18:14 public-keys
-rw-r--r-- 1 www-data www-data  14 Jun 11 18:14 service-offering
-rw-r--r-- 1 www-data www-data  36 Jun 11 18:14 vm-id
```

@rhtyd are you able to reproduce the issue @matheusdaluz mentioned ?

matheusdaluz:
@ustcweizhou yes, you added tags to vm ?

server/src/test/java/com/cloud/vm/UserVmManagerImplTest.java
matheusdaluz: fixed

ustcweizhou:
> @ustcweizhou yes, you added tags to vm ?

@matheusdaluz yes. I added tags, no change in virtual routers. then stop/start the vm, still did not see any change. 
I also read the code and did not find any code to add the resource tags to metadata.  strange. https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/network/router/CommandSetupHelper.java#L1008

matheusdaluz:
@ustcweizhou sorry, i use a modified cloudstack that save the tags in virtual router, it is a good feature to acs or not ?

ustcweizhou:
@matheusdaluz 
it might be useful for some other users.
so if you want to merge it into master, it would be better to add a global setting to indicate whether the vm tags will be saved in virtual routers.

Issue #3379, opened 2019-06-05
SRX Plugin Cannot add Port Forwarding Rule


richardlawley:
<!--
Verify first that your issue/request is not already reported on GitHub.
Also test if the latest release and master branch are affected too.
Always add information AFTER of these HTML comments, but no need to delete the comments.
-->

##### ISSUE TYPE
<!-- Pick one below and delete the rest -->
 * Bug Report

##### COMPONENT NAME
<!--
Categorize the issue, e.g. API, VR, VPN, UI, etc.
-->
~~~
SRX Plugin
~~~

##### CLOUDSTACK VERSION
<!--
New line separated list of affected versions, commit ID for issues on master branch.
-->

~~~
4.11.2
~~~

##### CONFIGURATION
<!--
Information about the configuration if relevant, e.g. basic network, advanced networking, etc.  N/A otherwise
-->
Advanced Networking

##### OS / ENVIRONMENT
<!--
Information about the environment if relevant, N/A otherwise
-->
Model: srx345-dual-ac
Junos: 15.1X49-D110.4
JUNOS Software Release [15.1X49-D110.4]

##### SUMMARY
<!-- Explain the problem/feature briefly -->
When trying to add a port forwarding rule, it fails with a syntax error.  If this ever worked, then I suspect the syntax of the API command has changed since JunOS 10, which from the hardware compatibility list may be what worked.

##### STEPS TO REPRODUCE
<!--
For bugs, show exactly how to reproduce the problem, using a minimal test-case. Use Screenshots if accurate.

For new features, show how the feature would be used.
-->
Add a port forwarding rule
<!-- Paste example playbooks or commands between quotes below -->

<!-- You can also paste gist.github.com links for larger files -->

##### EXPECTED RESULTS
<!-- What did you expect to happen when running the steps above? -->
Port Forwarding rule should be added, creating a destination nat rule on the SRX


##### ACTUAL RESULTS
<!-- What actually happened? -->

<!-- Paste verbatim command output between quotes below -->
~~~
2019-06-05 16:44:44,452 DEBUG [c.c.n.r.JuniperSrxResource] (DirectAgent-287:ctx-8a4e6665) (logid:be9407ed) Sending request:
<rpc>                                                                                                                                                                                                                                
    <load-configuration>                                                                                                                                                                                                             
    <configuration>                                                                                                                                                                                                                  
        <security>                                                                                                                                                                                                                   
            <nat>                                                                                                                                                                                                                    
[...]

PR #3393, opened 2019-06-10
Fix removing SRX port forwarding rules, improve add/remove logic

richardlawley:
## Description

This PR partially fixes the logic around port forwarding rules on the Juniper SRX plugin.  The code in the plugin is based on JunOS 10, which is very old.  The changes here should not break compatibility, but should enable the plugin to be used on newer devices.  Note that an additional change to a script file is required to be able to add port forwarding rules, but as this PR was targetted for 4.11.3, I thought it best not to include this change as it might break compatibility for anyone still using JunOS 10.

I've made the logic better and consistent for adding/removing static nat and port forwarding rules - these were multi-step processes which did not check each individual step.  This would aid in manually fixing rules in case of further problems.

I've also improved the logging for communication with the SRX by stripping out the Apache header before sending it, and indicating the name of the template filename in use.

To be able to add port forwarding rules, the `<dst-port>` tags in dest-nat-rule-add.xml must be changed to `<low>`.

Fixes: #3379

## Types of changes

## How Has This Been Tested?

Test environment consisting of simulator CloudStack connected to an SRX345 device.  Test process:

* Add Static NAT Rule
* Confirm rule was added (show | compare rollback 1)
* Remove Static NAT Rule
* Confirm rule was removed (show | compare rollback 1)
* Confirm that Add/Remove left config as before (show | compare rollback2 = empty)
* Add Port Forwarding Rule
* Confirm rule was added (show | compare rollback 1)
* Remove Port Forwarding Rule
* Confirm rule was removed (show | compare rollback 1)
* Confirm that Add/Remove left config as before (show | compare rollback2 = empty)

### DIFF
--- a/plugins/network-elements/juniper-srx/src/main/java/com/cloud/network/resource/JuniperSrxResource.java
+++ b/plugins/network-elements/juniper-srx/src/main/java/com/cloud/network/resource/JuniperSrxResource.java
@@ -177,7 +177,15 @@ public class JuniperSrxResource implements ServerResource {
         private static final Logger s_logger = Logger.getLogger(JuniperSrxResource.class);
 
         private SrxXml(String filename) {
-            xml = getXml(filename);
+            String contents = getXml(filename);
+
+            // Strip the apache header and add the filename as a header to aid debugging
+            contents = contents.replaceAll( "(?s)<!--.*?-->", "" ).trim();
+            if (!contents.startsWith("<?")) {
+                contents = "<!-- " + filename + " -->" + contents;
+            }
+
+            xml = contents;
         }
 
         public String getXml() {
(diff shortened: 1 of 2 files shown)


rhtyd:
Our smoketests don't run the srx at all, so no smoketests against vmware/kvm/xenserver are needed. We can merge based on Travis/simulator smoke tests and general packaging.
@blueorangutan package

rhtyd:
@richardlawley can you rebase against latest master?

To the wider community - do we know any srx users/devs who can review this?

richardlawley:
I've rebased this, and started a conversation on the dev list about bumping up the supported version of SRX to 15, which would mean I'd finish off these changes (which were originally hoped to go into 4.11.3) so that no changes to files are required for users.

rhtyd:
Okay @richardlawley let me know when you're done with your changes. I'll help kick regression tests.

rhtyd:
On hold until @richardlawley confirm changes and testing.

richardlawley:
@rhtyd This is ready now.  I don't the travis failures can be related to this.

This change will likely prevent CS from being able to add Port Forwarding rules on the originally supported 10.4 version due to the change in syntax for destination nat rules.  It has been tested on an SRX345 running JUNOS 15.1X49-D170.4.  I cannot find documentation for which version of JunOS introduced the changes - related changes came in 11.4 to static NAT.

I had originally intended to also improve the handling of port forwarding rules where port ranges are specified - currently this uses 2 rules + 1 dest nat pool for every port, which might become a problem if a large range is specified.  However, documentation on the port ranges is very unclear from Juniper, and this would also significantly complicate the PR.  I may revisit it in future if we find it problematic.

rhtyd:
Thanks @richardlawley I'll kick some tests.
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Closed reopened PR to rekick Travis.
@richardlawley can you provide details or summary of tests you performed against srx?

richardlawley:
Essentially adding and removing port forwarding and static NAT rules, then running `show | compare rollback 2` to verify that config was restored to same state as before the add.

Before changes, add pf rule would fail (without change the XML file) and remove would appear to complete but the rule would not actually have been removed.

rhtyd:
@borisstoyanov @anuragaw @shwstppr can you review based on regression test results?

PR #3403, opened 2019-06-13
ui: Fix quick view tooltip title on multiselect list views

NicoWohlfarth:
## Description

Fixed an issue, where the quick view tooltip title would not have been displayed properly on list views with a multiselect row.
This was because the html content of the first `td` of a row was used to render the title of the quick view tooltip, which would not work if the first columns content was a checkbox.
Now the `td` with class `first` will be used for this (this will be the second column if there is a multiselect column, otherwise it will remain the first column).

## Types of changes

## Screenshots:
Before
before
After
after

## How Has This Been Tested?
Deployed to local environment and tested functionality.

### DIFF
--- a/ui/scripts/ui/widgets/listView.js
+++ b/ui/scripts/ui/widgets/listView.js
@@ -1585,10 +1585,10 @@
                             $('<span>').html(_l('label.quickview') + ': '),
                             $('<span>').addClass('title').html(
                                 cloudStack.concat(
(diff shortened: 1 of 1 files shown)


svenvogel:
@shwstppr @anuragaw can you take a look? :)

anuragaw:
Tested fine locally too

> Based on PR Description and reading style in CSS LGTM.
> Will do a quick test locally and get back if I find any issues.

PR #3405, opened 2019-06-13
kvm: fix qemu hook race condition

rhtyd:
This fixes the qemu hooks `mkdir` race condition which can happen when
too many VMs may launch on a KVM host executing the hooks script that
tries to `mkdir` for the custom directory. On exception (multiple scripts
trying to mkdir), the VM stops. When libvirtd restarts it triggers hooks, and
the issue can be reproduced, which can cause loss of running VMs. 

The custom directory need not be created if it does not exist, instead
the custom hooks should only execute when there is a custom directory.
Admins who want to use the custom scripts feature can create the directory
as per the feature documentation. 

Feature documentation:
http://docs.cloudstack.apache.org/en/4.11.2.0/adminguide/hosts.html#kvm-libvirt-hook-script-include

## Types of changes

### DIFF
--- a/agent/bindir/libvirtqemuhook.in
+++ b/agent/bindir/libvirtqemuhook.in
@@ -78,7 +78,9 @@ def handleMigrateBegin():
 
 
 def executeCustomScripts(sysArgs):
-    createDirectoryIfNotExists(customDir, customDirPermissions)
+    if not os.path.exists(customDir) or not os.path.isdir(customDir):
+        return
+
     scripts = getCustomScriptsFromDirectory()
 
     for scriptName in scripts:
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@DaanHoogland we currently have no way to detect and tell the user/admin that VM died due to the hooks race condition.
@blueorangutan package

DaanHoogland:
@rhtyd my question is/was: Do we need to send an alarm if the directory doesn't exist?

rhtyd:
@blueorangutan test

ustcweizhou:
code LGTM

PR #3412, opened 2019-06-20
Allow for the VM Hostname to be edited  when VM is switched off 

kioie:
## Description

This is a feature that allows end users to be able to change the hostname of their VM and the, if the vm.instancename.flag under the Global Settings is set to true

## Types of changes

Screen Shot 2019-06-28 at 3 37 13 PM
Screen Shot 2019-06-28 at 3 37 45 PM
Screen Shot 2019-06-28 at 3 38 04 PM

## How Has This Been Tested?

 Changed the vm.instancename.flag under the Global Settings to false to see if the VM hostname would still change

 Changed the  vm.instancename.flag under the Global Settings to true to see if the VM hostname would change.


kioie:
@blueorangutan package

svenvogel:
@kioie is it a console symbol/icon?

kioie:
> @kioie is it a console symbol/icon?

Yes it is

svenvogel:
1. why not here?

image

2. does your update VM Hostname change the Display Name too?

svenvogel:
3. maybe there is a another icon which is more fitting.

kioie:
> 1. why not here?
> 
> image
> 
> i think its more usable to put it there because for detailed settings we have this icon. normally the vm hostname will not be changed very often in infrastructure and two two more clicks should be okay. we should not overload the menu.
> 
My intention was to make it as easily accessible as possible (meaning you can switch off a vm, change the hostname, restart the VM, on the same page), but I can follow your suggestion and shift it if thats you're recommendation
> 1. does your update change the vm hostname and the display name too?

Doesn't change the display name, only the hostname

> 2. maybe there is a another icon which is more fitting.

Went through the icon list, couldn't find anything, but maybe you can recommend something and I can switch it up asap. Also, with the current icon, I think maybe it might bring some bit of confusion if I move it to the above page (someone might think its an icon to allow access to the console when the VM is powered off)

anuragaw:
Agree with @svenvogel 's suggestion that the edit menu should have host name change facility instead of quick view. The latter makes more sense to have frequently used options in the limited space it provides.

kioie:
> Agree with @svenvogel 's suggestion that the edit menu should have host name change facility instead of quick view. The latter makes more sense to have frequently used options in the limited space it provides. Also in that case we won't need new icon.

Changed as suggested

rhtyd:
LGTM but did not test it.
@anuragaw @svenvogel have you tested the latest change?

rhtyd:
Ping @anuragaw @svenvogel are you lgtm on this?

anuragaw:
Code LGTM but haven't tested yet. Will get back in sometime with a quick test.

rhtyd:
Based on the code it should work, did a quick one. Will merge now.

rhtyd:
Based on the code it should work, did a quick one. Will merge now.

PR #3414, opened 2019-06-21
Increase z-index for install-wizard step

pbankonier:
Increase z-index for install wizard step to make buttons clickable again.

## Description
After the latest refactoring of the styles there was a slight mistake which caused that the buttons for the initial install wizard in the beginning weren't clickable anymore, because they got overlaid by another box.

## Types of changes

## Screenshots:
Bildschirmfoto 2019-06-20 um 17 01 55

## How Has This Been Tested?
Tested on local environment.

### DIFF
--- a/ui/css/cloudstack3.css
+++ b/ui/css/cloudstack3.css
@@ -784,7 +784,7 @@ body.install-wizard {
 
 .install-wizard .step {
   position: relative;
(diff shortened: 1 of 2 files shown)


svenvogel:
@anuragaw and @shwstppr can you take a look? ;) Thanks!

Issue #3305, opened 2019-04-29
SSL SSVM not working when configured to use a URL instead of a domain


fpaillot:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

SSVM

##### CLOUDSTACK VERSION

4.11.2

##### CONFIGURATION

consoleproxy.url.domain	: myurl.mydomain.com
consoleproxy.sslEnabled	: true
secstorage.ssl.cert.domain : myurl.mydomain.com 
secstorage.encrypt.copy : true

##### OS / ENVIRONMENT

N/A

##### SUMMARY

Based on : https://www.shapeblue.com/securing-cloudstack-4-11-with-https-tls/, I tried to configure a shared URL for SSVM and CPVM.

It works fine for CPVM but not for SSVM.

When I try to upload a template from local computer, I see in the logs :
~~~
 DEBUG [o.a.c.s.e.DefaultEndPointSelector] (Upload-Monitor-1:ctx-3b853e8b) (logid:9f919e25) Coudn't find ssvm for urlhttps://A-B-C-Dmyurl.mydomain.com/upload/b2aa7a84-90dc-4a13-94ec-8c34970f16b8
~~~

It seems that the URL builder does not take into the account the fact I'm not a using a wildcard domain (*.mydomain.com) but a full URL (myurl.mydomain.com).

##### STEPS TO REPRODUCE

*Configure SSL for CPVM and SSVM with a static URL in the options
*Upload the matching certificate
*Restart/recreae the SSVM and CPVM
*Try to upload a template

##### EXPECTED RESULTS

~~~

The generated URL for the upload should be : https://myurl.mydomain.com/upload/

~~~

##### ACTUAL RESULTS

~~~
https://A-B-C-Dmyurl.mydomain.com/upload/

~~~

rhtyd:
The settings clearly ask to enter the domain and not the URL. @fpaillot you may use a service such as xip.io or sslip.io for dynamic resolution of domain-address.

PR #3420, opened 2019-06-24
ssvm: use secstorage.ssl.cert.domain as hostname if it does not start with '*' when upload a template or volume from local

ustcweizhou:
## Description

If secstorage.ssl.cert.domain does not start with '*', we should use it as host name in url when upload template/volume from local

Fixes: #3305

## Types of changes


rhtyd:
@blueorangutan package

anuragaw:
The PR is for 4.11 - just wanted to confirm it's on purpose.

rhtyd:
Sometimes authors send PR for older branches, we'll forward merge in that case @anuragaw

rhtyd:
@blueorangutan test

PR #3328, opened 2019-05-10
Enhancement scss refactoring

webermaximilian92:
## Description
This PR is an SCSS-refactoring addition to PR: https://github.com/apache/cloudstack/pull/3307#issuecomment-491314203

This enhances code-quality of scss code by
- including tool-tasks like sass-lint-fix (autodetects scss lint warnings and fix them if possible)
- adding jsbeautifyrc, that can be used by IDE extensions to unify scss code automatically 
- deleting old css files for IE7, that will no longer be in use today (almost 13 years old browser)
- rethinking of old "custom.css"-style-file. (it's not beneficial to generate own custom css file anymore. The less css, the less server requests. Rather now any CSS code can be included inside custom.scss file and will be compiled at bottom of cloudstack3.css by compiler.
- adding z-index variables to better organize treacherous interdependently z-index values.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?
By running compilers and check build css code and by visiting the page with new css file.
(This changes modify and unify the structure of css code only and deleting old unused code)
All we have to check yet, is if cloudstack3.css includes any effective changes.

It's very useful to reduce the diff to non whitespace changes here again:
Screenshot 2019-04-26 at 16 23 44

PR #3428, opened 2019-06-26
Build css from scss sources in client ui build

pbankonier:
## Description
* This PR integrates the compiling of the scss sources in the build of the client ui module, so the css files can be removed from the sourcecode.
I also moved the a folder which contains all the frontend resources from the root directory to the client a folder, because they get only used in this module.

* **Furthermore, I readded the empty custom.css file which was missing due to the changes in the refactoring if the styles. this regression was introduced in PR #3328** 

## Types of changes

## How Has This Been Tested?


rhtyd:
There are many conflicts, will this break UI building/usage and introduce a new dependency (npm etc.)?

pbankonier:
I will split the bugfix and the build changes in two PRs.

Issue #2895, opened 2018-10-11
ssh-keys instances ERROR


izenk:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1
~~~

##### CONFIGURATION

KVM, Advanced Networking

##### OS / ENVIRONMENT

CentOS 7.4

##### SUMMARY

When click to "View Instances" link in project View for ssh-keys - error is displayed instead of instances list 

##### STEPS TO REPRODUCE

Login to UI -> switch to project view -> click to "Accounts" -> choose "SSH keys" -> choose any ssh key -> click "View instances" [Screen](https://postimg.cc/DJ05Gwdz)

resmo:
similar problem fixed with network.js in bff7c01f19264e85c0da4a79c45f619761cfcc86

anuragaw:
Pushed a PR for fixing the bug @izenk

PR #3435, opened 2019-06-28
ui: don't pass account name when in project mode to VMs from sshkeypair reference

anuragaw:
When users try to list instances in project mode from Account > SSHKeyPairs > Key > View Instances, we don't need to include sshkeypairs account name into the API Param. It's redundant and causes error.

## Description

Fixes: #2895 
## Types of changes

## How Has This Been Tested?

Manually in UI try various combinations of reproducing bug in admin/user mode and switchng between regular and project view.

### DIFF
--- a/ui/scripts/instances.js
+++ b/ui/scripts/instances.js
@@ -750,9 +750,13 @@
                 if ("sshkeypairs" in args.context) {
                     $.extend(data, {
                         domainid: args.context.sshkeypairs[0].domainid,
-                        account: args.context.sshkeypairs[0].account,
                         keypair: args.context.sshkeypairs[0].name
                     });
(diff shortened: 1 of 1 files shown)


anuragaw:
ping for review - @rhtyd , @shwstppr , @svenvogel , @PaulAngus , @izenk

anuragaw:
@rhtyd - that's correct. When the above path is taken in UI the project name is passed into account name for view instances. We need to ignore the project account name since project id is already populated when we createURL

rhtyd:
Thanks @anuragaw so what you're saying is fix how to data is passed and rest is taken care of. Let's ask @svenvogel @DennisKonrad @pbankonier @NicoWohlfarth and other UI contributors for testing and review.

anuragaw:
That's correct @rhtyd .

PR #3441, opened 2019-06-28
Bug fix for distinting between string and map type tags in forms

anuragaw:
Some APIs consume 'tags' param as string and some consume as maps.
Since each API can have at most one 'tags' param the extraction of
map based tags should only happen when strings based tags are not
extracted from the form serialization.

## Types of changes

## How Has This Been Tested?

Try tagging in 2 different places -
1. Snapshot , Snapshot policies
2. Disk offering creation

Both places should work. Without this patch only former would work.

### DIFF
--- a/ui/scripts/ui/dialog.js
+++ b/ui/scripts/ui/dialog.js
@@ -778,7 +778,7 @@
 
             var complete = function($formContainer) {
                 var $form = $formContainer.find('form');
-                var data = cloudStack.serializeForm($form));
+                var data = cloudStack.serializeForm($form);
                 if (!data.tags) {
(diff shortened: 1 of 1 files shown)


anuragaw:
ping for review - @rhtyd , @shwstppr , @nvazquez , @borisstoyanov , @PaulAngus

anuragaw:
Tags extraction was a recent change I had worked on for supporting snapshot and snapshot policy tags. This was merged in the last few days . This PR change makes it more conservative and avoids overwriting of string based tags if any because a form should not contain multiple params with name 'tags'

rhtyd:
Merged based on reviews, manual test and screenshot.

Issue #2724, opened 2018-06-29
Service Offering Sorting is broken in UI


bwsw:
##### ISSUE TYPE
Bug Report

##### COMPONENT NAME
UI

##### CLOUDSTACK VERSION
4.11.1

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
Ubuntu 16.04

##### SUMMARY
Trying to order Compute service offerings in UI, but the result is the same as before sorting. UI controls allow moving offerings against each other, including drag and drop way, but after page reload they are still in the same order as before.

##### STEPS TO REPRODUCE
Create several service offerings, try to reorder them in UI. 

##### EXPECTED RESULTS
Items reordered

##### ACTUAL RESULTS
Items in the same order

bwsw:
Later investigation has shown that it also broke in Templates/ISO sorting order.

rhtyd:
@anuragaw can you verify fix this, since I think you've fixed sorting for couple of resources in the UI.

anuragaw:
Copy @rhtyd

anuragaw:
DB support for service offerings was existing but I discovered another bug while testing and fixed it in the above PR.

rhtyd:
As per remarks on https://github.com/apache/cloudstack/pull/3445, this is fixed. Closing.

PR #3445, opened 2019-06-29
Fix sorting order bug in UI Code with usage of sortkey.algorithm global config

anuragaw:
Current master has sorting broken and the order reverses as opposed to
what is set if sortkey.algorithm is set to true. The problem lies in the
way the update APIs were being called by the UI. The code was agnostic
to a global config that backend uses to set the order of the entities
in the corresponding list APIs.

We need to make UI aware of the global config and instead of changing
sign of sort key pass proper numbers so that DB isn't confusing when
some tables have positive sortkey and some have negative.

The fix is in 2 steps-
1) Make use of sortkey name in place where it's relevant. Mere row index
is not sufficient.
2) Reverse the sortkey if we are sorting by descending (when global
config is false)
Fix is general bugfix but I also verified that #2724 is also fixed after this change and verifiable. The DB support already existed.
## Types of changes

## How Has This Been Tested?

Flip the config sortkey.algorithm and verify the templates and offerings order is retained as the UI sets in either case.

### DIFF
--- a/ui/scripts/sharedFunctions.js
+++ b/ui/scripts/sharedFunctions.js
@@ -2419,7 +2419,7 @@ cloudStack.api = {
                     url: createURL(updateCommand),
                     data: {
                         id: args.context[objType].id,
-                        sortKey: g_sortKeyIsAscending ? (-1 * args.index) : args.index
+                        sortKey: args.sortKey
                     },
                     success: function(json) {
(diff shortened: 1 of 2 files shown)


anuragaw:
ping @shwstppr, @nvazquez , @borisstoyanov - for quick review on this small but important change.

PR #3451, opened 2019-07-01
[UI] Fix wrap text for accounts on project view

nvazquez:
## Description
Fix wrap text for account names on the project view, as it cannot be fully displayed

## Types of changes

Before the fix the account column does not wrap the full text:
image

After the fix:
image

## How Has This Been Tested?
Create Project
Select Project -> Accounts

### DIFF
--- a/ui/css/cloudstack3.css
+++ b/ui/css/cloudstack3.css
@@ -8154,7 +8154,6 @@ div.container div.panel div#details-tab-addloadBalancer.detail-group div.loadBal
   display: block;
   float: left;
   max-width: 90%;
-  white-space: nowrap;
   text-overflow: ellipsis;
   overflow: hidden;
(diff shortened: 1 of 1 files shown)


nvazquez:
@blueorangutan package

DaanHoogland:
@rhtyd for sure, all multiedit list views. not sure how many those are, if any. I think users would wnat this to be configurable.

nvazquez:
@rhtyd yes, as @DaanHoogland mentions it affects the multiEdit views. In my opinion, could improve each usage if the full text was not displayed, do you think it should only address the problem for the project view?

DaanHoogland:
@nvazquez No, I think it is fine for other cases as well. let's collect a list here if we encounter them.

PR #3464, opened 2019-07-04
Proposed refactor of memorykbs to memoryusedkbs to reflect description on API docs

kioie:
## Description

Based on the [API docs](https://cloudstack.apache.org/api/apidocs-4.12/apis/listVirtualMachines.html) this field `memorykbs` has the description `the memory used by the vm` . 
The refactor to `memoryusedkbs` enhances clarity in line with the naming scheme.

## Types of changes

Screen Shot 2019-07-04 at 9 30 54 AM

## How Has This Been Tested?

Similar to all tests for the `listvirtualmachines` api

### DIFF
--- a/api/src/org/apache/cloudstack/api/response/UserVmResponse.java
+++ b/api/src/org/apache/cloudstack/api/response/UserVmResponse.java
@@ -196,9 +196,9 @@ public class UserVmResponse extends BaseResponseWithTagInformation implements Co
     @Param(description = "the write (bytes) of disk on the vm")
     private Long diskKbsWrite;
 
-    @SerializedName("memorykbs")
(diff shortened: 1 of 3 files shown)


rhtyd:
This will break api response because right now the docs says:
```
memorykbs | the memory used by the vm
```

kioie:
> This will break api response because right now the docs says:
> 
> ```
> memorykbs | the memory used by the vm
> ```

Can we consider moving it to the next version?

rhtyd:
@kioie break API compatibility should be avoided. You may start a discussion thread on dev@. The possible middle ground could be to introduce a new variable and set it.

kioie:
@rhtyd noted...will raise proposal on mailing list

PR #3467, opened 2019-07-04
schema: add support for XenServer 7.1.2 (LTS)

### LINKED ISSUE #3461: guest_os_hypervisor missing entries for XenServer 7.1.2
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
DB
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2, 4.11.3
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

XenServer 7.1.2 (7.1 CU2) Hosts

##### SUMMARY

There are no entries in guest_os_hypervisor for XenServer 7.1.2 (7.1 CU2 - the current LTS XenServer) which means guests are booted using the "Other Install Media" template.  Amongst other things, this means that PV guests boot as HV, but also likely means guests won't get XenServer optimisations for their OS type.

Perhaps a better way rather than just inserting specific rows for this version would be to add logic to ignore the revision number in a version matching pattern x.y.z if there are no rows found, e.g. look for row with version = 7.1.2, fallback to row with version = 7.1.0 (would still work if XenServer is updated without CloudStack being updated).

Related: #3086

##### STEPS TO REPRODUCE

Install Fresh CloudStack instance with new XenServer 7.1.2 hosts.  Wait for CS to try to deploy system VMs.

##### EXPECTED RESULTS

System VMs should boot using the Debian Wheezy 7.0 (64-bit) XenServer template

##### ACTUAL RESULTS

System VMs should boot using the Other Install Media XenServer template (which fails to boot in a virtual environment using XenServer running on VMs, only capable of running PV).

rhtyd:
Fixes #3461

Adds upgrade path to add support for XenServer 7.1.2. A larger refactoring can be done in 4.14+ to treat all dot-releases as the same.

## Types of changes


rhtyd:
@blueorangutan package

rhtyd:

Issue #3400, opened 2019-06-12
Allow normal users to download template when template are not extractable


rhtyd:
On VMware env, a user is unable to download the template without having the extractable option enabled. 

It is reported from the user that the option to enable the extractable parameter in project view is unavailable for normal users and the same is in the default view. (may need an investigation to confirm). The same options may be provided while creating a template from a volume snapshot.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VMware, templates
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2.0
~~~

ustcweizhou:
as far as I know only the owner can change the option.

rhtyd:
I've not tested it, if it's not a valid bug, we can close it after reproduction.

rhtyd:
@Spaceman1984 can you have a look on it?

rhtyd:
Log examples:
```
2019-07-02 17:15:28,721 DEBUG [o.a.c.f.j.i.AsyncJobManagerImpl] (API-Job-Executor-13:ctx-d7ae5b77 job-274) (logid:ec555145) Executing AsyncJobVO {id:274, userId: 9, accountId: 14, instanceType: Template, instanceId: 203, cmd: org.apache.cloudstack.api.command.user.template.ExtractTemplateCmd, cmdInfo: {"mode":"HTTP_DOWNLOAD","response":"json","ctxUserId":"9","zoneid":"f873972d-d0b6-4122-9ac6-18663fd05bf3","httpmethod":"GET","ctxStartEventId":"1182","id":"57e47508-ca63-40eb-9a87-33c588ba8088","ctxDetails":"{\"interface com.cloud.dc.DataCenter\":\"f873972d-d0b6-4122-9ac6-18663fd05bf3\",\"interface com.cloud.template.VirtualMachineTemplate\":\"57e47508-ca63-40eb-9a87-33c588ba8088\"}","ctxAccountId":"14","uuid":"57e47508-ca63-40eb-9a87-33c588ba8088","cmdEventType":"TEMPLATE.EXTRACT","_":"1562067928698"}, cmdVersion: 0, status: IN_PROGRESS, processStatus: 0, resultCode: 0, result: null, initMsid: 20634575163998, completeMsid: null, lastUpdated: null, lastPolled: null, created: null}
[...]

PR #3475, opened 2019-07-05
Allowing template owner to download template

Spaceman1984:
## Description
Removed the download icon when a template is not extractable.

Modified the api to allow a user from the same account as the template, to change the extractable attribute on the template.

Fixes: #3400

## Types of changes

## How Has This Been Tested?

Tested using the UI.
Create a user, then login with that user. Navigate to templates, add a template.
Edit the template, click on the extractable checkbox. Save changes. 

At this point, there is no error.

### DIFF
--- a/ui/scripts/templates.js
+++ b/ui/scripts/templates.js
@@ -1378,12 +1378,7 @@
                                     isextractable: {
                                         label: 'label.extractable.lower',
                                         isBoolean: true,
-                                        isEditable: function() {
-                                            if (isAdmin())
-                                                return true;
-                                            else
-                                                return false;
-                                        },
+                                        isEditable: true,
                                         converter: cloudStack.converters.toBooleanText
                                     },
                                     passwordenabled: {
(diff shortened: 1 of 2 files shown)


ui/scripts/templates.js
```diff
@@ -1378,12 +1378,7 @@
                                     isextractable: {
                                         label: 'label.extractable.lower',
                                         isBoolean: true,
-                                        isEditable: function() {
-                                            if (isAdmin())
-                                                return true;
-                                            else
-                                                return false;
-                                        },
+                                        isEditable: true,
```
rhtyd: Can you add a check to only allow editable option when it's either the root admin of when template account matches the logged in user account?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@anuragaw @shwstppr can you review this PR?
@Spaceman1984 can you address review comments.

Thanks.

ui/scripts/templates.js
Spaceman1984: The edit button is not shown for public templates when logged in with a different account, so the extractable checkbox is not displayed.

rhtyd:
@borisstoyanov @shwstppr @anuragaw can you review and test. I've not tested it. Few cases; what happens to template wrt edit, download button for template registered by normal users under a domain, wrt root domain, domain admin, other accounts in the sub-domain etc.

rhtyd:
Smoketests LGTM.

rhtyd:
This is a known issue @shwstppr to go back and refresh the page. This is a general limitation of the UI.

Issue #3478, opened 2019-07-09
When system vm is deployed from a new secondary pool it may fail to live migrate because the new secondary storage pool is not mounted in target host.


borisstoyanov:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Storage
~~~

##### CLOUDSTACK VERSION

~~~
4.9
~~~

##### SUMMARY

With VMware, when a new secondary storage pool is added, and a system-vm is being deployed from that mount, cloudstack operator cannot live migrate the VM to a new host because the secondary storage pool is not mounted on the host. 

2019-07-02 17:18:20,490 DEBUG [o.a.c.f.j.i.AsyncJobManagerImpl] (API-Job-Executor-10:ctx-e0ee2e04 job-496401) (logid:4365c38d) Complete async job-496401, jobStatus: FAILED, resultCode: 530, result: org.apache.cloudstack.api.response.ExceptionResponse/null/{"uuidList":[],"errorcode":530,"errortext":"MigrationCommand failed due to Exception: java.lang.RuntimeException\nMessage: 現在接続しているデバイス「CD/DVD drive 1」が、アクセスできないバッキング「[acb61221d4cc3273852f15dea25b9c57] systemvm/systemvm-4.9.1.0.iso」を使用しています。\n"}

The message written in Japanese means "Device 'CD/DVD drive 1' uses backing '[acb61221d4cc3273852f15dea25b9c57] systemvm/systemvm-4.9.1.0.iso', which is not accessible.".

##### EXPECTED RESULTS

~~~
Migration should be successful 
~~~

##### ACTUAL RESULTS

~~~
Migration fails since the secondary storage pool is not mounted on the target host.
~~~

rhtyd:
I checked both 4.9, 4.11, master codebases prior to VM migration it does check that the secondary storage is mounted. However, the root cause is that when there are multiple secondary storages it picks one randomly which is mounted instead of the specific secondary storage.

Here it picks one datastore instead of a list for that zone:
https://github.com/apache/cloudstack/blob/master/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareManagerImpl.java#L495

rhtyd:
Related issue on secondary storage selection: https://github.com/apache/cloudstack/issues/3041

shwstppr:
[...]

PR #3302, opened 2019-04-29
server: sync templates on adding new secondary storage

shwstppr:
## Description
Fixes #2689 

With the current code, existing templates were not downloaded to the new secondary storage when it is added. SSVM needed to be restarted to start the download process. This PR starts templates sync for the new secondary storage when it is added.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?
Using KVM based monkeybox environment. New secondary storage was added and template download was verified by listing files for the storage.

PR #3480, opened 2019-07-10
engine, server, services: fix for respecting secondary storage threshold limit

### LINKED ISSUE #3287: Secondary storage pool 90pct limit ignored

shwstppr:
Retrieval of an image store using ImageStoreProviderManager has been refactored by introducing three different methods,
`DataStore getRandomImageStore(List<DataStore> imageStores);`
To get an image store for reading purpose. Threshold capacity check will not be used here.
`DataStore getImageStoreWithFreeCapacity(List<DataStore> imageStores);`
To get an image store for reading purpose. Threshold capacity check will be used here and the store with max free space will be returned. If no store with filled storage less than the threshold is found, the NULL value will be returned.
`List<DataStore> listImageStoresWithFreeCapacity(List<DataStore> imageStores);`
To get a list of image stores for writing purpose which fulfills threshold capacity check.

Correspondingly DataStoreManager methods have been refactored to return similar values for a given zone,
```
DataStore getRandomImageStore(long zoneId);
DataStore getImageStoreWithFreeCapacity(long zoneId);
List<DataStore> listImageStoresWithFreeCapacity(long zoneId);
```
Consumers have been refactored to accept NULL value and throw exception or log error if needed.

Fixes -
#3287 - NULL value will be returned when secondary storage is needed for writing but there is not store with free space.
#3041 - Rather than returning random secondary storage for writing, storage with max. free space will be returned.
#3478 - For migration on VMware, all writable secondary storage will be mounted while preparation.

## Types of changes

### DIFF
--- a/engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
+++ b/engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
@@ -33,11 +33,11 @@ public interface DataStoreManager {
 
     List<DataStore> getImageStoresByScope(ZoneScope scope);
 
-    DataStore getImageStoreForRead(long zoneId);
+    DataStore getRandomImageStore(long zoneId);
 
-    DataStore getImageStoreForWrite(long zoneId);
+    DataStore getImageStoreWithFreeCapacity(long zoneId);
 
-    List<DataStore> getImageStoresForWrite(long zoneId);
+    List<DataStore> listImageStoresWithFreeCapacity(long zoneId);
 
     List<DataStore> getImageCacheStores(Scope scope);
 
--- a/engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
+++ b/engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
@@ -37,6 +37,8 @@ public interface DataStoreManager {
 
     DataStore getImageStoreForWrite(long zoneId);
 
+    List<DataStore> getImageStoresForWrite(long zoneId);
+
     List<DataStore> getImageCacheStores(Scope scope);
 
     DataStore getImageCacheStore(long zoneId);
--- a/engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
+++ b/engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
@@ -33,7 +33,9 @@ public interface DataStoreManager {
 
     List<DataStore> getImageStoresByScope(ZoneScope scope);
 
-    DataStore getImageStore(long zoneId);
+    DataStore getImageStoreForRead(long zoneId);
+
+    DataStore getImageStoreForWrite(long zoneId);
 
     List<DataStore> getImageCacheStores(Scope scope);
 
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -197,7 +197,7 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
             Host destHost) {
         VMTemplateStoragePoolVO sourceVolumeTemplateStoragePoolVO = vmTemplatePoolDao.findByPoolTemplate(destStoragePool.getId(), srcVolumeInfo.getTemplateId());
         if (sourceVolumeTemplateStoragePoolVO == null && destStoragePool.getPoolType() == StoragePoolType.Filesystem) {
-            DataStore sourceTemplateDataStore = dataStoreManagerImpl.getImageStoreForRead(srcVolumeInfo.getDataCenterId());
+            DataStore sourceTemplateDataStore = dataStoreManagerImpl.getRandomImageStore(srcVolumeInfo.getDataCenterId());
             if (sourceTemplateDataStore != null) {
                 TemplateInfo sourceTemplateInfo = templateDataFactory.getTemplate(srcVolumeInfo.getTemplateId(), sourceTemplateDataStore);
                 TemplateObjectTO sourceTemplate = new TemplateObjectTO(sourceTemplateInfo);
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -197,7 +197,7 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
             Host destHost) {
         VMTemplateStoragePoolVO sourceVolumeTemplateStoragePoolVO = vmTemplatePoolDao.findByPoolTemplate(destStoragePool.getId(), srcVolumeInfo.getTemplateId());
         if (sourceVolumeTemplateStoragePoolVO == null && destStoragePool.getPoolType() == StoragePoolType.Filesystem) {
-            DataStore sourceTemplateDataStore = dataStoreManagerImpl.getImageStore(srcVolumeInfo.getDataCenterId());
+            DataStore sourceTemplateDataStore = dataStoreManagerImpl.getImageStoreForRead(srcVolumeInfo.getDataCenterId());
             if (sourceTemplateDataStore != null) {
                 TemplateInfo sourceTemplateInfo = templateDataFactory.getTemplate(srcVolumeInfo.getTemplateId(), sourceTemplateDataStore);
                 TemplateObjectTO sourceTemplate = new TemplateObjectTO(sourceTemplateInfo);
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -24,18 +24,17 @@ import java.util.Set;
 
 import javax.inject.Inject;
 
-import com.cloud.storage.ScopeType;
-import com.cloud.storage.Storage;
 import org.apache.cloudstack.engine.subsystem.api.storage.DataStore;
+import org.apache.cloudstack.engine.subsystem.api.storage.ObjectInDataStoreStateMachine;
 import org.apache.cloudstack.engine.subsystem.api.storage.StrategyPriority;
 import org.apache.cloudstack.engine.subsystem.api.storage.TemplateDataFactory;
 import org.apache.cloudstack.engine.subsystem.api.storage.TemplateInfo;
 import org.apache.cloudstack.engine.subsystem.api.storage.VolumeInfo;
-import org.apache.cloudstack.engine.subsystem.api.storage.ObjectInDataStoreStateMachine;
 import org.apache.cloudstack.storage.command.CopyCommand;
 import org.apache.cloudstack.storage.datastore.DataStoreManagerImpl;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
 import org.apache.cloudstack.storage.to.TemplateObjectTO;
+import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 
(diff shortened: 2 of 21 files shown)


engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
```diff
@@ -198,18 +198,20 @@ protected void copyTemplateToTargetFilesystemStorageIfNeeded(VolumeInfo srcVolum
         VMTemplateStoragePoolVO sourceVolumeTemplateStoragePoolVO = vmTemplatePoolDao.findByPoolTemplate(destStoragePool.getId(), srcVolumeInfo.getTemplateId());
         if (sourceVolumeTemplateStoragePoolVO == null && destStoragePool.getPoolType() == StoragePoolType.Filesystem) {
             DataStore sourceTemplateDataStore = dataStoreManagerImpl.getImageStore(srcVolumeInfo.getDataCenterId());
-            TemplateInfo sourceTemplateInfo = templateDataFactory.getTemplate(srcVolumeInfo.getTemplateId(), sourceTemplateDataStore);
-            TemplateObjectTO sourceTemplate = new TemplateObjectTO(sourceTemplateInfo);
+            if (sourceTemplateDataStore != null) {
```
shwstppr: @rhtyd Throwing an exception or returning null for Image store can fail this. This might not be needing a sec store for writing. Maybe need to handle this separately?

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2677,6 +2677,9 @@ private String orchestrateExtractVolume(long volumeId, long zoneId) {
         }
         // perform extraction
         ImageStoreEntity secStore = (ImageStoreEntity)dataStoreMgr.getImageStore(zoneId);
+        if (secStore == null) {
+            throw new InvalidParameterValueException(String.format("Secondary storage to satisfy storage needs cannot be found for zone: %d", zoneId));
+        }
         String value = _configDao.getValue(Config.CopyVolumeWait.toString());
```
kioie: What about the `VolumeVO uploadVolume` and the `GetUploadParamsResponse uploadVolume` , that also use  `getImageStore` , will you also handle the null returns there?

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -1118,7 +1118,10 @@ public boolean finalizeVirtualMachineProfile(VirtualMachineProfile profile, Depl
         vm.setDetails(details);
 
         DataStore secStore = _dataStoreMgr.getImageStore(dest.getDataCenter().getId());
-        assert (secStore != null);
+        if (secStore != null) {
```
kioie: `if (secStore == null){`

engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
kioie: Could this be the reason you are using `(sourceTemplateDataStore != null)` ? If so, you might need an else at the end to manage the nulls

shwstppr: @kioie It should never be getting null value for image store as it needs the store only for reading and threshold limit shouldn't be affecting this. Will have to make changes for it.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
shwstppr: Not seeing any reference for getImageStore(Long zoneId) here,
https://github.com/shapeblue/cloudstack/blob/secondary-storage-limit-3287/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java#L290-L388
`DataStore getImageStore(String storeUuid, Long zoneId)` from `TemplateManager.java` used in these methods already handles null value.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
shwstppr: Done

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
kioie: Yes, you are correct, hadn't seen the `CloudRuntimeException` inside the method.

engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/manager/ImageStoreProviderManagerImpl.java
```diff
@@ -146,17 +145,30 @@ public ImageStoreEntity getImageStore(String uuid) {
...
+                return imageStores.get(0);
+            }
         }
-        return imageStores.get(0);
+
+        // No store with space found
+        s_logger.error(String.format("Can't find an image storage in zone with less than %d usage",
+                Math.round(_statsCollector.getImageStoreCapacityThreshold()*100)));
+        return null;
     }
```
kioie: Besides checking for whether the store is below the threshold, are you also considering doing a check if the image will fit in the imagestore and what capacity remains afterwards....for example, if the current used capacity is at 89% and the image might move it to even upto 99%....or is that check happening elsewhere?

shwstppr: @kioie I'm not sure that is possible to implement for scenarios at this point with the current design. For example, while registering a template through URL we don't know actual size after download and install.
But such a check can be considered for copy template, snapshots, etc. I've to check the code for that. Anyway, the requirement of this method is to return an image store with capacity within the threshold and I don't think we can have a check for resultant size here.

kioie: That's okay, just wanted to know the possibility of it happening

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test

shwstppr:
@blueorangutan test

rhtyd:
@shwstppr were the vmware specific fixes also done? You can kick tests now.
@blueorangutan package

shwstppr:

shwstppr:
@rhtyd yes I made the changes for vmware migration issue. Would help if you can review last 2 commits

engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
```diff
@@ -33,7 +33,11 @@
 
     List<DataStore> getImageStoresByScope(ZoneScope scope);
 
-    DataStore getImageStore(long zoneId);
+    DataStore getImageStoreForRead(long zoneId);
+
+    DataStore getImageStoreForWrite(long zoneId);
+
+    List<DataStore> getImageStoresForWrite(long zoneId);
```
rhtyd: Rename it something more meaningful like `listImagesStoresWithFreeCapacity`?

engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
```diff
@@ -33,7 +33,11 @@
 
     List<DataStore> getImageStoresByScope(ZoneScope scope);
 
-    DataStore getImageStore(long zoneId);
+    DataStore getImageStoreForRead(long zoneId);
+
+    DataStore getImageStoreForWrite(long zoneId);
```
rhtyd: How about getImagesStoresWithFreeCapacity?

engine/api/src/main/java/org/apache/cloudstack/engine/subsystem/api/storage/DataStoreManager.java
```diff
@@ -33,7 +33,11 @@
 
     List<DataStore> getImageStoresByScope(ZoneScope scope);
 
-    DataStore getImageStore(long zoneId);
+    DataStore getImageStoreForRead(long zoneId);
```
rhtyd: `getRandomImageStore` or something that conveys meaning? You may also put javadocs on them to convey which method does what?

shwstppr:
@rhtyd done the refactoring
@blueorangutan package

rhtyd:
Can you check the build failure @shwstppr  on backend jenkins?

shwstppr:
@blueorangutan package

shwstppr:
@rhtyd Not sure why it failed the first time. Logs showed checkstyle errors related to files unrelated to this PR. Packaging seems fine now.

shwstppr:

rhtyd:
Re-kicking Travis

rhtyd:
I've rekicked Travis jobs, please check again in case they fail @shwstppr 
@blueorangutan test matrix

rhtyd:

shwstppr:
@borisstoyanov will look into it

> I think I'm hitting an issue with this one. I cannot sync the templates across a newly added image store. Here's my steps:
> 
> Originally created environement with 4 secondary storages
> deleted 3 of them so I just have 1 running.
> On the deleted storages I've clean all files.
> Registered a few templates in cloudstack and they ended up on the running storage
> Added back one of the storages and destroyed the SSVM in a while
> After a day I went back to the machine and the newly added templates were not on the new image store, just the default ones were added. Also the new secondary storage is not being mounted on any of the hosts

rhtyd:
Tests LGTM, any comments @shwstppr @borisstoyanov  ?
@shwstppr can you test/fix Bobby's issue?

shwstppr:
@rhtyd looking into the issue

@blueorangutan package

shwstppr:
> I think I'm hitting an issue with this one. I cannot sync the templates across a newly added image store. Here's my steps:
> 
> Originally created environement with 4 secondary storages
> deleted 3 of them so I just have 1 running.
> On the deleted storages I've clean all files.
> Registered a few templates in cloudstack and they ended up on the running storage
> Added back one of the storages and destroyed the SSVM in a while
> After a day I went back to the machine and the newly added templates were not on the new image store, just the default ones were added. Also the new secondary storage is not being mounted on any of the hosts

@borisstoyanov I've verified, the issue you are facing is unrelated to this PR. Currently, code skips downloading a template to new secondary store if is not public and not featured and not a system template.
https://github.com/apache/cloudstack/blob/master/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java#L517-L520
If you can repeat your steps while making new template public/fetured(or remove your new stores, edit template to make it public and re-add store) it will be downloaded instantly after storage addition. There is no issue with template sync and it should start as soon as a new storage is added. I fixed that with #3302 
Not sure if it is by design that a normal template(not featured, not public) will remain on a single secondary store for a zone. @rhtyd @nvazquez any thoughts?

borisstoyanov:
Sure, thanks on the investigation @shwstppr, will check

rhtyd:
@borisstoyanov @shwstppr have you tested this simple case:
(a) deploy a VM on a host
(b) add a new secondary storage to a zone, then check that the secondary storage is not mounted on any of the hosts
(c) then migrate that VM to another host, check that the secondary storage is mounted on that host.

PR #3494, opened 2019-07-15
Fix hardcoded max data volumes when VM has been created but not started before

nvazquez:
## Description
When VM is created but not started on any host, attaching a volume on it causes the maximum number of allowed volumes to be 6 (hardcoded)

## Types of changes

## How Has This Been Tested?
- Deploy VM with startvm=false parameter
- Attach more than 6 volumes to VM (in case of KVM and VMware)

### DIFF
--- a/engine/schema/src/main/java/com/cloud/hypervisor/dao/HypervisorCapabilitiesDaoImpl.java
+++ b/engine/schema/src/main/java/com/cloud/hypervisor/dao/HypervisorCapabilitiesDaoImpl.java
@@ -16,6 +16,7 @@
 // under the License.
 package com.cloud.hypervisor.dao;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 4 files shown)


nvazquez:
Ping @anuragaw @shwstppr can you please review?

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -3007,6 +3009,13 @@ private int getMaxDataVolumesSupported(UserVmVO vm) {
         if (host != null) {
             _hostDao.loadDetails(host);
             maxDataVolumesSupported = _hypervisorCapabilitiesDao.getMaxDataVolumesLimit(host.getHypervisorType(), host.getDetail("product_version"));
+        } else {
+            List<HypervisorType> supportingDefaultHV = Arrays.asList(HypervisorType.XenServer, HypervisorType.VMware,
+                    HypervisorType.KVM, HypervisorType.Ovm, HypervisorType.LXC);
+            HypervisorType hypervisorType = vm.getHypervisorType();
+            if (hypervisorType != null && supportingDefaultHV.contains(hypervisorType)) {
+                maxDataVolumesSupported = _hypervisorCapabilitiesDao.getMaxDataVolumesLimit(hypervisorType, "default");
```
rhtyd: Could version specific values cause regression or bug?

DaanHoogland: if these are lower the "default", yes @rhtyd. But that should never be the case, should it?

DaanHoogland:
hudson couldn't reach github :(

DaanHoogland:
try again

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
nvazquez: Good point, I've verified any version contains max limit lower than the default for each hypervisor

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan package

engine/schema/src/main/java/com/cloud/hypervisor/dao/HypervisorCapabilitiesDaoImpl.java
```diff
@@ -106,4 +107,16 @@ public Boolean isVmSnapshotEnabled(HypervisorType hypervisorType, String hypervi
         HypervisorCapabilitiesVO result = getCapabilities(hypervisorType, hypervisorVersion);
         return result.getVmSnapshotEnabled();
     }
+
+    @Override
+    public List<HypervisorType> getHypervisorsWithDefaultEntries() {
```
anuragaw: Is there a possibility of high frequency calls to this method? If so we can perhaps optimize query calls with some caching? If not LGTM

nvazquez: Thanks @anuragaw, have refactored it to invoke this only once

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test matrix

rhtyd:
Tests LGTM, they are failing irrespective of this PR.

Issue #3488, opened 2019-07-12
cloud0 subnet should not be hardcoded to 169.254.0.1/16


wido:
We are setting up a BGP+EVPN+VXLAN setup using Frr and BGP Unnumbered and this is causing some problems with the *cloud0* bridge created by the cloudstack agent.

Although the global setting *control.cidr* can be modified, the KVM Agent will still create this bridge with a hardcoded subnet:

**com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java**
<pre>
        if (!foundLinkLocalBr) {
            Script.runSimpleBashScript("ip address add 169.254.0.1/16 dev " + linkLocalBr + ";" + "ip route add " + NetUtils.getLinkLocalCIDR() + " dev " + linkLocalBr + " src " +
                    NetUtils.getLinkLocalGateway());
        }
</pre>

When using BGP Unnumbered it will try to create a route pointing to *169.254.0.1*

This works until the CloudStack Agent is started:

<pre>
10.255.255.1 via 169.254.0.1 dev enp81s0f1 proto bgp metric 20 onlink 
10.255.255.2 via 169.254.0.1 dev enp81s0f1 proto bgp metric 20 onlink 
10.255.255.3 via 169.254.0.1 dev enp81s0f1 proto bgp metric 20 onlink 
10.255.255.4 via 169.254.0.1 dev enp81s0f1 proto bgp metric 20 onlink 
10.255.255.6 via 169.254.0.1 dev enp81s0f1 proto bgp metric 20 onlink 
10.255.255.7 via 169.254.0.1 dev enp81s0f1 proto bgp metric 20 onlink 
10.255.255.8 via 169.254.0.1 dev enp81s0f1 proto bgp metric 20 onlink 
</pre>

After the CloudStack Agent is started the 169.254.0.1/16 is added to cloud0 and not allowing Frr to create these routes:

<pre>
zebra[5010]: 0:10.255.255.8/32: Route install failed
zebra[4562]: Extended Error: Nexthop has invalid gateway
</pre>

The solution would be that through *agent.properties* this CIDR can be controlled and isn't hardcoded.

rhtyd:
Sounds like a config issue, @wido can you ensure that BGP is not automatically enforced for cloud0. I think from KVM agent perspective it is not aware what kind of network topology/model exists, it simply works on the idea of a network device.

wido:
[...]

PR #151, opened 2015-04-08
CLOUDSTACK-6697: bigswitch networking plugin update

kwanggithub:
1. provide compatibility with the Big Cloud Fabric (BCF) controller
   L2 Connectivity Service in both VPC and non-VPC modes
2. virtual network terminology updates: VNS --> BCF_SEGMENT
3. uses HTTPS with trust-always certificate handling
4. topology sync support with BCF controller
5. support multiple (two) BCF controllers with HA
6. support VM migration
7. support Firewall, Static NAT, and Source NAT with NAT enabled option
8. add VifDriver for Indigo Virtual Switch (IVS)

PR #3500, opened 2019-07-17
kvm/bridge: Allow Link Local Cidr (cloud0 interface) to be configured

wido:
## Description
There are certain scenarios where the 169.254.0.0/16 subnet is used for different
purposes then CloudStack on a hypervisor.

Once of such scenarios is a BGP+EVPN+VXLAN setup using BGP Unnumbered where the
169.254.0.1 address is used by Frr/Zebra BGP routing to send traffic to the
neighboring router.

The following settings can be changed in the agent.properties (default values added):

control.cidr=169.254.0.0/16

Make sure the global setting 'control.cidr' matches the values defined in the agent.propeties!

In the future the mgmt server can send this parameter to a KVM Agent on startup, but at the moment
this framework is not in place and thus these values can't be send to the Agent in a proper manner.

Fixes: #3488

## Types of changes

## How Has This Been Tested?
Ran local test on our cloud environment and verified cloud0 now has a different address.

<pre>
root@hv-138-a05-23:~# ip addr show cloud0
58: cloud0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether fe:00:a9:fe:f0:c9 brd ff:ff:ff:ff:ff:ff
    inet 169.254.240.1/20 scope global cloud0
       valid_lft forever preferred_lft forever
    inet6 fe80::8c3c:82ff:fe5a:77c/64 scope link 
       valid_lft forever preferred_lft forever
root@hv-138-a05-23:~#
</pre>

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
@@ -28,17 +28,17 @@ import java.util.regex.Pattern;
 
 import javax.naming.ConfigurationException;
 
+import com.cloud.utils.StringUtils;
+import com.cloud.utils.net.NetUtils;
+import com.cloud.utils.script.OutputInterpreter;
+import com.google.common.base.Strings;
 import org.apache.log4j.Logger;
 import org.libvirt.LibvirtException;
-
-import com.google.common.base.Strings;
 
 import com.cloud.agent.api.to.NicTO;
 import com.cloud.exception.InternalErrorException;
 import com.cloud.network.Networks;
 import com.cloud.utils.NumbersUtil;
-import com.cloud.utils.net.NetUtils;
-import com.cloud.utils.script.OutputInterpreter;
 import com.cloud.utils.script.Script;
 
 public class BridgeVifDriver extends VifDriverBase {
@@ -49,7 +49,7 @@ public class BridgeVifDriver extends VifDriverBase {
     private final Object _vnetBridgeMonitor = new Object();
     private String _modifyVlanPath;
     private String _modifyVxlanPath;
-    private String bridgeNameSchema;
+    private String _controlCidr = NetUtils.getLinkLocalCIDR();
     private Long libvirtVersion;
 
     @Override
@@ -67,7 +67,10 @@ public class BridgeVifDriver extends VifDriverBase {
             networkScriptsDir = "scripts/vm/network/vnet";
         }
 
-        bridgeNameSchema = (String)params.get("network.bridge.name.schema");
+        String controlCidr = (String)params.get("control.cidr");
+        if (StringUtils.isNotBlank(controlCidr)) {
+            _controlCidr = controlCidr;
+        }
 
         String value = (String)params.get("scripts.timeout");
         _timeout = NumbersUtil.parseInt(value, 30 * 60) * 1000;
@@ -384,7 +387,7 @@ public class BridgeVifDriver extends VifDriverBase {
     private void deleteExistingLinkLocalRouteTable(String linkLocalBr) {
         Script command = new Script("/bin/bash", _timeout);
         command.add("-c");
-        command.add("ip route | grep " + NetUtils.getLinkLocalCIDR());
+        command.add("ip route | grep " + _controlCidr);
         OutputInterpreter.AllLinesParser parser = new OutputInterpreter.AllLinesParser();
         String result = command.execute(parser);
         boolean foundLinkLocalBr = false;
@@ -397,15 +400,16 @@ public class BridgeVifDriver extends VifDriverBase {
                 }
                 final String device = tokens[2];
                 if (!Strings.isNullOrEmpty(device) && !device.equalsIgnoreCase(linkLocalBr)) {
-                    Script.runSimpleBashScript("ip route del " + NetUtils.getLinkLocalCIDR() + " dev " + tokens[2]);
+                    Script.runSimpleBashScript("ip route del " + _controlCidr + " dev " + tokens[2]);
                 } else {
                     foundLinkLocalBr = true;
                 }
             }
         }
+
         if (!foundLinkLocalBr) {
-            Script.runSimpleBashScript("ip address add 169.254.0.1/16 dev " + linkLocalBr + ";" + "ip route add " + NetUtils.getLinkLocalCIDR() + " dev " + linkLocalBr + " src " +
-                    NetUtils.getLinkLocalGateway());
+            Script.runSimpleBashScript("ip address add " + NetUtils.getLinkLocalAddressFromCIDR(_controlCidr) + " dev " + linkLocalBr);
+            Script.runSimpleBashScript("ip route add " + _controlCidr + " dev " + linkLocalBr + " src " + NetUtils.getLinkLocalGateway(_controlCidr));
         }
     }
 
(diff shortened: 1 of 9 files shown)


RPDiep:
In addition, also tested and verified: systemvms do get their ips from within the control.cidr and they are fully functional.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
```diff
@@ -397,15 +403,16 @@ private void deleteExistingLinkLocalRouteTable(String linkLocalBr) {
...
                     foundLinkLocalBr = true;
                 }
             }
         }
+
         if (!foundLinkLocalBr) {
-            Script.runSimpleBashScript("ip address add 169.254.0.1/16 dev " + linkLocalBr + ";" + "ip route add " + NetUtils.getLinkLocalCIDR() + " dev " + linkLocalBr + " src " +
-                    NetUtils.getLinkLocalGateway());
+            Script.runSimpleBashScript("ip address add " + NetUtils.getLinkLocalAddressFromCIDR(_controlCidr) + " dev " + linkLocalBr);
+            Script.runSimpleBashScript("ip route add " + _controlCidr + " dev " + linkLocalBr + " src " + NetUtils.getLinkLocalFirstAddressFromCIDR(_controlCidr));
```
nvazquez: How about keeping `NetUtils.getLinkLocalGateway()` but extending it to accept a CIDR parameter? As internally it invokes `getLinkLocalFirstAddressFromCIDR` as well

plugins/hypervisors/kvm/src/test/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriverTest.java
```diff
@@ -55,4 +55,5 @@ public void isValidProtocolAndVnetId() {
         Assert.assertTrue(driver.isValidProtocolAndVnetId("123", "vlan"));
         Assert.assertTrue(driver.isValidProtocolAndVnetId("456", "vxlan"));
     }
+
```
nvazquez: Extra line :)

utils/src/main/java/com/cloud/utils/net/NetUtils.java
```diff
@@ -962,26 +962,29 @@ public static String getLinkLocalNetMask() {
     }
 
     public static String getLinkLocalGateway() {
-        return "169.254.0.1";
+        return getLinkLocalFirstAddressFromCIDR(getLinkLocalCIDR());
     }
 
     public static String getLinkLocalCIDR() {
         return "169.254.0.0/16";
```
nvazquez: Same as the comment above, instead of returning a hardcoded value, this method can return the configured CIDR and can make `getLinkLocalGateway()` above to be consistent with it.

rhtyd:
@wido cc @GabrielBrascher - can you review the oustanding comments. Tests LGTM.
@blueorangutan package

rhtyd:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/IvsVifDriver.java
```diff
@@ -1,296 +0,0 @@
...
-import com.cloud.agent.api.to.NicTO;
-import com.cloud.exception.InternalErrorException;
-import com.cloud.hypervisor.kvm.resource.LibvirtVMDef.InterfaceDef;
-import com.cloud.network.Networks;
-import com.cloud.utils.NumbersUtil;
-import com.cloud.utils.net.NetUtils;
-import com.cloud.utils.script.OutputInterpreter;
-import com.cloud.utils.script.Script;
-
-public class IvsVifDriver extends VifDriverBase {
```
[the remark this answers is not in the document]
rhtyd: Ping @wido @GabrielBrascher is this unused class? If you not sure, can you not remove it?

rhtyd:
Left some remark, some changes I'll need to re-review, meanwhile I'll wait for regression tests to come back.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/IvsVifDriver.java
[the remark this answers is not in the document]
wido: This file is removed. It seems also that it was a type or some case.

It was unused, so that's why this PR removes it.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
wido: Could be, but there aren't many methods calling it. I see what you mean, I don't have a try preference for it though.

plugins/hypervisors/kvm/src/test/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriverTest.java
wido: Argh!

utils/src/main/java/com/cloud/utils/net/NetUtils.java
wido: Do we have access to the configuration here? I don't think so. Therefor I changed it this way.

wido:
I think I addressed the comments above. Some seem to be a coding style thing but not really a technical change.

Let me know if this PR is LGTM

rhtyd:
@wido can you comment on the IvsVifDriver removal?

wido:
@rhtyd I did? The file is not referenced anywhere. I also think that it was a typo when the file was created.

rhtyd:
@wido nope, that's not a typo it's a driver for the indigo virtual switch introduced in https://github.com/apache/cloudstack/pull/151 and it's not used anywhere because this is something one configures for the `libvirt.vif.driver=` in the agent.properties file. Kindly revert that change.

wido:
@rhtyd I see! The file is back with a few modifications.

@nvazquez I also took another look at your comments and they should be adressed now.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/IvsVifDriver.java
```diff
@@ -70,9 +71,9 @@ public void configure(Map<String, Object> params) throws ConfigurationException
         }
         _ivsIfUpPath = Script.findScript(utilScriptsDir, "qemu-ivs-ifup");
 
-        libvirtVersion = (Long) params.get("libvirtVersion");
```
nvazquez: Looks like libvirtVersion was not being used on this class, is that right?

utils/src/main/java/com/cloud/utils/net/NetUtils.java
nvazquez: Looks good

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/IvsVifDriver.java
wido: Not being used indeed. Dead variable which was never called

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

Issue #3497, opened 2019-07-16
VMs stuck in Stopping state due to NPE at org.apache.cloudstack.network.topology.BasicNetworkVisitor.visit(BasicNetworkVisitor.java:201)


server24:
Bug Report

After upgrading from 4.11.2.0 to 4.11.3.0 every operation that requires deleting an entry in a basic network seems to fail due to a NPE and therefore the VMs are stuck in Stopping state.

Possibly caused by commit b2b99ca63eecab6c0be40015ee54c37b55e07bac regarding server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkTopology.java (but this is just my guess)

CloudStack version 4.11.3.0
with Basic Networking

How to reproduce: Create a VM and then try to stop or destroy it. The procedure will not complete and the VM stays in "Stopping" state. In the CloudStack Mgm Server the following error message is shown:

```
2019-07-16 13:37:52,781 DEBUG [o.a.c.n.t.BasicNetworkTopology] (AgentManager-Handler-2:null) (logid:) REMOVING DHCP ENTRY RULE
2019-07-16 13:37:52,781 DEBUG [o.a.c.n.t.BasicNetworkTopology] (AgentManager-Handler-2:null) (logid:) Applying dhcp entry in network Ntwk[212|Guest|6]
2019-07-16 13:37:52,784 ERROR [o.a.c.f.m.MessageDispatcher] (AgentManager-Handler-2:null) (logid:) Unexpected exception when calling com.cloud.vm.ClusteredVirtualMachineManagerImpl.HandlePowerStateReport
java.lang.reflect.InvocationTargetException
        at sun.reflect.GeneratedMethodAccessor77.invoke(Unknown Source)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.apache.cloudstack.framework.messagebus.MessageDispatcher.dispatch(MessageDispatcher.java:75)
        at org.apache.cloudstack.framework.messagebus.MessageDispatcher.onPublishMessage(MessageDispatcher.java:45)
        at org.apache.cloudstack.framework.messagebus.MessageBusBase$SubscriptionNode.notifySubscribers(MessageBusBase.java:441)
        at org.apache.cloudstack.framework.messagebus.MessageBusBase.publish(MessageBusBase.java:178)
[...]

PR #3501, opened 2019-07-18
Fix stop VM issue on basic zones

nvazquez:
## Description
CloudStack reports an error while stopping/expunging VMs on basic zones

Fixes: #3497 

## Types of changes

## How Has This Been Tested?
VM lifecycle

### DIFF
--- a/server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
+++ b/server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
@@ -198,7 +198,9 @@ public class BasicNetworkVisitor extends NetworkTopologyVisitor {
         final DeployDestination destination = dhcp.getDestination();
         final boolean remove = dhcp.isRemove();
 
-        if (destination != null && router.getPodIdToDeployIn().longValue() == destination.getPod().getId()) {
+        if (destination != null && destination.getPod() != null &&
+                router != null && router.getPodIdToDeployIn() != null &&
+                router.getPodIdToDeployIn().longValue() == destination.getPod().getId()) {
             _commandSetupHelper.createDhcpEntryCommand(router, userVM, nicVo, remove, commands);
 
             return _networkGeneralHelper.sendCommandsToRouter(router, commands);
--- a/server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
+++ b/server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
@@ -198,7 +198,7 @@ public class BasicNetworkVisitor extends NetworkTopologyVisitor {
         final DeployDestination destination = dhcp.getDestination();
         final boolean remove = dhcp.isRemove();
 
-        if (router.getPodIdToDeployIn().longValue() == destination.getPod().getId()) {
+        if (destination != null && router.getPodIdToDeployIn().longValue() == destination.getPod().getId()) {
             _commandSetupHelper.createDhcpEntryCommand(router, userVM, nicVo, remove, commands);
 
             return _networkGeneralHelper.sendCommandsToRouter(router, commands);
--- a/server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
+++ b/server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
@@ -198,9 +198,9 @@ public class BasicNetworkVisitor extends NetworkTopologyVisitor {
         final DeployDestination destination = dhcp.getDestination();
         final boolean remove = dhcp.isRemove();
 
-        if (destination != null && destination.getPod() != null &&
-                router != null && router.getPodIdToDeployIn() != null &&
-                router.getPodIdToDeployIn().longValue() == destination.getPod().getId()) {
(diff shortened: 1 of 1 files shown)


nvazquez:
@blueorangutan package

server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
```diff
@@ -198,7 +198,7 @@ public boolean visit(final DhcpEntryRules dhcp) throws ResourceUnavailableExcept
         final DeployDestination destination = dhcp.getDestination();
         final boolean remove = dhcp.isRemove();
 
-        if (router.getPodIdToDeployIn().longValue() == destination.getPod().getId()) {
+        if (destination != null && router.getPodIdToDeployIn().longValue() == destination.getPod().getId()) {
```
anuragaw: Might as well be more defensive and add `router != null && router.getPodIdToDeployIn() != null`. In general being defensive is safer I think.

nvazquez: Thanks @anuragaw, added defensive checks

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

server/src/main/java/org/apache/cloudstack/network/topology/BasicNetworkVisitor.java
server24: @anuragaw @nvazquez 
This would avoid the command being executed at all, which is not the original intention. The destination is not necessary when removing something, as the DhcpRule in this case should already be deployed. So wouldn't it make more sense to change the if to:

```
if (router != null &&
            (remove || (destination != null &&
                                    destination.getPod() != null && router.getPodIdToDeployIn() != null &&
                                    router.getPodIdToDeployIn().longValue() == destination.getPod().getId()))) {
```

server24:
> Did not test it

@rhtyd @anuragaw @nvazquez 
This would avoid the command being executed at all, which is not the original intention. The destination is not necessary when removing something, as the DhcpRule in this case should already be deployed. So wouldn't it make more sense to change the if to:

```
if (router != null &&
            (remove || (destination != null &&
                              destination.getPod() != null &&
                              router.getPodIdToDeployIn() != null &&
                              router.getPodIdToDeployIn().longValue() == destination.getPod().getId()))) {
```

I have recompiled with this test and the DHCP-entries are correct in the Basic Networking (create, start, stop, destroy, expunge VM tested).

server24:
> IDK if this is expected but if you try to restart the network with clean up in a basic zone you get the 
> following exception:

this is an issue that I had already reported the 3rd of January, 2019 to the developers list, relevant excerpt:

###
...
a Network restart triggered using the UI fails with:

"Can't find all necessary running routers!"

I have looked up the source code for it and came to the following relevant snippet in

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java

starting at line 234:

```
        if (offering.isRedundantRouter() || network.isRollingRestart()) {
            expectedRouters = 2;
        }

```

From the log files I see that a rollingRestart is indeed triggered, but why does it expect two routers then? Because there might be two routers running simultaneously during the restart? If that was the case, then why in the following lines, relevant snippet (starting at line 237):

```
        if (routers == null || routers.size() < expectedRouters) {
            //we might have a router which is already deployed and running.
            //so check the no of routers in network currently.
            List<DomainRouterVO> current_routers = _routerDao.listByNetworkAndRole(network.getId(), Role.VIRTUAL_ROUTER);
            if (current_routers.size() < 2) {
                updateToFailedState(network);
                throw new ResourceUnavailableException("Can't find all necessary running routers!", DataCenter.class, network.getDataCenterId());
            }
        }

```

is throwing an Exception when current_routers.size() is smaller than 2 and not smaller than 1?

I understand that the logic is designed towards more complex networks, but will this logic not always fails for an already implemented BASIC network on KVM, with no redudant routers enabled and one VR already active? 
###

Back then nobody replied or gave feedback, glad that finally somebody is looking at this.

Generally I have the feeling that no testing is ever done for Basic Networking... is everybody just using Advanced Networking with CS?

rhtyd:
@nvazquez can you check @server24 's comment?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland:
Aren't we going to drop support for basic zones? How does this fit in that scenario? cc @GabrielBrascher @wido @rhtyd

rhtyd:
@DaanHoogland it has been discussed but a concrete plan has not been proposed. I think that's a very long term thing, till then the feature should be supported like other features.

rhtyd:
@server24 can you test the PR?

nvazquez:
@rhtyd @server24 I'll address the open comments today

nvazquez:
@server24 I have pushed a fix based on your previous comment, thanks!

Regarding network restart with cleanup, it is possible by previously changing the setting 'network.rolling.restart' to false

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test matrix

nvazquez:
@rhtyd there are still vmware and xenserver test jobs running

PR #3508, opened 2019-07-22
Enable KVM storage data motion on KVM hypervisor_capabilities

GabrielBrascher:
## Description

Considering that we have addressed a considerable effort on allowing KVM to perform storage data motion, this PR proposes updating the 'hypervisor_capabilities' table setting the 'storage_motion_supported' to '1' for KVM.

**PRs that implemented KVM storage motion features:**
- Non-managed storages
#2997 KVM VM live migration with ROOT volume on file storage type
#2983 KVM live storage migration intra cluster from NFS source and destination
- Managed storages
#2298 CLOUDSTACK-9620: Enhancements for managed storage

## Types of changes

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41200to41300.sql
@@ -386,3 +386,5 @@ CREATE TABLE `cloud`.`direct_download_certificate_host_map` (
   CONSTRAINT `fk_direct_download_certificate_host_map__certificate_id` FOREIGN KEY (`certificate_id`) REFERENCES `direct_download_certificate` (`id`) ON DELETE CASCADE
 ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
+-- KVM: enable storage data motion on KVM hypervisor_capabilities
+UPDATE `cloud`.`hypervisor_capabilities` SET `storage_motion_supported` = 1 WHERE `hypervisor_capabilities`.`hypervisor_type` = 'KVM';


rhtyd:
@GabrielBrascher Moved to 4.14 as this is not a blocker changes and in 4.14 when we'll drop support for centos6 and older kvm based distros. Storage motion was not enabled by default in 4.13 and older because it may not work on centos6 as it works with newer libvirt/distros such on centos7 with qemu-ev and with Ubuntu 16.04/18.04+.

svenvogel:
> LGTM, to move as part of the 4.13->4.14 upgrade path. As noted, we cannot accept this in general as storage motion may not necessarily work as tested/reviewed against older distros.

@rhtyd can you explain a little bit more?

rhtyd:
@svenvogel live storage motion does not work on all supported kvm distro/hosts for example on CentOS6 with a very old libvirt/qemu pkgs.

GabrielBrascher:
@rhtyd I see your point.

@svenvogel @rhtyd What do you think of adding something on documentation pointing to the need of update the table?

For instance, on the Host installation guide add something similar to [volume-snapshots](http://docs.cloudstack.apache.org/en/4.12.0.0/installguide/hypervisor/kvm.html#volume-snapshots)

And on the upgrade 4.11.x [database-preparation](http://docs.cloudstack.apache.org/en/4.12.0.0/upgrading/upgrade/upgrade-4.11.html#database-preparation) session add the database changes in case of CentOS 6 (disabling the storage motion)

rhtyd:
Thanks @GabrielBrascher sure you may send a PR, and after 4.13 RC1 is cut and master is available for merging I can work with you to get this into the 4.13->4.14 upgrade path as well as remove the support for centos6 pkging among few other deprecations as previously discussed on dev@ ML.

rhtyd:
Moved/merged in https://github.com/apache/cloudstack/pull/3591
Closing this now.

andrijapanicsb:
@rhtyd @GabrielBrascher -  storage migration does NOT work with CentOS7, neither the stock qemu/libvirt versions nor the qemu ev from the CentOS SIG repo (but it works when manually doing storage migration, via virsh --copy-all-storage in CentOS 6.4 at least)

For CentOS7, only when you use oVirt repo for qemu ev, only then it works.

In short: this works out of the box for Ubuntu only.
What I recall, this should have been added already in 4.11 or so.

PR #3278, opened 2019-04-17
systemvm: new qemu-guest-agent based patching for KVM

rhtyd:
It was reported that systemvm patching broke with qemu-ev 2.12, this fixes for that reported issue:

This introduces a new patching script for patching systemvms on KVM
using qemu-guest-agent that runs inside the systemvm on startup. This
also removes the vport device which was previously used by the legacy
patching script and instead uses the modern and new uniform guest
agent vport for host-guest communication.

Also updates the sytemvmtemplate build config to use the latest Debian
9.8.0 iso.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

PR #3524, opened 2019-07-25
Fix VR bootstrapping/connection state in KVM

svenvogel:
## Description

* After PR #3278 there was some changes on the channels. this does not seem like the problem. if we look to the VR XML file we see the following. the qemu guest agent must be the problem because the state of the VR is "disconnected".

```xml
   <channel type='unix'>
     <source mode='bind' path='/var/lib/libvirt/qemu/r-1750-VM.org.qemu.guest_agent.0'/>
     <target type='virtio' name='org.qemu.guest_agent.0' state='disconnected'/>
     <alias name='channel0'/>
     <address type='virtio-serial' controller='0' bus='0' port='1'/>
   </channel>
```
after investigation of the bootstrap process i think i found this problem. there is no startup from the qemu-guest-agent. we need them in connected state for the patch.sh script.

image

the problems seems to be the startup of the "qemu-guest-agent". since this is not an really systemd service more than an sysvinit. the service will be started with "systemctl enable --now qemu-guest agent". this will be redirected to systemd-sysv-install. normally "--now" should enable and start the service. i think with this redirection to systemd-sysv-install there is no option for the startup.

image
image

* last one is to capture the error from the module load kvm pci hotplug

Use actually master, build a SystemVM and try to deploy them

## Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/setup/bootstrap.sh
+++ b/systemvm/debian/opt/cloud/bin/setup/bootstrap.sh
@@ -85,9 +85,14 @@ config_guest() {
           sed -i "s/%/ /g" $CMDLINE
           ;;
      kvm)
-          # Configure hot-plug
-          modprobe acpiphp || true
-          modprobe pci_hotplug || true
+          # Configure kvm hotplug support
+          if grep -E 'CONFIG_HOTPLUG_PCI=y|CONFIG_HOTPLUG_PCI_ACPI=y' /boot/config-`uname -r`; then
+            log_it "acpiphp and pci_hotplug module already compiled in"
+          else
+            modprobe acpiphp 2> /dev/null && log_it "acpiphp module loaded" || true
+            modprobe pci_hotplug 2> /dev/null && log_it "pci_hotplug module loaded" || true
+          fi
+
           sed -i -e "/^s0:2345:respawn.*/d" /etc/inittab
           sed -i -e "/6:23:respawn/a\s0:2345:respawn:/sbin/getty -L 115200 ttyS0 vt102" /etc/inittab
           systemctl enable qemu-guest-agent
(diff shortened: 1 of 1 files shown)


systemvm/debian/opt/cloud/bin/setup/bootstrap.sh
```diff
@@ -85,12 +85,18 @@ config_guest() {
...
+          if grep -E 'CONFIG_HOTPLUG_PCI=y|CONFIG_HOTPLUG_PCI_ACPI=y' /boot/config-`uname -r`; then
+            log_it "acpiphp and pci_hotplug module already compiled in"
+          else
+            modprobe acpiphp 2> /dev/null && log_it "acpiphp module loaded" || true
+            modprobe pci_hotplug 2> /dev/null && log_it "pci_hotplug module loaded" || true
+          fi
+
           sed -i -e "/^s0:2345:respawn.*/d" /etc/inittab
           sed -i -e "/6:23:respawn/a\s0:2345:respawn:/sbin/getty -L 115200 ttyS0 vt102" /etc/inittab
-          systemctl enable --now qemu-guest-agent
```
rhtyd: Minor nit: The `--now` does the same, enables and starts it.

rhtyd:
@blueorangutan package

rhtyd:
@svenvogel is this work in progress or ready for review?

rhtyd:
@blueorangutan test

svenvogel:
@rhtyd yes its ready. there is nothing from me. :)

systemvm/debian/opt/cloud/bin/setup/bootstrap.sh
svenvogel: @rhtyd yes normally it should but qemu-guest-agent it a sysvinit and unfortunately with this redirect there is no startup. 

```sh
[root@kvm08 /]# /lib/systemd/systemd-sysv-install --help
systemd-sysv-install version 1.7.4 - Copyright (C) 1997-2000 Red Hat, Inc.
This may be freely redistributed under the terms of the GNU Public License.

usage:   systemd-sysv-install <enable|disable|is-enabled> [name]
```

rhtyd:
@blueorangutan test

rhtyd:
Tests lgtm. Can you review @shwstppr @anuragaw @borisstoyanov

rhtyd:
Tests LGTM

PR #3525, opened 2019-07-26
NioServer: retain links by address string to minimize resource leak

rhtyd:
Every time a client connects, the NioServer will retain the link against
the InetSocketAddress object. If the same agent/client reconnects, it
will grow older links over time and in case of denial of service attack
or a client/script/monitoring-service reconnecting aggressively against
port 8250 will cause the `_links` weak hashmap to grow over time and
very quickly.

The fix will ensure that only one Link gets weakly retained for an
incoming client based on its address string.

This is what is seen when only one kvm/agent connects to the management server:
Screenshot from 2019-07-26 16-24-05

When I restart the cloudstack-agent couple of times, I see the `_links` grow:
Screenshot from 2019-07-26 16-24-17

## Types of changes

### DIFF
--- a/utils/src/main/java/com/cloud/utils/nio/Link.java
+++ b/utils/src/main/java/com/cloud/utils/nio/Link.java
@@ -489,7 +489,7 @@ public class Link {
             try {
                 sslEngine.closeInbound();
             } catch (SSLException e) {
-                s_logger.warn("This SSL engine was forced to close inbound due to end of stream.");
+                s_logger.warn("This SSL engine was forced to close inbound due to end of stream.", e);
             }
             sslEngine.closeOutbound();
             // After closeOutbound the engine will be set to WRAP state,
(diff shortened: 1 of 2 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Caveat: for multiple-mgmt servers behind an LB, it will end up creating only one weak entry in the weak hash map. I'm not sure if it could cause any issues.

svenvogel:
@rhtyd sounds intresting. LGTM. I think when anybody use a lb maybe the transparent mode can help?

DaanHoogland:
> Caveat: for multiple-mgmt servers behind an LB, it will end up creating only one weak entry in the weak hash map. I'm not sure if it could cause any issues.

Do you mean that the two servers would share that hashmap, @rhtyd ? that seem strange to me. And if you mean on the agent side, that would only be a problem if the MSses are natted behind the same address, would it?

rhtyd:
In case of mgmt servers behind an LB, this will likely create a single entry in the weak hashmap. I'll revisit this towards 4.14 or 4.13.1; it's also possible to add a garbage collector to remove old/unreachable links from the heap.

rhtyd:
Sorry, reverting this PR because I remembered why it was put On-Hold, as the change would cause a single Link to be added to a Map when the mgmt server is behind a LB. So the changes are not optimal for all cases.

PR #3527, opened 2019-07-26
[WIP] [REFRESH] CLOUDSTACK-10106: GPU/vGPU Support on VMware

onitake:
## Description
:warning: **This is still work in progress.**

This branch implements GPU passthrough for VMware 6.0 and later hypervisors.
It is based on a previous pull request that currently has conflicts with master and hasn't been updated since September 2018.

Ticket: https://issues.apache.org/jira/browse/CLOUDSTACK-10106
Based on: https://github.com/apache/cloudstack/pull/2340

Fixes: #2340

## Types of changes

## How Has This Been Tested?

I haven't had chance to test these changes since I don't have the appropriate hardware available.

However, I'd still like to see this feature merged, with the option to add more hardware at a later point of time.


rhtyd:
is this ready @onitake ?

andrijapanicsb:
This is present in ACP 4.11 (just FYI)

svenvogel:
@andrijapanicsb thats true, but its a pick up from @onitake from @nitin-maharana Accelerite.

@nitin-maharana added newer grid cards to the config. i dont know that @onitake changed here but the text above in the introduction will explain it. so the question from @rhtyd to @onitake is this ready is correct.

same from me @onitake is it ready?

andrijapanicsb:
Unless it has been tested by someone on the real HW - it's not ready.

onitake:
It's completely untested and there has been no update since I posted it.
So no, I don't think it's ready.

#2340 hasn't received any updates either, so we're stuck with support for legacy cards only. Not very useful.
I think the whole patch needs to be redesigned from scratch, and not just around the VMware API.

svenvogel:
>#2340 hasn't received any updates either, so we're stuck with support for legacy cards only. Not very useful.
I think the whole patch needs to be redesigned from scratch, and not just around the VMware API.

@onitake  you dont have a card for testing or what do you mean with legacy card? i thought its for grid cards... 
do you see tests in the future? we would help but we dont have a card too. :(

onitake:
The patch specifically targets the older Kepler generation GRID cards, which are no longer supported by NVIDIA or hardware vendors. Newer cards may work via PCI passthrough, which seems to be supported by this patch. But I don't know for sure, because I currently don't have a system where I could test this.

On the other hand, NVIDIA has released newer data centre frameworks, such as [vComputeServer](https://www.nvidia.com/en-us/data-center/virtual-gpu-technology/), which don't seem to be covered by the patch. I'm also not sure how they integrate with VMware and other hypervisors.

In case you wan't to give it a shot yourself: I recently did a bit of research on how much a DC-grade system would cost and which hardware is suitable. The most affordable current-generation solution seems to be the Tesla T4, and you can get a 1RU Supermicro machine with four T4 cards, 2 Xeon Gold CPUs and 384GB RAM for about 20'000$. If you want to use vComputeServer, that will cost you another 50$ per GPU per year.

andrijapanicsb:
I'll move this 20.000 $ PR to 4.15 milestone - I don't see a point to merge it if nobody can test it - that means nobody can make use of it.

PR #3534, opened 2019-07-31
Misc fixes around API permissions, global settings and template UX

rhtyd:
- This allows getUploadParamsForIso for all user role types, also fixes
authorised field for getUploadParamsForTemplate API.
- Fix global setting description to say what is used when value is empty/blank.
- For VM running/allocated usage description, use parenthesis to return the instance name and ID.
- Display template download progress when template is added to a project

## Types of changes

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/query/QueryService.java
+++ b/api/src/main/java/org/apache/cloudstack/query/QueryService.java
@@ -92,7 +92,7 @@ public interface QueryService {
 
     static final ConfigKey<String> UserVMBlacklistedDetails = new ConfigKey<String>("Advanced", String.class,
             "user.vm.blacklisted.details", "rootdisksize, cpuOvercommitRatio, memoryOvercommitRatio, Message.ReservedCapacityFreed.Flag",
-            "Determines whether users can view certain VM settings", true);
+            "Determines whether users can view certain VM settings. When set to empty, default value used is: rootdisksize, cpuOvercommitRatio, memoryOvercommitRatio, Message.ReservedCapacityFreed.Flag.", true);
 
     static final ConfigKey<String> UserVMReadOnlyUIDetails = new ConfigKey<String>("Advanced", String.class,
             "user.vm.readonly.ui.details", "dataDiskController, rootDiskController",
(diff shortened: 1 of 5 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Pinging for review @shwstppr @anuragaw

api/src/main/java/org/apache/cloudstack/query/QueryService.java
```diff
@@ -92,7 +92,7 @@
 
     static final ConfigKey<String> UserVMBlacklistedDetails = new ConfigKey<String>("Advanced", String.class,
             "user.vm.blacklisted.details", "rootdisksize, cpuOvercommitRatio, memoryOvercommitRatio, Message.ReservedCapacityFreed.Flag",
-            "Determines whether users can view certain VM settings", true);
+            "Determines whether users can view certain VM settings. When set to empty, default value used is: rootdisksize, cpuOvercommitRatio, memoryOvercommitRatio, Message.ReservedCapacityFreed.Flag.", true);
```
anuragaw: Slight addition could be there optinally - "To allow all settings to be visible set this to `none`"?

rhtyd: The main issue with setting to a fake value is that, what is the none is used as a valid detail (most likely that's never going to happen but for sanity-sake). Which is why I chose to simply state that behaviour.

anuragaw: Sounds good.

borisstoyanov:
These changes seem fine, there's one problem related to uploading template and iso from local, which I'm not sure if it will be fixed in this PR? 
cc: @rhtyd

rhtyd:
I'll check/test @borisstoyanov

anuragaw:
@borisstoyanov, @rhtyd - I also see a format not matching the selected format when I upload a local macchina qcow2 template on KVM based monkey box. Investigating bit more.

rhtyd:
For the upload iso/template fix, I'll send another PR Bobby. Thanks.
Merging based on test results and lgtms.

Issue #3518, opened 2019-07-24
Unhandled NPE in VMSnapShotManagerImpl.orchestrateCreateVMSnapshot produces snapshots that cannot be deleted


onitake:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Server
~~~

##### CLOUDSTACK VERSION

~~~
at least since 4.11, still present in master
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

VMware ESX 6.5
CentOS 7

##### SUMMARY

Under certain conditions (races or similar), CloudStack may bails out in the middle of a snapshot creation process. Due to incorrect handling of this condition, a stray snapshot entry may be left in the database, even though no snapshot exists.

When trying to delete this bogus entry, another error is thrown (because the snapshot doesn't actually exist) and the entry stays in state "Expunged" indefinitely.

I have traced the source of the problem to the `return null` in https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/vm/snapshot/VMSnapshotManagerImpl.java#L516-L519

This causes a NullPointerException in https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/vm/snapshot/VMSnapshotManagerImpl.java#L1197-L1199 when `getId()` is called.

In my opinion, `orchestrateCreateVMSnapshot` should never return null, but rethrow the exception instead. This would cause the problem to be handled properly later, and an error message returned to the client.

It's unclear, however, under which circumstances `takeVMSnapshot` would throw an exception exactly, and if cleanup may be needed. In my opinion, such cleanup should happen right in `takeVMSnapshort`, not be forgotten later.

##### STEPS TO REPRODUCE

N/A, because we don't know the exact reason why the snapshot creation process may fail.

The error printed in the log is as follows:
~~~
2019-07-24 09:58:04,419 ERROR [cloud.api.ApiAsyncJobDispatcher] (API-Job-Executor-64:ctx-8c081a18 job-2196835) Unexpected exception while executing org.apache.cloudstack.api.command.user.vmsnapshot.CreateVMSnapshotCmd
java.lang.RuntimeException: Unexpected exception
       at com.cloud.vm.snapshot.VMSnapshotManagerImpl.createVMSnapshot(VMSnapshotManagerImpl.java:482)
       at sun.reflect.GeneratedMethodAccessor817.invoke(Unknown Source)
[...]

PR #3546, opened 2019-08-07
[FIX] Rethrow takeVMSnapshot() exception

onitake:
## Description

This patch changes the behaviour of `VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(Long vmId, Long vmSnapshotId, Boolean quiescevm)` to rethrow the exception when `takeVMSnapshot()` fails, instead of returning null.

If null is returned, this will lead to a NullPointerException in `VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VmWorkCreateVMSnapshot work)` in the `snapshot.getId()` call, and causes an incomplete snapshot entry to remain in the database.

Fixes: #3518

See there for more details.

## Types of changes

## How Has This Been Tested?

This change has not been tested yet.

At this point, it is unknown if it will cause unexpected side effects.

### DIFF
--- a/server/src/main/java/com/cloud/vm/snapshot/VMSnapshotManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/snapshot/VMSnapshotManagerImpl.java
@@ -515,7 +515,7 @@ public class VMSnapshotManagerImpl extends MutualExclusiveIdsManagerBase impleme
             return snapshot;
         } catch (Exception e) {
             s_logger.debug("Failed to create vm snapshot: " + vmSnapshotId, e);
-            return null;
+            throw new CloudRuntimeException("Failed to create vm snapshot: " + vmSnapshotId, e);
         }
     }
 


onitake:
`VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(Long vmId, Long vmSnapshotId, Boolean quiescevm)` is used in two places:
- `VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VmWorkCreateVMSnapshot work)` where the null value will lead to the NPE and stray DB entry
- `VMSnapshotManagerImpl.createVMSnapshot(Long vmId, Long vmSnapshotId, Boolean quiescevm)` where it is returned directly

`createVMSnapshot` in turn is used in two places:
- `CreateVMSnapshotCmd.execute()`, where the null value is checked and a different exception is thrown (`ServerApiException`)
- `StorageSystemSnapshotStrategy.takeHypervisorSnapshot(VolumeInfo volumeInfo)` where the null value is also checked and a `CloudRuntimeException` is thrown.

An alternative fix for the NPE would be to modify `VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VmWorkCreateVMSnapshot work) in the snapshot.getId()` to check for a null value before calling `getId()`.

onitake:
I suspect that `orchestrateDeleteVMSnapshot(VmWorkDeleteVMSnapshot work)` and `orchestrateDeleteAllVMSnapshots(VmWorkDeleteAllVMSnapshots work)` have similar problems.

onitake:
Sorry, scratch that last remark. These function only get a boolean return value. No risk of an NPE there.

rhtyd:
@blueorangutan package

onitake:
What do you think, @rhtyd ? Would a null value check in `VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VmWorkCreateVMSnapshot work)` be better?

Otherwise, the other null checks could be removed completely, as we would never return a null value.

rhtyd:
The only thing I would check is if throwing an exception would cause any snapshot/item in the db to be set in incorrect state, i.e. any cleanup that may be necessary when the method throws the exception.

onitake:
The problem here is that this fairly difficult, as I don't have a reliable way to trigger throwing an exception, in a real life scenario.
As I'm also new to the CloudStack code base, I don't know if there is a good way to test the behaviour in a unit test.

A test scenario should (roughly) look like this:
- prepare a test VM
- prime takeVMSnapshot so it throws an exception
- create a VM snapshot
- verify that: no NPE occurs, the error is propagated back to the user and/or log, no snapshot was created, no invalid database entry for the snapshot remains (i.e. no DB entry at all, or a DB entry in error state)

server/src/main/java/com/cloud/vm/snapshot/VMSnapshotManagerImpl.java
```diff
@@ -515,7 +515,7 @@ private VMSnapshot orchestrateCreateVMSnapshot(Long vmId, Long vmSnapshotId, Boo
             return snapshot;
         } catch (Exception e) {
             s_logger.debug("Failed to create vm snapshot: " + vmSnapshotId, e);
-            return null;
+            throw new CloudRuntimeException("Failed to create vm snapshot: " + vmSnapshotId, e);
```
DaanHoogland: I think setting the snapshot to a good state should happen here. not sure what it would be, but probably 'Error'.

DaanHoogland:
@onitake can you add setting the snapshot state before throwing the exception?

onitake:
There are other sources for exceptions, not just the part that did `return null`, so I would assume that there is already some sort of error handling and setting error state in place.
Is this not the case? I want to avoid duplicating something that's already done...

The NPE was not handled properly anywhere AFAIK, so nothing would properly set an error state.

DaanHoogland:
ok, fair point but than at this point it can also not deal with CloudRuntimeException expecting not snapshot being created in the first place or expect that no error state needs to set at that point.

onitake:
@DaanHoogland Setting the snapshot state to "Error" would involve something like this:
```java
            try {
                VMSnapshot.State.getStateMachine().transitTo(vmSnapshot, VMSnapshot.Event.OperationFailed, null, _vmSnapshotDao);
            } catch (NoTransitionException ex) {
                throw new CloudRuntimeException("Cannot transit snapshot to failed state", ex);
            }
```
I'm not sure putting this into the exception handler is a good idea.

Maybe it would be better to ensure the CloudRuntimeException is accounted for. There should already be code for setting the error state somewhere.

onitake:
So, after some more digging, I came to the conclusion that there are two major code paths how the snapshot creation can be reached:

- via `org.apache.cloudstack.storage.vmsnapshot.DefaultVMSnapshotStrategy#takeVMSnapshot`: In this case, there is already a finally clause that ensures the snapshot transitions to Error state
- via `com.cloud.vm.snapshot.VMSnapshotManagerImpl#orchestrateCreateVMSnapshot(com.cloud.vm.snapshot.VmWorkCreateVMSnapshot)`: This is where the NPE happens, and I have absolutely no idea how this method is called. There are no direct references to it, and the `@ReflectionUse` annotation indicates that something generates dynamic calls to it. @rhtyd Do you have any idea where this routine is called and how errors are handled?

server/src/main/java/com/cloud/vm/snapshot/VMSnapshotManagerImpl.java
DaanHoogland: ok, we'll have to find out in testing (or bad experience) I can not in good conscience hold this given your answers in the comment, @onitake . Let's be aware in case anything happens to the snapshot functionality

DaanHoogland:
please not that this still needs testing!

rhtyd:
Smoketests/Travis LGTM, and OK.

onitake:
@rhtyd is this only going into 4.14? No backport to 4.13?

rhtyd:
The PR was opened for master so it will go into 4.14, if you'd want in 4.13 (4.13.1.0 milestone) please send a PR against 4.13 branch @onitake

onitake:
All right, will do.

PR #3576, opened 2019-09-02
[WIP] vmware: try to find exact file when multiple files are returned

rhtyd:
On searching a file in the datastore when multiple files are found
by the datastore MO, try and loop through the result and find the file
exactly matching the requested file name.

## Types of changes

### DIFF
--- a/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/DatastoreMO.java
+++ b/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/DatastoreMO.java
@@ -403,8 +403,6 @@ public class DatastoreMO extends BaseMO {
             return rootDirectoryFilePath;
         }
 
-        String parentFolderPath;
-        String absoluteFileName = null;
         s_logger.info("Searching file " + fileName + " in " + datastorePath);
 
         HostDatastoreBrowserMO browserMo = getHostDatastoreBrowserMO();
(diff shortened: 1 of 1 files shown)


rhtyd:
cc @PaulAngus @andrijapanicsb @DaanHoogland 
This is a work-in-progress PR, the idea is to ensure that when multiple results are returned while searching a datastore, it tries to further loop and find exact-ness in the file being searched. I'm not sure if that breaks any feature for older styled VMs/volumes/snapshots etc.

vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/DatastoreMO.java
```diff
@@ -416,22 +414,39 @@ public String searchFileInSubFolders(String fileName, boolean caseInsensitive, S
...
             if (info != null && info.size() > 0) {
                 for (FileInfo fi : info) {
-                    absoluteFileName = parentFolderPath = result.getFolderPath();
-                    s_logger.info("Found file " + fileName + " in datastore at " + absoluteFileName);
-                    if (parentFolderPath.endsWith("]"))
-                        absoluteFileName += " ";
-                    absoluteFileName += fi.getPath();
-                    if(isValidCloudStackFolderPath(parentFolderPath, searchExcludedFolders)) {
+                    String possibleFilename = null;
+                    possibleFilename = parentFolderPath = result.getFolderPath();
```
DaanHoogland: should this double assignment really be here? it does not make the code clearer.

rhtyd: it's copied/refactored code (nothing new I wrote mostly), see old code line 423:
```
                    absoluteFileName = parentFolderPath = result.getFolderPath();
```

rhtyd:
@blueorangutan package

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
Can you pls give/link some example (I recall related inhouse issues...) on how to test this (i.e. example of the problem)?

rhtyd:
@andrijapanicsb I think it has to do with a VM with many number of data disks, and we perform vmotion/storage migration that changes the volume names (this was from a support issue).

rhtyd:
@andrijapanicsb  I think the issue was of deleting a volume or a snapshot or a failed snapshot? or something (sorry don't recall) where the code is written in such a way that it might find the first file matching the string instead of an exact match and could potentially remove a useful vmdk/file.
@borisstoyanov @PaulAngus @DaanHoogland @nvazquez do you remember anything?

DaanHoogland:
@andrijapanicsb @rhtyd This was noticed when Vmware would rename disk to ROOT-000001.vmdk when the hundred thousand mark was reached. The disks would no longer have a unique name and the wrong disk would be removed.

rhtyd:
Yes, I remember faintly now thanks @DaanHoogland - does that give you enough to test it @andrijapanicsb @borisstoyanov @PaulAngus ?

andrijapanicsb:
not really - I can experiment with trying to reproduce first, but that's gonna take time.

rhtyd:
@andrijapanicsb @DaanHoogland @PaulAngus @borisstoyanov - I'm not sure how we should test this, should we push to 4.15 instead if we're not sure?

DaanHoogland:
@rhtyd, this can be tested by setting in value = '999999' the record sequence with name == 'vm_instance_seq' and than create a couple of VMs with disks. It is quite elaborate (manual setup), so if we want to spend the time fine, otherwise push to 4.15.

rhtyd:
Okay I've pushed it to 4.15 for now, given we've a lot of PRs/issues to deal with. (maybe we can revisit this for a 4.14.1/* in future)

rhtyd:
@harikrishna-patnala with your VMware PR, is this still valid - that is, is there code to fuzzy search vmdk files (instead of exact search). In which case is something like this PR still valid?

rhtyd:
Closing - this may not be necessary with recent move to object/storage manager where low-level operations are not needed/performed.

PR #3582, opened 2019-09-04
systemvmtemplate: Fix Debian 9 iso url

dpassante:
## Description

Since version 10.0.0 of Debian has become stable, the URL of the Debian 9.9.0 ISO files has changed from `current` to `archive`.

The old URL returns a 404 and crash the build of systemvm templates.

Affects: 4.11, 4.13.

## Types of changes

## How Has This Been Tested?

I built a systemvm template without any errors with this change.

### DIFF
--- a/tools/appliance/systemvmtemplate/template.json
+++ b/tools/appliance/systemvmtemplate/template.json
@@ -38,7 +38,7 @@
       "disk_interface": "virtio",
       "net_device": "virtio-net",
 
-      "iso_url": "https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/debian-9.9.0-amd64-netinst.iso",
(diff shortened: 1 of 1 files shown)


GabrielBrascher:
@dpassante thanks for this PR. Can you please target this change to branch 4.13?
The current LTS is (will be soon released) 4.13.0.0. Or does it make sense add this to 4.11 and then push forward from 4.11 :arrow_right: 4.13 :arrow_right: master @wido @PaulAngus @rhtyd?

svenvogel:
it always the same if the iso will change. i think we have any of this PRs in the past. its lesser a bug more than a normal change :)

dpassante:
@GabrielBrascher Done !
If the 4.13 will be soon released as the new TLS, it probably doesn't make sense to add this small fix to 4.11.

dpassante:
The CI ran on 4.11 branch when I rebased the PR. I probably should have changed the target branch before rebasing.
How can we get CI run again on 4.13 branch?

Issue #3195, opened 2019-02-26
Python 3 compatibility


wido:
A lot of Python code inside CloudStack is still Python 2 and we have just 10 months left before Python 2 is no longer supported: https://pythonclock.org/

Various items need to be looked at and checked for Python 3 compatibility. A list, but probably not complete:

- [x] KVM Security Groups
- [x] Xen Security Groups
- [ ] Virtual Router scripts (addressed in #3730)
- [x] Various setup tools
- [x] build tools (addressed in #4030)
- [x] marvin framework and smoke tests (#4727)
- [x] marvin component test sets (#5082)
- [x] packaging (cloud.spec) (#4513 is asking for it)

These are isolated projects and can usually be taken care of separately and also properly tested.

Creating this ticket so it can be referred to and also this is out there.

rhtyd:
Is there an argument to say, these may be implemented purely in Java (if possible)? Currently, many of these scripts esp on KVM are forked via the cloudstack-agent and called.
For VR, I played with GRPC and I think it can be secured against the current CA-framework introduced in 4.11; here's a PoC: https://github.com/shapeblue/cloudstack/pull/44
I'll propose more details on dev@ once I can put together design/idea details.

falcon78921:
@rhytd @wido I would love to help with moving over to Python 3, if that's still the plan.

wido:
Although I agree with @rhtyd that a lot can be done in Java, it's not always the easiest. Python is sometimes a lot easier to develop and talk directly with the system for executing commands.

We haven't started on this @falcon78921 , but any help is welcome! Just pick a Python script and start working on it.

rhtyd:
@wido changing the env dependency is a big step, python3.x is not available on all distros by default. I think it's better to discuss on dev@ for future steps. In my proposal, for the VR for example I've proposed that we move to a lightweight agent and as PoC I've shared a Go-Java approach based on grpc.

I think the best time would be to discuss/start doing things after 4.13, as we've decided to deprecate CentOS6 after 4.13.

wido:
@rhtyd Python 3 is available on ALL distros. It has been for a very long time. It's on RHEL, CentOS, Debian and Ubuntu. No need to install anything :)

We can always talk about this, but the fact is that per January Python 2 is EOL and it's not so super difficult to write things in Python 3.

I'm not against an Agent in the VR for example, sounds like a good thing!

PaulAngus:
[...]

PR #3589, opened 2019-09-09
kvm/security_group: Make Security Group Python 3 compatible

wido:
This script only runs on the KVM Hypervisors and these all support
Python 3.

As Python 2 is deprecated at the end of 2019 we need to fix these
scripts to work under Python 3.

CentOS 7, 8 and Ubuntu 16.04 and 18.04 all have Python 3 installed
by default.

Ubuntu 20.04 will no longer have Python 2 installed and therefor
this script needs to be modified to work with Python 3.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,13 +3,13 @@ Section: libs
 Priority: extra
 Maintainer: Wido den Hollander <wido@widodh.nl>
 Build-Depends: debhelper (>= 9), openjdk-8-jdk | java8-sdk | java8-jdk | openjdk-9-jdk, genisoimage,
- python-mysql.connector, maven (>= 3) | maven3, python (>= 2.7), lsb-release, dh-systemd, python-setuptools
+ python-mysql.connector, maven (>= 3) | maven3, python (>= 2.7), python3 (>= 3), lsb-release, dh-systemd, python-setuptools
 Standards-Version: 3.8.1
 Homepage: http://www.cloudstack.org/
 
 Package: cloudstack-common
 Architecture: all
-Depends: ${misc:Depends}, ${python:Depends}, genisoimage, nfs-common, python-netaddr
+Depends: ${misc:Depends}, ${python:Depends}, genisoimage, nfs-common
 Conflicts: cloud-scripts, cloud-utils, cloud-system-iso, cloud-console-proxy, cloud-daemonize, cloud-deps, cloud-python, cloud-setup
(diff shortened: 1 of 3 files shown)


GabrielBrascher:
Just adding, for the record, that this is the first PR addressing the effort raised on issue #3195 
Thanks for the PR @wido! I will be reviewing it and re-testing it.

rhtyd:
@wido @GabrielBrascher can you add dependency of `python3` in `packaging/centos7/cloud.spec` to cloudstack-common and cloudstack-agent packages?

debian/control
```diff
@@ -3,13 +3,13 @@ Section: libs
...
 Maintainer: Wido den Hollander <wido@widodh.nl>
 Build-Depends: debhelper (>= 9), openjdk-8-jdk | java8-sdk | java8-jdk | openjdk-9-jdk, genisoimage,
- python-mysql.connector, maven (>= 3) | maven3, python (>= 2.7), lsb-release, dh-systemd, python-setuptools
+ python-mysql.connector, maven (>= 3) | maven3, python (>= 2.7), python3 (>= 3), lsb-release, dh-systemd, python-setuptools
 Standards-Version: 3.8.1
 Homepage: http://www.cloudstack.org/
 
 Package: cloudstack-common
 Architecture: all
-Depends: ${misc:Depends}, ${python:Depends}, genisoimage, nfs-common, python-netaddr
```
rhtyd: @wido  @GabrielBrascher Do we no longer need python-netaddr? (for all/any py scripts?)

packaging/centos7/cloud.spec
```diff
@@ -54,7 +54,7 @@ intelligent IaaS cloud implementation.
 %package management
 Summary:   CloudStack management server UI
 Requires: java-1.8.0-openjdk
-Requires: python
+Requires: python3
```
rhtyd: We probably want both python and python3 like the debian/control change @GabrielBrascher, please add a new Requires line

rhtyd:
Thanks @GabrielBrascher left some minor remarks

packaging/centos7/cloud.spec
GabrielBrascher: @rhtyd ah sure. Updated it, thanks!

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3597, opened 2019-09-13
kvm: Logrotate should not touch agent.log

wido:
Logrotate should only touch security_group.log and resizevolume.log
as the agent.log is already rotated by log4j inside the Agent.

Having two systems trying to rotate agent.log leads to all kinds of
issues like having binary (compressed) data in the middle of a plain-text
log file.

### DIFF
--- a/agent/conf/cloudstack-agent.logrotate.in
+++ b/agent/conf/cloudstack-agent.logrotate.in
@@ -15,12 +15,11 @@
 # specific language governing permissions and limitations
 # under the License.
 
-@AGENTLOG@
-/var/log/cloudstack/agent/security_group.log
-{
+/var/log/cloudstack/agent/security_group.log /var/log/cloudstack/agent/resizevolume.log {
     copytruncate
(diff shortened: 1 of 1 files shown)


rhtyd:
(el6 failure expected)
@blueorangutan package

rhtyd:
Config and packaging only change, wont' need smoketests.

onitake:
Isn't the commit message a bit misleading?
Aside from removing agent.log from the logrotate config, it also added /var/log/cloudstack/agent/resizevolume.log - was this part of  @AGENTLOG@ before?

ustcweizhou:
@onitake no, resizevolume.log is new in logrotate
I think it would be good to add security_group.log to logrotate.

onitake:
Ah, ok.
Just thought it may be good if this is mentioned somewhere.

ustcweizhou:
@rhtyd can you cherry-pick this commit to 4.13 branch ?

PR #3604, opened 2019-09-23
Fix Policy Based Routing for private gateway static routes

DennisKonrad:
## Description

We noticed that VMs that are target of a static nat cannot use private gateways.

Investigation showed that these VMs behave differently than all other VMs in a VPC because their whole traffic gets marked and is only allowed to leave through the public ip provided by the static nat.

Could be related to: https://github.com/apache/cloudstack/pull/3366
Maybe @richardlawley @rhtyd can contribute if this is an unwanted side-effect.

To not break anything we propose a solution that does only mark the nat portion of the traffic and allow the VM use the private gateway like normal.

## Types of changes

## How Has This Been Tested?
We built and manually tested this to check if function is restored and nothing else breaks.

### DIFF
--- a/systemvm/debian/opt/cloud/bin/configure.py
+++ b/systemvm/debian/opt/cloud/bin/configure.py
@@ -931,10 +931,10 @@ class CsForwardingRules(CsDataBag):
 
         self.fw.append(["mangle", "front",
                         "-A PREROUTING -s %s/32 -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" %
-                        rule["internal_ip"]])
+                        rule["public_ip"]])
         self.fw.append(["mangle", "front",
                         "-A PREROUTING -s %s/32 -m state --state NEW -j MARK --set-xmark %s/0xffffffff" %
-                        (rule["internal_ip"], hex(100 + int(device[len("eth"):])))])
+                        (rule["public_ip"], hex(100 + int(device[len("eth"):])))])
         self.fw.append(["nat", "front",
                         "-A PREROUTING -d %s/32 -j DNAT --to-destination %s" % (rule["public_ip"], rule["internal_ip"])])
         self.fw.append(["nat", "front",
--- a/systemvm/debian/opt/cloud/bin/configure.py
+++ b/systemvm/debian/opt/cloud/bin/configure.py
@@ -930,10 +930,10 @@ class CsForwardingRules(CsDataBag):
             raise Exception("Ip address %s has no device in the ips databag" % rule["public_ip"])
 
         self.fw.append(["mangle", "front",
-                        "-A PREROUTING -s %s/32 -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" %
+                        "-A PREROUTING -d %s/32 -m state --state NEW -j CONNMARK --save-mark --nfmask 0xffffffff --ctmask 0xffffffff" %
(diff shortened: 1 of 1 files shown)


richardlawley:
I doubt this was introduced in #3366 - all I did there was change to the proper way of adding rules to the front.  Before this change, rules were still added to the front, but this happened multiple times as the code couldn't detect it as a duplicate rule.

DennisKonrad:
Thanks you for helping to find the source of this issue @richardlawley. 
We noticed the behaviour of the nat'ed VM changed in the passed year somehow. Our systemvm where the private gatways were working (for the natted vm) was about a year old. Maybe something else changed that I did not find/see?

Or could the issue solved by #3366 have caused the packet mark not to work somehow?

richardlawley:
I'm not sure. I made a few changes around the same time for static nat, especially when there are multiple public networks, but some of these were in code branches which don't get hit on a vpc.

Unfortunately I'm only running 4.11.2 in production, but have all of my fixes ported into our systemvm iso. If I get a chance I'll see if I can reproduce the problem.

DennisKonrad:
That would be a great help. I think most cs users don't use the private gateways and maybe this nat issue was not noticed?

What do you think of the proposed changes? Do you think it breaks functionality?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DennisKonrad:
@rhtyd I looked into the failed test and I think it's failed because of:
`deletenetworkoffering failed, due to: errorCode: 431, errorText:Can't delete network offering 75 as its used by 1 networks. To make the network offering unavaiable, disable it`

It seems this is not caused by the changes in this PR. Can you confirm?

andrijapanicsb:
@blueorangutan test

DaanHoogland:
should this go on 4.13 as well, @andrijapanicsb @DennisKonrad ?

andrijapanicsb:
Makes sense @DaanHoogland

andrijapanicsb:
@DennisKonrad I would like to test this manually.
Can you please give EXACT explanation on the setup to reproduce the issue (create this, create that, etc....)

Thanks!

DennisKonrad:
@andrijapanicsb I'll give you the exact case the problem can be reproduced. In short:

If you use VPC with private Gateways and static NAT, the one VM that the NAT points to isn't able to use the private gateway anymore.
The private gateway is used to connect to an outside network (with static routes).

This worked up to a change that I cannot pin down exactly and after we updated the NATed VM lost connection to the machines behind the private gateway. Therefor the fix

rhtyd:
@DennisKonrad can you change the PR base to 4.13 and rebase against origin/4.13?

DaanHoogland:
> @DennisKonrad can you change the PR base to 4.13 and rebase against origin/4.13?

and mark milestone 4.13.1, please?

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DennisKonrad:
@DaanHoogland @rhtyd After a lot of pain: 
Rebased to origin/4.13
Changed PR base to 4.13
Changed Milestone to 4.13

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
tests look good, second pair of eyes, @andrijapanicsb @wido @rhtyd @weizhouapache ?

DaanHoogland:
tnx guys

PR #3351, opened 2019-05-24
Have persistent DHCP leases file on VRs and cleanup /etc/hosts on VM deletion

nvazquez:
## Description

Since the CloudStack virtual router was redesigned on version 4.6 it has been observed that the DHCP leases file is not persistent across network operations. This causes conflicts on guest VMs static IPs, causing these static IPs to not be renewed by the DHCP server running on isolated and VPC networks' virtual routers (dnsmasq). On stopping or destroying a VM, its dhcp/dns records are not removed from the virtual router causing ghost effects.

Fixes #3272 
Fixes #3354 

## Types of changes
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?
Tested on local environment, 1x KVM host, Isolated network
- Create VMs on isolated network
- Restart network

PR #3608, opened 2019-09-25
server: Cleanup dhcp and dns entries only on expunging VM

rhtyd:
This fixes a behaviour to not cleanup DHCP and DNS rules for NICs of a
VM in the VR when it is stopped, but instead when VM is expunged because
stopped VMs in CloudStack still retain the IPs and records.

## Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -40,8 +40,6 @@ import java.util.stream.Stream;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
FYI - this partially reverts and refactors changes introduced in #3351 to cleanup VM dhcp/dns rules in the VR only when a VM is expunged. A VM hostname and IP address is used as constraint and checks even of a VM is stopped, i.e. you cannot deploy a VM with the same hostname and IP in a network with an existing VM with those attributes. This was tested against a local monkeybox KVM env and regression tested agains Trillian/kvm.

PR #3611, opened 2019-09-26
[WIP] Support for CentOS8

rhtyd:
This adds support for CentOS8 packaging. The first would be to support CentOS8 as KVM host, and then as mgmt+usage server host.

Tasks:
- [ ] Create BO/docker container for packaging
- [ ] Test/support packaging for management and usage server, on CentOS8 host
- [ ] Test/support packaging for KVM agent, on CentOS8 KVM host
- [x] Depends on #3601 (jdk changes, mysql dependency support/version)

## Types of changes

### DIFF
--- a/packaging/centos8/cloud.limits
+++ b/packaging/centos8/cloud.limits
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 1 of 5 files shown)


rhtyd:
@andrijapanicsb cc @DaanHoogland @PaulAngus @borisstoyanov - we won't have time, moved to the next 4.15 milestone.

NuxRo:
It would be interesting to see how this goes as CentOS8 (like Ubuntu 20.04) has switched from iptables to nftables, though the command "iptables" is still present, it's just using nftables underneath. This might impact security group functionality.

If this is the case, it might be a good occasion to check whether SG functionality can be moved entirely into libvirt which supports (I think).

https://libvirt.org/firewall.html

rhtyd:
Closed in favour of the new PR.

PR #3616, opened 2019-10-03
Reduce verbosity of Async Job Manager log messages

richardlawley:
## Description

On a Cloudstack instance with log verbosity set to INFO, there are 2 messages generated every 10 seconds about cleaning up expired async jobs.  This makes the log messages much noisier than they need to be.

image

This PR reduces that message to trace level, since there are actually useful messages at INFO level in that class.

## Types of changes

## How Has This Been Tested?

Check compilation and tests


rhtyd:
@blueorangutan package

rhtyd:
Logging change, explicit full regression test not needed. Test against simulator/Travis is OK.

PR #3623, opened 2019-10-07
kvm: Use 'ip' instead of 'brctl'

wido:
The command 'brctl' is deprecated and should no longer be used.

iproute2 supports all the features we need and therefor we should use
this instead of the old commands.

Feature wise this does not change anything. It just makes the code more
robust towards the future.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
@@ -273,12 +273,12 @@ public class BridgeVifDriver extends VifDriverBase {
 
     @Override
     public void attach(LibvirtVMDef.InterfaceDef iface) {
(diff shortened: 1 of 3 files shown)


ustcweizhou:
LGTM
tested the commands on ubuntu 16

RPDiep:
Seems good to me, except this won't run under CentOS 6 according to the manpages of iproute (running currently iproute 2.6.32).
Backporting this should therefore not be done.

rhtyd:
@wido can you also fix `scripts/vm/network/vnet/modifyvlan.sh`?
With CentOS6 support removed from master/4.14, we should try and move all network related commands to use `iproute2` instead of ifconfig/brctl etc.

wido:
Done @rhtyd

rhtyd:
Nice, thanks.

@blueorangutan package

rhtyd:
@blueorangutan test

GabrielBrascher:
This one looks ready to merge.
@rhtyd Is there any issue with the **blueorangutan** test?

rhtyd:
Something went wrong at the backend, I'll rekick test.
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3635, opened 2019-10-15
server: acquire IPv4 address when add secondary IP to nic if IP is not specified

ustcweizhou:
## Description

After commit fbf488497fb863c13fc0908281e3f4f86906df43, admin need to specify an ipv4 or ipv6 addresses when add IP to nic which breaks backward compatibity. If IP is not specified, a IPv4 address should be returned.

## Types of changes

## How Has This Been Tested?

When I add an IP to a nic and do not specify an IP, 

Before this PR, it gives an error "Allocating ip to guest nic 111 failed"
After this PR, it succeeds.

### DIFF
--- a/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
+++ b/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
@@ -732,11 +732,10 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService {
             }
 
             try {
-                if (ipv4Address != null) {
-                    ipaddr = _ipAddrMgr.allocatePublicIpForGuestNic(network, podId, ipOwner, ipv4Address);
-                }
                 if (ipv6Address != null) {
                     ip6addr = ipv6AddrMgr.allocatePublicIp6ForGuestNic(network, podId, ipOwner, ipv6Address);
+                } else {
+                    ipaddr = _ipAddrMgr.allocatePublicIpForGuestNic(network, podId, ipOwner, ipv4Address);
                 }
                 if (ipaddr == null && ipv6Address == null) {
                     throw new InvalidParameterValueException("Allocating ip to guest nic " + nicId + " failed");


ustcweizhou:
@wido @GabrielBrascher @rhtyd 
can you please review and test it ?

rhtyd:
@blueorangutan package

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -732,11 +732,10 @@ public NicSecondaryIp allocateSecondaryGuestIP(final long nicId, IpAddresses req
             }
 
             try {
-                if (ipv4Address != null) {
-                    ipaddr = _ipAddrMgr.allocatePublicIpForGuestNic(network, podId, ipOwner, ipv4Address);
-                }
                 if (ipv6Address != null) {
                     ip6addr = ipv6AddrMgr.allocatePublicIp6ForGuestNic(network, podId, ipOwner, ipv6Address);
+                } else {
```
DaanHoogland: I'm a bit confused, the null check for ipv4address is gone and it is not assigned (even when not null) when it isn't but ipv6Address is not null. This means that if both are available ipv6 is the *only one* used. Is that what you intend @ustcweizhou ?
I see how this fixes backwards compatibility btw, just wondering if we would want to allow dualstack!?

weizhouapache: @DaanHoogland 
the method is called only when add a nic to vm.
https://github.com/apache/cloudstack/blob/master/api/src/main/java/org/apache/cloudstack/api/command/user/vm/AddIpToVmNicCmd.java#L159-L162

at most one ip (ipv4 or ipv6) is passed.

the idea of this PR is, if ipv6 is passed, check and assign an ipv6 address. It ipv4 is passed or no ip is passed, then assign an ipv4 address.

DaanHoogland: 👍

DaanHoogland:
@blueorangutan test

PR #3627, opened 2019-10-08
[do not merge] server: Do NOT cleanup dhcp and dns when stop a vm

ustcweizhou:
## Description

According comment in PR #3608, dhcp and dns entries are cleaned up only when a VM is expunged.
Revert part of commit 8fb388e9312b917a8f36c7d7e3f45985a95ce773.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

PR #3662, opened 2019-11-03
Increase DHCP lease time to infinite

andrijapanicsb:
Lease time set to effectively infinite (36000+ days) since we fully control VM lifecycle via CloudStack.
Infinite time helps avoid some edge cases which could cause DHCPNAK being sent to VMs since
(RHEL) system lose routes when they receive DHCPNAK.
When VM is expunged, its active lease and DHCP/DNS config is properly removed (#3627 and #3608) from related files in VR, so the infinite duration of lease does not cause any issues or garbage.
There will be soon a PR which also regenerates the /var/lib/misc/dnsmasq.leases (active lease DB file) in the new VR (when restarting network with cleanup), which will help around RHEL edge cases (described above) for the VMs who are already running in productions systems with 30d lease time.

## Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
@@ -160,10 +160,14 @@ class CsDhcp(CsDataBag):
 
     def add(self, entry):
         self.add_host(entry['ipv4_address'], entry['host_name'])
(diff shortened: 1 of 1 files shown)


andrijapanicsb:
CC @ustcweizhou (your username/name does NOT appear when requesting review...)

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan test

PR #3669, opened 2019-11-07
server: Fix resource count of primary storage/volume because of Expunged volumes

rakgenius:
## Description

If the volume of a VM is in "Expunged" state then it should not be
considered towards total resource count of "primarystoragetotal"
field of a domain.\
Currently, CloudStack takes into resource calculation even if the
volume is expunged. The volume itself doesn't exist in primary
storage and hence it should not be considered towards resource
calculation.

Steps to reproduce the issue:

1 . Get the resource count of "primarystoragetotal" of a particular domain.
2 . Create a VM with 5GB root disk size and stop it.
3 . Now the value of "primarystoragetotal" should be intitial value plus 5.
4 . Navigate to "volumes" of the VM and select "Download Volume" option.
5 . Once the volume is downloaded, expunge the VM.
6 . Get the resource count of "primarystoragetotal". it will be same value as in step 3
    But it should be same as initial value obtained in step 1.

## Types of changes

## How Has This Been Tested?
This has been tested using the cloudmonkey api commands.

```
list domains id=<ID> filter=primarystorageavailable,primarystoragelimit,primarystoragetotal
```

We can also run ```update resourcecount domainid=<ID>``` to update the resource before running the above command.

### DIFF
--- a/engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
+++ b/engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
@@ -122,7 +122,7 @@ public class VolumeDaoImpl extends GenericDaoBase<VolumeVO, Long> implements Vol
     public List<VolumeVO> findByPoolId(long poolId) {
         SearchCriteria<VolumeVO> sc = AllFieldsSearch.create();
         sc.setParameters("poolId", poolId);
-        sc.setParameters("notDestroyed", Volume.State.Destroy);
+        sc.setParameters("notDestroyed", Volume.State.Destroy, Volume.State.Expunged);
         sc.setParameters("vType", Volume.Type.ROOT.toString());
         return listBy(sc);
     }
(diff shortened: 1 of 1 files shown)


weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan help

DaanHoogland:
@rakgenius @weizhouapache makes perfect sense. Have you looked at the usage server and billing verification? This could be needed if events don't get logged for the volumes.

weizhouapache:
@rhtyd @DaanHoogland @andrijapanicsb can you kick off a test ?

by the way, is it possible to grant me access to kick a test ?

DaanHoogland:
@weizhouapache I am afraid not, These tests are run in our internal lab and we need to keep small degree of control on how busy we let it get.
@blueorangutan test

weizhouapache:
The PR has two approvals , and BO tests passed.
is it ready for merge ?

DaanHoogland:
@weizhouapache I am looking at the travis run; once again a maximum log length error. I restarted it just to see. I am awaiting it but I think we are good to go.

Issue #3602, opened 2019-09-20
Using custom protocol in ACL rule, breaks ACL


andrijapanicsb:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
mgmt, ACL rules handling
~~~

##### CLOUDSTACK VERSION

~~~
tested 4.11.2.0, 4.13.0.0
~~~

##### CONFIGURATION

VPC, custom ACL with a few (ingress) rules applied to a network

##### OS / ENVIRONMENT

NA

##### SUMMARY

Adding an ACL rule with a custom **protocol number** (i.e. protocol 50, 51 or any other) - will break ACL/iptables chain inside the VR in following way
- all the rules BEFORE this one (rule number lower than the custom protocol rule number) will get properly applied
- this specific rule (custom protocol number) will NOT be injected at all inside the VR (seems it is not sent by mgmt server to the VR ) 
- all the rules that come AFTER this one, will be injected only AFTER the default DENY rule is injected

##### STEPS TO REPRODUCE

Create a VPC with custom ACL on the network i.e.:
![image](https://user-images.githubusercontent.com/45762285/65315803-06ad2300-db99-11e9-980b-184eab9468e0.png)

Confirm rules are properly applied inside the VR:
```
root@r-47-VM:~# iptables-save | grep INBOUND_eth2
:ACL_INBOUND_eth2 - [0:0]
-A FORWARD -d 10.10.10.0/24 -o eth2 -j ACL_INBOUND_eth2
-A ACL_INBOUND_eth2 -d 225.0.0.50/32 -j ACCEPT
-A ACL_INBOUND_eth2 -d 224.0.0.18/32 -j ACCEPT
-A ACL_INBOUND_eth2 -s 6.6.6.6/32 -p tcp -m tcp --dport 666 -j ACCEPT
-A ACL_INBOUND_eth2 -s 2.2.2.2/32 -p tcp -m tcp --dport 222 -j ACCEPT
-A ACL_INBOUND_eth2 -s 3.3.3.3/32 -p tcp -m tcp --dport 333 -j ACCEPT
-A ACL_INBOUND_eth2 -j DROP
```

Now, add a custom rule in the middle of the rules list (rule number 31 in the example below), as follows:

![image](https://user-images.githubusercontent.com/45762285/65315955-4d9b1880-db99-11e9-8f51-56b12b6fb087.png)

Again examine the rules in the VR:

```
[...]

PR #3678, opened 2019-11-12
vpc: fix acl rule with protocol number is not applied correctly in vpc vr

ustcweizhou:
## Description

When add a acl rule with protocol number, the iptables rules in vpc vr is not applied correctly.
for example, when add an ingress acl rule (protocol number:50, cidr: 2.2.2.2/32), we expect to have a iptables rule: "-A ACL_INBOUND_eth2 -s 2.2.2.2/32 -p esp -j ACCEPT"
the actual rule is "-A ACL_INBOUND_eth2 -j DROP"

It is because the rules in json are not correct.
network_acl.json.a8c52dca-0278-4e1c-b72b-987ca7121f4f.gz:{"device":"eth2","mac_address":"02:00:7d:27:00:02","private_gateway_acl":false,"nic_ip":"192.168.11.12","nic_netmask":"28","ingress_rules":[{"type":"protocol","protocol":50,"cidr":"ACCEPT","allowed":false},{"type":"all","cidr":"0.0.0.0/0","allowed":true},],"egress_rules":[],"type":"networkacl"}

Fixes: #3602 

## Types of changes

## How Has This Been Tested?

Add acl rule in VPC, then check the iptables rules in VPC VR. all are good with this change.

### DIFF
--- a/core/src/main/java/com/cloud/agent/resource/virtualnetwork/facade/SetNetworkAclConfigItem.java
+++ b/core/src/main/java/com/cloud/agent/resource/virtualnetwork/facade/SetNetworkAclConfigItem.java
@@ -79,7 +79,7 @@ public class SetNetworkAclConfigItem extends AbstractConfigItemFacade {
                 // If we check the size of the array, it will fail to setup the network.
                 // So, let's catch the exception and continue in the loop.
                 try {
-                    aclRule = new ProtocolAclRule(ruleParts[5], false, Integer.parseInt(ruleParts[1]));
+                    aclRule = new ProtocolAclRule(ruleParts[4], "ACCEPT".equals(ruleParts[5]), Integer.parseInt(ruleParts[1]));
                 } catch (final Exception e) {
                     s_logger.warn("Problem occured when reading the entries in the ruleParts array. Actual array size is '" + ruleParts.length + "', but trying to read from index 5.");
                     continue;


andrijapanicsb:
@blueorangutan package

weizhouapache:
@blueorangutan test

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
@weizhouapache seems that blueorangutan doesn't honour your request, will need to kick his bash :) 
thx for the fix, I'm testing manually as we speak.

andrijapanicsb:
LGTM

tested manually, results below.

Setup:
image

Results as expected:
image

andrijapanicsb:
Actually, by checking some strongswan documentation, that is how rules should be added (no PORT definition in the rule) - so that is all OK.

weizhouapache:
@andrijapanicsb thanks for testing !

dport is not supported by some protocols , for example ah/esp/gre (50/51/47)
According to iptables document (https://linux.die.net/man/8/iptables) , dport is supported by some other protocols  besides tcp/udp , for example dccp, sctp. The protocols are rarely used.

weizhouapache:
@andrijapanicsb 
I have tested in a virtual router, it gives the same result, only tcp/udp/dccp/sctp support dport.

```
root@r-597-VM:~# for i in `seq 0 142`;do iptables -I ACL_INBOUND_eth2 -p $i --dport 1000 -j ACCEPT >/dev/null 2>&1;if [ "$?" == "0" ];then echo $i;fi;done
6
17
33
132
```
https://www.iana.org/assignments/protocol-numbers/protocol-numbers.xhtml

andrijapanicsb:
yes @weizhouapache that's what I mean, that's fine.

weizhouapache:
@andrijapanicsb nice. 
maybe we can take dccp/sctp into consideration afterwards.
They are not in use on our platforms.

andrijapanicsb:
your call @weizhouapache  (for now, I'm fine with just this fix here). Thx

weizhouapache:
@rhtyd @DaanHoogland can you please review it ?

PR #3695, opened 2019-11-19
debian: fix symlink issue post install/upgrade

andrijapanicsb:
## Types of changes

- [x ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

About to be tested!

### DIFF
--- a/debian/cloudstack-usage.postinst
+++ b/debian/cloudstack-usage.postinst
@@ -21,13 +21,8 @@ set -e
 
 case "$1" in
     configure)
-        # We copy the db.properties file from the management server to the usage server
-        # This used to be a symlink, but we don't do that anymore
-        if [ -f "/etc/cloud/management/db.properties" ]; then
-            cp -a /etc/cloud/management/db.properties /etc/cloudstack/usage/db.properties
-        fi
-
-        # Replacing db.properties with management server db.properties
+     
+        # Linking usage server db.properties to management server db.properties
         if [ -f "/etc/cloudstack/management/db.properties" ]; then
             rm -rf /etc/cloudstack/usage/db.properties
             ln -s /etc/cloudstack/management/db.properties /etc/cloudstack/usage/db.properties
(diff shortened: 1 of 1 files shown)


andrijapanicsb:
@blueorangutan package

rhtyd:
@blueorangutan package

debian/cloudstack-usage.postinst
```diff
@@ -21,15 +21,10 @@ set -e
...
-        if [ -f "/etc/cloud/management/db.properties" ]; then
-            cp -a /etc/cloud/management/db.properties /etc/cloudstack/usage/db.properties
-        fi
-
-        # Replacing db.properties with management server db.properties
+     
+        # Linking usage server db.properties to management server db.properties
         if [ -f "/etc/cloudstack/management/db.properties" ]; then
-            rm -rf /etc/cloudstack/usage/db.properties
+            rm -f /etc/cloudstack/usage/db.properties
```
DaanHoogland: this would work, but wouldn't ```ln -sf ...``` as well?

debian/cloudstack-usage.postinst
```diff
@@ -40,7 +35,7 @@ case "$1" in
 
         # Replacing key with management server key
         if [ -f "/etc/cloudstack/management/key" ]; then
-            rm -rf /etc/cloudstack/usage/key
+            rm -f /etc/cloudstack/usage/key
             ln -s /etc/cloudstack/management/key /etc/cloudstack/usage/key
```
DaanHoogland: same as above

rhtyd: Yes. I'm not sure if the file being removed may not be a symlink, this approach should work as intended.

andrijapanicsb: KISS, as far as I'm concerned :)

andrijapanicsb:
I've manually tested this one, and it works fine as long as we respect the expectation that the usage server is installed once both the db.properties and the key files are present in the mgmt server folder.

PR #3701, opened 2019-11-21
security_group.py: check cidr unstrictly to accept cidrs like 1.1.1.1/24

ustcweizhou:
## Description

When I add a security group rule with cidr like 1.1.1.1/24, the rule is not applied on kvm hypervisor.
Ths issue does not exist in 4.13.0.0 and previous versions.

## Types of changes

## How Has This Been Tested?

It has been tested in advanced zone with security group.
It can be verified very easily in python3, see results below.

```
# python3
Python 3.6.7 (default, Oct 22 2018, 11:32:17)
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.

>>> import ipaddress
>>> ipaddress.ip_network("1.1.1.1/28")
IPv4Network('1.1.1.1/28')

>>> ipaddress.ip_network("1.1.1.1/24")
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/ipaddress.py", line 74, in ip_network
    return IPv4Network(address, strict)
  File "/usr/lib/python3.6/ipaddress.py", line 1519, in __init__
    raise ValueError('%s has host bits set' % self)
ValueError: 1.1.1.1/24 has host bits set

>>> ipaddress.ip_network("1.1.1.1/24", False)
IPv4Network('1.1.1.0/24')
```

### DIFF
--- a/scripts/vm/network/security_group.py
+++ b/scripts/vm/network/security_group.py
@@ -967,7 +967,7 @@ def parse_network_rules(rules):
         ipv6 = []
         for ip in cidrs.split(","):
             try:
-                network = ipaddress.ip_network(ip)
+                network = ipaddress.ip_network(ip, False)
                 if network.version == 4:
                     ipv4.append(ip)
                 else:


wido:
@rhtyd Yes, we have that in the RPM/DEB dependencies.

Python 3 is however also present by default on both CentOS and Ubuntu.

rhtyd:
Manually tested against python3, BO doesn't run test against SG env so I'll merge this based on reviews and manual testing.

PR #3704, opened 2019-11-22
utils: use iproute to get default network interface

rhtyd:
This uses `ip route` than `route` without an explicit path (/bin, /sbin
etc.) to query the default network interface on the system.

## Types of changes

Tested on Ubuntu:
> ip route | grep default | head -1 | grep -Po '(?<=dev )(\S+)'
wlp0s20f3

### DIFF
--- a/utils/src/main/java/com/cloud/utils/net/NetUtils.java
+++ b/utils/src/main/java/com/cloud/utils/net/NetUtils.java
@@ -268,7 +268,7 @@ public class NetUtils {
             final String defDev = Script.runSimpleBashScript("/sbin/route -n get default 2> /dev/null | grep interface | awk '{print $2}'");
             return defDev;
         }
-        return Script.runSimpleBashScript("ip route | grep default | head -1 | grep -Po '(?<=dev )(\\S+)'");
+        return Script.runSimpleBashScript("awk '$2 == 00000000 { print $1 }' /proc/net/route");
     }
 
     public static String getLocalIPString() {
--- a/utils/src/main/java/com/cloud/utils/net/NetUtils.java
+++ b/utils/src/main/java/com/cloud/utils/net/NetUtils.java
@@ -268,7 +268,7 @@ public class NetUtils {
             final String defDev = Script.runSimpleBashScript("/sbin/route -n get default 2> /dev/null | grep interface | awk '{print $2}'");
             return defDev;
         }
-        return Script.runSimpleBashScript("awk '$2 == 00000000 { print $1 }' /proc/net/route");
+        return Script.runSimpleBashScript("ip route show default 0.0.0.0/0 | head -1 | awk '{print $5}'");
     }
 
     public static String getLocalIPString() {
--- a/utils/src/main/java/com/cloud/utils/net/NetUtils.java
+++ b/utils/src/main/java/com/cloud/utils/net/NetUtils.java
@@ -268,19 +268,7 @@ public class NetUtils {
             final String defDev = Script.runSimpleBashScript("/sbin/route -n get default 2> /dev/null | grep interface | awk '{print $2}'");
             return defDev;
         }
-        final String defaultRoute = Script.runSimpleBashScript("/sbin/route | grep default");
-
-        if (defaultRoute == null) {
-            return null;
-        }
-
-        final String[] defaultRouteList = defaultRoute.split("\\s+");
-
-        if (defaultRouteList.length != 8) {
-            return null;
-        }
-
-        return defaultRouteList[7];
+        return Script.runSimpleBashScript("ip route | grep default | head -1 | grep -Po '(?<=dev )(\\S+)'");
     }
 
     public static String getLocalIPString() {
--- a/debian/control
+++ b/debian/control
@@ -15,7 +15,7 @@ Description: A common package which contains files which are shared by several C
 
 Package: cloudstack-management
 Architecture: all
-Depends: ${python:Depends}, openjdk-8-jre-headless | java8-runtime-headless | java8-runtime | openjdk-9-jre-headless, cloudstack-common (= ${source:Version}), sudo, python-mysql.connector, libmysql-java, augeas-tools, mysql-client, adduser, bzip2, ipmitool, file, lsb-release, init-system-helpers (>= 1.14~)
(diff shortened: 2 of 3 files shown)


rhtyd:
@andrijapanicsb no, we don't need net-tools (that's for older tools like ifconfig, route etc.) My PR switches to iproute (uses ip route and not `route`).

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

utils/src/main/java/com/cloud/utils/net/NetUtils.java
```diff
@@ -268,19 +268,7 @@ public static String getDefaultEthDevice() {
...
-        }
-
-        final String[] defaultRouteList = defaultRoute.split("\\s+");
-
-        if (defaultRouteList.length != 8) {
-            return null;
-        }
-
-        return defaultRouteList[7];
+        return Script.runSimpleBashScript("ip route | grep default | head -1 | grep -Po '(?<=dev )(\\S+)'");
```
weizhouapache: @rhtyd 
when I run the command on ubuntu 18.04, I got empty result. is that normal ?

```
root@mgt118-4:~# ip route | grep default | head -1 | grep -Po '(?<=dev )(\\S+)'
root@mgt118-4:~# ip route | grep default | head -1
default via 10.11.118.254 dev eth0
```

rhtyd: @weizhouapache yes, since it's a java string I've to use the double slash `\\`.
It would run this: (note the `\\S` in the string vs `\S` that it runs on bash)
`ip route | grep default | head -1 | grep -Po '(?<=dev )(\S+)'`

wido: I would suggest:

<pre>
ip route show default 0.0.0.0/0|head -1|awk '{print $3}'
</pre>

If a Host has a large routing table you can run into problems here. By fetching the route for 0.0.0.0/0 you query the routing table faster.

weizhouapache: agrees. it should be better

DaanHoogland: ```ip route show default``` does not work with ip from brew on mac (which is not a big issue but would be nice.

DaanHoogland: on mac
```
baloujak:var dahn$ ip route get default
default via 192.168.144.99 dev en0
baloujak:var dahn$ ip route get default | awk '{print $5}'
en0
```
have to test if that would work on other platforms

andrijapanicsb: kvm-centos7-kvm1 ~]# ip route get default
need at least a destination address

andrijapanicsb:
> code looks good, needs testing on a two ms instance with ssl enabled to see if it solves the cert issue. Analytically I'd say it does (so it probably doesn't @andrijapanicsb ;).

What does it mean "ssl enabled"? I did test with 2 MS instances but didn't actually start the second one...which perhaps it not good enough test - let me do that.

utils/src/main/java/com/cloud/utils/net/NetUtils.java
DaanHoogland: I don't know where that commen t went but @wido we need the device on which the default route is so we need to {print $5} not $3

DaanHoogland:
@andrijapanicsb,

> What does it mean "ssl enabled"?

the default (self signed) rootCA is relying on this code.

andrijapanicsb:
All good, and both mgmt servers fine, single KMV agent also fine:

grep -i "certificate ownership verification failed for client" /var/log/cloudstack/management/management-server.log 

returns no results on either MS servers.


LGTM

utils/src/main/java/com/cloud/utils/net/NetUtils.java
andrijapanicsb: While there might be some speed improvements in Wido's comment, I don't expect one is going to have tons of routes on a blank management server (where the needed code we are patching is being executed), and I definitively will not have time testing it all over again.

wido: I don't mind that much. I would just think about these things are you don't need to print the whole routing table.

Especially on a Hypervisor/Host this is something to think of.

andrijapanicsb: @wido , I do recall that ip route work as part of refactoring VXLAN scripts, so that makes sense perfectly.

rhtyd: If speed is a requirement, I won't bother iproute2, just use `procfs`: `awk '$2 == 00000000 { print $1 }' /proc/net/route`

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

utils/src/main/java/com/cloud/utils/net/NetUtils.java
wido: I'd just stick with ip route. That's good and reliable.

You might want to add the '-o' flag which makes it better to parse.

utils/src/main/java/com/cloud/utils/net/NetUtils.java
```diff
@@ -268,19 +268,7 @@ public static String getDefaultEthDevice() {
...
-        }
-
-        final String[] defaultRouteList = defaultRoute.split("\\s+");
-
-        if (defaultRouteList.length != 8) {
-            return null;
-        }
-
-        return defaultRouteList[7];
+        return Script.runSimpleBashScript("awk '$2 == 00000000 { print $1 }' /proc/net/route");
```
DaanHoogland: in @wido 's style i'd prefer: (whatever happened to "suggest change"?)
```ip route show default 0.0.0.0/0|head -1|awk '{print $5}'```
note that we do need the fifth field, i.e. return the device not an address.

File: utils/src/main/java/com/cloud/utils/net/NetUtils.java [hunk 3]
```diff
@@ -268,19 +268,7 @@ public static String getDefaultEthDevice() {
...
-        }
-
-        final String[] defaultRouteList = defaultRoute.split("\\s+");
-
-        if (defaultRouteList.length != 8) {
-            return null;
-        }
-
-        return defaultRouteList[7];
+        return Script.runSimpleBashScript("ip route show default 0.0.0.0/0 | head -1 | awk '{print $5}'");
```
rhtyd: @wido @DaanHoogland while the `/proc` subsystem is stable enough, I've used iproute now - how about now ^^ ?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3710, opened 2019-11-24
[Work in progress - do not merge] Replace builtin CentOS 5.x with 8.0 templates

andrijapanicsb:
## Description

New default/bultin templates for major hypervisors:

- CentOS 8.0 (VMware/KVM/XenServer) 
- Minimal install with "standard" / "core" groupinstall (a.k.a. additional network tools - ifconfig, route, etc.)
- nano, iftop, iotop, htop, bash-completion, traceroute
- ssh running, systemctl stop/disable firewalld, SELinux disabled
- disabled "predictive" interface names (net.ifnames=0 added to kernel boot line)
- stripped UUID/MAC from eth0 network config file
- VMware: opensource variant "open-vm-tools" installed - recommended by VMware (instead of using the one from ESXi hosts); both vmw_pvscsi and vmxnet3 drivers present and details added to vm_template_table to spin VMs with those devices
- XenServer: XenServer Tools 6.5sp1 installed (xs-tools confirmed as running, considering this is CentOS8 which the xs-tools installation script was not aware of)
- KVM: no special configuration, VirtIO present by default

## Types of changes
- [ x] Enhancement (improves an existing feature and functionality)

## How Has This Been Tested?
Templates were deployed on ACS install with old versions of hypervisors: VMware 5.5, XenServer 6.5sp1 fully patched up to patch 1066 and CentOS 7.3 (3 clusters, mixed env) and verified to work on those versions of hypervisors.

Yet to be tested against VMware 6.0/6.5/6.7, XenServer 7.0, 7.1, 7.x and Ubuntu 16/18, CentOS 7.6 maybe - but if it works on older ones than any newer hypervisor should be no problem.

## Comment:
- The only concern is that XenServer 6.5 must have all the patches installed (confirmed that patch XS65ESP1066 is needed) - otherwise, CentOS 8 (actually, anything newer than CentOS 7.2) will not boot successfully.

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-41300to41400-cleanup.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41300to41400-cleanup.sql
@@ -19,6 +19,6 @@
 -- Schema upgrade cleanup from 4.13.0.0 to 4.14.0.0
 --;
 
-# Delete an ancient template added in 2.2, which was set as removed in 4.5. For sake of cleanup and easier overview of builtin templates.
-# This DOES assume one is not running any VMs from this template!
+-- Delete an ancient template added in 2.2, which was set as removed in 4.5. For sake of cleanup and easier overview of builtin templates.
+-- This DOES assume one is not running any VMs from this template!
 DELETE FROM `cloud`.`vm_template` WHERE `id`=2 AND `unique_name`="centos53-x86_64" AND `name`="CentOS 5.3(64-bit) no GUI (XenServer)";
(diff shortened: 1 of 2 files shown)


andrijapanicsb:
@blueorangutan package

andrijapanicsb:
I need "help" confirming that I've done the proper job of removing an ancient template "" which was added in 2.2 and then marked as  removed in 4.5 - this is purely cleanup so we can more eassily observe those in vm_templates table (I've issues with garbage collection)
- I've added the delete SQL statement for schema upgrade cleanup file from 4.13 to 4.14 - that is fine for sure
- I've removed 2 lines from the setup/db/templates.sql - so that the template is NOT added in the dirst place for the new installations on master (to be 4.14) - this one is what I would like someone to confirm is OK.

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
(original templates.sql file might not be the right place to change things, will need to redo some changes later)

andrijapanicsb:
Wrong places to edit the files, closing this one. Work continued in #3712

Issue #3700, opened 2019-11-21
update default user vm template to later version


borisstoyanov:
##### ISSUE TYPE

 * Improvement Request
 * Enhancement Request

##### COMPONENT NAME

~~~
Templates
~~~

##### CLOUDSTACK VERSION

~~~
4.14
~~~

##### SUMMARY

Currently cloudstack uses CentOS 5.5 as default user VM template, can we update that to latest CentOS 7 or even 8 versions if possible?

andrijapanicsb:
Not sure we have CentOS8 mapping/guest OS for all hypervisors - but sounds like a very valid proposition (that I had in my head last few days as well). Let me see if I have time to do something around that.

borisstoyanov:
is this being fixed in #3712 right?

andrijapanicsb:
yep

andrijapanicsb:
@borisstoyanov is using CentOS8 going to break any Marvin tests (Inside CentOS8 - ssh still works, and all  basic network utils available inside, including "old" ifconfig/route etc - I'm talking about #3712 ) -

DaanHoogland:
@borisroman @andrijapanicsb does it make sense to let the operator/user decide on the default usertemplate by prioviding a url and guestos type at install time? or through an API?

andrijapanicsb:
I don't think so @DaanHoogland  - the more "freedom" you give to the user, the more issues you might face.
We need to see what is the "newest" guest OS, supported by the OLDEST hypervisor we officially support in ACS - we can't use a newer OS than that (and we should not be using older, like the current one).
Remeber - this template is barely to validate/test basic ACS functionality and if the deployment is successful (can deploy a VM in a clean/new ACS installation)

One thing to notice is that we also run Marvin tests using this default template, so.... need to take care on steps forward.

rohityadavcloud:
We should bring this discussion on dev@, which I see has been tried before. The purpose of the default template is just to cloudstack in a new installation it works. With EL9 around, it may make sense to discuss it again on dev@. I'll start that discussion but close this 2yr+ old ticket.

PR #3712, opened 2019-11-24
[WORK IN PROGRES - do NOT merge] Replace builtin CentOS 5.x with 8.0 templates

andrijapanicsb:
## Description

New default/bultin templates for major hypervisors:
(new PR, instead of #3710 ) 
- CentOS 8.0 (VMware/KVM/XenServer) 
- Minimal install with "standard" / "core" groupinstall (a.k.a. additional network tools - ifconfig, route, etc.)
- nano, iftop, iotop, htop, bash-completion, traceroute
- ssh running, systemctl stop/disable firewalld, SELinux disabled
- disabled "predictive" interface names (net.ifnames=0 added to kernel boot line)
- stripped UUID/MAC from eth0 network config file
- VMware: opensource variant "open-vm-tools" installed - recommended by VMware (instead of using the one from ESXi hosts); both vmw_pvscsi and vmxnet3 drivers present and details added to vm_template_table to spin VMs with those devices
- XenServer: XenServer Tools 6.5sp1 installed (xs-tools confirmed as running, considering this is CentOS8 which the xs-tools installation script was not aware of)
- KVM: no special configuration, VirtIO present by default

## Types of changes

## How Has This Been Tested?
Templates were deployed on ACS install with old versions of hypervisors: VMware 5.5, XenServer 6.5sp1 fully patched up to patch 1066 and CentOS 7.3 (3 clusters, mixed env) and verified to work on those versions of hypervisors.

Yet to be tested against VMware 6.0/6.5/6.7, XenServer 7.0, 7.1, 7.x and Ubuntu 16/18, CentOS 7.6 maybe - but if it works on older ones than any newer hypervisor should be no problem.

## Comment:
- The only concern is that XenServer 6.5 must have all the patches installed (confirmed that patch XS65ESP1066 is needed) - otherwise, CentOS 8 (actually, anything newer than CentOS 7.2) will not boot successfully.


andrijapanicsb:
@blueorangutan package

borisstoyanov:
@blueorangutan test matrix

andrijapanicsb:
Those tests failed @borisstoyanov (as you are aware) - interesting it doesn't report back...

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
Looking good - ACS deploys successfully and downloads the correct template (base on the HV present in the Zone).

From my side, this is LGTM (though this doesn't count).
Will test deploying same templates on multiple other versions of vSphere/XenServer/KVM just to ensure the OS itself can boot properly.

andrijapanicsb:
@blueorangutan test matrix

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan test matrix

engine/schema/src/main/resources/META-INF/db/schema-41300to41400-cleanup.sql
```diff
@@ -19,3 +19,6 @@
 -- Schema upgrade cleanup from 4.13.0.0 to 4.14.0.0
 --;
 
+-- Delete an ancient template added in 2.2, which was set as removed in 4.5. For the sake of cleanup and easier overview of builtin templates in DB.
+-- This DOES assume one is not running any VMs from this template!
+DELETE FROM `cloud`.`vm_template` WHERE `id`=2 AND `unique_name`="centos53-x86_64" AND `name`="CentOS 5.3(64-bit) no GUI (XenServer)";
```
GabrielBrascher: Considering:
> "This DOES assume one is not running any VMs from this template!"

Does it make sense to add an extra verification that ensures the deletion only if there is no VMs on with that template ID (therefore deleting only if does not exists VMs with _removed NULL_)?

For instance:
```
DELETE FROM `cloud`.`vm_template` WHERE `id`=2 AND `unique_name`="centos53-x86_64" AND `name`="CentOS 5.3(64-bit) no GUI (XenServer)" AND NOT EXISTS (SELECT 1 FROM `cloud`.`vm_instance` WHERE vm_template_id=2 AND removed IS NULL);
```

andrijapanicsb: Thx @GabrielBrascher - done.

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
Don't run tests until I sort out Trillian usage of templates (we currently replace template DB URLs with custom/internal repo which is pointing to CentOS 5.5, so it makes whole testing useless). I have work in progress here.

andrijapanicsb:
Tested with latest hypervisors we support atm - and all works fine:
- VMware 6.7 U3
- XenServer 7.6
- CentOS 7.7 (stock qemu/libvirt versions)

andrijapanicsb:
This one needs to be parked for some time - MANY Marvin tests are failing since we are relying on builtin CentOS 5.x templates during testing - and now when using CentOS 8, with just 128MB of RAM Offerings, there is kernel panic inside CentOS8 due to low RAM so Marvin tests will fail.

This will need some more planning around Marvin testing/templates used.

rhtyd:
@andrijapanicsb use something lightweight like alpinelinux?

andrijapanicsb:
That makes sense @rhtyd - just needs some time and effort to be changed - but, again, can be done irrespective of this PR - since Trillian replaces URLs always, so this PR will NOT affect Marvin tests.
I agree for something much more lightweight (or worst case, just keep using same CentOS 5.5, if there is nothing even MORE lightweight, considering we give those VMs just 128MB of RAM)

DaanHoogland:
moving to 4.15 (for now)
btw, it is not a bad idea to combine default SVM template with builtin, imnsho

DaanHoogland:
are we abandonning this as it is centos8, @andrijapanicsb @rhtyd @GabrielBrascher ?
And will we plan a new implementation for #3700?

rhtyd:
Agree we should abandon this and revisit something else. The purpose of the build-in template is for testing only, I don't think most users/service providers etc would use the built-in tempalte in prod, so maybe an argument could be we don't need the built-in template (the current one is okay) or may be have the debian/systemvmtemplate itself be used as a built-in template.

Issue #3625, opened 2019-10-07
ui allows disabled API call and then ends session


DaanHoogland:
In 4.9 a user that did a call that was not allowed from the UI would get a permission denied error in the console. With 4.11.2 a 401 is thrown and the session is ended
 
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI, API
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2.0, others to be tested
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

when creating special permissions for a role a user would expect to get a comprehensible error message if he tries to call an API that is DENY'd them. Using cloudmonkey (after sync) the API will not be available. Before sync'ing the call results in "Error: (HTTP 401, error code <nil>) unable to verify user credentials and/or request signature"

##### STEPS TO REPRODUCE

~~~
for instance:
- login as an admin,
- set the rule for domainAdmin of disableUser to permission == DENY
- log on as a domain admin and try to disable a user.
~~~

##### EXPECTED RESULTS

~~~
an error message specifying that the call is not allowed. I.E. "The user is not allowed to request the API command or the API command does not exist" was shown in 4.9 (both in UI and cloudmonkey)
~~~

##### ACTUAL RESULTS

~~~
a session reset without clear reason. 
~~~

DaanHoogland:
9988c269b259b84c0b8436bad17f88dbc1d706e7 introduces a check and than returns false, instead of throwing ServerApiException. I suspect this is resulting in the session being dropped instead of the exception message getting to the user as error response.

DaanHoogland:
encoutered 9a13227a78acb14d9fd4f9ad5246ce67396cef7a that fixes this.

DaanHoogland:
@wido @rhtyd after analysing I think the session is invalidated to rigorously in the mentioned commits. Please consider:

1. if a valid user does a request they are not allowed, an error response must be given and the session maintained
2. if a user makes a request from a disallowed ip, the session must be aborted/invalidated
3. if an unauthenticated user makes a request the session should have not existed in the first place (and I think we have no issue here)

the first .1 is old behaviour which no longer works. .2 is new behaviour that makes sense and should be preserved. If you agree I'll create a PR for this.

rhtyd:
@DaanHoogland okay here are my thoughts:
[...]

PR #3713, opened 2019-11-25
break session only on illegal origin

DaanHoogland:
## Description

Added a new CloudRuntime Exception to distinguish between command not allowed and command issued from disallowed origin, so users that are authenticated still get a proper error message instead of having their session revoked bluntly.

this was a regression in 4.11 (compared to 4.9) so it is on the 4.11 branch and will have to be merged forward.

Fixes: #3625 
## Types of changes


DaanHoogland:
error'd out because nuagevsp is no longer available in maven central?
@fmaximus is this a fluke or genuine?

DaanHoogland:
we need to decide on 4.11 and maybe backport

PR #3724, opened 2019-11-28
Storage-based Snapshots for KVM VMs

slavkap:
## Description

Cloudstack, with KVM as a hypervisor, provides live VM snapshots only with volumes whose image format is QCOW. 
This is a limitation for storage providers with disks in RAW format.

link to ML: http://mail-archives.apache.org/mod_mbox/cloudstack-dev/201911.mbox/%3cCAA6FghF7eaY-A3XGN5zSwKVvp7zrcuNzv5nFAQKaF+et3zH-ag@mail.gmail.com%3e#archives

With this feature
- CloudStack users will be able to do live consistent snapshots of all disks attached to the VM 
- Is using respective storage provider plugin implementation to take the snapshot
- For NFS/Local storage - now this functionality is limited until #5297 gets merged
- You can create a VM snapshot without its memory

## Types of changes

## Documentation

https://github.com/apache/cloudstack-documentation/pull/153

## Prerequisites

guest agent enabled - used to freeze/thaw the VM

## How Has This Been Tested?

Environment:

- CloudStack - management and agent version - main
- OS for management and hypervisor - CentOS7
- Hypervisor - KVM
- Primary storages - NFS, StorPool, Local storage and Ceph. 

Enable the global configuration setting - `kvm.vmstoragesnapshot.enabled`

Take snapshot:
- Take VM snapshot without memory on running virtual machine
- We have tested this feature with these storage providers - NFS, StorPool, Local storage and Ceph. 
- There is one smoke test, which creates VM, installs qemu-guest-agent, takes, reverts and deletes VM snapshot. 

Revert snapshot
- Stop the virtual machine and revert the VM snapshot


DaanHoogland:
@blueorangutan package

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan test

weizhouapache:
@slavkap @andrijapanicsb 
this is a nice feature.
I will review and test it . It might take some days, depends on my availability.

slavkap:
Thank you @weizhouapache that you will spend time on this! I am available if you have any comments or questions.

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
@slavkap - nice stuff as already commented!

1. I see "lots" of dependency which has to be documented properly in upstream documentation (preferably) but also I would like to see it documented in the description of the global setting - "...requires qemu 1.6+ and the qemu-guest-agent installed inside a VM" 

2. Do you know how well qemu-guest-agent is supported/possible on Windows guests - lot's of people will be running Windows VMs as well - covering just Linux workloads would be suboptimal.

3. You are mentioning QCOW format and behaviour about it. Ceph uses RAW format from Qemu/Libvirt perspective, but possibly is marked as QCOW2 in DB (I recall some hassle around DB format in past) - can you please elaborate about this - how does it work with Ceph vs other "real" QCOW2 formats? Store

4. Currently, due to someone noticing bugs in QCOW2 snapshots (or restore of) where the volume/filesystem of the guest gets corrupted - volume snapshots have been disabled for running VMs for KVM. @weizhouapache can you comment on this? I'm wondering if we might hit the same issue here as well

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFreezeThawVMCommandWrapper.java
```diff
@@ -0,0 +1,92 @@
...
+                return new FreezeThawVMAnswer(command, false, String.format("Failed to %s due to %s was not found",
+                        command.getOption().toUpperCase(), vmName));
+            }
+            DomainState domainState = domain.getInfo().state ;
+            if (domainState != DomainState.VIR_DOMAIN_RUNNING) {
+                return new FreezeThawVMAnswer(command, false,
+                        String.format("%s of VM failed due to vm %s is in %s state", command.getOption().toUpperCase(),
+                                vmName, domainState));
+            }
+            Script cmd = new Script("virsh", s_logger);
```
wido: Using virsh is not reliable. In the CloudStack agent we use the JNA bindings for libvirt and I suggest to use those here as well.

We should not execute virsh from inside Java

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
```diff
@@ -947,25 +953,57 @@ public Answer backupSnapshot(final CopyCommand cmd) {
...
-                    s_logger.debug("Failed to backup snaptshot: " + result);
-                    return new CopyCmdAnswer(result);
-                }
-                final File snapFile = new File(snapshotDestPath + "/" + descName);
-                if(snapFile.exists()){
-                    size = snapFile.length();
+                if (isKVMEnabled) {
+                    OutputInterpreter.OneLineParser parser = new OutputInterpreter.OneLineParser();
+                    Script sc = new Script("virsh");
+                    sc.add(String.format("qemu-monitor-command %s '{ \"execute\" : \"query-block-jobs\" }'", vmName));
```
wido: This can also be executed through libvirt's Java bindings and we should not do this through a separate command.

Avoid any execution from Java as it's unreliable.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
```diff
@@ -1456,22 +1496,65 @@ public Answer createSnapshot(final CreateObjectCommand cmd) {
...
                 final String vmUuid = vm.getUUIDString();
-                final Object[] args = new Object[] {snapshotName, vmUuid};
-                final String snapshot = SnapshotXML.format(args);
+                boolean isKVMEnabled = cmd.getContextParam("kvmsnapshot") != null ? Boolean.parseBoolean(cmd.getContextParam("kvmsnapshot")) : false;
+                s_logger.debug(String.format("Snapshots on KVM is enabled %s", isKVMEnabled));
+                if (isKVMEnabled) {
+                    long size = 0;
+                    OutputInterpreter.OneLineParser parser = new OutputInterpreter.OneLineParser();
+                    Script sc = new Script("virsh");
+                    sc.add(String.format("qemu-monitor-command %s '{ \"execute\" : \"query-block\" }'", vmName));
```
wido: Same here as my other comments

andrijapanicsb:
moving to 4.15, I'm afraid we won't be able to test in time.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
weizhouapache: @wido libvirt-java is not updated for quite long time.
'virsh' provides more functions.

weizhouapache:
> 4\. urrently, due to someone noticing bugs in QCOW2 snapshots (or restore of) where the volume/filesystem of the guest gets corrupted - volume snapshots have been disabled for running VMs for KVM. @weizhouapache can you comment on this? I'm wondering if we might hit the same issue here as well

@andrijapanicsb as far as I remember, vm is paused not only when we create snapshot but also delete a snapshot. I think corrupted volume issue should be fixed.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
wido: Yes, but then we fix that. It's a matter of sending a patch to the libvirt project and they are happy to accept it.

I think cloudstack is the only project really using those bindings. But they are not so difficult, so we can send patches when we need something.

We should not just work around it by not using them. Executing commands is just not reliable enough.

slavkap: Thanks @wido for the comments!
We will check what will be needed for the patch, and we will start working on it. We will ask for help, if we get stuck somewhere.
Could you please give me and example why "executing commands is just not reliable enough"? And if there are a problems with commands execution whether there will be the same with execution of virsh in helper scripts?

wido: All the calls to libvirt from the agent are done through the Java bindings. If we suddenly start to execute virsh from Java we work around all the existing code which is there to have an open connection with libvirt.

In addition you can only check on exit status and not handle exceptions properly coming back from libvirt.

If we are lacking features in libvirt-java those need to be added through the libvirt project.

slavkap:
Hello @andrijapanicsb, sorry for the late response! Bellow you can find answers of your questions

>  I see "lots" of dependency which has to be documented properly in upstream documentation (preferably) but also I would like to see it documented in the description of the global setting - "...requires qemu 1.6+ and the qemu-guest-agent installed inside a VM"

We'll document the functionality, but could you please tell me what should be included? Also I will update the global setting with:

>I would like to see it documented in the description of the global setting - "...requires qemu 1.6+ and the qemu-guest-agent installed inside a VM"

>    Do you know how well qemu-guest-agent is supported/possible on Windows guests - lot's of people will be running Windows VMs as well - covering just Linux workloads would be suboptimal.
	
	For Windows freeze/thaw of guest agent is included in versions from 2013

>    You are mentioning QCOW format and behaviour about it. Ceph uses RAW format from Qemu/Libvirt perspective, but possibly is marked as QCOW2 in DB (I recall some hassle around DB format in past) - can you please elaborate about this - how does it work with Ceph vs other "real" QCOW2 formats? Store

Ceph is marked in DB as RAW format, but this doesn't affect snapshot and revert of Ceph's volumes, because we are using its implementation for this. For each primary storage we are using appropriate plugin with take/revert snapshot implementation.

   
    Did you guys test this on the busy VMs (IO and CPU heavy workloads)?

We've completed our tests including on busy VMs with different primary storages. They've been successfully reverted and it works as expected.

> moving to 4.15, I'm afraid we won't be able to test in time.

Unfortunately we'll need time to implement qemu-agent and qemu-monitor commands functionality in libvirt java api, and don't know when it will be accepted.

slavkap:
Hello,

I have proposed to libvirt-java project the virDomainQemuMonitorCommand and virDomainQemuAgentCommand implementation (as @wido requested), which were accepted and merged to the project. I'll be happy if some of you can help me with information, what should I do, those changes  to go to maven central repo?

Thanks in advance!

wido:
Great news @slavkap ! We should now request a new release of libvirt-java in Maven so we can depend on that.

slavkap:
Thanks @wido ! I'll work on that.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFreezeThawVMCommandWrapper.java
slavkap: replaced with libvirt's Java API

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
slavkap: here also is replaced the use of virsh with libvirt's Java API

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
slavkap: replaced with libvirt's Java bindings

alamintech:
I have some problem. I am used ceph storage but after create vm then reboot ceph server osd service can't start. How can I solve  this. this is ceph issue or cloudstack issue ? I am not sure.

slavkap:
> I have some problem. I am used ceph storage but after create vm then reboot ceph server osd service can't start. How can I solve this. this is ceph issue or cloudstack issue ? I am not sure.

Hello @alamintech I'm not deeply familiar with CEPH, but I guess you should ask this question in users or dev mailing list. This pull request does not affect the creation of VM or Ceph's services

alamintech:
ok, Thanks.

rhtyd:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java
```diff
@@ -0,0 +1,66 @@
+package org.apache.cloudstack.utils.qemu;
```
GabrielBrascher, line 1: Can you please add the Apache license header?

GabrielBrascher (changes requested):
Thanks for the PR @slavkap.

I raised a few comments and will be taking a deeper look at it this week as well.

slavkap:
> Thanks for the PR @slavkap.
> 
> I raised a few comments and will be taking a deeper look at it this week as well.

Thank you @GabrielBrascher!
I did the changes regarding your comments

server/src/main/java/com/cloud/vm/snapshot/VMSnapshotManagerImpl.java
```diff
@@ -358,9 +359,13 @@ public VMSnapshot allocVMSnapshot(Long vmId, String vsDisplayName, String vsDesc
             throw new InvalidParameterValueException("Can not snapshot memory when VM is not in Running state");
         }
 
+        boolean isKVMsnapshotsEnabled = SnapshotManager.VMsnapshotKVM.value() == null || SnapshotManager.VMsnapshotKVM.value();
```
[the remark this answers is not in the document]
slavkap: Thanks! I've missed this scenario.

GabrielBrascher:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland:
@slavkap can you look at the logs for those errors and assess them? @GabrielBrascher can you give a second look (code freeze imminent)

DaanHoogland:
and @weizhouapache can you give this a look as well?

slavkap:
> @slavkap can you look at the logs for those errors and assess them? @GabrielBrascher can you give a second look (code freeze imminent)

@DaanHoogland I think I've found the problem, which is related to unused import that cannot find module. But I think that there will be another problem - in the tests I'm installing "qemu-guest-agent" on the quest VM, is there a way to point out a template which is higher than CentOS 5.5 (I mean which is not out of support)

DaanHoogland:
@slavkap i think there is a PR out to up the default template. not sure if it is ready or merged. for test purposes look at http://dl.openvm.eu/cloudstack/macchinina/x86_64/

GabrielBrascher:
@DaanHoogland this PR looks like a good one to get into 4.15. I will help to review/test it :slightly_smiling_face:

GabrielBrascher:
@slavkap I tested but the Freeze command returns code 1 causing the VM snapshot to fail.

#### 1. Management log:
```
2020-10-29 19:07:56,030 INFO [o.a.c.s.v.StorageVMSnapshotStrategy] (Work-Job-Executor-23:ctx-15f06ac9 job-56793/job-56794 ctx-ad9cccca) (logid:a6a700e8) Freeze of virtual machine took 0 miliseconds. 
2020-10-29 19:07:56,034 DEBUG [c.c.a.t.Request] (Work-Job-Executor-23:ctx-15f06ac9 job-56793/job-56794 ctx-ad9cccca) (logid:a6a700e8) Seq 1-2747758722649426716: Sending { Cmd , MgmtId: 2482717785957, via: 1(n01.2019.cldin.net), Ver: v1, Flags: 100011, [{"com.cloud.agent
.api.FreezeThawVMCommand":{"vmName":"i-2-13022-VM","option":"thaw","wait":"0"}}] } 
2020-10-29 19:07:56,092 DEBUG [c.c.a.t.Request] (AgentManager-Handler-9:null) (logid:) Seq 1-2747758722649426716: Processing: { Ans: , MgmtId: 2482717785957, via: 1, Ver: v1, Flags: 10, [{"com.cloud.agent.api.FreezeThawVMAnswer":{"result":"false","details":"Failed to TH
AW vm i-2-13022-VM due to result status is: {"return":1}","wait":"0"}}] } 
2020-10-29 19:07:56,093 DEBUG [c.c.a.t.Request] (Work-Job-Executor-23:ctx-15f06ac9 job-56793/job-56794 ctx-ad9cccca) (logid:a6a700e8) Seq 1-2747758722649426716: Received: { Ans: , MgmtId: 2482717785957, via: 1(n01.2019.cldin.net), Ver: v1, Flags: 10, { FreezeThawVMAnswe
r } } 
2020-10-29 19:07:56,099 DEBUG [c.c.v.s.VMSnapshotManagerImpl] (Work-Job-Executor-23:ctx-15f06ac9 job-56793/job-56794 ctx-ad9cccca) (logid:a6a700e8) Failed to create vm snapshot: 12 
com.cloud.utils.exception.CloudRuntimeException: Could not freeze VM.Failed to FREEZE vm i-2-13022-VM due to result status is: {"return":1} 
       at org.apache.cloudstack.storage.vmsnapshot.StorageVMSnapshotStrategy.takeVMSnapshot(StorageVMSnapshotStrategy.java:233) 
       at com.cloud.vm.snapshot.VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VMSnapshotManagerImpl.java:551) 
       at com.cloud.vm.snapshot.VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VMSnapshotManagerImpl.java:1234) 
       at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) 
       at com.cloud.vm.VmWorkJobHandlerProxy.handleVmWorkJob(VmWorkJobHandlerProxy.java:107) 
       at com.cloud.vm.snapshot.VMSnapshotManagerImpl.handleVmWorkJob(VMSnapshotManagerImpl.java:1261) 
       at jdk.internal.reflect.GeneratedMethodAccessor347.invoke(Unknown Source)
```

#### 2. message when checking _freezeAnswer.getResult()_ via remote debug:
`Failed to FREEZE vm i-2-13029-VM due to result status is: {"return":1}`

It is interesting that the VM got frozen, with virsh I checked the status before and after the command being executed. It seems to work but the result code was 1, which leads to CloudStack ending the process assuming that it failed to freeze the VM.

#### 3. Environment

KVM running on Ubuntu 18.04: 
Compiled against library: libvirt 4.0.0
Using library: libvirt 4.0.0
Using API: QEMU 4.0.0
Running hypervisor: QEMU 2.11.1

#### 4. Manual tests
When running qemu via command line on the same KVM node then it worked fine:

```
~# virsh qemu-agent-command 82 '{"execute": "guest-fsfreeze-status"}'|jq 
{ 
 "return": "thawed" 
}
~# virsh qemu-agent-command 82 '{"execute": "guest-fsfreeze-freeze"}'|jq 
{ 
 "return": 2 
} 
~# virsh qemu-agent-command 82 '{"execute": "guest-fsfreeze-status"}'|jq 
{ 
 "return": "frozen" 
}
```

#### 5. Bypassing the freeze code result

Via remote debug I changed the result code from 1 to 2 and then all worked fine. The VM got freeze, and then the thaw command worked fine, unfreezing it; with that bypass the snapshot was taken as expected.

#### 6. Conclusion

The overall process seems to be working fine; however, I got caught by weird behavior when freezing the VM. I will take another look at it. But if you have any clue about what might be happening please let me know.

GabrielBrascher (commented):
One small thing that I noticed when setting the configuration in global settings is that it is not dynamic. Is there any specific motivation for this behavior?

wido:
> @GabrielBrascher thank you for the feedback and that you spend time to test!
> Unfortunately I haven't faced such a problem. But I will do a research
> 
> My environment is:
> KVM running on CentOS 7
> libvirt version - libvirtd (libvirt) 4.5.0
> qemu-kvm-ev-2.12.0-33.1.el7_7.4

Gabriel tested with Ubuntu 18.04 and Ubuntu VM’s as well.

Do we know what returncode 1 or 2 mean? Might both be valid?

slavkap:
Yes, I saw what is Gabriel's environment, but his manual tests aren't failing and the return code is 2 on success.

slavkap:
in documentation of Qemu quest says this:

> Command: guest-fsfreeze-freeze
> Sync and freeze all freezable, local guest filesystems. If this command succeeded, you may call guest-fsfreeze-thaw later to unfreeze.
> 
> Note: On Windows, the command is implemented with the help of a Volume Shadow-copy Service DLL helper. The frozen state is limited for up to 10 seconds by VSS.
> 
> Returns: Number of file systems currently frozen. On error, all filesystems will be thawed. If no filesystems are frozen as a result of this call, then guest-fsfreeze-status will remain "thawed" and calling guest-fsfreeze-thaw is not necessary.
> 
> Since: 0.15.0

maybe I do not have to check the return code, and just in case to execute "guest-fsfreeze-status" to be sure that is freeze/thawed

GabrielBrascher:
@slavkap maybe I was not clear on the manual tests that passed.

I got result code of 2 when executed via command line.
When executing the snapshot via CloudStack, via remote debug, I saw that it returns code 1 and then trhows the exception.

With that said, I am checking on documentation to look for what it means. Also, the idea of checking for the status instead of result code looks like a good option.

GabrielBrascher:
Looking at the qemu documentation, it looks that a result code means the _Number of file systems currently frozen_. Not sure why it resulted in **2** when running the command line, and it resulted in **1** via CloudStack execution flow, but I don't think that 1 would be a problem.

#### guest-fsfreeze-freeze:
_Sync and freeze all freezable, local guest filesystems. If this command succeeded, you may call guest-fsfreeze-thaw later to unfreeze.

Note: On Windows, the command is implemented with the help of a Volume Shadow-copy Service DLL helper. The frozen state is limited for up to 10 seconds by VSS.

Returns: Number of file systems currently frozen. On error, all filesystems will be thawed. If no filesystems are frozen as a result of this call, then guest-fsfreeze-status will remain "thawed" and calling guest-fsfreeze-thaw is not necessary._

server/src/main/java/com/cloud/storage/snapshot/SnapshotManager.java
```diff
@@ -59,6 +59,8 @@
     public static final ConfigKey<Boolean> BackupSnapshotAfterTakingSnapshot = new ConfigKey<Boolean>(Boolean.class, "snapshot.backup.to.secondary",  "Snapshots", "true",
             "Indicates whether to always backup primary storage snapshot to secondary storage. Keeping snapshots only on Primary storage is applicable for KVM + Ceph only.", false, ConfigKey.Scope.Global, null);
 
+    public static final ConfigKey<Boolean> VMsnapshotKVM = new ConfigKey<>(Boolean.class, "kvm.vmstoragesnapshot.enabled", "Snapshots", "false", "For live snapshot of virtual machine instance on KVM hypervisor without memory. Requieres qemu version 1.6+ (on NFS or Local file system) and qemu-guest-agent installed on guest VM", false, ConfigKey.Scope.Global, null);
```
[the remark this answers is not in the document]
GabrielBrascher: Dynamically normally is a good option considering that it is not necessary to reboot the management server to load the configuration. In some cases, it is better to restart the management node though.

To set it to be dinamically it is necessary to create a ConfigKey with the boolean before `ConfigKey.Scope` as `true`.

For instance:
```
public static final ConfigKey<Boolean> VMsnapshotKVM = new ConfigKey<>(Boolean.class, "kvm.vmstoragesnapshot.enabled", "Snapshots", "false", "For live snapshot of virtual machine instance on KVM hypervisor without memory. Requieres qemu version 1.6+ (on NFS or Local file system) and qemu-guest-agent installed on guest VM", true, ConfigKey.Scope.Global, null); 
```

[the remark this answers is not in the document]
slavkap: Thanks! I didn't know that. I will change it to dynamic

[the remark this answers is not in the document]
slavkap: have set the configuration setting to dynamic

GabrielBrascher (approved):
LGTM based on code and manual tests.

api/src/main/java/com/cloud/storage/Snapshot.java
```diff
@@ -48,7 +48,7 @@ public boolean equals(String snapshotType) {
     }
 
     public enum State {
-        Allocated, Creating, CreatedOnPrimary, BackingUp, BackedUp, Copying, Destroying, Destroyed,
+        Allocated, AllocatedKVM, Creating, CreatingForVM, CreatedOnPrimary, CreatedOnPrimaryForVM, BackingUp, BackingUpForVM, BackedUp, BackedUpForVM, Copying, Destroying, Destroyed,
```
sureshanaparti, line 51: @slavkap Snapshot states should be generic, irrespective of hypervisor. Any other thoughts?

api/src/main/java/com/cloud/vm/snapshot/VMSnapshot.java
```diff
@@ -31,7 +31,7 @@
     enum State {
         Allocated("The VM snapshot is allocated but has not been created yet."), Creating("The VM snapshot is being created."), Ready(
                 "The VM snapshot is ready to be used."), Reverting("The VM snapshot is being used to revert"), Expunging("The volume is being expunging"), Removed(
-                "The volume is destroyed, and can't be recovered."), Error("The volume is in error state, and can't be recovered");
+                "The volume is destroyed, and can't be recovered."), CreatingKVM("Creating KVM VM snapshot"), Error("The volume is in error state, and can't be recovered");
```
sureshanaparti, line 34: VM snapshot states should be also  generic. Can this be addressed in any other way?

api/src/main/java/com/cloud/storage/Snapshot.java
slavkap: Hello @sureshanaparti  , thanks for the review! The design of the current implementation was to change as less as possible in CloudStack, that's why we are reusing a lot of the functionality. The additional snapshot's states are added to distinguish with the existing functionality of VM snapshots. In this PR we are lying to CloudStack that a VM snapshot is created, actually we create a separate snapshot for each disk of the VM (and they are consistent, because of the freeze of VM). And the additional states are needed to forbid the user at the end to delete one of the volume's snapshots (by mistake or on purpose). 
I'd be glad if you can give me an advice with another option

api/src/main/java/com/cloud/vm/snapshot/VMSnapshot.java
slavkap: I'll check if I can omit those changes, thanks @sureshanaparti

api/src/main/java/com/cloud/storage/Snapshot.java
GabrielBrascher: @sureshanaparti @slavkap changing from _AllocatedKVM_ to _AllocatedVM_ would be better indeed. What do you think @sureshanaparti?

slavkap: @GabrielBrascher I've renamed it to AllocatedVM, @sureshanaparti is it OK to change only the name or I have to think to implement it in different way?

api/src/main/java/com/cloud/vm/snapshot/VMSnapshot.java
slavkap: removed the new state and events related to it

DaanHoogland:
@wido all your comments seem addressed. Are you fine with this?
@blueorangutan package

DaanHoogland:
@blueorangutan test

slavkap:
> Only failed tests results shown below:
> 
> Test	Result	Time (s)	Test File
> test_01_create_vm_snapshots	`Error`	32.64	test_vm_snapshot_kvm.py
> test_02_revert_vm_snapshots	`Failure`	30.16	test_vm_snapshot_kvm.py

@DaanHoogland I'll look at the logs, but is there a way that I can see the management log, to check the exception?

DaanHoogland:
> > Only failed tests results shown below:
> > Test	Result	Time (s)	Test File
> > test_01_create_vm_snapshots	`Error`	32.64	test_vm_snapshot_kvm.py
> > test_02_revert_vm_snapshots	`Failure`	30.16	test_vm_snapshot_kvm.py
> 
> @DaanHoogland I'll look at the logs, but is there a way that I can see the management log, to check the exception?

it should be included in the zip, https://github.com/blueorangutan/acs-prs/releases/download/trillian/pr3724-t3143-kvm-centos7.zip

slavkap:
Unfortunately I could not reproduce the failure. According to the logs the create snapshot has failed on execution of qemu-monitor-command "drive-backup". I could add to log the result, but not sure if you rerun the tests it will fail again. If I add the result in log, is it possible to run only this test - test_vm_snapshot_kvm.py?

DaanHoogland:
> Unfortunately I could not reproduce the failure. According to the logs the create snapshot has failed on execution of qemu-monitor-command "drive-backup". I could add to log the result, but not sure if you rerun the tests it will fail again. If I add the result in log, is it possible to run only this test - test_vm_snapshot_kvm.py?

not from here, but i started it from our lab. will let you know.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@slavkap the env was destroyed so running the full suit. if you want to run just that test, you are a bit on your own.
@blueorangutan test

slavkap:
> @slavkap the env was destroyed so running the full suit. if you want to run just that test, you are a bit on your own.

Thanks @DaanHoogland I did not have bad intentions with the question whatever you could rerun only this test. I just wanted to check the result without to wait all tests to be executed. And if it fails again with this setup, what is the reason. Anyway I know that I'm on my own, but I've executed this test so many times on my dev, without any failures

DaanHoogland:
> > @slavkap the env was destroyed so running the full suit. if you want to run just that test, you are a bit on your own.
> 
> Thanks @DaanHoogland I did not have bad intentions with the question whatever you could rerun only this test. I just wanted to check the result without to wait all tests to be executed. And if it fails again with this setup, what is the reason. Anyway I know that I'm on my own, but I've executed this test so many times on my dev, without any failures

We don't have that limiting feature but I am working on getting the testrun smaller, that won't be for 4.15 though

slavkap:
I've checked the error and it couldn't create a VM because of this:

> 2020-11-12 16:44:36,206 DEBUG [c.c.a.m.a.i.FirstFitAllocator] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91 FirstFitRoutingAllocator) (logid:144793be) Host Allocator returning 0 suitab
> le hosts
> 2020-11-12 16:44:36,206 DEBUG [c.c.d.DeploymentPlanningManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) No suitable hosts found
> 2020-11-12 16:44:36,206 DEBUG [c.c.d.DeploymentPlanningManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) No suitable hosts found under this Cluster: 1
> 2020-11-12 16:44:36,207 DEBUG [c.c.d.DeploymentPlanningManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Could not find suitable Deployment Destination for t
> his VM under any clusters, returning. 
> 2020-11-12 16:44:36,208 DEBUG [c.c.d.FirstFitPlanner] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Searching resources only under specified Cluster: 1
> 2020-11-12 16:44:36,208 DEBUG [c.c.d.FirstFitPlanner] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) The specified cluster is in avoid set, returning.
> 2020-11-12 16:44:36,212 DEBUG [c.c.d.DeploymentPlanningManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) DeploymentPlanner allocation algorithm: null
> 2020-11-12 16:44:36,212 DEBUG [c.c.d.DeploymentPlanningManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Trying to allocate a host and storage pools from dc:
> 1, pod:1,cluster:null, requested cpu: 500, requested ram: (512.00 MB) 536870912
> 2020-11-12 16:44:36,212 DEBUG [c.c.d.DeploymentPlanningManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Is ROOT volume READY (pool already allocated)?: No
> 2020-11-12 16:44:36,212 DEBUG [c.c.d.DeploymentPlanningManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Deploy avoids pods: [], clusters: [1], hosts: [1, 2]
> 2020-11-12 16:44:36,213 DEBUG [c.c.d.FirstFitPlanner] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Searching resources only under specified Pod: 1
> 2020-11-12 16:44:36,214 DEBUG [c.c.d.FirstFitPlanner] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Listing clusters in order of aggregate capacity, that have (at lea
> st one host with) enough CPU and RAM capacity under this Pod: 1
> 2020-11-12 16:44:36,216 DEBUG [c.c.d.FirstFitPlanner] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) Removing from the clusterId list these clusters from avoid set: [1
> ]
> 2020-11-12 16:44:36,218 DEBUG [c.c.d.FirstFitPlanner] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) No clusters found after removing disabled clusters and clusters in
>  avoid list, returning.
> 2020-11-12 16:44:36,228 DEBUG [c.c.c.CapacityManagerImpl] (Work-Job-Executor-58:ctx-8d676e31 job-2472/job-2475 ctx-e703fb91) (logid:144793be) VM state transitted from :Starting to Stopped with event: Oper
> ationFailedvm's original host id: null new host id: null host id before state transition: 1

rhtyd:
@slavkap sorry it hadn't got a lot of attention, let's try and fix that. Is the PR ready for present `main`, do you need to make any further improvements.
@blueorangutan package

slavkap:
Thanks, @rhtyd, that you've reopened a discussion for this! I'll have to rebase and test it again with what's in the main branch.

rhtyd:
@blueorangutan test

slavkap:
Hi @rhtyd, @DaanHoogland, sorry to bother you both, but what are the versions of qemu/ libvirt on the agents? I'm pretty sure that creating snapshot was passing before, but now it can't find the command 'drive-backup', which is included in qemu-kvm-ev

DaanHoogland:
> Hi @rhtyd, @DaanHoogland, sorry to bother you both, but what are the versions of qemu/ libvirt on the agents? I'm pretty sure that creating snapshot was passing before, but now it can't find the command 'drive-backup', which is included in qemu-kvm-ev

I didn't review @slavkap , is that a new command used? If so it means we are putting more limits on what versions can be used. I don't know the version, I'll have a look in some running env.

slavkap:
@DaanHoogland, thanks! For CentOS, qemu drive-backup is included in the `qemu-kvm-ev` package, and maybe I'm wrong that the test passed once in the past. I remember a discussion that you didn't want to set `qemu-kvm-ev` on Trillian, but I'm not sure.

DaanHoogland:
@slavkap, the installed packages;
```
# yum list | grep qemu
ipxe-roms-qemu.noarch                    20180825-3.git133f4c.el7       @base   
libvirt-daemon-driver-qemu.x86_64        4.5.0-36.el7_9.5               @updates
qemu-img.x86_64                          10:1.5.3-175.el7_9.4           @updates
qemu-kvm.x86_64                          10:1.5.3-175.el7_9.4           @updates
qemu-kvm-common.x86_64                   10:1.5.3-175.el7_9.4           @updates
```
```
# yum list | grep libvirt
libvirt.x86_64                           4.5.0-36.el7_9.5               @updates
libvirt-bash-completion.x86_64           4.5.0-36.el7_9.5               @updates
libvirt-client.x86_64                    4.5.0-36.el7_9.5               @updates
libvirt-daemon.x86_64                    4.5.0-36.el7_9.5               @updates
libvirt-daemon-config-network.x86_64     4.5.0-36.el7_9.5               @updates
libvirt-daemon-config-nwfilter.x86_64    4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-interface.x86_64   4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-lxc.x86_64         4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-network.x86_64     4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-nodedev.x86_64     4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-nwfilter.x86_64    4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-qemu.x86_64        4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-secret.x86_64      4.5.0-36.el7_9.5               @updates
libvirt-daemon-driver-storage.x86_64     4.5.0-36.el7_9.5               @updates
...
libvirt-daemon-driver-storage-rbd.x86_64 4.5.0-36.el7_9.5               @updates
...
libvirt-libs.x86_64                      4.5.0-36.el7_9.5               @updates
python36-libvirt.x86_64                  4.5.0-1.el7                    @epel   
```

slavkap:
Thank you, @DaanHoogland. Then it's normal mine Marvin's tests to fail. Do I have to remove the tests from this PR or skip them?

DaanHoogland:
It's allright to test for requirements and execute based on availability, so by no means remove them, @slavkap .
My biggest worry is the big interval of functionality differences we have based on qemu versions and how we can handle those in a operator friendly way.

slavkap:
Thanks, @DaanHoogland. I will add these checks depending on OS and qemu version

slavkap:
@blueorangutan package

sureshanaparti:
@blueorangutan test

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

slavkap:
Hi all, this PR is ready for review. I'll appreciate it if somebody has time to test it.
According to the test results, I think they are not related to this PR

nvazquez:
Thanks @slavkap, will schedule some time for testing it. Quick question: I've noticed this PR was tested on 4.14 could you test it on the current main branch as well?

slavkap:
Thanks, @nvazquez, that you will schedule this for testing! Actually, I'm testing it whenever I push something new

nvazquez:
@weizhouapache can you please test this PR?

weizhouapache:
@blueorangutan package

weizhouapache:
@slavkap @nvazquez 
I believe this pr is good.

Unfortunately when I tested on ubuntu 20.04, it fails.

the reason is, ubuntu 20.04 uses -blockdev instead -device when start a vm via libvirt.
```
# on ubuntu 20.04
# virsh qemu-monitor-command i-2-85-VM '{"execute":"query-block"}' --pretty |jq -r ".return[].device"

```
same issue as http://qemu.11.n7.nabble.com/query-block-command-is-missing-device-name-Deprecated-td714506.html
I tried with code change below, still does not work
```
diff --git a/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java b/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java
index 8461d31f08..6766291eaf 100644
--- a/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java
+++ b/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java
@@ -57,6 +57,9 @@ public class QemuCommand {
                     if (arr.get(i).getAsJsonObject().has("inserted") && arr.get(i).getAsJsonObject().get("inserted").isJsonObject()) {
                         JsonObject inserted = arr.get(i).getAsJsonObject().get("inserted").getAsJsonObject();
                         if (inserted.get("file").getAsString().contains(path)) {
+                            if (org.apache.commons.lang.StringUtils.isBlank(deviceName)) {
+                                return inserted.get("node-name").getAsString();
+                            }
                             return deviceName;
                         }
                     }
```

on centos7 it works.

```
# on centos
# virsh qemu-monitor-command i-2-5-VM '{"execute":"query-block"}' --pretty |jq -r '.return[].device'
drive-virtio-disk0
drive-ide0-1-0
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
```diff
@@ -1525,22 +1606,62 @@ public Answer createSnapshot(final CreateObjectCommand cmd) {
...
-                final Object[] args = new Object[] {snapshotName, vmUuid};
-                final String snapshot = SnapshotXML.format(args);
+                boolean isKVMEnabled = cmd.getContextParam("kvmsnapshot") != null ? Boolean.parseBoolean(cmd.getContextParam("kvmsnapshot")) : false;
+                s_logger.debug(String.format("Snapshots on KVM is enabled %s", isKVMEnabled));
+                if (isKVMEnabled) {
+                    long size = 0;
+                    String queryBlockResult = vm.qemuMonitorCommand(new Gson().toJson(QemuCommand.executeQemuCommand(QemuCommand.QEMU_BLOCK, null)).toString(), 0);
+                    String path = disk.getPath();
+                    String snapshotDestPath = primaryPool.getLocalPath() + File.separator + "tmp";
+                    if (queryBlockResult != null) {
```
weizhouapache, line 1616: @slavkap it is better to use org.apache.commons.lang.StringUtils.isNotBlank in line 1616 (queryBlockResult) and line 1618 (deviceName)

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
slavkap: Thanks, @weizhouapache, for the code review! I'll change this

slavkap:
Thank you, @weizhouapache, for testing this PR! I will research for a fix of that problem, but unfortunately, all of my dev. environments are with centos7

slavkap:
@weizhouapache, can you share the error you've got with your change of `device` to `node-name`?

weizhouapache:
> @weizhouapache, can you share the error you've got with your change of `device` to `node-name`?

@slavkap the error is same as posted on http://qemu.11.n7.nabble.com/query-block-command-is-missing-device-name-Deprecated-td714506.html

```
2021-07-16 19:25:02,243 DEBUG [c.c.v.s.VMSnapshotManagerImpl] (Work-Job-Executor-1:ctx-0db99c99 job-684/job-685 ctx-d351d56e) (logid:cee7bf47) Failed to create vm snapshot: 8
com.cloud.utils.exception.CloudRuntimeException: Failed to take snapshot: {"id":"libvirt-2170","error":{"class":"GenericError","desc":"Invalid job ID ''"}}
	at org.apache.cloudstack.storage.vmsnapshot.StorageVMSnapshotStrategy.takeVMSnapshot(StorageVMSnapshotStrategy.java:234)
	at com.cloud.vm.snapshot.VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VMSnapshotManagerImpl.java:566)
	at com.cloud.vm.snapshot.VMSnapshotManagerImpl.orchestrateCreateVMSnapshot(VMSnapshotManagerImpl.java:1249)
```

slavkap:
Thanks, @weizhouapache. I think that the problem is fixed. I've tested it only on agents with CentOS7 but did the tests with qemu-monitor-command on my workstation (Ubuntu), and it's working with the solution from the link you shared

weizhouapache:
> Thanks, @weizhouapache. I think that the problem is fixed. I've tested it only on agents with CentOS7 but did the tests with qemu-monitor-command on my workstation (Ubuntu), and it's working with the solution from the link you shared

@slavkap great.
this is tested ok on ubuntu 20.04. @nvazquez

weizhouapache (approved):
approved.

@slavkap a small bug on ui: vm snapshot revert should be disabled if vm is running, hypervisor is KVM and vm snapshot type is Disk. you can fix it after the pr is merged. @nvazquez

slavkap:
Thank you, @weizhouapache, for your time to test this!

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

nvazquez:
@davidjumani please check if you can create a separate PR for the UI issue

nvazquez:
@sureshanaparti can you please review?

api/src/main/java/com/cloud/storage/Snapshot.java
```diff
@@ -48,7 +48,7 @@ public boolean equals(String snapshotType) {
     }
 
     public enum State {
-        Allocated, Creating, CreatedOnPrimary, BackingUp, BackedUp, Copying, Destroying, Destroyed,
+        Allocated, AllocatedVM, Creating, CreatingForVM, CreatedOnPrimary, CreatedOnPrimaryForVM, BackingUp, BackingUpForVM, BackedUp, BackedUpForVM, Copying, Destroying, Destroyed,
```
sureshanaparti, line 51: @slavkap Some questions.
- Does the snapshot operation here triggered through VM snapshot operation (as there are changes in VM snapshot strategies)?
- Are the actual volume snapshots taken (for the volumes of the VM) are maintained at volume level (in snapshots table)?
- Does the snapshots taken here are listed with VM snapshots or volume snapshots?

If the snapshots here are VM level snapshots, i think it is better to move these new states (and associated state transitions) to  VM Snapshot object.

slavkap: > * Does the snapshot operation here triggered through VM snapshot operation (as there are changes in VM snapshot strategies)?

Yes, it's triggered through the VM snapshot operation.
> 
> * Are the actual volume snapshots taken (for the volumes of the VM) are maintained at volume level (in snapshots table)?
> 

They are maintained mostly at the VM level. You will see a volume snapshot, but you don't have any options to operate separately on it.

> * Does the snapshots taken here are listed with VM snapshots or volume snapshots?

You'll see both a VM snapshot and also a volume snapshot for each volume of the group snapshot.

image
image

sureshanaparti: > > * Does the snapshot operation here triggered through VM snapshot operation (as there are changes in VM snapshot strategies)?
> 
> Yes, it's triggered through the VM snapshot operation.
> 
> > * Are the actual volume snapshots taken (for the volumes of the VM) are maintained at volume level (in snapshots table)?
> 
> They are maintained mostly at the VM level. You will see a volume snapshot, but you don't have any options to operate separately on it.
> 
> > * Does the snapshots taken here are listed with VM snapshots or volume snapshots?
> 
> You'll see both a VM snapshot and also a volume snapshot for each volume of the group snapshot.
> 
> image
> image

@slavkap As no operations can be performed on the volume snapshots (of the VM) and these are read-only, I feel it is better to keep the snapshot as VM snapshot only and maintain the relation of these volume snapshots to VM snapshot at the backend (in DB). Also, better to move these new states (and associated state transitions) to VM Snapshot object.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
We can rule out suse failure - likely jenkins build issue.
@blueorangutan test

rhtyd:
Did it overwrite packaging, why do we've three comments cc @davidjumani ?
@blueorangutan package

rhtyd:
@blueorangutan test matrix

rhtyd:

rhtyd:
Tests LGTM. Ping @sureshanaparti can you help review this PR?

DaanHoogland:
@slavkap can you add a javadoc to the most important classes using the term freeze-thaw to explain what is meant by that term in this context, please?

sureshanaparti:
@slavkap As discussed offline, Can you try to create the VM Snapshot as a volume group snapshot (with different snapshot type), maintaining the same (current) states of snapshot. This way, the same snapshots state machine, db table can be re-used and try not to list the snapshots which are part of volume group snapshot.

slavkap:
Thank you, @sureshanaparti, for your help and time! I marked the PR as a draft until I remove the new states and do some changes. Also, I will remove the functionality for the NFS/Local storage. It's hard to set it up on CentOS hypervisors. The PR proposed by Daniel [5297](https://github.com/apache/cloudstack/pull/5297) is cleaner, and it's better to use it for NFS/Local storage.

rhtyd:
@slavkap just to confirm, is this PR on-hold until your changes or are you saying we should go ahead with reviewing/testing https://github.com/apache/cloudstack/pull/5297 ?

slavkap:
Hi @rhtyd, we discussed with @sureshanaparti that I should make a few changes to the snapshot states. I'm working on them, and it will be great if #5297 gets in first. With it, I won't limit this PR with the functionality for NFS

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

rp-:
I tested this PR today with the Linstor plugin and creating/reverting and deleting works with Linstor.

nvazquez:
Thanks @rp- can you please approve on the Files changed tab -> Submit review?

rp- (approved):
Tested with Linstor primary storage:
create/revert and delete worked.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class QemuCommand {
+    //Qemu agent commands
+    public static final String AGENT_FREEZE = "guest-fsfreeze-freeze";
+    public static final String AGENT_THAW = "guest-fsfreeze-thaw";
+    public static final String AGENT_FREEZE_STATUS = "guest-fsfreeze-status";
+    //Qemu monitor commands
+    public static final String QEMU_QUERY_BLOCK_JOBS = "query-block-jobs";
```
nvazquez, line 30: These 3 are not used, could be removed?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java
```diff
@@ -0,0 +1,44 @@
...
+    public static final String QEMU_BLOCK = "query-block";
+    public static final String QEMU_DRIVE_BACKUP = "drive-backup";
+
+    public static final String QEMU_CMD = "execute";
+
+    public static Map<String, Object> executeQemuCommand(String command, Map<String, String> args ){
```
nvazquez, line 36: Can this be renamed? It is a but misleading since nothing is executed at this time

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFreezeThawVMCommandWrapper.java
```diff
@@ -0,0 +1,104 @@
...
+                    s_logger.trace("Ingore error ", e);
+                }
+            }
+        }
+    }
+
+    private String getResultOfQemuCommand(String cmd, Domain domain) throws LibvirtException {
+        String result = null;
+        if (cmd.equals(FreezeThawVMCommand.FREEZE)) {
+            result = domain.qemuAgentCommand(new Gson().toJson(QemuCommand.executeQemuCommand(QemuCommand.AGENT_FREEZE, null)).toString(), 10, 0);
```
nvazquez, line 96: I think readability could be increased on these 3 calls by making the `QemuCommand.executeQemuCommand` (or renamed) method return a string instead of a map.

Also, these 3 calls could be reduced to only 1, by defining a collection that returns the agent "guest-fsfreeze-*" command based on the `cmd`

server/src/main/java/com/cloud/storage/snapshot/SnapshotManager.java
```diff
@@ -59,6 +59,8 @@
     public static final ConfigKey<Boolean> BackupSnapshotAfterTakingSnapshot = new ConfigKey<Boolean>(Boolean.class, "snapshot.backup.to.secondary",  "Snapshots", "true",
             "Indicates whether to always backup primary storage snapshot to secondary storage. Keeping snapshots only on Primary storage is applicable for KVM + Ceph only.", false, ConfigKey.Scope.Global, null);
 
+    public static final ConfigKey<Boolean> VMsnapshotKVM = new ConfigKey<>(Boolean.class, "kvm.vmstoragesnapshot.enabled", "Snapshots", "false", "For live snapshot of virtual machine instance on KVM hypervisor without memory. Requieres qemu version 1.6+ (on NFS or Local file system) and qemu-guest-agent installed on guest VM", true, ConfigKey.Scope.Global, null);
```
nvazquez, line 62: Could it be renamed to something like 'VMStorageSnapshotKVM'? Could the scope be reduced to zone?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuCommand.java
slavkap: Thanks, @nvazquez, for the review! I will rename it

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFreezeThawVMCommandWrapper.java
slavkap: @nvazquez, I prefer to leave it like this. In the feature, we could reuse this for commands that accept arguments, and for me, it will be more readable than assembling different strings.

server/src/main/java/com/cloud/storage/snapshot/SnapshotManager.java
slavkap: @nvazquez, I will rename it, but why limit this to a Zone?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFreezeThawVMCommandWrapper.java
nvazquez: @slavkap ok, since qemuAgentCommand accepts a string as a first parameter I would be +1 on having a helper class or similar that will retrieve the string in the expected format based on the params, but not a blocker

server/src/main/java/com/cloud/storage/snapshot/SnapshotManager.java
nvazquez: @slavkap I was thinking it would make sense for admins to enable/disable the feature for a zone and not for all of them at the same time - just a suggestion

nvazquez (approved):
Thanks @slavkap - code LGTM

nvazquez:
Hi @wido @GabrielBrascher @svenvogel would it be possible for you to test this PR on Ceph?

nvazquez:
@blueorangutan test

nvazquez:
Merging based on approvals and tests results

rohityadavcloud:
Fantastic, this is finally merged! Thank for your work and patience @slavkap - do raise a documentation PR to as required - https://github.com/apache/cloudstack-documentation

Issue #3683, opened 2019-11-13
Stop "mgmt server needs restart" message on global setting that doesn't require it


andrijapanicsb:
Currently, when we change ANY global setting, there is a popup window saying "Please restart your management server(s) for your new settings to take effect"

But on i.e. 4.11 release, Out of  589 different settings, 188 of those do not require mgmt server restart. (SELECT * FROM cloud.configuration WHERE is_dynamic=1;)

Would be nice to stop that false popup for those global settings that do not require mgmt server restart.

weizhouapache:
@andrijapanicsb we have a fix for this issue
I will port it to 4.13 and create a pull request.

andrijapanicsb:
thanks (as usual) @weizhouapache !

weizhouapache:
@andrijapanicsb 
I have created the pull request #3729 
could you please test it when you have time ? thanks !

PR #3729, opened 2019-11-29
config: add isdynamic flag in configuration response

ustcweizhou:
## Description

Some configurations are dynamic so it does not require restart of mgt server when we change it.
Add isdynamic in configuration response, and does not pop up dialog "restart mgt server" if config is dynamic on UI.

Fixes: #3683 

## Types of changes

## How Has This Been Tested?

Tested with global configuration "xen.vm.vcpu.max", it works fine (no pop up dialog).

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
+++ b/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
@@ -187,6 +187,7 @@ public class ApiConstants {
     public static final String IP_LIMIT = "iplimit";
     public static final String IP_TOTAL = "iptotal";
     public static final String IS_CLEANUP_REQUIRED = "iscleanuprequired";
+    public static final String IS_DYNAMIC = "isdynamic";
     public static final String IS_EXTRACTABLE = "isextractable";
     public static final String IS_FEATURED = "isfeatured";
     public static final String IS_PORTABLE = "isportable";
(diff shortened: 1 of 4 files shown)


DaanHoogland:
@blueorangutan package

andrijapanicsb:
thx Wei.

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
LGTM

Tested manually on 10-15 different settings.

rhtyd:
This will largely work for most global settings, however, there are configkeys that are defined to not be dynamic in nature and sometimes configkey which are are dynamic are used while creating a scheduled thread or executor service whose values only apply when a cloudstack component (managerbase etc.) is configured but not during/after the component has started. To accept this, will require codebase wide search and changing any configkeys which are declared as dynamic but not used dynamically.

weizhouapache:
@rhtyd 
this change is based on the assumption that the "dynamic" is set correctly for the configurations.
We need to check all global configurations , especially the configurations which were added in Config.java before ConfigKey is introduced in cloudstack. it is another storage I think.

DaanHoogland:
good catch @rhtyd , however, will this be harmful? Or can we merge and plan the small fixes that will follow as they come up?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
cc @PaulAngus this adds feature to know if restarting mgmt server is needed after updating a global setting. I'll see what we can do to detect/fix this in Primate.

DaanHoogland:
discussed with @rohit offline. this is a significant improvement over the current situation in spite of his remarks. We'll need to deal with the 10% of exceptions as we encounter them.

PR #3736, opened 2019-12-03
Add protocol number support for security group rules

ravening:
## Description

Currently while creating ingress/egress rule for a security group,
we can specify only TCP/UDP/ICMP. Sometimes we need to add rules
for different protocol numbers or rules for all the above three
mentioned protocols.

In this new feature, users can specify the protocol number or select
"ALL" option which will apply rules for TCP/UDP/ICMP in one shot

## Types of changes

Navigate to "Networks" and select "Security groups" from the drop-down menu on the top left
Select any security group name and click on either "ingress" or "egress" tab

Screenshot 2019-12-03 at 09 44 05

Screenshot 2019-12-03 at 09 45 11

Screenshot 2019-12-03 at 09 45 44

## How Has This Been Tested?

Adding a rule for a particular protocol number

1. Create a VM and note down the security group name to which it belongs to
2. Navigate to "Network" and select "Security groups" from the drop-down menu on the top left.
3. Select the security group which is obtained from step 1.
4. Click on either ingress or egress tab
5. Try to create a new rule by selecting "Protocol number" from the drop-down list.
     For ex: Enter protocol number as 47 and cidr as "99.99.99.99/32"

Screenshot 2019-12-03 at 10 12 57

6. Now get the hostname on which the VM is running and ssh to it.
7. Run the iptables-save command to ensure that there is a new rule for the protocol number you mentioned above.
    The output should be as shown below

```
# iptables-save | grep 99.99.99
-A i-2-118-VM -s 99.99.99.99/32 -p gre -j ACCEPT
```

Adding a rule to all TCP/UDP/ICMP at a time

1. Create a VM and note down the security group name to which it belongs to
2. By default, there will be a drop action for the three protocols.
2. Navigate to "Network" and select "Security groups" from the drop-down menu on the top left.
3. Select the security group which is obtained from step 1.
4. Click on either ingress or egress tab
5. Try to create a new rule by selecting "ALL" from the drop-down list.

Screenshot 2019-12-03 at 10 17 06

6. Now get the hostname on which the VM is running and ssh to it.
7. Run the iptables-save command to ensure that the below rule should be applied.

```
# iptables-save | grep i-2-118-VM
-A i-2-118-VM -s 88.88.88.88/32 -m state --state NEW -j ACCEPT
```

New integration test has been added in ```test/integration/component/test_protocol_number_security_group.py```

Integration test can be run using

```nosetests --with-marvin --marvin-config=<config file> test/integration/component/test_protocol_number_security_group.py```

### DIFF
--- a/scripts/vm/network/security_group.py
+++ b/scripts/vm/network/security_group.py
@@ -1032,7 +1032,18 @@ def add_network_rules(vm_name, vm_id, vm_ip, vm_ip6, signature, seqno, vmMac, ru
                 action = "ACCEPT"
                 direction = "-s"
 
-            range = str(start) + ':' + str(end)
+            if start == 0 and end == 0:
+                dport = ""
+            else:
+                dport = " --dport " + str(start) + ":" + str(end)
+
+            if protocol != 'all' and protocol != 'icmp' and protocol != 'tcp' and protocol != 'udp':
+                protocol_all = " -p " + protocol
+                protocol_state = " "
+            else:
+                protocol_all = " -p " + protocol + " -m " + protocol
+                protocol_state = " -m state --state NEW "
+
             if 'icmp' == protocol:
                 range = str(start) + '/' + str(end)
                 if start == -1:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: scripts/vm/network/security_group.py [hunk 1]
```diff
@@ -1032,7 +1032,18 @@ def add_network_rules(vm_name, vm_id, vm_ip, vm_ip6, signature, seqno, vmMac, ru
                 action = "ACCEPT"
                 direction = "-s"
 
-            range = str(start) + ':' + str(end)
+            if start == 0 and end == 0:
+                dport = ""
+            else:
+                dport = " --dport " + str(start) + ":" + str(end)
+
+            if protocol != 'all' and protocol != 'icmp' and protocol != 'tcp' and protocol != 'udp':
```
wido: For ip6tables it's not 'icmp' which is used for the protocol, it's icmpv6, so don't we need to add a check there?

rakgenius: In the UI, we only have options for "TCP", "UDP", "ICMP" and "ALL". There is no option for "ICMPv6" in UI. Once this is added in ui, we can have that check here as well.

rakgenius:
@wido @rhtyd @GabrielBrascher guys, can you review this when you get time?

scripts/vm/network/security_group.py
wido: @rakgenius I'm aware, but underneath we translate ICMP to ICMPv6 when the CIDR/IP is an IPv6 adress. That's what we need to check.

weizhouapache: @wido the protocol is "icmp" in cloudstack, no matter the ip/cidr is ipv4 or ipv6.

look at the code below
https://github.com/apache/cloudstack/pull/3736/files/c27fdfee8c9f5ab06cf18dd6dc41ba356d75f4c4#diff-d10df45583c471e8de4a64128f95c802R1066-R1070

weizhouapache:
@DaanHoogland this PR has two approvals. could you kick off a trillian test ?

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

PR #3738, opened 2019-12-03
Load Average for KVM

DennisKonrad:
## Description
This adds the **load average for the kvm hypervisor**. The average Load already seems to be available for Xen.

In the _listHosts_ the [averageload](https://cloudstack.apache.org/api/apidocs-4.13/apis/listHosts.html) is already present. In _listHostMetrics_ it appears as "cpuloadaverage".

It adds the 1-min load average from /proc/loadavg to the
_listHosts
listHostsMetrics_
API calls.

image

## Types of changes

## How Has This Been Tested?
Built and tested in kvm cluster.
Tested manually via API calls

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
@@ -116,7 +116,7 @@ public class HostResponse extends BaseResponse {
 
     @SerializedName("averageload")
     @Param(description = "the cpu average load on the host")
-    private Long averageLoad;
+    private Double averageLoad;
 
     @SerializedName("networkkbsread")
     @Param(description = "the incoming network traffic on the host")
@@ -337,7 +337,7 @@ public class HostResponse extends BaseResponse {
         this.cpuUsed = cpuUsed;
     }
 
-    public void setAverageLoad(Long averageLoad) {
+    public void setCpuAverageLoad(Double averageLoad) {
         this.averageLoad = averageLoad;
     }
 
@@ -570,7 +570,7 @@ public class HostResponse extends BaseResponse {
         return cpuUsed;
     }
 
-    public Long getAverageLoad() {
+    public Double getAverageLoad() {
         return averageLoad;
     }
 
--- a/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
@@ -114,9 +114,9 @@ public class HostResponse extends BaseResponse {
     @Param(description = "the amount of the host's CPU after applying the cpu.overprovisioning.factor ")
     private String cpuWithOverprovisioning;
 
-    @SerializedName("cpuloadaverage ")
+    @SerializedName("cpuloadaverage")
     @Param(description = "the cpu average load on the host")
-    private Double cpuloadaverage ;
+    private Double cpuloadaverage;
 
     @SerializedName("networkkbsread")
     @Param(description = "the incoming network traffic on the host")
--- a/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
@@ -114,7 +114,7 @@ public class HostResponse extends BaseResponse {
     @Param(description = "the amount of the host's CPU after applying the cpu.overprovisioning.factor ")
     private String cpuWithOverprovisioning;
 
-    @SerializedName("cpuloadaverage")
+    @SerializedName(ApiConstants.CPU_LOAD_AVERAGE)
     @Param(description = "the cpu average load on the host")
     private Double cpuloadaverage;
 
--- a/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
@@ -114,9 +114,9 @@ public class HostResponse extends BaseResponse {
     @Param(description = "the amount of the host's CPU after applying the cpu.overprovisioning.factor ")
     private String cpuWithOverprovisioning;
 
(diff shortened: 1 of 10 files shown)


plugins/metrics/src/main/java/org/apache/cloudstack/response/HostMetricsResponse.java
```diff
@@ -43,6 +43,10 @@
     @Param(description = "the total cpu allocated in Ghz")
     private String cpuAllocated;
 
+    @SerializedName("loadAverage")
+    @Param(description = "the average cpu load the last minute")
+    private String loadAverage;
```
wido: Shouldn't this be a Double?

DennisKonrad: I don't think so. I think it's just for output. The other variables here are either Booleans or Strings

wido: Ah, yes. I misread. This is the API response to the outside world.

GabrielBrascher:
@blueoragutan package

GabrielBrascher:
@blueorangutan package

plugins/metrics/src/main/java/org/apache/cloudstack/response/HostMetricsResponse.java
```diff
@@ -43,6 +43,10 @@
     @Param(description = "the total cpu allocated in Ghz")
     private String cpuAllocated;
 
+    @SerializedName("loadAverage")
```
weizhouapache: is cpuloadaverage better ?

weizhouapache:
@DennisKonrad 
I have tested it , the value it looks fine.
The only problem is, I spent some time to look for the value in the api responses.  it would be better to change them to same name

(local) 🐵 > list hosts filter=averageload
(local) 🐵 > list hostsmetrics filter=loadAverage

plugins/metrics/src/main/java/org/apache/cloudstack/response/HostMetricsResponse.java
DennisKonrad: you are right, will change it to that

DennisKonrad: For easier usage I changed both the results in 
list hosts
list hostsmetrics
to cpuloadaverage.

DennisKonrad:
@weizhouapache I changed both of the keys to cpuloadaverage:

image

image

GabrielBrascher:
@blueorangutan package

DaanHoogland:
I would start a regression suite, but...
Do we need test automation for this?
/cc @weizhouapache @GabrielBrascher @DennisKonrad

GabrielBrascher:
@DaanHoogland I don't think that automation tests are strictly necessary in this case. It is a simple change, and we already have code reviews and one manual tests by @weizhouapache.

I am assuming that the centos6 failing to package is not related to this PR.

wido:
@DennisKonrad Why do the Host Metrics on your screenshots show a String and not a Double?

We should return this in the API as a Double and not a String.

GabrielBrascher:
Good point @wido. It is set as double or Double on `HostStats.java`, `HostResponse.java`, `GetHostStatsAnswer.java`, and `HostStatsEntry.java`. Additionally, at `CPUStat.java` the method `getCpuLoadAverage()` returns Double.

Therefore, it looks OK on the PR. However, `list hostmetrics` returns it as String; `list hosts` looks fine though. I will take a look around the `list hostmetrics`.

On a 5.0 we can normalize API values and fix issues such as % values returned as Strings instead of double.

DennisKonrad:
@wido Good catch. Will change that to output as double

DaanHoogland:
@GabrielBrascher centos6 is broken on a lot (most if not all) PRs. not to worry about that one.
so a package after @DennisKonrad 's next addition will be enough to carry on.

DennisKonrad:
@wido I think it's better now.

image

wido:
Looking good!

api/src/main/java/org/apache/cloudstack/api/response/HostResponse.java
```diff
@@ -114,9 +114,9 @@
     @Param(description = "the amount of the host's CPU after applying the cpu.overprovisioning.factor ")
     private String cpuWithOverprovisioning;
 
-    @SerializedName("averageload")
+    @SerializedName("cpuloadaverage")
```
rhtyd: Can you move this to APIConstants, also changes in the API key response breaks the response compatibility if xen-specific users are using the key for something in their scripts, tool, external integrations. Can you revert the change?

DennisKonrad: It seems to me the guys with xen can't use it anyway:
https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java#L2258-L2272

I will add it to the APIConstanst and leave the API key changed for consistency. Sounds good to you @rhtyd

rhtyd: I'm okay with your rename change if the field is not set previously.

DennisKonrad: Okay, I think [this](https://github.com/apache/cloudstack/pull/3738/commits/e05667692505989001764324a165831420628b05) should do it. I couldn't find any code where the old name is used

GabrielBrascher:
Looking good to merge.
Merging it considering that all checks have passed, we have 4 LGTMs, a couple of manual tests, and the packaging results are ok for Ubuntu and CentOS.

PR #3771, opened 2019-12-17
Revert "Fix virtual template size for managed storage for KVM / refactor cloud-install-sys-tmplt"

andrijapanicsb:
Reverts apache/cloudstack#3371


andrijapanicsb:
Revert

Issue #3544, opened 2019-08-07
When scrolling down the VM settings, they keep on loading again and again endlessly


borisstoyanov:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
Master
~~~

##### SUMMARY

Observed in vm settings and template settings. When settings are a lot, it loads them over and over endlessly when scrolling down

##### STEPS TO REPRODUCE

~~~

~~~

##### EXPECTED RESULTS

~~~

~~~

##### ACTUAL RESULTS

~~~

~~~

DaanHoogland:
#3690 and #3691 relate to this

MNettesheim:
Hi Boris, 
Isn't this about the OVF properties page, rather than the settings page.

borisstoyanov:
Hi @MNettesheim, 
It's about the settings at template and vm, but we're considering the ovf settings at #3690

DaanHoogland:
merged in 4.13 and master

PR #3778, opened 2019-12-19
Endless settings on templates and instances

Spaceman1984:
## Description

When loading the vm template settings from the ui, if there are more than twelve items in the list, the settings are loaded again if you scroll to the last setting.

This fixes the reload of the settings when scrolling to the end of the list.

Fixes: #3544 
## Types of changes

## How Has This Been Tested?

This was tested by going to the settings tab of an instance and a template and adding many settings then navigating away and back and then scrolling down.


ui/scripts/instances.js
```diff
@@ -3437,6 +3437,13 @@
 						custom: cloudStack.uiCustom.granularDetails({
                         resourceType: 'UserVm',
 							dataProvider: function(args) {
+                                // no paging for listVirtualMachines details
+                                if (args.page > 1){
+                                    args.response.success({
+                                        data: []
+                                    });
+                                    return;
+                                }
```
anuragaw: Can you fix indentation @Spaceman1984 ?

ui/scripts/templates.js
```diff
@@ -2489,6 +2489,13 @@
 								custom: cloudStack.uiCustom.granularDetails({
                                     resourceType: 'Template',
 									dataProvider: function(args) {
+                                        // no paging for listTemplates details
+                                        if (args.page > 1){
+                                            args.response.success({
+                                                data: []
+                                            });
+                                            return;
+                                        }
```
anuragaw: same here.

ui/scripts/instances.js
DaanHoogland: What do you mean @anuragaw ? I see no problem with the indentation here.

ui/scripts/templates.js
DaanHoogland: or here?!

ui/scripts/instances.js
anuragaw: Isn't this supposed to be inside dataProvider @DaanHoogland ?
image

Spaceman1984: The indentation is fine when viewing with vscode and intellij, I think this is a github display issue.

DaanHoogland:
@borisstoyanov if you agree i'll merge

ui/scripts/instances.js
anuragaw: What you see in IDEs is actually rendered as per configuration of whitespace. However, this is old UI and I've seen this broken at many files.

anuragaw:
I've fixed indentation on the PR and LGTM

DaanHoogland:
ui issue and 2 approvals, will merge.

Issue #3783, opened 2019-12-20
Create volume or template from snapshot is broken


NuxRo:
Hi,

I'm on 4.13 KVM + local storage, trying to create a volume or a template from a snapshot.
Cloudmonkey says:
```
(localcloud)  > create volume snapshotid=985f20b1-b57e-458e-983d-ce69848fe06c
 Error: (HTTP 530, error code 9999) <nil>
```

management-server.log says:

```
2019-12-20 16:17:58,558 DEBUG [o.e.j.s.HttpChannel] (qtp504527234-813:null) (logid:) HttpChannelOverHttp@5d83425d{r=1,c=false,a=IDLE,uri=//192-168-140-220.cloud.tld:8443/client/api?apiKey=-Rt233Gc8hogf2d-mLvno6fzoTueIja8W_io56YM1Bh4
35Rk6dXex0w_t29X_sv2MMhbuM1k2GbS63chKglc4A&command=createVolume&response=json&snapshotid=d603641c-dff4-43aa-8da1-07afaa0a35d3&signature=zcUn58qI0MLg2y2%2FJC30D8ThDc8%3D} onContentComplete
2019-12-20 16:17:58,558 DEBUG [o.e.j.s.HttpChannel] (qtp504527234-813:null) (logid:) HttpChannelOverHttp@5d83425d{r=1,c=false,a=IDLE,uri=//192-168-140-220.cloud.tld:8443/client/api?apiKey=-Rt233Gc8hogf2d-mLvno6fzoTueIja8W_io56YM1Bh4
35Rk6dXex0w_t29X_sv2MMhbuM1k2GbS63chKglc4A&command=createVolume&response=json&snapshotid=d603641c-dff4-43aa-8da1-07afaa0a35d3&signature=zcUn58qI0MLg2y2%2FJC30D8ThDc8%3D} onRequestComplete
2019-12-20 16:17:58,558 DEBUG [o.e.j.s.HttpInput] (qtp504527234-813:null) (logid:) HttpInputOverHTTP@23b3db24[c=0,q=0,[0]=null,s=STREAM] addContent EOF
2019-12-20 16:17:58,558 DEBUG [o.e.j.s.HttpConnection] (qtp504527234-813:null) (logid:) HttpConnection@12ce2981[p=HttpParser{s=END,0 of -1},g=HttpGenerator@5598102f{s=START}]=>HttpChannelOverHttp@5d83425d{r=1,c=false,a=IDLE,uri=//109-70-
[...]

PR #3786, opened 2019-12-24
[WIP] server: fix checking disk offering access for snapshot volume

shwstppr:
Fixes https://github.com/apache/cloudstack/issues/3783
As reported in the issue, creating volumes from pure snapshot fails with NPE. This is due to order of calls where disk offering access is checked before checking disk offering value. This PR fixes the same.

## Types of changes

## How Has This Been Tested?
In progress


shwstppr:
@blueorangutan package

DaanHoogland:
@shwstppr code looks good, but the bug existed IMHO because of the method being 200 lines long and thus illegible. I will let this pass to honour history but if you have time please factor out the method contents to a helper class or a set of helper methods?
Also, i think we want this on 4.13.
can you give a second look? tnx.

shwstppr:
Closing in favour of https://github.com/apache/cloudstack/pull/3791
@DaanHoogland will address long method refactoring changes there

PR #3803, opened 2020-01-13
Bug fix : set restart_required to 0 after restarting network

ravening:
## Description

After restarting the network with or without cleanup option, the ```restart_required``` field in ```networks``` table should be reset to 0

## Types of changes

## How Has This Been Tested?

1 . Check if the ```restart_required``` field in the ```networks``` table is 1 or not. If not you can set it to 1
2 . Restart the network with or without cleanup option.
3 . The ```restart_required``` field in the ```networks``` table should be reset to 0 once the network restarts successfully.

### DIFF
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -2936,7 +2936,8 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
     @Override
     public boolean restartNetwork(final Long networkId, final Account callerAccount, final User callerUser, final boolean cleanup) throws ConcurrentOperationException, ResourceUnavailableException,
     InsufficientCapacityException {
-
+        boolean status = true;
+        boolean restartRequired = false;
         final NetworkVO network = _networksDao.findById(networkId);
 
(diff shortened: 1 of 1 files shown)


weizhouapache:
@DaanHoogland 
when restart a network (shared/isolated) without cleanup, "Restart Required" will be "Yes" after it. This seems to be a regression issue of your commit 8218ce0fe243d85d0f4cb6280d2310a81c8e99de

Even if we have restarted the network with/without cleanup,  "Restart Required" is still "Yes"

DaanHoogland:
wow @weizhouapache that is from oct 2014. I'm pretty sure some context has changed since then. for instance the zero downtime restart efforts. I don't think that change is relevant to still consider in this fix. preventing setting restart required to false just after setting it to true. probably it isn't needed any more.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan test

andrijapanicsb:
@DaanHoogland this fixes one of our internal issues. Thx @ravening

DaanHoogland:
ok, @andrijapanicsb , cherry-picking this on 4.13, as it won't hurt others

andrijapanicsb:
@ravening will this also fix all networks of the VPC when you restart VPC (not network) with cleanup?
cc/ @weizhouapache

weizhouapache:
@andrijapanicsb in my testing, no...

PR #3807, opened 2020-01-13
Enhancement: Allow creating network with duplicate name

ravening:
Add a global setting to disable creating networks with
same name in an account

Add a global setting to disable creating network without
mentioning the start and end IPv4 or IPv6 address

## Description

By default we can create networks with the same name in the account. Sometimes we should not create the networks with same name. This change adds a global setting which prevents creating the network with same name. The default value is true and set it to false to prevent creating network with same names.

Also its possible to create a shared network without mentioning the start and the end IPv4 or IPv6 address. This change adds a global setting which prevents creating a shared network without specifying the start and the end IPv4 or IPv6 address

## Types of changes

## How Has This Been Tested?

Creating a network with same names
1 . Set the global setting ```allow.duplicate.networkname``` to true
2. Create a new network with the name "test".
3. Try to create another network with the same name. This should create the second network with the same name
4. Now set the above global setting value to false
5. Try to create another network with same name.
6. No exception will be thrown while creating the network with same name

Creating a network without specifying the start and end IPv4 or IPv6 address

1. Set the global setting ```allow.empty.start.end.ipaddress``` to true
2. Create a shared network without mentioning the start and the end IPv4 or IPv6 address
3. You will be able to create network successfully
4. Now set the above global setting value to false
5. Try to create another shared network without mentioning the start or the end IPv4 or IPv6 address
6. An exception will be thrown while creating the shared network

### DIFF
--- a/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
+++ b/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
@@ -55,6 +55,8 @@ import org.apache.cloudstack.api.command.user.vm.ListNicsCmd;
 import org.apache.cloudstack.api.response.AcquirePodIpCmdResponse;
 import org.apache.cloudstack.context.CallContext;
 import org.apache.cloudstack.engine.orchestration.service.NetworkOrchestrationService;
+import org.apache.cloudstack.framework.config.ConfigKey;
+import org.apache.cloudstack.framework.config.Configurable;
 import org.apache.cloudstack.framework.config.dao.ConfigurationDao;
 import org.apache.cloudstack.framework.messagebus.MessageBus;
 import org.apache.cloudstack.framework.messagebus.PublishScope;
@@ -107,6 +109,7 @@ import com.cloud.network.dao.FirewallRulesDao;
 import com.cloud.network.dao.IPAddressDao;
 import com.cloud.network.dao.IPAddressVO;
 import com.cloud.network.dao.LoadBalancerDao;
+import com.cloud.network.dao.NetworkAccountDao;
 import com.cloud.network.dao.NetworkDao;
 import com.cloud.network.dao.NetworkDetailVO;
 import com.cloud.network.dao.NetworkDetailsDao;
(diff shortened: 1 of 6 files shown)


server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -200,9 +203,14 @@
...
  */
-public class NetworkServiceImpl extends ManagerBase implements NetworkService {
+public class NetworkServiceImpl extends ManagerBase implements NetworkService, Configurable {
     private static final Logger s_logger = Logger.getLogger(NetworkServiceImpl.class);
 
+    private static final ConfigKey<Boolean> AllowDuplicateNetworkName = new ConfigKey<Boolean>("Advanced", Boolean.class,
+            "allow.duplicate.networkname", "true", "Allow creating networks with same name in account", true);
+    private static final ConfigKey<Boolean> AllowEmptyStartEndIpAddress = new ConfigKey<Boolean>("Advanced", Boolean.class,
+            "allow.empty.start.end.ipaddress", "true", "Allow creating network without mentioning start and end IP address",
+            true);
```
DaanHoogland: these are global scope only settings? is account level not more applicable or something in between?

ravening: I have made the necessary changes it include the account level setting

DaanHoogland:
@blueorangutan package

ravening:
@DaanHoogland can this be merged?

DaanHoogland:
@blueorangutan package

DaanHoogland:
@ravening "pudding needs eating" so i am starting a full regression suite. (if it comes back good i'm fine with it.)

DaanHoogland:
@blueorangutan test

DaanHoogland:
I am not investing time in testing this now, but regrtession passes on all platforms. feel free to merge, @wido @weizhouapache ?

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -200,9 +203,14 @@
...
  */
-public class NetworkServiceImpl extends ManagerBase implements NetworkService {
+public class NetworkServiceImpl extends ManagerBase implements NetworkService, Configurable {
     private static final Logger s_logger = Logger.getLogger(NetworkServiceImpl.class);
 
+    private static final ConfigKey<Boolean> AllowDuplicateNetworkName = new ConfigKey<Boolean>("Advanced", Boolean.class,
+            "allow.duplicate.networkname", "true", "Allow creating networks with same name in account", true, ConfigKey.Scope.Account);
+    private static final ConfigKey<Boolean> AllowEmptyStartEndIpAddress = new ConfigKey<Boolean>("Advanced", Boolean.class,
+            "allow.empty.start.end.ipaddress", "true", "Allow creating network without mentioning start and end IP address",
+            true, ConfigKey.Scope.Account);
```
rhtyd: @ravening for backward compatibility, should this be false?

ravening: @rhtyd In the current/default behavior we can create network without mentioning the start and end IP address. So the default value is "true". If the default value is "false" then it means we cant create a network with an empty start and end IP address which is not backward compatible.

rhtyd:
@ravening if the newly added test `test/integration/component/test_network_create_with_dup_name.py` does not take more than 5-10mins to run, please move it under `smoke` and also add to Travis (see `.travis.yml).

ravening:
> @ravening if the newly added test `test/integration/component/test_network_create_with_dup_name.py` does not take more than 5-10mins to run, please move it under `smoke` and also add to Travis (see `.travis.yml).

@rhtyd moved the marvin test to `smoke` folder and added an entry in travis.yml as well

rhtyd:
Thanks @ravening I'll merge this as soon as Travis (simulator) passes (rekicked the job).

PR #3808, opened 2020-01-13
reload dnsmasq instead of restart when adding new rules

### LINKED ISSUE #3613: New rules cause dnsmasq to be restarted instead of reloaded
When adding/configuring new dhcp/dns rules, the dnsmasq process is restarted:
https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py#L71

Explore if it's possible to simply reload the dnsmasq process than restart it.

(If dnsmasq restarts at the same time monitoring service check kicks in, it might show to recover the dnsmasq again)

##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
VR
~~~

andrijapanicsb:
Fixes #3613 
Optimistic fix, to be tested.

## Description

Reload dnsmasq when new dhcp rules are added, don't restart.
Avoids "downtime" as observed by VR monitoring scripst in environemnts with lot's of VMs being started/stopped/created in parallel, etc.

## Types of changes

## How Has This Been Tested?

To be tested.

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
@@ -69,7 +69,7 @@ class CsDhcp(CsDataBag):
 
         if not self.cl.is_redundant() or self.cl.is_master():
             if restart_dnsmasq:
-                CsHelper.service("dnsmasq", "restart")
+                CsHelper.service("dnsmasq", "reload")
             else:
                 CsHelper.start_if_stopped("dnsmasq")
                 CsHelper.service("dnsmasq", "reload")


andrijapanicsb:
reload does work fine when /etc/dhcphosts.txt is changed, and it does issue a new IP (changed IP in there for the sake of the quick test, before this PR)

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan test

svenvogel:
@andrijapanicsb i think this should work. not tested but logical for me.

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
damn it monkey...

@blueorangutan test

rhtyd:
I've tried this in the past, but then the issue was that dnsmasq doesn't read all the configs (changed files). @andrijapanicsb please test

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
@blueorangutan test

wido:
Seems good to me.

@rhtyd might your experience be from an older version of dnsmasq and that this has been fixed right now?

andrijapanicsb:
Too many identical failures on 2 runes... will test manually, but it doesn't look good...

systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
```diff
@@ -69,7 +69,7 @@ def process(self):
 
         if not self.cl.is_redundant() or self.cl.is_master():
             if restart_dnsmasq:
-                CsHelper.service("dnsmasq", "restart")
+                CsHelper.service("dnsmasq", "reload")
```
weizhouapache: @andrijapanicsb 
"reload" works only if the service is started.
this is why we have line 74

andrijapanicsb: I'll need to test this all @weizhouapache , but afaik dnsmasq is started when VR is created?
Do you see a problem here? (save me some time from testing all this...)

weizhouapache: > I'll need to test this all @weizhouapache , but afaik dnsmasq is started when VR is created?
@andrijapanicsb it seems dnsmasq is disabled in routers. it will be started when dnsmasq.conf is generated.
https://github.com/apache/cloudstack/blob/master/systemvm/debian/opt/cloud/bin/setup/common.sh#L588-L595


> Do you see a problem here? (save me some time from testing all this...)

andrijapanicsb:
ok, fair enough - closing this one.

PR #3818, opened 2020-01-17
Display numeric value in exception instead of variable name

ravening:
## Description

If the disk size of the VM to be created is greater
then the volume size, then the exception message should
display the numeric value instead of the variable name

## Types of changes

Below is the exception which is thrown when error happens while creating a VM

Screenshot 2020-01-17 at 13 14 54

After the fix, below is the message which is displayed

Screenshot 2020-01-17 at 13 18 29

## How Has This Been Tested?

Set the global setting ```custom.diskoffering.size.max``` to lower value and try to create a vm with disk size greater the value of setting ```custom.diskoffering.size.max```


weizhouapache:
I think we do not need trillian test for this, right ? @DaanHoogland @rhtyd

DaanHoogland:
I'm fine with it like this but just an earmark: any scrapping of logs or exception messages might break. not a big chance as it in not the essence of the message and it is a fix. I'm fine with merging it like this.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

DaanHoogland:
@rhtyd do you need to see smoke tests for this (log message improvement only)?

rhtyd:
@DaanHoogland I'm okay to merge this once Travis passes.

PR #3822, opened 2020-01-19
set TCP as default protocol in lb list

andrijapanicsb:
## Description
Instead of "SSL" being set as a default LB protocol when configuring a LB rule, set TCP as the default one - move SSL to the bottom of the list.

Reason: SSL is supported only with NetScaller and should not even be there if VPC is not using NetScaller for LB services... so stop offering it as the default protocol and confusing VPC VR (i.e. non-NesScaller) users

## Types of changes

Old setup:
image

New setup:
image

### DIFF
--- a/ui/scripts/network.js
+++ b/ui/scripts/network.js
@@ -3441,21 +3441,21 @@
                                                 isEditable: true,
                                                 select: function(args) {
                                                     var data = [{
-                                                            id: 'ssl',
-                                                            name: 'ssl',
-                                                            description: _l('label.lb.protocol.ssl')
-                                                        }, {
                                                             id: 'tcp',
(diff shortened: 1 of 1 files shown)


andrijapanicsb:
Can be tested by downloading the patch file to mgmt server https://github.com/apache/cloudstack/commit/b9f39757506e0831ebcf015c408f6e93f7198c0d.patch
and applying a patch:

cd /usr/share/cloudstack-management/webapp/scripts
patch < b9f39757506e0831ebcf015c408f6e93f7198c0d.patch
refresh the gui with ctrl +alt + R

DaanHoogland:
@andrijapanicsb not for 4.13? (lgtm otherwise)

Issue #3831, opened 2020-01-23
The metrics view API response is not super-set of resources response keys


rhtyd:
The API responses for metrics views APIs, such as listZonesMetrics don't have all the same keys in the API response as the listZones. If this gets fixed in 4.14, Primate can leverage one API to do both metrics and resource view listings. Would be great if this is fixed in 4.13.1.0/4.14.0.0 milestone.

Illustration, output of listZonesMetrics:
```
    {
      "allocationstate": "Disabled",
      "clusters": "1 / 1",
      "cputotal": "0.00 Ghz",
      "dhcpprovider": "VirtualRouter",
      "id": "b00e2dca-9032-4f1d-aada-8b2cdc19e745",
      "localstorageenabled": true,
      "memorytotal": "0.00 GB",
      "name": "DummyInvalidZone",
      "securitygroupsenabled": true,
      "state": "Disabled",
      "tags": [],
      "zonetoken": "1dbe40a9-46e0-33a8-9b93-44066c94fdfa"
    }
```

Output of listZones:
```
    {
      "allocationstate": "Disabled",
      "dhcpprovider": "VirtualRouter",
      "dns1": "8.8.8.8",
      "id": "b00e2dca-9032-4f1d-aada-8b2cdc19e745",
      "internaldns1": "8.8.8.8",
      "localstorageenabled": true,
      "name": "DummyInvalidZone",
      "networktype": "Basic",
      "securitygroupsenabled": true,
      "tags": [],
      "zonetoken": "1dbe40a9-46e0-33a8-9b93-44066c94fdfa"
    }
```

The `networktype` is missing. There are also disprepancies seen in the listHosts, listHostsMetrics API responses, but did not check other APIs related to volume, vm, cluster and storage pool. It's possible the way a java object is parsed and serialised needs checking/fixing.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API
~~~

##### CLOUDSTACK VERSION

~~~
4.13, 4.14/master
~~~

rhtyd:
This must be a quick issue/fix somewhere, have asked @Pearl1594 to have a look. cc @andrijapanicsb @DaanHoogland

PR #3834, opened 2020-01-24
Fix: The metrics view API response is not super-set of resources response keys

Pearl1594:
## Description
The metrics API has few properties missing that are present in the corresponding resource. This addressed the following issue: https://github.com/apache/cloudstack/issues/3831

## Types of changes

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/response/VolumeResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/VolumeResponse.java
@@ -293,7 +293,7 @@ public class VolumeResponse extends BaseResponseWithTagInformation implements Co
         return this.getId();
     }
 
-    public Boolean isDestroyed() {
+    public Boolean getDestroyed() {
         return destroyed;
     }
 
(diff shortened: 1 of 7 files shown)


api/src/main/java/org/apache/cloudstack/api/response/VolumeResponse.java
```diff
@@ -293,7 +293,7 @@ public String getObjectId() {
         return this.getId();
     }
 
-    public Boolean isDestroyed() {
+    public Boolean getDestroyed() {
```
DaanHoogland: boolean getter should adhere to is<attr>() pattern

DaanHoogland:
I do not care much either way but we agreed on the getter convention for booleans a while back. thoughts @rhtyd ?

weizhouapache:
@DaanHoogland my suggestion is, we can use lombok.
https://projectlombok.org/

DaanHoogland:
@weizhouapache looks cool. @Pearl1594 @rhtyd is there a specific reason to go back on our getter convention, or any other reason we shouldn't use lombok? (name of the neightbourhood I live in;) after the indonesian isleland \o/)

DaanHoogland:
@Pearl1594 @rhtyd can you meddle in on @weizhouapache 's idea? (lombok would actually make this a get<>() as it is Boolean and not boolean)

Pearl1594:
@DaanHoogland I've reverted the particular function to it's original naming convention. the getter function need not be modified for this case.

rhtyd:
That's how BeanUtils::copyProperties and most IDEs works. The problem of booleans having `is` than `get` breaks the use case to copy bean/properties.

rhtyd:
@DaanHoogland this is a special case for the metrics apis that extend over resource list apis. Without using getters instead of issers, it cannot be fixed. Turns out the previously referred agreement is merely our own preference but not a standard convention for IDEs, and libraries.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

PR #3465, opened 2019-07-04
vr: Fix vpc router in UNKNOWN state

ustcweizhou:
## Description

If there are more than 10 vpc tiers or public ip subnets in a VPC, eth1X will be added in vpc router.
The redundant state is UNKNOWN in this case.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

creaete more than 10 vpc tiers, then vpc router will be UNKNOWN
it can be fixed by this commit.

PR #3848, opened 2020-01-28
vr: fix vr in unknown state (more)

ustcweizhou:
## Description

This fixes similar issue with #3465 

Meanwhile change log level of CsHelper.execute2 from DEBUG to INFO and fix some typo

## Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsHelper.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsHelper.py
@@ -29,8 +29,8 @@ from netaddr import *
 
 PUBLIC_INTERFACES = {"router": "eth2", "vpcrouter": "eth1"}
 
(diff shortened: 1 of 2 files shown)


DaanHoogland:
code lgtm but it is only an improvement, is that right @weizhouapache or does it really solve a bug?

weizhouapache:
@DaanHoogland I believe there are some issues caused by this when vpc has more than 10 tiers.
This change in STATE_COMMANDS is similar as #3465 which should be a bug fix.

The change on log level can be regarded as an improvement.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

DaanHoogland:
> @DaanHoogland I believe there are some issues caused by this when vpc has more than 10 tiers.

grep eth10 or -eth11 etc, got it thanks @weizhouapache

rhtyd:
@blueorangutan test

Issue #2783, opened 2018-08-02
Migrating router with public interface fails


marcowi:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR,QEMU
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1
~~~

##### CONFIGURATION

Both hosts have identical configuration:
cloudbr0   - private
cloudbr1   - storage
cloudbr2   - guest (VLAN 2101)
cloudbr3   - public (VLAN 99)

##### SUMMARY

Migrating a virtual router from one host to another results in error "org.libvirt.LibvirtException: Cannot get interface MTU on 'brbond2-99': no such device".
Digging into it shows that the qemu hook (https://github.com/apache/cloudstack/blob/master/agent/bindir/libvirtqemuhook.in) iterating over the interfaces; including the public one which has the name brbond3-99. In this iteration it gets the guest interface name from the agent.properties and concatenates it with the vlan-id 99 so it results in brbond2-99.

##### STEPS TO REPRODUCE

Migrate a router with a public interface which has a vlan.

##### EXPECTED RESULTS

Since the public interface is needed on the other host as well one should handle the public interface the same but with the public.network.device value from agent.properties.

##### ACTUAL RESULTS

Public interfaces are treated as guest interfaces.

rhtyd:
Do you have bonded nics on the destination host? @marcowi what are the names of your private, public and guest network devices of the kvm hosts (as configured in agent.properties)?

marcowi:
Yes, they are bonded on every host and the hosts are identical in hard- and software.

agent.properties:
private.network.device=cloudbr0
public.network.device=cloudbr3
guest.network.device=cloudbr2

rhtyd:
@PaulAngus @DagSonsteboSB any comments on this?

PaulAngus:
how are you assigning these vlans (99 and 2101) ?  as you have a public, this suggests advanced networking. the way have written you configuration suggest to me that you have hardcoded these VLANs on the hosts.   for advanced networking, this should not be the case.  They should be connected to trunked switch ports, and CloudStack allowed to apply the tags.  This sounds like CloudStack is now trying to add VLAN tags to already VLAN tagged interfaces.

marcowi:
It is advanced networking and the 99 is configured in Home>Infrastructure>Zones>myzone>eno5>Public as vlan://99 for the range of public IPs.
[...]

PR #3855, opened 2020-01-30
kvm: Fix router migration issue when router has control/public nics onother physical network than guest

ustcweizhou:
## Description

In VM migration on KVM, libvirt qemu hook script will change the bridge name to bridges for guest networks. It works for user vm. However for virtual router, it has nics on control network and public network. If control/public use different physical networks than guest network, virtual router cannot be migrated.

Fixes: #2783 

## Types of changes

## How Has This Been Tested?

We have more than 1 physical networks and faced same issue before.
We have this change to fix the issue in old cloudstack versions. 
I did not test in latest master or 4.13 but I believe it should work as well.


DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

Issue #3859, opened 2020-02-03
test_privategw_acl.py failed in some Trillian tests


weizhouapache:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME

~~~
VPC private gateway
~~~

##### CLOUDSTACK VERSION

~~~
4.14
~~~

##### CONFIGURATION

Advanced zone 

##### SUMMARY

~~~
Smoke test  test_privategw_acl.py failed in some Trillian tests.
~~~

##### STEPS TO REPRODUCE

~~~
see more details in https://github.com/apache/cloudstack/blob/master/test/integration/smoke/test_privategw_acl.py#L317-L381
~~~

##### EXPECTED RESULTS

~~~
No errors/failures
~~~

##### ACTUAL RESULTS

~~~
Some tests failed. For example

https://github.com/apache/cloudstack/pull/3477#issuecomment-571288662
https://github.com/apache/cloudstack/pull/3836#issuecomment-579981259
https://github.com/apache/cloudstack/pull/3851#issuecomment-580891071

~~~

weizhouapache:
https://github.com/apache/cloudstack/pull/3871#issuecomment-583599818

weizhouapache:
https://github.com/apache/cloudstack/pull/3601#issuecomment-584493660

rhtyd:
Failure seen in latest master smoketests - https://github.com/apache/cloudstack/pull/3850#issuecomment-585511968

@weizhouapache I see this issue, are you working on it?

weizhouapache:
@rhtyd yes
found the root cause: https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/network/NetworkServiceImpl.java#L4472

private gateway in two VPCs use the same isolated network, if they have same vlan and cidr.

rhtyd:
Okay @weizhouapache do we need to fix that in the test data config, test or in CloudStack java code?

rhtyd:
@weizhouapache are you send a fix? Thanks.

weizhouapache:
@rhtyd we need to fix it in java code. it is not simple.

rhtyd:
I think it is expected that vlan shouldn't be same? @weizhouapache  - should we fix this in the test or test's config instead?

weizhouapache:
> I think it is expected that vlan shouldn't be same? 
[...]

PR #3861, opened 2020-02-04
test: check more connectivity in test_privategw_acl.py

ustcweizhou:
## Description

in test_privategw_acl.py, we need to check connectivity not only vpc1 to vpc2, but also vpc2 to vpc1.

Related to: #3859 

## Types of changes


ustcweizhou:
@rhtyd @DaanHoogland can you kick a test ?
The test should succeed, but I suspect this will fail because I found a bug in the investigation on #3859.

rhtyd:
Sure.
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3862, opened 2020-02-04
Userdata to display static NAT as public ip instead of VR ip

ravening:
## Description

If static nat is enabled on VM then metadata service should return
the static nat instead of gateway IP.
If static not is not enabled then it should return the gateway IP
as the public IP

## Types of changes

## How Has This Been Tested?

Step to reproduce:

1. Create a vm
2. Ssh to vm.
3. Run the below command inside the vm
   wget http://\<VR public ip\>/latest/meta-data/public-ipv4

   Note down the output of the above command
4. Now acquire a new public and enable static NAT on that IP to this vm
5. Now run the same command mentioned above in the VM
   This should display the static NAT ip instead of VR public IP

Output:

Before enabling static nat

wget http://10.10.10.40/latest/meta-data/public-ipv4
$ cat public-ipv4
10.10.10.29

After enabling static nat

wget http://10.10.10.40/latest/meta-data/public-ipv4
$ cat public-ipv4
10.11.10.30

### DIFF
--- a/server/src/main/java/com/cloud/network/router/CommandSetupHelper.java
+++ b/server/src/main/java/com/cloud/network/router/CommandSetupHelper.java
@@ -188,9 +188,11 @@ public class CommandSetupHelper {
     public void createVmDataCommand(final VirtualRouter router, final UserVm vm, final NicVO nic, final String publicKey, final Commands cmds) {
         final String serviceOffering = _serviceOfferingDao.findByIdIncludingRemoved(vm.getId(), vm.getServiceOfferingId()).getDisplayText();
         final String zoneName = _dcDao.findById(router.getDataCenterId()).getName();
+        final IPAddressVO staticNatIp = _ipAddressDao.findByVmIdAndNetworkId(nic.getNetworkId(), vm.getId());
         cmds.addCommand(
                 "vmdata",
-                generateVmDataCommand(router, nic.getIPv4Address(), vm.getUserData(), serviceOffering, zoneName, nic.getIPv4Address(), vm.getHostName(), vm.getInstanceName(),
+                generateVmDataCommand(router, nic.getIPv4Address(), vm.getUserData(), serviceOffering, zoneName,
+                        staticNatIp == null ? null : staticNatIp.getAddress().addr(), vm.getHostName(), vm.getInstanceName(),
                         vm.getId(), vm.getUuid(), publicKey, nic.getNetworkId()));
     }
 
@@ -1010,7 +1012,7 @@ public class CommandSetupHelper {
     }
 
     private VmDataCommand generateVmDataCommand(final VirtualRouter router, final String vmPrivateIpAddress, final String userData, final String serviceOffering,
-            final String zoneName, final String guestIpAddress, final String vmName, final String vmInstanceName, final long vmId, final String vmUuid, final String publicKey,
+            final String zoneName, final String publicIpAddress, final String vmName, final String vmInstanceName, final long vmId, final String vmUuid, final String publicKey,
             final long guestNetworkId) {
         final VmDataCommand cmd = new VmDataCommand(vmPrivateIpAddress, vmName, _networkModel.getExecuteInSeqNtwkElmtCmd());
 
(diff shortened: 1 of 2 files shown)


rhtyd:
@blueorangutan package

rhtyd:
if @ravening does decide to change to 4.13, we can change the milestone to 4.13.1 @DaanHoogland  - it's not exactly a bug fix is it?

DaanHoogland:
It is marked as a bug and could be considered so by a user that has a static address assigned. The want to be visible by that address in the outside world, @rhtyd .

rhtyd:
@ravening cc @weizhouapache - can you explain how the user-data would be used by the user VM? Could that cause regressions for VMs which were previously expecting the VR/gateway public IP instead?

andrijapanicsb:
When static nat is configured for a VM - then it actually shows it's IP address clearly, so this is in line with the fix I believe:

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
@blueorangutan test

rhtyd:
@blueorangutan package

ravening:
> @ravening cc @weizhouapache - can you explain how the user-data would be used by the user VM? Could that cause regressions for VMs which were previously expecting the VR/gateway public IP instead?

If VM's dont have static NAT enabled, it will still display VR gateway as public ip. Only when static NAT is enabled, it will display static NAT as public ip. I guess that should be the expected behavior and shouldnt break regression

ravening:
> @ravening this looks great, thanks. Can we rebase this on 4.13, please?

@DaanHoogland rebased to 4.13

rhtyd:
@blueorangutan package

andrijapanicsb:
will test once I'm available for the sake of "somebody else tested as well" thx.

rhtyd:
@blueorangutan test

rhtyd:
Can anyone confirm manual testing? Are we breaking any compatibility?

DaanHoogland:
@wido @GabrielBrascher Are using this kind of configuration? @pdion891 @swill ?

@rhtyd I don't see any related error in that report from the blue ape. I am fine with merging.

/cc @ravening @weizhouapache

andrijapanicsb:
testing it now

andrijapanicsb:
Tested fine, except for a single issue I noticed (might not be related, but it also might)
Happens in both VPC and Isolated network
- Configure Static NAT, check meta-data - correct Static NAT IP reported
- Release the IP, without previously disabling the Static NAT - and still the same Public IP is reported in meta-data, which is wrong (should be the source NAT now)
- FYI: Disabling the Static NAT works fine, again Source NAT is reported in meta-data

The issue has something to do with IPtables and the way we (don't) clean those rules properly, see the image below.

@weizhouapache if you think this is not related to the PR, then we can move forward with this one (and this is kind of edge case, you are usually supposed to disable Static NAT, but again - a valid issue anyway)

image

weizhouapache:
> Tested fine, except for a single issue I noticed (might not be related, but it also might)
> Happens in both VPC and Isolated network
> 
> * Configure Static NAT, check meta-data - correct Static NAT IP reported
> * Release the IP, without previously disabling the Static NAT - and still the same Public IP is reported in meta-data, which is wrong (should be the source NAT now)
> * FYI: Disabling the Static NAT works fine, again Source NAT is reported in meta-data
> 
> The issue has something to do with IPtables and the way we (don't) clean those rules properly, see the image below.
> 
> @weizhouapache if you think this is not related to the PR, then we can move forward with this one (and this is kind of edge case, you are usually supposed to disable Static NAT, but again - a valid issue anyway)
> 
> image

@andrijapanicsb thanks for testing !
It is definitely an issue. We will fix it.

server/src/main/java/com/cloud/network/router/CommandSetupHelper.java
```diff
@@ -1024,18 +1026,21 @@ private VmDataCommand generateVmDataCommand(final VirtualRouter router, final St
...
+        if (dcVo.getNetworkType() == NetworkType.Basic || network.getGuestType() == Network.GuestType.Shared) {
+            cmd.addVmData("metadata", "public-ipv4", vmPrivateIpAddress);
             cmd.addVmData("metadata", "public-hostname", StringUtils.unicodeEscape(vmName));
         } else {
-            if (router.getPublicIpAddress() == null) {
-                cmd.addVmData("metadata", "public-ipv4", guestIpAddress);
+            if (publicIpAddress != null) {
+                cmd.addVmData("metadata", "public-ipv4", publicIpAddress);
+                cmd.addVmData("metadata", "public-hostname", publicIpAddress);
             } else {
```
weizhouapache: @ravening you missed one check here
```
-            } else {
+            } else if (router.getPublicIpAddress() != null) {
```

ravening: Made necessary chagnes

ravening:
> Tested fine, except for a single issue I noticed (might not be related, but it also might)
> Happens in both VPC and Isolated network
> 
> * Configure Static NAT, check meta-data - correct Static NAT IP reported
> * Release the IP, without previously disabling the Static NAT - and still the same Public IP is reported in meta-data, which is wrong (should be the source NAT now)
> * FYI: Disabling the Static NAT works fine, again Source NAT is reported in meta-data
> 
> The issue has something to do with IPtables and the way we (don't) clean those rules properly, see the image below.
> 
> @weizhouapache if you think this is not related to the PR, then we can move forward with this one (and this is kind of edge case, you are usually supposed to disable Static NAT, but again - a valid issue anyway)
> 
> image

@andrijapanicsb Thanks for testing. I have updated the commit. Can you please test it again and let me know if you face any issues?

weizhouapache:
@ravening the new commits is not related to the issue @andrijapanicsb reported.
can you test following what @andrijapanicsb said ?

```
Release the IP, without previously disabling the Static NAT - and still the same Public IP is reported in meta-data, which is wrong (should be the source NAT now)
```

ravening:
> Release the IP, without previously disabling the Static NAT - and still the same Public IP is reported in meta-data, which is wrong (should be the source NAT now)

@weizhouapache tested the steps which @andrijapanicsb mentioned above.
Yes it does report the public static nat ip instead of source nat ip if I release an IP without disabling the static nat.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland:
@andrijapanicsb can you re-test?

rhtyd:
@andrijapanicsb can you re-test?

andrijapanicsb:
incoming...

andrijapanicsb:
Merging based on 3 x LGTMs/Approvals and extensive manual testing, and passing regression tests.

PR #3867, opened 2020-02-05
Usage event to store zone id while uploading template and volume

Pearl1594:
## Description
After a local template is uploaded via browser, the generated usage event with type = "TEMPLATE.CREATE" is persisted with the data store ID instead of the zone ID on the zone_id column. The fix will refactor the upload monitor logic, as after the upload completes, it sets the datastore ID on the zone ID column for the created "TEMPLATE.CREATE" usage event. This refactor will query the DB for the data store and will set its associated zone ID in the usage field.
The fix produces the same behaviour as when registering a template from URL.
FIx is also for uploading VOLUME from local/via browser.

## Types of changes


andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan test

PR #3870, opened 2020-02-07
systemvm: list systemvm does not return agent state and version

rhtyd:
This makes the listSystemVms API to return the host status (agent state),
version and last pinged information. This makes it possible for UIs
to call a single API to get this information.

This is largely for Primate.

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/response/SystemVmResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/SystemVmResponse.java
@@ -19,14 +19,14 @@ package org.apache.cloudstack.api.response;
 import java.util.Date;
 import java.util.List;
 
-import com.google.gson.annotations.SerializedName;
-
 import org.apache.cloudstack.api.ApiConstants;
 import org.apache.cloudstack.api.BaseResponse;
 import org.apache.cloudstack.api.EntityReference;
 
+import com.cloud.host.Status;
 import com.cloud.serializer.Param;
 import com.cloud.vm.VirtualMachine;
+import com.google.gson.annotations.SerializedName;
 
 @EntityReference(value = VirtualMachine.class)
 public class SystemVmResponse extends BaseResponse {
@@ -138,6 +138,10 @@ public class SystemVmResponse extends BaseResponse {
     @Param(description = "the state of the system VM")
     private String state;
 
+    @SerializedName("agentstate")
+    @Param(description = "the agent state of the system VM")
+    private String agentState;
+
     @SerializedName("activeviewersessions")
     @Param(description = "the number of active console sessions for the console proxy system vm")
     private Integer activeViewerSessions;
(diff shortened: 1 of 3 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@shwstppr @Pearl1594 please review as well.

Pearl1594:
LGTM - agentstate, lastpinged and version fields seem to be added to the list systemvms response

api/src/main/java/org/apache/cloudstack/api/response/SystemVmResponse.java
```diff
@@ -138,6 +138,10 @@
     @Param(description = "the state of the system VM")
     private String state;
 
+    @SerializedName("agentstate")
+    @Param(description = "the agent state of the system VM")
```
shwstppr: can we add `since` key here as these are new response params

rhtyd:
@DaanHoogland the technical preview targets 4.14, without this fix the implementation in Primate would be complicated.

weizhouapache:
code LGTM

@rhtyd can you make some changes on 4.13/4.14 UI ? the agent state is determined  in a complicated way.

rhtyd:
That's not very critical, but let me check if it's a quick one. I originally wanted to limit the scope just to apis, wrt Primate.

DaanHoogland:
ok, @rhtyd I'm fine with it. merge at will.

rhtyd:
@shwstppr thanks, fixed the `since` in params response keys.
@weizhouapache I've fixed the UI as well. Can you review and test the API/UI changes? Thanks.

rhtyd:
Thanks @DaanHoogland will merge after smoketests and some UI confirmation (cc @weizhouapache)
@blueorangutan package

DaanHoogland:
@weizhouapache can you confirm the ui?

weizhouapache:
@DaanHoogland ui works.

However, lastping returns wrong date/time. we have same issue on other api response. we can fix them together afterwards.

DaanHoogland:
sounds like a ugly bug @weizhouapache shift by ten bits != /1000. I think we should move away from the bit shifting when dealing with milliseconds.
Would you consider this not to be a blocker for this PR?

DaanHoogland:
@blueorangutan test

rhtyd:
@weizhouapache cc @DaanHoogland I used the same logic as here: https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/api/query/dao/HostJoinDaoImpl.java#L105

weizhouapache:
@DaanHoogland what I said is not a blocker. I am ok with this PR.

@rhtyd yes, it has same issue. In response of listhosts and listsystemvms, both shows the last ping time is "1970-01-XXXX" instead of "2020-02-XXXX". we can fix it in the future, not today.

Issue #3860, opened 2020-02-03
UI: Resize ROOT volume does not invoke resizeVolume API


nvazquez:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
UI
~~~

##### CLOUDSTACK VERSION
~~~
4.13.0
~~~

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### SUMMARY
The API invokation to resizeVolume API does not happen as the UI breaks due to an error (seen with developer tools enabled):

![image](https://user-images.githubusercontent.com/5295080/73660153-656dc780-4676-11ea-9c73-e1e1366a7175.png)

##### STEPS TO REPRODUCE
~~~
Create a VM
Select ROOT disk
Resize disk to a new size
~~~

##### EXPECTED RESULTS
~~~
Resize API is invoked
~~~

##### ACTUAL RESULTS
~~~
API is not invoked
~~~

rhtyd:
@Spaceman1984 can you have a look at this as well?

Spaceman1984:
Sure @rhtyd

PR #3873, opened 2020-02-07
Fixed root volume resize from ui

Spaceman1984:
## Description

This PR fixes the volume resize action for a root volume from the UI.

Fixes: #3860 #3874
## Types of changes

volume resize
## How Has This Been Tested?

This has been tested by creating an instance and resizing the root volume. The root volume size is then verified in storage metrics.

### DIFF
--- a/ui/scripts/storage.js
+++ b/ui/scripts/storage.js
@@ -1575,7 +1575,7 @@
                                     if (newDiskOffering != null && newDiskOffering.length > 0) {
                                         array1.push("&diskofferingid=" + encodeURIComponent(newDiskOffering));
                                     }
-                                    if ((args.context.volumes[0].type == "ROOT" || selectedDiskOfferingObj.iscustomized == true) {
+                                    if (args.context.volumes[0].type == "ROOT" || selectedDiskOfferingObj.iscustomized == true) {
                                         cloudStack.addNewSizeToCommandUrlParameterArrayIfItIsNotNullAndHigherThanZero(array1, args.data.newsize);
                                     }
 
--- a/ui/scripts/storage.js
+++ b/ui/scripts/storage.js
@@ -1575,7 +1575,7 @@
                                     if (newDiskOffering != null && newDiskOffering.length > 0) {
                                         array1.push("&diskofferingid=" + encodeURIComponent(newDiskOffering));
                                     }
-                                    if (selectedDiskOfferingObj != null && selectedDiskOfferingObj.iscustomized == true) {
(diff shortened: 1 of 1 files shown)


Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

Spaceman1984:
@blueorangutan package

weizhouapache:
@Spaceman1984 
I have tested it ,it works.

however, I think we can use the following change on line 1578
```
if (selectedDiskOfferingObj == null || selectedDiskOfferingObj.iscustomized == true) {
```
the change on line 1597-1601 are not needed then.

the change on line 1584 are needed.

Spaceman1984:
> @Spaceman1984
> I have tested it ,it works.
> 
> however, I think we can use the following change on line 1578
> 
> ```
> if (selectedDiskOfferingObj == null || selectedDiskOfferingObj.iscustomized == true) {
> ```
> 
> the change on line 1597-1601 are not needed then.
> 
> the change on line 1584 are needed.
@weizhouapache 
I agree with all, accept, I think line 1601 can be removed, but we need to keep line 1597-1600 - it's an error message about not allowing a smaller volume size.

Spaceman1984:
All good @weizhouapache, I will make those changes.

Spaceman1984:
@blueorangutan package

borisstoyanov:
@blueorangutan package

ui/scripts/storage.js
```diff
@@ -1575,13 +1575,13 @@
                                     if (newDiskOffering != null && newDiskOffering.length > 0) {
                                         array1.push("&diskofferingid=" + encodeURIComponent(newDiskOffering));
                                     }
-                                    if (selectedDiskOfferingObj.iscustomized == true) {
+                                    if (selectedDiskOfferingObj == null || selectedDiskOfferingObj.iscustomized == true) {
```
DaanHoogland: aren't we making an assumption here about the reason the offering is missing? We only want to allow it if it is missing because it is a root-disk. It looks like we are allowing some client side hacking; set the `selectedDiskOfferingObj` to `null` and you are good to go. (i didn't check the API)

Spaceman1984: I have changed the logic to check the volume type instead.

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan package

rhtyd:
@blueorangutan package

borisstoyanov:
I guess we could merge this one as it does not require marvin tests (UI fix).

PR #3879, opened 2020-02-11
kvm: Enable virtio drivers based on guest os display name

ustcweizhou:
## Description

When we add new guest os, sometimes we missed the records in guest_os_hypervisor. 
However, the guest disk model (virtio/ide) is determined by record in the table.
It causes the issue that some new guest os(eg Debian 8/9) uses e1000 instead of virtio nic, and ide disk instead of virtio disk.

To fix the issue permanantly, pass the guest os name in guest_os if the record for kvm is not found in guest_os_hypervisor.

Related commit:7ac9f00eeeb4cd37ec39efeba066e799b581b1a0

## Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/hypervisor/KVMGuru.java
+++ b/server/src/main/java/com/cloud/hypervisor/KVMGuru.java
@@ -132,7 +132,7 @@ public class KVMGuru extends HypervisorGuruBase implements HypervisorGuru {
             guestOsMapping = _guestOsHypervisorDao.findByOsIdAndHypervisor(guestOS.getId(), getHypervisorType().toString(), host.getHypervisorVersion());
         }
         if (guestOsMapping == null || host == null) {
-            to.setPlatformEmulator("Other");
+            to.setPlatformEmulator(guestOS.getDisplayName());
         } else {
             to.setPlatformEmulator(guestOsMapping.getGuestOsName());
         }
--- a/server/src/main/java/com/cloud/hypervisor/KVMGuru.java
+++ b/server/src/main/java/com/cloud/hypervisor/KVMGuru.java
@@ -132,7 +132,7 @@ public class KVMGuru extends HypervisorGuruBase implements HypervisorGuru {
             guestOsMapping = _guestOsHypervisorDao.findByOsIdAndHypervisor(guestOS.getId(), getHypervisorType().toString(), host.getHypervisorVersion());
         }
         if (guestOsMapping == null || host == null) {
-            to.setPlatformEmulator(guestOS.getDisplayName());
+            to.setPlatformEmulator(guestOS.getDisplayName() == null ? "unknown" : guestOS.getDisplayName());
         } else {
             to.setPlatformEmulator(guestOsMapping.getGuestOsName());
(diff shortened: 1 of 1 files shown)


svenvogel:
@weizhouapache can you explain it me a little bit better. if i understand it correctly if guestosmapping and host is null then we set guestOS.getDisplayName(). why display name? maybe i stand on the line 😕

weizhouapache:
> @weizhouapache can you explain it me a little bit better. if i understand it correctly if guestosmapping and host is null then we set guestOS.getDisplayName(). why display name? maybe i stand on the line 😕

@svenvogel 
here is an example:
in guest_os, we have Debian 9
```
mysql> select id,name,display_name from guest_os where id in (289,290);
+-----+------+-----------------------------+
| id  | name | display_name                |
+-----+------+-----------------------------+
| 289 | NULL | Debian GNU/Linux 9 (32-bit) |
| 290 | NULL | Debian GNU/Linux 9 (64-bit) |
+-----+------+-----------------------------+
2 rows in set (0.01 sec)
```
however, in guest_os_hypervisor, there is no record for these guest oses on kvm
```
mysql> select * from guest_os_hypervisor where guest_os_id in (289,290) and hypervisor_type='KVM';
Empty set (0.00 sec)
```

when we start a vm with Debian 9, in StartCommand it shows
```
"arch":"x86_64","os":"Debian GNU/Linux 9 (64-bit)","platformEmulator":"Other"
```

so vm is started with ide disk and e1000 nic.


with this change, "platformEmulator" will be set to same value as "os"

rhtyd:
Some testing required - @DaanHoogland @andrijapanicsb

andrijapanicsb:
@blueorangutan package

rhtyd:
@blueorangutan test

server/src/main/java/com/cloud/hypervisor/KVMGuru.java
```diff
@@ -132,7 +132,7 @@ public VirtualMachineTO implement(VirtualMachineProfile vm) {
             guestOsMapping = _guestOsHypervisorDao.findByOsIdAndHypervisor(guestOS.getId(), getHypervisorType().toString(), host.getHypervisorVersion());
         }
         if (guestOsMapping == null || host == null) {
-            to.setPlatformEmulator("Other");
+            to.setPlatformEmulator(guestOS.getDisplayName());
```
rhtyd: @ustcweizhou how about we check and fallback to `"Other"` in case `guestOS.getDisplayName()` is null or empty (unlikely but just to make code more defensive)?

rhtyd:
@blueorangutan test

DaanHoogland:
sorry @weizhouapache, i editted according to @rhtyd 's comment (kinda, revert if you want)

weizhouapache:
@DaanHoogland thanks Daan !
I changed unknown to Other

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan test

rhtyd:
Tests LGTM, the failures are known issue

PR #3888, opened 2020-02-16
jdk versions

DaanHoogland:
## Description

try if the toolchains plugin works ;)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -52,6 +52,7 @@
 
         <!-- Build properties -->
         <cs.jdk.version>14</cs.jdk.version>
+        <cs.jdk.vendor>openjdk</cs.jdk.vendor>
         <cs.target.dir>target</cs.target.dir>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -51,7 +51,8 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
         <!-- Build properties -->
-        <cs.jdk.version>11</cs.jdk.version>
+        <cs.jdk.version>14</cs.jdk.version>
```
rhtyd: This would cause PR to be built against jdk14, given what got broken and fixed to make it work for jdk11 - moving to a newer jdk may be more work.

rhtyd:
Let's close this and focus on only making jdk11 work for now @DaanHoogland ?

DaanHoogland:
closing, note @rhtyd that the focus of this is not making any one jdk work but have a jenkins/maven equivalent of jenv.

Issue #3891, opened 2020-02-17
Upgrade script - vpc_offering_details add 'if not exist' 


borisstoyanov:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
DB Upgrade
~~~

##### CLOUDSTACK VERSION

~~~

~~~

##### SUMMARY

The upgrade fails with the following exception
```
2020-02-16 16:46:28,907 DEBUG [c.c.u.d.ScriptRunner] (main:null) (logid:) CREATE TABLE `vpc_offering_details` (   `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,   `offering_id` bigint(20) unsigned NOT NULL COMMENT 'vpc offering id',   `name` varchar(255) NOT NULL,   `value` varchar(1024) NOT NULL,   `display` tinyint(1) NOT NULL DEFAULT '1' COMMENT 'True if the detail can be displayed to the end user',   PRIMARY KEY (`id`),   KEY `fk_vpc_offering_details__vpc_offering_id` (`offering_id`),   CONSTRAINT `fk_vpc_offering_details__vpc_offering_id` FOREIGN KEY (`offering_id`) REFERENCES `vpc_offerings` (`id`) ON DELETE CASCADE ) ENGINE=InnoDB DEFAULT CHARSET=utf8
2020-02-16 16:46:28,918 ERROR [c.c.u.d.ScriptRunner] (main:null) (logid:) Error executing: CREATE TABLE `vpc_offering_details` (   `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,   `offering_id` bigint(20) unsigned NOT NULL COMMENT 'vpc offering id',   `name` varchar(255) NOT NULL,   `value` varchar(1024) NOT NULL,   `display` tinyint(1) NOT NULL DEFAULT '1' COMMENT 'True if the detail can be displayed to the end user',   PRIMARY KEY (`id`),   KEY `fk_vpc_offering_details__vpc_offering_id` (`offering_id`),   CONSTRAINT `fk_vpc_offering_details__vpc_offering_id` FOREIGN KEY (`offering_id`) REFERENCES `vpc_offerings` (`id`) ON DELETE CASCADE ) ENGINE=InnoDB DEFAULT CHARSET=utf8
[...]

PR #3892, opened 2020-02-17
Add if not exists constraint to create table on upgrade paths

nvazquez:
## Description
Simply add 'IF NOT EXISTS' constraint to 'CREATE TABLE' SQLs on upgrade paths. This avoids upgrade paths to fail if there had been previous failures on the upgrade process.

Fixes: #3891 

## Types of changes


borisstoyanov:
@blueoranguran package

borisstoyanov:
@blueoranguran package

rhtyd:
@blueorangutan test

andrijapanicsb:
wow.... what, why?

-1 

If there was previous failure on upgrade, bunch of other things will fail, like extending the existing table or such. I saw the issue which this PR is supposed to fix, and I vote for closing both.

The correct action is failed upgrade --> roll back DB --> start again.

andrijapanicsb:
To further explain myself - if you attempt the upgrade on top of DB which previously failed to be upgraded, you might end up with duplicate entries trying to be inserted and that would be chaos.
The only "normal" way (acceptable in production) to proceed with the failed upgrade, is to start clean (drop DB, restore DB) and try again.

DaanHoogland:
@andrijapanicsb I almost agree with you. There is a possibility that two upgrade paths produce conflicting table definitions. In those cases (and in principle) the "CREATE IF NOT EXISTS" pattern is very good to hav as a standard. Double entries are a whole different issue.

PR #3903, opened 2020-02-20
VR: Send VM password to all Running VRs in network/vpc

ravening:
## Description

Currently, the cloudstack sends VM password only to the first
router in the network even if its the backup and return the result.

In some cases the first router will be back up and the second will be master.
Since password server is not running in backup, when the user resets the password,
it is sent to the first router which can be backup.
In that case, the new password is not stored in the password server and users cant log in with a new password.

This change ensures that we send the password to both the routers instead
of the first router so that a new password is stored in the master router.

## Types of changes

## How Has This Been Tested?

1 . Create and isolated network with redundant VR and a VM in it.
2 . Note down the password of the newly created and try to ssh to it from VR
3 . Ssh works with the existing password.
4 . Now reboot the master VR so that it will become BACKUP and the BACKUP will become MASTER
5 . Now stop the VM and reset it's password.
6 . Try to login to VM using the new password

Expected Result:

Ssh should work with new password

Actual result:

Ssh wont work

### DIFF
--- a/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
+++ b/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
@@ -707,6 +707,11 @@ NetworkMigrationResponder, AggregatedCommandExecutor, RedundantResource, DnsServ
         for (final VirtualRouter router : routers) {
             if (router.getState() == State.Running) {
                 final boolean result = networkTopology.savePasswordToRouter(network, nic, uservm, router);
+                if (!result) {
+                    s_logger.error("Unable to save password for VM " + vm.getInstanceName() +
+                            " on router " + router.getInstanceName());
+                    return false;
+                }
                 isVrRunning = true;
                 savePasswordResult = savePasswordResult && result;
             }
--- a/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
+++ b/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
@@ -702,18 +702,27 @@ NetworkMigrationResponder, AggregatedCommandExecutor, RedundantResource, DnsServ
 
         // If any router is running then send save password command otherwise
         // save the password in DB
+        boolean savePasswordResult = true;
+        boolean isVrRunning = false;
         for (final VirtualRouter router : routers) {
             if (router.getState() == State.Running) {
                 final boolean result = networkTopology.savePasswordToRouter(network, nic, uservm, router);
-                if (result) {
-                    // Explicit password reset, while VM hasn't generated a password yet.
-                    final UserVmVO userVmVO = _userVmDao.findById(vm.getId());
-                    userVmVO.setUpdateParameters(false);
-                    _userVmDao.update(userVmVO.getId(), userVmVO);
-                }
-                return result;
+                isVrRunning = true;
+                savePasswordResult = savePasswordResult && result;
             }
         }
+
+        // return the result only if one of the vr is running
+        if (isVrRunning) {
+            if (savePasswordResult) {
+                // Explicit password reset, while VM hasn't generated a password yet.
+                final UserVmVO userVmVO = _userVmDao.findById(vm.getId());
+                userVmVO.setUpdateParameters(false);
+                _userVmDao.update(userVmVO.getId(), userVmVO);
+            }
+            return savePasswordResult;
+        }
+
         final String password = (String) uservm.getParameter(VirtualMachineProfile.Param.VmPassword);
         final String password_encrypted = DBEncryptionUtil.encrypt(password);
         final UserVmVO userVmVO = _userVmDao.findById(vm.getId());


weizhouapache:
@ravening it looks the title is not correct. vm password is sent to all Running VRs , not MASTER VR.

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
```diff
@@ -702,18 +702,27 @@ public boolean savePassword(final Network network, final NicProfile nic, final V
 
         // If any router is running then send save password command otherwise
         // save the password in DB
+        boolean savePasswordResult = true;
+        boolean isVrRunning = false;
```
GabrielBrascher: @ravening can you please make both variables private?

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
```diff
@@ -702,18 +702,27 @@ public boolean savePassword(final Network network, final NicProfile nic, final V
...
                 final boolean result = networkTopology.savePasswordToRouter(network, nic, uservm, router);
-                if (result) {
-                    // Explicit password reset, while VM hasn't generated a password yet.
-                    final UserVmVO userVmVO = _userVmDao.findById(vm.getId());
-                    userVmVO.setUpdateParameters(false);
-                    _userVmDao.update(userVmVO.getId(), userVmVO);
-                }
-                return result;
+                isVrRunning = true;
+                savePasswordResult = savePasswordResult & result;
```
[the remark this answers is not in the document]
weizhouapache: @DaanHoogland 👍 
maybe it is a typo, we should use && instead of &.

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
weizhouapache: @GabrielBrascher isnot variables private by default ?

DaanHoogland: yes @weizhouapache @GabrielBrascher , and i don't even think the keyword is valid in that place. so unless you are asking to move them to fields...?

ravening: @GabrielBrascher @DaanHoogland do you still want me to add private keyword?

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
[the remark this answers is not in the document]
ravening: will change it

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
GabrielBrascher: You are right, I did not expand correctly the diff and saw it as on the class instead of method.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

GabrielBrascher:
Additionally to what I mentioned on the previous comment:

`BasicNetworkTopology.applyRules(Network, VirtualRouter, String, boolean, Long, boolean, RuleApplierWrapper<RuleApplier>)` is the method that will return true or false on `savePasswordToRouter`. Most of the issues that can happen there will either raise an exception or log it so I don't see much problem; however, I got interested in the fact that it might make sense to break the loop (and maybe add a log to that explaining).

Again, overall it looks OK :)

weizhouapache:
> Is it OK to have one of the routers returning false? Would it make sense do break de loop and log an error/warn/debug message?

@GabrielBrascher good point. if result is false, we can simply log a error message and return false.

ravening:
> > Is it OK to have one of the routers returning false? Would it make sense do break de loop and log an error/warn/debug message?
> 
> @GabrielBrascher good point. if result is false, we can simply log a error message and return false.
> @ravening what do you think ?

@GabrielBrascher Looked at the code and the function `applyRules` is throwing exception if anything bad happens. So yeah returning false makes sense if it fails for the first router and no need to continue

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
```diff
@@ -702,18 +702,32 @@ public boolean savePassword(final Network network, final NicProfile nic, final V
...
-                    userVmVO.setUpdateParameters(false);
-                    _userVmDao.update(userVmVO.getId(), userVmVO);
+                if (!result) {
+                    s_logger.error("Unable to save password for VM " + vm.getInstanceName() +
+                            " on router " + router.getInstanceName());
+                    return false;
                 }
-                return result;
+                isVrRunning = true;
+                savePasswordResult = savePasswordResult && result;
```
weizhouapache: @ravening it seems savePasswordResult  is not needed any more, because it is always true.

PR #3904, opened 2020-02-20
Avoid duplicate alerts when router state changes

ravening:
## Description

When both router of VPC are in MASTER state
then multiple alerts are sent equal to the number of tiers in the VPC.
If the VPC has 3 tiers then 6 alerts will be sent. This is not good
if VPC has more than 10 networks in it.

Instead of checking the router status for all the tiers in the VPC,
just check the status of rotuer for one tier in a VPC so that
multiple deplicate alerts can be avoided

## Types of changes

Alerts Dashboard when both routers are in MASTER state in VPC

Screenshot 2020-02-20 at 15 00 20

After the fix, when both routers are in master state

Screenshot 2020-02-20 at 15 09 16

## How Has This Been Tested?

1 . Create a VPC with redundant VR's and multiple tiers in it
2. Change the state of the BACKUP VR to MASTER.
3 . You will see multiple alerts in the Cloudstack dashboard


DaanHoogland:
Trusting Wei's test I'll run regression suites and merge
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

PR #3905, opened 2020-02-20
Fix network rules issue if default egress policy is Allow

ravening:
# Description

When the network service offering of a network is changed from
"defualt deny" to "default allow", the ping from VM to outside
world is stopped even after adding egress rule.

If the default policy is allow then the ping should never stop
working

## Types of changes

## How Has This Been Tested?

Steps to reproduce the issue

1 . Create a network offering with static nat enabled and default policy as "DENY"
2 . Create a vm in it and try to ping 8.8.8.8 from the vm
3 . Ping will fail
4. Now add an ICMP egress rule with code -1
5. Try to ping 8.8.8.8 from vm and this time ping will work
6. Add or remove any other egress rule and the ping should still work
7. Now create a new network offering with static nat enable and default egress policy as "ALLOW"
8. Change the network offering of the above created network to the newly created network offering
9. Now ssh to vm and try to ping 8.8.8.8
10. Ping wont work because of the existing ICMP egress rule.
11. Now remove the ICMP egress rule and try to ping 8.8.8.8 from VM
12. The ping still DOES NOT work

Expected result:

The ping should work at step 12 since default egress policy is ALLOW

Actual result:

The ping doesn't work even though the default egress policy is ALLOW

Now apply the patch and try to reproduce steps 1 to 12
Ping from VM to 8.8.8.8 will work at step 12.

### DIFF
--- a/server/src/main/java/com/cloud/network/router/VirtualNetworkApplianceManagerImpl.java
+++ b/server/src/main/java/com/cloud/network/router/VirtualNetworkApplianceManagerImpl.java
@@ -1812,10 +1812,8 @@ Configurable, StateListener<VirtualMachine.State, VirtualMachine.Event, VirtualM
         // Fetch firewall Egress rules.
         if (_networkModel.isProviderSupportServiceInNetwork(guestNetworkId, Service.Firewall, provider)) {
             firewallRulesEgress.addAll(_rulesDao.listByNetworkPurposeTrafficType(guestNetworkId, Purpose.Firewall, FirewallRule.TrafficType.Egress));
-            if (firewallRulesEgress.isEmpty()) {
-                //create egress default rule for VR
-                createDefaultEgressFirewallRule(firewallRulesEgress, guestNetworkId);
-            }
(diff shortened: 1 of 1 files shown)


weizhouapache:
> code looks good
> one question, we are now unconditionally adding a default rule for the network to the existing list (empty or not) is there any guarantee we don't add conflicting rules _in the right order_? (i.e. deny for some prot./port as well as allow.)

@DaanHoogland the rule for default egress policy is always added to the end of firewall rules list (it is not saved to database). it should be good.

https://github.com/apache/cloudstack/blob/18ad783f65d0c8e5a2df08c4c1e504b5963e210a/server/src/main/java/com/cloud/network/router/VirtualNetworkApplianceManagerImpl.java#L1949-L1969

DaanHoogland:
@blueorangutan package

DaanHoogland:
@andrijapanicsb @borisstoyanov @rhtyd any of you care to
@blueorangutan test

PR #3907, opened 2020-02-21
Allow port 80/8080 accessible only from guest network

ravening:
# Description

The port 80/8080 in the vr of the guest network is accessible from
outisde the world. Restrict it so that only the vm's in the guest
network can access them

Before fix

$ nc -vz 10.11.118.172 8080
Connection to 10.11.118.172 port 8080 [tcp/http] succeeded!

After fix
$ nc -vz 10.11.118.172 8080
^C

## Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsAddress.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsAddress.py
@@ -412,9 +412,9 @@ class CsIP:
             self.fw.append(
                 ["filter", "", "-A INPUT -i %s -p tcp -m tcp --dport 53 -s %s -j ACCEPT" % (self.dev, guestNetworkCidr)])
             self.fw.append(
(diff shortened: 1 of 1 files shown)


rhtyd:
In case of SSVM we may want port 80 connectivity to be able to download templates/isos/volumes.
@ravening can you check if you change adds regression for SSVM?

@blueorangutan package

weizhouapache:
@rhtyd this PR adds a rule only in method fw_vpcrouter and fw_router. ssvm should not be impacted.

rhtyd:
@blueorangutan test

rhtyd:
Thanks @weizhouapache - have you tested this in your env?

weizhouapache:
> Thanks @weizhouapache - have you tested this in your env?

@rhtyd yes, we have used this change in production for 3 years.

PR #3908, opened 2020-02-22
Alert on certain host status transitions

Slair1:
## Description

Add ability to create alert for host status transitions, such as from Alert->Up or Up->Disconnected, etc.  Right now some transitions may alert, but we wanted to know about any major host status transition.  

We also created a ConfigKey (alert.on.host.transiton) that defaults to the current behavior of not alerting on transitions.  But, by changing the setting to True, we are alerted if there are host transitions.  Which we find helpful - especially as once in a great while we'll have a host transition out of an Up status to something else.  But we never get an alert that it transitioned back to Up (even if it happened 1 second later)....

The ConfigKey works in conjunction with a flag set on the FSM's transitions.  If the TransitionAlertFlag for a particular transition is set to True AND the ConfigKey is set to True then the alert will be raised.

Let me know if 4.13 is OK or if you prefer a different branch instead of 4.13.

## Types of changes

Our environment has it set to True, but it defaults to False.
image
image

## How Has This Been Tested?

Tested putting hosts through various states.  Made sure we received the emails and the Alerts appear in the UI.

### DIFF
--- a/api/src/main/java/com/cloud/host/Status.java
+++ b/api/src/main/java/com/cloud/host/Status.java
@@ -97,6 +97,9 @@ public enum Status {
     public Status getNextStatus(Event e) throws NoTransitionException {
         return s_fsm.getNextState(this, e);
     }
+    public boolean getTransitionAlertFlag(Event e) throws NoTransitionException {
+        return s_fsm.getTransitionAlertFlag(this, e);
+    }
 
     public Status[] getFromStates(Event e) {
         List<Status> from = s_fsm.getFromStates(this, e);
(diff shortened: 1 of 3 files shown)


api/src/main/java/com/cloud/host/Status.java
```diff
@@ -119,45 +122,45 @@ public Status getNextStatus(Event e) throws NoTransitionException {
     static {
```
Slair1: if a true flag is passed at the end, that is the transitionAlertFlag.  If it is True then an alert should be raised.  If it is False or not passed at all, then an alert should NOT be raised.

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -1432,14 +1436,29 @@ public boolean agentStatusTransitTo(final HostVO host, final Status.Event e, fin
...
+                if (status_logger.isDebugEnabled()) {
+                    status_logger.debug("agentStatusTransitTo: old status - " + hostStatus + ", event - " + e + ", new status - " + nextStatus + " (" + (didTransit ? "success)" : "failed)"));
+                }
+                if (AlertOnHostTransitions.value() && hostStatus.getTransitionAlertFlag(e)) {
+                    final DateTimeFormatter dateFormatter = DateTimeFormatter.ofPattern("yyyy-MM-dd H:m:s,SSS");
+                    final String hostShortDesc = host.getName() + " (id:" + host.getId() + ")";
+                    _alertMgr.sendAlert(AlertManager.AlertType.ALERT_TYPE_HOST, host.getDataCenterId(), host.getPodId(), host.getName() + " transit from " + hostStatus + " to " + nextStatus +  " (" + (didTransit ? "success)" : "failed)"),
+                        (dateFormatter.format(LocalDateTime.now())) + "\r\nTransit agent status with event " + e + " for host " + hostShortDesc + " from " + hostStatus + " to " + nextStatus + (didTransit ? " was successful" : " failed"));
+                }
+
```
Slair1: Above code is what raises the alert if both the ConfigKey is set and the TransitionAlertFlag is set for the particular Transition

utils/src/main/java/com/cloud/utils/fsm/StateMachine2.java
```diff
@@ -50,7 +50,11 @@ public void addInitialTransition(E event, S toState) {
     }
 
     public void addTransition(S currentState, E event, S toState) {
-      addTransition(new Transition<S, E>(currentState, event, toState, null));
+      addTransition(currentState, event, toState, false);
+    }
+
+    public void addTransition(S currentState, E event, S toState, boolean alertFlag) {
+      addTransition(new Transition<S, E>(currentState, event, toState, null, alertFlag));
     }
```
Slair1: Old constructor just calls new constructor with False for the alertFlag.  So we don't break any functionality using the previous constructor

utils/src/main/java/com/cloud/utils/fsm/StateMachine2.java
```diff
@@ -171,15 +179,22 @@ public String toString() {
...
       public static enum Impact {
         USAGE
       }
 
       public Transition(S currentState, E event, S toState, List<Impact> impacts) {
+        this(currentState, event, toState, impacts, false);
+      }
+
+      public Transition(S currentState, E event, S toState, List<Impact> impacts, boolean alertFlag) {
         this.currentState = currentState;
```
Slair1: Old constructor just calls new constructor with False for the alertFlag. So we don't break any functionality using the previous constructor

DaanHoogland:
@Slair1 you marked this as new feature/enhancement. These only go on new releases. can you rebase to master?

Slair1:
> @Slair1 you marked this as new feature/enhancement. These only go on new releases. can you rebase to master?

yep sure can.  we are also getting alerts from this feature for our secondary storage VM in one of our environments.  We are tracking down if it is a problem with this PR or if this PR just revealed an issue in that environment that we were previously unaware of.

PR #3910, opened 2020-02-24
Jdk11 auto select on upgrade

rhtyd:
## Description

On install or upgrade, in the post-installation/configure, the packages can check and update the default java to JDK11. On most systems, the java is set to `auto` or the best/latest version available, but for in-place upgrades, users may miss reading the release notes.

## Types of changes

### DIFF
--- a/debian/cloudstack-management.postinst
+++ b/debian/cloudstack-management.postinst
@@ -63,6 +63,13 @@ if [ "$1" = configure ]; then
     grep -s -q "db.cloud.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS} || sed -i -e "\$adb.cloud.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS}
     grep -s -q "db.usage.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS} || sed -i -e "\$adb.usage.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS}
     grep -s -q "db.simulator.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS} || sed -i -e "\$adb.simulator.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS}
+
+    # Check and switch to Java 11
+    JDK11_NAME=$(update-java-alternatives --list | grep 11 | head -1 |  grep -Eo '^[^ ]+')
+    if [ -z ${JDK11_NAME} ]; then
+      echo "Setting default JAVA to ${JDK11_NAME}"
+      update-java-alternatives -s ${JDK11_NAME} > /dev/null
+    fi
 fi
 
 #DEBHELPER#
--- a/debian/cloudstack-management.postinst
+++ b/debian/cloudstack-management.postinst
@@ -66,9 +66,9 @@ if [ "$1" = configure ]; then
 
     # Check and switch to Java 11
     JDK11_NAME=$(update-java-alternatives --list | grep 11 | head -1 |  grep -Eo '^[^ ]+')
-    if [ -z ${JDK11_NAME} ]; then
+    if [ -n ${JDK11_NAME} ]; then
       echo "Setting default JAVA to ${JDK11_NAME}"
-      update-java-alternatives -s ${JDK11_NAME} > /dev/null
+      update-java-alternatives --jre-headless -s ${JDK11_NAME} > /dev/null
     fi
 fi
 
--- a/debian/cloudstack-agent.postinst
+++ b/debian/cloudstack-agent.postinst
@@ -53,6 +53,13 @@ case "$1" in
         fi
         cp -a /usr/share/cloudstack-agent/lib/libvirtqemuhook /etc/libvirt/hooks/qemu
 
+        # Check and switch to Java 11
+        JDK11_NAME=$(update-java-alternatives --list | grep 11 | head -1 |  grep -Eo '^[^ ]+')
+        if [ -z ${JDK11_NAME} ]; then
+          echo "Setting default JAVA to ${JDK11_NAME}"
+          update-java-alternatives -s ${JDK11_NAME} > /dev/null
+        fi
+
         ;;
 esac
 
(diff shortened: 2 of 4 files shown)


rhtyd:
Agree @DaanHoogland, we've already advised users via release/upgrade notes. I'm okay to close this PR if we would like to go via the docs way. Alternatively on post-install we can simply ask the admin to check and ensure that jdk11+ is configured without actually charging the default automatically. Thoughts?

DaanHoogland:
It's a difficult one @rhtyd . If I have hundreds of hosts and I didn't upgrade the jdk for some other reason, I don't want to be bothered with the question. But if i do have some agent/tooling installed that requires a newer jdk than we do, I don't want it to automatically be overruled leaving me with a disfunctional cloud. Can we easily decide that the active one is jdk8 and set the alternative if it is?

rhtyd:
@DaanHoogland like we did for jdk8, we've a section in the docs/upgrade notes on the Java11 requirement: https://github.com/apache/cloudstack-documentation/pull/87/files#diff-4bb612af30005312756eb8a61325f71fR18

I'm not sure if the automatic switch may cause errors. Should the doc change be enough, if you agree I'll close this?

DaanHoogland:
@rhtyd I agree, but we had questions. should we involve more people?

rhtyd:
@DaanHoogland let me ping a few more people if they've any thoughts - @PaulAngus @andrijapanicsb @borisstoyanov @weizhouapache @svenvogel @wido @GabrielBrascher @nathanejohnson @kiwiflyer @nvazquez ?

debian/cloudstack-management.postinst
```diff
@@ -63,6 +63,13 @@ if [ "$1" = configure ]; then
     grep -s -q "db.cloud.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS} || sed -i -e "\$adb.cloud.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS}
     grep -s -q "db.usage.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS} || sed -i -e "\$adb.usage.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS}
     grep -s -q "db.simulator.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS} || sed -i -e "\$adb.simulator.driver=jdbc:mysql" ${CONFDIR}/${DBPROPS}
+
+    # Check and switch to Java 11
+    JDK11_NAME=$(update-java-alternatives --list | grep 11 | head -1 |  grep -Eo '^[^ ]+')
+    if [ -z ${JDK11_NAME} ]; then
```
weizhouapache: @rhtyd it should be "-n" not "-z"

debian/cloudstack-management.postinst
```diff
@@ -63,6 +63,13 @@ if [ "$1" = configure ]; then
...
+      echo "Setting default JAVA to ${JDK11_NAME}"
+      update-java-alternatives -s ${JDK11_NAME} > /dev/null
```
weizhouapache: on Ubuntu 18.04, this command does not work.
the following command works
```update-java-alternatives --jre-headless -s java-1.11.0-openjdk-amd64```

GabrielBrascher:
@rhtyd I think that automating the Java upgrade process would be really cool. However, I agree with @DaanHoogland considerations in case the user needs to run another application on same host with a higher Java version.

One alternative would be to execute the java update as an optional step. In such a case, the user would be able to:
1. Accept: the person is aware that needs to change its Java version to 11 and is OK with that
2. Deny/Ignore: the person has a higher version of java installed
3. Stop (if possible to stop at that moment): upgrade process is stopped and the operator can review the upgrade process and get ready for Java +11

The easier way would be to put it as an upgrade step on documentation. For me, it would work fine as well.

debian/cloudstack-management.postinst
rhtyd: Ah yes, we want to do not empty, I'll fix.

rhtyd:
Thanks @weizhouapache I've addressed the comments.

@GabrielBrascher cc @DaanHoogland I think we have to enforce that CloudStack depends and only uses jre-11. The upgrade steps will echo when/if they need to switch default java/jre to 11. If a person has installed a higher version > 11 this will enforce and change to 11, because it will break one of the the env either for CloudStack (which may not say work against JRE14) or the app (which needs a jre other than version 11).

Perhaps there could be a middleground where we define JAVA in one of the /etc/default/cloudstack-* files and systemd uses JAVA JRE as defined in those default files than the global/default JRE. Thoughts - @DaanHoogland @andrijapanicsb @weizhouapache @GabrielBrascher @PaulAngus @borisstoyanov ?

rhtyd:
@blueorangutan package

DaanHoogland:
i kinda dissagree, @rhtyd . not strongly but i think a newer jre should be able to run ACS (as opposed to a newer jdk being able to build) I won't block for this but am still hanging towards doc instruction.

rhtyd:
@DaanHoogland it's the java/jre team that removed stuff in java11 for example. We cannot run jdk8 built mgmt server against jre-11; we expect it should work but it doesnot. Besides, per the LTS schedule jdk11 is the latest lts, other java releases 12,13,14 are fast releases not and we shouldn't use them in production.

rhtyd:
@blueorangutan test

rhtyd:
I'm closing this @DaanHoogland @andrijapanicsb @weizhouapache @GabrielBrascher as the release notes/doc already explain the Java11 requirement: https://github.com/apache/cloudstack-documentation/pull/87/files#diff-4bb612af30005312756eb8a61325f71f

I'm afraid any automatic switch may break few environments, we cannot assume a higher JRE version will allow mgmt server to run (jdk8 build mgmt server doesnot run against jre11 for example) and there is no other way to mitigate the upgrade issue other than the docs ^^. Reopen if we can agree with the PR which should solve for most environments except for very few which may run other java apps on mgmt server, or kvm hosts.

PR #3912, opened 2020-02-24
Fix infinite lease time

andrijapanicsb:
The previous setup of many hours would not work, due to some internal dnsmasq issues - lease was set correclty, but dnsmasq was setting the dhcp-renew-time (and rebind time) to less than 2 years from the date the lease was issued.

Using "infinite" as the value (instead of the number) works as expected - and (atm) the renew date is set to year 2088, etc..

## Types of changes

Confirmation inside user VM of correct renew/rebind/expire time
image

Configuration inside VRs
image
Different one:
image

## How Has This Been Tested?

See screenshot above


rhtyd:
Since this is a bugfix, @andrijapanicsb can you send for 4.13 instead?

andrijapanicsb:
done @rhtyd

andrijapanicsb:
Closing, will open against 4.13

PR #3913, opened 2020-02-24
Fix dhcp infinite lease time

andrijapanicsb:
The previous setup of many hours would not work, due to some internal dnsmasq issues - lease was set correclty, but dnsmasq was setting the dhcp-renew-time (and rebind time) to less than 2 years from the date the lease was issued.

Using "infinite" as the value (instead of the number) works as expected - and (atm) the renew date is set to year 2088, etc..

## Types of changes

Confirmation inside user VM of correct renew/rebind/expire time
image

Configuration inside VRs
image
Different one:
image

## How Has This Been Tested?

See screenshot above

### DIFF
--- a/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
@@ -183,9 +183,9 @@ class CsDhcp(CsDataBag):
 
         if entry['default_entry']:
             self.cloud.add("%s,%s,%s,%s" % (entry['mac_address'],
-                                             entry['ipv4_address'],
-                                             entry['host_name'],
-                                             lease))
+                                            entry['ipv4_address'],
+                                            entry['host_name'],
+                                            lease))
         else:
             tag = entry['ipv4_address'].replace(".", "_")
             self.cloud.add("%s,set:%s,%s,%s,%sh" % (entry['mac_address'],
--- a/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
+++ b/systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
@@ -174,18 +174,15 @@ class CsDhcp(CsDataBag):
 
     def add(self, entry):
         self.add_host(entry['ipv4_address'], entry['host_name'])
-        # Lease time set to effectively infinite (36000+ days) since we properly control all DHCP/DNS config via CloudStack.
+        # Lease time set to "infinite" since we properly control all DHCP/DNS config via CloudStack.
         # Infinite time helps avoid some edge cases which could cause DHCPNAK being sent to VMs since
         # (RHEL) system lose routes when they receive DHCPNAK.
         # When VM is expunged, its active lease and DHCP/DNS config is properly removed from related files in VR,
         # so the infinite duration of lease does not cause any issues or garbage.
-        # There will be soon a PR which also regenerates the /var/lib/misc/dnsmasq.leases (active lease DB file)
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

andrijapanicsb:
> removing `lease = randint(870000, 870010)` is certainly OK. Not sure how `infinite` works. code lgtm

It works as one can imagine (I was intrusted by the dnsmasq ML guy to use that word and tested it works fine)

andrijapanicsb:
@blueorangutan test

svenvogel:
@andrijapanicsb LGTM and a small question. To ensure that the ip addresses are static in Cloudstack we use such a high lease time and they are only be changed from the operator. right?

andrijapanicsb:
@svenvogel yes and not - I mean we can not change this (unless one edits the script/systemVM.iso) - here we are ensuring renew time is in far, far future ("infinite" as defined by the dnsmasq options) - thus effectively VM will never attempt to renew it's IP and thus will not be possible to fail to renew IP from the VR which has no knowledge of the issues lease (new VR that is created as part of the network restart with clean up - the /var/lib/misc/dnsmasq.leases is missing in this new VR - so VR will reject any DHCPRENEW request, and VM will eventually fail back to to  "initialisation" process with DHCPDISCOVER etc - but some OS (RHEL) will remove/delete their static routes when this happens...

svenvogel:
> thus effectively VM will never attempt to renew it's IP and thus will not be possible to fail to renew IP from the VR which has no knowledge of the issued lease 

@andrijapanicsb Thanks for your explaination. 👍 I didn't have this sentence on my screen in particular.

DaanHoogland:
@blueorangutan test

DaanHoogland:
@weizhouapache @wido @GabrielBrascher @rhtyd @nvazquez and anybody else,...
I am going to merge this on no objections. It has quite an impact, so i'll leave it for a day for you all to have a chance to comment.

andrijapanicsb:
(I've tested also on Windows 2019 that it gets the renewal time of "-1", as opposed to proper numbers - so will have to assume it's OK for this fellow as well)

(Also tested CentOS 8, more for fun than for anything else)

systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
```diff
@@ -174,18 +174,15 @@ def write_hosts(self):
...
         # so the infinite duration of lease does not cause any issues or garbage.
-        # There will be soon a PR which also regenerates the /var/lib/misc/dnsmasq.leases (active lease DB file)
-        # in the new VR (when restarting network with cleanup), which will help around RHEL edge cases (described above)
-        # for the VMs who are already running in productions systems with 30d lease time.
-        lease = randint(870000, 870010)
+        lease = 'infinite'
 
         if entry['default_entry']:
-            self.cloud.add("%s,%s,%s,%sh" % (entry['mac_address'],
+            self.cloud.add("%s,%s,%s,%s" % (entry['mac_address'],
```
DaanHoogland: pylint is not accepting this, @andrijapanicsb

systemvm/debian/opt/cloud/bin/cs/CsDhcp.py
```diff
@@ -174,18 +174,15 @@ def write_hosts(self):
...
                                              entry['ipv4_address'],
                                              entry['host_name'],
                                              lease))
```
DaanHoogland: you need to unindent by one space as you made the base statement one char shorter as well. :P

rhtyd:
I see @DaanHoogland has addressed his own review comment and travis is green. Tests LGTM.

Issue #3893, opened 2020-02-18
VR Health check errors seen in logs


rhtyd:
Saw this on latest master env, this may need checking - 

```
2020-02-18 07:58:01,574 ERROR [c.c.n.r.VirtualNetworkApplianceManagerImpl] (RouterStatusMonitor-1:ctx-3bbeb667) (logid:e9e4832d) Fail to complete the FetchRouterHealthChecksResultTask! 
com.google.gson.JsonIOException: java.io.EOFException: End of input
	at com.google.gson.Streams.parse(Streams.java:49)
	at com.google.gson.Gson.fromJson(Gson.java:497)
	at com.google.gson.Gson.fromJson(Gson.java:467)
	at com.google.gson.Gson.fromJson(Gson.java:417)
	at com.cloud.network.router.VirtualNetworkApplianceManagerImpl.updateDbHealthChecksFromRouterResponse(VirtualNetworkApplianceManagerImpl.java:1518)
	at com.cloud.network.router.VirtualNetworkApplianceManagerImpl$FetchRouterHealthChecksResultTask.getFailingChecks(VirtualNetworkApplianceManagerImpl.java:1271)
	at com.cloud.network.router.VirtualNetworkApplianceManagerImpl$FetchRouterHealthChecksResultTask.runInContext(VirtualNetworkApplianceManagerImpl.java:1250)
	at org.apache.cloudstack.managed.context.ManagedContextRunnable$1.run(ManagedContextRunnable.java:49)
	at org.apache.cloudstack.managed.context.impl.DefaultManagedContext$1.call(DefaultManagedContext.java:56)
	at org.apache.cloudstack.managed.context.impl.DefaultManagedContext.callWithContext(DefaultManagedContext.java:103)
	at org.apache.cloudstack.managed.context.impl.DefaultManagedContext.runWithContext(DefaultManagedContext.java:53)
	at org.apache.cloudstack.managed.context.ManagedContextRunnable.run(ManagedContextRunnable.java:46)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
[...]

PR #3919, opened 2020-02-28
Handle EOFException during VR Health Check

Pearl1594:
## Description
VR health check throws EOFException due to malformed/Invalid JSON.

Addresses issue: https://github.com/apache/cloudstack/issues/3893

## Types of changes

### DIFF
--- a/utils/src/main/java/com/cloud/utils/ssh/SshHelper.java
+++ b/utils/src/main/java/com/cloud/utils/ssh/SshHelper.java
@@ -34,7 +34,6 @@ import com.trilead.ssh2.ChannelCondition;
 import com.trilead.ssh2.Connection;
 import com.trilead.ssh2.Session;
 import com.cloud.utils.Pair;
-import com.trilead.ssh2.StreamGobbler;
 
 public class SshHelper {
     private static final int DEFAULT_CONNECT_TIMEOUT = 180000;
@@ -185,11 +184,11 @@ public class SshHelper {
 
             sess.execCommand(command);
 
-            StreamGobbler stdout = new StreamGobbler(sess.getStdout());
-            StreamGobbler stderr = new StreamGobbler(sess.getStderr());
+            InputStream stdout = sess.getStdout();
+            InputStream stderr = sess.getStderr();
 
-            byte[] buffer = new byte[16384];
-            StringBuffer sbResult = new StringBuffer(214748364);
+            byte[] buffer = new byte[8192];
+            StringBuffer sbResult = new StringBuffer();
             int currentReadBytes = 0;
             while (true) {
                 throwSshExceptionIfStdoutOrStdeerIsNull(stdout, stderr);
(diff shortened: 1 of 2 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

utils/src/main/java/com/cloud/utils/ssh/SshHelper.java
```diff
@@ -184,12 +185,11 @@ public static void scpTo(String host, int port, String user, File pemKeyFile, St
...
-            InputStream stdout = sess.getStdout();
-            InputStream stderr = sess.getStderr();
-
-            byte[] buffer = new byte[8192];
-            StringBuffer sbResult = new StringBuffer();
+            StreamGobbler stdout = new StreamGobbler(sess.getStdout());
+            StreamGobbler stderr = new StreamGobbler(sess.getStderr());
 
+            byte[] buffer = new byte[16384];
+            StringBuffer sbResult = new StringBuffer(214748364);
```
rhtyd: @Pearl1594 what's this magic number?

Pearl1594: @rhtyd the number represents the number of characters that corresponds to 200MB

utils/src/main/java/com/cloud/utils/ssh/SshHelper.java
```diff
@@ -207,22 +206,18 @@ public static void scpTo(String host, int port, String user, File pemKeyFile, St
                     if (canEndTheSshConnection(waitResultTimeoutInMs, sess, conditions)) {
                         break;
                     }
-
                 }
 
-                while (stdout.available() > 0) {
-                    currentReadBytes = stdout.read(buffer);
-                    sbResult.append(new String(buffer, 0, currentReadBytes));
+               while((currentReadBytes = stdout.read(buffer)) > 0) {
```
rhtyd: @Pearl1594 instead of `> 0` we probably want to check for readbytes `!= -1`. For example: (comments only for your understanding)
```
currentReadBytes = stdout.read(buffer);
while (currentReadBytes >= 0) { // or check for != -1
  sbResult.append(new String(buffer, 0, currentReadBytes));
  currentReadBytes = stdout.read(buffer); // read() will return -1 on eof/complete
}
```
The other option is to use a more modern java-way to use `BufferedReader` for example like:
```
 BufferedReader bufferedStdOut = new BufferedReader(stdout);
.... use
```

rhtyd:
Changes lgtm, some further improvements advised. Let's test them now.
@blueorangutan package

utils/src/main/java/com/cloud/utils/ssh/SshHelper.java
Pearl1594: @rhtyd I did consider using that, however decided against it so at to avoid wrapping the Input Stream. Also used the readAllBytes() but that would increase the memory footprint as indicated earlier. So, went ahead with this approach.

rhtyd:
@blueorangutan package

utils/src/main/java/com/cloud/utils/ssh/SshHelper.java
DaanHoogland: @Pearl1594 long shot but you are using a test for '> 0'. What if the stream is blocking due to timing issues but not ended yet? Wouldn't we still end reading before end of stream?

DaanHoogland: never mind, guess you already changed that.

DaanHoogland:
@blueorangutan test

DaanHoogland:
Anybody testing? @PaulAngus @Pearl1594 @borisstoyanov @andrijapanicsb

andrijapanicsb:
I can, if I'm told how to reproduce the original bug (once I tried this some time ago, I couldn't reproduce it - but I know it's a valid issue)

rhtyd:
it's a rare case, perhaps @PaulAngus can advice who was able to see/reproduce it @andrijapanicsb @DaanHoogland

rhtyd:
Merging this.

PR #3960, opened 2020-03-11
Rename max.retries setting

nvazquez:
## Description
Simply rename a global setting name to a more descriptive one:

`max.retries` to `vm.ha.migration.max.retries`

Also, as it is used in global context, have changed the scope from Cluster to Global.

This setting has been introduced before 4.14, so no need to update DB on upgrade

Setting improvement as well as #3961 

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

Issue #3942, opened 2020-03-05
Host reports RAM allocated, when no running VMs


andrijapanicsb:
**Problem**: host reported as being fully RAM allocated when all VMs are stopped on this host.

Reproduce:
Host has 6 VMs running.
Stop all 6 VMs - zero VMs running on this host.
Fetch the latest metrics via Dashboard "Fetch latest" button.
Go to host, click on it, open the Statistics tab - it will show RAM memory as allocated (sum of RAM memory of all VMs) - so no new VMs cant be started on it.

I've traced this to the last_host_id value in vm_instance table.

Test workaround:
- Open DB editor, find all VMs where last_host_id = <hostID> and change this to another host ID - so that no VMs are referencing this particular host. Fetch the latest statistics again - it will show reduction of RAM allocated.

CloudStack Master (4.14) (one of the PRs - needs to be tested in the actual Master as well)
KVM (Ubunut18, should not be relevant)

andrijapanicsb:
I've observed this in the Rolling Maintenance PR - so need to test Master as well.

andrijapanicsb:
@weizhouapache haven't you fixed something similar - so that the stopped VMs doesn't "count" against the capacities?

andrijapanicsb:
Confirmed the issue in 4.13

weizhouapache:
@andrijapanicsb might the resource be reserved for stopped vm?

weizhouapache:
@andrijapanicsb 
just tested it, when vm is stopped, the resources (cpu/ram) will be reserved for 1 hour (configurable in global setting).

1. before stopping vm
```
mysql> select host_id,used_capacity,reserved_capacity,total_capacity,capacity_type from op_host_capacity where host_id=14;
+---------+---------------+-------------------+----------------+---------------+
| host_id | used_capacity | reserved_capacity | total_capacity | capacity_type |
+---------+---------------+-------------------+----------------+---------------+
|      14 |    3221225472 |                 0 |     7297568768 |             0 |
|      14 |          2500 |                 0 |           9196 |             1 |
|      14 |             4 |                 0 |              4 |            90 |
+---------+---------------+-------------------+----------------+---------------+
3 rows in set (0.01 sec)
```

2. after stopping vm
```
mysql> select host_id,used_capacity,reserved_capacity,total_capacity,capacity_type from op_host_capacity where host_id=14;
+---------+---------------+-------------------+----------------+---------------+
[...]

PR #3961, opened 2020-03-11
Reduce nonsence value for capacity.skipcounting.hours

andrijapanicsb:
When doing rapid host maintenance one after the other, and when previously stopping a huge number of VMs on a single host (i.e. all VMs on that host are stopped) - no new VMs can be started on this empty host since those VMs' capacities are reserved for 3600sec - so when doing maintenance of another host, VMs can't be migrated to this empty host.

Not sure who and why has this setting default value to such high value (1h) (doesn't make sense)
Not to mention the name of the setting is wrong, should be "capacity.skipcounting.second"

Another PR to rename this setting is most welcome @nvazquez @DaanHoogland @weizhouapache

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-41300to41400.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41300to41400.sql
@@ -19,6 +19,11 @@
 -- Schema upgrade from 4.13.0.0 to 4.14.0.0
 --;
 
+-- Don't reserve capacities for stopped VMs by default as this influences normal capacity checks when doing i.e. rolling host maintenance, etc.
+-- One is free to increase this to a high value in case it's needed in his environment, otherwise makes no sense in most environments,
+-- as new VM can't be started on an empty host which was consumed previously by many VMS which are now stopped
+UPDATE `cloud`.`configuration` SET `default_value`='10' WHERE  `name`='capacity.skipcounting.hours';
+
 -- KVM: enable storage data motion on KVM hypervisor_capabilities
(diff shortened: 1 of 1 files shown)


weizhouapache:
@andrijapanicsb is is better to remove this configuration from database and add new config via Configkey ?

andrijapanicsb:
@weizhouapache if it doesn't exists in config key form, I guess that's OK? How about #3960

weizhouapache:
> @weizhouapache if it doesn't exists in config key form, I guess that's OK? How about #3960

@andrijapanicsb I think these are different.
max.retries was added in 4.13->4.14, since 4.14 is not released yet, we do not need extra sql change if we rename it.
capacity.skipcounting.hours was added quite long ago (before 2013), so extra sql change is needed if we rename it. we have two options (1) rename it in sql (2) delete it in sql and add new ConfigKey.

andrijapanicsb:
sorry @weizhouapache my bad (read something wrong...) - so whatever works for you - if the config key-based addition of this one is not there, yes please (don't expect me to know how to add config key :) - so feel free to edit my PR please?

andrijapanicsb:
@weizhouapache I was expecting a PR from @nvazquez to rename this setting to xxx..seconds (not hours) - that's what I was mentioning (linked to wrong PR)... so we can put everything in a single PR?

DaanHoogland:
👎 no config key should be added in this way

andrijapanicsb:
Closed due to not being able to accomplish the full goal - as discussed in #3942

PR #3965, opened 2020-03-13
server: Honor vm.destroy.forcestop when expunge a vm

ustcweizhou:
## Description

When expunge a Running vm, vm will be stopped with forcestop=false which does not make sense. we should honor vm.destroy.forcestop in global setting, or always set forcestop=true.

## Types of changes


rhtyd:
Usually we would expunged a destroyed VM which must be already stopped?
LGTM nonetheless.

weizhouapache:
> Usually we would expunged a destroyed VM which must be already stopped?
> LGTM nonetheless.

actually there are two apis doing the same thing.

```
expunge virtualmachine id=<vm id>
destroy virtualmachine id=<vm id> expunge=True
```

vm can be Running, Stopped, Destroyed.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3649, opened 2019-10-21
Garbage snapshots are left on Primary and Secondary

GabrielBrascher:
## Description

Tested both cases:
- Snapshot backed up on Secondary (stored on primary and secondary) 
- Snapshot stored only on primary storage.

This fix works for KVM  and XenServer.

**Note**: I changed the name of the strategy from _XenserverSnapshotStrategy_ to _DefaultSnapshotStrategy_ due to the fact that the "Xenserver" strategy handles also Ceph, KVM, and do return `StrategyPriority.DEFAULT` in any case except _REVERT_.

```
    @Override
    public StrategyPriority canHandle(Snapshot snapshot, SnapshotOperation op) {
        if (SnapshotOperation.REVERT.equals(op)) {
            long volumeId = snapshot.getVolumeId();
            VolumeVO volumeVO = volumeDao.findById(volumeId);

            if (volumeVO != null && ImageFormat.QCOW2.equals(volumeVO.getFormat())) {
                return StrategyPriority.DEFAULT;
            }
            return StrategyPriority.CANT_HANDLE;
        }
        return StrategyPriority.DEFAULT;
    }
```

Fixes: #3646

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

PR #3969, opened 2020-03-16
Snapshot deletion issues

### LINKED ISSUE #3646: Garbage snapshots are left on Primary and Secondary storage after a volume snapshot is deleted
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
snapshots handling
~~~

##### CLOUDSTACK VERSION

~~~
4.11.2, 4.13.0.0 tested
~~~

##### CONFIGURATION

Xs and NFS    or    KVM and NFS/Ceph (have not tested VMware)

##### SUMMARY

When a snapshot is deleted (manually, or as part of the scheduled snap policy), snapshot is NOT deleted from Primary Storage (except KVM + Ceph, which is being fixed in https://github.com/apache/cloudstack/pull/3615) and is also NOT deleted from Secondary Storage. 
DB entries are inconsistent/partially cleared.

##### STEPS TO REPRODUCE

XS+NFS tested as following
Make snap (XS plus NFS), delete snap - snap is left on Primary Storage AND on Secondary Storage.
Same goes for scheduled snap policy. "Primary" store ref row is deleted from the snapshots_store_ref but not really deleted on Primary Store. "image" store ref row is NOT deleted from snapshots_store_ref and image is NOT deleted on Secondary Store (after muiltiple runs of storage clean up thread i.e. storage.cleanup.delay=60, storage.cleanup.delay=60.

KVM+Ceph tested as per the updates in https://github.com/apache/cloudstack/pull/3615 and will be tracker here.

NOTE: XenserverSnapshotStrategy is NOT handling only XS, also some KVM
Pay attention to the managed storage.
This whole "garbage" problem was not there in 4.8 at least (possibly also 4.9).

##### EXPECTED RESULTS

~~~
snapshot is deleted in both primary and secondary storage (after the cleanup thread runs)
~~~

##### ACTUAL RESULTS

~~~
garbage is left, left and right
~~~

DaanHoogland:
## Description

_copy of PR #3649 on a joint account for easier cooperation._

Tested both cases:
- Snapshot backed up on Secondary - manual and scheduled
- Snapshot stored only on primary storage (works with KVM + Ceph only)

**Note**: @GabrielBrascher changed the name of the strategy from _XenserverSnapshotStrategy_ to _DefaultSnapshotStrategy_ due to the fact that the "Xenserver" strategy handles also Ceph, KVM, and do return `StrategyPriority.DEFAULT` in any case except _REVERT_.

```
    @Override
    public StrategyPriority canHandle(Snapshot snapshot, SnapshotOperation op) {
        if (SnapshotOperation.REVERT.equals(op)) {
            long volumeId = snapshot.getVolumeId();
            VolumeVO volumeVO = volumeDao.findById(volumeId);

            if (volumeVO != null && ImageFormat.QCOW2.equals(volumeVO.getFormat())) {
                return StrategyPriority.DEFAULT;
            }
        }
        return StrategyPriority.DEFAULT;
    }
```

Fixes: #3646

## Types of changes

## WHAT HAS BEEN FIXED / WORKs FINE:

- VMware, KVM - no garbage on Secondary/Primary store (snapshot.backup.to.secondary=TRUE)👍 
-- snapshot.backup.to.secondary=FALSE - works (by design) only with KVM+Ceph - no garbage on Secondary/Primary storage 👍 
---  snapshot.backup.to.secondary=FALSE  does NOT work (by design) with other hypervisors/storage combo - so don't bother trying 😄 
- ad-hoc single XenServer volume snapshot (i.e.NOT consecutive snaps/scheduled snaps 👎 )

## WHAT HAS NOT BEEN FIXED (tracked via #4018 ):

- **Major**: XenServer consecutive/scheduled volume snapshots, since having parent/child relation _(vs. VMware and KVM where consecutive/scheduled snaps are NOT chained (no parent/child relation))_ - continues to be broken - i.e. one can continue to create/use XS volume snaps, but they are not removed properly on Primary and Secondary storage and thus leave garbage all over the place - eventually, there might happen the "Snapshot chain too long" error from Xen (see #4018 for the possible workaround/fix and a bit more explanation on how it works atm) 👎 👎 👎 
- **Minor**: VMware and KVM PRIMARY references (role_type=primary) are NOT removed from the snapshot_store_ref table - minor garbage, but to be fixed eventually 👎

### DIFF
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -290,7 +290,7 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
         if (snapshotVO != null && snapshotVO.getState() == Snapshot.State.Destroyed) {
             deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
         } else {
-            // This is to handle snapshots which are created only on primary when snapshot.backup.to.secondary is set to false. 
+            // This is to handle snapshots which are created only on primary when snapshot.backup.to.secondary is set to false.
             SnapshotObject obj = (SnapshotObject) snapshotOnPrimaryInfo;
             try {
                 obj.processEvent(Snapshot.Event.DestroyRequested);
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -288,10 +288,12 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
             // This may occur for instance when they are created only on primary because
             // snapshot.backup.to.secondary was set to false.
             if (snapshotOnPrimaryInfo == null) {
-                s_logger.debug("Snapshot id:" + snapshotId + " not found on primary storage, skipping snapshot deletion on primary and marking it destroyed");
+                if (s_logger.isDebugEnabled()) {
+                    s_logger.debug(String.format("Snapshot (id: %d) not found on primary storage, skipping snapshot deletion on primary and marking it destroyed", snapshotId));
+                }
                 snapshotVO.setState(Snapshot.State.Destroyed);
                 snapshotDao.update(snapshotId, snapshotVO);
-                return true;
+                deletedOnPrimary = true;
             } else {
                 SnapshotObject obj = (SnapshotObject) snapshotOnPrimaryInfo;
                 try {
@@ -365,15 +367,21 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
     private boolean deleteSnapshotOnPrimary(Long snapshotId, SnapshotInfo snapshotOnPrimaryInfo) {
         SnapshotDataStoreVO snapshotOnPrimary = snapshotStoreDao.findBySnapshot(snapshotId, DataStoreRole.Primary);
         if (isSnapshotOnPrimaryStorage(snapshotId)) {
-            s_logger.debug("Snapshot reference is found on primary storage for snapshot id:" + snapshotId + ", performing snapshot deletion on primary");
+            if (s_logger.isDebugEnabled()) {
+                s_logger.debug(String.format("Snapshot reference is found on primary storage for snapshot id: %d, performing snapshot deletion on primary", snapshotId));
+            }
             if (snapshotSvr.deleteSnapshot(snapshotOnPrimaryInfo)) {
                 snapshotOnPrimary.setState(State.Destroyed);
                 snapshotStoreDao.update(snapshotOnPrimary.getId(), snapshotOnPrimary);
-                s_logger.debug("Successfully deleted snapshot id:" + snapshotId + " on primary storage");
+                if (s_logger.isDebugEnabled()) {
+                    s_logger.debug(String.format("Successfully deleted snapshot id: %d on primary storage", snapshotId));
+                }
                 return true;
             }
         } else {
-            s_logger.debug("Snapshot reference is not found on primary storage for snapshot id:" + snapshotId + ", skipping snapshot deletion on primary");
+            if (s_logger.isDebugEnabled()) {
+                s_logger.debug(String.format("Snapshot reference is not found on primary storage for snapshot id: %d, skipping snapshot deletion on primary", snapshotId));
+            }
             return true;
         }
         return false;
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -308,7 +308,7 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
         Boolean deletedOnSecondary = false;
         SnapshotInfo snapshotOnImage = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Image);
         if (snapshotOnImage == null) {
-            s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on backup storage", snapshotId));
+            s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on secondary storage", snapshotId));
         } else {
             SnapshotObject obj = (SnapshotObject)snapshotOnImage;
             try {
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -267,7 +267,6 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
         boolean deletedOnSecondary = false;
         if (snapshotOnImage == null) {
             s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on backup storage", snapshotId));
-            snapshotDao.remove(snapshotId);
         } else {
             SnapshotObject obj = (SnapshotObject)snapshotOnImage;
             try {
@@ -285,7 +284,27 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
             }
         }
 
-        boolean deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId);
+        boolean deletedOnPrimary = false;
+        snapshotVO = snapshotDao.findById(snapshotId);
+        SnapshotInfo snapshotOnPrimaryInfo = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Primary);
+        if (snapshotVO != null && snapshotVO.getState() == Snapshot.State.Destroyed) {
+            deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
+        } else {
+            // This is to handle snapshots which are created only on primary when snapshot.backup.to.secondary is set to false. 
+            SnapshotObject obj = (SnapshotObject) snapshotOnPrimaryInfo;
+            try {
+                obj.processEvent(Snapshot.Event.DestroyRequested);
+                deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
+                if (!deletedOnPrimary) {
+                    obj.processEvent(Snapshot.Event.OperationFailed);
+                } else {
+                    obj.processEvent(Snapshot.Event.OperationSucceeded);
+                }
+            } catch (NoTransitionException e) {
+            s_logger.debug("Failed to set the state to destroying: ", e);
+            return false;
+            }
+        }
         if (deletedOnPrimary) {
             s_logger.debug(String.format("Successfully deleted snapshot (id: %d) on primary storage.", snapshotId));
         } else if (deletedOnSecondary) {
@@ -321,9 +340,8 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
      * Deletes the snapshot on primary storage. It can return false when the snapshot was not stored on primary storage; however this does not means that it failed to delete the snapshot. </br>
      * In case of failure, it will throw one of the following exceptions: CloudRuntimeException, InterruptedException, or ExecutionException. </br>
      */
-    private boolean deleteSnapshotOnPrimary(Long snapshotId) {
+    private boolean deleteSnapshotOnPrimary(Long snapshotId, SnapshotInfo snapshotOnPrimaryInfo) {
         SnapshotDataStoreVO snapshotOnPrimary = snapshotStoreDao.findBySnapshot(snapshotId, DataStoreRole.Primary);
-        SnapshotInfo snapshotOnPrimaryInfo = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Primary);
         if (isSnapshotOnPrimaryStorage(snapshotId) && snapshotSvr.deleteSnapshot(snapshotOnPrimaryInfo)) {
             snapshotOnPrimary.setState(State.Destroyed);
             snapshotStoreDao.update(snapshotOnPrimary.getId(), snapshotOnPrimary);
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -267,6 +267,7 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
         boolean deletedOnSecondary = false;
         if (snapshotOnImage == null) {
             s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on backup storage", snapshotId));
+            snapshotDao.remove(snapshotId);
         } else {
             SnapshotObject obj = (SnapshotObject)snapshotOnImage;
             try {
@@ -326,7 +327,6 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
         if (isSnapshotOnPrimaryStorage(snapshotId) && snapshotSvr.deleteSnapshot(snapshotOnPrimaryInfo)) {
             snapshotOnPrimary.setState(State.Destroyed);
             snapshotStoreDao.update(snapshotOnPrimary.getId(), snapshotOnPrimary);
-            snapshotDao.remove(snapshotId);
             return true;
         }
         return false;
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -262,9 +262,49 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
             throw new InvalidParameterValueException("Can't delete snapshotshot " + snapshotId + " due to it is in " + snapshotVO.getState() + " Status");
         }
 
-        SnapshotInfo snapshotOnImage = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Image);
+        boolean deletedOnSecondary = deleteOnSecondaryIfNeeded(snapshotId);
+        boolean deletedOnPrimary = deleteOnPrimaryIfNeeded(snapshotId);
 
+        if (deletedOnPrimary) {
+            s_logger.debug(String.format("Successfully deleted snapshot (id: %d) on primary storage.", snapshotId));
+        } else if (deletedOnSecondary) {
+            s_logger.debug(String.format("The snapshot was deleted on secondary storage. Could not find/delete snapshot (id: %d) on primary storage.", snapshotId));
+        }
+
+        return deletedOnSecondary || deletedOnPrimary;
+    }
+
+    private boolean deleteOnPrimaryIfNeeded(Long snapshotId) {
+        SnapshotVO snapshotVO;
+        boolean deletedOnPrimary = false;
+        snapshotVO = snapshotDao.findById(snapshotId);
+        SnapshotInfo snapshotOnPrimaryInfo = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Primary);
+        if (snapshotVO != null && snapshotVO.getState() == Snapshot.State.Destroyed) {
+            deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
+        } else {
+            // Here we handle snapshots which are to be deleted but are not marked as destroyed yet.
+            // This may occur for instance when they are created only on primary because
+            // snapshot.backup.to.secondary was set to false.
+            SnapshotObject obj = (SnapshotObject) snapshotOnPrimaryInfo;
+            try {
+                obj.processEvent(Snapshot.Event.DestroyRequested);
+                deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
+                if (!deletedOnPrimary) {
+                    obj.processEvent(Snapshot.Event.OperationFailed);
+                } else {
+                    obj.processEvent(Snapshot.Event.OperationSucceeded);
+                }
+            } catch (NoTransitionException e) {
+                s_logger.debug("Failed to set the state to destroying: ", e);
+                deletedOnPrimary = false;
+            }
+        }
+        return deletedOnPrimary;
+    }
+
+    private boolean deleteOnSecondaryIfNeeded(Long snapshotId) {
         boolean deletedOnSecondary = false;
+        SnapshotInfo snapshotOnImage = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Image);
         if (snapshotOnImage == null) {
             s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on backup storage", snapshotId));
         } else {
@@ -280,37 +320,10 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
                 }
             } catch (NoTransitionException e) {
                 s_logger.debug("Failed to set the state to destroying: ", e);
-                return false;
+                deletedOnSecondary = false;
             }
         }
-
-        boolean deletedOnPrimary = false;
-        snapshotVO = snapshotDao.findById(snapshotId);
-        SnapshotInfo snapshotOnPrimaryInfo = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Primary);
-        if (snapshotVO != null && snapshotVO.getState() == Snapshot.State.Destroyed) {
-            deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
-        } else {
-            // This is to handle snapshots which are created only on primary when snapshot.backup.to.secondary is set to false.
-            SnapshotObject obj = (SnapshotObject) snapshotOnPrimaryInfo;
-            try {
-                obj.processEvent(Snapshot.Event.DestroyRequested);
-                deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
-                if (!deletedOnPrimary) {
-                    obj.processEvent(Snapshot.Event.OperationFailed);
-                } else {
-                    obj.processEvent(Snapshot.Event.OperationSucceeded);
-                }
-            } catch (NoTransitionException e) {
-            s_logger.debug("Failed to set the state to destroying: ", e);
-            return false;
-            }
-        }
-        if (deletedOnPrimary) {
-            s_logger.debug(String.format("Successfully deleted snapshot (id: %d) on primary storage.", snapshotId));
-        } else if (deletedOnSecondary) {
-            s_logger.debug(String.format("The snapshot was deleted on secondary storage. Could not find/delete snapshot (id: %d) on primary storage.", snapshotId));
-        }
-        return deletedOnSecondary || deletedOnPrimary;
+        return deletedOnSecondary;
     }
 
     /**
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -305,7 +305,7 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
     }
 
     private Boolean deleteOnSecondaryIfNeeded(Long snapshotId) {
-        Boolean deletedOnSecondary = false;
+        Boolean deletedOnSecondary = null;
         SnapshotInfo snapshotOnImage = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Image);
         if (snapshotOnImage == null) {
             s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on secondary storage", snapshotId));
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -257,21 +257,23 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
             return true;
         }
 
-        if (!Snapshot.State.BackedUp.equals(snapshotVO.getState()) && !Snapshot.State.Error.equals(snapshotVO.getState()) &&
+        if (!Snapshot.State.BackedUp.equals(snapshotVO.getState()) &&
                 !Snapshot.State.Destroying.equals(snapshotVO.getState())) {
             throw new InvalidParameterValueException("Can't delete snapshotshot " + snapshotId + " due to it is in " + snapshotVO.getState() + " Status");
         }
 
-        boolean deletedOnSecondary = deleteOnSecondaryIfNeeded(snapshotId);
+        Boolean deletedOnSecondary = deleteOnSecondaryIfNeeded(snapshotId);
         boolean deletedOnPrimary = deleteOnPrimaryIfNeeded(snapshotId);
 
         if (deletedOnPrimary) {
             s_logger.debug(String.format("Successfully deleted snapshot (id: %d) on primary storage.", snapshotId));
-        } else if (deletedOnSecondary) {
-            s_logger.debug(String.format("The snapshot was deleted on secondary storage. Could not find/delete snapshot (id: %d) on primary storage.", snapshotId));
+        } else {
+            s_logger.debug(String.format("The snapshot (id: %d) could not be found/deleted on primary storage.", snapshotId));
         }
-
-        return deletedOnSecondary || deletedOnPrimary;
+        if (null == deletedOnSecondary && deletedOnSecondary) {
+            s_logger.debug(String.format("Successfully deleted snapshot (id: %d) on secondary storage.", snapshotId));
+        }
+        return (deletedOnSecondary != null) && deletedOnSecondary || deletedOnPrimary;
     }
 
     private boolean deleteOnPrimaryIfNeeded(Long snapshotId) {
@@ -302,8 +304,8 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
         return deletedOnPrimary;
     }
 
-    private boolean deleteOnSecondaryIfNeeded(Long snapshotId) {
-        boolean deletedOnSecondary = false;
+    private Boolean deleteOnSecondaryIfNeeded(Long snapshotId) {
+        Boolean deletedOnSecondary = false;
         SnapshotInfo snapshotOnImage = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Image);
         if (snapshotOnImage == null) {
             s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on backup storage", snapshotId));
@@ -320,7 +322,7 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
                 }
             } catch (NoTransitionException e) {
                 s_logger.debug("Failed to set the state to destroying: ", e);
-                deletedOnSecondary = false;
+//                deletedOnSecondary remain null
             }
         }
         return deletedOnSecondary;
--- a/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
+++ b/engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
@@ -287,18 +287,25 @@ public class DefaultSnapshotStrategy extends SnapshotStrategyBase {
             // Here we handle snapshots which are to be deleted but are not marked as destroyed yet.
             // This may occur for instance when they are created only on primary because
             // snapshot.backup.to.secondary was set to false.
-            SnapshotObject obj = (SnapshotObject) snapshotOnPrimaryInfo;
-            try {
-                obj.processEvent(Snapshot.Event.DestroyRequested);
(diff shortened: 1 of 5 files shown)


DaanHoogland:
@blueorangutan package

DaanHoogland:
using KVM/NFS, I tested this scenario:

1. created VM with data disk
2. both disks snapshotted
3. both deleted
4. both marked deleted in the DB and removed secondary
5. no changes in primary (not sure if the snapshots were removed in between

basically different results than reported by @GabrielBrascher .

this is with @GabrielBrascher 's fix in this PR.
that would mean we are good to go, but it could also mean there are unknown circumstances that create garbage. (cc @andrijapanicsb )

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -269,63 +262,87 @@ public boolean deleteSnapshot(Long snapshotId) {
...
-                    obj.processEvent(Snapshot.Event.OperationFailed);
-                } catch (NoTransitionException e1) {
-                    s_logger.debug("Failed to change snapshot state: " + e1.toString());
+        boolean deletedOnSecondary = false;
+        if (snapshotOnImage == null) {
+            s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on backup storage", snapshotId));
+        } else {
+            SnapshotObject obj = (SnapshotObject)snapshotOnImage;
+            try {
+                deletedOnSecondary = deleteSnapshotOnSecundaryStorage(snapshotId, snapshotOnImage, obj);
```
GabrielBrascher: going to fix the "secundary" typo

GabrielBrascher:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
Not good (still testing a few more scenarios and will update this comment once done):

## XENSERVER (7.1):

### snapshot.backup.to.secondary = TRUE (default)
MANUAL SNAPS: LGTM
- deletion works fine - removed from both primary/secondary store, DB records in snapshost_store_reg marked as DESTROYED, all good.
SCHEDULED SNAPS: MAYBE OK - someone should advise
- for disk snaps we are making snap chains (parent_snapshot_id pointing to the ID of the previous scheduled snapshot) and while DB says that older snaps are deleted, it's not really true on storage side, but as those are snap diff snaps (chain)  - it makes sense that it doesn't delete the older/parent snap while there are child/linked snaps), but an older snap can be selectively deleted manually - which is a bit strange, as it would break the snap chain)

### snapshot.backup.to.secondary = FALSE
MANUAL SNAPS: LGTM
- deletion works fine - removed from both primary store, DB records in snapshost_store_ref marked as DESTROYED, all good.
SCHEDULED SNAPS:
- TO BE TESTED - will edit here once it completes+




## KVM (CENT7.1)

### snapshot.backup.to.secondary TRUE (default)
MANUAL SNAPS: LGTM
- deletion works fine - removed from both primary/secondary store, DB records in snapshost_store_reg marked as DESTROYED, all good.
- I also see we are NOT keeping the snap on the QCOW2 file (good, as we made a copy to sec storage) - and even though the "PRIMARY" record (store_role) is still created and kept as "Ready" in snapshost_store_ref - it does get properly marked as DESTROYED when snap is deleted with no exception
SCHEDULED SNAPS:
- works fine, created and deleted as expected in both DB and on the primary/secondary stor.

### snapshot.backup.to.secondary = FALSE
MANUAL SNAPS: BROKEN
- can't even create manual KVM volume snap
SCHEDULED SNAPS: NA
- not tested as manual ones fail


### VMware not tested, 
but will re-test that it's NOT affected once we have everything else fixed.

GabrielBrascher:
@andrijapanicsb thanks for all the tests, I am going to check this issue.

GabrielBrascher:
@andrijapanicsb thank you very much for all the tests. Here follows an update on  `snapshot.backup.to.secondary = FALSE`.

I have been testing this implementation (tested with 4.13.1 and also on a port to 4.14.0). The snapshotting with `snapshot.backup.to.secondary = FALSE` seems to be working fine for me. Created and deleted on (i) XenServer + NFS, (ii) KVM + NFS, (iii) KVM + Ceph.

Environment:
- CloudStack management node: Ubuntu 18.04
- KVM nodes: Ubuntu 18.04
- XenServer 6.5

Can you please share more details on the issue that you had? Are there exceptions on management or the agent node? Is the DB  being correctly updated (snapshots, volumes, snapshot_store_ref)?

Note that snapshots on NFS as primary storage is a bit different than on secondary. The snapshot handling for NFS (primary) is executed with the help of the [managesnapshot.sh](https://github.com/apache/cloudstack/blob/da865f08d30df5e177154049215fb6f4b54e5252/scripts/storage/qcow2/managesnapshot.sh#L72) script. That script on KVM + NFS (primary) creates a lvm, which makes a bit tricky to find the snapshot volume compared to other approaches (e.g. ceph: `rbd snap ls foo`, nfs secondary: `ls -lah /export/secondary/snapshots/`, XenServer: `xe vdi-list`).

Basically the script runs a lvm lvcreate: `lvm lvcreate --size ${lv_bytes}b --name "${snapshotname}-cow" ${vg} >&2 || return 2`. Let's say that the volume path is `/mnt/<uuid primary>/<uuid volume>`.

Thus, the snapshot path will become something as `/mnt/<uuid primary>/<uuid volume>/<uuid snapshot>`. However, one looking at the NFS storage via `ls /mnt/<uuid primary>/` will see something as `/mnt/<uuid primary>/<uuid volume>`.

Is that happening on your test environment?

davidjumani:
@GabrielBrascher That is happening in our environment. However, when I create a first snapshot of a data volume, it also makes a copy in the secondary storage. Successive snapshots do not get created in the secondary storage but when I try to restore them or create a new volume from them it fails, while I can only use the first snapshot that I had created to create a new volume / restore the snapshot.

GabrielBrascher:
Got it @davidjumani, thanks for the feedback.

andrijapanicsb:
So - yes, when using KVM+NFS and snapshot.backup.to.secondary=false - there is all kind of mess with snapshots happening - both in 4.11.3 and in master - which implies lack of proper code implementation for this combo.

The rest looks good and I'm tempted to give this specific PR a LGTM (after some more testing of what was already tested before as good)

andrijapanicsb:
I proposed that we continue/merge this PR as it is (I want to confirm a few more things with testing, specifically that VMware is still not affected) and close the "garbage" problem as it seems solved.

For the KVM+NFS/local very bad behaviour when snapshot.backup.to.secondary=false - that is traced back to the original implementation, which seems only done for KVM + CEPH, and so I propose that we simply change the description of this global settinh to ... Applicable only for KVM = Ceph"

@GabrielBrascher sounds ok to you?

GabrielBrascher:
@andrijapanicsb sounds good to me, considering the raised context and all tests done so far. Let's run a few more tests then.

andrijapanicsb:
@blueorangutan test matrix

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -269,63 +262,87 @@ public boolean deleteSnapshot(Long snapshotId) {
...
+     * Deletes the snapshot on primary storage. It can return false when the snapshot was not stored on primary storage; however this does not means that it failed to delete the snapshot. </br>
+     * In case of failure, it will throw one of the following exceptions: CloudRuntimeException, InterruptedException, or ExecutionException. </br>
+     */
+    private boolean deleteSnapshotOnPrimary(Long snapshotId) {
+        SnapshotDataStoreVO snapshotOnPrimary = snapshotStoreDao.findBySnapshot(snapshotId, DataStoreRole.Primary);
+        SnapshotInfo snapshotOnPrimaryInfo = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Primary);
+        if (isSnapshotOnPrimaryStorage(snapshotId) && snapshotSvr.deleteSnapshot(snapshotOnPrimaryInfo)) {
+            snapshotOnPrimary.setState(State.Destroyed);
+            snapshotStoreDao.update(snapshotOnPrimary.getId(), snapshotOnPrimary);
+            snapshotDao.remove(snapshotId);
```
davidjumani: Perhaps this is leading to the record having a removed date on the snapshots table which later on causes issues during GC ?

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -269,63 +262,87 @@ public boolean deleteSnapshot(Long snapshotId) {
...
-            volumesFromSnapshot = _volumeDetailsDaoImpl.findDetails("SNAPSHOT_ID", String.valueOf(snapshotId), null);
 
-            if (volumesFromSnapshot.size() > 0) {
-                try {
-                    obj.processEvent(Snapshot.Event.OperationFailed);
-                } catch (NoTransitionException e1) {
-                    s_logger.debug("Failed to change snapshot state: " + e1.toString());
+        boolean deletedOnSecondary = false;
+        if (snapshotOnImage == null) {
+            s_logger.debug(String.format("Can't find snapshot [snapshot id: %d] on backup storage", snapshotId));
```
davidjumani: Can remove the snapshot entry here instead!

harikrishna-patnala: Yes David, snapshotDao.remove(snapshotId) needs to be moved here. 
cc @GabrielBrascher

andrijapanicsb:
@blueorangutan package

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
GabrielBrascher: That looks like the issue indeed, I will take a second look at this.

GabrielBrascher (commented):
Thanks for the observations @davidjumani @harikrishna-patnala.

I got offline yesterday and had a few technical problems on my Dev environment. I will take another look at this today.

andrijapanicsb:
when NOT backing up the snap to secondary - the snap fails to be deleted....damn it 😢 

2020-04-01 15:29:22,117 DEBUG [c.c.a.ApiServlet] (qtp504527234-17:ctx-3353a255 ctx-cc5b2b2c) (logid:0f9171aa) ===END===  10.1.0.1 -- GET  command=deleteSnapshot&id=1e63d646-eadd-4a1d-b8b8-ddec90fc9e0e&response=json&_=1585754941759
2020-04-01 15:29:22,133 DEBUG [o.a.c.s.s.DefaultSnapshotStrategy] (API-Job-Executor-3:ctx-c4bb5e71 job-284 ctx-75e649e7) (logid:37b859e0) Can't find snapshot
[snapshot id: 19] on backup storage
2020-04-01 15:29:22,141 DEBUG [o.a.c.s.s.SnapshotObject] (API-Job-Executor-3:ctx-c4bb5e71 job-284 ctx-75e649e7) (logid:37b859e0) Failed to update state:java.lang.NullPointerException
2020-04-01 15:29:22,142 DEBUG [c.c.s.s.SnapshotManagerImpl] (API-Job-Executor-3:ctx-c4bb5e71 job-284 ctx-75e649e7) (logid:37b859e0) Failed to delete snapshot: 19:java.lang.NullPointerException
2020-04-01 15:29:22,147 ERROR [c.c.a.ApiAsyncJobDispatcher] (API-Job-Executor-3:ctx-c4bb5e71 job-284) (logid:37b859e0) Unexpected exception while executing org.apache.cloudstack.api.command.user.snapshot.DeleteSnapshotCmd
com.cloud.utils.exception.CloudRuntimeException: Failed to delete snapshot:java.lang.NullPointerException
        at com.cloud.storage.snapshot.SnapshotManagerImpl.deleteSnapshot(SnapshotManagerImpl.java:611)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:338)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:197)
        at org.apache.cloudstack.network.contrail.management.EventUtils$EventInterceptor.invoke(EventUtils.java:107)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:174)
        at com.cloud.event.ActionEventInterceptor.invoke(ActionEventInterceptor.java:51)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:174)
        at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:92)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:185)
        at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212)
        at com.sun.proxy.$Proxy199.deleteSnapshot(Unknown Source)
        at org.apache.cloudstack.api.command.user.snapshot.DeleteSnapshotCmd.execute(DeleteSnapshotCmd.java:103)
        at com.cloud.api.ApiDispatcher.dispatch(ApiDispatcher.java:156)
        at com.cloud.api.ApiAsyncJobDispatcher.runJob(ApiAsyncJobDispatcher.java:108)
        at org.apache.cloudstack.framework.jobs.impl.AsyncJobManagerImpl$5.runInContext(AsyncJobManagerImpl.java:586)
        at org.apache.cloudstack.framework.jobs.impl.AsyncJobManagerImpl$5.run(AsyncJobManagerImpl.java:534)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)

andrijapanicsb:
I've made other extensive tests and everything works fine when snapshot.backup.to.secondary=TRUE - including GC on the snapshot_store_ref table.

But snap can't be deleted when snapshot.backup.to.secondary=FALSE

harikrishna-patnala:
Have seen couple of issues when handling deletesnapshot for snapshots backed up only in primary storage and snapshots backed up in secondary storage. I have made a fix for that but unable to push the changes due to some access issues. I'll update the PR once I resolve my access issue.0

Following are the test cases I covered after the fix is made
1. Create data volumes vol1 and vol2
2. Set global parameter snapshots.backup.to.secondary = true
3. Create snapshot snap1vol1 from vol1
4. Set global parameter snapshots.backup.to.secondary = false
5. Create snapshot snap2vol2 from vol2 
6. Create snapshot snap3vol1 form vol1
7. Delete snap2vol2 // This is deletion of snapshot which is created only in primary and no entries in snapshot_store_ref table against store_role ‘Image’
8. Delete snap3vol1 // This is deletion of snapshot which is created only in primary but MS can get a snapshot_store_ref entry against store_role ‘Image’ with id of vol1 because snap1vol1 are created in both primary and secondary storage from vol1
9. Delete snap1vol1 // This is deletion of snapshot which is created both in primary and secondary

Attached the patch if someone can add it on behalf of me.
[0001-Fix-deletesnapshot-worflow-to-handle-both-snapshots-.patch.zip](https://github.com/apache/cloudstack/files/4428067/0001-Fix-deletesnapshot-worflow-to-handle-both-snapshots-.patch.zip)

GabrielBrascher:
@harikrishna-patnala thanks for sharing that commit. Can you please share a branch with that commit?

The access issue is due to the fact that Apache branches are open only for committers.

nvazquez:
@GabrielBrascher @harikrishna-patnala I have pushed Hari's commit to this PR
@blueorangutan package

andrijapanicsb:
@blueorangutan test

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -285,7 +284,27 @@ public boolean deleteSnapshot(Long snapshotId) {
...
+                if (!deletedOnPrimary) {
+                    obj.processEvent(Snapshot.Event.OperationFailed);
+                } else {
+                    obj.processEvent(Snapshot.Event.OperationSucceeded);
+                }
+            } catch (NoTransitionException e) {
+            s_logger.debug("Failed to set the state to destroying: ", e);
+            return false;
+            }
+        }
```
DaanHoogland: @harikrishna-patnala can you factor this bit out, please?

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -285,7 +284,27 @@ public boolean deleteSnapshot(Long snapshotId) {
...
+                deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
+                if (!deletedOnPrimary) {
+                    obj.processEvent(Snapshot.Event.OperationFailed);
+                } else {
+                    obj.processEvent(Snapshot.Event.OperationSucceeded);
+                }
+            } catch (NoTransitionException e) {
+            s_logger.debug("Failed to set the state to destroying: ", e);
+            return false;
+            }
```
DaanHoogland: and than factor this bit out again?

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -285,7 +284,27 @@ public boolean deleteSnapshot(Long snapshotId) {
...
         }
 
-        boolean deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId);
+        boolean deletedOnPrimary = false;
+        snapshotVO = snapshotDao.findById(snapshotId);
+        SnapshotInfo snapshotOnPrimaryInfo = snapshotDataFactory.getSnapshot(snapshotId, DataStoreRole.Primary);
+        if (snapshotVO != null && snapshotVO.getState() == Snapshot.State.Destroyed) {
+            deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
+        } else {
+            // This is to handle snapshots which are created only on primary when snapshot.backup.to.secondary is set to false.
```
DaanHoogland: this comment does not seem to be in alignment with the code. How do we conclude that we are at this point in code because of the global setting 'snapshot.backup.to.secondary'?

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -285,7 +284,27 @@ public boolean deleteSnapshot(Long snapshotId) {
...
+                obj.processEvent(Snapshot.Event.DestroyRequested);
+                deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
+                if (!deletedOnPrimary) {
+                    obj.processEvent(Snapshot.Event.OperationFailed);
+                } else {
+                    obj.processEvent(Snapshot.Event.OperationSucceeded);
+                }
+            } catch (NoTransitionException e) {
+            s_logger.debug("Failed to set the state to destroying: ", e);
+            return false;
```
DaanHoogland: indent?

DaanHoogland (commented):
a refactor is desireable

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
harikrishna-patnala: There is will be only 'Primary' store entry in snapshot_store_ref table for a snapshot created when snapshot.backup.to.secondary is set to false. so deletion of snapshot on secondary store will fail and it won't mark snapshot as destroyed and then goes to 'else' case. In other case (snapshot.backup.to.secondary is set to true) 'Image' and 'Primary' entries will be there in which it goes to 'if' case.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

harikrishna-patnala:
Code changes on the last commit from @DaanHoogland looks good to me.

andrijapanicsb:
Latest testing of the deletion stuff reports below

(previously I have tested normal operations works fine - convert to this to that, revert (KVM only), etc, for all 3 hypervisors)

## snapshot.backup.to.secondary=FALSE

Expected to work ONLY with KVM+Ceph (not other variants, not other hypervisors)
(but the basic create/delete snap was tested with KVM+NFS to make sure nothing broken in that specific flow)
create, delete snap works fine - GC kicks in, no exceptions - after GC runs, the "PRIMARY" (role_type=primary) row in snapshots_store_ref is left undeleted (with other snapshot.backup.to.secondary=TRUE testing as well)

@rafaelweingartner please test with Ceph if possible

## snapshot.backup.to.secondary=TRUE

- snaps can be created, operated with, deleted fine
-- deleted from file systems as expected (State changed to Destroyed in snapshots and snapshot_store_ref tables - "removed"date NOT set in the main "snapshots" table
--- GC will remove the IMAGE row (role_type=image) from the snapshot_store_ref table and set the "removed" date in the main "snapshots" table
-- GC will NOT remove the PRIMARY row from the snapshot_store_ref table (any of possibly many rows for the same volume (i.e. many different snapss of the same volume)
-- When a volume (root/data) is removed, then only first/latest (not sure) PRIMARY row in the snapshot_store_ref table- all other PRIMARY rows for a particular volume are left - i.e. DB garbage is left

### GC garbage workaround

Remove rows from snapshot_store_ref table:

> delete from snapshot_store_ref where snapshot_id in (select id from snapshots where status="Destroyed" and removed is NOT NULL); 

(removed not NULL = means that the GC has run and left the garbage behind)

or the safer one (volume is removed)

> delete from snapshot_store_ref where snapshot_id in (select id from snapshots where status="Destroyed" and removed is NOT NULL) and volume_id in (select id from volumes where state="Expunged" and removed is NOT NULL);

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
yes should be !=, tnx (it's late)

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -257,21 +257,23 @@ public boolean deleteSnapshot(Long snapshotId) {
...
         if (deletedOnPrimary) {
             s_logger.debug(String.format("Successfully deleted snapshot (id: %d) on primary storage.", snapshotId));
-        } else if (deletedOnSecondary) {
-            s_logger.debug(String.format("The snapshot was deleted on secondary storage. Could not find/delete snapshot (id: %d) on primary storage.", snapshotId));
+        } else {
+            s_logger.debug(String.format("The snapshot (id: %d) could not be found/deleted on primary storage.", snapshotId));
         }
-
-        return deletedOnSecondary || deletedOnPrimary;
+        if (null == deletedOnSecondary && deletedOnSecondary) {
```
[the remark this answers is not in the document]
DaanHoogland: yes should be !=, tnx (it's late)

GabrielBrascher:
I did a few tests and it does not delete on secondary storage.

Tests considered the following scenarios (with `snapshot.backup.to.secondary=TRUE`):
1.  KVM + Ceph (primary) & NFS (scondary).

Result: snapshot deleted properly on primary storage; snapshot stays on secondary. The snapshot state is Ready at table `snapshot_store_ref`.

2.  KVM + NFS (primary & secondary).

Result: snapshot deleted properly on primary storage; snapshot stays on secondary. The snapshot state is Destroying at table `snapshot_store_ref`.

I am debugging and re-checking previous commits to find at which point this stopped working.

GabrielBrascher:
@DaanHoogland @andrijapanicsb please ignore the previous test results.

The issues that I got were related with SSVM not working (VM running, but agent was down). Therefore, those failures when deleting snapshot on secondary storage are not related with this PR.

The fact that the deletion worked without any exception even with SSVM failing to delete de snapshot on secondary is a problem. It could lead to snapshot garbage issue on secondary storage. But that is not critical and we can (should) leave such issue for a future release.

DaanHoogland:
@GabrielBrascher @andrijapanicsb (@harikrishna-patnala ) that last commit seems to be a state never tested. I don't think the code allows or should allow for a NoSuchTransitionException so I do not know how we can force it to happen.

DaanHoogland:
@blueorangutan package

harikrishna-patnala:
> @GabrielBrascher @andrijapanicsb (@harikrishna-patnala ) that last commit seems to be a state never tested. I don't think the code allows or should allow for a NoSuchTransitionException so I do not know how we can force it to happen.

NoSuchTransitionException won't occur at all if all state transitions are defined properly. So only way I can think of testing is in dev testing that too by changing the state intermediately on setting breakpoints. 
Also here we are not doing anything in the code, returning null or returning false does the same logging.

DaanHoogland:
@blueorangutan test matrix

andrijapanicsb:
@GabrielBrascher @DaanHoogland I've tested deleting the snaps while SSVM was destroyed - indeed the delete action is "successful" but the file is left on Secondary Storage - however, it gets deleted once GC kicks in (assuming SSVM is up again at that moment) - I've confirmed this explicitly.

To me this all looks good - certainly MUCH better than what we started with a few months ago - and if there are still some edge cases, we can touch that subject in 4.15 or similar.

@harikrishna-patnala @GabrielBrascher @DaanHoogland - if anyone is OK with this, please LGTM it so we have a consensus - if not, please explain why

andrijapanicsb:
(will wait for test matrix to finish)

harikrishna-patnala:
Based on the last testing that I have done and @andrijapanicsb's test results this is in good shape now.
LGTM

GabrielBrascher (approved):
LGTM based on tests.

All worked fine with Ceph as primary storage; snapshots are deleted on both cases (`snapshot.backup.to.secondary=TRUE` or `snapshot.backup.to.secondary=FALSE`)

andrijapanicsb:
XS test that needs investigation:

2020-04-07 16:45:40,456 - CRITICAL - EXCEPTION: test_03_list_snapshots: ['Traceback (most recent call last):\n', '  File "/usr/lib/python2.7/site-packages/nose/suite.py", line 228, in run\n    self.tearDown()\n', '  File "/usr/lib/python2.7/site-packages/nose/suite.py", line 351, in tearDown\n    self.teardownContext(ancestor)\n', '  File "/usr/lib/python2.7/site-packages/nose/suite.py", line 367, in teardownContext\n    try_run(context, names)\n', '  File "/usr/lib/python2.7/site-packages/nose/util.py", line 471, in try_run\n    return func()\n', '  File "/marvin/tests/smoke/test_list_ids_parameter.py", line 209, in tearDownClass\n    raise Exception("Warning: Exception during cleanup : %s" % e)\n', "Exception: Warning: Exception during cleanup : Job failed: {jobprocstatus : 0, created : u'2020-04-07T16:46:13+0000', jobresult : {errorcode : 530, errortext : u**'Failed to delete snapshot'**}, cmd : u'org.apache.cloudstack.api.command.user.snapshot.DeleteSnapshotCmd', userid : u'964f52cb-78d6-11ea-bc32-1e00400108b3', jobstatus : 2, jobid : u'8806aed0-896c-4264-801e-d570217f4ae0', jobresultcode : 530, jobinstanceid : u'4cdc1214-942c-436c-b6ae-738f3220b1cf', jobresulttype : u'object', jobinstancetype : u'Snapshot', completed : u'2020-04-07T16:46:13+0000', accountid : u'964f4d54-78d6-11ea-bc32-1e00400108b3'}\n"]

VMware tests that failed:

'org.apache.cloudstack.api.command.admin.vm.MigrateVMCmd\', userid : u\'3d686233-78da-11ea-8ab1-1e009101089c\', jobstatus : 2, jobid : u\'aace648a-e269-49f8-982b-3c07bf5baa11\', jobresultcode : 530, jobresulttype : u\'object\', jobresult : {errorcode : 530, errortext : u"MigrationCommand failed due to Exception: java.lang.RuntimeException\\nMessage: **The target host does not support the virtual machine\'s current hardware requirements.**\\n"}, accountid : u\'3d685acc-78da-11ea-8ab1-1e009101089c\'}\n']
That is funny...

andrijapanicsb:

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -287,18 +287,25 @@ private boolean deleteOnPrimaryIfNeeded(Long snapshotId) {
...
-                deletedOnPrimary = deleteSnapshotOnPrimary(snapshotId, snapshotOnPrimaryInfo);
-                if (!deletedOnPrimary) {
-                    obj.processEvent(Snapshot.Event.OperationFailed);
-                } else {
-                    obj.processEvent(Snapshot.Event.OperationSucceeded);
+            if (snapshotOnPrimaryInfo == null) {
+                s_logger.debug("Snapshot id:" + snapshotId + " not found on primary storage, skipping snapshot deletion on primary and marking it destroyed");
+                snapshotVO.setState(Snapshot.State.Destroyed);
+                snapshotDao.update(snapshotId, snapshotVO);
+                return true;
```
DaanHoogland, line 294: no extra return please; this means `deletedOnPrimary = true` and we fall through?

engine/storage/snapshot/src/main/java/org/apache/cloudstack/storage/snapshot/DefaultSnapshotStrategy.java
```diff
@@ -352,14 +359,21 @@ private boolean deleteSnapshotOnSecondaryStorage(Long snapshotId, SnapshotInfo s
...
+     * Deletes the snapshot on primary storage. It returns true when the snapshot was not found on primary storage; </br>
+     * In case of failure while deleting the snapshot, it will throw one of the following exceptions: CloudRuntimeException, InterruptedException, or ExecutionException. </br>
      */
     private boolean deleteSnapshotOnPrimary(Long snapshotId, SnapshotInfo snapshotOnPrimaryInfo) {
         SnapshotDataStoreVO snapshotOnPrimary = snapshotStoreDao.findBySnapshot(snapshotId, DataStoreRole.Primary);
-        if (isSnapshotOnPrimaryStorage(snapshotId) && snapshotSvr.deleteSnapshot(snapshotOnPrimaryInfo)) {
-            snapshotOnPrimary.setState(State.Destroyed);
-            snapshotStoreDao.update(snapshotOnPrimary.getId(), snapshotOnPrimary);
+        if (isSnapshotOnPrimaryStorage(snapshotId)) {
+            s_logger.debug("Snapshot reference is found on primary storage for snapshot id:" + snapshotId + ", performing snapshot deletion on primary");
```
DaanHoogland, line 368: can you surround debug statements that include string manupulation in teh paramater list with `if (LOG.isDebugEnables()) {}` please

DaanHoogland (commented):
style remarks, changes/logic look good.

Slair1:
Is this branch (4.13) missing this related PR? #3194

andrijapanicsb:
@blueorangutan package

DaanHoogland:
@Slair1 :
```
$ git branch --contains c94ee1454d3ecdf1bf371fd18c8e3ec0631d3092
  4.11
  4.12
  4.13
...
```

andrijapanicsb:
So it does @Slair1 ...

andrijapanicsb:
Ape, do your thing!  @blueorangutan test matrix

Slair1:
> So it does @Slair1 ...

@andrijapanicsb could you double-check the code?  When i manually check 4.13, i don't see the lines added in the screenshot below from #3194 .  
image

andrijapanicsb:
@Slair1 I can't, I'm off as from 10min ago (late night, time for beer) - perhaps @DaanHoogland can confirm once more tomorrow....

andrijapanicsb:
Is this for volume or VM snapshots @Slair1 ? - as volume snapshots are NOT possible for KVM while the VM is running (someone thought that's interesting to do that, due to qcow2 safety...so now it's useless for KVM...)

Slair1:
> Is this for volume or VM snapshots @Slair1 ? - as volume snapshots are NOT possible for KVM while the VM is running (someone thought that's interesting to do that, due to qcow2 safety...so now it's useless for KVM...)

It is for volume snapshots, but and you're correct the global setting kvm.snapshot.enabled needs to be enabled.  But this brings up another point, we are wondering if we need to do a suspend before deleting VM Snapshots also.  From our examination of the code, it looks like both VM Snapshots and Volume Snapshots use the save libvirt libraries to take and delete the snapshot... so I would think they could both have the same corruption issue if the snapshot is deleted while the VM is powered-on and not suspended during the snapshot delete...

andrijapanicsb:

andrijapanicsb:
@blueorangutan test

DaanHoogland:
those were spurious failures obviously, @andrijapanicsb as they passed after applying insanity. If you manual verification checks out we should merge and release.

andrijapanicsb (approved):
LGTM from my side:

- 3 x LGTMs (2 x Approvals), test matrix (KVM,VMware,XS) runs fine (known privategtw test issues, fixed in master)
- Manual tests fine (with some outstanding fix/refactoring work to be done via issue #4018 

Will update the PR description to explain what HAS and HAS NOT been fixed.

Merging.

Slair1:
> > So it does @Slair1 ...
> 
> @andrijapanicsb could you double-check the code? When i manually check 4.13, i don't see the lines added in the screenshot below from #3194 .
> image


@andrijapanicsb just wanted to ping you on this again thanks!

andrijapanicsb:
I don't see it myself, despite @DaanHoogland confirming the commit being present in 4.13 branch... Some regression by someone?

andrijapanicsb:
but you can test 4.13.1 RC1 and see if VMs are suspended as expected - perhaps some parts of the code were refractored or so...

Slair1:
> but you can test 4.13.1 RC1 and see if VMs are suspended as expected - perhaps some parts of the code were refractored or so...

@andrijapanicsb @DaanHoogland from our testing it is just gone... so strange - i hope other commits didn't get dropped somehow.  I even checked the "blame" and I don't see a refactor - the last change in branch 4.13 was 4-years ago for the section of code.

andrijapanicsb:
Agree, it's gone/messed up (trying to understand what happened with some of my colleagues who are good with git) - some code refactoring happened between from 4.11 to 4.12, and it seems gone during this... (from 4.12 and onwards) - so it's NOT new issue (that code missing)

andrijapanicsb:
@Slair1 do you mind creating a PR for MASTER (not 4.13 - we already have RC1 for vote today) to add back those few lines... it's incredible - the global desc that the VM will be suspended is STILL there in 4.12/4.13/master, but the "vm.suspend" code is 
@DaanHoogland your opinion?

andrijapanicsb:
@Slair1 - the bug you shared (RHEL bug), was marked as WON'T FIX - and was raised against qemu-kvm-0.12.1.2-2.355.el6_4.1.x86_64

Can you please describe in which env (versions of qemu) you are seeing the corruption (as master/4.14 does NOT even support centos6)
The issue happened during emerging 4.11 into 4.12 branch - we identified WHEN it was messed up, but not HOW (which is now irrelevant)

I would like to first REPRODUCE this issue on centos7+ instead of just suspending the VM (now that we think to AGAIN introduce the suspend thing)

Slair1:
> @Slair1 - the bug you shared (RHEL bug), was marked as WON'T FIX - and was raised against qemu-kvm-0.12.1.2-2.355.el6_4.1.x86_64
> 
> Can you please describe in which env (versions of qemu) you are seeing the corruption (as master/4.14 does NOT even support centos6)
> The issue happened during emerging 4.11 into 4.12 branch - we identified WHEN it was messed up, but not HOW (which is now irrelevant)
> 
> I would like to first REPRODUCE this issue on centos7+ instead of just suspending the VM (now that we think to AGAIN introduce the suspend thing)

Thanks @andrijapanicsb .  We have always been on CentOS7 - never 6.  We have had many corrupt VMs from using the scheduled volume snapshot functionality...  We had to turn off the ability to do volume snapshots while a VM is powered-on...  We also stopped using VM Snapshots since the Delete VM Snapshot functionality deletes snapshots the same way as volume snapshots are deleted from the qcow2 files (after the snap is backed up to secondary storage).  We are hoping the pause VM, delete snapshot, resume VM fixes this - we are still testing

andrijapanicsb:
Clear, I did a bit of reading myself. Say, what about corruption when creating snapshots? - in recent qemu releases (2.10) qemu-img actually opens qcow2 with read-write access instead of read-only previously (don't ask me why, documented behavior). This would probably mean that even backing up the snap to Secondary Storage (with qemu 2.10) will not work (requires a use of --force-share flag, which we don't have implemented in ACS) unless the VM is suspended - did you have a change to test qemu 2.10 (that is separate Centos SIG repo), etc? What about corruption while creating snapshots with stock qemu version 1.53?

andrijapanicsb:
tested, confirmed the issue with latest centos7 host (centos5.5 ext3 guest - simulated high IO load while deleting the snap and broke the VM)

Raised PR (for master only) https://github.com/apache/cloudstack/pull/4029

ravening:
@DaanHoogland @GabrielBrascher @andrijapanicsb @rhtyd 

Found an issue with this commit. Below are the steps to reproduce

1. Create VM snapshot from running vm
2. Create a snapshot from the above VM snapshot
3. Delete the snapshot
4. Try to create snapshot again. This will fail.
This is because when we delete the snapshot for the first time, its deleting the VM snapshot as well in the primary storage.
So when we try to create snapshot again, its not able to find the vm snapshot and throws an exception


I have made a fix so that it checks if the vm snapshot from which the snapshot is created, still exists on primary storage or not. If it exists then its not deleted from primary but only from secondary

andrijapanicsb:
@ravening for which hypervisor is this?
if VM snapshots is deleted, we should also delete it from the Primary Storage - that is fine. 
This sounds more like some DB garbage, due to which we can't create a new one (chained, parent/child snaps?)
Which PR did you create?

Issue #2736, opened 2018-07-09
If a VM resource is requested to transition in a state where it's already in, it returns success, instead it should return 'the resource is in the same state' 


borisstoyanov:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VM handler 

~~~

##### CLOUDSTACK VERSION

~~~
4.11
~~~

##### SUMMARY

If a VM is already running on HostA and the Admin requests a Start on HostB, it will return a success with the VM being started on HostA. Instead it should return an error to the administrator that the resource is already in running state.

##### STEPS TO REPRODUCE

1. Start a VM on HostA
2. With cloudmonkey, start the same VM on HostB
3. Observe the result

##### EXPECTED RESULTS

~~~
Response should fail the job and say VM is already in the same state. 
~~~

##### ACTUAL RESULTS

~~~
Job succeeds with VM not being the requested host 
~~~

DaanHoogland:
I would argue that the expected behaviour is a migration to host B but agree that the endstate does not qualify as a success. Do you request that this is implemented as failure result, @borisstoyanov ?

borisstoyanov:
Yes @DaanHoogland, I do believe that it should return a failure if cloudstack is not able to execute the request with all required parameters.

PR #3979, opened 2020-03-18
Limit API from trying to start a VM that is already running

kioie:
## Description

Enhancement for issue #2736

## Types of changes

Screenshot 2020-03-18 at 16 06 54

## How Has This Been Tested?

1. Start a VM
2. With cloudmonkey, start the same VM
3. Observe the result
Response should fail the job and say VM is already in running state.


DaanHoogland:
I don't mind this code but it seems to me that there is no state transition from started to started and hence this should already fail. If it doesn't there is a bug in the statemachine implementation or - design.

GabrielBrascher:
I just tested on the master and it indeed prompts no exception when starting a VM that is already Running; therefore it makes sense to have this PR merged.

@DaanHoogland are you OK with this one?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #3839, opened 2020-01-27
FEATURE-3823: kvm agent hooks

bwsw:
## Description

The PR introduces the new KVM agent extension interface in the form of hooks. Every hook is implemented in Groovy like

```
package groovy

class BaseTransform {
	String transform(Object logger, String xml) {
		File file = new File("/tmp/hooks.txt")
		file.append("------------------------------------------------------------------------\n")
		file.append(xml)
		return xml
	}
}

new BaseTransform()
```

There are 3 hooks are implemented now:
1. `XML transformer` which is called right before VM is launched and it allows modifying XML somehow (example above)

2. `onStart` and `onStop` hooks which are called right after VM state changes to started/stopped.

```
package groovy

class StateHandler {
	Object onStart(Object logger, String vmName) {
		File file = new File("/tmp/hooks-start.txt")
		file.append(vmName)
		return null
	}

        Object onStop(Object logger, String vmName) {
		File file = new File("/tmp/hooks-stop.txt")
		file.append(vmName)
                return null
        }
}

new StateHandler()
```

Every hook is run inside `try {} catch (Exception e) {}`, so it can not cause agent misbehavior. If hooks are not defined, they are skipped.

Also, the PR adds initial support for GitLab CICD for those, who does WIPs with GitLab, not in GitHub.

Initial RFC/Proposal: https://github.com/apache/cloudstack/issues/3823

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

Tested in unit tests and manually for master.

PR #3986, opened 2020-03-23
Change listVpcOfferings endpoint and createVpcOffering endpoint

radu-todirica:
## Description

Currently in cloudstack there is no way to create a VPC Offering with a selected service offering, the service offering is always null. When a VPC router is created if the service offering is null a default Service Offering is chosen.

The following changes have been added:

- listVpcOfferings endpoint will list what system service offering they were created with
- the cloudstack ui will let you pick what serviceOfferingId to use
- the cloudstack ui will let you edit the serviceOffering used by the vpcOffering

## Types of changes

addVpcOffering
editOffering

## How Has This Been Tested?

Wrote a python test for testing create VPC offering with selected service offering
Steps for validation

- Create a system offering
- Create VPC Offering by specifying all supported Services and your system offering
- VPC offering should be created successfully

[results.txt](https://github.com/apache/cloudstack/files/4371015/results.txt)
[runinfo.txt](https://github.com/apache/cloudstack/files/4371016/runinfo.txt)

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/admin/vpc/UpdateVPCOfferingCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/admin/vpc/UpdateVPCOfferingCmd.java
@@ -25,6 +25,7 @@ import org.apache.cloudstack.api.ApiErrorCode;
 import org.apache.cloudstack.api.BaseAsyncCmd;
 import org.apache.cloudstack.api.Parameter;
 import org.apache.cloudstack.api.ServerApiException;
+import org.apache.cloudstack.api.response.ServiceOfferingResponse;
 import org.apache.cloudstack.api.response.VpcOfferingResponse;
 import org.apache.log4j.Logger;
 
@@ -71,6 +72,12 @@ public class UpdateVPCOfferingCmd extends BaseAsyncCmd {
 
     @Parameter(name = ApiConstants.SORT_KEY, type = CommandType.INTEGER, description = "sort key of the VPC offering, integer")
     private Integer sortKey;
+
+    @Parameter(name = ApiConstants.SERVICE_OFFERING_ID,
+            type = CommandType.UUID,
+            entityType = ServiceOfferingResponse.class,
+            description = "the ID of the service offering for the VPC router appliance")
+    private Long serviceOfferingId;
 
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
@@ -154,6 +161,10 @@ public class UpdateVPCOfferingCmd extends BaseAsyncCmd {
 
     public Integer getSortKey() {
         return sortKey;
+    }
+
+    public Long getServiceOfferingId() {
+        return serviceOfferingId;
     }
 
     /////////////////////////////////////////////////////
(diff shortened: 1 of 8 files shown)


weizhouapache:
@radu-stefanache all service/disk/network/vpc offerings are not editable (except name and display text). maybe you can research on other solutions. for example, add "router offering" to vpc as a new field, or update vpc to another offering (similar as update network offering).

DaanHoogland:
@weizhouapache i agree that offerings should not be editable as they can be in use, but what is the harm in adding the offering to the create API?

weizhouapache:
> @weizhouapache i agree that offerings should not be editable as they can be in use, but what is the harm in adding the offering to the create API?

@DaanHoogland  I am ok with adding router offering to createVpcOffering, similar as createNetworkOffering.

However, the network/vpc will be shutdown if we change a network/vpc to another offering. It would be nice to create a new VR before we destroy current running VR, like @rhtyd did when restart Network/vpc.

radu-todirica:
Thanks @weizhouapache and @DaanHoogland for your feedback. I will do the changes.

DaanHoogland:
hope you manage @radu-stefanache . point to keep in mind uis that we can not ever allow altering offerings as they might be in use elsewhere and by other people in other locations in the system and unpredictable things happen if you allow that. 
happy coding

nathanejohnson:
@DaanHoogland so regarding the "no edit to service / disk offerings" , I'm not sure I completely understand your stance on this?  In fact, there is a PR we're considering that would allow users to modify the hypervisor based disk iops / rate limiting parameters?

DaanHoogland:
@nathanejohnson that would impact all running instances, how would you control which get impacted and in fact for which it is meant? such changes should be done on the instances or by selecting a new offering, not by altering an offering of which you don't know by who it is used.
btw this is not some secret principle I adhere, but a long standing design proinciple of cloudstack as i understand it to be. (/me loves discussion)

nathanejohnson:
@DaanHoogland it would impact them once they restarted, yes.  And I can see this being an argument around cores, ram etc.  But for hypervisor based IO limits, we have no current concept of "custom", which is how people get around static cpu and memory limits today.

DaanHoogland:
sorry @nathanejohnson , i'm on staycation so not reacting to promptly.
should the concept of custom than not be implemented, instead of trying to correctly implement hypervisor specific knobs in offerings? how you say it, it seems like it is something that should be read from the host during deploy???

nathanejohnson:
This isn't host specific, in this case "hypervisor specific" implies that for kvm these knobs are part of the domain xml, instead of from some external storage provider.  I do think in a perfect world these would have been implemented in such a way that it would be easy to override on a per-VM basis, such as with "custom", but that's now how it's done today, and it wouldn't be the easiest lift in the world to do it either.

DaanHoogland:
@nathanejohnson is #3839 in line with your requirements? your last description let me to associate to that.

rhtyd:
ping @radu-todirica cc @kiwiflyer is this still relevant? Can you fix the conflict if so. Thanks.

DaanHoogland:
@radu-todirica @weizhouapache @rohityadavcloud @nathanejohnson closing this as it is antiquated and a conflict. Please reopen if it is updated and still relevant.

PR #3999, opened 2020-03-30
Update schema-41310to41400.sql

andrijapanicsb:
Keeping the snapshot on Primary storage only, was originally implemented for KVM+Ceph specifically as it seems, since it's seriously broken for KVM+NFS/local in 4.11+ (and not applicable/doesn't affects VMware at all, while it accidentally works for XS+NFS as they share the same "strategy")

Thus update the desc so that KVM users don't find themselves scratching their head.
We would want to fix/implement this in 4.15 (?) since when using qcow2 the snaps can be kept "attached" to it, the same way they are now attached/part of the rbd image (Ceph)

ping @GabrielBrascher @wido @weizhouapache 

cc @nathanejohnson (you guys implemented this, right?)

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-41310to41400.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41310to41400.sql
@@ -21,7 +21,7 @@
 
 -- Update the description to indicate this only works with KVM + Ceph 
 -- (not implemented properly atm for KVM+NFS/local, and it accidentaly works with XS + NFS. Not applicable for VMware)
-UPDATE `cloud`.`configuration` SET `description`='Indicates whether to always backup primary storage snapshot to secondary storage. Applicable for KVM + Ceph only.' WHERE  `name`='snapshot.backup.to.secondary';
+UPDATE `cloud`.`configuration` SET `description`='Indicates whether to always backup primary storage snapshot to secondary storage. Keeping snapshots only on Primary storage is applicable for KVM + Ceph only.' WHERE  `name`='snapshot.backup.to.secondary';
 
 -- KVM: enable storage data motion on KVM hypervisor_capabilities
 UPDATE `cloud`.`hypervisor_capabilities` SET `storage_motion_supported` = 1 WHERE `hypervisor_capabilities`.`hypervisor_type` = 'KVM';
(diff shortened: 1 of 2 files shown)


engine/schema/src/main/resources/META-INF/db/schema-41310to41400.sql
```diff
@@ -19,6 +19,10 @@
 -- Schema upgrade from 4.13.1.0 to 4.14.0.0
 --;
 
+-- Update the description to indicate this only works with KVM + Ceph 
+-- (not implemented properly atm for KVM+NFS/local, and it accidentaly works with XS + NFS. Not applicable for VMware)
+UPDATE `cloud`.`configuration` SET `description`='Indicates whether to always backup primary storage snapshot to secondary storage. Applicable for KVM + Ceph only.' WHERE  `name`='snapshot.backup.to.secondary';
```
DaanHoogland: nice to update in the db, but it needs updating in code: https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/storage/snapshot/SnapshotManager.java#L59

DaanHoogland (dismissed):
:-1:

DaanHoogland (changes requested):
please see my drunken comment

andrijapanicsb:
I love you comments ❤️  don't worry 😉  
say, what makes you think it's needed? this global setting is already in the DB - so my idea is to edit the DB.
The config key you've shared (AFAIK!?) would create a setting that doesn't exist, right? but since it's already there from 4.11.3, I thought to just update it?

andrijapanicsb:
@blueorangutan package

weizhouapache (approved):
code LGTM

nvazquez (approved):

andrijapanicsb:
image
This was on both clean install and when 4.13.1-snapshot was upgraded to master (this PR code) -so it looks good <--- Works without the config key update 

Will rebuild just for the sake of re-testing (and upgrade my envs)

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
Doesn't count from my side - but this one is LGTM after manual testing

andrijapanicsb:
for the sake of fun, I don't find the tests needed for this change.

@blueorangutan test

DaanHoogland:
this PR looks good, despite travis :(

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
Looking good, 2 x LGTMs, manual testing regression testing OK (failed test nothing to do with the PR).

Merging.

PR #3800, opened 2020-01-10
Revert "Extract systemvm.iso using bsdtar (#3536)"

anuragaw:
This reverts commit 8a55c931e806a64f7f00153d4a58c4463caa7a47.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

PR #3536, opened 2019-08-01
Extract systemvm.iso using bsdtar

cl-k-takahashi:
Signed-off-by: Kai Takahashi <k-takahashi@creationline.com>

## Description

Extract systemvm.iso using bsdtar in injectkeys.sh if available.
With bsdtar, injectkeys.sh doesn't need to mount systemvm.iso and works even on unprivileged containers.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

Checked that a new key is injected to systemvm.iso by running injectkeys.sh manually.

PR #4015, opened 2020-04-06
Inject keys into systemvm.iso using bsdtar (related to #3536)

cl-k-takahashi:
I fixed an issue of #3536 which was reverted on #3800

At first, systemvm_mnt dir is owned by root, this makes bsdtar unable to work.
By removing systemvm_mnt first, bsdtar can work properly.
systemvm_mnt can be removed by user cloud as /var/[lib/]cloudstack/management is owned by cloud.

Signed-off-by: Kai Takahashi <k-takahashi@creationline.com>

## Types of changes


rhtyd:
cc @weizhouapache if we pass the sshpublic key to VMware & XenServer systemvm/VR via the agent do we even need to patch the systemvm.iso like this at all?

weizhouapache:
> cc @weizhouapache if we pass the sshpublic key to VMware & XenServer systemvm/VR via the agent do we even need to patch the systemvm.iso like this at all?

@rhtyd I think we can pass the encoded sshkey in cmdline instead of injecting it to the systemvm.iso. let's discuss later

rohityadavcloud:
Thanks @cl-k-takahashi for the PR, we've moved to a new patching approach that won't use systemvm.iso. Closing this due to the change.

PR #4016, opened 2020-04-06
Fixed private gateway can't be deleted

### LINKED ISSUE #2908: Private gateway can't be deleted
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Management Node
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1
~~~

##### CONFIGURATION

~~~
Advanced Network, KVM bridged
~~~

##### OS / ENVIRONMENT

~~~
CentOS 4.11.1
~~~

##### SUMMARY

Its possible to create VPC offering without services, then add VPC based on this offering, then add Private Gateway for this VPC, then add route for this Private Gateway.
During route creation error is displayed:
screen shot 2018-10-18 at 18 39 45

**But** route still being created. Then if I try to delete route - I can't, because of the same error.
So created route can't be deleted and as a result Private Gateway can't be deleted also and VPC can't be deleted too.

Spaceman1984:
## Description

When the static route service is not available on the VPC and a static route is created, the static route is created in a revoked state.

Currently, the UI doesn't distinguish between active or revoked static routes.

This PR adds the missing state filter to the list routes command and only lists active routes in the UI.
It also ignores revoked routes when the private gateway is being removed but clears out the inactive routes before the gateway is removed.

Fixes #2908 
## Types of changes

## How Has This Been Tested?

This has been tested by creating a VPC with no services, adding a private gateway, adding a static route and deleting the gateway.

### DIFF
--- a/engine/schema/src/main/java/com/cloud/network/vpc/dao/StaticRouteDao.java
+++ b/engine/schema/src/main/java/com/cloud/network/vpc/dao/StaticRouteDao.java
@@ -30,7 +30,7 @@ public interface StaticRouteDao extends GenericDao<StaticRouteVO, Long> {
 
     List<StaticRouteVO> listByVpcId(long vpcId);
 
-    List<StaticRouteVO> listByGatewayId(long vpcId);
+    List<StaticRouteVO> listByGatewayId(long gatewayId);
 
     long countRoutesByGateway(long gatewayId);
 
--- a/engine/schema/src/main/java/com/cloud/network/vpc/dao/StaticRouteDao.java
+++ b/engine/schema/src/main/java/com/cloud/network/vpc/dao/StaticRouteDao.java
@@ -30,6 +30,8 @@ public interface StaticRouteDao extends GenericDao<StaticRouteVO, Long> {
 
     List<StaticRouteVO> listByVpcId(long vpcId);
 
+    List<StaticRouteVO> listByGatewayId(long vpcId);
+
     long countRoutesByGateway(long gatewayId);
 
 }
--- a/ui/scripts/vpc.js
+++ b/ui/scripts/vpc.js
@@ -2652,7 +2652,8 @@
                                                     url: createURL('listStaticRoutes'),
                                                     data: {
                                                         gatewayid: args.context.vpcGateways[0].id,
-                                                        listAll: true
+                                                        listAll: true,
+                                                        state: "Active"
                                                     },
                                                     success: function(json) {
                                                         var items = json.liststaticroutesresponse.staticroute;
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vpc/CreateStaticRouteCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vpc/CreateStaticRouteCmd.java
@@ -109,7 +109,7 @@ public class CreateStaticRouteCmd extends BaseAsyncCreateCmd {
             routeResponse.setResponseName(getCommandName());
         } finally {
             if (!success || route == null) {
-                _vpcService.revokeStaticRoute(getEntityId());
+                _entityMgr.remove(StaticRoute.class, getEntityId());
                 throw new ServerApiException(ApiErrorCode.INTERNAL_ERROR, "Failed to create static route");
             }
         }
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vpc/ListStaticRoutesCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vpc/ListStaticRoutesCmd.java
@@ -48,6 +48,9 @@ public class ListStaticRoutesCmd extends BaseListTaggedResourcesCmd {
     @Parameter(name = ApiConstants.GATEWAY_ID, type = CommandType.UUID, entityType = PrivateGatewayResponse.class, description = "list static routes by gateway id")
     private Long gatewayId;
 
+    @Parameter(name = ApiConstants.STATE, type = CommandType.STRING, description = "list static routes by state")
+    private String state;
+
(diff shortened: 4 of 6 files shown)


Spaceman1984:
@blueorangutan package

engine/schema/src/main/java/com/cloud/network/vpc/dao/StaticRouteDao.java
```diff
@@ -30,6 +30,8 @@
 
     List<StaticRouteVO> listByVpcId(long vpcId);
 
+    List<StaticRouteVO> listByGatewayId(long vpcId);
```
weizhouapache, line 33: gatewayId

Spaceman1984: Renamed

Spaceman1984:
> In the case as you said "When the static route service is not available on the VPC when a static route is created, the static route is created in a revoked state", could you add some changes to avoid it ?

@weizhouapache The state of the route is changed to revoked after failing validation. The other option here would be to delete the record at this point.

DaanHoogland:
@blueorangutan package

weizhouapache:
> > In the case as you said "When the static route service is not available on the VPC when a static route is created, the static route is created in a revoked state", could you add some changes to avoid it ?
> 
> @weizhouapache The state of the route is changed to revoked after failing validation. The other option here would be to delete the record at this point, which I think was the intention, but the delete method is also failing validation. I thought to leave things this way, to keep some history until the gateway is removed. I'm happy to force delete the record after failing validation if you believe it's a better way to go?

@Spaceman1984 I think it would be better to remove the record if there is error/exception in the creation.

Spaceman1984:
@blueorangutan package

Spaceman1984:
> > > In the case as you said "When the static route service is not available on the VPC when a static route is created, the static route is created in a revoked state", could you add some changes to avoid it ?
> > 
> > 
> > @weizhouapache The state of the route is changed to revoked after failing validation. The other option here would be to delete the record at this point, which I think was the intention, but the delete method is also failing validation. I thought to leave things this way, to keep some history until the gateway is removed. I'm happy to force delete the record after failing validation if you believe it's a better way to go?
> 
> @Spaceman1984 I think it would be better to remove the record if there is error/exception in the creation.

Done

rhtyd:
@Spaceman1984 this looks like deletes pg in CloudStack DB, have you verified if this would also cleanup rules on the VR properly?

Spaceman1984:
> @Spaceman1984 this looks like deletes pg in CloudStack DB, have you verified if this would also cleanup rules on the VR properly?

No, I'll check.

Spaceman1984:
> @Spaceman1984 this looks like deletes pg in CloudStack DB, have you verified if this would also cleanup rules on the VR properly?


@rhtyd I have checked iptables on the VR before and after trying to add a static route, there is no difference, also this code removes the static route, not the private gateway. Having broken static routes referencing the gateway prevented it from being deleted.

rhtyd:
@Spaceman1984 private gateway would mean a nic, IP and route on the VR, not iptables rules

rhtyd:
ping @Spaceman1984 no update?

Spaceman1984:
Sorry @rhtyd, I can't give this issue much attention right now, I'm currently busy with a different project.

rhtyd:
@Spaceman1984 please check comments ^^ thanks

Spaceman1984:
@blueorangutan test

rhtyd:
@Spaceman1984 ping, any update on this?

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

Spaceman1984:
@rhtyd I have tested creating and deleting a private gateway, the nic, ip and route is created and removed.

DaanHoogland:
@weizhouapache @rhtyd further doubts/questions or are we good to merge?

borisstoyanov:
@blueorangutan test

rhtyd:
@DaanHoogland @Spaceman1984 - my concern is that whether removing privategateway also remove the nic/configuration from the VPC VR or not? Based on code, I think it will solve the issue or removing it from the DB. @Spaceman1984 did you test the case by adding privategateway and check if VPC VR has a new nic and then removing it removes it both from VR and DB?

Spaceman1984:
> @DaanHoogland @Spaceman1984 - my concern is that whether removing privategateway also remove the nic/configuration from the VPC VR or not? Based on code, I think it will solve the issue or removing it from the DB. @Spaceman1984 did you test the case by adding privategateway with some static routes (also deploy VMs as VR programming is lazy) and check if VPC VR has a new nic and then removing it removes it both from VR and DB?

I'll do some more testing

Spaceman1984:
> @DaanHoogland @Spaceman1984 - my concern is that whether removing privategateway also remove the nic/configuration from the VPC VR or not? Based on code, I think it will solve the issue or removing it from the DB. @Spaceman1984 did you test the case by adding privategateway with some static routes (also deploy VMs as VR programming is lazy) and check if VPC VR has a new nic and then removing it removes it both from VR and DB?

@rhtyd I have confirmed that it works as expected.

ui/scripts/vpc.js
```diff
@@ -2652,7 +2652,8 @@
                                                     url: createURL('listStaticRoutes'),
                                                     data: {
                                                         gatewayid: args.context.vpcGateways[0].id,
-                                                        listAll: true
+                                                        listAll: true,
+                                                        state: "Active"
```
rhtyd, line 2656: @Spaceman1984 can you send the related change to Primate as well cc @davidjumani

PR #4029, opened 2020-04-14
Bring back vm.suspend during deleting VM snapshot

andrijapanicsb:
Original commit, merged to 4.11.3, got mysteriously missing when merging branch 4.11 to 4.12 on the very same day that original PR was merged - so bringing back the same "suspend the VM before the KVM VM snapshot is deleted" code (we already have "resume if it's suspended, after VM snap deletion is finished"

I'm talking about the PR #3194 (the commit itself is there, but NOT the content/code - possibly a wrongly resolved merge conflicts etc)

/cc @weizhouapache  (you approved the original PR) and @rhtyd (you also approved the original PR)

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
@@ -1006,9 +1006,10 @@ public class KVMStorageProcessor implements StorageProcessor {
                     if (state == DomainInfo.DomainState.VIR_DOMAIN_RUNNING && !primaryStorage.isExternalSnapshot()) {
                         final DomainSnapshot snap = vm.snapshotLookupByName(snapshotName);
                         try {
+                            s_logger.info(String.format("Suspending VM '%s' to delete snapshot,", vm.getName()));
                             vm.suspend();
                         } catch (final LibvirtException e) {
-                            s_logger.debug("Failed to suspend the VM: " + e);
+                            s_logger.error("Failed to suspend the VM", e);
                             throw e;
                         }
                         snap.delete(0);
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
@@ -1010,7 +1010,7 @@ public class KVMStorageProcessor implements StorageProcessor {
                         } catch (final LibvirtException e) {
                             s_logger.debug("Failed to suspend the VM: " + e);
                             throw e;
-                        }                      
+                        }
                         snap.delete(0);
 
                         /*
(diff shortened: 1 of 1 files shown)


andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan package

Slair1:
@andrijapanicsb thanks for opening this PR up.  From your testing were you able to reproduce the corruption using a "VM Snapshot" also (create followed by a delete)?  Or just a volume-based snapshot?

I ask because the delete of a VM snapshot still does not have the pause/resume.

andrijapanicsb:
I'm testing with the latest master first, trying to corrupt the VM (delete VM snap during high IO load inside a VM) - but can't manage to do that (hm, my cent7 template is EXT4, not the gentle XFS - perhaps that's the case) - tired of testing...

Anyway, it's safe to suspend the VM before the snap restore, so we shall proceed with it - so I will verify the VM is suspended (this PR) and move from there.

andrijapanicsb:
@Slair1 I don't see VM being paused when snapshots (VM snapshot) is being deleted...

Sorry to ask, did you ever observe/confirm code works on 4.11.3?
i.e. "virsh list" should show VM as "paused" for a few moments, right? (like it is while the snap is created in the first place) ?

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@weizhouapache this PR is no good.
The snap delete seems to NOT happen here, as I don't see the log line "Suspending..." while the VM is being suspended at all (added to help debug why the vm.suspend is not run/doesn't get suspended)

I'm moving this PR to 4.15, unless you miraculously come (wouldn't be the first time :) with proper code and add that vm.suspend whereever the snap IS actually deleted...)
thx

Slair1:
@andrijapanicsb you're correct, we also need the suspend and resume when a VM Snapshot is deleted also - that is not currently done.  This PR is partially correct though as is.   Snapshots are also deleted after a Backup to Secondary Storage operation is finished.  @ggoodrich-ipp perhaps you can push your PR and reference the new PR here?

@ggoodrich-ipp has a more complete PR with the pause/resume in both applicable locations.

andrijapanicsb:
Guys, we are touching here, as I understood – VM snapshots?
And keep in mind that we need to craft 4.14 RC1 on Friday – not sure we have any time for new stuff, unless a VERY quick fix.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
```diff
@@ -1005,6 +1005,13 @@ public Answer backupSnapshot(final CopyCommand cmd) {
                             primaryStore.getUuid());
                     if (state == DomainInfo.DomainState.VIR_DOMAIN_RUNNING && !primaryStorage.isExternalSnapshot()) {
                         final DomainSnapshot snap = vm.snapshotLookupByName(snapshotName);
+                        try {
+                            s_logger.info(String.format("Suspending VM '%s' to delete snapshot,", vm.getName()));
+                            vm.suspend();
+                        } catch (final LibvirtException e) {
+                            s_logger.error("Failed to suspend the VM", e);
+                            throw e;
```
DaanHoogland, line 1013: same comment as in #4033: first we suspend unconditionaly, and than, we just assume it need resuming without checking if it was running in the first place. Are there situations we should guard against more diligently? (cc @weizhouapache @rhtyd )

rhtyd: @DaanHoogland I see on line 1007, we are checking that domain is running? Did I miss anything?

DaanHoogland: yes, we are, i am confusing PRs. this one actually doesn't do anything in the delete vmshapshot scenario.

PR #4030, opened 2020-04-15
some minor changes to make the build run with python 3

DaanHoogland:
## Description

print as method
async keyword use in variable name fixed
urllib openurl from urllib2 to urllib.request

This PR covers partially issue #3195.

## Types of changes

### DIFF
--- a/tools/apidoc/gen_toc.py
+++ b/tools/apidoc/gen_toc.py
@@ -245,10 +245,10 @@ for f in sys.argv:
 
 def xml_for(command):
     name = command['name']
-    async = command['async'] and ' (A)' or ''
+    asyncmethod = command['async'] and ' (A)' or ''
     dirname = command['dirname']
     return '''<xsl:if test="name=\'%(name)s\'">
-<li><a href="%(dirname)s/%(name)s.html"><xsl:value-of select="name"/>%(async)s</a></li>
+<li><a href="%(dirname)s/%(name)s.html"><xsl:value-of select="name"/>%(asyncmethod)s</a></li>
 </xsl:if>
 ''' % locals()
 
--- a/tools/marvin/marvin/codegenerator.py
+++ b/tools/marvin/marvin/codegenerator.py
@@ -21,7 +21,7 @@ from optparse import OptionParser
 from textwrap import dedent
 import os
 import sys
-import urllib2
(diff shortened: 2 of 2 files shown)


nvazquez:
@DaanHoogland in my tests compile works fine, but I'm having this problem with the deploy script:

````
$  python tools/marvin/marvin/deployDataCenter.py -i ../adv-kvm.cfg
  File "tools/marvin/marvin/deployDataCenter.py", line 77
    print "\n=== Data Center Settings are dumped to %s===" % dc_file_path
                                                         ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print("\n=== Data Center Settings are dumped to %s===" % dc_file_path)?
````

**Comment by nvazquez** [comment]
This will need refactor in many places, maybe considering this library? https://docs.python.org/2/library/2to3.html

**Comment by DaanHoogland** [comment]
yes @nvazquez this only addresses the build. There is another PR out for python3 changes (#3730 by @GabrielBrascher) for the system vm. I don't want to merge, as this is kind of independent. I did hit that deploy script as well, we can put that in a separate PR. or here as well, i just want to move this quickly and smaller chunks move faster ;)

File: tools/apidoc/gen_toc.py [hunk 1]
```diff
@@ -245,10 +245,10 @@ def choose_category(fn):
 
 def xml_for(command):
     name = command['name']
-    async = command['async'] and ' (A)' or ''
+    asyncmethod = command['async'] and ' (A)' or ''
```
GabrielBrascher, line 248: @DaanHoogland shouldn't the header of this python file be changed from `#!/cygdrive/c/Python27` to `#!/cygdrive/c/Python3`?

GabrielBrascher: Sorry if this comment is redundant, I see that @nvazquez mentioned about changing other pieces as well.

DaanHoogland: I think the cygwin specific python is not a good idea in any case `#! /bin/env python` would be better.
I think this is being taken into account in ther PRs so maybe i should close, @nvazquez @GabrielBrascher ?

DaanHoogland:
The changes here are in #4479 as well, will probably close this one.

PR #4032, opened 2020-04-15
Suspending the VM prior to deleting snapshots to avoid corruption, th…

ggoodrich-ipp:
…en resuming

## Description

These changes are related to PR #3194, but include suspending/resuming the VM when doing a VM snapshot as well, when deleting a VM snapshot, as it is performing the same operations via Libvirt. Also, there was an issue with the UI/localization changes in the prior PR, as that PR was altering the Volume snapshot behavior, but was altering the VM snapshot wording. Both have been altered in this PR.

Issuing this in response to the work happening in PR #4029.

## Types of changes


andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@weizhouapache does this one makes more sense/checks than the #4033  - i.e. shall we work on this one?

DaanHoogland:
@ggoodrich-ipp my suggestion there ^^ won't catch all possible conflicts, and won't give us a full mutex, but for the case we are handling now it should do. (cc @weizhouapache @rhtyd again)

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
```diff
@@ -51,14 +52,28 @@ public Answer execute(final DeleteVMSnapshotCommand cmd, final LibvirtComputingR
...
         try {
             final LibvirtUtilitiesHelper libvirtUtilitiesHelper = libvirtComputingResource.getLibvirtUtilitiesHelper();
-            Connect conn = libvirtUtilitiesHelper.getConnection();
+            conn = libvirtUtilitiesHelper.getConnection();
             dm = libvirtComputingResource.getDomain(conn, vmName);
 
             snapshot = dm.snapshotLookupByName(cmd.getTarget().getSnapshotName());
 
+            // Suspend the VM prior to deleting the snapshot to guard against corruption
+            dm.suspend();
```
rhtyd, line 66: Should we check if VM is running? (i.e. not in any other states)

rhtyd: May wrap in a try-with?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
```diff
@@ -93,12 +108,26 @@ public Answer execute(final DeleteVMSnapshotCommand cmd, final LibvirtComputingR
...
+                return new DeleteVMSnapshotAnswer(cmd, cmd.getVolumeTOs());
             }
 
             s_logger.warn(msg, e);
             return new DeleteVMSnapshotAnswer(cmd, false, msg);
         } finally {
             if (dm != null) {
+                // Make sure if the VM is paused, then resume it, in case we got an exception during our delete() and didn't have the chance before
+                try {
+                    dm = libvirtComputingResource.getDomain(conn, vmName);
```
rhtyd, line 122: dm is already not null, why overwrite it again?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
DaanHoogland: does this mean you agree with my statement above, @rhtyd ?
the try catch won't work, it will succeed leaving a corrupt disk image (according to @andrijapanicsb 's tests)

DaanHoogland:
as we are in release phase already and #4033 is simpler let's go with that. the logic is essentially the same.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
ggoodrich-ipp: I was under the impression that dm needed to be refreshed after a state change, as I saw this same thing being done in KVMStorageProcessor -> backupSnapshot(). Rather than validate that it was necessary or not, I decided that safer is always better, as refreshing the domain variable won't cause harm.

rhtyd:
@ggoodrich-ipp can you fix the conflicts, thanks.

ggoodrich-ipp:
> @ggoodrich-ipp can you fix the conflicts, thanks.

Conflicts resolved.

rhtyd:
@blueorangutan package

rhtyd:
@ggoodrich-ipp could you check if `quiescevm` is true in the `createSnapshot` API? Would that be better than doing it by default?

ggoodrich-ipp:
> @ggoodrich-ipp could you check if `quiescevm` is true in the `createSnapshot` API? Would that be better than doing it by default?

That seems questionable, due to the potential for corruption, in this case. I wouldn't think we would want an option that could lead to corruption.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
```diff
@@ -97,12 +108,26 @@ public Answer execute(final DeleteVMSnapshotCommand cmd, final LibvirtComputingR
...
+                    state = dm.getInfo().state;
+                    if (state == DomainInfo.DomainState.VIR_DOMAIN_PAUSED) {
+                        dm.resume();
+                    }
+                } catch (LibvirtException e) {
+                    s_logger.error("Failed to resume vm after delete snapshot " + cmd.getTarget().getSnapshotName() + " on vm: " + vmName + " return true : " + e);
+                }
+
```
DaanHoogland, line 130: can you extract this in a separate method please?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
```diff
@@ -63,6 +66,14 @@ public Answer execute(final DeleteVMSnapshotCommand cmd, final LibvirtComputingR
...
 
             snapshot.delete(0); // only remove this snapshot, not children
+            didDelete = true;
+
+            // Resume the VM
+            dm = libvirtComputingResource.getDomain(conn, vmName);
+            state = dm.getInfo().state;
+            if (state == DomainInfo.DomainState.VIR_DOMAIN_PAUSED) {
+                dm.resume();
+            }
```
DaanHoogland, line 76: resume action leads to an implicit assumption which must be guarded during maintenance by future generations: failing resume is concluded by examining a var named `didDelete`. dangerous assumption even with the current logic computing. see below.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
```diff
@@ -97,12 +108,26 @@ public Answer execute(final DeleteVMSnapshotCommand cmd, final LibvirtComputingR
             } else if (snapshot == null) {
                 s_logger.debug("Can not find vm snapshot " + cmd.getTarget().getSnapshotName() + " on vm: " + vmName + ", return true");
                 return new DeleteVMSnapshotAnswer(cmd, cmd.getVolumeTOs());
+            } else if (didDelete) {
+                s_logger.error("Failed to resume vm after delete snapshot " + cmd.getTarget().getSnapshotName() + " on vm: " + vmName + " return true : " + e);
+                return new DeleteVMSnapshotAnswer(cmd, cmd.getVolumeTOs());
```
DaanHoogland, line 113: here a var `didDelete` leads us to conclude that the current exception thrown is caused by `dm.resume()`. this boolean should be named some thing like `tryingResume` instead. better would be to investigate the nature of the exception.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStorageProcessor.java
```diff
@@ -1029,7 +1029,7 @@ public Answer backupSnapshot(final CopyCommand cmd) {
                         }
                     }
                 } catch (final Exception ex) {
-                    s_logger.debug("Failed to delete snapshots on primary", ex);
+                    s_logger.error("Failed to delete snapshots on primary", ex);
```
DaanHoogland, line 1032: :+1:

ggoodrich-ipp:
@DaanHoogland @rhtyd I have made some adjustments to this PR. Thoughts?

shwstppr:
@ggoodrich-ipp can you please resolve conflicts

weizhouapache:
@ggoodrich-ipp what issue does this pr solve ?
Delete a vm snapshot twice ?

ggoodrich-ipp:
> @ggoodrich-ipp what issue does this pr solve ?
> Delete a vm snapshot twice ?

Well, I think another PR managed to get in there to suspend the VM, but this PR is specifically around suspending, then resuming the VM when doing a snapshot.

ggoodrich-ipp:
I see @DaanHoogland re-targeted this to 4.15, which must have resolved the conflicts. Are there any other issues with this PR that need to be addressed prior to approval, @weizhouapache @shwstppr ? Let me know, and I'll try to accommodate. Thanks!

DaanHoogland:
travis is reporting a checkstyle error but i cannot find it closing

DaanHoogland:
re-opening to re-kick travis

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
```diff
@@ -52,18 +52,33 @@ public Answer execute(final DeleteVMSnapshotCommand cmd, final LibvirtComputingR
         final KVMStoragePoolManager storagePoolMgr = libvirtComputingResource.getStoragePoolMgr();
         Domain dm = null;
         DomainSnapshot snapshot = null;
+        DomainInfo.DomainState oldState = null;
```
weizhouapache, line 55: @ggoodrich-ipp 
should it be "DomainState" instead of "DomainInfo.DomainState" ?

weizhouapache (approved):

weizhouapache:
@ggoodrich-ipp could you please fix the build error ?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test

shwstppr (approved):
LGTM on code changes and test results. May need manual testing

rhtyd:
@weizhouapache have you done any manual testing on this?

rhtyd:
Did another round of code review, LGTM

PR #4033, opened 2020-04-15
kvm: suspend/resume in deleting vm snapshot on kvm

ustcweizhou:
## Description

To void qcow2 image corruption, we'd better suspend vm when delete a vmsnapshot, and resume it when vmsnapshot is removed.
Fixes: #3193 
related to: #3194 #4029 

## Types of changes

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
@@ -24,6 +24,7 @@ import org.apache.cloudstack.storage.to.VolumeObjectTO;
 import org.apache.log4j.Logger;
 import org.libvirt.Connect;
 import org.libvirt.Domain;
+import org.libvirt.DomainInfo.DomainState;
 import org.libvirt.DomainSnapshot;
 import org.libvirt.LibvirtException;
 
@@ -57,6 +58,9 @@ public final class LibvirtDeleteVMSnapshotCommandWrapper extends CommandWrapper<
             dm = libvirtComputingResource.getDomain(conn, vmName);
 
             snapshot = dm.snapshotLookupByName(cmd.getTarget().getSnapshotName());
+
+            s_logger.debug("Suspending domain " + vmName);
+            dm.suspend(); // suspend the vm to avoid image corruption
 
             snapshot.delete(0); // only remove this snapshot, not children
 
@@ -100,6 +104,10 @@ public final class LibvirtDeleteVMSnapshotCommandWrapper extends CommandWrapper<
         } finally {
             if (dm != null) {
                 try {
+                    if (dm.getInfo().state == DomainState.VIR_DOMAIN_PAUSED) {
+                        s_logger.debug("Resuming domain " + vmName);
+                        dm.resume();
+                    }
                     dm.free();
                 } catch (LibvirtException l) {
                     s_logger.trace("Ignoring libvirt error.", l);


andrijapanicsb:
@weizhouapache being Wei Zhou, as usual... :) thanks!

Shall we try to squeeze this in master only (4.13 is already out for voting) - or shall we craft another 4.13 RC2 due to this? (I don't see a problem there) - this seems to be a serious enough issue/fix to warrant RC2? (and simple enough change to NOT ask for a serious full-blown retesting of everything) ?

/cc @DaanHoogland @rhtyd

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
Good boy., ape....

luhaijiao:
Would expect to have it in 4.13 RC2 given it's potential but serious impact.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
```diff
@@ -58,6 +59,9 @@ public Answer execute(final DeleteVMSnapshotCommand cmd, final LibvirtComputingR
 
             snapshot = dm.snapshotLookupByName(cmd.getTarget().getSnapshotName());
 
+            s_logger.debug("Suspending domain " + vmName);
+            dm.suspend(); // suspend the vm to avoid image corruption
```
DaanHoogland, line 63: should we do this conditionally and remember whether we did, to then only start it up again in the finally clause if it wasn't suspended in the first place?

andrijapanicsb:
Not sure ai understand your question @DaanHoogland?
In general, we should first check if VM is running, and then suspend/delete/resume.
Not sure this would work for a stopped VM? (about to test the PR now)

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
rhtyd: @weizhouapache can you wrap this in a state check?

rhtyd:
@andrijapanicsb @DaanHoogland this also LGTM, smaller than the other two. The code to suspend is okay, but a check may be preferable.

@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtDeleteVMSnapshotCommandWrapper.java
andrijapanicsb: Will have to leave this for another time - otherwise, the "manually suspended" VM does indeed get's unsuspended - just managed to test it. WHY would someone in the first place even touch the VM manually, then expect ACS to handle it all?

andrijapanicsb:
2 x LGTMs/Approvals, manual testing done fine, regression tests are fine.

Ready for merge @DaanHoogland

andrijapanicsb:
@weizhouapache ... should we also suspend the VM while we delete **VM snapshot** right after Backuping up **volume snap** to Secondary Storage, below:

https://github.com/ustcweizhou/cloudstack/blob/6f30fdfed2986a52c78fb095b2cdc917f1e8f216/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtBackupSnapshotCommandWrapper.java#L175

This PR code 4033 is executed when we delete a VM snapshot that we manually created previously : (i.e. proper resume during creation and deletion of the VM snap) - all good here.

But when kvm.snapshots.enabled=true, and you snap **just the volume** of the running VM - I could see that the whole VM is being paused, whole VM snapshot was taken (all volumes and RAM) and then VM is being resumed and then a single volume snapshots (from the qcow2) is being exported via qemu-img to Secondary STorage and then the original VM snapshot is deleted - during this VM SNAP deletion I could not confirm the VM is being ALWAYS paused/suspended (no logs messages ever, but I could see it being paused when DATA volume snap is deleted, most of the time, but not paused when ROOT volumes are snapped)- i.e. I could see that the code of this PR is NOT executed, but some other code (what I shared above I guess?) - since I can't see for example that this logging is NOT done:    s_logger.debug("Suspending domain " + vmName); (from this PR)

The only thing I would like to see verified is that the VM is always suspended (like in this PR) when we delete VOLUME SNAPSHOT, due to the whole VM SNAPSHOTS is taken when we do just a volume snapshots (running VM, with kvm.snapshots.enable=true)

Could you please advise?

PR #4047, opened 2020-04-28
Look for active templates for VR deployment

ravening:
## Description
If the template from which VR is created got deleted, the state
is set to inactive and removed to null.
Since the template is already deleted, the VR can't be created
using this template again.

If someone restarts network with cleanup then it will try to
deploy the vr from the old non existing template again.
So search only for active template which are not yet deleted.

## Types of changes

## How Has This Been Tested?
1. Create a network and this will in turn create VR's
2. Note down the template from which the VR was created
3. Delete that template
4. Now restart the network with cleanup option
5. The new router deployment should be successful using another active template from the platform

### DIFF
--- a/engine/schema/src/main/java/com/cloud/storage/dao/VMTemplateDaoImpl.java
+++ b/engine/schema/src/main/java/com/cloud/storage/dao/VMTemplateDaoImpl.java
@@ -363,6 +363,7 @@ public class VMTemplateDaoImpl extends GenericDaoBase<VMTemplateVO, Long> implem
         tmpltTypeHyperSearch2.and("templateType", tmpltTypeHyperSearch2.entity().getTemplateType(), SearchCriteria.Op.EQ);
         tmpltTypeHyperSearch2.and("hypervisorType", tmpltTypeHyperSearch2.entity().getHypervisorType(), SearchCriteria.Op.EQ);
         tmpltTypeHyperSearch2.and("templateName", tmpltTypeHyperSearch2.entity().getName(), SearchCriteria.Op.EQ);
+        tmpltTypeHyperSearch2.and("state", tmpltTypeHyperSearch2.entity().getState(), SearchCriteria.Op.EQ);
 
         tmpltTypeSearch = createSearchBuilder();
         tmpltTypeSearch.and("state", tmpltTypeSearch.entity().getState(), SearchCriteria.Op.EQ);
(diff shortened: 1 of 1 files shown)


svenvogel:
Hi @ravening small question. 🙏 how do you delete the template? api or ui?

ravening:
> Hi @ravening small question. 🙏 how do you delete the template? api or ui?

Hello @svenvogel , its mostly through UI but can be done through api as well I guess but I have not tried using api

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland:
as per your description @ravening, "If the template from which VR is created got deleted, the state
is set to inactive and removed to null.", it seems the 'real' bug is that removed is set to null whilst it should be set to the time of deletion. Your code will work, but is it not hiding this 'real' problem? Being very defensive, we might want to address both issues.

ravening:
> as per your description @ravening, "If the template from which VR is created got deleted, the state
> is set to inactive and removed to null.", it seems the 'real' bug is that removed is set to null whilst it should be set to the time of deletion. Your code will work, but is it not hiding this 'real' problem? Being very defensive, we might want to address both issues.

I will see if I can reproduce the issue to check which part of the code is setting it to null instead of removed timestamp

DaanHoogland:
@nvazquez do you have an opinion as well on this?

nvazquez:
I think the fix makes sense as it is now. I don't know the reason why we set the template to Inactive but not removed instead of removing it (timestamp on the removed column) as we normally do with other entities, probably it is not a bug but the expected behaviour by design

nvazquez (approved):
Code LGTM, have not tested it

DaanHoogland:
ok tnx @nvazquez . As said I do have my doubt as I think we might be hiding the issue, but will not 👎 .

GabrielBrascher (approved):
Code LGTM

shwstppr:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test

rhtyd (approved):

shwstppr:
@blueorangutan test

shwstppr (approved):
lgtm

PR #4051, opened 2020-04-29
db: add support for MySQL 8

rhtyd:
- Splits commands to create user and grant access on database, the old
statement is no longer supported by MySQL 8.x
- `NO_AUTO_CREATE_USER` is no longer supported by MySQL 8.x so remove
that from db.properties conn parameters

For mysql-server 8.x setup the following changes were added/tested to
make it work with CloudStack in /etc/mysql/mysql.conf.d/mysqld.cnf and
then restart the mysql-server process:

    server_id = 1
    sql-mode="STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION,ERROR_FOR_DIVISION_BY_ZERO,NO_ZERO_DATE,NO_ZERO_IN_DATE,NO_ENGINE_SUBSTITUTION"
    innodb_rollback_on_timeout=1
    innodb_lock_wait_timeout=600
    max_connections=1000
    log-bin=mysql-bin
    binlog-format = 'ROW'

    default-authentication-plugin=mysql_native_password

Notice the last line above, this is to reset the old password-based
authentication used by MySQL 5.x.

Developers can set empty password as follows:

    > sudo mysql -u root
    ALTER USER 'root'@'localhost' IDENTIFIED BY '';
mysql8-deploydb

Tested on Ubuntu 20.04 for a local Primate setup.

### DIFF
--- a/utils/conf/db.properties
+++ b/utils/conf/db.properties
@@ -44,8 +44,7 @@ db.cloud.testWhileIdle=true
 db.cloud.timeBetweenEvictionRunsMillis=40000
 db.cloud.minEvictableIdleTimeMillis=240000
 db.cloud.poolPreparedStatements=false
-db.cloud.url.params=prepStmtCacheSize=517&cachePrepStmts=true&prepStmtCacheSqlLimit=4096&sessionVariables=sql_mode='STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION'
-
+db.cloud.url.params=prepStmtCacheSize=517&cachePrepStmts=true&prepStmtCacheSqlLimit=4096&sessionVariables=sql_mode='STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION,ERROR_FOR_DIVISION_BY_ZERO,NO_ZERO_DATE,NO_ZERO_IN_DATE,NO_ENGINE_SUBSTITUTION'
 # usage database settings
 db.usage.username=cloud
 db.usage.password=cloud
--- a/utils/conf/db.properties
+++ b/utils/conf/db.properties
@@ -44,7 +44,8 @@ db.cloud.testWhileIdle=true
 db.cloud.timeBetweenEvictionRunsMillis=40000
 db.cloud.minEvictableIdleTimeMillis=240000
 db.cloud.poolPreparedStatements=false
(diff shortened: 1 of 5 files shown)


rhtyd:
@blueorangutan package

rhtyd:
Smoke tests or some test cases:
- MySQL 5.7 + CloudStack against this PR
- MySQL 8.x + CloudStack against this PR
- Travis/developers to test mvn deploydb on their MySQL 5.7 env
- Travis/developers to test mvn deploydb on their MySQL 8.x env
- Upgrade from MySQL 5.7 to 8.x; against CloudStack (optional)

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

utils/conf/db.properties
```diff
@@ -44,8 +44,7 @@ db.cloud.testWhileIdle=true
 db.cloud.timeBetweenEvictionRunsMillis=40000
 db.cloud.minEvictableIdleTimeMillis=240000
 db.cloud.poolPreparedStatements=false
-db.cloud.url.params=prepStmtCacheSize=517&cachePrepStmts=true&prepStmtCacheSqlLimit=4096&sessionVariables=sql_mode='STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION'
-
+db.cloud.url.params=prepStmtCacheSize=517&cachePrepStmts=true&prepStmtCacheSqlLimit=4096&sessionVariables=sql_mode='STRICT_TRANS_TABLES,NO_ENGINE_SUBSTITUTION,ERROR_FOR_DIVISION_BY_ZERO,NO_ZERO_DATE,NO_ZERO_IN_DATE,NO_ENGINE_SUBSTITUTION'
```
weizhouapache, line 47: two NO_ENGINE_SUBSTITUTION

rhtyd: Good catch I'll fix that

rhtyd:
@blueorangutan test

DaanHoogland:
> Smoke tests or some test cases that should be validated:
> 
>     * [ ]  MySQL 5.7 + CloudStack against this PR
> 
>     * [ ]  MySQL 8.x + CloudStack against this PR with (a) single mgmt/usage server and (b) multiple mgmt servers and usage servers
> 

You are trying to keep this compatible with mysql 5 and 8, @rhtyd ? Does it not make more sense to actively seek support mariadb ?
look good otherwise.

rhtyd:
@DaanHoogland yes we should support both versions until 5.x reaches EOL.
MariaDB is largely compatible with MySQL 5.x at least per their docs. 
We need to check if trillian default test on CentOS7 uses mariadb. 
From an applications point of view, the mysql-java library and wire protocol stays the same as well as support for (ansi) SQL.

rhtyd:
cc @shwstppr @davidjumani - while deploying database, and a zone against mysql 8.0 (on Ubuntu 20.04 on raspberrypi4) I found an issue with the `nics` table where `update_time` was being set as null which it didn't like (the default column was not null), I could alter/modify the default param; but this needs checking against mysql8/CentOS8.

rhtyd:
Closing in favour of https://github.com/apache/cloudstack/pull/4068/commits/b895cf98f3927b8fbae7c927367e2c1d3bb68e0b

PR #4065, opened 2020-05-09
Enable revocation checking for uploaded certificates

artem-smotrakov:
## Description

This update turns on certificate revocation checking for uploaded certificates:

- Updated `CertServiceImpl` to be able to enable revocation checking.
- Introduced a new parameter `ENABLED_REVOCATION_CHECK` for `UploadSslCertCmd`.
- Updated `CertServiceTest`.

Even if no CLRs are specified via `PKIXParameters`, the certificates
themselves may still provide info for revocation checking:

- The AIA extension may contains a URL to the OCSP responder.
- The CLRDP extension contains a URL to the CLR.

Those extensions may need to be explicitly enabled by setting the system properties `com.sun.security.enableAIAcaIssuers` and `com.sun.security.enableCRLDP` to true. See [Java PKI Programmer's Guide](https://docs.oracle.com/en/java/javase/11/security/java-pki-programmers-guide.html).

Using a revoked certificate may be dangerous. One of the most common reasons why a certificate authority (CA) revokes a certificate is that the private key has been compromised. For example, the private key might have been stolen by an adversary.

If I understand correctly, the `CertServiceImpl` bean is used for operations with certificates on a load balancer. In particular, it validates a certificate chain without revocation checking while uploading a certificate. If a compromised revoked certificate is then used by the load balancer, then it may result to compromising TLS connections. However, the attacker has to be able to implement man-in-the-middle attack to compromise the connections. So the attacker has to be quite powerful. Therefore, such an attack is definitely not easy to implement. On the other hand, the impact may be significant because of loss of confidentiality.

This has been discussed on security@cloudstack.apache.org

## Types of changes

## How Has This Been Tested?

Updated one test case and added a new one in `CertServiceTest`.

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/loadbalancer/UploadSslCertCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/loadbalancer/UploadSslCertCmd.java
@@ -76,6 +76,9 @@ public class UploadSslCertCmd extends BaseCmd {
     @Parameter(name = ApiConstants.NAME , type = CommandType.STRING, required = true, description = "Name for the uploaded certificate")
     private String name;
 
+    @Parameter(name = ApiConstants.ENABLED_REVOCATION_CHECK, type = CommandType.BOOLEAN, description = "Enables revocation checking for certificates", since = "4.15")
+    private Boolean enabledRevocationCheck = Boolean.TRUE;
+
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
     /////////////////////////////////////////////////////
@@ -110,6 +113,10 @@ public class UploadSslCertCmd extends BaseCmd {
 
     public String getName() {
         return name;
+    }
+
+    public Boolean getEnabledRevocationCheck() {
+        return enabledRevocationCheck;
     }
 
     /////////////////////////////////////////////////////
--- a/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
+++ b/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
@@ -55,6 +55,7 @@ public class ApiConstants {
     public static final String CERTIFICATE_CHAIN = "certchain";
     public static final String CERTIFICATE_FINGERPRINT = "fingerprint";
     public static final String CERTIFICATE_ID = "certid";
+    public static final String ENABLED_REVOCATION_CHECK = "enabledrevocationcheck";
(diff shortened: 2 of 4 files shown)


artem-smotrakov:
The failed jobs in Travis don't seem to be related to the update. If I understand correctly, they failed due to too many log messages. The update doesn't introduce new output.

DaanHoogland:
@artem-smotrakov the travis jobs often fail because of the log lengths. There is no way around that but restarting and reading the logs to make sure there is nothing in it that log. I restarted the two failed jobs.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

api/src/main/java/org/apache/cloudstack/api/command/user/loadbalancer/UploadSslCertCmd.java
```diff
@@ -76,6 +76,9 @@
     @Parameter(name = ApiConstants.NAME , type = CommandType.STRING, required = true, description = "Name for the uploaded certificate")
     private String name;
 
+    @Parameter(name = ApiConstants.ENABLED_REVOCATION_CHECK, type = CommandType.BOOLEAN, description = "Enables revocation checking for certificates")
```
DaanHoogland, line 79: can you add `@since` please?

DaanHoogland:

api/src/main/java/org/apache/cloudstack/api/command/user/loadbalancer/UploadSslCertCmd.java
artem-smotrakov: @DaanHoogland Do you mean `since` parameter for the `@Parameter` annotation?

```@Parameter(..., since = "4.14" ...)```

Or, do you mean `@since` in javadoc?

Which version should it be? 4.14?

DaanHoogland: no this one, it should be 4.15 (the `since` attribute is maybe a less confusing term for the same thing, forgive my semantic imprecision and follow up nitpicking;)

artem-smotrakov: No problem, I've updated the code.

rhtyd:
We've two lgtms and test pass, I'll merge this.

PR #4066, opened 2020-05-09
kvm: newer qemu-img and libvirt 6.0 expects backing file format

rhtyd:
Newer libvirt 6.0 expects the backing file to be specified, or when creating volume with qemu-img specify backing file format otherwise VM deployment fails.

https://libvirt.org/kbase/backing_chains.html

Tested on Ubuntu 20.04/arm64 on a raspberrypi4.

TODO: check and use the -F format parameter depending on libvirt version, otherwise this param would fail on older systems.

## Types of changes

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
+++ b/plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/utils/qemu/QemuImg.java
@@ -150,6 +150,8 @@ public class QemuImg {
         s.add("-f");
         if (backingFile != null) {
             s.add(backingFile.getFormat().toString());
+            s.add("-F");
+            s.add(backingFile.getFormat().toString());
             s.add("-b");
             s.add(backingFile.getFileName());
         } else {


rhtyd:
@wido @GabrielBrascher @kiwiflyer @weizhouapache - have you explored any plans on supporting Ubuntu 20.04 as kvm host, with a very newer libvirt ?

rhtyd:
cc @shwstppr @davidjumani - I've not checked libvirtd version on CentOS8, pinging for reference.

weizhouapache:
> @wido @GabrielBrascher @kiwiflyer @weizhouapache - have you explored any plans on supporting Ubuntu 20.04 as kvm host, with a very newer libvirt ?

@rhtyd I am working on ubuntu 20.04 support.
Besides this change, we have some other changes. 
For example, different python version in packaging, libvirtd config, set backing file format when migrate a vm, etc.

rhtyd:
Great @weizhouapache my colleagues @davidjumani and @shwstppr will be working to support CentOS8 as management server and KVM host for cloudstack, there will be some overlap esp around migrating the python scripts to python3. I've asked them to open a draft PR where you're welcome to collaborate.
cc @PaulAngus

weizhouapache:
@rhtyd 
python3 support and ubuntu 20.04 support are two unrelated tasks I think.
python3 can be installed on centos7 and ubuntu 18.04.
python2 can be installed on ubuntu 20.04 and it looks good in my testing until now.

wido:
This is on the to-do of @GabrielBrascher and me for the 4.15 release.

We should stay away from Python 2 @weizhouapache as we are already working on getting all the code Py3 ready. Py2 is EOL and shouldn't be used anymore.

Most of the code for on the KVM Agent, like the Security Grouping has already been fixed to work under Py3.

But we haven't looked into all the details for Ubuntu 20.04 yet.

weizhouapache:
@rhtyd @wido 
I have fixed some issues in ubuntu 20.04. it works fine with 4.14.
https://github.com/apache/cloudstack/pull/4069

Questions
(1) will we support ubuntu 20.04 in 4.14.1.0 ?
(2) will we move to python3 in 4.14.1.0 (all python scripts on hypervisor) ?

wido:
4.14 will not be a LTS, so why will there be a .1 release?

And I think we shouldn't make a Py3 change in a minor release. This should be done in 4.15.

20.04 support in 4.14 is too late now I think.

weizhouapache:
@wido @rhtyd @PaulAngus @andrijapanicsb 
I have the impression that 4.14 will be next LTS. We have discussed it before. Is there conclusion for the discussion ?

wido:
@weizhouapache This is maybe not the place for that discussion, but I thought that because 4.13 was an LTS that 4.14 wouldn't be one and 4.15 would become the next LTS.

And again, changing a lot of things in a minor release, like 4.14.1 would not be something I would recommend/like.

andrijapanicsb:
@weizhouapache , as 4.14.1 would be a fix release for the 4.14, so in short NO new features should go in, as far as I understand - 4.14.1. (4.13.1 or 4.15.1) would always have to be just a fix releases - so better target this against 4.15?

rhtyd:
I'll close this in favour of @weizhouapache 's https://github.com/apache/cloudstack/pull/4069

PR #4076, opened 2020-05-11
Support searching vm's by multiple id

ravening:
## Description
Add support to list virtual machines by multiple ids (not uuid)
In older version of cloudstacj, the vm id's are stored as
integer and not as UUID's. Currently it supports searching for
VM's using multiple UUID's but not possible to search VM's
using multiple ID's

## Types of changes

## How Has This Been Tested?
Before the fix below api calls didnt work

```
(local) > list virtualmachines ids=1697,1698,1699 listall=true
```

After the fix it will work

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
@@ -70,7 +70,7 @@ public class ListVMsCmd extends BaseListTaggedResourcesCmd implements UserCmd {
     @Parameter(name = ApiConstants.ID, type = CommandType.UUID, entityType = UserVmResponse.class, description = "the ID of the virtual machine")
     private Long id;
 
-    @Parameter(name=ApiConstants.IDS, type=CommandType.LIST, collectionType=CommandType.UUID, entityType=UserVmResponse.class, description="the IDs of the virtual machines, mutually exclusive with id", since = "4.4")
+    @Parameter(name=ApiConstants.IDS, type=CommandType.LIST, collectionType=CommandType.UUID, entityType=UserVmResponse.class, description="the IDs of the virtual machines, mutually exclusive with id", since = "")
     private List<Long> ids;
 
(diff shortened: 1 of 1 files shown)


svenvogel:
@ravening Hi. is this an enhancement or bug fix? i saw you removed the 4.4. how does this fix the problem?

ravening:
> @ravening Hi. is this an enhancement or bug fix? i saw you removed the 4.4. how does this fix the problem?

Hi @svenvogel this is a bug fix. After removing the "4.4", it works fine. We are using 4.7 version and had this issue

DaanHoogland:
@ravening, did you check that searching for UUIDs still works?

ravening:
> @ravening, did you check that searching for UUIDs still works?

yes search multiple uuid or single uuid still works

svenvogel:
@ravening the 4.4 sound like a search blocker only for a specific version or what do you think?

ravening:
> @ravening the 4.4 sound like a search blocker only for a specific version or what do you think?

@svenvogel "since=4.4" indicates that its only supported from version 4.4 onwards. Since there was "id's" instead of "uuid" before 4.4, search for list of "id's" is not possible. Removing it will ensure that we can search it in any version

DaanHoogland:
@ravening are you running a pre-4.4 version?

ravening:
> @ravening are you running a pre-4.4 version?

Not anymore. There were few vm's which were created on 4.2 version and when upgraded to 4.7, listing vm's with multiple id's(not uuid's) didnt work anymore.

DaanHoogland:
@ravening, so did you find out what caused this to work this way? The chance seems so unrelated to your objectives that it reeks of an undesirable side effect.
also, did you try removing the since attribute altogether?

rhtyd:
@ravening we don't want end-users to be able to list resources by integer IDs; the support we've is largely to keep some of the APIs/params compatible. Given the API indeed works when multiple comma-separated IDs (uuids/strings) are passed, the support for searching by multiple IDs already works.

rhtyd:
Closing on previous remark and no further communication.

PR #4077, opened 2020-05-11
Disable searching by instance name for customers

ravening:
## Description
Only admins should be able to search VM by instance name
Customers should not see or serach VM's using the instance name (i-)

## Types of changes

## How Has This Been Tested?
Login as "admin".
Note down the i- name of the customer VM.
Now login to the customer domain
Try to search for VM using the i- name
It should not display the VM. It should display only for admin

### DIFF
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -993,7 +993,9 @@ public class QueryManagerImpl extends MutualExclusiveIdsManagerBase implements Q
             SearchCriteria<UserVmJoinVO> ssc = _userVmJoinDao.createSearchCriteria();
             ssc.addOr("displayName", SearchCriteria.Op.LIKE, "%" + keyword + "%");
             ssc.addOr("name", SearchCriteria.Op.LIKE, "%" + keyword + "%");
-            ssc.addOr("instanceName", SearchCriteria.Op.LIKE, "%" + keyword + "%");
+            if (isRootAdmin) {
+                ssc.addOr("instanceName", SearchCriteria.Op.LIKE, "%" + keyword + "%");
+            }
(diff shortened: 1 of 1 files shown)


DaanHoogland:
I appriciate that you considder this an enhancement/bug fix for your site @ravening , but I'm not convinced that this is true for every site. For instance a corporate cloud used in internal development might rely in their automation on instance names. Can you guard this with a setting allowing operators to let users search in the old fashion if they so wish?
cc @PaulAngus @rhtyd @andrijapanicsb

andrijapanicsb:
As long as the customer can not search instances by i-xxx-yy-vm name that are OUTSIDE of his domain, I don't see an issue here? Optionally, this should be configurable by a global setting, with defaulting to the current behaviour (unless there is some security concern which I did not read here?)

ravening:
@DaanHoogland @andrijapanicsb yes I will add a global setting so that users can enable/disable it

rhtyd:
No need @ravening - PR LGTM
@DaanHoogland @andrijapanicsb I checked the API response we don't return the `instancename` to the user as part of the API response, so we can safely assume that end users wouldn't know the internal name of the VM. In case of admins, this is returned such as:
```
      "instancename": "i-2-94-QA",
```

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
Business layer only change, Travis/smoketests are enough I think.

andrijapanicsb:
Overall LGTM, but have not tested it.

DaanHoogland:
ok @rhtyd @ravening I'm fine, i hadn't thought of instancename vs name. let's merge

rhtyd:
Wait - please HOLD on merging - I probably found a regression, need to check something.

rhtyd:
Checked and reported a bug in listVirtualMachinesMetrics API - https://github.com/apache/cloudstack/issues/4143
This PR is okay, let's merge.

Issue #2861, opened 2018-09-21
SSVM NFS4 does not support minor version setting e.g. 4.1


resmo:
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
SecondaryStorage VM
~~~

##### CLOUDSTACK VERSION

~~~
4.11.1
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

Debian 9 system VM default NFS version is 4.2. If your NFS server does not support 4.2, the mount may fail (is does in our case).

There is a global config `secstorage.nfs.version` to specify the version, but it does only support int values. If you set it to `4` the NFS4 version used is `4.0`. 

There is currently no way to set a minor version. e.g. `4.1`.

:warning: Note that version 3 has no "minor versions". 

Probably worth considering is to add new configuration field for nfs4 minor versions. e.g. `secstorage.nfs4.minorversion` defaulted to null.  NFS4 version shoulld be a combination of 
`secstorage.nfs.version + '.' + secstorage.nfs4.minorversion`. NFS3 only `secstorage.nfs.version`

rhtyd:
Looks like a bug, if it was previously supported/working. However, not a major blocker, I'll try and fix it before RC2.

resmo:
not a big issue, could be in 4.11.3.

PaulAngus:
looks like it would be a simple enhancement to add to nfs mount commands:

mount -t nfs4 -o rw,nolock,nfsvers=4,minorversion=1 nfsserver:/path /mnt/path

But this needn't hold up 4.11.2

rhtyd:
cc @Spaceman1984

rhtyd:
Fixed in https://github.com/apache/cloudstack/pull/4180

PR #4084, opened 2020-05-15
Added nfs minor version support on SSVM

Spaceman1984:
## Description

This PR adds minor version support when mounting nfs on the SSVM as requested in #2861 

The global setting "secstorage.nfs.version" has been changed to use String data type which allows any minor version to be specified.

## Types of changes

## How Has This Been Tested?

This was tested by changing the global setting values and restarting the SSVM, then verifying if the secondary storage was mounted with the specified version by logging into the SSVM and typing "mount" at the cli.

Tested with SSVM running on KVM, Xenserver and VMWare hosts.

### DIFF
--- a/engine/components-api/src/main/java/com/cloud/capacity/CapacityManager.java
+++ b/engine/components-api/src/main/java/com/cloud/capacity/CapacityManager.java
@@ -73,9 +73,9 @@ public interface CapacityManager {
                     "If set to true, creates VMs as full clones on ESX hypervisor",
                     true,
                     ConfigKey.Scope.StoragePool);
-    static final ConfigKey<Integer> ImageStoreNFSVersion =
-            new ConfigKey<Integer>(
-                    Integer.class,
+    static final ConfigKey<String> ImageStoreNFSVersion =
+            new ConfigKey<String>(
+                    String.class,
                     "secstorage.nfs.version",
                     "Advanced",
                     null,
--- a/api/src/main/java/com/cloud/agent/api/to/NfsTO.java
+++ b/api/src/main/java/com/cloud/agent/api/to/NfsTO.java
@@ -24,7 +24,7 @@ public class NfsTO implements DataStoreTO {
     private DataStoreRole _role;
     private String uuid;
     private static final String pathSeparator = "/";
-    private Integer nfsVersion;
+    private String nfsVersion;
 
     public NfsTO() {
 
@@ -73,11 +73,11 @@ public class NfsTO implements DataStoreTO {
         return pathSeparator;
     }
 
-    public Integer getNfsVersion() {
+    public String getNfsVersion() {
         return nfsVersion;
     }
 
-    public void setNfsVersion(Integer nfsVersion) {
+    public void setNfsVersion(String nfsVersion) {
         this.nfsVersion = nfsVersion;
     }
 }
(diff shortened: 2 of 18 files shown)


engine/components-api/src/main/java/com/cloud/capacity/CapacityManager.java
```diff
@@ -73,16 +73,26 @@
                     "If set to true, creates VMs as full clones on ESX hypervisor",
                     true,
                     ConfigKey.Scope.StoragePool);
-    static final ConfigKey<Integer> ImageStoreNFSVersion =
+    static final ConfigKey<Integer> ImageStoreNFSMajorVersion =
```
rhtyd, line 76: @Spaceman1984 can you revert the change, why not parse for the major and minor version? Something like `x.y` could be parsed by splitting on the `.` if provided.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareManagerImpl.java
```diff
@@ -831,8 +832,12 @@ protected String mount(String path, String parent, Integer nfsVersion) {
         String result = null;
         Script command = new Script(true, "mount", _timeout, s_logger);
         command.add("-t", "nfs");
-        if (nfsVersion != null){
-            command.add("-o", "vers=" + nfsVersion);
+        if (nfsMajorVersion != null){
+            String nfsVersion = "vers=" + nfsMajorVersion;
+            if (nfsMinorVersion != null && nfsMajorVersion >= 4) {
```
rhtyd, line 837: Is minor version only supported for version 4 and above? If so, why not simply parse and pass the major version when it is less than `3`?

engine/components-api/src/main/java/com/cloud/capacity/CapacityManager.java
Spaceman1984: @rhtyd If I was to split on `.` I would have to change the Integer data type to String, basically undoing all the changes I have made. Doing it this way preserves backward compatibility if a value has already been stored for NFS version and adding a second variable for minor version is as per the suggestion in the original issue.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareManagerImpl.java
Spaceman1984: This code does exactly that, I don't understand what you would want to change.

rhtyd: @Spaceman1984 why not simply pass the version (a single variable) than two?

rhtyd:
ping @Spaceman1984

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/manager/VmwareManagerImpl.java
Spaceman1984: I'll change it to 1 field

Spaceman1984:
@blueorangutan package

Spaceman1984:
Closing PR due to git issue

PR #4092, opened 2020-05-20
engine/schema: add empty DB upgrade path from 4.14.0.0 to 4.15.0.0

shwstppr:
## Description
Empty DB upgrade path from 4.14.0.0 to 4.15.0.0

## Types of changes


engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41400to41500.java
```diff
@@ -0,0 +1,248 @@
...
+        return new InputStream[] {script};
+    }
+
+    @Override
+    public void performDataMigration(Connection conn) {
+        updateSystemVmTemplates(conn);
+    }
+
+    @SuppressWarnings("serial")
+    private void updateSystemVmTemplates(final Connection conn) {
```
rhtyd, line 71: @shwstppr can you remove any similar check in the previous upgrade path 4.13->4.14 for example? We don't want multiple instances of systemvm migration code in older upgrade paths.

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41400to41500.java
shwstppr: @rhtyd removed the systemvm code from 41310to41400. Please confirm if it is fine now

DaanHoogland:
I am having second thoughts about this one @rhtyd @shwstppr ! we shouldn't have to install the 4.14 templates when upgrade from 4.14, should we? these should already be in place.

rhtyd:
@DaanHoogland you're right I actually wanted that to be moved to 4.14-4.15 upgrade path so I can fix it in here - https://github.com/apache/cloudstack/pull/4104/files We'll have a new systemvmtemplate for 4.15 (Debian 10 based). It won't be problem if we merge #4104

DaanHoogland:
ok, @rhtyd that means we must give that one high priority as no update (addmittedly of dev envs only) will work

Issue #4127, opened 2020-06-03
Update proper error msg and status when download template/ISO fails due to invalid file format


sureshanaparti:
### Description

The template download fails immediately with download state: DOWNLOAD_ERROR and error msg: Steam closed, for templates with invalid file format. The proper error message has to be reported when the download template/ISO fails due to invalid file format. The template status & download state should be set accordingly so that the SSVM never tries to re-download the template.

The file format is validated in SSVM using the output of cmd: "file <template-file> | cut -d: -f2", after downloading atleast 1MB of data.

##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
Template, SSVM
~~~

##### CLOUDSTACK VERSION

~~~
4.13.0
~~~

##### SUMMARY

The Http template downloader in SSVM is aborting the connection immediately for the templates with invalid file format, and the template download fails with error msg: Stream closed as shown in below logs. The proper error message has to be reported when the download template/ISO fails due to invalid file format. Also, update download state & status of the template so that the SSVM never tries to re-download the template. Handle any cleanups if required.

The file format is validated in SSVM using the output of cmd: "file <template-file> | cut -d: -f2", after downloading atleast 1MB of data.

logs:
~~~
Jun  3 07:17:22 s-2-VM _run.sh[1312]: 07:17:22,750  INFO HttpTemplateDownloader:207 - Starting download from http://***/****.ova to /mnt/SecStorage/802dfb3b-f7ef-3aec-9f46-c3dc12358ac8/template/tmpl/2/210/dnld982911992150124927tmp_ remoteSize=17536495476 , max size=53687091200
Jun  3 07:17:23 s-2-VM _run.sh[1312]: 07:17:23,372  INFO DownloadManagerImpl:270 - Download Completion for jobId: 0a4b329e-7698-4878-9d7d-0ea27c9796bd, status=UNRECOVERABLE_ERROR
Jun  3 07:17:23 s-2-VM _run.sh[1312]: 07:17:23,372  INFO DownloadManagerImpl:271 - local: /mnt/SecStorage/802dfb3b-f7ef-3aec-9f46-c3dc12358ac8/template/tmpl/2/210/dnld982911992150124927tmp_, bytes=1104917, error=Stream closed, pct=0
~~~

![TemplateDownload_Steam-close](https://user-images.githubusercontent.com/12028987/83661688-1adc6780-a5e4-11ea-8c14-684ec510e6e6.png)

[...]

PR #4138, opened 2020-06-10
Fixed incorrect error message on invalid template type download

Spaceman1984:
## Description

When a template is downloaded, the first 1MB of the template is validated to determine if the template is of correct file type. On failure, the download is aborted and the input stream is set to null. This leads to a second error when the try-with-resources block tries to auto-close the stream and throws an ioexception. The "stream closed" error message is then written to the db. 

This PR checks if an error has been stored before setting a new error message.

Fixes: #4127 
## Types of changes

## How Has This Been Tested?

This has been tested by downloading an invalid template and checking the error message from the UI or in cloud.template_store_ref.error_str in the db.
The cleanup has been tested by restarting the management server to trigger template sync to verify if the dead template has been removed.


Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

core/src/main/java/com/cloud/storage/template/HttpTemplateDownloader.java
```diff
@@ -218,7 +218,10 @@ public long download(boolean resume, DownloadCompleteCallback callback) {
             errorString = hte.getMessage();
         } catch (IOException ioe) {
             status = TemplateDownloader.Status.UNRECOVERABLE_ERROR; //probably a file write error?
-            errorString = ioe.getMessage();
+            // Let's not overwrite the original error message.
+            if (errorString.isEmpty()){
```
DaanHoogland, line 222: errorString contains a single space initially. String.isEmpty returns (this.value.length == 0)
so this is never happening.

Spaceman1984: You are correct @DaanHoogland, I have made the change.

core/src/main/java/com/cloud/storage/template/HttpTemplateDownloader.java
```diff
@@ -219,7 +219,7 @@ public long download(boolean resume, DownloadCompleteCallback callback) {
         } catch (IOException ioe) {
             status = TemplateDownloader.Status.UNRECOVERABLE_ERROR; //probably a file write error?
             // Let's not overwrite the original error message.
-            if (errorString.isEmpty()){
+            if (errorString.equals(" ")){
```
DaanHoogland, line 222: i'd go for `errorString.trim().isEmpty()` to be honest (to be more defensive, but this should work as well. And it might be wise to alter the initialisation statement as well???

Spaceman1984: What if we get rid of the check altogether and append the error message to what was there already?

engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
```diff
@@ -489,6 +489,10 @@ public void handleTemplateSync(DataStore store) {
                                     s_logger.info("Removing leftover template " + uniqueName + " entry from template store table");
                                     // remove those leftover entries
                                     _vmTemplateStoreDao.remove(tmpltStore.getId());
+                                    // remove from zones
+                                    _vmTemplateZoneDao.deletePrimaryRecordsForTemplate(tmplt.getId());
+                                    // remove template
+                                    _templateDao.remove(tmplt.getId());
```
rhtyd, line 495: @Spaceman1984 this probably needs checking, what if the templates are OK for other secondary storages within the same zone; and across zones?

Spaceman1984: I'll test with multiple secondary storages

Spaceman1984: You are correct @rhtyd, this code removes the template if there is an error in any of the storages. I have reverted.

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

Spaceman1984:
@blueorangutan test

core/src/main/java/com/cloud/storage/template/HttpTemplateDownloader.java
```diff
@@ -218,7 +219,10 @@ public long download(boolean resume, DownloadCompleteCallback callback) {
             errorString = hte.getMessage();
```
nvazquez, line 219: Why not set it if errorString is null as below?

core/src/main/java/com/cloud/storage/template/HttpTemplateDownloader.java
```diff
@@ -218,7 +219,10 @@ public long download(boolean resume, DownloadCompleteCallback callback) {
...
         } catch (IOException ioe) {
             status = TemplateDownloader.Status.UNRECOVERABLE_ERROR; //probably a file write error?
-            errorString = ioe.getMessage();
+            // Let's not overwrite the original error message.
+            if (errorString == null){
```
nvazquez, line 223: Minor format one: can you leave a space between the closing parenthesis and the opening brace?

core/src/main/java/com/cloud/storage/template/HttpTemplateDownloader.java
Spaceman1984: > Why not set it if errorString is null as below?

It may be an easy code change, but then the scope of this issue will be greater, which will require more testing as well.

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

rhtyd:
@nvazquez are you LGTM on this?

core/src/main/java/com/cloud/storage/template/HttpTemplateDownloader.java
```diff
@@ -243,7 +247,6 @@ private boolean copyBytes(File file, InputStream in, RandomAccessFile out) throw
                 offset = writeBlock(bytes, out, block, offset);
                 if (!verifyFormat.isVerifiedFormat() && (offset >= 1048576 || offset >= remoteSize)) { //let's check format after we get 1MB or full file
                     verifyFormat.invoke();
-                    if (verifyFormat.isInvalid()) return true;
```
rhtyd, line 246: @Spaceman1984 cc @nvazquez why are we not validating format now?

Spaceman1984: @rhtyd If validation fails, an exception is thrown, so there is no setting of an invalid state that has to be checked.

nvazquez: In short, prior to these changes the invoke() method only set the boolean variable retrieved in case of valid/invalid. After these changes, if it is invalid invoke() throws an exception, so no need to check isInvalid() in this line

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@Spaceman1984 can you take a look at the failing tests?

Spaceman1984:
@nvazquez I had a look at the errors, doesn't seem related to template downloads. I will fire the tests again.

Spaceman1984:
@blueorangutan test

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

PR #4144, opened 2020-06-11
Fix Usage failed to get pid

div8cn:
## Description

cloudstack-usage use System.getProperty("pid") to get the process PID
System.getProperty cannot get the Pid, which will cause the service to stay in the config stage. Cannot continue to start, and there will be no log output.

This PR uses ManagementFactory instead of System.getProperty to get Pid, and throws an exception when Pid cannot be obtained

## Types of changes

### DIFF
--- a/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
+++ b/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
@@ -278,12 +278,14 @@ public class UsageManagerImpl extends ManagerBase implements UsageManager, Runna
             throw new ConfigurationException("Unhandled exception configuring UsageManager " + e.toString());
         }
 
+        String processName;
         try {
-            String processName = ManagementFactory.getRuntimeMXBean().getName();
+            processName = ManagementFactory.getRuntimeMXBean().getName();
             _pid = Integer.parseInt(processName.split("@")[0]);
         } catch (Exception e) {
-            s_logger.error("Unable to get process pid ", e);
-            throw new ConfigurationException("Unable to get process pid " + e.toString());
+            String msg = String.format("Unable to get process Id for %s!", processName);
+            s_logger.debug(msg , e);
+            throw new ConfigurationException(msg, e);
         }
         return true;
     }
--- a/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
+++ b/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
@@ -33,6 +33,8 @@ import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
+import java.lang.management.ManagementFactory;
+
 import org.apache.cloudstack.quota.QuotaAlertManager;
 import org.apache.cloudstack.quota.QuotaManager;
 import org.apache.cloudstack.quota.QuotaStatement;
@@ -275,7 +277,14 @@ public class UsageManagerImpl extends ManagerBase implements UsageManager, Runna
             s_logger.error("Unhandled exception configuring UsageManger", e);
             throw new ConfigurationException("Unhandled exception configuring UsageManager " + e.toString());
         }
-        _pid = Integer.parseInt(System.getProperty("pid"));
+
+        try {
+            String processName = ManagementFactory.getRuntimeMXBean().getName();
+            _pid = Integer.parseInt(processName.split("@")[0]);
+        } catch (Exception e) {
+            s_logger.error("Unable to get process pid ", e);
+            throw new ConfigurationException("Unable to get process pid " + e.toString());
+        }
         return true;
     }
 
--- a/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
+++ b/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
@@ -33,8 +33,6 @@ import java.util.concurrent.TimeUnit;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
-import java.lang.management.ManagementFactory;
-
 import org.apache.cloudstack.quota.QuotaAlertManager;
 import org.apache.cloudstack.quota.QuotaManager;
 import org.apache.cloudstack.quota.QuotaStatement;
@@ -280,8 +278,7 @@ public class UsageManagerImpl extends ManagerBase implements UsageManager, Runna
 
         String processName = null;
         try {
-            processName = ManagementFactory.getRuntimeMXBean().getName();
-            _pid = Integer.parseInt(processName.split("@")[0]);
+            _pid = (int) ProcessHandle.current().pid();
         } catch (Exception e) {
             String msg = String.format("Unable to get process Id for %s!", processName);
             s_logger.debug(msg , e);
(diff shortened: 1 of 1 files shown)


rhtyd:
@div8cn where did you reproduce the issue, OS/distro, CloudStack version etc? Thanks

div8cn:
I encountered this problem when building cloudstack-usage into docker and started using supervisord。

DaanHoogland:
@blueorangutan package

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -275,7 +277,14 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
...
             throw new ConfigurationException("Unhandled exception configuring UsageManager " + e.toString());
         }
-        _pid = Integer.parseInt(System.getProperty("pid"));
+
+        try {
+            String processName = ManagementFactory.getRuntimeMXBean().getName();
+            _pid = Integer.parseInt(processName.split("@")[0]);
+        } catch (Exception e) {
+            s_logger.error("Unable to get process pid ", e);
+            throw new ConfigurationException("Unable to get process pid " + e.toString());
```
ravening, line 286: Should we add a more meaningful message here so its more clear?

rhtyd:
@div8cn can you check if in the docker container the pid parameter was passed? Depending on how the service runs perhaps this was not passed, look at the cloudstack-usage systemd service for example.

DaanHoogland:
@blueorangutan package

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -275,7 +277,14 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
...
+        }
```
DaanHoogland, line 287: 1. twice the same message should be in a string constand
1. rethrown should not abandon the stack trace for the original problem
```suggestion
        String processName;
        try {
            processName = ManagementFactory.getRuntimeMXBean().getName();
            _pid = Integer.parseInt(processName.split("@")[0]);
        } catch (Exception e) {
            String msg = String.format("Unable to get process Id for %s!", processName);
            s_logger.error(msg);
            throw new ConfigurationException(msg, e);
        }
```
or
```suggestion
        String processName;
        try {
            processName = ManagementFactory.getRuntimeMXBean().getName();
            _pid = Integer.parseInt(processName.split("@")[0]);
        } catch (Exception e) {
            String msg = String.format("Unable to get process Id for %s!", processName);
            s_logger.debug(msg , e);
            throw new ConfigurationException(msg, e);
        }
```

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
DaanHoogland: I'm all for clarity @ravening . what do you suggest?

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -275,7 +277,16 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
...
-        _pid = Integer.parseInt(System.getProperty("pid"));
+
+        String processName = null;
+        try {
+            processName = ManagementFactory.getRuntimeMXBean().getName();
+            _pid = Integer.parseInt(processName.split("@")[0]);
+        } catch (Exception e) {
+            String msg = String.format("Unable to get process Id for %s!", processName);
+            s_logger.debug(msg , e);
+            throw new ConfigurationException(msg + " " + e.toString());
```
GabrielBrascher, line 288: @DaanHoogland @div8cn What do you think of using `e.printStackTrace()` to add the stack on the log/ConfigurationException?

With `toString` (or simply s_logger.debug(msg , e)) it logs only the exception name and the error message (if there is any message). On the other hand, `printStackTrace()` presents the whole stack trace of an exception, which is very helpful for debugging.

DaanHoogland: `logger.debug(msg, e)` should log the entire stacktrace @GabrielBrascher . and it gives the user to filter or direct the output to a certain stream/file. `e.printStacktrace()` only logs to stdout (or stderr, i forgot). in short i think not too much of it. I agree toString does not seem the best, but we have alreadyt output the stacktrace by then. A change from `toString()` to `getLocalizedMessage()` maybe, @div8cn? Of course we can add the entire exception again i.e. `throw new ConfigurationException(msg, e)`, that's fine as well.

rhtyd:
@blueorangutan package

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -275,7 +277,16 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
             s_logger.error("Unhandled exception configuring UsageManger", e);
             throw new ConfigurationException("Unhandled exception configuring UsageManager " + e.toString());
         }
-        _pid = Integer.parseInt(System.getProperty("pid"));
+
+        String processName = null;
+        try {
+            processName = ManagementFactory.getRuntimeMXBean().getName();
```
rhtyd, line 283: @div8cn master is on java 11 now, can you use `ProcessHandle.current().pid()`? I did a quick test:
```
> jshell 
|  Welcome to JShell -- Version 11.0.7
|  For an introduction type: /help intro

jshell> ProcessHandle.current().pid()
$1 ==> 20380

```

div8cn:
@rhtyd  Thanks, I've updated it to ProcessHandle.current ().pid().

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -280,8 +278,7 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
 
         String processName = null;
         try {
-            processName = ManagementFactory.getRuntimeMXBean().getName();
-            _pid = Integer.parseInt(processName.split("@")[0]);
+            _pid = (int) ProcessHandle.current().pid();
         } catch (Exception e) {
             String msg = String.format("Unable to get process Id for %s!", processName);
```
weizhouapache, line 283: processName is always null .

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -275,7 +275,15 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
             s_logger.error("Unhandled exception configuring UsageManger", e);
             throw new ConfigurationException("Unhandled exception configuring UsageManager " + e.toString());
         }
-        _pid = Integer.parseInt(System.getProperty("pid"));
+
+        String processName = null;
```
rhtyd, line 279: processName is not used and can be removed @div8cn

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd (approved):

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
though the ha error is a bit offsetting I see not how any of these errors can be related, @rhtyd agree?

rhtyd (approved):

rhtyd:
Yup LGTM

PR #4148, opened 2020-06-16
server: Do not resize volume of running vm on KVM host if host is not Up or not Enabled

ustcweizhou:
##  Description

If we resize a volume of a vm running on a host which is not Up or not Enable, the job will be scheduled to another normal host. Then the volume will be resized by "qemu-img resize" instead of "virsh blockresize", the image might be corrupted after resize.

## Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
@@ -1192,8 +1192,13 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
 
             /* Do not resize volume of running vm on KVM host if host is not Up or not Enabled */
             if (currentSize != newSize && userVm.getState() == State.Running && userVm.getHypervisorType() == HypervisorType.KVM) {
+                if (userVm.getHostId() == null) {
+                    throw new InvalidParameterValueException("Cannot find the hostId of running vm " + userVm.getUuid());
+                }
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1187,6 +1189,16 @@ private VolumeVO orchestrateResizeVolume(long volumeId, long currentSize, long n
             if (currentSize != newSize && _volsDao.getHypervisorType(volume.getId()) == HypervisorType.XenServer && !userVm.getState().equals(State.Stopped)) {
                 throw new InvalidParameterValueException(errorMsg);
             }
+
+            /* Do not resize volume of running vm on KVM host if host is not Up or not Enabled */
+            if (currentSize != newSize && userVm.getState() == State.Running && userVm.getHypervisorType() == HypervisorType.KVM) {
+                HostVO host = _hostDao.findById(userVm.getHostId());
+                if (host.getStatus() != Status.Up) {
```
rhtyd, line 1196: @ustcweizhou should you do a null check, what if the host was removed?

weizhouapache: @rhtyd added null check

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
LGTM

PR #4149, opened 2020-06-16
set python explicit during transistion

DaanHoogland:
this will allow for easier switching back and forth during development on python 3 functionality

## Types of changes

## How Has This Been Tested?

regular use on a default python3 platform


yadvr (approved):
I'm not sure if we use this as pyenv is not used generally in our workflows

DaanHoogland:
we don't use pyenv from mvn do we?

GabrielBrascher (commented):
@DaanHoogland unfortunately mvn build is failing due to pyenv version:

```
pyenv: version `2.7.18' is not installed (set by /home/travis/build/apache/cloudstack/.python-version)
```

DaanHoogland:
@GabrielBrascher then you have to change your version to the one you did install, or not set it at all. I have to switch between 2 and 3 as well. That is exactly the point of the change. after transistion it should read some 3 version or be deleted if we want to... I'm not sure if just 2 or just 3 works.

if there is no interest in this, I'll just close.

GabrielBrascher:
@DaanHoogland I have no particular issue with this change, considering it is meant to the transition from Python 2 to 3.

The issue that I raised is that if this PR gets merged then Travis will fail for all PRs based on the master.

We should then check what version Travis has installed and set to that version or see what we can do to change Travis version. Otherwise, all PRs will have the following issue ```pyenv: version '2.7.18' is not installed (set by /home/travis/build/apache/cloudstack/.python-version).```

GabrielBrascher (approved):
I don't see why not having it for this period of transition Python 2 to 3.
LGTM.

rhtyd:
@DaanHoogland can this be closed?

Issue #4129, opened 2020-06-05
Certificate auto renewal fails after agent disconnection or re-provision certificate to host


sureshanaparti:
### Description

Agent disconnection event is purging the certificates from the active certificates map, being maintained by the management server in memory. Re-provisioning of the certificate also triggers agent disconnect, which will result the same.

As the active certificates map doesn't hold any certificate details for the host, certificate authority framework will fail to auto renew the host's certificate after the expiry alert period configured using the setting _ca.framework.cert.expiry.alert.period_

For clustered management servers, the latest certificates should be owned/maintained (in memory) by the resource owner i.e the MS who owns the host, irrespective of the change in resource ownership or provisioning certificate from the other MS who doesn't own the host. Currently, the latest copy is being maintained in the MS who triggers the provision certificate.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
CA Manager
~~~

##### CLOUDSTACK VERSION

~~~
master
~~~

##### CONFIGURATION

MS, using Securing Agents Communication

##### SUMMARY

The certificates details of the host are removed from the active certificates map maintained by the management server when agent is disconnected. The certificate re-provisioning of host would also trigger agent disconnect, which will delete the certificate for the host.

As the active certificates map doesn't hold any certificate details for the host, certificate authority framework will fail to auto renew the host's certificate after the expiry alert period configured using the setting _ca.framework.cert.expiry.alert.period_

For clustered management servers, the latest certificates should be owned/maintained (in memory) by the resource owner i.e the MS who owns the host, irrespective of the change in resource ownership or provisioning certificate from the other MS who doesn't own the host.

##### STEPS TO REPRODUCE

1) Configure the CA config parameters and restart MS.
- Enable auto certificate renewal by setting the global/cluster config parameter to "_ca.framework.cert.automatic.renewal_" to true.
(Set minimal possible values for validity period and  expiry alert period.)
- Set validity period using config parameter "_ca.framework.cert.validity.period_" to 2 days, at global level.
- Set the expiry alert period using config parameter "_ca.framework.cert.expiry.alert.period_" to 1 day, at global/respective cluster level.
2) Provision host security keys (or Disconnect the agent)

Enable debug logs and the below logs are not observed after expiry alert period. Also, no alerts on certificate expiry.

~~~
[...]

PR #4156, opened 2020-06-19
Fixed removal of hosts from certsmap when running certificate auto-renew

Spaceman1984:
## Description

When a host connects to a management server, the host IP address and the certificate are stored in memory on the management server. This mapping is checked periodically to determine if any certificates are due to expire.

Before a certificate is renewed, a few checks are done to determine if the host is connected to the management server by fetching the host record from the database. The problem here is if the wrong record is fetched, the host is not checked for renewal.

This PR improves the host record fetch from the database by looking only at hosts that are not removed.

Fixes: #4129 
## Types of changes

## How Has This Been Tested?

This has been tested by setting the ca.framework.cert.validity.period and ca.framework.cert.expiry.alert.period to the same value. This is to ensure that a certificate is up for renewal as soon as it is issued.
Then watch the management server logs to see if auto-renewal happens.

This has also been tested by using two management servers and reprovisioning host security keys from the second management server and still having the certs auto-renew.

### DIFF
--- a/engine/schema/src/main/java/com/cloud/host/dao/HostDao.java
+++ b/engine/schema/src/main/java/com/cloud/host/dao/HostDao.java
@@ -97,6 +97,12 @@ public interface HostDao extends GenericDao<HostVO, Long>, StateDao<Status, Stat
 
     List<HostVO> listByType(Type type);
 
+    /**
+     * Finds a host by ip address, excludes removed hosts.
+     *
+     * @param ip The ip address to match on
+     * @return One matched host
+     */
     HostVO findByIp(String ip);
 
     /**
--- a/engine/schema/src/main/java/com/cloud/host/dao/HostDaoImpl.java
+++ b/engine/schema/src/main/java/com/cloud/host/dao/HostDaoImpl.java
@@ -90,7 +90,7 @@ public class HostDaoImpl extends GenericDaoBase<HostVO, Long> implements HostDao
     protected SearchBuilder<HostVO> DcPrivateIpAddressSearch;
     protected SearchBuilder<HostVO> DcStorageIpAddressSearch;
     protected SearchBuilder<HostVO> PublicIpAddressSearch;
-    protected SearchBuilder<HostVO> AnyIpAddressSearch;
+    protected SearchBuilder<HostVO> UnremovedIpAddressSearch;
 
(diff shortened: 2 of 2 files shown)


Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

rhtyd:
@Spaceman1984 can you kick packaging and tests again; was purgeHost being called on a host already removed in DB? But how that does affect hosts that are already connected to a management server (i.e. host not removed in DB)?

Spaceman1984:
@blueorangutan package

Spaceman1984:
@rhtyd I didn't see purgeHost being called on a host already removed in my testing. What I observed, was the host being ignored because a check was done on an empty management server field.

Spaceman1984:
@blueorangutan test

Spaceman1984:
@nvazquez there is no problem with the certs map losing certs when hosts disconnect because the cert map is repopulated when a host connects again.

Spaceman1984:
Auto-renewal can only happen if the host has a valid certificate. As soon as the certificate becomes invalid the host can't communicate with the management server anymore and therefore wouldn't be able to get a new certificate. So before the validity period of the certificate would run out and the host is not able to communicate with the management server, auto-renewal must happen. - The auto-renewal process was failing and therefore certificates were not being renewed. If auto-renewal is fixed, then there wouldn't be a problem with disconnected hosts.

nvazquez:
Thanks @Spaceman1984, tested manually and looks good.

@rhtyd I noticed the internal active certificates map uses the host IP as the key, which is then used for querying in the DB for that IP. Is there any reason for not using the internal host ID or the UUID as the map key and using the host IP instead?

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

rhtyd:
@nvazquez yes it's because the code which adds the X509 cert to the hashmap (currently) uses the incoming client's IP address as the key. It may need checking or use host UUID/ID instead of IP address if that corrects the implementation.

rhtyd:
ping @nvazquez please review

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

Issue #4191, opened 2020-07-01
Wrong count in listUsageRecords response


olivierlemasle:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
API
~~~

##### CLOUDSTACK VERSION
~~~
4.13
4.14
~~~

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### SUMMARY
The `count` in listUsageRecords reponse is wrong; it is equal to the number of usage records in the returned set (a single page), and not the total amount of usage records for a given request.

This is a regression due to #3235 (specifically https://github.com/apache/cloudstack/pull/3235/commits/c0295ad20e19a2ddb0bfb7470220fbba17f41115).

##### STEPS TO REPRODUCE

Query the listUsageRecords API, and look at the field "count".

##### EXPECTED RESULTS

`count` is expected to be the total number of usage records with the specified filter.
This is the behaviour for the other CloudStack API commands; a recent fix (#3894) was done to align with this behaviour, following issue #3890.

With the wrong `count` value, an API user does not know if there's other pages in the result set.

Actually, the same bug was already fixed back in 2013 with commit 78b1ac4126aa1d.

##### ACTUAL RESULTS

~~~
$ cmk list usageRecords startdate=2020-07-01 enddate=2020-07-01 pagesize=10 page=1 | jq ".count"
10
~~~

#### Additional issue

In the PR causing the issue (#3235), usage records of IP addresses on shared networks flagged "Hide IP address usage" were filtered out _after_ the search. I guess that's why the `count` value was updated. However, it also mean that the number of returned results in a page is not always equal to the `pagesize` parameter.

E.g.: 
```
$ cmk list usageRecords startdate=2020-07-01 enddate=2020-07-01 type=3 pagesize=10 page=25 | jq ".count"
6
$ cmk list usageRecords startdate=2020-07-01 enddate=2020-07-01 type=3 pagesize=10 page=26 | jq ".count"
4
$ cmk list usageRecords startdate=2020-07-01 enddate=2020-07-01 type=3 pagesize=10 page=27

```

Here, 4 usage records were filtered on page 25.
It is even possible to have **all** usages records filtered out on a page, which means that the result will be:

```json
{"listusagerecordsresponse":{}}
```

A page with all usage records filtered out is indistinguishable from a non-existing page (e.g. page 27 above), so there's absolutely no way to know how many pages must be fetched.

olivierlemasle:
[...]

PR #3235, opened 2019-03-25
network: allow ability to specify if network's ipaddress usage need to be hidden

rhtyd:
Problem: Admins don’t want to charge for IP address usage on certain (shared) networks.
Root Cause: There is no flag or detail for admins to provide using UI or API when creating networks to specify if they want IP address usage of the network hidden.
Solution: A new boolean hideipaddressusage flag is added to the createNetwork API and a checkbox in the ‘Add guest network’ UI for the root admins to specify if they want the shared network’s IP address usage to be hidden in the listUsageRecords API response. The provided flag is saved as the ‘hideIpAddressUsage’ detail in the cloud.network_details table for the network. For existing (shared) networks, root admins can also specify the same boolean API parameter hideipaddressusage with the updateNetwork API request to configure the behaviour for an existing network. When the detail/flag is true, the IP address usage for the (shared) network is not exported in the listUsageRecords API response. The listNetworks API response will include the details of a network for root admin only. (note usage is still recorded in the usage database but not return by the listUsageRecords API)

The API flag works for any kind of network via the API, but the checkbox is only shown while creating shared networks in the UI.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

This adds a new API arg only accessible to admins to specify if they
want the network's ip address usage hidden. This then saves this setting
in the network_details table for a network, and the listUsageRecords
API response creator checks for an IP address if it needs to be exported
or skipped/hidden.

The setting is available only to root admin via the listNetworks API response (details key).
Root admin can also update existing networks by using updateNetwork API and passing `hideipaddressusage=true|false`

UI screenshot, that adds this checkbox only for admins creating a shared network:
51316923-21645080-1a7c-11e9-8c78-9769aa84b7c1

**note: it's possible for all other kinds of networks to have their ip address usages skipped as well, available via the API.

PR #4193, opened 2020-07-01
Fix usage record count

Pearl1594:
## Description
listUsageRecords returns the count equal to the number of objects returned in the page, rather than the total count
Addresses: https://github.com/apache/cloudstack/issues/4191

## Types of changes


Pearl1594:
@blueorangutan package

olivierlemasle:
Hi @Pearl1594 @rhtyd ,

Thank you for the quick fix.
If I'm not wrong, this will return the size of the _unfiltered_ list. If there's some usage records for IP addresses on shared networks with "Hide IP address usage", the displayed "count" will be greater than the real number of returned usage records. Would that not break pagination in tools like Primate (if Primate displayed usage records)?

The best would probably be to move the filtering from the ApiResponseHelper to the actual search in UsageServiceImpl, but that could be difficult and/or inefficient.

Pearl1594:
@blueorangutan package

server/src/main/java/com/cloud/usage/UsageServiceImpl.java
```diff
@@ -328,9 +333,20 @@ public boolean generateUsageRecords(GenerateUsageRecordsCmd cmd) {
...
-                        usageDbId = ip.getId();
+                    if (ip == null) {
+                        break;
+                    }
+                    Long networkId = ip.getAssociatedWithNetworkId();
+                    if (networkId == null) {
+                        networkId = ip.getSourceNetworkId();
+                    }
+                    NetworkDetailVO networkDetail = _networkDetailsDao.findDetail(networkId, Network.hideIpAddressUsage);
+                    if (networkDetail != null && networkDetail.getValue() != null && networkDetail.getValue().equals("true")) {
```
olivierlemasle, line 344: As this code is in a test `if (usageId != null)`, it will remove IP address usage only if the API command `listUsageRecords` was called with a specific `usageId` parameter.
For API calls like `command=listUsageRecords&startdate=xxx&enddate=xxx`, it will include all usage records, including for IP addresses on networks with "Hide usage".

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan test

rhtyd:
Can you @davidjumani review this? Thanx

Pearl1594:
@rhtyd since startdate and enddate are mandatory fields in the API , the usageRecords fetched from the database will be within the (date) range provided and further processing is then done on that list obtained

rhtyd:
Thanks for checking the mandatory params @Pearl1594, you're right with the argument I hadn't considered start/end date as mandatory parameters.

In that case, perhaps we can spend some time in future to optimise it, for example, get all the IPs given a list of IPs (unique usage IDs) and then join against the details table to check how many IPs need to be hidden, and run a quick loop to find number of counts we've to reduce if any; the current code may significantly impact the list API response which would require many DB calls.

olivierlemasle:
Hi @Pearl1594 @rhtyd What are the required steps before merging this?

Pearl1594:
@olivierlemasle It will probably require another round of testing before it can be merged.

RodrigoDLopez:
Hi @Pearl1594,
I ran some manual tests using a 4.13.1 environment

This one is before I update my environment with your fix.
```
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 | jq ".count, .usagerecord | length"
500
500
```

I don't know why the usage recods length is limited to 500, because in this environment I have more than one thousand usage records.
If I put paginate, you can see that my correct account is 1148
```
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 pagesize=400 page=1 | jq ".count, .usagerecord | length"
400
400
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 pagesize=400 page=2 | jq ".count, .usagerecord | length"
400
400
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 pagesize=400 page=3 | jq ".count, .usagerecord | length"
348
348
```
To me, this count is wrong, and without pagesize we have those limit of 500 records. I think that those limit cannot exist, since I asked unfiltered usage records, just startdate and enddate.

Now, with your fix... when I list usage records without pagesize, the API brings to me the correct count, but the length of `.usagerecors` is limited to 500
I think I got something wrong here. Does this limit exist due to performance issues?
```
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 | jq ".count, .usagerecord | length"
1148
500
```

with pagesize and page number we have this as response.
```
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 pagesize=400 page=1 | jq ".count, .usagerecord | length"
1148
400
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 pagesize=400 page=2 | jq ".count, .usagerecord | length"
1148
400
cloudmonkey -p acs listusagerecords startdate=2020-07-24 enddate=2020-07-24 pagesize=400 page=3 | jq ".count, .usagerecord | length"
1148
348
```

with your fix, the count is correct, but I think those limit of 500 cannot exist when the requested command is an unfiltered usage records
I didn't test with `hideIpAddressUsage` as @olivierlemasle because he has already pointed out this problem

olivierlemasle:
@RodrigoDLopez The limit to 500 usage records when `pagesize` is not defined is "normal".
There's a Global setting named "default.page.size" (description: "Default page size for API list* commands") which defaults to 500.

Pearl1594:
@RodrigoDLopez @olivierlemasle  Thanks for testing
@RodrigoDLopez as @olivierlemasle mentioned the upper limit is set to 500 when not specified
@olivierlemasle I'll look into what's messing up the count when pagination is done. Thanks!

olivierlemasle:
Hi @Pearl1594, did you get any chance to look into the pagination issue? Thank you!

Pearl1594:
Hi @olivierlemasle , I've been tied up with other stuff and haven't been able to give time to this. However, to get the right count when hideIpAddressUsage is true, can be a bit involved inorder to make it efficient. 
As a first level fix, we can get the pagination issue sorted, as it currently works, and for handling hiding IP usage records, we can have another PR?

olivierlemasle:
@Pearl1594 @rhtyd Is that what you propose?

- First, fix the `count` value with https://github.com/apache/cloudstack/pull/4193/commits/5b5a2e54dd2016481385d3b8558f23280d5047d1 AND return all usage records (incl. usages for IP addresses on networks flagged with "Hide IP address usage", which means reverting temporarily #3235)

- Then, in a separate PR, handle the hiding with a proper solution. I think the best solution (performance-wise) would require a DB schema update: 
  - Adding a flag (e.g. `hidden`) in table `cloud_usage.cloud_usage` ; `listUsageRecords` would then simply filter out these usages.
  - It would also require adding a new column in `cloud_usage.usage_ip_address` to know if the corresponding usage should be hidden. This new flag would be set in`UsageManagerImpl`.

@Pearl1594  Is that what you thought? Do you want me to do the PRs (both or only the second) or do you want to do it?

olivierlemasle:
@Pearl1594 I suggest you revert in this PR the change on `UsageServiceImpl.java`.
This PR will then do two things:
- Fix the `count` issue;
- Return all usages (even usages for IP addresses on networks flagged with "Hide IP address usage").
This would fix #4191.

On my side, I've just opened a new PR, #4327, which fixes the issue for hidden IP address usages. However, this PR includes a DB schema change, so I guess it could not be merged in all impacted branches (since 4.13).

@Pearl1594 @rhtyd Is that ok?

Pearl1594:
> @Pearl1594 @rhtyd Is that what you proposed?
> 
>     * First, fix the `count` value with [5b5a2e5](https://github.com/apache/cloudstack/commit/5b5a2e54dd2016481385d3b8558f23280d5047d1) AND return all usage records (incl. usages for IP addresses on networks flagged with "Hide IP address usage", which means reverting temporarily #3235)
> 
>     * Then, in a separate PR, handle the hiding with a proper solution. I think the best solution (performance-wise) would require a DB schema update:
>       
>       * Adding a flag (e.g. `hidden`) in table `cloud_usage.cloud_usage` ; `listUsageRecords` would then simply filter out these usages.
>       * It would also require adding a new column in `cloud_usage.usage_ip_address` to know if the corresponding usage should be hidden. This new flag would be set in`UsageManagerImpl`.
> 
> 
> @Pearl1594 Is that what you thought? Do you want me to do the PRs (both or only the second) or do you want to do it?

@olivierlemasle Apologies for the delay in response, but this was precisely what I meant.
Sounds like a good plan - to revert my PR to https://github.com/apache/cloudstack/commit/5b5a2e54dd2016481385d3b8558f23280d5047d1.
Thanks a lot for the PR!

rhtyd:
Ping me when you've reverted changes and kick tests @Pearl1594 thnx
I'll look at the new PR soon @olivierlemasle

olivierlemasle:
>  Sounds like a good plan - to revert my PR to 5b5a2e5.

@Pearl1594  Don't you want to also remove https://github.com/apache/cloudstack/blob/caefb0c9b5f3b7234291f13d272c15a928489b3f/server/src/main/java/com/cloud/api/ApiResponseHelper.java#L3437-L3441 (as you did in 69eeac38)?

Pearl1594:
@blueorangutan package

rhtyd:
@blueorangutan package

olivierlemasle (approved):

Pearl1594:
@blueorangutan package

olivierlemasle:
@Pearl1594 Any chance merging this PR?

DaanHoogland:
this is a 4.13 based change @rhtyd do we need to push this towards 4.16?

olivierlemasle:
IMHO, this should be backported in 4.13, 4.14 and recent releases, as it fixes a regression in 4.13.
The "long-term fix" #4327 should be included in the next release.

olivierlemasle:
@Pearl1594 Any new on this?

Pearl1594:
It's good to go @olivierlemasle

Pearl1594:
@blueorangutan test

Pearl1594:
@blueorangutan test

olivierlemasle:
@blueorangutan test

olivierlemasle:
Thanks @DaanHoogland

PR #4212, opened 2020-07-16
Migrate/Stop VMs with local storage when preparing host for maintenance

GabrielBrascher:
## Description

This PR adds a global settings parameter that configures the strategy for handling VMs in local storage when putting a host in maintenance.

Global settings name:  `host.maintenance.local.storage.strategy`
Description: _Defines the strategy towards VMs with volumes on local storage when putting a host in maintenance. The default strategy is 'Error', preventing maintenance in such a case. To migrate away VMs running on local storage choose 'Migrating' strategy. To stop VMs, choose 'Stopping' strategy._

## Types of changes

## How Has This Been Tested?

- Strategy **Error**:
1. put a host into maintenance
2. Maintenance fails and the host gets into ErrorInMaintenance state

- Strategy **Stopping**:
1. put a host into maintenance
2. VMs with volumes on shared storage are migrated, VMs in local storage are stopped, the host gets into Maintenance 

- Strategy **Migrating**:
1. put a host into maintenance
2. VMs with volumes on shared and on local storage are migrated, the host gets into Maintenance

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -5775,10 +5775,10 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
 
     }
 
-    private boolean isVMUsingLocalStorage(VMInstanceVO vm) {
+    public boolean isVMUsingLocalStorage(long vmId) {
         boolean usesLocalStorage = false;
 
-        List<VolumeVO> volumes = _volsDao.findByInstance(vm.getId());
+        List<VolumeVO> volumes = _volsDao.findByInstance(vmId);
         for (VolumeVO vol : volumes) {
             DiskOfferingVO diskOffering = _diskOfferingDao.findById(vol.getDiskOfferingId());
             if (diskOffering.isUseLocalStorage()) {
@@ -5838,7 +5838,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
             throw new InvalidParameterValueException("Unsupported Hypervisor Type for User VM migration, we support XenServer/VMware/KVM/Ovm/Hyperv/Ovm3 only");
         }
 
-        if (isVMUsingLocalStorage(vm)) {
+        if (isVMUsingLocalStorage(vm.getId())) {
             if (s_logger.isDebugEnabled()) {
                 s_logger.debug(vm + " is using Local Storage, cannot migrate this VM.");
             }
@@ -6262,6 +6262,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
 
         List<VolumeVO> vmVolumes = _volsDao.findUsableVolumesForInstance(vm.getId());
         Map<Long, Long> volToPoolObjectMap = new HashMap<Long, Long>();
+
         if (!isVMUsingLocalStorage(vm) && MapUtils.isEmpty(volumeToPool)
             && (destinationHost.getClusterId().equals(srcHost.getClusterId()) || isVmVolumesOnZoneWideStore(vm))){
             // If volumes do not have to be migrated
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -5775,10 +5775,10 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
 
     }
 
-    public boolean isVMUsingLocalStorage(long vmId) {
+    public boolean isVMUsingLocalStorage(VMInstanceVO vm) {
         boolean usesLocalStorage = false;
 
-        List<VolumeVO> volumes = _volsDao.findByInstance(vmId);
+        List<VolumeVO> volumes = _volsDao.findByInstance(vm.getId());
         for (VolumeVO vol : volumes) {
             DiskOfferingVO diskOffering = _diskOfferingDao.findById(vol.getDiskOfferingId());
             if (diskOffering.isUseLocalStorage()) {
@@ -5828,7 +5828,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         }
 
         if (!isOnSupportedHypevisorForMigration(vm)) {
-            s_logger.error(vm + " is not XenServer/VMware/KVM/Ovm/Hyperv, cannot migrate this VM form hypervisor type " + vm.getHypervisorType());
+            s_logger.error(vm + " is not XenServer/VMware/KVM/Ovm/Hyperv, cannot migrate this VM from hypervisor type " + vm.getHypervisorType());
             throw new InvalidParameterValueException("Unsupported Hypervisor Type for VM migration, we support XenServer/VMware/KVM/Ovm/Hyperv/Ovm3 only");
         }
 
@@ -5836,7 +5836,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
             throw new InvalidParameterValueException("Unsupported Hypervisor Type for User VM migration, we support XenServer/VMware/KVM/Ovm/Hyperv/Ovm3 only");
         }
 
-        if (isVMUsingLocalStorage(vm.getId())) {
+        if (isVMUsingLocalStorage(vm)) {
             s_logger.error(vm + " is using Local Storage, cannot migrate this VM.");
             throw new InvalidParameterValueException("Unsupported operation, VM uses Local storage, cannot migrate");
         }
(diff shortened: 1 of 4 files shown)


GabrielBrascher:
@blueorangutan package

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -5619,7 +5619,7 @@ public VirtualMachine migrateVirtualMachine(Long vmId, Host destinationHost) thr
             throw new InvalidParameterValueException("Unsupported Hypervisor Type for User VM migration, we support XenServer/VMware/KVM/Ovm/Hyperv/Ovm3 only");
         }
 
-        if (isVMUsingLocalStorage(vm)) {
+        if (isVMUsingLocalStorage(vm.getId())) {
             if (s_logger.isDebugEnabled()) {
                 s_logger.debug(vm + " is using Local Storage, cannot migrate this VM.");
```
wido, line 5624: Is this debug line still valid?

GabrielBrascher: That is a good point. It depends on the strategy configured at the ResourceService, I will take a look at the best approaches to adapt it.

GabrielBrascher: @wido I did check it again and noticed that the log message as it is makes sense.

The `migrateVirtualMachine` method does not work with volume migration. When a VM with volumes, e.g. when migrating VMs with local storage, is migrated it is used `migrateWithStorage` method. Therefore, we need to keep it that way.

GabrielBrascher: @wido I think that I answered the _wrong_ question on previous comments ...

Regarding the debug line, personally, I would remove it. The idea behind it would be to avoid costly String operations and therefore optimize CloudStack processing. However, I think that there is a little optimization, and most of the cases CloudStack already runs in Debug mode.

I think that I saw @DaanHoogland and @rafaelweingartner discussing about it on the past.

DaanHoogland: 1. As for the message, looks like an extra check for the strategy is in order, but the flow might already guarantee it's validity.
2. As for the second interpretation of the question, as a matter of principle it is always good to check the logging level as parameters are always evaluated before passing and you never know how often the code is being executed.
3. As for the level, this is an obvious mis-use of debug. An error or at least warning or info situation is occurring here; the user might expect a migration but there is something impairing it. This is not a message for a developer, hence not a debug level message.

my €0,03 ;)

GabrielBrascher: @DaanHoogland your €0,03 is very appreciated ;)
I will change the log level from debug to error (maybe warn, but it looks more like an error).

GabrielBrascher: @DaanHoogland I changed the log level from debug to error on those cases.

GabrielBrascher:
@DaanHoogland I did not answer your Marvin question, sorry for the delay.

I added Unit tests but, to be honest, I had no plans on creating Marvin tests for this. I can take a look at it, though.

GabrielBrascher:
Ping for review @rhtyd, @nvazquez, @andrijapanicsb, @borisstoyanov, @DaanHoogland, @svenvogel, @weizhouapache, @RodrigoDLopez, @svenvogel, @kiwiflyer, and others :-)

andrijapanicsb:
@GabrielBrascher what was the behaviour so far? I believe we were stopping all VMs, right?

I would be happy to see the this new feature to implement the old behaviour as the defaul behaviour - to keep backward compatibility, if that makes sense

GabrielBrascher:
@andrijapanicsb normally VMs are (live) migrated when a host is put into Maintenance; however, if the host has VMs with local storage the host is placed in the "`ErrorInMaintenance`" state; the ADMIN then needs to manually stop/migrate these VMs and then re-try the Maintenance.

Following the same concern that you raised, this implementation keeps the current behavior by default, therefore does not cause backward compatibility to any deployed Zone.

To change the behavior the Root Admin needs to configure the global settings parameter `host.maintenance.local.storage.strategy` on its own risk.

This parameter defines the strategy towards VMs with volumes on local storage when putting a host in maintenance.

1. The default strategy is 'Error', preventing maintenance in such a case.
2. To migrate away VMs running on local storage it is necessary to set `Migrating` strategy.
3. To stop VMs with local storage instead of placing the host into `ErrorInMaintenance`, it is necessary to change the strategy to  `Stopping`.

andrijapanicsb:
that sounds good @GabrielBrascher - thx for confirming.
Logic/behaviour is LGTM - but I haven't tested it.
How does it behave with other hypervisors? I trust you've tested KVM only?

GabrielBrascher:
@andrijapanicsb I tested with KVM only indeed.

I just re-checked the maintenance behavior with VMs running on shared storage. The zone holds one of the recent commits at the master (not older than one week).

Tested on KVM node with NFS as primary storage:
1. the host has only VMs running with Root disk placed on shared storage
2. call `prepareHostForMaintenance` API command
3. all VMs are live migrated to another host
4. host gets into Maintenance state

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

wido (approved):
LGTM based on code and discussions with Gabriel

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -5609,9 +5609,7 @@ public VirtualMachine migrateVirtualMachine(Long vmId, Host destinationHost) thr
         }
 
         if (!isOnSupportedHypevisorForMigration(vm)) {
-            if (s_logger.isDebugEnabled()) {
-                s_logger.debug(vm + " is not XenServer/VMware/KVM/Ovm/Hyperv, cannot migrate this VM form hypervisor type " + vm.getHypervisorType());
-            }
+            s_logger.error(vm + " is not XenServer/VMware/KVM/Ovm/Hyperv, cannot migrate this VM form hypervisor type " + vm.getHypervisorType());
```
DaanHoogland, line 5612: nit picking: the target host can be specified and small spello
```suggestion
            s_logger.error(vm + " is not XenServer/VMware/KVM/Ovm/Hyperv, cannot migrate this VM from hypervisor type " + vm.getHypervisorType() + " to hypervisor type " + destinationHost.getHypervisorType());
```

DaanHoogland (approved):

GabrielBrascher:
@DaanHoogland I added your contribution, that log message is better indeed. Also, I fixed the conflicts with the master.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@DaanHoogland quick update here: code stills the same, however, I rebased it to the current master.

Ping for reviewers @andrijapanicsb @rhtyd @RodrigoDLopez @ravening @weizhouapache and others.

GabrielBrascher:
@blueorangutan package

RodrigoDLopez (approved):
LGTM based on code review.
I'm +1 with this Enhancement

shwstppr:
@GabrielBrascher conflict here

GabrielBrascher:
Code has been updated adding the option to Stop VM with force stop or a simple stop. Strategies are the following:

1. The default strategy is **'Error'**, preventing maintenance in such a case
2. Choose **'Migration'** strategy to migrate away VMs running on local storage
3. To stop VMs, choose **'Stop'** strategy : `_haMgr.scheduleStop(vm, hostId, WorkType.Stop);`
4. To force-stop VMs, choose **'ForceStop'** strategy: `_haMgr.scheduleStop(vm, hostId, WorkType.ForceStop);`

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
Running tests. The strategy added for Stop is failing; StopForce works fine.

DaanHoogland:
@blueorangutan test keepEnv

DaanHoogland:
@GabrielBrascher sounds like you are not ready to have this merged, do you?

GabrielBrascher:
@DaanHoogland I am still checking the python changes on the system VM. I should be soon giving a definitive answer (good or "bad"). Good to see that all tests are green.

The failing tests (test_01_migrate_VM_and_root_volume and test_02_migrate_VM_with_two_data_disks) are failing on other PRs as well.

GabrielBrascher:
Code has been updated reverting the change to allow choosing between Stop or ForceStop. It turns out that the implementation got a bit tricky when HA management calls a VM `Stop` at the maintenance workflow. For now, keeping only VM `ForceStop`.

Fitting the "graceful" Stop into the current maintenance workflow might be a good enhancement for another PR, but for now, keeping the strategy feature as it is:
1. The default strategy is 'Error', preventing maintenance in such a case
2. Choose 'Migration' strategy to migrate away VMs running on local storage
3. To force-stop VMs, choose the 'ForceStop' strategy

Manual tests are looking good.

GabrielBrascher:
@blueorangutan package

rhtyd:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
Considering the this implementation only touches at `doMaintaint`and `prepareForMaintenance` methods I think that the test failures at `test_01_migrate_VM_and_root_volume` and `test_02_migrate_VM_with_two_data_disks` are not related to any of the added commits in this PR.

Thanks to all the reviewers @DaanHoogland @andrijapanicsb @rhtyd @wido @RodrigoDLopez; this one looks ready for merge. Are there any concerns that are missing my attention?

PR #4228, opened 2020-07-28
Dont add host back after agent service restart

### LINKED ISSUE #4218: Removed kvm host is added back to cloudstack when restart cloudstack-agent on it
When we remove a KVM host from cloudstack, and restart cloudstack-agent on it, the host is added back to cloudstack.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
KVM agent
~~~

##### CLOUDSTACK VERSION

~~~
4.14
~~~

##### CONFIGURATION

basic/advanced networking

##### OS / ENVIRONMENT

KVM

##### STEPS TO REPRODUCE

~~~
1. add a kvm host  to clodustack
2. put host to maintenance in cloudstack
3. remove host from cloudstack
4. restart cloudstack-agent on the host

~~~

##### EXPECTED RESULTS

~~~

5. host is not added back to cloudstack
~~~

##### ACTUAL RESULTS

~~~
5. host is added back to cloudstack
~~~

ravening:
## Description

Fixes #4218 

If a host is removed from cloudstack, it will be
added back if we restart the agent service on
the host. It should not be added back if manualy
removed.

Provide a global setting "add.host.on.service.restart" with
default value of true. If set to false, the zoneid,
clusterid and podid in agent.properties will be set to
null so that when service is restarted, the agent is not
added back

If true these values will not be set to null so that it
can be still added back on service restart

## Types of changes

## How Has This Been Tested?
1. Add a host from ui. It will be successfully added
2. Now remove the host
3. Set the global setting `add.host.on.service.restart` to false
3. ssh to host and restart cloudstack-agent service
4. The host is not added back
5. Try to add host from ui again. It will be added

### DIFF
--- a/agent/src/main/java/com/cloud/agent/Agent.java
+++ b/agent/src/main/java/com/cloud/agent/Agent.java
@@ -420,6 +420,15 @@ public class Agent implements HandlerFactory, IAgentControl {
         }
     }
 
+    protected void cleanupAgentZoneProperties() {
+        // Unset zone, cluster and pod values so that host is not added back
+        // when service is restarted. This will be set to proper values
+        // when host is added back
+        _shell.setPersistentProperty(null, "zone", "");
+        _shell.setPersistentProperty(null, "cluster", "");
+        _shell.setPersistentProperty(null, "pod", "");
+    }
+
     public synchronized void lockStartupTask(final Link link) {
         _startup = new StartupTask(link);
         _timer.schedule(_startup, _startupWait);
@@ -603,6 +612,9 @@ public class Agent implements HandlerFactory, IAgentControl {
                         final ShutdownCommand shutdown = (ShutdownCommand)cmd;
                         s_logger.debug("Received shutdownCommand, due to: " + shutdown.getReason());
                         cancelTasks();
+                        if (shutdown.isRemoveHost()) {
+                            cleanupAgentZoneProperties();
+                        }
                         _reconnectAllowed = false;
                         answer = new Answer(cmd, true, null);
                     } else if (cmd instanceof ReadyCommand && ((ReadyCommand)cmd).getDetails() != null) {
--- a/agent/src/main/java/com/cloud/agent/Agent.java
+++ b/agent/src/main/java/com/cloud/agent/Agent.java
@@ -420,10 +420,14 @@ public class Agent implements HandlerFactory, IAgentControl {
         }
     }
 
+    /**
+     * Cleanup agent zone properties.
+     *
+     * Unset zone, cluster and pod values so that host is not added back
+     * when service is restarted. This will be set to proper values
+     * when host is added back
+     */
     protected void cleanupAgentZoneProperties() {
-        // Unset zone, cluster and pod values so that host is not added back
-        // when service is restarted. This will be set to proper values
-        // when host is added back
         _shell.setPersistentProperty(null, "zone", "");
         _shell.setPersistentProperty(null, "cluster", "");
         _shell.setPersistentProperty(null, "pod", "");
--- a/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
+++ b/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
@@ -410,7 +410,7 @@ public class ConfigurationManagerImpl extends ManagerBase implements Configurati
     public final static ConfigKey<Long> IOPS_MAX_WRITE_LENGTH = new ConfigKey<Long>(Long.class, "vm.disk.iops.maximum.write.length", "Advanced", "0",
             "Maximum IOPS write burst duration (seconds). If '0' (zero) then does not check for maximum burst length.", true, ConfigKey.Scope.Global, null);
 
-    public static final ConfigKey<Boolean> ADD_HOST_ON_SERVICE_RESTART = new ConfigKey<Boolean>(Boolean.class, "add.host.on.service.restart", "Advanced", "true",
+    public static final ConfigKey<Boolean> ADD_HOST_ON_SERVICE_RESTART_KVM = new ConfigKey<Boolean>(Boolean.class, "add.host.on.service.restart.kvm", "Advanced", "true",
             "Indicates whether the host will be added back to cloudstack after restarting agent service on host. If false it wont be added back even after service restart",
             true, ConfigKey.Scope.Global, null);
 
@@ -6383,6 +6383,6 @@ public class ConfigurationManagerImpl extends ManagerBase implements Configurati
     @Override
     public ConfigKey<?>[] getConfigKeys() {
         return new ConfigKey<?>[] {SystemVMUseLocalStorage, IOPS_MAX_READ_LENGTH, IOPS_MAX_WRITE_LENGTH,
-                BYTES_MAX_READ_LENGTH, BYTES_MAX_WRITE_LENGTH, ADD_HOST_ON_SERVICE_RESTART};
+                BYTES_MAX_READ_LENGTH, BYTES_MAX_WRITE_LENGTH, ADD_HOST_ON_SERVICE_RESTART_KVM};
     }
 }
--- a/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
+++ b/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
@@ -410,7 +410,7 @@ public class ConfigurationManagerImpl extends ManagerBase implements Configurati
     public final static ConfigKey<Long> IOPS_MAX_WRITE_LENGTH = new ConfigKey<Long>(Long.class, "vm.disk.iops.maximum.write.length", "Advanced", "0",
             "Maximum IOPS write burst duration (seconds). If '0' (zero) then does not check for maximum burst length.", true, ConfigKey.Scope.Global, null);
 
-    public static final ConfigKey<Boolean> AddHostOnServiceRestart = new ConfigKey<Boolean>(Boolean.class, "add.host.on.service.restart", "Advanced", "true",
+    public static final ConfigKey<Boolean> ADD_HOST_ON_SERVICE_RESTART = new ConfigKey<Boolean>(Boolean.class, "add.host.on.service.restart", "Advanced", "true",
             "Indicates whether the host will be added back to cloudstack after restarting agent service on host. If false it wont be added back even after service restart",
             true, ConfigKey.Scope.Global, null);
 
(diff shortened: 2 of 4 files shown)


server/src/main/java/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
```diff
@@ -348,11 +362,19 @@ private void setupAgentSecurity(final Connection sshConnection, final String age
             _hostDao.saveDetails(connectedHost);
             return resources;
         } catch (DiscoveredWithErrorException e) {
+            s_logger.error("Exception: "+ e.getMessage());
```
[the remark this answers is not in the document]
ravening: I have no idea why it's there in first place. So I'm displaying message whatever we catch

server/src/main/java/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
```diff
@@ -348,11 +362,19 @@ private void setupAgentSecurity(final Connection sshConnection, final String age
...
             throw e;
         } catch (Exception e) {
             String msg = " can't setup agent, due to " + e.toString() + " - " + e.getMessage();
             s_logger.warn(msg);
         } finally {
+            // set guid to null once host is added
```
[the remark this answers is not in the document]
ravening: Initialguid will always beer null since I'm fetching the value for a removed host. For non removed hosts it will be non null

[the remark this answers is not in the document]
DaanHoogland: exactly. it is not necessarily null. I think you'd better remove this comment, it is confusing.

server/src/main/java/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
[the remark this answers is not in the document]
DaanHoogland: ok, than let's at least make the text describing our course of action:
```suggestion
            s_logger.error("DiscoveredWithErrorException caught and rethrowing, message; " + e.getMessage());
```

rhtyd:
@ravening this will break the host-add feature for a lot of users who use the functionality to add a KVM host out-of-band without using the addHost API.

Instead a middle-ground fix could be to remove a host's certificates when a KVM host is removed (all the /etc/cloudstack/agent/cloud.* files) which will stop KVM agent connections to be accepted by the management server for most deployments (i.e. CA plugin and auth strictness enabled).

server/src/main/java/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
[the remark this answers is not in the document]
ravening: @DaanHoogland this code is executed only for the newly added hosts. When the host is removed previously, the guid is set to null. So the initial value of `guid` will always be `null` in this case.
Before adding the host, I set the guid to a temporary value and at the end I set it back to null which is the original value.
I hope to have cleared your confusion

ravening:
> @ravening this will break the host-add feature for a lot of users who use the functionality to add a KVM host out-of-band without using the addHost API.
> 
> Instead a middle-ground fix could be to remove a host's certificates when a KVM host is removed (all the /etc/cloudstack/agent/cloud.* files) which will stop KVM agent connections to be accepted by the management server for most deployments (i.e. CA plugin and auth strictness enabled).

@rhtyd this will work only if CA plugin and auth strictness is enabled but what if we have not configured both in some setup. In that case, the host will still be added

rhtyd:
Well @ravening fixing it otherwise would break some people's way of deploying/adding KVM host. Perhaps there's an easy way to reset the host (domain/IP) and zone/pod/cluster settings in the agent.properties (i.e. reset or remove them) when a host is removed?

ravening:
> Well @ravening fixing it otherwise would break some people's way of deploying/adding KVM host. Perhaps there's an easy way to reset the host (domain/IP) and zone/pod/cluster settings in the agent.properties (i.e. reset or remove them) when a host is removed?

I'm thinking of setting guid to null in agent.properties when host is removed and in connection handler we search for a host based on guid whenever we get a new connection request. Any other suggestions @rhtyd

rhtyd:
@ravening I think resetting the host, zone, cluster, etc to default (or remove) should be enough. Post agent restart the agent will not know where to connect.

DaanHoogland:
@ravening @rhtyd how about hiding this behaviour behind a global setting `auto-add-hosts`? (default would have to be true for backwards compatibility)

ravening:
> @ravening @rhtyd how about hiding this behaviour behind a global setting `auto-add-hosts`? (default would have to be true for backwards compatibility)

Lets say, I add a global setting `add.host.on.service.restart` with default value of true
When we delete the host, if the value is false then we set zone/cluster/pod value to null else retain as it is

Since default value is true, we dont change any value and host will be added back if service is restarted

ravening:
@rhtyd @DaanHoogland please review again

server/src/main/java/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
```diff
@@ -474,7 +477,8 @@ public DeleteHostAnswer deleteHost(HostVO host, boolean isForced, boolean isForc
 
         _resourceMgr.deleteRoutingHost(host, isForced, isForceDeleteStorage);
         try {
-            ShutdownCommand cmd = new ShutdownCommand(ShutdownCommand.DeleteHost, null);
+            ShutdownCommand cmd = AddHostOnServiceRestart.value() ? new ShutdownCommand(ShutdownCommand.DeleteHost, null, false) :
+                    new ShutdownCommand(ShutdownCommand.DeleteHost, "Cleaning up zone/pod/cluster details for host", true);
```
DaanHoogland, line 481: general jist looks good, i'd rather have the tristate operator in the parameter passing then the whole contructor invocation ? note the ?)
```suggestion
            ShutdownCommand cmd = new ShutdownCommand(ShutdownCommand.DeleteHost, null, AddHostOnServiceRestart.value() ? false : true);
```

DaanHoogland:
@wido @GabrielBrascher I think this will impact you guys most. Can you review?

wido:
I have never run into this situation to be honest. But the change seems sane to me.

server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
```diff
@@ -410,6 +410,10 @@
     public final static ConfigKey<Long> IOPS_MAX_WRITE_LENGTH = new ConfigKey<Long>(Long.class, "vm.disk.iops.maximum.write.length", "Advanced", "0",
             "Maximum IOPS write burst duration (seconds). If '0' (zero) then does not check for maximum burst length.", true, ConfigKey.Scope.Global, null);
 
+    public static final ConfigKey<Boolean> AddHostOnServiceRestart = new ConfigKey<Boolean>(Boolean.class, "add.host.on.service.restart", "Advanced", "true",
```
GabrielBrascher, line 413: Considering that this is defined as a `static final`, I would recommend keeping the code convention such as described on [Oracle Java Naming Conventions](https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html).

In this case `AddHostOnServiceRestart` should be changed to something as `ADD_HOST_ON_SERVICE_RESTART`.

agent/src/main/java/com/cloud/agent/Agent.java
```diff
@@ -419,6 +419,15 @@ protected void cancelTasks() {
         }
     }
 
+    protected void cleanupAgentZoneProperties() {
+        // Unset zone, cluster and pod values so that host is not added back
+        // when service is restarted. This will be set to proper values
+        // when host is added back
```
GabrielBrascher, line 425: These commented lines could be a nice Javadoc for the method; what do you think of moving them to a Javadoc block?

server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
ravening: @GabrielBrascher I made the necessary changes. Please review again

server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
```diff
@@ -410,6 +410,10 @@
     public final static ConfigKey<Long> IOPS_MAX_WRITE_LENGTH = new ConfigKey<Long>(Long.class, "vm.disk.iops.maximum.write.length", "Advanced", "0",
             "Maximum IOPS write burst duration (seconds). If '0' (zero) then does not check for maximum burst length.", true, ConfigKey.Scope.Global, null);
 
+    public static final ConfigKey<Boolean> ADD_HOST_ON_SERVICE_RESTART = new ConfigKey<Boolean>(Boolean.class, "add.host.on.service.restart", "Advanced", "true",
```
rhtyd, line 413: If this is KVM specific can we add kvm in the global setting name @ravening ?

ravening: @rhtyd i don't think this is KVM specific. The code path is same for all HV's I guess. Didn't test it on other Hv types.

rhtyd: The agent only runs on KVM host, for the other two major hypervisors (VMware and XenServer) we don't run an agent on them.

ravening: Didnt know that :P will add KVM in the name

ravening: @rhtyd made the change

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland:
@GabrielBrascher @wido @rhtyd are we 🆗 with this?

rhtyd (approved):
LGTM, I've not manually tested this

rhtyd:
@ravening @weizhouapache can you share your manual test results if any?

ravening:
@rhtyd Sure... I will share the results by tomorrow

rhtyd:
ping @ravening any update? thanks.

ravening:
> ping @ravening any update? thanks.

@rhtyd apart from logs what else can I share? Manual testing involves

Add a host to cloudstack
Put it in maintenance mode
Remove it
Restart agent service on it
The host should not be added back

DaanHoogland (approved):

PR #4231, opened 2020-07-30
kvm/ceph: Only if a port number has been specified define in the XML

wido:
## Description
Ceph used to use port 6789 (no need to specify it), but with the messenger v2
from Ceph it switched to port 3300 while 6789 still works.

librados/librbd/libvirt will automatically figure out the ports to use if none is
specified.

Therefor there is no need for CloudStack to explicitely define the port in the XML
passed to Libvirt or Qemu.

Leave blank if no port number has been defined by the user.

## Types of changes

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMPhysicalDisk.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMPhysicalDisk.java
@@ -28,7 +28,7 @@ public class KVMPhysicalDisk {
 
         rbdOpts = "rbd:" + image;
         rbdOpts += ":mon_host=" + monHost;
-        if (monPort != 6789) {
+        if (monPort > 0) {
             rbdOpts += "\\\\:" + monPort;
         }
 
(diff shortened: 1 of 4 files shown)


rhtyd:
@wido reminder - can you help do the next https://github.com/ceph/rados-java release? It would be great if we can do it soon/before 4.15 to support the latest ceph release. Thanks.
@blueorangutan package

wido:
@rhtyd I actually have rados-java 0.6.0 ready. Just need to have somebody upload it to maven central.

rhtyd:
@blueorangutan test

DaanHoogland:
seems ready to merge, right @wido @rhtyd ?

wido:
For me it is

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMPhysicalDisk.java
```diff
@@ -28,7 +28,7 @@ public static String RBDStringBuilder(String monHost, int monPort, String authUs
 
         rbdOpts = "rbd:" + image;
         rbdOpts += ":mon_host=" + monHost;
-        if (monPort != 6789) {
+        if (monPort > 0) {
             rbdOpts += "\\\\:" + monPort;
```
DaanHoogland, line 32: this double escape seems to byte us now ;P during release of 4.15

DaanHoogland: @wido @weizhouapache can you comment on if this is the only issue?

PR #3730, opened 2019-11-29
Update SystemVM scripts from python2 to python3

GabrielBrascher:
## Description

- Considering that Python 2.7 will not be maintained in 2020 (will be retired on 1st January 2020) next VR template should have Python3 installed to support python3 scripts; additionally, VR's Python2 codebase should be migrated to Python3.

- This PR covers partially issue #3195.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

PR #4243, opened 2020-08-04
Update SystemVM debian iso from 10.4.0 to 10.5.0

GabrielBrascher:
## Description
Debian 10.4.0 is not available, therefore SystemVM templates will fail to build with a missing iso.

Fixed updating with the Debian 10.5.0 iso located at https://cdimage.debian.org/debian-cd/current/amd64/iso-cd/

## Types of changes

## How Has This Been Tested?
Executed command to build the template:
```
$ sudo bash build.sh systemvmtemplate
```

Check that the template was successfully generated with script `tools/appliance/build.sh`


GabrielBrascher:
@blueorangutan package

rhtyd:
Needs separate systemvm build if necessary
cc @DaanHoogland

rhtyd:
Checked this, just URL/checksum change. Full smoketests not necessary, @GabrielBrascher have you found any reason we should build a new systemvmtemplate towards 4.15 release? Or keep using the build from June: http://download.cloudstack.org/systemvm/4.15/ cc @wido @DaanHoogland

DaanHoogland:
@GabrielBrascher change gtm.
Do we expect debian to just remove older 10 releases when they update? This may lead to failures after we release as well. At least for anybody building their own systemVMs. What if we release and the day after they upgrade to 10.6?

GabrielBrascher:
@rhtyd this happened while I was building a System VM template for testing Python3 development at PR #3730.
Depending on how #3730 progress we might or might not need a new template.

GabrielBrascher:
@DaanHoogland I was thinking the same. Do we need to constantly update that any time that there is another Debian 10.X?

That might be interesting thinking about new developers playing around with the system VM template. However, it might require some manual effort to keep it up to date.

GabrielBrascher:
@DaanHoogland that is a good question. What do you think @rhtyd? 

I can close this PR if you guys think it should be addressed inside a PR when the respective requires a new system VM template.

The downside is that it could catch some developer that is not used with the building template process; but on the other hand, we don't need to keep an eye looking for changes on the Debian 10.X.Y.

Alternatives that I see:
1. Update iso just if a new template is necessary
2. Create some fancy way to automatically update
3. Update at any time if someone struggle with building and sees it (no automation checking from time to time the ISO)

DaanHoogland:
if you need it you need it and i am fine with it. lgtm in that respect. i do not like the flimsy procedure we use, but i can live with it.

rhtyd:
The systemvm py scripts are patched via systemvm.iso, so we don't need a new template unless there's something in the 10.5 update. I'll merge this as it breaks build and review/compare changes against the June build.

PR #4249, opened 2020-08-06
Host SSVM Debian ISO on download.cloudstack.org

wido:
## Description
This way we do not depend on Debian keeping the ISO online or changing
URLs which can break our build system(s).

See also: #4243


rhtyd:
I think we can use the archive link https://cdimage.debian.org/mirror/cdimage/archive/ for older ISOs, but LGTM

PR #4255, opened 2020-08-10
Prevent null pointer on listPublicIpAddress cmd

RodrigoDLopez:
## Description

I reproduced this in a production environment that had been upgraded from 4.9 to 4.11 and then to 4.13.  

If the environment has `data centers` and `VLANs` removed, sends an API request with the command `listPublicIpAddress` and allocatedOnly=false we can see this error:
```
2020-07-29 08:15:16,224 ERROR [c.c.a.ApiServer] (qtp504527234-1307:ctx-d9b6eb3f ctx-a0add444) (logid:b94eb35a) unhandled exception executing api command: [Ljava.lang.String;@6177c5d4
java.lang.NullPointerException
```

This occurs because `ManagementService.searchForIPAddresses` returns IP addresses from those deleted `data centers`. And then we have a Null Pointer exception when building the `createIPAddressResponse`.  
This PR prevents this null pointer exception by doing an `INNER_JOIN` with `data_center` table were `removed` is null.

## Types of changes

## How Has This Been Tested?

To test this, I run the `listPublicIpAddrescomand` with `allocatedOnly = false` using `cloudmonkey` to make the request
**Before the fix:**
```
cloudmonkey listpublicipaddresses allocatedonly=false
Error: (HTTP 530, error code 9999) <nil>
```

**After the fix:**
```
cloudmonkey listpublicipaddresses allocatedonly=false
{
  "count": 755,
  "publicipaddress": [...]
}
```

### DIFF
--- a/server/src/main/java/com/cloud/server/ManagementServerImpl.java
+++ b/server/src/main/java/com/cloud/server/ManagementServerImpl.java
@@ -1983,11 +1983,8 @@ public class ManagementServerImpl extends ManagerBase implements ManagementServe
 
         final SearchBuilder<VlanVO> vlanSearch = _vlanDao.createSearchBuilder();
         vlanSearch.and("vlanType", vlanSearch.entity().getVlanType(), SearchCriteria.Op.EQ);
+        vlanSearch.and("removed", vlanSearch.entity().getRemoved(), SearchCriteria.Op.NULL);
         sb.join("vlanSearch", vlanSearch, sb.entity().getVlanId(), vlanSearch.entity().getId(), JoinBuilder.JoinType.INNER);
-
-        SearchBuilder<DataCenterVO> zoneSearchBuilder = _dcDao.createSearchBuilder();
-        zoneSearchBuilder.and("removed", zoneSearchBuilder.entity().getRemoved(), SearchCriteria.Op.NULL);
-        sb.join("removed_zone_filter", zoneSearchBuilder, sb.entity().getDataCenterId(), zoneSearchBuilder.entity().getId(), JoinBuilder.JoinType.INNER);
 
         if (isAllocated != null && isAllocated == true) {
             sb.and("allocated", sb.entity().getAllocatedTime(), SearchCriteria.Op.NNULL);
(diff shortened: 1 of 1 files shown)


RodrigoDLopez:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
redundant vpc would not have much ado with listIps but to be sure
@blueorangutan test

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -1985,6 +1985,10 @@ private boolean hasSuitablePoolsForVolume(final VolumeVO volume, final Host host
         vlanSearch.and("vlanType", vlanSearch.entity().getVlanType(), SearchCriteria.Op.EQ);
         sb.join("vlanSearch", vlanSearch, sb.entity().getVlanId(), vlanSearch.entity().getId(), JoinBuilder.JoinType.INNER);
 
+        SearchBuilder<DataCenterVO> zoneSearchBuilder = _dcDao.createSearchBuilder();
```
GabrielBrascher, line 1988: I think that it would be better to implement this via the VLAN Search Builder instead of adding the proposed search builder. Here are a few points of why I think it would be better:

1. All removed zones should have its networks removed anyway, therefore VLANs of a removed zone are marked as removed as well;
2. this could also prevent issues of selecting IP addresses of removed VLANs on an existing Zone;
3. and, at the end, it reduces costs of one extra join.

Currently lines 1984 - 1986:
```
final SearchBuilder<VlanVO> vlanSearch = _vlanDao.createSearchBuilder();
vlanSearch.and("vlanType", vlanSearch.entity().getVlanType(), SearchCriteria.Op.EQ);
sb.join("vlanSearch", vlanSearch, sb.entity().getVlanId(), vlanSearch.entity().getId(), JoinBuilder.JoinType.INNER);
```

Changing current VLAN search builder adding a verification "_WHERE removed IS NULL_":
```
final SearchBuilder<VlanVO> vlanSearch = _vlanDao.createSearchBuilder();
vlanSearch.and("vlanType", vlanSearch.entity().getVlanType(), SearchCriteria.Op.EQ);
vlanSearch.and("removed", vlanSearch.entity().getRemoved(), SearchCriteria.Op.NULL);
sb.join("vlanSearch", vlanSearch, sb.entity().getVlanId(), vlanSearch.entity().getId(), JoinBuilder.JoinType.INNER);SearchCriteria.Op.NULL);
```

RodrigoDLopez: I did not see that. thanks for the hint
work like a charm adding this removed filter for Vlan Search.

GabrielBrascher: That's great! :slightly_smiling_face:

PR #4256, opened 2020-08-10
Fix comparison using nullable objects

RodrigoDLopez:
## Description

This PR fix some wrongs comparison using nullable objects. Preventing null pointer exception

## Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
@@ -2199,7 +2199,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
         // OfflineVmwareMigration: check storage tags on disk(offering)s in comparison to destination storage pool
         // OfflineVmwareMigration: if no match return a proper error now
         DiskOfferingVO diskOffering = _diskOfferingDao.findById(vol.getDiskOfferingId());
-        if (diskOffering.equals(null)) {
+        if (diskOffering == null) {
             throw new CloudRuntimeException("volume '" + vol.getUuid() + "', has no diskoffering. Migration target cannot be checked.");
         }
         if (!doesTargetStorageSupportDiskOffering(destPool, diskOffering)) {


utils/src/test/java/com/cloud/utils/TernaryTest.java
```diff
@@ -28,7 +28,7 @@ public void testEquals() {
         Assert.assertEquals(new Ternary<String, String, String>("a", "b", "c"), new Ternary<String, String, String>("a", "b", "c"));
         Assert.assertFalse(new Ternary<String, String, String>("a", "b", "c").equals(new Ternary<String, String, String>("a", "b", "d")));
         Assert.assertFalse(new Ternary<String, String, String>("a", "b", "c").equals(""));
-        Assert.assertFalse(new Ternary<String, String, String>("a", "b", "c").equals(null));
```
rafaelweingartner, line 31: You do not need to remove the `equals` here. In this very specific use case, there is no null pointer that can happen here.

engine/orchestration/src/test/java/com/cloud/agent/manager/ConnectedAgentAttacheTest.java
```diff
@@ -44,7 +44,7 @@ public void testEqualsFalseNull() throws Exception {
 
         ConnectedAgentAttache agentAttache1 = new ConnectedAgentAttache(null, 0, null, link, false);
 
-        assertFalse(agentAttache1.equals(null));
```
rafaelweingartner, line 47: The same here. I do not understand why people coded this check, but it does not cause a null pointer. Moreover, if you change the equals to `==` here, you are changing the test case.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -2199,7 +2199,7 @@ public Volume migrateVolume(MigrateVolumeCmd cmd) {
         // OfflineVmwareMigration: check storage tags on disk(offering)s in comparison to destination storage pool
         // OfflineVmwareMigration: if no match return a proper error now
         DiskOfferingVO diskOffering = _diskOfferingDao.findById(vol.getDiskOfferingId());
-        if (diskOffering.equals(null)) {
+        if (diskOffering == null) {
```
rafaelweingartner, line 2202: This is the only one that makes sense to change.

RodrigoDLopez: thanks for the hint, I did not know about this other specific use case.

utils/src/test/java/com/cloud/utils/TernaryTest.java
RodrigoDLopez: thanks for the hint, I did not know about this other specific use case.

RodrigoDLopez:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

PR #4266, opened 2020-08-14
Adding os type id to the usage record response for virtual machines

ggoodrich-ipp:
## Description

Adding the Os Type Id to the CloudStack Api response for listUsageRecords for running and allocated VMs. When attempting to determine what type of VM was being used for a removed VM, the listVirtualMachines Api will not return a result, as the record has been marked removed. For our purposes, the type of OS is still required for usage purposes. This change simply adds that additional element to the output.

## Types of changes

## How Has This Been Tested?

I have been leveraging the Api for retrieving usage information for virtual machines of specific types fairly extensively.

This change simply adds an additional element to the output, and thus should be considered a non-breaking change.

### DIFF
--- a/server/src/main/java/com/cloud/api/ApiResponseHelper.java
+++ b/server/src/main/java/com/cloud/api/ApiResponseHelper.java
@@ -3386,6 +3386,7 @@ public class ApiResponseHelper implements ResponseGenerator {
                 resourceType = ResourceTag.ResourceObjectType.UserVm;
                 usageRecResponse.setUsageId(vm.getUuid());
                 resourceId = vm.getId();
+                usageRecResponse.setOsTypeId(vm.getGuestOSId());
             }
             //Hypervisor Type
             usageRecResponse.setType(usageRecord.getType());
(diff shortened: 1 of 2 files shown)


davidjumani (approved):
Code LGTM

davidjumani:
@blueorangutan package

rhtyd (approved):

rhtyd:
Simple API setter fix, merging based on Travis tests and packaging.

server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -3386,6 +3386,7 @@ public UsageRecordResponse createUsageResponse(Usage usageRecord, Map<String, Se
                 resourceType = ResourceTag.ResourceObjectType.UserVm;
                 usageRecResponse.setUsageId(vm.getUuid());
                 resourceId = vm.getId();
+                usageRecResponse.setOsTypeId(vm.getGuestOSId());
```
rhtyd, line 3389: @ggoodrich-ipp we just caught that this sets the DB internal ID and not UUID, should we fix that in 4.15.1?

ggoodrich-ipp: I guess I must've realized this, as I'm storing this off as an int in a local billing database. Is it an easy fix to change this to get the UUID? It appears where I'm using it, I fetch this int value, and then use it to go get the UUID, so it seems the UUID is ultimately what I need. Maybe I used the int as that was already in the model? It has been far enough back that I don't recall the specifics.

PR #4317, opened 2020-09-10
Display acl name in listNetworks response

ravening:
## Description

Display network acl name along with its id so that we
dont need to make extra api call `listNetworksAcllist`
 to get acl name

## Types of changes

## How Has This Been Tested?

Through cloudmonkey api

```
(local) mgt01 > list networks id=81f5-5f17-488a-be61-7d6ad99 filter=aclname
{
  "count": 1,
  "network": [
    {
      "aclname": "default_allow"
    }
  ]
}
```

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/response/NetworkResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/NetworkResponse.java
@@ -239,7 +239,7 @@ public class NetworkResponse extends BaseResponse implements ControlledEntityRes
     private Boolean redundantRouter;
 
     @SerializedName(ApiConstants.ACL_NAME)
-    @Param(description = "ACL name associated with the VPC network")
+    @Param(description = "ACL name associated with the VPC network", since = "4.15.0")
     private String aclName;
 
     public Boolean getDisplayNetwork() {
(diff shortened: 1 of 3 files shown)


ravening:
@rhtyd @davidjumani please review

api/src/main/java/org/apache/cloudstack/api/response/NetworkResponse.java
```diff
@@ -238,6 +238,10 @@
     @Param(description = "If the network has redundant routers enabled", since = "4.11.1")
     private Boolean redundantRouter;
 
+    @SerializedName(ApiConstants.ACL_NAME)
+    @Param(description = "ACL name associated with the VPC network")
```
shwstppr, line 242: @ravening will it be better to add a since tag here?

shwstppr (approved):
Code LGTM except a minor nit

@blueorangutan package

davidjumani (approved):
LGTM!

api/src/main/java/org/apache/cloudstack/api/response/NetworkResponse.java
ravening: @shwstppr made the change

PR #4327, opened 2020-09-14
Re-enable IP address usage hiding

olivierlemasle:
## Description
With this commit, listUsageRecords filters out usages made on IP addresses on networks with "Hide IP Address Usage". This was previously introduced in #3235.

This requires a DB schema change, to introduce a flag `is_hidden` in table `cloud_usage.cloud_usage`.

Cf. #4193

## Types of changes

## How Has This Been Tested?
Multiple IP addresses have been created.
- Those on shared networks without "Hide IP Address Usage" generate usages normally;
- Those on shared networks with "Hide IP Address Usage" generate usages with flag `is_hidden`, and are not returned by the API.

### DIFF
--- a/server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
+++ b/server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
@@ -101,6 +101,8 @@ import com.cloud.network.dao.IPAddressVO;
 import com.cloud.network.dao.LoadBalancerDao;
 import com.cloud.network.dao.NetworkAccountDao;
 import com.cloud.network.dao.NetworkDao;
+import com.cloud.network.dao.NetworkDetailsDao;
+import com.cloud.network.dao.NetworkDetailVO;
 import com.cloud.network.dao.NetworkDomainDao;
 import com.cloud.network.dao.NetworkServiceMapDao;
 import com.cloud.network.dao.PhysicalNetworkDao;
@@ -211,6 +213,8 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
     NetworkOfferingDao _networkOfferingDao;
     @Inject
     NetworkDao _networksDao;
+    @Inject
+    NetworkDetailsDao _networkDetailsDao;
     @Inject
     NicDao _nicDao;
     @Inject
@@ -924,9 +928,10 @@ public class IpAddressManagerImpl extends ManagerBase implements IpAddressManage
                                     VlanVO vlan = _vlanDao.findById(addr.getVlanId());
                                     String guestType = vlan.getVlanType().toString();
                                     if (!isIpDedicated(addr)) {
+                                        final boolean usageHidden = isUsageHidden(addr);
                                         UsageEventUtils.publishUsageEvent(EventTypes.EVENT_NET_IP_ASSIGN, owner.getId(), addr.getDataCenterId(), addr.getId(),
-                                                addr.getAddress().toString(),
-                                                addr.isSourceNat(), guestType, addr.getSystem(), addr.getClass().getName(), addr.getUuid());
+                                                addr.getAddress().toString(), addr.isSourceNat(), guestType, addr.getSystem(), usageHidden,
+                                                addr.getClass().getName(), addr.getUuid());
                                     }
                                     if (updateIpResourceCount(addr)) {
                                         _resourceLimitMgr.incrementResourceCount(owner.getId(), ResourceType.public_ip);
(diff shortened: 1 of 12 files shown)


Pearl1594:
@blueorangutan package

DaanHoogland:
@blueorangutan test

olivierlemasle:
Kindly ping @rhtyd

DaanHoogland:
@blueorangutan package

DaanHoogland:
@olivierlemasle Do you have a smoke or integration test for this? it would help guard the feature in the future.
@blueorangutan test

olivierlemasle:
@DaanHoogland I must say I do not know CloudStack smoke tests well; are there any smoke tests on usages data in CloudStack currently, that I could use as a basis?

DaanHoogland:
look in the test/integration/ subdirectory. I imagine there are marvin tests for related functionality, if you want to go that way. there is https://github.com/apache/cloudstack/blob/master/test/integration/smoke/test_usage.py and I hav e to admit I don't know how relevant it is for your case.

svenvogel:
@olivierlemasle which milestone version should it merged? we can set it to 4.16?

DaanHoogland:
@svenvogel as it is marked a bug I wouldimagine 4.15 or before to be appropriate. unless we decide it is too risky and not major..?

olivierlemasle:
:+1: for 4.15! Without it, IP addresses will generate usages on networks flagged "Hide IP address usages"...

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

olivierlemasle:
@DaanHoogland @svenvogel I just rebased on master to fix the conflict in `schema-41400to41500.sql`.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

Pearl1594 (approved):
Code LGTM

DaanHoogland:
@andrijapanicsb what do you think about this?

andrijapanicsb:
what is being different with this vs the original PR?

olivierlemasle:
@andrijapanicsb 
In #3235, usages were filtered out on the fly by the `ApiResponseHelper`, each time a `listUsageRecords` result was returned. It means that that filter ran only for the requested `page` of usages.

However, like all `listXxx` requests, `listUsageRecords`'s response has a `count` field which is equal to the total number of usages (all pages included). It then caused the regession #4191, which was fixed in #4193 by reverting the filtering.

To filter the "hidden" IP address usages with a correct `count`, it is necessary to know how many usages were filtered out _in the other pages_. This PR moves the filter from the API layer to the usage server, to get the filter right without damaging performances.

NB: Since today's rebase, a timeout in smoke tests cause Travis build to fail.

server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -2238,4 +2245,18 @@ public boolean isIpEqualsGatewayOrNetworkOfferingsEmpty(Network network, String
...
     }
+
+    @Override
+    public boolean isUsageHidden(IPAddressVO ip) {
+        Long networkId = ip.getAssociatedWithNetworkId();
+        if (networkId == null) {
+            networkId = ip.getSourceNetworkId();
+        }
+        if (networkId == null) {
+            // Should not happen
```
DaanHoogland, line 2256: if it shouldn't happen, isn't it really an exception?

DaanHoogland (commented):
one little question, i don't like null returns. otherwise good.

server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
olivierlemasle: @DaanHoogland You're right, I could certainly return a `CloudRuntimeException`.

olivierlemasle:
@DaanHoogland I've just updated the PR after your comment.

DaanHoogland:
@blueorangutan package

Pearl1594:
Verified fix:

Created a VM in a shared n/w with hiding IP Usage enabled, and one in a shared network without hiding IP usage -  3 records were created for both networks, While in all there are 33 records, listusagerecords filters out listing ip usage for the n/w with hide ip usage enabled.

```
select count(*) from cloud_usage\G
*************************** 1. row ***************************
count(*): 33
1 row in set (0.00 sec)
``` 

```
list usagerecords startdate="2020-10-26 00:00:00" enddate="2020-11-02 00:00:00" 
{
  "count": 30,
  "usagerecord": [
    {....
    }
  ]
}
```

DaanHoogland:
@blueorangutan test

DaanHoogland (approved):

PR #4353, opened 2020-09-29
Fixing unable to create account when passing account parameter

davidjumani:
## Description
While creating an account, and passing the account parameter, it does not create a user within the given account. Instead it throws an error that the account exists

## Types of changes

## How Has This Been Tested?
TODO


rhtyd:
@davidjumani but that is expected I think - the idea is what you add an account; the account parameter allows creating an account with a different name compared to the passed `username`.

PR #4111, opened 2020-05-28
API-call to declare host as Degraded

GabrielBrascher:
## Description

Sometimes a host is down (admins know) but CS keeps VMs in the 'Running' state to avoid data corruption. An example of that is when a KVM host using Ceph halts. CloudStack cannot check if VM still writing on disk, therefore it leaves it in the "Running" state and HA will not be triggered for those VMs.

With that in mind, this PR proposes an API-call to declare a Host as "Degraded". All "Running"  VMs on that Host should then be set to 'Stopped' so HA can kick in.

1. Command name: `declareHostAsDegraded`; parameters: id (id of the host); description: _Declare host as 'Degraded'. The host must be on 'Disconnected' or 'Alert' state._
2. Command name: `cancelHostAsDegraded`; parameters: id (id of the host); description: _Cancel host status from 'Degraded'. The host will transit back to status 'Enabled'._

### Note:

Admin **MUST** check if there are VMs running on the problematic host and call such API command just if he/she is **100%** sure that can deploy the Running VMs on another host.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## Screenshots (if appropriate):

Example of how host Resource State is shown on UI when a host is declared 'Degraded' (taken when this PR was using state "Dead"):
image

## How Has This Been Tested?

**Note:** tests were performed on two KVM hosts with HA enabled: _Host01_ and _Host02_.

#### Test 01: declareHostAsDegraded
**ha.force = false**
1. Deploy VMs on host Host01
- deploy 2 VMs shared storage, HA enabled (VM01, VM02)
- deploy 1 VM shared storage, HA disabled (VM03)
- deploy 1 VM local storage, HA disabled (VM04)
2. Stop cloudstack-agent.service on Host1
3. Run Virsh command to shutdown/destroy all VMs in Host10
4. All VMs are in "Running" state, the host is in "Disconnected"
5. call API command declareHostsAsDegraded
- Host state migrated to "Degraded"
- VMs with HA + Shared started on Host02 (transit from Running to Stopped, Stopped -> Starting, Starting -> Running)
- VMs with no HA were marked as Stopped (transit from Running to Stopped, restart Scheduler ignores non-HA VMs)

**ha.force=true**
[...]

PR #4363, opened 2020-10-01
Ability to put a server in Down state to maintenance

ravening:
## Description

Adds a new global setting set.host.down.to.maintenance which when set
to true will allow setting a host in downstate into maintenance mode
so that when the host comes back up, it will be automatically set to
maintenance mode and the resource state will not be set to Enabled.

## Types of changes

## How Has This Been Tested?

1, Power off one hypervisor so that it enters DOWN state
2. set the global setting `set.host.down.to.maintenance` to true
3. Enable maintenance mode on that host
4. Start the hypervisor
5. When it comes back up, it will be in maintenance mode

### DIFF
--- a/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
+++ b/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
@@ -1305,7 +1305,7 @@ public class ResourceManagerImpl extends ManagerBase implements ResourceManager,
             throw new CloudRuntimeException("Host is already in state " + host.getResourceState() + ". Cannot recall for maintenance until resolved.");
         }
 
-        if (SET_HOST_DOWN_TO_MAINTENANCE.value() && (host.getStatus() == Status.Down)) {
+        if (SET_HOST_DOWN_TO_MAINTENANCE.valueIn(host.getDataCenterId()) && (host.getStatus() == Status.Down)) {
             if (host.getResourceState() == ResourceState.Enabled) {
                 _hostDao.updateResourceState(ResourceState.Enabled, ResourceState.Event.AdminAskMaintenance, ResourceState.PrepareForMaintenance, host);
                 _hostDao.updateResourceState(ResourceState.PrepareForMaintenance, ResourceState.Event.InternalEnterMaintenance, ResourceState.Maintenance, host);
--- a/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
+++ b/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
@@ -180,6 +180,9 @@ import com.cloud.vm.dao.UserVmDetailsDao;
 import com.cloud.vm.dao.VMInstanceDao;
 import com.google.gson.Gson;
 
+
+import static com.cloud.configuration.ConfigurationManagerImpl.SET_HOST_DOWN_TO_MAINTENANCE;
+
 @Component
 public class ResourceManagerImpl extends ManagerBase implements ResourceManager, ResourceService, Manager {
     private static final Logger s_logger = Logger.getLogger(ResourceManagerImpl.class);
@@ -1300,6 +1303,17 @@ public class ResourceManagerImpl extends ManagerBase implements ResourceManager,
         }
         if (!ResourceState.canAttemptMaintenance(host.getResourceState())) {
             throw new CloudRuntimeException("Host is already in state " + host.getResourceState() + ". Cannot recall for maintenance until resolved.");
+        }
+
+        if (SET_HOST_DOWN_TO_MAINTENANCE.value() && (host.getStatus() == Status.Down)) {
+            if (host.getResourceState() == ResourceState.Enabled) {
+                _hostDao.updateResourceState(ResourceState.Enabled, ResourceState.Event.AdminAskMaintenance, ResourceState.PrepareForMaintenance, host);
+                _hostDao.updateResourceState(ResourceState.PrepareForMaintenance, ResourceState.Event.InternalEnterMaintenance, ResourceState.Maintenance, host);
+                return _hostDao.findById(hostId);
+            } else if (host.getResourceState() == ResourceState.ErrorInMaintenance) {
+                _hostDao.updateResourceState(ResourceState.ErrorInMaintenance, ResourceState.Event.InternalEnterMaintenance, ResourceState.Maintenance, host);
+                return _hostDao.findById(hostId);
+            }
         }
 
         if (_hostDao.countBy(host.getClusterId(), ResourceState.PrepareForMaintenance, ResourceState.ErrorInPrepareForMaintenance) > 0) {
--- a/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
+++ b/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
@@ -415,10 +415,12 @@ public class ConfigurationManagerImpl extends ManagerBase implements Configurati
             "Maximum IOPS read burst duration (seconds). If '0' (zero) then does not check for maximum burst length.", true, ConfigKey.Scope.Global, null);
     public final static ConfigKey<Long> IOPS_MAX_WRITE_LENGTH = new ConfigKey<Long>(Long.class, "vm.disk.iops.maximum.write.length", "Advanced", "0",
             "Maximum IOPS write burst duration (seconds). If '0' (zero) then does not check for maximum burst length.", true, ConfigKey.Scope.Global, null);
-
     public static final ConfigKey<Boolean> ADD_HOST_ON_SERVICE_RESTART_KVM = new ConfigKey<Boolean>(Boolean.class, "add.host.on.service.restart.kvm", "Advanced", "true",
             "Indicates whether the host will be added back to cloudstack after restarting agent service on host. If false it wont be added back even after service restart",
             true, ConfigKey.Scope.Global, null);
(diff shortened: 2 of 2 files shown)


weizhouapache (approved):

GabrielBrascher:
@ravening what do you think PR https://github.com/apache/cloudstack/pull/4111? Does that PR fit the context of this PR?

ravening:
> @ravening what do you think PR https://github.com/apache/cloudstack/pull/4111? Does that PR fit the context of this PR?

@GabrielBrascher looked at it and have few questions


The declare degrade API works only if host is in disconnected state? 
What if we set host to maintenance mode and after that for some reason it goes down, we need to execute the api.
But with global setting we don't need to execute another API.


Also to declare host as not dead, we need to call another API and then cancel maintenance mode but with global setting , you need to just cancel maintenance mode

Apart from these two extra steps, both are doing same functionality

GabrielBrascher:
@ravening the declare host as degraded API command would serve to put a problematic host on a special state that would allow to mark the VMs as stopped and trigger HA to start VMs (when service offering has _haenabled=true_) on another hosts.

Your implementation is nice and covers some interesting ground. The API command requires intervention from ADMIN with care. This is because if the host is Disconnected but there are VMs running in it and writing on storage it may cause corruption if VMs start on another host and write on same storage. Therefore that command should only be executed if the host is down.

API description: _Declare host as 'Degraded'. Host must be on 'Disconnected' or 'Alert' state. The ADMIN must be sure that there are no VMs running on the respective host otherwise this command might corrupted VMs that were running on the 'Degraded' host_.

Threfore, it cannot be automated. I think that both implementations cover different gaps on some corner cases.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -1302,6 +1305,17 @@ public Host maintain(final PrepareForMaintenanceCmd cmd) {
             throw new CloudRuntimeException("Host is already in state " + host.getResourceState() + ". Cannot recall for maintenance until resolved.");
         }
 
+        if (SET_HOST_DOWN_TO_MAINTENANCE.value() && (host.getStatus() == Status.Down)) {
```
GabrielBrascher, line 1308: @ravening what do you think of extracting lines 1308 - 1318 to a method? With that, it would be possible to cover these cases with JUnit test methods?

GabrielBrascher (commented):
Thanks for the PR @ravening. The code and feature are looking good.

I am not against merging it the way it is; however, a big plus would be adding a couple of test cases covering the added logics.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
sureshanaparti, line 1308: @ravening consider the value in zone, as the config scope is specified as Zone.

```suggestion
        if (SET_HOST_DOWN_TO_MAINTENANCE.valueIn(host.getDataCenterId()) && (host.getStatus() == Status.Down)) {
```

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -1302,6 +1305,17 @@ public Host maintain(final PrepareForMaintenanceCmd cmd) {
...
+            if (host.getResourceState() == ResourceState.Enabled) {
+                _hostDao.updateResourceState(ResourceState.Enabled, ResourceState.Event.AdminAskMaintenance, ResourceState.PrepareForMaintenance, host);
+                _hostDao.updateResourceState(ResourceState.PrepareForMaintenance, ResourceState.Event.InternalEnterMaintenance, ResourceState.Maintenance, host);
+                return _hostDao.findById(hostId);
+            } else if (host.getResourceState() == ResourceState.ErrorInMaintenance) {
+                _hostDao.updateResourceState(ResourceState.ErrorInMaintenance, ResourceState.Event.InternalEnterMaintenance, ResourceState.Maintenance, host);
+                return _hostDao.findById(hostId);
+            }
```
sureshanaparti, line 1316: @ravening can also consider state "_ResourceState.ErrorInPrepareForMaintenance_" ?

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
ravening: > @ravening consider the value in zone, as the config scope is specified as Zone.
> 
> 
> 
> ```suggestion
> 
>         if (SET_HOST_DOWN_TO_MAINTENANCE.valueIn(host.getDataCenterId()) && (host.getStatus() == Status.Down)) {
> 
> ```

@sureshanaparti i don't think we need that because the host can exist only in one zone at any time but I can change the setting to zone scope instead of global scope

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
ravening: > @ravening can also consider state "_ResourceState.ErrorInPrepareForMaintenance_" ?

@sureshanaparti no. This is applicable only for hosts in maintenance state

sureshanaparti: @ravening any specific reason to skip "_ResourceState.ErrorInPrepareForMaintenance_", and consider "_ResourceState.ErrorInMaintenance_", to put the host in maintenance? What if the host is back Up when the resource state is "ErrorInPrepareForMaintenance" ?

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
sureshanaparti: @ravening Currently, the config scope is Zone, do you want to mark that as Global scope?

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
ravening: > @ravening any specific reason to skip "_ResourceState.ErrorInPrepareForMaintenance_", and consider "_ResourceState.ErrorInMaintenance_", to put the host in maintenance? What if the host is back Up when the resource state is "ErrorInPrepareForMaintenance" ?

@sureshanaparti  this code doesn't take care of that... This is only for hosts already in maintenance mode

sureshanaparti: ok @ravening

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
ravening: @sureshanaparti I dont think I need to make changes here as config is at zone level..setting in one zone doesnt mess up setting in other zones

sureshanaparti: @ravening May be some confusion here. My point is, use the config value specified at the scope defined in the config parameter, else update the config scope accordingly.

The config "SET_HOST_DOWN_TO_MAINTENANCE" defined above uses Zone Scope. So, In order to pick the Zone specific value of that config, you have to use "_valueIn(<zoneId>)_", not _value()_ (this picks the global config value). If your use case is to set the config at Cloud level, then change Scope to Global.

svenvogel:
@ravening small question. down means not enabled?

ravening:
> @ravening small question. down means not enabled?

@svenvogel need not be. The host is still enabled but is in down state

DaanHoogland:
@ravening can you look at the conflict?

ravening:
@GabrielBrascher so whats your opinion on this pr? shall I close it and take yours further?

GabrielBrascher:
@ravening your PR covers a different scenario, very similar though. From what I understand, this PR avoids a problematic host to get back from down to Enabled.

I don't see any problem in merging this one, but here follows a small summary on what #4111 does:

In summary, it presents an API command to manually fence a problematic host. The command declares a host as degraded moving it from Alert or Disconnected to "Degraded" state. All VMs that were running and had HA enabled will be started on another host, VMs with HA disabled will triansit from "Running" state (obviously they are not running, but CloudStack does not know that) to Stopped.

In summary, this implementation is a simple way of avoiding a problematic host to get back and does not require admins to take action. The other PR requires admins to execute an API command when they know a Disconnected/Alert host is down and has no running VMs.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
GabrielBrascher: This observation makes sense. Can you please take a look at _org.apache.cloudstack.framework.config.ConfigKey.valueIn()_ @ravening?

GabrielBrascher (approved):
Overall code LGTM, left a small comment regarding the config key being used with _value()_ vs _valueIn()_.

@DaanHoogland @svenvogel @rhtyd @andrijapanicsb what do you think of this funcionality?

GabrielBrascher:
@ravening can you please fix the conflicts?

ravening:
> @ravening can you please fix the conflicts?

@GabrielBrascher have resolved the conflicts

GabrielBrascher:
Thanks, @ravening!

GabrielBrascher:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@GabrielBrascher functionality makes sense and is guarded by a setting. 2 lgtms and test-report, so merging

Issue #4350, opened 2020-09-27
modifyvxlan.sh expect /var/run/cloud directory for LOCK, but directory require manual creation


levindecaro:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
cloudstack-agent
~~~

##### CLOUDSTACK VERSION

~~~
4.14
~~~

##### CONFIGURATION

advanced networking, vxlan

##### OS / ENVIRONMENT

centos 7.8.2003

##### SUMMARY

Error when create guest instance, agent log show 

Sep 27 12:38:59 compute-12 java: WARN  [resource.wrapper.LibvirtPlugNicCommandWrapper] (agentRequest-Handler-3:) (logid:32fe0cb1)  Plug Nic failed due to com.cloud.exception.InternalErrorException: Failed to create vnet 2726: /usr/share/cloudstack-common/scripts/vm/network/vnet/modifyvxlan.sh: line 145: /var/run/cloud/vxlan.lock: No such file or directory

modifyvxlan.sh expect /var/run/cloud directory for persist LOCK file, but directory require manual creation

##### STEPS TO REPRODUCE

~~~
create zone with advanced networking without security group
import a centos/debian vm template
create a VPC 10.0.0.0/8
create a network zone 10.1.1.1/24
launch a vm instance
~~~

##### EXPECTED RESULTS

~~~
vm should be created
~~~

##### ACTUAL RESULTS

~~~
creation fail with "Command failed due to Internal Server Error"
~~~

DaanHoogland:
fixed in #4373
can you confirm @levindecaro ?

DaanHoogland:
verified by reported

PR #4373, opened 2020-10-05
Handles creation /var/run/cloud folder for creation of lock file while modifyvxlan.sh script is run

Pearl1594:
## Description
modifyvxlan.sh script fails run as it isn't able to find the /var/run/cloud directory to create the lock file
Fixes: https://github.com/apache/cloudstack/issues/4350

## Types of changes

## How Has This Been Tested?
- Deployed an Advanced zone with VXLAN as isolation method
- deployed a VM in an isolated network

### DIFF
--- a/scripts/vm/network/vnet/modifyvxlan.sh
+++ b/scripts/vm/network/vnet/modifyvxlan.sh
@@ -131,6 +131,9 @@ fi
 
 LOCKFILE=/var/run/cloud/vxlan.lock
 
+# ensures that parent directories exists and prepares the lock file
+mkdir -p "${LOCKFILE%/*}"
+
 (
     flock -x -w 10 200 || exit 1
     if [[ "$OP" == "add" ]]; then


scripts/vm/network/vnet/modifyvxlan.sh
```diff
@@ -131,6 +131,9 @@ fi
 
 LOCKFILE=/var/run/cloud/vxlan.lock
 
+# ensures that parent directories exists and prepares the lock file
+mkdir -p "${LOCKFILE%/*}" && touch "$LOCKFILE"
```
DaanHoogland, line 135: we do not need the touch right?

DaanHoogland (approved):

DaanHoogland:
@blueorangutan package

scripts/vm/network/vnet/modifyvxlan.sh
Pearl1594: Wouldn't actually require the file creation

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan  package

wido:
The big question: Why doesn't this exist?

I have never seen these issues with Ubuntu systems.

rhtyd:
Ideally this must be created on pkg installation.
@blueorangutan package

wido:
@rhtyd /var/run is a tmpfs on many systems nowadays, so the package can't fix this.

I just wonder why we (PCextreme) never run into this.

rhtyd (approved):
LGTM, this will ensure the folder exists

rhtyd:
Manually tested the logic on bash, LGTM. Merging based on travis tests and lgtms.

levindecaro:
@DaanHoogland 

I would like to report #4350 fixed create directory issue. thx

Issue #4372, opened 2020-10-05
list roles API returns a wrong count


DaanHoogland:
##### ISSUE TYPE

 * Bug Report
 * Improvement Request
 * Enhancement Request
 * Feature Idea
 * Documentation Report
 * Other

##### COMPONENT NAME

~~~

~~~

##### CLOUDSTACK VERSION

~~~

~~~

##### SUMMARY

deleted roles are counted in the total for listRoles.

##### STEPS TO REPRODUCE

~~~
create extra roles
delete the extra roles again
list roles
~~~

##### EXPECTED RESULTS

~~~
```
(localcloud) SBCM5> > list roles filter=name
{
  "count": 8,
  "role": [
    {
      "name": "Root Admin"
    },
    {
      "name": "Resource Admin"
    },
    {
      "name": "Domain Admin"
    },
    {
      "name": "User"
    },
    {
      "name": "Read-Only Admin - Default"
    },
    {
      "name": "Read-Only User - Default"
    },
    {
      "name": "Support Admin - Default"
    },
    {
      "name": "Support User - Default"
    }
  ]
}
```
~~~

##### ACTUAL RESULTS

~~~
```
(localcloud) SBCM5> > list roles filter=name
{
  "count": 71,
  "role": [
    {
      "name": "Root Admin"
    },
    {
      "name": "Resource Admin"
    },
    {
      "name": "Domain Admin"
    },
    {
      "name": "User"
    },
    {
      "name": "Read-Only Admin - Default"
    },
    {
      "name": "Read-Only User - Default"
    },
    {
      "name": "Support Admin - Default"
    },
    {
      "name": "Support User - Default"
    }
  ]
}
```
~~~

PR #4374, opened 2020-10-05
Fixing searchAndCount searchAndDistinctCount when sc is null

davidjumani:
## Description
Fixes https://github.com/apache/cloudstack/issues/4372

Ensures that the count returned by searchAndCount searchAndDistinctCount do not include removed items
Before, it would return the count of removed items  as well when the search criteria is null (since sc is created and modified in search which doesn't reflect when it tires to get the count)
Also refactored a bit to make life easieer for the next person who comes along

## Types of changes

## How Has This Been Tested?
Before :
```
(localcloud) SBCM5> > list roles filter=
{
  "count": 11,
  "role": [
    {},
    {},
    {},
    {},
    {},
    {},
    {},
    {},
    {}
  ]
}
```
After : 
```
(localcloud) SBCM5> > list roles filter=
{
  "count": 9,
  "role": [
    {},
    {},
    {},
    {},
    {},
    {},
    {},
    {},
    {}
  ]
}

```

### DIFF
--- a/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
+++ b/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
@@ -344,23 +344,13 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
 
     @DB()
     protected List<T> search(SearchCriteria<T> sc, final Filter filter, final Boolean lock, final boolean cache) {
-        if (_removed != null) {
-            if (sc == null) {
-                sc = createSearchCriteria();
-            }
-            sc.addAnd(_removed.second().field.getName(), SearchCriteria.Op.NULL);
-        }
+        sc = checkAndSetRemovedIsNull(sc);
         return searchIncludingRemoved(sc, filter, lock, cache);
     }
 
     @DB()
     protected List<T> search(SearchCriteria<T> sc, final Filter filter, final Boolean lock, final boolean cache, final boolean enableQueryCache) {
-        if (_removed != null) {
-            if (sc == null) {
-                sc = createSearchCriteria();
-            }
-            sc.addAnd(_removed.second().field.getName(), SearchCriteria.Op.NULL);
-        }
+        sc = checkAndSetRemovedIsNull(sc);
         return searchIncludingRemoved(sc, filter, lock, cache, enableQueryCache);
     }
 
@@ -519,7 +509,6 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
         if (_removed != null) {
             sc.addAnd(_removed.second().field.getName(), SearchCriteria.Op.NULL);
         }
-
         return customSearchIncludingRemoved(sc, filter);
     }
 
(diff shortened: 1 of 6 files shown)


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

DaanHoogland (approved):
code looks good. I think the search and count method should however have a single set of criteria for both the results and the count and not need any adjustments down the line. starting with an sc only containing "true", could be a solution (as well or on top of this, no pref)

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -519,7 +509,6 @@ public T lockOneRandomRow(final SearchCriteria<T> sc, final boolean exclusive) {
         if (_removed != null) {
             sc.addAnd(_removed.second().field.getName(), SearchCriteria.Op.NULL);
```
sureshanaparti, line 510: can use _checkAndSetRemovedIsNull()_ here ?

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1935,6 +1918,23 @@ public boolean configure(final String name, final Map<String, Object> params) th
...
+            if (sc == null) {
+                sc = createSearchCriteria();
+            }
+            sc.addAnd(_removed.second().field.getName(), SearchCriteria.Op.NULL);
+        }
+        return sc;
+    }
+
+    public Integer getDistinctCount(SearchCriteria<T> sc, boolean removed) {
+        if (!removed) {
```
sureshanaparti, line 1932: `removed` & `_removed` seems to be redundant here & wherever used in similar way, can you directly set / unset `_removed` as per its usage ? If the purpose is to include / exclude removed records, you can use methods with naming _*IncludingRemovedBy()_ [for ex: _getDistinctCountIncludingRemovedBy()_] to be in sync with the other methods.

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
davidjumani: Makes it a bit difficult since it is of a different type M not T

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
davidjumani: As of now `getDistinctCount` returns including removed so changing it would require changing where it's been referenced, would it be better to create `getDistinctCountExcludingRemoved`  ?

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
sureshanaparti: ok @davidjumani

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
sureshanaparti: _getDistinctCountExcludingRemoved_ is not in line with other methods, so I do not recommend.

sureshanaparti: Default method should not include removed records, explicitly mentioned methods "_*IncludingRemovedBy()_" should.

davidjumani: Added! Thanks @sureshanaparti

sureshanaparti: Perfect, changes LGTM.

sureshanaparti (approved):

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

davidjumani:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan test centos7 xs71
@blueorangutan test centos7 kvmcentos7

sureshanaparti:

DaanHoogland:

DaanHoogland:

Issue #4371, opened 2020-10-05
Incorrect count returned by findHostsForMigration


davidjumani:
findHostsForMigration returns wrong count

##### ISSUE TYPE

 * Bug Report

##### CLOUDSTACK VERSION
master

##### STEPS TO REPRODUCE
~~~
(localcloud) SBCM5> > find hostsformigration virtualmachineid=3a2e2d93-2105-4b77-a5df-c2c07caab43c filter=
{
  "count": 6,
  "host": [
    {},
    {}
  ]
}

~~~

##### EXPECTED RESULTS
The right count

DaanHoogland:
#relates to #4372

PR #4375, opened 2020-10-05
Fixing count for findHostsForMigration

davidjumani:
## Description
Fixes https://github.com/apache/cloudstack/issues/4371

In the code, all servers, and their count belonging to the same zone are returned. Then they are filtered for suitability.
This is fine except for the point where VM volumes existing on managed storage pools or hosts on which those storage pools are not suitable. In this case it removes the hosts rather than marking them as not suitable. This messes up the count. Subtracting the non suitable hosts count does not fix this since we're paginating the response so unsuitable hosts and thier count from the current page will be removed but will not fix the total count since hosts on other pages can not be filtered.
Another place is that the source host is removed from the list but the count is not adjusted

## Types of changes

## How Has This Been Tested?
Before :
```
{
  "count": 4,
  "host": [
    {},
    {},
    {}
  ]
}

```

After :
```
{
  "count": 3,
  "host": [
    {},
    {},
    {}
  ]
}

```

### DIFF
--- a/server/src/main/java/com/cloud/server/ManagementServerImpl.java
+++ b/server/src/main/java/com/cloud/server/ManagementServerImpl.java
@@ -1275,20 +1275,20 @@ public class ManagementServerImpl extends ManagerBase implements ManagementServe
         final Type hostType = srcHost.getType();
         Pair<List<HostVO>, Integer> allHostsPair = null;
         List<HostVO> allHosts = null;
-        List<HostVO> hostsWithStorageMigration = null;
+        List<HostVO> hostsForMigrationWithStorage = null;
         final Map<Host, Boolean> requiresStorageMotion = new HashMap<Host, Boolean>();
         DataCenterDeployment plan = null;
         if (canMigrateWithStorage) {
             allHostsPair = searchForServers(startIndex, pageSize, null, hostType, null, srcHost.getDataCenterId(), null, null, null, keyword,
                 null, null, srcHost.getHypervisorType(), srcHost.getHypervisorVersion(), srcHost.getId());
             allHosts = allHostsPair.first();
-            hostsWithStorageMigration = new ArrayList<>(allHosts);
+            hostsForMigrationWithStorage = new ArrayList<>(allHosts);
 
             for (final VolumeVO volume : volumes) {
                 StoragePool storagePool = _poolDao.findById(volume.getPoolId());
                 Long volClusterId = storagePool.getClusterId();
 
-                for (Iterator<HostVO> iterator = hostsWithStorageMigration.iterator(); iterator.hasNext();) {
+                for (Iterator<HostVO> iterator = hostsForMigrationWithStorage.iterator(); iterator.hasNext();) {
                     final Host host = iterator.next();
 
                     if (volClusterId != null) {
@@ -1358,7 +1358,7 @@ public class ManagementServerImpl extends ManagerBase implements ManagementServe
 
         for (final HostAllocator allocator : hostAllocators) {
             if (canMigrateWithStorage) {
-                suitableHosts = allocator.allocateTo(vmProfile, plan, Host.Type.Routing, excludes, hostsWithStorageMigration, HostAllocator.RETURN_UPTO_ALL, false);
+                suitableHosts = allocator.allocateTo(vmProfile, plan, Host.Type.Routing, excludes, hostsForMigrationWithStorage, HostAllocator.RETURN_UPTO_ALL, false);
             } else {
                 suitableHosts = allocator.allocateTo(vmProfile, plan, Host.Type.Routing, excludes, HostAllocator.RETURN_UPTO_ALL, false);
             }
--- a/server/src/main/java/com/cloud/server/ManagementServerImpl.java
+++ b/server/src/main/java/com/cloud/server/ManagementServerImpl.java
@@ -1172,8 +1172,8 @@ public class ManagementServerImpl extends ManagerBase implements ManagementServe
         final Object resourceState = cmd.getResourceState();
         final Object haHosts = cmd.getHaHost();
 
-        final Pair<List<HostVO>, Integer> result = searchForServers(cmd.getStartIndex(), cmd.getPageSizeVal(), name, type, state, zoneId, pod, cluster, id, keyword, resourceState, haHosts, null,
-                null);
+        final Pair<List<HostVO>, Integer> result = searchForServers(cmd.getStartIndex(), cmd.getPageSizeVal(), name, type, state, zoneId, pod,
+            cluster, id, keyword, resourceState, haHosts, null, null);
         return new Pair<List<? extends Host>, Integer>(result.first(), result.second());
     }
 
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

sureshanaparti:
@davidjumani can you please add the root cause for the count mismatch and the fix to the PR description (and also commit msg).

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -1278,33 +1277,24 @@ private HypervisorType getHypervisorType(VMInstanceVO vm, StoragePool srcVolumeP
...
+                for (HostVO host : allHosts) {
                     if (volClusterId != null) {
                         if (storagePool.isLocal() || !host.getClusterId().equals(volClusterId) || usesLocal) {
-                            if (storagePool.isManaged()) {
-                                // At the time being, we do not support storage migration of a volume from managed storage unless the managed storage
-                                // is at the zone level and the source and target storage pool is the same.
-                                // If the source and target storage pool is the same and it is managed, then we still have to perform a storage migration
-                                // because we need to create a new target volume and copy the contents of the source volume into it before deleting the
-                                // source volume.
-                                iterator.remove();
```
sureshanaparti, line 1301: @davidjumani any reason, why unsuitable hosts are not removed from the hosts list ?

davidjumani: This is so that the total count and hosts returned remain the same. Rather than removing them, they're marked as not suitable
https://github.com/apache/cloudstack/blob/master/api/src/main/java/org/apache/cloudstack/api/command/admin/host/FindHostsForMigrationCmd.java#L94

sureshanaparti: in actual production, there can be lot of unsuitable hosts. say 60 out of 100 are unsuitable, in this case, list all these and marking unsuitable doesn't sound good. thoughts?

sureshanaparti: @davidjumani better to update the count result from _searchForServers()_, when unsuitable hosts are removed, instead changing the existing code for suitable hosts, can cause regression.

davidjumani: Agreed, but since we're paginating the result, we can not remove the hosts in page 2 when we've only fetched the ones in page 1. So marking it as not compatible so that the total count is not broken

sureshanaparti: Pagination is the UI thing, may be some other alternative that can be worked upon. If someone uses the API, doesn't look good to return all hosts, marking suitable & unsuitable.

davidjumani: Again I agree, but that's the way it's already being done, so just fixing the count
https://github.com/apache/cloudstack/blob/master/api/src/main/java/org/apache/cloudstack/api/command/admin/host/FindHostsForMigrationCmd.java#L89

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan test centos7 kvm-centos7 keepEnv

DaanHoogland:

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -1278,33 +1277,24 @@ private HypervisorType getHypervisorType(VMInstanceVO vm, StoragePool srcVolumeP
...
-                            } else {
-                                if (hasSuitablePoolsForVolume(volume, host, vmProfile)) {
-                                    requiresStorageMotion.put(host, true);
-                                } else {
-                                    iterator.remove();
-                                }
+                            // At the time being, we do not support storage migration of a volume from managed storage unless the managed storage
```
rhtyd, line 1291: My only concern here is that now that we're not removing hosts, what is those hosts which are not removed now as shown as suitable, would that cause any regression?

davidjumani: Will mark them as not suitable based on the `requiresStorageMotion` map

davidjumani:
@blueorangutan package

server/src/main/java/com/cloud/server/ManagementServerImpl.java
davidjumani: @sureshanaparti Sorted in the UI via https://github.com/apache/cloudstack-primate/pull/788/commits/a4732a334bd4be2df755dbe704e1894b1200bc69

davidjumani:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan test centos7 kvm-centos7 keepEnv

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -1275,19 +1275,20 @@ private HypervisorType getHypervisorType(VMInstanceVO vm, StoragePool srcVolumeP
         final Type hostType = srcHost.getType();
         Pair<List<HostVO>, Integer> allHostsPair = null;
         List<HostVO> allHosts = null;
+        List<HostVO> hostsWithStorageMigration = null;
```
sureshanaparti, line 1278: @davidjumani do you mean hosts for migration with storage? why another list is required here?

DaanHoogland:
@blueorangutan test centos7 kvm-centos7 keepEnv

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -1326,10 +1327,9 @@ private HypervisorType getHypervisorType(VMInstanceVO vm, StoragePool srcVolumeP
             if (s_logger.isDebugEnabled()) {
                 s_logger.debug("Searching for all hosts in cluster " + cluster + " for migrating VM " + vm);
             }
-            allHostsPair = searchForServers(startIndex, pageSize, null, hostType, null, null, null, cluster, null, keyword, null, null, null, null);
-            // Filter out the current host.
+            allHostsPair = searchForServers(startIndex, pageSize, null, hostType, null, null, null, cluster, null, keyword, null, null, null,
+                null, srcHost.getId());
             allHosts = allHostsPair.first();
-            allHosts.remove(srcHost);
```
sureshanaparti, line 1332: @davidjumani didn't see any change with respect to hosts count, does these changes fix the count issue?

server/src/main/java/com/cloud/server/ManagementServerImpl.java
davidjumani: Previously the `allHosts` list was altered (if storage migration was required) and then used to find the suitable hosts. Instead, now it's being cloned into `hostsWithStorageMigration` and using the cloned list to find the suitable hosts so that the original list is not altered.

server/src/main/java/com/cloud/server/ManagementServerImpl.java
davidjumani: It does since I've added an excludes param in the `searchForServers` method and passing the source host ID to it. This excludes the source host in the query, as well as ensures that the allHosts list is never altered. So the count remains consistant but unsuitable hosts are marked accordingly

server/src/main/java/com/cloud/server/ManagementServerImpl.java
sureshanaparti: Ok @davidjumani 

```suggestion
        List<HostVO> hostsForMigrationWithStorage = null;
```

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -1326,10 +1327,9 @@ private HypervisorType getHypervisorType(VMInstanceVO vm, StoragePool srcVolumeP
...
             plan = new DataCenterDeployment(srcHost.getDataCenterId(), srcHost.getPodId(), srcHost.getClusterId(), null, null, null);
         }
```
sureshanaparti, line 1335: @davidjumani can you check if `otherHosts` here is not required as there is no change in the `allHosts` & its count, which you can return as it is (as a List), instead of Pair (with count). You can use use the `allHosts` count wherever applicable.

DaanHoogland:
@blueorangutan test centos7 kvm-centos7 keepEnv

DaanHoogland (approved):

server/src/main/java/com/cloud/server/ManagementServerImpl.java
davidjumani: It's needed here since the data type is different `Pair<List<HostVO>, Integer>` vs `Pair<List<? extends Host>, Integer>`

sureshanaparti: thanks for confirming that @davidjumani

sureshanaparti (approved):
Changes LGTM, not tested though

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@davidjumani can you have a quick look at the errors above?

davidjumani:
@DaanHoogland The host fencing errors are due to `com.google.gson.stream.MalformedJsonException: Unterminated object near assword":"p*****},"resourcestate":"Enabl'}`
Think that's beeen fixed in https://github.com/apache/cloudstack/pull/4387
Need to look into the vpc issues

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

PR #4380, opened 2020-10-06
RBD: support multiple MON hosts

ustcweizhou:
## Types of changes

ceph-rbd

details

ceph-rbd-details

### DIFF
--- a/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
+++ b/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
@@ -141,6 +141,9 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
 
         URI uri = null;
         try {
+            if (url != null && url.startsWith("rbd://")) {
+                url = url.replace(",", "/");
+            }
             uri = new URI(UriUtils.encodeURIComponent(url));
             if (uri.getScheme() == null) {
                 throw new InvalidParameterValueException("scheme is null " + url + ", add nfs:// (or cifs://) as a prefix");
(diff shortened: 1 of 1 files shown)


DaanHoogland (approved):
the changes look good but they are in a 225 lines method that should be restructured to contain scheme handlers for different schemes with init and prep methods.

wido:
What does this add? Librados/RBD already handles this by creating a Round Robin DNS name pointing to all three IPs.

There is no need to try and hack this into the URL. Redundancy can already be achieved with DNS.

weizhouapache:
> What does this add? Librados/RBD already handles this by creating a Round Robin DNS name pointing to all three IPs.

@wido libvirt also supports multiple monitors. this provides another option.

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -141,6 +141,9 @@ public DataStore initialize(Map<String, Object> dsInfos) {
 
         URI uri = null;
         try {
+            if (url != null && url.startsWith("rbd://")) {
```
sureshanaparti, line 144: in case the url scheme is not case sensitive, can use:

```suggestion
            if (url != null && url..toLowerCase().startsWith("rbd://")) {
```

(or)

```suggestion
            if (StringUtils.startsWithIgnoreCase(url, "rbd://")) {
```

wido:
The PR currently only contains one commit and just a few changes. This seems incomplete to me.

Where do we generate the Libvirt XML?

weizhouapache:
> The PR currently only contains one commit and just a few changes. This seems incomplete to me.
> 
> Where do we generate the Libvirt XML?

@wido this does not require any other code changes in kvm plugin.
in libvirt pool definition, it uses host ips instead of single host ip.

```
<pool type='rbd'>
  <name>uuid</name>
  <uuid>uuid</uuid>
  <capacity unit='bytes'>aaaaaa</capacity>
  <allocation unit='bytes'>aaaaaa</allocation>
  <available unit='bytes'>aaaaaa</available>
  <source>
    <host name='10.11.118.42,10.11.118.43,10.11.118.44' port='6789'/>
    <name>ceph-rbd</name>
    <auth type='ceph' username='cloudstack'>
      <secret uuid='60a41767-1196-3ff3-9105-19de5dc99fce'/>
    </auth>
  </source>
</pool>
```

weizhouapache:
closed this ticket.

I do not have a ceph testing env for now.
I will open this ticket when I have verified everything works as expected.

rohityadavcloud:
@weizhouapache I've verified this, it works but not in the syntax you've shared but:
```
<pool type='rbd'>
  <name><name uuid here></name>
  ....
  <source>
    <host name='192.168.1.11'/>
    <host name='192.168.1.12'/>
    <host name='192.168.1.13'/>
    <name><name of ceph rbd pool here></name>
    <auth type='ceph' username='cloudstack'>
      <secret uuid='<uuid here>'/>
    </auth>
  </source>
</pool>
```
https://libvirt.org/storage.html#StorageBackendRBD

DNS is good solution @wido but some people including me may like this solution.

weizhouapache:
re-work on this feature and created a pull request
https://github.com/apache/cloudstack/pull/6792

Issue #4355, opened 2020-09-29
JsonSyntaxException when creating API command response


GabrielBrascher:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API
~~~

##### CLOUDSTACK VERSION

~~~
master/4.15
~~~

##### CONFIGURATION

~~~
N/A
~~~
##### OS / ENVIRONMENT

~~~
N/A
~~~

##### SUMMARY

Issues when API command response is generated. Exception at the log and API command result. The command works fine, but to whom using the API it seems to have failed.

##### STEPS TO REPRODUCE

~~~
Call an API command, e.g. prepareHostMaintenance or cancelHostMaintenance
~~~

##### EXPECTED RESULTS

~~~
Proper API response and no log exceptions on the management node
~~~

##### ACTUAL RESULTS

1. When calling API command via CloudMonkey:
```
> prepare hostformaintenance id=host-id
Error: async API failed for job b43bbdbf-fe65-436c-9e7e-7c48e4fbfea1
```

2. log at the management node:
```
2020-09-29 16:55:15,853 DEBUG [c.c.a.t.Request] (API-Job-Executor-4:ctx-4d783c5f job-55326 ctx-7d98f674) (logid:b43bbdbf) Seq 24-3700551518814994444: Received:  { Ans: , MgmtId: 2482717785957, via: 24(n06.2019.cldin.net), Ver: v1, Flags: 110, { MaintainAnswer } }
2020-09-29 16:55:15,874 DEBUG [c.c.r.ResourceState] (API-Job-Executor-4:ctx-4d783c5f job-55326 ctx-7d98f674) (logid:b43bbdbf) Resource state update: [id = 24; name = n06.2019.cldin.net; old state = Enabled; event = AdminAskMaintenance; new state = PrepareForMaintenance]
2020-09-29 16:55:15,917 DEBUG [c.c.r.ResourceManagerImpl] (API-Job-Executor-4:ctx-4d783c5f job-55326 ctx-7d98f674) (logid:b43bbdbf) Sent resource event EVENT_PREPARE_MAINTENANCE_AFTER to listener CapacityManagerImpl
2020-09-29 16:55:15,950 ERROR [c.c.a.ApiAsyncJobDispatcher] (API-Job-Executor-4:ctx-4d783c5f job-55326) (logid:b43bbdbf) Unexpected exception while executing org.apache.cloudstack.api.command.admin.host.PrepareForMaintenanceCmd
[...]

PR #4387, opened 2020-10-06
Fix JsonSyntaxException when creating API command response #4355 

GabrielBrascher:
## Description

Added missing `\"` (from `"*****"` to `"*****\""`) at _obfuscatePassword method from AsyncJobManagerImpl.java#L485_
https://github.com/apache/cloudstack/blob/0f3f2a09370a18301db28ec3d28efe746b6437c9/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java#L485

The lack of `"` caused a JsonSyntaxException exception when the host response had a password (e.g. if the host has out of band enabled e.g. IPMI or Redfish).

Fixes: #4355

## Types of changes

## How Has This Been Tested?

- JUnit test
- manual tests

### DIFF
--- a/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
+++ b/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
@@ -482,7 +482,12 @@ public class AsyncJobManagerImpl extends ManagerBase implements AsyncJobManager,
                 if (result.contains(pattern)) {
                     String[] resp = result.split(pattern);
                     String psswd = resp[1].toString().split(",")[0];
-                    result = resp[0] + pattern + psswd.replace(psswd.substring(2, psswd.length() - 1), "*****\"") + "," + resp[1].split(",", 2)[1];
+                    if (psswd.endsWith("}")) {
+                        psswd = psswd.substring(0, psswd.length() - 1);
+                        result = resp[0] + pattern + psswd.replace(psswd.substring(2, psswd.length() - 1), "*****") + "}," + resp[1].split(",", 2)[1];
+                    } else {
+                        result = resp[0] + pattern + psswd.replace(psswd.substring(2, psswd.length() - 1), "*****") + "," + resp[1].split(",", 2)[1];
+                    }
                 }
             }
         }
--- a/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
+++ b/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
@@ -475,7 +475,7 @@ public class AsyncJobManagerImpl extends ManagerBase implements AsyncJobManager,
         return job;
     }
 
-    private String obfuscatePassword(String result, boolean hidePassword) {
+    public String obfuscatePassword(String result, boolean hidePassword) {
         if (hidePassword) {
             String pattern = "\"password\":";
             if (result != null) {
--- a/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
+++ b/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
@@ -482,7 +482,7 @@ public class AsyncJobManagerImpl extends ManagerBase implements AsyncJobManager,
                 if (result.contains(pattern)) {
                     String[] resp = result.split(pattern);
                     String psswd = resp[1].toString().split(",")[0];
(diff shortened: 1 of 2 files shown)


GabrielBrascher:
@blueorangutan package

DaanHoogland (approved):
lgtm

DaanHoogland:
@blueorangutan test

DaanHoogland:
did work @GabrielBrascher 
```
======================================================================
ERROR: Tests degraded HA state when agent is stopped/killed
  File "/marvin/tests/smoke/test_hostha_kvm.py", line 355, in test_hostha_kvm_host_degraded
    self.assertIssueCommandState('ON', 'On')
  File "/marvin/tests/smoke/test_hostha_kvm.py", line 493, in assertIssueCommandState
    self.issuePowerActionCmd(command)
  File "/marvin/tests/smoke/test_hostha_kvm.py", line 539, in issuePowerActionCmd
    raise e
Exception: Job failed: {jobprocstatus : 0, created : u'2020-10-08T02:10:21+0000', completed : u'2020-10-08T02:10:21+0000', cmd : u'org.apache.cloudstack.api.command.admin.outofbandmanagement.IssueOutOfBandManagementPowerActionCmd', userid : u'a8138930-08c3-11eb-8574-1e00ca0138c9', jobstatus : 2, jobid : u'f8439839-f821-4bd6-b7a3-820345992038', jobresultcode : 530, jobresulttype : u'object', jobresult : {errorcode : 530, errortext : u'com.google.gson.stream.MalformedJsonException: Unterminated object near "password":"p*****"","action":"ON","desc'}, accountid : u'a812be97-08c3-11eb-8574-1e00ca0138c9'}
```
note that in this case the replacement yield a double double quote, which is one in excess.

GabrielBrascher:
@DaanHoogland what about removing the password from the JSON response?

DaanHoogland:
@GabrielBrascher If sometimes an extra '\"' is added and sometimes not, this is still going to be an issue if you try to remove it. I'd rather replace the exact "something" with "****" or "s***". it seems that we are brushing a bug under the carpet otherwise.

framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
```diff
@@ -475,14 +475,14 @@ public AsyncJob queryJob(final long jobId, final boolean updatePollTime) {
...
-    private String obfuscatePassword(String result, boolean hidePassword) {
+    public String obfuscatePassword(String result, boolean hidePassword) {
         if (hidePassword) {
             String pattern = "\"password\":";
             if (result != null) {
                 if (result.contains(pattern)) {
                     String[] resp = result.split(pattern);
                     String psswd = resp[1].toString().split(",")[0];
-                    result = resp[0] + pattern + psswd.replace(psswd.substring(2, psswd.length() - 1), "*****") + "," + resp[1].split(",", 2)[1];
+                    result = resp[0] + pattern + psswd.replace(psswd.substring(2, psswd.length() - 1), "*****\"") + "," + resp[1].split(",", 2)[1];
```
rhtyd, line 485: Why not complete replace with the `****`?

GabrielBrascher:
@DaanHoogland I proposed removing the password because I don't see much of use in having it on the response, it seems that does not aggregate relevant information.

DaanHoogland:
It is adding value but only if having a password is optional, @GabrielBrascher. And i think hiding it should be optional as well as a private cloud operator might opt for having it visible. I am not going to argue that anymore, btw.

framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
GabrielBrascher: I am not sure if replacing the password with `****` would then affect the intention of the password response. 
I think that the idea was to offer admins a hint on the password.

But that is good to keep in mind, thanks for the heads up @rhtyd. What do you think @DaanHoogland?

GabrielBrascher:
@DaanHoogland I found out why some cases the regex failed and some not. Updated code to fix it (for good, I hope). Also added test case addressing both situations.

The regex uses comma to set the split separator: `String psswd = resp[1].toString().split(",")[0];`
This works on cases such as:
```
...,"password":"password","action":"OFF"...
```

However, the specific case that I got with OOBM it has the following JSON String:

```
"username":"root","password":"password"},"resourcestate":"PrepareForMaintenance","hahost":false
```

With that said, the code has been updated to deal with both cases: `"password":"password"},` and `"password":"password",`.

GabrielBrascher:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@rhtyd no further comments?

rhtyd (approved):

Issue #4292, opened 2020-08-27
VMware: template VMs not removed during storage GC/cleanup


andrijapanicsb:
Current master (testing specifically PR #4250), vSphere 6.7 (but I believe not important, as the same has been observed in earlier VMware versions)

Issue: when all VMs from a specific template are deleted from a particular Primary Storage (Datastore), the template should be also removed from the Primary Storage after the storage GC kicks in - but this never happens, and template-VMs (with UUID-like names) are left on the Primary Storage, consuming space and junk-ing around.

![image](https://user-images.githubusercontent.com/45762285/91502101-8d996e00-e8c7-11ea-89a8-62e5f348c219.png)

rhtyd:
With @Spaceman1984 to attempt

Spaceman1984:
I'm taking a look at this

rhtyd:
cc #4307 now

andrijapanicsb:
this is true for every ACS release in past few years
deploy a VM from a new template, notice the template-VM created (UUID-like name).
destroy all VMs from that template
reduce the storage cleanup period (global settings to i..e. 60sec) and restart mgmt
once it runs, nothing is removed from the primarty storage, though records ARE removed from the template_spool_ref

Spaceman1984:
I found the code where the problem is in VmwareResource.java. 
```
public Answer execute(DestroyCommand cmd) { ...

if (vmMo != null && vmMo.isTemplate()) {
    if (s_logger.isInfoEnabled()) {
        s_logger.info("Destroy template volume " + vol.getPath());
    }
    vmMo.destroy();
    } else {
        if (s_logger.isInfoEnabled()) {
            s_logger.info("Template volume " + vol.getPath() + " is not found, no need to delete.");
        }
}
```
Volumes will never be deleted by looking at this. (`if (vmMo != null && vmMo.isTemplate()) {`)

The specific line of code was added with this issue https://issues.apache.org/jira/browse/CLOUDSTACK-8405
I will have to run through this issue to fix it correctly I think the logic for deciding whether to delete a volume or template should be in a different place and not here.

sureshanaparti:
@Spaceman1984 `vmMo`'s objects can be never be a volume, it is either a VM or Template.

Spaceman1984:
Ok, @sureshanaparti, it's not volume, it's a VM, but, but not a template and it is failing on that line of code. If it is supposed to be a template, then that could be the issue. I will need to do more digging.

PR #4389, opened 2020-10-07
Fixed vm-templates not being removed from primary storage with storag…

Spaceman1984:
…e garbage collection

## Description

Fixes: #4292 

## Types of changes

## How Has This Been Tested?

This has been tested by changing the storage cleanup global var to 30 seconds, creating a single instance from a template, removing the instance and checking if the template gets removed from Primary Storage.

### DIFF
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -6785,6 +6785,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                 if (s_logger.isInfoEnabled()) {
                     s_logger.info("Destroy template volume " + vol.getPath());
                 }
+                vmMo.markAsVirtualMachine(hyperHost.getHyperHostOwnerResourcePool(), hyperHost.getMor());
                 vmMo.destroy();
             } else {
                 if (s_logger.isInfoEnabled()) {
--- a/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
+++ b/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
@@ -407,7 +407,7 @@ public class VirtualMachineMO extends BaseMO {
     }
 
     public void markAsVirtualMachine(ManagedObjectReference resourcePool, ManagedObjectReference host) throws Exception {
-        _context.getService().markAsVirtualMachine(_mor, pool, host);
+        _context.getService().markAsVirtualMachine(_mor, resourcePool, host);
     }
 
     public boolean isTemplate() throws Exception {
--- a/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
+++ b/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
@@ -406,7 +406,7 @@ public class VirtualMachineMO extends BaseMO {
         _context.getService().markAsTemplate(_mor);
     }
 
-    public void markAsVirtualMachine(ManagedObjectReference pool, ManagedObjectReference host) throws Exception {
+    public void markAsVirtualMachine(ManagedObjectReference resourcePool, ManagedObjectReference host) throws Exception {
         _context.getService().markAsVirtualMachine(_mor, pool, host);
     }
 
--- a/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
+++ b/vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
@@ -406,6 +406,10 @@ public class VirtualMachineMO extends BaseMO {
         _context.getService().markAsTemplate(_mor);
     }
 
+    public void markAsVirtualMachine(ManagedObjectReference pool, ManagedObjectReference host) throws Exception {
+        _context.getService().markAsVirtualMachine(_mor, pool, host);
+    }
+
     public boolean isTemplate() throws Exception {
         VirtualMachineConfigInfo configInfo = getConfigInfo();
         return configInfo.isTemplate();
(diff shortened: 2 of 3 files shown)


Spaceman1984:
@blueorangutan package

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -6785,6 +6785,7 @@ public Answer execute(DestroyCommand cmd) {
                 if (s_logger.isInfoEnabled()) {
                     s_logger.info("Destroy template volume " + vol.getPath());
                 }
+                vmMo.markAsVirtualMachine(hyperHost.getHyperHostOwnerResourcePool(), hyperHost.getMor());
```
sureshanaparti, line 6788: @Spaceman1984 why to mark as VirtualMachine before destroying it ? will the destroy operation fail without it?

vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
```diff
@@ -406,6 +406,10 @@ public void markAsTemplate() throws Exception {
         _context.getService().markAsTemplate(_mor);
     }
 
+    public void markAsVirtualMachine(ManagedObjectReference pool, ManagedObjectReference host) throws Exception {
```
sureshanaparti, line 409:
```suggestion
    public void markAsVirtualMachine(ManagedObjectReference resourcePool, ManagedObjectReference host) throws Exception {
```

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
Spaceman1984: Yes, it did @sureshanaparti,

Spaceman1984:
@blueorangutan test ?

vmware-base/src/main/java/com/cloud/hypervisor/vmware/mo/VirtualMachineMO.java
```diff
@@ -406,6 +406,10 @@ public void markAsTemplate() throws Exception {
         _context.getService().markAsTemplate(_mor);
     }
 
+    public void markAsVirtualMachine(ManagedObjectReference resourcePool, ManagedObjectReference host) throws Exception {
+        _context.getService().markAsVirtualMachine(_mor, pool, host);
```
sureshanaparti, line 410: _pool_ => _resourcePool_

Spaceman1984: Fixed

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
sureshanaparti: I see delete/destory operation works for VM & Template in the vCenter. so, I think _destroyTask()_ API can be used with the template mor as well, no need to explicitly mark as VM. What are the errors seen in the logs for failed destroy task without marking as VM?

sureshanaparti: If _destroy()_ works only when marked as a VM, then the same should be ensured across the usages of vmMO _destroy()_

Spaceman1984: The error was:
Unable to delete template with snapshots - convert the template to a Virtual Machine first.

Spaceman1984:
@blueorangutan test centos7  vmware-67u3

Spaceman1984:

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
rhtyd: @Spaceman1984 @sureshanaparti - in case of linked clones, it makes sense (the error). Do we/CloudStack check that template is not used by any of the VMs, prior to removing it. In that case it perfectly makes sense, if we don't delete the template from primary storage. Thoughts?

sureshanaparti: Yes, in this case, better not to mark the VM as template while spooling it from the secondary storage. So, no need to mark it back to VM for deletion.

Spaceman1984: The error happened for all templates I tested, even systemvm. I speculate that it is referring to the internal command that vmware calls when you click on the delete button from the UI.

Spaceman1984: > Do we/CloudStack check that template is not used by any of the VMs, prior to removing it. 

Yes

nvazquez: Looks good. I tried `vmMo.removeAllSnapshots();` before the destroy but fails. I wonder if templates GC worked before not being marked as templates.

nvazquez (approved):
LGTM

sureshanaparti (approved):
Changes LGTM

DaanHoogland:
@Spaceman1984 as this is not a draft-pr and there are two lgtm; is this good to go or does it need more testing? (cc @borisstoyanov )

rhtyd (approved):

andrijapanicsb:
@Spaceman1984 do you want to advise, in form of the official documentation AND here - how one can fix his env, by marking the existing "template VMs" (which are just VMs in vSphere) as real VMware templates in vSphere - to ensure a proper GC next time for all the old garbage

If I understood this PR correctly, this should be possible - and should allow for an automatical garbage cleanup of tons of older "template VMs" from vSphere (i.e. this PR, afaik, will work only when a new VM is deployed for a VERY first time from a template, as we mark that template VM as a template in vSphere)

cc @rhtyd and @nvazquez for opinion

Spaceman1984:
@andrijapanicsb, I agree with @nvazquez.

PR #4407, opened 2020-10-14
packaging: enable Parallel Collector GC for management server

rhtyd:
The default GC algorithm G1 that is enabled by default with Java11
serves well on multiprocessor machines with large amount of memory where
GC is probablistic with low pauses, where response time is more
important than throughput and GC is kept shorter.

The CloudStack management server is largely a multi-threaded server
application that handles and orchestrates several network requests, and
has the default max. heap size of only 2G that can be considered a
small/medium application from a heap size perspective. Perhaps a more
aggresive GC algorithm such as ParallelGC as used in Java8 and before
(that is previous CloudStack releases) would serve better for throughput
and cause more aggressive GC.

This PR propose a change in default GC algorithm to avoid OOM issues.

Reference: https://docs.oracle.com/en/java/javase/11/gctuning/available-collectors.html#GUID-13943556-F521-4287-AAAA-AE5DE68777CD

## Types of changes

# How this was tested

Heap activity for Java11 with G1 (default GC algorithm):
Screenshot from 2020-10-15 04-03-40

Heap activity seen after switching to ParallelGC:
Screenshot from 2020-10-15 06-19-04
... and over time with some constant load (listApis calls to simulate Primate logins every 1s, all GC/heap graphs look stable):
Screenshot from 2020-10-15 10-47-16

Heap, thread etc activities with the ParallelGC and max pause time setting (500ms):
Screenshot from 2020-10-15 16-00-56

### DIFF
--- a/packaging/systemd/cloudstack-management.default
+++ b/packaging/systemd/cloudstack-management.default
@@ -15,7 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-JAVA_OPTS="-Djava.security.properties=/etc/cloudstack/management/java.security.ciphers -Djava.awt.headless=true -Dcom.sun.management.jmxremote=false -XX:+UseParallelGC -Xmx2G -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/var/log/cloudstack/management/ -XX:ErrorFile=/var/log/cloudstack/management/cloudstack-management.err "
+JAVA_OPTS="-Djava.security.properties=/etc/cloudstack/management/java.security.ciphers -Djava.awt.headless=true -Dcom.sun.management.jmxremote=false -Xmx2G -XX:+UseParallelGC -XX:MaxGCPauseMillis=500 -XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=/var/log/cloudstack/management/ -XX:ErrorFile=/var/log/cloudstack/management/cloudstack-management.err "
 
 CLASSPATH="/usr/share/cloudstack-management/lib/*:/etc/cloudstack/management:/usr/share/cloudstack-common:/usr/share/cloudstack-management/setup:/usr/share/cloudstack-management:/usr/share/java/mysql-connector-java.jar"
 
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

rhtyd:
Smoketests against vmware kicked

sureshanaparti:
@rhtyd Good to see change in the GC for management server process. Even though Parallel GC is the fastest, there are some side effects of using Parallel GC.

- Parallel GC is the default GC in Java 8, which is changed to G1 from Java 9 [1], due to higher GC pauses.
- All application threads are stopped during GC, can impact response time / latency. Check some use cases for Parallel GC here [2].

[1] http://openjdk.java.net/jeps/248
[2] https://www.informit.com/articles/article.aspx?p=2496621&seqNum=2

I think, it is better to control such pause times using option "-XX:MaxGCPauseMillis", when we opt for Parallel GC.

rhtyd:
@sureshanaparti okay I'll add that and do some tests, however, the max. heap size is very small (2G) the pauses would affect memory-heavy application such as databases or some sort of application that have huge heaps (10-100s of GBs). I think given CloudStack is largely a network application with small heap of 2G (default) even the pauses won't affect it (or its threads) much.

rhtyd:
@sureshanaparti can you check the changes now? I've added the `-XX:MaxGCPauseMillis=500` Thanks

@blueorangutan package

rhtyd:

rhtyd:
Let's do one KVM too
@blueorangutan test

DaanHoogland (approved):
I'm ok, it seems that users/operators can tweak this however they wish so I don't think we are hurting anybody.

rhtyd:
Heap status seen under control, after over 24hrs:
Screenshot from 2020-10-16 18-02-27

Pearl1594 (approved):
LGTM

nvazquez (approved):
LGTM

harikrishna-patnala (approved):
I've been keeping an eye on a setup with these changes and MS is stable. So LGTM.

rhtyd:
After 94+hrs with the changes:
Screenshot from 2020-10-19 13-06-02

PR #4412, opened 2020-10-19
Validating type parameter and including all types

davidjumani:
## Description
Validating the `type` parameter for the listNetworks API.
Also allowing all type to be passed

## Types of changes

## How Has This Been Tested?
```
(localhost) 🐱 > list networks type=usolated
🙈 Error: (HTTP 431, error code 4350) Unexpected Guest type : usolated
list networks type=all

(localhost) 🐱 > list networks type=all filter=name
{
............
}
```

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/network/ListNetworksCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/network/ListNetworksCmd.java
@@ -36,6 +36,7 @@ import org.apache.cloudstack.api.response.ZoneResponse;
 
 import com.cloud.network.Network;
 import com.cloud.utils.Pair;
+import com.google.common.base.Strings;
 
 @APICommand(name = "listNetworks", description = "Lists all available networks.", responseObject = NetworkResponse.class, responseView = ResponseView.Restricted, entityType = {Network.class},
         requestHasSensitiveInfo = false, responseHasSensitiveInfo = false)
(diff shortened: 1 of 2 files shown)


davidjumani:
@blueorangutan package

api/src/main/java/org/apache/cloudstack/api/command/user/network/ListNetworksCmd.java
```diff
@@ -101,7 +102,13 @@ public Long getZoneId() {
...
 
     public String getGuestIpType() {
-        return guestIpType;
+        if (!Strings.isNullOrEmpty(guestIpType)) {
+            if (guestIpType.equalsIgnoreCase("all")) {
+                return null;
+            }
+            return Network.GuestType.fromValue(guestIpType).toString();
+        }
+        return null;
```
DaanHoogland, line 111: so in the end empty and "all" are treated the same?

DaanHoogland (approved):
not sure why we need the "all" but code looks good

api/src/main/java/org/apache/cloudstack/api/command/user/network/ListNetworksCmd.java
davidjumani: that's right!

davidjumani:
@blueorangutan test

GabrielBrascher (approved):
Code LGTM.

I have not much clue on why we need "all" @DaanHoogland.  What about simply remove "all" and document it? (maybe on another PR, this would require some discussion)

davidjumani:
I think that all should be there as a filter. Will be useful in Primate to simplify the existing code once merged

DaanHoogland:
I'm fine with leaving it there

PR #4418, opened 2020-10-21
Create Event in case of OOBM failure

GabrielBrascher:
## Description

Out of band power operations are constantly issued to assess the host power status. However, if such actions fail the Admins receive a warning message on the log. In order to help Admins to easily track such failures, this PR creates an Event each time an Out-of-band power operation fails.

## Types of changes

### DIFF
--- a/server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
+++ b/server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
@@ -47,11 +47,11 @@ public class PowerOperationTask implements Runnable {
         try {
             service.executePowerOperation(host, powerOperation, null);
         } catch (Exception e) {
-            LOG.warn(String.format("Out-of-band management background task operation=%s for host id=%d name=%s failed with: %s",
-                    powerOperation.name(), host.getId(), host.getName(), e.getMessage()));
+            LOG.warn(String.format("Out-of-band management background task operation=%s for host %s failed with: %s",
+                    powerOperation.name(), host.getName(), e.getMessage()));
 
             String eventMessage = String
-                    .format("Error while issuing out-of-band management action (%s) for host (id: %d, name: %s)", powerOperation.name(), host.getId(), host.getName());
+                    .format("Error while issuing out-of-band management action %s for host: %s", powerOperation.name(), host.getName());
 
             ActionEventUtils.onCreatedActionEvent(CallContext.current().getCallingUserId(), CallContext.current().getCallingAccountId(), EventVO.LEVEL_WARN,
                     EventTypes.EVENT_HOST_OUTOFBAND_MANAGEMENT_ACTION, true, eventMessage);
--- a/server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
+++ b/server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
@@ -17,7 +17,11 @@
 
 package org.apache.cloudstack.outofbandmanagement;
 
+import com.cloud.event.ActionEventUtils;
+import com.cloud.event.EventTypes;
+import com.cloud.event.EventVO;
 import com.cloud.host.Host;
+import org.apache.cloudstack.context.CallContext;
 import org.apache.log4j.Logger;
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
```diff
@@ -43,8 +53,26 @@ public void run() {
         try {
             service.executePowerOperation(host, powerOperation, null);
         } catch (Exception e) {
-            LOG.warn(String.format("Out-of-band management background task operation=%s for host id=%d failed with: %s",
-                    powerOperation.name(), host.getId(), e.getMessage()));
+            LOG.warn(String.format("Out-of-band management background task operation=%s for host id=%d name=%s failed with: %s",
+                    powerOperation.name(), host.getId(), host.getName(), e.getMessage()));
```
wido, line 57: In general in CloudStack we try to not log the 'hostid' as it doesn't say anything. The ID can only be fetched from the DB and not from the API and thus UI.

server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
```diff
@@ -43,8 +53,26 @@ public void run() {
...
+
+            if(CREATE_EVENT_ON_OOBM_FAILURE.value()) {
+                String eventMessage = String.format("Error while issuing out-of-band management action (%s) for host (id: %d, name: %s)",
+                        powerOperation.name(), host.getId(), host.getName());
```
wido, line 61: Same applies here

server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
```diff
@@ -17,16 +17,26 @@
...
 
-public class PowerOperationTask implements Runnable {
+public class PowerOperationTask implements Runnable, Configurable {
     public static final Logger LOG = Logger.getLogger(PowerOperationTask.class);
 
     final private OutOfBandManagementService service;
     final private Host host;
     final private OutOfBandManagement.PowerOperation powerOperation;
 
+    public static final ConfigKey<Boolean> CREATE_EVENT_ON_OOBM_FAILURE = new ConfigKey<Boolean>("Advanced", Boolean.class, "create.event.on.oob.failure", "true",
```
wido, line 36: I know there is no naming scheme here, but isn't it easier if OOB related config keys are prefixed with *oob*?

<pre>oob.create.event.on.failure</pre>

For example

server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
GabrielBrascher: Makes sense, I am going to keep only the host name then.

server/src/main/java/org/apache/cloudstack/outofbandmanagement/PowerOperationTask.java
GabrielBrascher: Thanks for the suggestion, @wido.
I am removing that global settings config key. There are too many config keys already, I am leaving to add a new one only if it is really necessary.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@DaanHoogland I took the liberty to aim this one to 4.15.
Please let me know if this one is getting too late then I move it to 4.16 ;)

DaanHoogland:
no tests added but for regression,
@blueorangutan test

rhtyd (approved):
LGTM - did not test it

wido (approved):
LGTM

Admins should be notified if CloudStack can't reach the OOB of a node. As this will be used not very often it's good that CloudStack keeps testing the reachability of the OOB so that when it's needed it also works.

GabrielBrascher:
@rhtyd @DaanHoogland @wido @PaulAngus here follows a screenshot with the warning events:
image

DaanHoogland:
@PaulAngus can we put this in for 4.15?

DaanHoogland:
@PaulAngus very small and meets all criteria, merging

Issue #3392, opened 2019-06-10
Dedicate VMware hosts are still used for template seeding


rhtyd:
When a VMware host is dedicated to an account, the host may still be used for preparing templates. A dedicated host owner/account may not prefer that their host is used for template preparations.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

##### COMPONENT NAME

~~~
VMware, Storage
~~~

##### CLOUDSTACK VERSION

~~~
4.9.x, 4.11.x
~~~

rhtyd:
cc @Spaceman1984 - good to have

Spaceman1984:
@rhtyd what do you mean by "preparing templates"? Is this the process of adding an image template to Cloudstack and then the secondary storage agent does the download?

Would that then mean that a dedicated host should not have the system VMs running on it?

rhtyd:
refer prepareTemplate API

PR #4428, opened 2020-10-27
Moved dedicated hosts to the end of the resultset when selecting an e…

Spaceman1984:
…ndpoint

## Description

When Cloudstack selects an endpoint to do various storage-related tasks,  a dedicated host will be lower in priority.

Fixes: #3392 
## Types of changes

## How Has This Been Tested?

Dedicate a host to an account and then call the prepareTemplate API - depending on your setup - if you have 2 hosts and the 1 is dedicated, the dedicated host should not do the seeding of the template.
If there is only 1 host, then it will do the seeding.


engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -128,9 +131,15 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
...
+        List<Long> dedicatedHosts = dedicatedResourceDao.listAllHosts();
+
         // TODO: order by rand() is slow if there are lot of hosts
         sbuilder.append(") t where t.value<>'true' or t.value is null");    //Added for exclude cluster's subquery
-        sbuilder.append(" ORDER by rand() limit 1");
+        sbuilder.append(" ORDER by ");
+        for (Long hostId: dedicatedHosts){ // put dedicated hosts at the end of the result set
+            sbuilder.append("field(t.id, '" + hostId +"')," );
+        }
+        sbuilder.append(" rand() limit 1");
```
DaanHoogland, line 142: if we randomise and limit to one result, what does it matter what the ordering of the set is?

Spaceman1984: We just don't want dedicated hosts ending up being the one selected.

Spaceman1984: The limit is only applied after the sorting.

DaanHoogland: ok, really not very clear code than. What would the resulting sql of this be?

Spaceman1984:
```
select t.id from (
    select h.id, cd.value 
    from host h 
    join storage_pool_host_ref s on h.id = s.host_id  
    join cluster c on c.id=h.cluster_id 
    left join cluster_details cd on c.id=cd.cluster_id 
    and cd.name='cluster.storage.operations.exclude' 
    where h.status = 'Up' 
    and h.type = 'Routing' 
    and h.resource_state = 'Enabled' 
    and s.pool_id = 1
    and h.data_center_id = 1
) t 
where t.value<>'true' 
or t.value is null ORDER BY field(t.id, '2'), field(t.id, '5'), rand()
```

DaanHoogland: so why this complicated way to filter out specific values? It isn't very readible and we can add tham to the where clause as `not t.id in (<hostID>, ...)`

Spaceman1984: We don't want to filter them out, just reorder because of the case of all hosts being dedicated.

DaanHoogland: ok, next question, why two separate field() calls and not a combined one?

Spaceman1984: Yes, I have been thinking about that, I may be able to combine this into 1 field call. - I'll see what I can do.

rhtyd:
@Spaceman1984 pl run packaging and test

Spaceman1984:
@blueorangutan package

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -128,9 +132,24 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
...
+            sbuilder.append("field(t.id, ");
+            Iterator<Long> hostIt = dedicatedHosts.iterator();
+            while (hostIt.hasNext()) { // put dedicated hosts at the end of the result set
+                sbuilder.append("'" + hostIt.next() + "'");
+                if (hostIt.hasNext()) {
+                    sbuilder.append(",");
+                }
+            }
+            sbuilder.append(")," );
+        }
```
DaanHoogland, line 151: can this be a separate method (i.e. `moveDedicatedHostsToBackForExclusion(sbuilder,dedicatedHosts);`) ?

rhtyd:

rhtyd (approved):
LGTM, did not test it

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
DaanHoogland: pretty please, @Spaceman1984

Spaceman1984: I'll see what I can do @DaanHoogland

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -128,9 +132,24 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
             }
         }
 
+        List<Long> dedicatedHosts = dedicatedResourceDao.listAllHosts();
```
sureshanaparti, line 135: @Spaceman1984 get the dedicated hosts list based on the scope here instead all hosts

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -128,9 +132,24 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
...
+
         // TODO: order by rand() is slow if there are lot of hosts
         sbuilder.append(") t where t.value<>'true' or t.value is null");    //Added for exclude cluster's subquery
-        sbuilder.append(" ORDER by rand() limit 1");
+        sbuilder.append(" ORDER by ");
+        if (dedicatedHosts.size() > 0) {
+            Collections.shuffle(dedicatedHosts); // Randomize dedicated hosts as well.
+            sbuilder.append("field(t.id, ");
+            Iterator<Long> hostIt = dedicatedHosts.iterator();
+            while (hostIt.hasNext()) { // put dedicated hosts at the end of the result set
```
sureshanaparti, line 144: @Spaceman1984 the ordering based on the dedicated host ids here would put them on the top, can you confirm the result here?

Spaceman1984: @sureshanaparti  I ran the SQL multiple times, the dedicated hosts were always at the end of the result set.

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
Spaceman1984: Done

Spaceman1984:
@blueorangutan package

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
sureshanaparti: > @sureshanaparti I ran the SQL multiple times, the dedicated hosts were always at the end of the result set.

Let's consider host ids: 3, 4 are dedicated, then the query does " _ORDER by field(t.id, '3','4'), rand() limit 1_". Am I correct?

Spaceman1984: Yes @sureshanaparti

sureshanaparti: thanks for the confirmation @Spaceman1984

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
Spaceman1984: This call doesn't get all hosts, it gets all dedicated hosts, then builds up a random list of dedicated hosts. To remove some of the entries I will have to iterate through another list of hosts from another resultset which I think would cause more overhead than doing it this way. Or am I misunderstanding your intention?

sureshanaparti: > This call doesn't get all hosts, it gets all dedicated hosts, which is then used to build up a random list of dedicated hosts to move to the back of the result set. To remove some of the entries I will have to iterate through another list of hosts from another resultset which I think would cause more overhead than doing it this way. Or am I misunderstanding your intention?

@Spaceman1984 I mean to say, pick only the dedicated hosts based on the scope. For example, whenever the pool scope is Cluster, pick all the dedicated hosts in that cluster only, not the entire list of dedicated hosts. Got it?

Spaceman1984: Yes, I understand. Making this change is not going to reduce the size of the result set, it will reduce the number of ids to order against. I don't think adding this complexity is worth the effort as this functionality is not available in the DAO, I will have to write a few more functions to make that happen. If you feel that we need this, then I will go ahead.

Spaceman1984: I mean, if a host is not in the order list it gets ignored.

sureshanaparti: @Spaceman1984 New method in dao would be better. Consider a case with 2 clusters, one has 'N' dedicated hosts and the other has None. When you process the pool with second cluster, you have to iterate over the 'N' dedicated hosts (belongs to first cluster) without any use go them. and also, you are passing them to the sql knowingly that these host ids doesn't exist in the result set.

DaanHoogland:
@sureshanaparti are all you concerns addressed?

DaanHoogland:
@blueorangutan test

sureshanaparti:
> @sureshanaparti are all you concerns addressed?

@DaanHoogland changes not updated yet. advised to get the dedicated hosts for the given scope, using new dao method.

DaanHoogland:
ping @Spaceman1984

Spaceman1984:
I'll look into it @DaanHoogland

DaanHoogland:
@blueorangutan package

DaanHoogland:
@sureshanaparti can you have look again?
@blueorangutan test

DaanHoogland (approved):
code lgtm

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -115,22 +123,46 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
...
+                    for (HostVO hostVO: clusterHosts){
+                        dedicatedHosts.add(hostVO.getId());
+                    }
+                }
             } else if (scope.getScopeType() == ScopeType.ZONE) {
                 sbuilder.append(" and h.data_center_id = ");
                 sbuilder.append(scope.getScopeId());
+                DedicatedResourceVO dedicatedHost = dedicatedResourceDao.findByZoneId(scope.getScopeId());
+                if (dedicatedHost != null){
+                    List<HostVO> zoneHosts = hostDao.findByClusterId(scope.getScopeId());
```
sureshanaparti, line 147: @Spaceman1984 scope id in this case is zone id, and the hosts are found using cluster id method, please check.

Spaceman1984: Right, copy-paste error. I'll fix it.

DaanHoogland: good catch @sureshanaparti . guess my squared-eyes filtered this out 👍

Spaceman1984:
@blueorangutan package

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -115,22 +123,46 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
...
             } else if (scope.getScopeType() == ScopeType.CLUSTER) {
                 sbuilder.append(" and h.cluster_id = ");
                 sbuilder.append(scope.getScopeId());
+                DedicatedResourceVO dedicatedHost = dedicatedResourceDao.findByClusterId(scope.getScopeId());
+                if (dedicatedHost != null){
+                    List<HostVO> clusterHosts = hostDao.findByClusterId(scope.getScopeId());
+                    for (HostVO hostVO: clusterHosts){
+                        dedicatedHosts.add(hostVO.getId());
+                    }
+                }
```
sureshanaparti, line 141: @Spaceman1984 what if cluster is not dedicated, but some hosts in that cluster are dedicated?

Spaceman1984: This is handled on line 155

engine/storage/src/main/java/org/apache/cloudstack/storage/endpoint/DefaultEndPointSelector.java
```diff
@@ -115,22 +123,46 @@ protected EndPoint findEndPointInScope(Scope scope, String sqlBase, Long poolId)
...
+                    List<HostVO> zoneHosts = hostDao.findByDataCenterId(scope.getScopeId());
+                    for (HostVO hostVO: zoneHosts){
+                        dedicatedHosts.add(hostVO.getId());
+                    }
+                }
             }
         }
+        // We didn't find any hosts specifically dedicated to a zone or cluster
+        if (dedicatedHosts.size() == 0) {
+            dedicatedHosts = dedicatedResourceDao.listAllHosts();
```
sureshanaparti, line 156: @Spaceman1984  the scope is not considered here, this returns all the dedicated hosts same as the earlier case.

If the cluster is not dedicated and some hosts in that cluster are dedicated, then get the dedicated hosts in that cluster only, which you can move to the end of the result set. Got me? Didn't see any dao method for this?

Spaceman1984: If the cluster is not dedicated, then we cannot know which cluster to look at.

Spaceman1984: This is not possible

Spaceman1984: Ok, I'll make some changes.

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test

DaanHoogland:
@sureshanaparti can you check if your concerns are met? tnx

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -1672,10 +1672,13 @@ protected boolean sendStop(final VirtualMachineGuru guru, final VirtualMachinePr
...
                 guru.finalizeStop(profile, answer);
+
+                final UserVmVO userVm = _userVmDao.findById(vm.getId());
                 if (vm.getType() == VirtualMachine.Type.User) {
-                    final UserVmVO userVm = _userVmDao.findById(vm.getId());
-                    userVm.setPowerState(PowerState.PowerOff);
-                    _userVmDao.update(userVm.getId(), userVm);
+                    if (userVm != null){
+                        userVm.setPowerState(PowerState.PowerOff);
+                        _userVmDao.update(userVm.getId(), userVm);
```
sureshanaparti, line 1680: @Spaceman1984 these changes are part of your fix?

Spaceman1984: Unrelated, just a missed null pointer check I picked up while testing.

sureshanaparti (approved):
code LGTM

PR #4436, opened 2020-10-29
CKS CoreOS EOL update 

Pearl1594:
## Description
With CoreOS having reached End of life, support has been provided CKS to consume systemVM template to deploy kubernetes clusters

## Types of changes


Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan package

engine/schema/src/main/resources/META-INF/db/schema-41400to41500.sql
```diff
@@ -833,3 +833,7 @@ INSERT INTO `cloud`.`guest_os_hypervisor` (uuid,hypervisor_type, hypervisor_vers
 
 -- Fix OS category for Guest OS 'Other PV Virtio-SCSI (64-bit)'
 UPDATE `cloud`.`guest_os` SET category_id = 7 WHERE id = 275 AND display_name = 'Other PV Virtio-SCSI (64-bit)';
+
+ALTER TABLE `cloud`.`user_vm` ADD COLUMN `user_vm_type` varchar(255) DEFAULT "UserVM" COMMENT 'Defines the type of UserVM';
```
sureshanaparti, line 837: @Pearl1594 is this not redundant of _Type_ column in _vm_instance_ table? This _type_ column already used to persist different system VMs  (ConsoleProxy, SecondaryStorageVm, DomainRouter), may be you can use this for different user VMs as well.

engine/schema/src/main/java/com/cloud/vm/UserVmVO.java
```diff
@@ -48,6 +48,9 @@
     @Column(name = "update_parameters", updatable = true)
     protected boolean updateParameters = true;
 
+    @Column(name = "user_vm_type", updatable = true)
```
sureshanaparti, line 51: can use "type" from VMInstanceVO instead?

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
```diff
@@ -824,7 +824,7 @@ private DiskProfile allocateTemplatedVolume(Type type, String name, DiskOffering
         int volumesNumber = 1;
         List<DatadiskTO> templateAsIsDisks = null;
         String configurationId = null;
-        if (template.isDeployAsIs()) {
+        if (template.isDeployAsIs() && vm.getType() != VirtualMachine.Type.SecondaryStorageVm) {
```
sureshanaparti, line 827: please note there is already a specific PR #4437 with this change

engine/schema/src/main/resources/META-INF/db/schema-41400to41500.sql
Pearl1594: Intention of adding a user vm type - as a sub-type of VMInstance type - User is to basically reduce any regression - changing User to other type to accommodate CKS / K8s nodes would possibly require a lot more changes

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
Pearl1594: This was added to verify the fix @sureshanaparti

server/src/main/java/com/cloud/vm/UserVmManager.java
```diff
@@ -55,6 +55,10 @@
 
     static final int MAX_USER_DATA_LENGTH_BYTES = 2048;
 
+    public static enum UserVmType {
```
sureshanaparti, line 58: @Pearl1594  there is a common enum "VirtualMachine.Type" where all the VM types are defined, can use the same?

engine/schema/src/main/resources/META-INF/db/schema-41400to41500.sql
sureshanaparti: Ok @Pearl1594 I think it is better to use single source for defining the instance type.

Pearl1594:
@blueorangutan package

PR #4443, opened 2020-11-05
Fixed VM scaling test on Xenserver

Spaceman1984:
## Description

If a VM is started before setting the global var enable.dynamic.scale.vm to true, scaling is not possible on the specific VM until it is stopped and started again. Setting the global var before starting the VM solves the problem. The template also needs dynamic scaling enabled. But after checking those variables the VM instance would still fail due to the default max vCPUs being set to 16.

Fixes: #4237 

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

Run the Marvin test - scaling works fine.

PR #4447, opened 2020-11-05
Fixed failed to start instances on Xenserver with dynamic scaling due to default max cpus.

Spaceman1984:
## Description

When trying to start an instance with dynamic scaling on a Xenserver Hypervisor with less than 16 cores, the VM fails to start. This is due to the default max value being set to 16. This PR limits the max vCPUs to what the host is capable of.

## Types of changes

## How Has This Been Tested?

Set enable.dynamic.scale.vm to true
Try to start an instance on a Xenserver host with less than 16 cores.

### DIFF
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/XenServerGuru.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/XenServerGuru.java
@@ -98,7 +98,12 @@ public class XenServerGuru extends HypervisorGuruBase implements HypervisorGuru,
         if (userVmVO != null) {
             HostVO host = hostDao.findById(userVmVO.getHostId());
             if (host != null) {
-                to.setVcpuMaxLimit(MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()));
+                // Max cpu cannot be more than host capability
+                if (host.getCpus() < MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()))
+                    to.setVcpuMaxLimit(host.getCpus());
+                else {
+                    to.setVcpuMaxLimit(MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()));
+                }
             }
         }
 
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/XenServerGuru.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/XenServerGuru.java
@@ -99,9 +99,16 @@ public class XenServerGuru extends HypervisorGuruBase implements HypervisorGuru,
             HostVO host = hostDao.findById(userVmVO.getHostId());
             if (host != null) {
                 // Max cpu cannot be more than host capability
-                if (host.getCpus() < MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()))
+                if (host.getCpus() < MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId())) {
+                    String message = String.valueOf(new StringBuilder()
+                            .append("Ignoring global max vCPUs limit (from global setting xen.vm.vcpu.max) on vm ")
+                            .append(vm.getUuid()).append(" and setting VM max vCPU to host pCPU max: ")
+                            .append(host.getCpus())
+                            .append(". Requested number of virtual CPUs exceeds number of host physical CPUs"));
+                    logger.info(message);
                     to.setVcpuMaxLimit(host.getCpus());
-                else {
+                } else {
+                    logger.info("Setting vm: " + vm.getUuid() + " max vCPU limit (from global setting xen.vm.vcpu.max) to: " + MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()));
                     to.setVcpuMaxLimit(MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()));
                 }
             }


Spaceman1984:
@blueorangutan package

Spaceman1984:
@blueorangutan test centos7 xcpng76

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/XenServerGuru.java
```diff
@@ -98,7 +98,12 @@ public VirtualMachineTO implement(VirtualMachineProfile vm) {
         if (userVmVO != null) {
             HostVO host = hostDao.findById(userVmVO.getHostId());
             if (host != null) {
-                to.setVcpuMaxLimit(MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()));
+                // Max cpu cannot be more than host capability
+                if (host.getCpus() < MaxNumberOfVCPUSPerVM.valueIn(host.getClusterId()))
+                    to.setVcpuMaxLimit(host.getCpus());
```
harikrishna-patnala, line 103: I think host.getCpus() get the actual CPU on the physical host, but xenserver host is capable of allocating virtual CPUs more than the actual count.
Here is the list of configuration limits for XS 7.1 version. https://docs.citrix.com/en-us/xenserver/7-1/system-requirements/configuration-limits.html
It says "Virtual CPUs per VM (Linux)" 32. So we might need to set the value according to this.

Spaceman1984: I tested this with a host with 8 CPUs and it failed trying to allocate 16 on XCP-ng 7.6

Spaceman1984: I'll do some testing on Xenserver

Spaceman1984: I tested XCP-NG 7.6 and Xenserver 7.6, I was unable to start a VM with more vCPUs than pCPUs.
XCP-NG
Xenserver

Spaceman1984: Seems like overcommit on xenserver was removed according to this article @harikrishna-patnala https://support.citrix.com/article/CTX236977

harikrishna-patnala: It might also depend on the supported limits on guest OS. but setting the max vcpu limit less than or equal to host cpus does not give complete leverage to scale the VM. Also this setting was there since long time so I think this is something to do with OS type of XCP ng version limitation.

harikrishna-patnala: @Spaceman1984 got it, thanks for the article and yes it seems over committing CPU is disabled.
Do you think we have to also provide some global config "vcpu-unrestricted" to allow overcommit of CPU like xenserver does this way "xe vm-param-set uuid=<uuid> platform:vcpu-unrestricted=true". This gives admins to choose whether to overcommit or not.

Otherwise If we have to go with by setting only actual pCPUs, we need to remove this configuration parameter "xen.vm.vcpu.max" as we dont need it now.

Spaceman1984: I kept the global var in to allow things like if you have a host with 32 pCPUs but want to set a lower max vCPU limit for a VM for the purpose of dynamic scaling.
As for the ability to force overcommit - This is not reccommended in the article and I would think not best practice, that's why I would vote no.

harikrishna-patnala: Okay LGTM.

harikrishna-patnala: please put a log message which explans why we are setting that value, otherwise user may think why a different value is set other than "xen.vm.vcpu.max"

Spaceman1984: Alright, awesome, on it.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

Spaceman1984:
@blueorangutan test centos7 xcpng76

DaanHoogland:
looking by the discussion here, we should move this to 4.16 and make a good functional definition of what is and isn't possible on different xen platforms.

DaanHoogland:
> looking by the discussion here, we should move this to 4.16 and make a good functional definition of what is and isn't possible on different xen platforms.

ping @Spaceman1984 @harikrishna-patnala . cc @rhtyd

Spaceman1984:
@DaanHoogland I have determined this PR is related to #4443, I will merge this branch into that one and then close this PR.

harikrishna-patnala:
> > looking by the discussion here, we should move this to 4.16 and make a good functional definition of what is and isn't possible on different xen platforms.
> 
> ping @Spaceman1984 @harikrishna-patnala . cc @rhtyd

@DaanHoogland agreed to keep it for 4.16 since this makes a substantial change in the existing flow or values assigned.

PR #4448, opened 2020-11-05
Make internaldns1 optional for zone creation

ravening:
## Description

Make internaldns1 optional for zone creation

    As of now internaldns1 is mandatory field while creating
    a zone. Sometimes we dont have internaldns1 and so make
    this field as optional

Since this is optional, the original behaviour is still retained and nothing breaks

## Types of changes

## How Has This Been Tested?

Before the change

```
(local) mgt01 > create zone
Missing arguments: networktype internaldns1 dns1 name
(local) mgt01 >
```

After the change
```
(local) 🐱 > create zone
💩 Missing required parameters:  dns1, name, networktype
(local) 🐱 >
```

In systemvms
```
root@s-175-VM:~# cat /etc/resolv.conf
nameserver 8.8.4.4
nameserver 8.8.8.8
root@s-175-VM:~# vi /var/log/messages
root@s-175-VM:~# cat /etc/dnsmasq-resolv.conf
nameserver 8.8.4.4
nameserver 8.8.8.8
```

After destroying
```
root@s-176-VM:~# cat /etc/resolv.conf
nameserver 8.8.8.8
root@s-176-VM:~#
root@s-176-VM:~# cat /etc/dnsmasq-resolv.conf
nameserver 8.8.8.8
root@s-176-VM:~#
```

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/admin/zone/CreateZoneCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/admin/zone/CreateZoneCmd.java
@@ -58,7 +58,7 @@ public class CreateZoneCmd extends BaseCmd {
     @Parameter(name = ApiConstants.GUEST_CIDR_ADDRESS, type = CommandType.STRING, description = "the guest CIDR address for the Zone")
     private String guestCidrAddress;
 
-    @Parameter(name = ApiConstants.INTERNAL_DNS1, type = CommandType.STRING, required = true, description = "the first internal DNS for the Zone")
+    @Parameter(name = ApiConstants.INTERNAL_DNS1, type = CommandType.STRING, description = "the first internal DNS for the Zone")
     private String internalDns1;
 
     @Parameter(name = ApiConstants.INTERNAL_DNS2, type = CommandType.STRING, description = "the second internal DNS for the Zone")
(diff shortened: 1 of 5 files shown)


DaanHoogland:
to be honest, this has been a wonder of mine for a while, but i'm a bit weary of side effects of not having an internal dns. tnx for the PR anyway.

ravening:
> to be honest, this has been a wonder of mine for a while, but i'm a bit weary of side effects of not having an internal dns. tnx for the PR anyway.

@DaanHoogland you can still have internal dns1 but its optional now.. not removed entirely
sometimes people may not have internal dns and its difficult for them to proceed further if its mandatory

DaanHoogland:
> > to be honest, this has been a wonder of mine for a while, but i'm a bit weary of side effects of not having an internal dns. tnx for the PR anyway.
> 
> @DaanHoogland you can still have internal dns1 but its optional now.. not removed entirely
> sometimes people may not have internal dns and its difficult for them to proceed further if its mandatory

yes, i understand that, @ravening . Would it make sense, to address possible side effects, to set the internal dns to the dns in case it is missing as default?

PaulAngus:
@ravening I don't see the probably with requiring an internal DNS entry.  The operator can always enter the external DNS IP for it if they need to.  I can only see this causing pain elsewhere in the code.

Issue #4449, opened 2020-11-05
vr_cfg.sh use lots of memory failed to start if there are many firewall rules.


echoidcf:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
All version after 4.7 affected
~~~

##### CONFIGURATION

Vmware

##### SUMMARY

Script writing style in /opt/cloud/bin/vr_cfg.sh in VR will consume HUGE memory if there are more than 1,000 firewall rules.

When VR is restarted, cloudstack will pass a aggregation of JSON files to VR and use vr_cfg.sh to unpack it and run python script to load them. The problem is cloudstack will generate each JSON file to one-line-file, which is ok. But vr_cfg.sh is written in following style:
~~~
while read line; do
...
done < $cfg
~~~
If JSON file is big, hundreds of kilo-byte for example, there will be a very LOOOOOOOOONG line for the script to read. This will use a lot of memory and result in memory exhausted and will fail the async job of start VR.
And of course , VR will not come up after that.

As a result of my test, 2k firewall rules will cause this problem on a VR with 1GB memory. 6K firewall rules will kill a VR with 8GB memory, and 10k firewall rules will kill the VR even with 16GB memory.

The fix is simple, just change it to
~~~
cat $cfg | while read line; do
...
done
~~~
will fix this problem.

##### STEPS TO REPRODUCE

1. add 2000 firewall rules to a VR with 1GB memory
2. stop and start it

it will fail to start.

##### EXPECTED RESULTS

~~~
VR failed to start
~~~

##### ACTUAL RESULTS

~~~
Job will failed and a memory out log in the management server log.
~~~

DaanHoogland:
tnx for the report @echoidcf . Are you making a PR for this?  and do you think this style of coding is present at more locations?

rhtyd:
This explains why read line is slow (https://stackoverflow.com/questions/13762625/bash-while-read-loop-extremely-slow-compared-to-cat-why); we can explore - https://stackoverflow.com/questions/4642191/read-line-by-line-in-bash-script

DaanHoogland:
@echoidcf I added the PR for you as it is quite simple. Do you have any test results that proof this will help you, as in my understanding the amount of memory consumed in both variations won't differ much.

Also please see #4451 to check if this is the change you are suggesting.

davidjumani:
[...]

PR #4451, opened 2020-11-06
loop optimisation in bash

DaanHoogland:
## Description

A small optimisation as suggested by the user in #4449 . I'm not convinced this is actually reducing the memory usage as they suggest. the code itself seems rathrer risk-free. let's discuss.

Fixes: #4449 

## Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/vr_cfg.sh
+++ b/systemvm/debian/opt/cloud/bin/vr_cfg.sh
@@ -33,7 +33,7 @@ while getopts 'c:' OPTION; do
 esac; done
 
 export DEFER_CONFIG=true
-while read line; do
+cat $cfg | while read line; do
     #comment
     if [[ $line == \#* ]]; then
         continue
@@ -74,7 +74,7 @@ while read line; do
 
     fi
 
-done < $cfg
+done 
 
 # archive the configuration file
 mv $cfg /var/cache/cloud/processed/


DaanHoogland:
@GabrielBrascher @weizhouapache @rhtyd @wido let's discuss this. I've seen issues in large isolated nets on startup but I doubt this is going to be a big help. thoughts?

rhtyd (approved):
LGTM, needs testing and maybe new tests that could simulate that scale (1k+ fw rules?)

wido:
I'm not sure if this fixes it. Not against changing it, but I don't know if this is the root-cause.

davidjumani (approved):
Code works, LGTM

davidjumani:
@rhtyd @DaanHoogland Would splitting the file and processing each part sequentially work ?

DaanHoogland:
@davidjumani I like the idea, do we have a good logical splitting of the file? the problem states that many firewall rules cause the issue , so splitting over functional parts wouldn't solve this.

davidjumani:
@DaanHoogland Assuming that the rules are in multiple lines, loading all the lines in memory and processing them would kill the VR. Instead splitting the file into smaller ones and reading them could solve the memory issue

DaanHoogland:
Again @davidjumani I like the idea. What is the criterium? (no of lines, sections some combination of criterea) can we be sure that lines are self contained and do not require context or order of executing and can we guarantee that?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

weizhouapache:
@DaanHoogland @rhtyd 
I suggest not to merge this pr into 4.15.0.0

rhtyd:
Yes @weizhouapache, it's marked against 4.15.1.0; (I'm) just kicking tests on PRs.

DaanHoogland:
> @DaanHoogland @rhtyd
> I suggest not to merge this pr into 4.15.0.0

I agree, and would say let's not merge at all until we have knowledge if this is making a difference at all.

rhtyd:
@DaanHoogland @weizhouapache @wido should we merge this for 4.15.1.0 (targetting Q2)?

DaanHoogland:
I don't care for this code @rhtyd . I don't know but i don't think it is an improvement/optimisation

weizhouapache:
> I don't care for this code @rhtyd . I don't know but i don't think it is an improvement/optimisation

same to me.

rhtyd:
Considering all I could find, the pipe based thing would be better compared to redirection as it would avoid memory exhaustion; the pipe/buffer synchronisation (https://stackoverflow.com/questions/9553628/piping-and-redirection). Let's merge this in 4.15 and test over next few weeks, we can revisit and revert if necessary.

PR #4466, opened 2020-11-12
VR: fix logging is not working and logs are not appended to /var/log/cloud.log

ustcweizhou:
## Description

There are no logs in /var/log/cloud.log when something are changed in VRs (for example start vms, add/remove rules).

## Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/update_config.py
+++ b/systemvm/debian/opt/cloud/bin/update_config.py
@@ -26,8 +26,6 @@ import os.path
 import configure
 import json
 
-logging.basicConfig(filename='/var/log/cloud.log', level=logging.INFO, format='%(asctime)s  %(filename)s %(funcName)s:%(lineno)d %(message)s')
-
 # first commandline argument should be the file to process
 argc = len(sys.argv)
 if argc != 2 and argc != 3:
--- a/systemvm/debian/opt/cloud/bin/update_config.py
+++ b/systemvm/debian/opt/cloud/bin/update_config.py
@@ -16,17 +16,17 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import logging
+logging.basicConfig(filename='/var/log/cloud.log', level=logging.INFO, format='%(asctime)s  %(filename)s %(funcName)s:%(lineno)d %(message)s')
+
 import sys
 from merge import QueueFile
-import logging
 import subprocess
 from subprocess import PIPE, STDOUT
 import os
 import os.path
 import configure
 import json
-
-logging.basicConfig(filename='/var/log/cloud.log', level=logging.INFO, format='%(asctime)s  %(filename)s %(funcName)s:%(lineno)d %(message)s')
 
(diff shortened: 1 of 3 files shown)


DaanHoogland:
code looks good @weizhouapache . Are we to understand that any modules imported don't get the changed logging config if it is set after importing?

weizhouapache:
> code looks good @weizhouapache . Are we to understand that any modules imported don't get the changed logging config if it is set after importing?

@DaanHoogland yes. actually all python scripts in /opt/cloud/bin/cs does not have logging settings, so there is no log appended to /var/log/cloud.log
update_config.py is the entry to update VR configurations (including iptables rules/nics/configs,etc). 
we have similar code in elder versions which work fine. After systemvm refactoring by @rhtyd , it does not work any more.  I have no idea what caused it.

DaanHoogland (approved):

rhtyd (approved):

rhtyd:
@blueorangutan package

DaanHoogland:
pylint doesn't approve of this, @weizhouapache @rhtyd (see 1st travis job) maybe we can mashup the order some more

systemvm/debian/opt/cloud/bin/update_config.py
```diff
@@ -16,18 +16,18 @@
 # specific language governing permissions and limitations
 # under the License.
 
+import logging
+logging.basicConfig(filename='/var/log/cloud.log', level=logging.INFO, format='%(asctime)s  %(filename)s %(funcName)s:%(lineno)d %(message)s')
+
```
DaanHoogland, line 21: not really helpfull, maybe. we can solve the pylint issue by tickering the order, something like:
```suggestion
import sys
import logging
import subprocess
import os
import os.path
import configure
import json

logging.basicConfig(filename='/var/log/cloud.log', level=logging.INFO, format='%(asctime)s  %(filename)s %(funcName)s:%(lineno)d %(message)s')

from subprocess import PIPE, STDOUT
```
Maybe with more `from x import y`'s. Or have some option on pylint.

weizhouapache:
> pylint doesn't approve of this, @weizhouapache @rhtyd (see 1st travis job) maybe we can mashup the order some more

@DaanHoogland @rhtyd found another fix, tested ok. please review and test it

rhtyd:
@blueorangutan package

rhtyd (approved):
Lgtm, but needs checking if we need to specify explicit log file path

rhtyd:
@blueorangutan test

DaanHoogland:
great @weizhouapache more elegant and the travis errors so far are unrelated. it's easy to overlook the package init possibility, thanks for this.

DaanHoogland:
marking this as major as it stops operators from being able to do proper diagnosis on the system/network

davidjumani (approved):
LGTM
Able to see logs when rules added

Screenshot from 2020-11-20 14-53-01

Issue #3934, opened 2020-03-04
Zones are returned in a random order


dpassante:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API
~~~

##### CLOUDSTACK VERSION

~~~
4.13.0.0
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

`listZones` returns the zones in a "random" order when it has not been explicitly defined with the new sort key feature.

This impacts the correct functioning of the Ansible Cloudstack modules which use the first zone if no zone parameter is given.

##### STEPS TO REPRODUCE

Keep zones unsorted:
~~~
mysql> select id, name, sort_key from data_center;
+----+----------------------------+----------+
| id | name                       | sort_key |
+----+----------------------------+----------+
|  1 | Sandbox-simulator-basic    |        0 |
|  2 | Sandbox-simulator-advanced |        0 |
+----+----------------------------+----------+
2 rows in set (0.00 sec)
~~~

And use a playbook in which the `zone` param is not given.
Several calls to `listZones` also do the trick.

##### EXPECTED RESULTS

A list of zones sorted by SQL ID when no sort key are configured to avoid breaking some existing Ansible deployments.

Ensure that a sort key is automatically incremented for new zones and existing zones in the case of a Cloudstack upgrade could also be a nice feature.

##### ACTUAL RESULTS

~~~
$ for in in `seq 1 1 10`; do cs listZones |grep -m1 name; done
      "name": "Sandbox-simulator-basic",
      "name": "Sandbox-simulator-basic",
      "name": "Sandbox-simulator-basic",
      "name": "Sandbox-simulator-basic",
      "name": "Sandbox-simulator-basic",
      "name": "Sandbox-simulator-basic",
      "name": "Sandbox-simulator-advanced",
      "name": "Sandbox-simulator-advanced",
      "name": "Sandbox-simulator-advanced",
      "name": "Sandbox-simulator-advanced",
~~~

dpassante:
In fact, the "problem" seems to come from the fact that zones are created with the sort key 0 which has priority by default.

~~~
+----+--------------------------------------+----------+
| id | name                                 | sort_key |
+----+--------------------------------------+----------+
|  1 | Sandbox-simulator-basic              |        0 |
|  2 | Sandbox-simulator-advanced           |        0 |
|  3 | ansible-test-dpassante-55511421-zone |        0 |
[...]

PR #4494, opened 2020-11-23
sql: Fix Zones are returned in a random order (#3934)

ustcweizhou:
### Description

This PR updates sort_key to id if sort_key of all zones are 0 (including removed zones, which means sort_key is not set by customers), so zones will be ordered by id in listZones response.

Fixes: #3934

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-41400to41500.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41400to41500.sql
@@ -837,3 +837,6 @@ UPDATE `cloud`.`guest_os` SET category_id = 7 WHERE id = 275 AND display_name = 
 -- Add flag 'hidden' in tables usage_ip_address and cloud_usage
 ALTER TABLE `cloud_usage`.`usage_ip_address` ADD COLUMN `is_hidden` smallint(1) NOT NULL DEFAULT '0' COMMENT 'is usage hidden';
 ALTER TABLE `cloud_usage`.`cloud_usage` ADD COLUMN `is_hidden` smallint(1) NOT NULL DEFAULT '0' COMMENT 'is usage hidden';
+
+-- Fix Zones are returned in a random order (#3934)
+UPDATE `cloud`.`data_center` JOIN (SELECT COUNT(1) AS count FROM `cloud`.`data_center` WHERE `sort_key` != 0) AS tbl_tmp SET `sort_key` = `id` WHERE count = 0;


rhtyd (approved):

rhtyd:
@blueorangutan package

engine/schema/src/main/resources/META-INF/db/schema-41400to41500.sql
```diff
@@ -837,3 +837,6 @@ UPDATE `cloud`.`guest_os` SET category_id = 7 WHERE id = 275 AND display_name =
 -- Add flag 'hidden' in tables usage_ip_address and cloud_usage
 ALTER TABLE `cloud_usage`.`usage_ip_address` ADD COLUMN `is_hidden` smallint(1) NOT NULL DEFAULT '0' COMMENT 'is usage hidden';
 ALTER TABLE `cloud_usage`.`cloud_usage` ADD COLUMN `is_hidden` smallint(1) NOT NULL DEFAULT '0' COMMENT 'is usage hidden';
+
+-- Fix Zones are returned in a random order (#3934)
+UPDATE `cloud`.`data_center` JOIN (SELECT COUNT(1) AS count FROM `cloud`.`data_center` WHERE `sort_key` != 0) AS tbl_tmp SET `sort_key` = `id` WHERE count = 0;
```
DaanHoogland, line 842: excuse my ignorance. why not just 
```
UPDATE `cloud`.`data_center` SET `sort_key` = `id` WHERE `sort_key` = 0;
```
this seems like it going to work as well, but i'm missing the point, i guess. at worst the first in the sortorder has to be put back at the fron, but this should be a onetime effort.

weizhouapache: @DaanHoogland assume the sort_key is 1,0, with your sql sort_key will be changed to 1,2.
the order of zones is changed by the sql.

DaanHoogland: so in the case above, `sort_key` is only changed if there is no `sort_key != 0`. This would mean that if there are three where it is 0 and one where it is 10, nothing gets changed, does it?

if this is at all possible btw!

weizhouapache: @DaanHoogland exactly.
IMHO, if there is a non-zero sort_key , which mean cloudstack user might use sort_key for ordering, we will not touch it.
if all sort_key are 0, we change sort_key to id.

DaanHoogland: ok, sounds reasonable.

DaanHoogland (approved):

rhtyd:
@blueorangutan test

davidjumani (approved):

rhtyd (approved):

PR #4507, opened 2020-11-30
Fix failure in validating IP address in case of multiple Management Servers

Pearl1594:
### Description
When there are multiple Management servers, their IPs are listed as comma separated string in /var/cache/cloud/cmdline file against the "host" parameter, and when setting up the SSVM,  SSVM fails to validate the IP of the management servers:
```
WARN  [storage.resource.NfsSecondaryStorageResource] (main:null) destIp is not a valid ip address or cidr destIp=x.x.x.x,y.y.y.y
```
In order to fix this, the IP addresses are split on the delimiter - ',' comma and then the further operations are performed.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
Deployed a setup with 2 management servers, and validated the fix. No longer fails validation of IP address

### DIFF
--- a/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
+++ b/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
@@ -46,7 +46,6 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -2588,8 +2587,10 @@ public class NfsSecondaryStorageResource extends ServerResourceBase implements S
         if (_inSystemVM) {
             _localgw = (String)params.get("localgw");
             if (_localgw != null) { // can only happen inside service vm
-                String mgmtHost = (String)params.get("host");
-                addRouteToInternalIpOrCidr(_localgw, _eth1ip, _eth1mask, mgmtHost);
+                String mgmtHosts = (String)params.get("host");
+                for (final String mgmtHost : mgmtHosts.split(",")) {
+                    addRouteToInternalIpOrCidr(_localgw, _eth1ip, _eth1mask, mgmtHost);
+                }
 
                 String internalDns1 = (String)params.get("internaldns1");
                 if (internalDns1 == null) {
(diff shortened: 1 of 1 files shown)


Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan test

DaanHoogland:
Though I see the need of this fix, i am a bit wary of it. It must be implemented somewhere for the kvm agent as well, so why is that code not called here. It seems like we are doubling/overlooking something.

weizhouapache:
@Pearl1594 why not change line 2592  to below ?

```
                String mgmtHosts = (String)params.get("host");
                for (final String mgmtHost : mgmtHosts.split(",")) {
                    addRouteToInternalIpOrCidr(_localgw, _eth1ip, _eth1mask, mgmtHost);
                }
```

weizhouapache (approved):
code LGTM

sureshanaparti (approved):
code changes LGTM, same behavior exists in _ConsoleProxyResource_.

Pearl1594:
> Though I see the need of this fix, i am a bit wary of it. It must be implemented somewhere for the kvm agent as well, so why is that code not called here. It seems like we are doubling/overlooking something.

This situation does not arise in case of KVM agent @DaanHoogland

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan test

DaanHoogland:
> > Though I see the need of this fix, i am a bit wary of it. It must be implemented somewhere for the kvm agent as well, so why is that code not called here. It seems like we are doubling/overlooking something.
> 
> This situation does not arise in case of KVM agent @DaanHoogland

I think I wasn't clear , @Pearl1594 . The kvm agent must have code to handle the case of multiple ip addresses for the MS, why isn't the same code used for both? The ip selection process during connecting is the same for all agent so the code should be the same bit as well.

Pearl1594:
@DaanHoogland wrt KVM agent, the workflow followed is a bit different:
https://github.com/apache/cloudstack/blob/master/agent/src/main/java/com/cloud/agent/Agent.java#L194
https://github.com/apache/cloudstack/blob/master/agent/src/main/java/com/cloud/agent/AgentShell.java#L152
And here, we correctly consider the possibility of having multiple MS and split them accordingly.

This piece of code aligns to how ConsoleProxyResource handles multiple MSs.

DaanHoogland:
ok, I'll shut my load mouth, @Pearl1594 
@PaulAngus this is minor but meet merge criteria. At your discretion.

shwstppr (approved):
LGTM

rhtyd (approved):

PR #4516, opened 2020-12-03
Fix hypervisor type cast to string

alexandru-bagu:
### Description

This PR addresses an error that appears when you try to add a new host. I don't even understand why there was a cast to String in the first place. I will assume some classes send HypervisorType and some send a string (empty or otherwise). Shouldn't this be addressed to use the same type everywhere? With this fix adding a new xenserver host works fine.

Here is the stack trace:

> WARN  [c.c.s.StorageManagerImpl] (AgentTaskPool-6:ctx-90ed1f24) (logid:77e97d5e) Unable to setup the local storage pool for Host[-4-Routing]
java.lang.ClassCastException: class com.cloud.hypervisor.Hypervisor$HypervisorType cannot be cast to class java.lang.String (com.cloud.hypervisor.Hypervisor$HypervisorType is in unnamed module of loader org.eclipse.jetty.webapp.WebAppClassLoader @3f69b166; java.lang.String is 
in module java.base of loader 'bootstrap')
        at org.apache.cloudstack.storage.datastore.lifecycle.CloudStackPrimaryDataStoreLifeCycleImpl.initialize(CloudStackPrimaryDataStoreLifeCycleImpl.java:135)
        at com.cloud.storage.StorageManagerImpl.createLocalStorage(StorageManagerImpl.java:627)
        ...
INFO  [c.c.u.e.CSExceptionErrorCode] (AgentTaskPool-6:ctx-90ed1f24) (logid:77e97d5e) Could not find exception: com.cloud.exception.ConnectionException in error code list for exceptions
WARN  [c.c.a.m.AgentManagerImpl] (AgentTaskPool-6:ctx-90ed1f24) (logid:77e97d5e) Monitor LocalStoragePoolListener says there is an error in the connect process for 4 due to Unable to setup the local storage pool for Host[-4-Routing]
INFO  [c.c.a.m.AgentManagerImpl] (AgentTaskPool-6:ctx-90ed1f24) (logid:77e97d5e) Host 4 is disconnecting with event AgentDisconnected
WARN  [c.c.r.ResourceManagerImpl] (AgentTaskPool-6:ctx-90ed1f24) (logid:77e97d5e) Unable to connect due to
com.cloud.exception.ConnectionException: Unable to setup the local storage pool for Host[-4-Routing]
        at com.cloud.storage.StorageManagerImpl.createLocalStorage(StorageManagerImpl.java:640)
        ...
Caused by: java.lang.ClassCastException: class com.cloud.hypervisor.Hypervisor$HypervisorType cannot be cast to class java.lang.String (com.cloud.hypervisor.Hypervisor$HypervisorType is in unnamed module of loader org.eclipse.jetty.webapp.WebAppClassLoader @3f69b166; java.lang.String is in module java.base of loader 'bootstrap')
        at org.apache.cloudstack.storage.datastore.lifecycle.CloudStackPrimaryDataStoreLifeCycleImpl.initialize(CloudStackPrimaryDataStoreLifeCycleImpl.java:135)
        at com.cloud.storage.StorageManagerImpl.createLocalStorage(StorageManagerImpl.java:627)
        ... 35 more

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
Create a basic cloudstack development build (more or less following the installation guide, no oss), try to add a new host with **local storage** (I assume any virtualization type, but the code itself is fishy. I imagine it worked fine for whoever added this bit when he tested with VMware otherwise he would have not sent the patch) and you get this error. I tested with xenserver.

### DIFF
--- a/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
+++ b/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
@@ -249,7 +249,7 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
             parameters.setPath(hostPath.replaceFirst("/", ""));
             parameters.setUserInfo(userInfo);
         } else if (scheme.equalsIgnoreCase("PreSetup")) {
-            if (hypervisorType.equals(HypervisorType.VMware)) {
+            if (HypervisorType.VMware.equals(hypervisorType)) {
                 validateVcenterDetails(zoneId, podId, clusterId,storageHost);
             }
             parameters.setType(StoragePoolType.PreSetup);
--- a/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
+++ b/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
@@ -337,7 +337,7 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
             uuid = (String)existingUuid;
         } else if (scheme.equalsIgnoreCase("sharedmountpoint") || scheme.equalsIgnoreCase("clvm")) {
             uuid = UUID.randomUUID().toString();
-        } else if (scheme.equalsIgnoreCase("PreSetup") && !hypervisorType.equals(HypervisorType.VMware)) {
+        } else if ("PreSetup".equalsIgnoreCase(scheme) && !HypervisorType.VMware.equals(hypervisorType)) {
             uuid = hostPath.replace("/", "");
         } else {
             uuid = UUID.nameUUIDFromBytes((storageHost + hostPath).getBytes()).toString();
--- a/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
+++ b/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
@@ -257,7 +257,7 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
             parameters.setPort(0);
             parameters.setPath(hostPath);
         } else if (scheme.equalsIgnoreCase("DatastoreCluster")) {
-            if (hypervisorType.equals(HypervisorType.VMware)) {
+            if (HypervisorType.VMware.equals(hypervisorType)) {
                 validateVcenterDetails(zoneId, podId, clusterId,storageHost);
             }
             parameters.setType(StoragePoolType.DatastoreCluster);
--- a/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
+++ b/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
@@ -64,7 +64,6 @@ import org.apache.cloudstack.engine.subsystem.api.storage.ZoneScope;
 import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
 import org.apache.cloudstack.storage.volume.datastore.PrimaryDataStoreHelper;
-import org.apache.commons.lang.StringUtils;
 import org.apache.log4j.Logger;
 
 import javax.inject.Inject;
@@ -132,7 +131,7 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
         Long zoneId = (Long)dsInfos.get("zoneId");
         String url = (String)dsInfos.get("url");
         String providerName = (String)dsInfos.get("providerName");
-        String hypervisorType = (String)dsInfos.get("hypervisorType");
+        HypervisorType hypervisorType = (HypervisorType)dsInfos.get("hypervisorType");
         if (clusterId != null && podId == null) {
             throw new InvalidParameterValueException("Cluster id requires pod id");
         }
@@ -250,7 +249,7 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
             parameters.setPath(hostPath.replaceFirst("/", ""));
             parameters.setUserInfo(userInfo);
         } else if (scheme.equalsIgnoreCase("PreSetup")) {
-            if (StringUtils.isNotBlank(hypervisorType) && HypervisorType.getType(hypervisorType).equals(HypervisorType.VMware)) {
+            if (hypervisorType.equals(HypervisorType.VMware)) {
                 validateVcenterDetails(zoneId, podId, clusterId,storageHost);
             }
             parameters.setType(StoragePoolType.PreSetup);
@@ -258,7 +257,7 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
             parameters.setPort(0);
             parameters.setPath(hostPath);
         } else if (scheme.equalsIgnoreCase("DatastoreCluster")) {
-            if (StringUtils.isNotBlank(hypervisorType) && HypervisorType.getType(hypervisorType).equals(HypervisorType.VMware)) {
+            if (hypervisorType.equals(HypervisorType.VMware)) {
                 validateVcenterDetails(zoneId, podId, clusterId,storageHost);
             }
             parameters.setType(StoragePoolType.DatastoreCluster);
@@ -338,7 +337,7 @@ public class CloudStackPrimaryDataStoreLifeCycleImpl implements PrimaryDataStore
             uuid = (String)existingUuid;
         } else if (scheme.equalsIgnoreCase("sharedmountpoint") || scheme.equalsIgnoreCase("clvm")) {
             uuid = UUID.randomUUID().toString();
-        } else if (scheme.equalsIgnoreCase("PreSetup") && !(StringUtils.isNotBlank(hypervisorType) && HypervisorType.getType(hypervisorType).equals(HypervisorType.VMware))) {
+        } else if (scheme.equalsIgnoreCase("PreSetup") && !hypervisorType.equals(HypervisorType.VMware)) {
             uuid = hostPath.replace("/", "");
         } else {
             uuid = UUID.nameUUIDFromBytes((storageHost + hostPath).getBytes()).toString();
--- a/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/VolumeServiceTest.java
+++ b/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/VolumeServiceTest.java
@@ -276,6 +276,7 @@ public class VolumeServiceTest extends CloudStackTestNGBase {
         params.put("path", uri.getPath());
         params.put("protocol", StoragePoolType.NetworkFilesystem);
         params.put("dcId", dcId.toString());
+        params.put("hypervisorType", HypervisorType.None);
         params.put("clusterId", clusterId.toString());
         params.put("name", this.primaryName);
         params.put("port", "1");
@@ -318,6 +319,7 @@ public class VolumeServiceTest extends CloudStackTestNGBase {
             params.put("path", uri.getPath());
             params.put("protocol", Storage.StoragePoolType.NetworkFilesystem);
             params.put("dcId", dcId.toString());
+            params.put("hypervisorType", HypervisorType.None);
             params.put("clusterId", clusterId.toString());
             params.put("name", this.primaryName);
             params.put("port", "1");
--- a/server/src/main/java/com/cloud/storage/StorageManagerImpl.java
+++ b/server/src/main/java/com/cloud/storage/StorageManagerImpl.java
@@ -722,7 +722,7 @@ public class StorageManagerImpl extends ManagerBase implements StorageManager, C
         params.put("zoneId", zone.getId());
         params.put("clusterId", clusterId);
         params.put("podId", podId);
-        params.put("hypervisorType", hypervisorType.toString());
+        params.put("hypervisorType", hypervisorType);
         params.put("url", cmd.getUrl());
         params.put("tags", cmd.getTags());
         params.put("name", cmd.getStoragePoolName());


sureshanaparti:
@alexandru-bagu why _host.getHypervisorType()_  is null in your case, here: https://github.com/apache/cloudstack/blob/fb1e903532d943bc98a1a8b95d4499013234a917/server/src/main/java/com/cloud/storage/StorageManagerImpl.java#L619

alexandru-bagu:
> @alexandru-bagu why _host.getHypervisorType()_ is null in your case, here:
> 
> https://github.com/apache/cloudstack/blob/fb1e903532d943bc98a1a8b95d4499013234a917/server/src/main/java/com/cloud/storage/StorageManagerImpl.java#L619

I am sure it's not null, not in my case anyway. If you are actually referring to the null verification I do in PR then I guess it's just a habit I picked up along the way. As you can read in my stacktrace, the hypervisorType is not null but rather not castable to string. I am assuming it was never accepted in java to cast an enum to string. Enums (at least in C#) are castable to integers not strings. If my assumption is wrong then is this code not meant to be ran with openjdk?

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -132,7 +132,11 @@ public DataStore initialize(Map<String, Object> dsInfos) {
         Long zoneId = (Long)dsInfos.get("zoneId");
         String url = (String)dsInfos.get("url");
         String providerName = (String)dsInfos.get("providerName");
-        String hypervisorType = (String)dsInfos.get("hypervisorType");
+        Object hypervisorTypeObj = dsInfos.get("hypervisorType");
```
weizhouapache, line 135: it would be better to define hypervisorType as
```
HypervisorType hypervisorType = (String)dsInfos.get("hypervisorType");
```
the other lines in this file needs to be changed accordingly.

alexandru-bagu: I will update the PR.

sureshanaparti:
> I updated the PR. I still have to check if the dsInfos contains the hypervisorType because some tests do not provide a hypervisorType.

what tests, which hypervisor ?

alexandru-bagu:
Are https://github.com/apache/cloudstack/blob/e5c29c392e5fa00653c7b7a1ea8d3fbd6b8b974f/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/VolumeServiceTest.java#L263 and https://github.com/apache/cloudstack/blob/e5c29c392e5fa00653c7b7a1ea8d3fbd6b8b974f/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/VolumeServiceTest.java#L311 not using the API in question?

Also looking through the code, this does not look ok:
using HypervisorType: https://github.com/apache/cloudstack/blob/e5c29c392e5fa00653c7b7a1ea8d3fbd6b8b974f/server/src/main/java/com/cloud/storage/StorageManagerImpl.java#L619
using String: https://github.com/apache/cloudstack/blob/e5c29c392e5fa00653c7b7a1ea8d3fbd6b8b974f/server/src/main/java/com/cloud/storage/StorageManagerImpl.java#L725

This is at first glance. Since I am new to CloudStack architecture I cannot say for certain that both those calls are to the same _initialize_ function but it looks like that.

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -133,7 +133,7 @@ public DataStore initialize(Map<String, Object> dsInfos) {
         String providerName = (String)dsInfos.get("providerName");
         HypervisorType hypervisorType = HypervisorType.None;
         if(dsInfos.containsKey("hypervisorType")) {
-            hypervisorType = (HypervisorType)dsInfos.get("hypervisorType");
+            hypervisorType = HypervisorType.getType((String)dsInfos.get("hypervisorType"));
```
weizhouapache, line 136: @alexandru-bagu in your latest commit, hypervisor type is casted to string again

alexandru-bagu: @weizhouapache The issue I created this PR for is that dsInfos.get("hypervisorType") is sometimes String or sometimes missing. As I highlighted in my previous post, both String and HypervisorType is used so I am not sure about how to proceed. We could check types but I don't believe that is correct.

DaanHoogland:
> > @alexandru-bagu why _host.getHypervisorType()_ is null in your case, here:
> > https://github.com/apache/cloudstack/blob/fb1e903532d943bc98a1a8b95d4499013234a917/server/src/main/java/com/cloud/storage/StorageManagerImpl.java#L619
> 
> I am sure it's not null, not in my case anyway. If you are actually referring to the null verification I do in PR then I guess it's just a habit I picked up along the way. As you can read in my stacktrace, the hypervisorType is not null but rather not castable to string. I am assuming it was never accepted in java to cast an enum to string. Enums (at least in C#) are castable to integers not strings. If my assumption is wrong then is this code not meant to be ran with openjdk?

@alexandru-bagu, it is a general accepted practice to have enum types be casted to string, but you'll need a `toString()` defined on the enum. casting to int is generally frowned upon because the string descriptors may be auto-numbered and an insert on the type will change the numbering. (in this case the enum range is small and also persisted so that can't be allowed anyway) hope this explains.

alexandru-bagu:
> > > @alexandru-bagu why _host.getHypervisorType()_ is null in your case, here:
> > > https://github.com/apache/cloudstack/blob/fb1e903532d943bc98a1a8b95d4499013234a917/server/src/main/java/com/cloud/storage/StorageManagerImpl.java#L619
> > 
> > 
> > I am sure it's not null, not in my case anyway. If you are actually referring to the null verification I do in PR then I guess it's just a habit I picked up along the way. As you can read in my stacktrace, the hypervisorType is not null but rather not castable to string. I am assuming it was never accepted in java to cast an enum to string. Enums (at least in C#) are castable to integers not strings. If my assumption is wrong then is this code not meant to be ran with openjdk?
> 
> @alexandru-bagu, it is a general accepted practice to have enum types be casted to string, but you'll need a `toString()` defined on the enum. casting to int is generally frowned upon because the string descriptors may be auto-numbered and an insert on the type will change the numbering. (in this case the enum range is small and also persisted so that can't be allowed anyway) hope this explains.

My comment was not about whether it's acceptable or not to cast enums to integers. It was merely about the fact that C# and Java compilers allow by default enum to int casts and vice versa but not enum to string or string to enum directly. That comment was obviously related to the fact that in the current source there is a cast of an Object to String but it turns out that Object is sometimes either HypervisorType or String. I opened this PR because casting an enum to string (not using helpers/built in functions of Enums) causes an exception.

The issue was introduced by https://github.com/apache/cloudstack/commit/6df819028e472dd9801873857fe20133381ca0fe and I guess no one had this issue because no tests were done using local storage.

And as a side note, it seems is is "acceptable"/"ok" to have known blocking errors in the upcoming release (4.15.0.0). I mean why not postpone the fix until 4.15.1.0? In the meantime anyone who wants to use 4.15.0.0 with local storage will just have to patch and build the software themselves I guess. I know I will have to.

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -133,7 +133,7 @@ public DataStore initialize(Map<String, Object> dsInfos) {
         String providerName = (String)dsInfos.get("providerName");
         HypervisorType hypervisorType = HypervisorType.None;
         if(dsInfos.containsKey("hypervisorType")) {
-            hypervisorType = HypervisorType.getType((String)dsInfos.get("hypervisorType"));
+            hypervisorType = HypervisorType.getType(dsInfos.get("hypervisorType").toString());
```
DaanHoogland, line 136: casting to String and calling toString() on an object would only be different if toString() is not implemented @alexandru-bagu . What are you aiming for here?

alexandru-bagu: Unless I am wrong, .toString is implemented by the compiler for enums. My aim was to provide a quick fix for this issue and I did a small test to show that the code above works. https://github.com/alexandru-bagu/java-enum-test

The point of .toString there is that if the object given is either String or HypervisorType then both will provide the same output for .toString unless as you have said one overrides the .toString for the enum. In this case it is not overriden as such the code will work.

Though I agree, this not the correct way to fix it. I will make it a cast to HypervisorType and change the function call that provides the wrong type.

alexandru-bagu:
I made the required changes to address the tests and wrong usage for StorageMangerImpl. Did a clean setup with xen with local storage and it worked fine.

DaanHoogland:
@alexandru-bagu if this blocks you it is a blocker.

DaanHoogland:
@weizhouapache @sureshanaparti are your concerns met?
@harikrishna-patnala can you please review asap?

DaanHoogland:
cc @rhtyd

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

rhtyd:
@harikrishna-patnala pl review.
LGTM, it's a blocker and an RC2 would be needed if local storage is broken. Can we also get some tests KVM + local storage, cc @GabrielBrascher @wido @weizhouapache ? Thanks.

DaanHoogland:
kubernetes errors are not related and also very intermittent. I'm not prepared to merge yet, for lack of user feedback. @alexandru-bagu we ask for 2 lgtm and an independent/3rd party test ok to merge as a general policy. In this case I will wait a while and merge if no feedback comes to not block 4.15 for too long. I'll stop the vote now with a link here.

GabrielBrascher:
@DaanHoogland @rhtyd I will run some tests with KVM + local storage this weekend.

weizhouapache (approved):
code lgtm

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -250,15 +249,15 @@ public DataStore initialize(Map<String, Object> dsInfos) {
             parameters.setPath(hostPath.replaceFirst("/", ""));
             parameters.setUserInfo(userInfo);
         } else if (scheme.equalsIgnoreCase("PreSetup")) {
-            if (StringUtils.isNotBlank(hypervisorType) && HypervisorType.getType(hypervisorType).equals(HypervisorType.VMware)) {
+            if (hypervisorType.equals(HypervisorType.VMware)) {
```
weizhouapache, line 252: I do not know if hypervisorType is 100% not null.
maybe it is better to use `if (HypervisorType.VMware.equals(hypervisorType)) {` ?

alexandru-bagu: Wouldn't that hide improper calls to this function?

rhtyd:
@alexandru-bagu can you add the null/blank defensive checks?

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
DaanHoogland: no, @alexandru-bagu , it is common practice using equals to call it on the object guaranteed not to be null, in this case the enum member
```suggestion
            if (HypervisorType.VMware.equals(hypervisorType)) {
```

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -250,15 +249,15 @@ public DataStore initialize(Map<String, Object> dsInfos) {
...
                 validateVcenterDetails(zoneId, podId, clusterId,storageHost);
             }
             parameters.setType(StoragePoolType.PreSetup);
             parameters.setHost(storageHost);
             parameters.setPort(0);
             parameters.setPath(hostPath);
         } else if (scheme.equalsIgnoreCase("DatastoreCluster")) {
-            if (StringUtils.isNotBlank(hypervisorType) && HypervisorType.getType(hypervisorType).equals(HypervisorType.VMware)) {
+            if (hypervisorType.equals(HypervisorType.VMware)) {
```
DaanHoogland, line 260:
```suggestion
            if (HypervisorType.VMware.equals(hypervisorType)) {
```

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -338,7 +337,7 @@ public DataStore initialize(Map<String, Object> dsInfos) {
             uuid = (String)existingUuid;
         } else if (scheme.equalsIgnoreCase("sharedmountpoint") || scheme.equalsIgnoreCase("clvm")) {
             uuid = UUID.randomUUID().toString();
-        } else if (scheme.equalsIgnoreCase("PreSetup") && !(StringUtils.isNotBlank(hypervisorType) && HypervisorType.getType(hypervisorType).equals(HypervisorType.VMware))) {
+        } else if (scheme.equalsIgnoreCase("PreSetup") && !hypervisorType.equals(HypervisorType.VMware)) {
```
DaanHoogland, line 340:
```suggestion
        } else if ("PreSetup".equalsIgnoreCase(scheme) && !HypervisorType.VMware.equals(hypervisorType)) {
```

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
alexandru-bagu: You misunderstand. If someone calls this function without providing this parameter then they wouldn't know because there would be no error reported. Alas I will commit your suggestions.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

GabrielBrascher (approved):
THanks for the PR @alexandru-bagu, Code LGTM.

I have not been able to test it (no VMware environment available on short time).

@DaanHoogland I have tested with KVM and things are looking good on both (I) upgrade environments with local storage already deployed or a zone deployed from scratch.

rhtyd (approved):

rhtyd:
Thanks @GabrielBrascher I'll merge based on your LGTM

PR #3945, opened 2020-03-05
server: update template to another template type

ustcweizhou:
## Description

this contains other changes
(1) add isrouting field for vm templates on UI
(2) show register URL of template/iso on UI
(3) set 'Bootable' field to changable for existing ISO

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

this provides a easy way to upload systemvm template
(1) register template
(2) update template type from USER to SYSTEM
(3) destroy ssvm/cpvm, new systemvms will be created from the new template (latest template with type=SYSTEM)
(4) change global setting router.template.kvm (or other hypervisor) to the name of new template, new VRs will be created from the new template.

PR #4522, opened 2020-12-08
template: Ensuring template is cross zone if type changed to system

davidjumani:
### Description

This PR enhances https://github.com/apache/cloudstack/pull/3945 by ensuring that all to-be System templates are cross zone

### Types of changes

### DIFF
--- a/server/src/main/java/com/cloud/template/TemplateManagerImpl.java
+++ b/server/src/main/java/com/cloud/template/TemplateManagerImpl.java
@@ -2108,6 +2108,9 @@ public class TemplateManagerImpl extends ManagerBase implements TemplateManager,
             if (templateType != null && cmd.isRoutingType() != null && (TemplateType.ROUTING.equals(templateType) != cmd.isRoutingType())) {
                 throw new InvalidParameterValueException("Please specify a valid templatetype (consistent with isrouting parameter).");
             }
+            if (templateType != null && (templateType == TemplateType.SYSTEM || templateType == TemplateType.BUILTIN) && !template.isCrossZones()) {
+                throw new InvalidParameterValueException("System and Builtin templates must be cross zone");
+            }
         }
 
         // update is needed if any of the fields below got filled by the user


server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2198,6 +2202,21 @@ private VMTemplateVO updateTemplateOrIso(BaseUpdateTemplateOrIsoCmd cmd) {
...
+                List<Long> zoneIds = zones.stream().map(zone -> zone.getId()).collect(Collectors.toList());
+                try {
+                    copyTemplate(id, null, zoneIds);
+                } catch (StorageUnavailableException e) {
+                    throw new CloudRuntimeException("Unable to copy template to other zones : ", e);
+                } catch (ResourceAllocationException e) {
+                    throw new CloudRuntimeException("Unable to copy template to other zones : ", e);
+                }
+            }
+        }
```
DaanHoogland, line 2218: is a good check but i'd like to see it in a separate method.

rhtyd (approved):

rhtyd:
cc @ravening @weizhouapache pl review
@blueorangutan package

ravening:
> cc @ravening @weizhouapache pl review
> @blueorangutan package

@rhtyd will review this when its ready

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2224,6 +2232,21 @@ else if (details != null && !details.isEmpty()) {
         return _tmpltDao.findById(id);
     }
 
+    private void copyTemplateToAllZones(Long id) {
+        s_logger.info("Copying template to all zones since it will be a system template");
+        List<DataCenterVO> zones = _dcDao.listEnabledZones();
+        if (zones != null) {
+            List<Long> zoneIds = zones.stream().map(zone -> zone.getId()).collect(Collectors.toList());
```
ravening, line 2239: Should we skip the zone on which the template is already in?

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2224,6 +2232,21 @@ else if (details != null && !details.isEmpty()) {
         return _tmpltDao.findById(id);
     }
 
+    private void copyTemplateToAllZones(Long id) {
+        s_logger.info("Copying template to all zones since it will be a system template");
```
ravening, line 2236: Better to display the template name?

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2224,6 +2232,21 @@ else if (details != null && !details.isEmpty()) {
...
+        List<DataCenterVO> zones = _dcDao.listEnabledZones();
+        if (zones != null) {
+            List<Long> zoneIds = zones.stream().map(zone -> zone.getId()).collect(Collectors.toList());
+            try {
+                copyTemplate(id, null, zoneIds);
+            } catch (StorageUnavailableException e) {
+                throw new CloudRuntimeException("Unable to copy template to other zones : ", e);
```
ravening, line 2243: Extract the message to string?

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
weizhouapache: this seems to be handled by https://github.com/apache/cloudstack/blob/c59d0c82137f2294c67d4dd77dd1f4a69b2edb09/server/src/main/java/com/cloud/template/TemplateManagerImpl.java#L931-L936

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2198,6 +2202,10 @@ private VMTemplateVO updateTemplateOrIso(BaseUpdateTemplateOrIsoCmd cmd) {
             template.setDynamicallyScalable(isDynamicallyScalable);
         }
 
+        if (templateType != null && templateType == TemplateType.SYSTEM) {
+            copyTemplateToAllZones(id);
+        }
```
weizhouapache, line 2207: @davidjumani should the template be changed to cross-zone when it is done ?

davidjumani:
@weizhouapache @ravening @DaanHoogland Would it be better to make the UpdateTemplate API an async one now that it copies templates between zones ?

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
davidjumani: Makes sense. Added the change

davidjumani:
Updated the PR to just add a cross zone check when trying to change a template to a System / Builtin template

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

weizhouapache:
@davidjumani now it is more clear, good.

weizhouapache (approved):

weizhouapache:
@davidjumani by the way, if there is only one zone, should it also be ok if it is not cross-zone ?

davidjumani:
@weizhouapache If all or no zones are passed when the template is uploaded, it should work fine even in a single zone scenario

weizhouapache:
> @weizhouapache If 'All zones' or no zoneids are passed when the template is uploaded, it should work fine even in a single zone scenario

@davidjumani yes, the template will be cross-zone. if there is only 1 zone, and upload template for the zone, should it work if change template type to 'SYSTEM' ?

I have another question you may know the answer . Assume there are two zones (zone1/zone2) run with cloudstack 4.14, if register new systemvm template for 4.15 (systemvm-XX-4.15.0) for zone1, will the upgrade to 4.15 work ? will the new systemvm template be copied to zone2 ?

davidjumani:
> > @weizhouapache If 'All zones' or no zoneids are passed when the template is uploaded, it should work fine even in a single zone scenario
> 
> @davidjumani yes, the template will be cross-zone. if there is only 1 zone, and upload template for the zone, should it work if change template type to 'SYSTEM' ?
> 
> I have another question you may know the answer . Assume there are two zones (zone1/zone2) run with cloudstack 4.14, if register new systemvm template for 4.15 (systemvm-XX-4.15.0) for zone1, will the upgrade to 4.15 work ? will the new systemvm template be copied to zone2 ?

@weizhouapache In the first case, I'm not sure it would work, since a zone was selected during the upload. Need to look at the code for a single zone scenario
The second, it will not upgrade since systemvm templates need to be uploaded in all zones that contain the respective hypervisor
@rhtyd Could you confirm ?

DaanHoogland:
I think this addresses the main point that @PaulAngus had with the change template type functionality. He hasn't given a full list indicating there is more, so can we think of more @weizhouapache @davidjumani @rhtyd @ravening ?
This fix looks good to me and should go in before 4.15.

PaulAngus:
> I think this addresses the main point that @PaulAngus had with the change template type functionality. He hasn't given a full list indicating there is more, so can we think of more @weizhouapache @davidjumani @rhtyd @ravening ?
> This fix looks good to me and should go in before 4.15.

PaulAngus:
sorry pressed wrong button. didn't mean to close

DaanHoogland:
:) np, but you did mean to comment I gather

PaulAngus:
yes, the comment was the 'only' intent.  still not blocking, just highlighting/correcting the issues as i see them.

DaanHoogland:
> > I think this addresses the main point that @PaulAngus had with the change template type functionality. He hasn't given a full list indicating there is more, so can we think of more @weizhouapache @davidjumani @rhtyd @ravening ?
> > This fix looks good to me and should go in before 4.15.
> 
> wow, I'm cloudstack QA now?
> I see 'should', 'assume', 'according to code', with no testing.
> There are about 5 different states, which means something like 20 permutations of state change ( as it can't change to itself).

exactly my point, so we need to assess whether for any other of the permutations extra safeguards are needed.

rhtyd:
@davidjumani can you kick pkg/test again?

davidjumani:
@blueorangutan package

DaanHoogland (approved):

davidjumani:
@blueorangutan test

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2130,7 +2129,7 @@ private VMTemplateVO updateTemplateOrIso(BaseUpdateTemplateOrIsoCmd cmd) {
             return template;
         }
 
-        template = _tmpltDao.createForUpdate(id);
+        template = _tmpltDao.findById(id);
```
rhtyd, line 2132: why is this changed?

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2219,7 +2222,7 @@ else if (details != null && !details.isEmpty()) {
             _tmpltDao.saveDetails(template);
         }
 
-        _tmpltDao.update(id, template);
+        _tmpltDao.persist(template);
```
rhtyd, line 2225: Here too?

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
davidjumani: createForUpdate creates an empty object with just the id https://github.com/apache/cloudstack/blob/master/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java#L295 , whereas a template object read from the db needed to check whether it's cross zone

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
davidjumani: Required since it's changed from createForUpdate to findById

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
rhtyd: Line 2075 already has that, we need to check why createForUpdate was used in the first place.

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
rhtyd: Persist is usually used for a new record, I'm concerned if this causes any regression or duplicates in DB.

rhtyd:
@davidjumani can you revert the two lines I've comments on, move the conditional up if that makes sense? Let's keep the changes minimal if possible, easier to review and test. Thanks.

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
davidjumani: There are several instances where findById has been used prior to persist. Tested and no duplicate DB entries were created

rhtyd: I was considering getting this towards 4.15RC2 but the additional changes don't give us confidence - only the conditional check was enough to handle the issue.

davidjumani: Makes sense, moving the check up

davidjumani:
@blueorangutan package

rhtyd (approved):

rhtyd:
@weizhouapache @DaanHoogland should we consider this for RC2, with current changes it addresses concerns from #3945
@blueorangutan test

weizhouapache:
> @weizhouapache @DaanHoogland should we consider this for RC2, with current changes it addresses concerns from #3945
> @blueorangutan test

@rhtyd I am ok to merge this into 4.15.0.0

PR #4531, opened 2020-12-10
novnc: Accept new novnc client and disconnect old session

### LINKED ISSUE #4550: Potential security issue with novnc console
The url in iframe of novnc console can be used without user login.

When open a vm console, and open the source of page, we get the real link of novnc console.
we are able to access the novnc console on other browsers/servers/ via the link, without user login.

This is a potential security issue.

We have two options
(1) add source IP of client to encrypted token, and check the client IP after token decryption when access the vm console.

we need to make sure
- IP to management server and CPVM (console proxy vm) are same.
- If the new browser/session has same outgoing IP as old session, it will still be able to access the vm console.

to be clear, these two files need to be changed
https://github.com/apache/cloudstack/blob/master/server/src/main/java/com/cloud/servlet/ConsoleProxyClientParam.java
and
https://github.com/apache/cloudstack/blob/master/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyClientParam.java

(2) save the access ticket to database
when vm console, access ticket will be generated and saved into database.

When access the console, management server will check the params to see if vm is valid and access ticket is not expired.,etc.
It will also check if token is valid and has been not used.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
NoVNC console
~~~

##### CLOUDSTACK VERSION

~~~
4.15/master
~~~

##### STEPS TO REPRODUCE

1. open a vm console
example:  
image

2. view the source of the page
example
```
<html><title>v-1-VM</title><frameset><frame src="https://10-135-126-207.cloud.leaseweb.net/resource/noVNC/vnc.html?port=8080&token=I-QW-IpdDhf3i82rJP_WctztYaq15O6ZP1Mya9F1VspNwDHDhXhQjRqO-ubZMM6YwJkfDNNEShT1K1XZmrnZvmv9ro504qoPqEuC9go5UlrQqBdJga-DAWaFKg0pv_QGo3xhGy4VSNCuj_N86ILC8wAQjmaqbSxyqFFfAlEc8UHxuyGiZG4YpHNnM6je3V82IV555AfnA1-3oyXh5tp9X-Bd7zeXRGIprt7_lwTnGxLgvdk06AoGseD4YabpB_aJLpvx3RfxuHD7hR1IrOrl9g"></frame></frameset></html>

```
3. open the link on other server. 
It will not work at the first time, but works after refresh (related to #4531 )

image

##### EXPECTED RESULTS

~~~
The link in page source should NOT work in other servers, without login
~~~

##### ACTUAL RESULTS

~~~
We are able to access the vm console via the URL in page source, without login.
~~~

ustcweizhou:
### Description

Fixes #4550 
This PR fixes an issue with novnc.

1. open a vm console
2. open the console in another session (other server, or different browser, or incognito window)
console will be stuck at Connecting
3. reload the page, vm console works. The console in step (1) is kicked out.

image

with this PR, the console in new session will work, the old session will be disconnected.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

with this pr, in new windows

image

### DIFF
--- a/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxy.java
+++ b/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxy.java
@@ -552,9 +552,13 @@ public class ConsoleProxy {
                     throw new AuthenticationException("Cannot use the existing viewer " + viewer + ": bad sid");
 
                 if (!viewer.isFrontEndAlive()) {
+                    s_logger.info("Initializing new novnc client as novnc view object is not alive: " + viewer);
                     authenticationExternally(param);
                     viewer.initClient(param);
                     reportLoadChange = true;
+                } else {
+                    s_logger.info("Rejecting new novnc client as novnc viewer object is still alive: " + viewer);
+                    return null;
                 }
             }
 
--- a/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxy.java
+++ b/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxy.java
@@ -551,15 +551,23 @@ public class ConsoleProxy {
                         !param.getClientHostPassword().equals(viewer.getClientHostPassword()))
                     throw new AuthenticationException("Cannot use the existing viewer " + viewer + ": bad sid");
 
-                if (!viewer.isFrontEndAlive()) {
-                    s_logger.info("Initializing new novnc client as novnc view object is not alive: " + viewer);
+                try {
                     authenticationExternally(param);
-                    viewer.initClient(param);
-                    reportLoadChange = true;
-                } else {
-                    s_logger.info("Rejecting new novnc client as novnc viewer object is still alive: " + viewer);
+                } catch (Exception e) {
+                    s_logger.error("Authencation failed for param: " + param);
                     return null;
                 }
+                s_logger.info("Initializing new novnc client and disconnecting existing session");
+                try {
+                    ((ConsoleProxyNoVncClient)viewer).getSession().disconnect();
+                } catch (IOException e) {
+                    s_logger.error("Exception while disconnect session of novnc viewer object: " + viewer, e);
+                }
+                removeViewer(viewer);
+                viewer = new ConsoleProxyNoVncClient(session);
+                viewer.initClient(param);
+                connectionMap.put(clientKey, viewer);
+                reportLoadChange = true;
             }
 
             if (reportLoadChange) {
--- a/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyNoVNCHandler.java
+++ b/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyNoVNCHandler.java
@@ -38,7 +38,7 @@ import org.eclipse.jetty.websocket.servlet.WebSocketServletFactory;
 @WebSocket
 public class ConsoleProxyNoVNCHandler extends WebSocketHandler {
 
-    private ConsoleProxyNoVncClient viewer;
+    private ConsoleProxyNoVncClient viewer = null;
     private static final Logger s_logger = Logger.getLogger(ConsoleProxyNoVNCHandler.class);
 
     public ConsoleProxyNoVNCHandler() {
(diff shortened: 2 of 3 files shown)


rhtyd (approved):

rhtyd:
Shouldn't new console kick the old one out?

weizhouapache:
> Shouldn't new console kick the old one out?

@rhtyd in old ajax viewer, no

rhtyd:
@ustcweizhou I think for the new novnc console, opening new window/session should close the previous on. 
@davidjumani can you review ?

weizhouapache:
> @ustcweizhou I think for the new novnc console, opening new window/session should close the previous on.
> @davidjumani can you review ?

@rhtyd 
there are 3 options
(1) reject new session, and keep old session.
(2) accept new session, and kick out old session
(3) keep both sessions

The current behavior (reject the first new session and accept the second new session) seems to be bug.

we have (3) in our production but the implementation is different from master.
we think (1) is better than (2).

@davidjumani @DaanHoogland @PaulAngus @andrijapanicsb your opinion ?

davidjumani:
@rhtyd Keeping both sessions should not be possible since an exclusive flag is sent when viewing the console
I'm for (1) - Reject the new and keep the old (just shorten the idle timeout)

davidjumani:
@blueorangutan package

rhtyd:
Last time I tried to do a remote vnd - if you try to connect, it usually bounces off older/other consoles and grants the new one to be opened. What's the behaviour in old console proxy? I think if we're rejecting the new console then a suitable msg but be shown.

DaanHoogland:
without any extra €0.02, option 1 has my vote but rohit's remark is valid, something like "concole session active" should do. we don't have to be verbose.

andrijapanic-dont-use-this-one:
Old VNC console is to kick out the previous session.

If we do the opposite, at very least a proper message should be shown to user, while making sure that an old session is terminated after some inactivity period, otherwise a colleague helping another colleague would not be able to take over the console session, which impacts usability. 

I'm personally for the old behaviour - kick off the old session (with a message explaining that another session has been established elsewhere) and allow the new session - it simply makes more sense and is more user-friendly.

andrijapanicsb:
um, that was me above ^^^

weizhouapache:
it looks difficult to get an decision.
would it be better to add global setting and let users choose ?

andrijapanicsb:
+1, with perhaps defaulting to the old VNC console behaviour (which should have been implemented in the first place)

rhtyd:
Agree, I would prefer the original behaviour; new console should kick the old one out.

davidjumani:
New console should not be created if an existing one is open. Kicking out the old session will cause issues especially when there are multiple users trying to access the console

andrijapanicsb:
That is how it worked so far @davidjumani and it allows colleagues who are jointly (remotely distanced people) troubleshooting a thing.

We all need to observe things from the user and usability perspective.

shwstppr:
Do we have a consensus on this? Can we have new global setting to switch between new session kicking our ol session and new session being rejected?

DaanHoogland:
> @rhtyd
> there are 3 options
> (1) reject new session, and keep old session.
> (2) accept new session, and kick out old session
> (3) keep both sessions
> 
> The current behavior (reject the first new session and accept the second new session) seems to be bug.

I think given discussions here we should go for a tristate configuration option, letting the operator decide. Given that (2) is the old behaviour it should then be the default. agree?

rhtyd:
+1 makes sense, go with backward compatibility

rhtyd:
ping @ustcweizhou can you make the changes? Or introduce a global setting to toggle behaviour?

weizhouapache:
> ping @ustcweizhou can you make the changes? Or introduce a global setting to toggle behaviour?

@rhtyd I will work on it.

weizhouapache:
@rhtyd @DaanHoogland @davidjumani 
as discussed, updated this pr to accept new vm console session and disconnect old session.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

davidjumani (approved):
Works like the old console. The old session is terminated when a new one is opened

rhtyd (approved):

PR #4537, opened 2020-12-13
CLOUDSTACK-10424:Potential sensitive information disclosure 

lujiefsi:
### Description

This PR fixing CLOUDSTACK-10424

### Types of changes

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
+++ b/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
@@ -612,7 +612,7 @@ public class AsyncJobManagerImpl extends ManagerBase implements AsyncJobManager,
                         } else {
                             // TODO, job wakeup is not in use yet
                             if (s_logger.isTraceEnabled())
-                                s_logger.trace("Unable to find a wakeup dispatcher from the joined job " + job.getId());
+                                s_logger.trace("Unable to find a wakeup dispatcher from the joined job:" + StringUtils.cleanString(job.toString()));
                         }
                     } else {
                         AsyncJobDispatcher jobDispatcher = getDispatcher(job.getDispatcher());
--- a/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
+++ b/framework/jobs/src/main/java/org/apache/cloudstack/framework/jobs/impl/AsyncJobManagerImpl.java
@@ -612,7 +612,7 @@ public class AsyncJobManagerImpl extends ManagerBase implements AsyncJobManager,
                         } else {
                             // TODO, job wakeup is not in use yet
                             if (s_logger.isTraceEnabled())
-                                s_logger.trace("Unable to find a wakeup dispatcher from the joined job: " + job);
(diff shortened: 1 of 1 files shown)


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland (commented):
I don't agree.

lujiefsi:
> I don't agree.

Hum, I give this patch for two reasons:

- this log just indicate that "Unable to find a wakeup dispatcher from the joined job". I have check the code, it finds the dispatcher by jobId. so there's no need to print all detail of the job in this case, just jobID is ok, as other TRACE level log in this file,
- if we enable TRACE level log for some reason, the password can be logged into files  which can then be propagated to other log systems, i.e Splunk where we can't make sure the log file are protected. 

I know current change is bad for debugging, but i still think the sensitive information  is more important, and this case, password is not usefull to debug. In order to make the log be useful to debug, i will change the patch as:
`s_logger.trace("Unable to find a wakeup dispatcher from the joined job:" + StringUtils.cleanString(job.toString()));`

Hope for more discussion!

rhtyd:
then trace logs are enabled, you'll even see SQL queries in the logs. Trace is supposed to show sensitive information and used for investigation/debugging when normal logs have failed and isn't enabled by default. I agree with Daan, -1.

lujiefsi:
> When trace logs are enabled, you'll even see SQL queries in the logs. Trace is supposed to show sensitive information and used for investigation/debugging when normal logs have failed and isn't enabled by default. I agree with Daan, -1.

OK, but I still think we should document this to tell sysadmin that sensitive information are logged and need carefully handle.

DaanHoogland:
@lujiefsi I thank you for your diligence and I can live with sanitised logs and I agree that a password is only useful when debugging the authentication. There are many things wrong with our logging and passwords being logged for the wrong reasons is one of them. If this change suits your need, i wont block it. cc @rhtyd

rhtyd:
I still don't agree, we don't enable trace logs in production unless we're investigating something. It may be a useful property to see unmasked/unsanitised data. Instead if there's any specific VO or job or command parameter that shouldn't be logged, it should be fixed there than a general fix; for example explicit log off annoation for a command can be done like this: https://github.com/apache/cloudstack/blob/master/core/src/main/java/org/apache/cloudstack/ca/SetupKeyStoreCommand.java#L31

lujiefsi:
> I still don't agree, we don't enable trace logs in production unless we're investigating something. It may be a useful property to see unmasked/unsanitised data. Instead if there's any specific VO or job or command parameter that shouldn't be logged, it should be fixed there than a general fix; for example explicit log off annoation for a command can be done like this: https://github.com/apache/cloudstack/blob/master/core/src/main/java/org/apache/cloudstack/ca/SetupKeyStoreCommand.java#L31
> 
> Let's revisit this towards 4.16.

It looks like a good solution!

rhtyd:
Hi, the PR got auto-closed due to no known source repository or some other Github-specific issue during the recent renaming of the default branch to `main`. If you wish kindly submit a new PR.

Issue #4547, opened 2020-12-16
wrong root disk during the ingest bug


DK101010:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
core,api
~~~

##### CLOUDSTACK VERSION

~~~
version > 4.14
~~~

##### OS / ENVIRONMENT

vmware env.

##### SUMMARY

VCenter api give a list of disks with a lexicographic order. CS expect that the root disk the first element of this list is. But this is not always the case.

##### STEPS TO REPRODUCE

~~~
create a vm with min. one data disk in vcenter, detach root disk, attach again, ingest it to cs
~~~

##### EXPECTED RESULTS

~~~
Check if first disk is really the root disk and if not search and use the root disk 
~~~

##### ACTUAL RESULTS

~~~
ingested vm can not start because wrong root disk
~~~

DK101010:
PR https://github.com/apache/cloudstack/pull/4548

shwstppr:
As discussed in #4548, there isn't any foolproof way to find root disk listUnmanagedInstance output or SDK APIs. PR #4591 has been created to add a new param in import API to allow the caller of API to assign a particular disk as ROOT volume.

shwstppr:
Fixed with #4591

PR #4548, opened 2020-12-16
Bug/disk order bug during ingest

DK101010:
### Description

This PR...

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Adapt mockito test to check the new behavior

### DIFF
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -7239,7 +7239,9 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                                     instanceDisk.setController(DiskControllerType.getType(device.getClass().getSimpleName()).toString());
                                     instanceDisk.setControllerUnit(((VirtualSCSIController) device).getBusNumber());
                                 } else {
-                                    instanceDisk.setController(DiskControllerType.none.toString());
+                                    String msg = String.format("Abort ingest process because controller of volume %s is unsupported", instanceDisk.getLabel());
+                                    s_logger.error(msg);
+                                    throw new Exception(msg);
                                 }
                                 break;
                             }
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -7239,7 +7239,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                                     instanceDisk.setController(DiskControllerType.getType(device.getClass().getSimpleName()).toString());
                                     instanceDisk.setControllerUnit(((VirtualSCSIController) device).getBusNumber());
                                 } else {
-                                    String msg = String.format("Abort ingest process because controller of volume %s is unsupported", instanceDisk.getLabel());
+                                    String msg = String.format("Controller of disk %s is unsupported", instanceDisk.getLabel());
                                     s_logger.error(msg);
                                     throw new Exception(msg);
                                 }
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -7266,6 +7266,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                         instanceDisks.add(instanceDisk);
                     }
                 } catch (Exception e) {
+                    instanceDisks.clear();
                     s_logger.info("Unable to retrieve unmanaged instance disk info. " + e.getMessage());
                 }
             }
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -7274,9 +7274,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                 }
 
                 int extractInt(UnmanagedInstanceTO.Disk disk) {
-                    String num = disk.getLabel().replaceAll("\\D", "");
-                    // return 0 if no digits found
-                    return num.isEmpty() ? 0 : Integer.parseInt(num);
+                    return disk.getPosition() + (disk.getControllerUnit() * 100);
                 }
             });
         }


server/src/main/java/org/apache/cloudstack/vm/UnmanagedVMsManagerImpl.java
```diff
@@ -942,7 +942,20 @@ private UserVm importVirtualMachineInternal(final UnmanagedInstanceTO unmanagedI
         if (CollectionUtils.isEmpty(unmanagedInstanceDisks)) {
             throw new ServerApiException(ApiErrorCode.INTERNAL_ERROR, String.format("No attached disks found for the unmanaged VM: %s", instanceName));
         }
-        final UnmanagedInstanceTO.Disk rootDisk = unmanagedInstance.getDisks().get(0);
+        UnmanagedInstanceTO.Disk rootDisk = unmanagedInstance.getDisks().get(0);
```
shwstppr, line 945: @DK101010 Since disks list for UnmanagedInstance is created by CloudStack only, will it be better to always prepare that list in a way that first disk is always ROOT disk.
Relevant code can be found in VmwareResource class.

DK101010: @shwstppr ahh ok, yes of course this is a better solution, i will check this.

server/src/main/java/org/apache/cloudstack/vm/UnmanagedVMsManagerImpl.java
```diff
@@ -1248,6 +1261,25 @@ public UserVmResponse importUnmanagedInstance(ImportUnmanagedInstanceCmd cmd) {
         return cmdList;
     }
 
+    /**
+     * VCenter send a list of disk in lexicographical order, but not in every
+     * case is hard disk one the root disk. Therefore we must check if hd one
+     * the root disk and if not we search it and add it to the top of the
+     * list
```
GabrielBrascher, line 1268: I see some points to improve here and took the liberty to propose the following Javadoc block:

```
VCenter sends a list of disks in lexicographical order, but not in every case the first volume on the list is the root volume. Therefore this method checks if the first element of the list is the root disk; if not, it searches it and adds it to the top of the list
```

Disclaimer 1: I changed from hard disk to "volume" due to the term hard disks being related to the technology/device used to store volumes. In many DCs Hard Disks are being mixed with (or replaced by) Solid State Disks.

Disclaimer 2: as far as I know, this method is not used only in VCenter cases. We have some situations on CloudStack where methods are used only by a hypervisor/storage technology, but I think that this is not the case. Therefore, maybe the documentation should be enhanced to avoid misunderstandings on that matter.

server/src/test/java/org/apache/cloudstack/vm/UnmanagedVMsManagerImplTest.java
```diff
@@ -368,6 +358,12 @@ public void importUnmanagedInstanceTest() {
         when(importUnmanageInstanceCmd.getName()).thenReturn("TestInstance");
         when(importUnmanageInstanceCmd.getAccountName()).thenReturn(null);
         when(importUnmanageInstanceCmd.getDomainId()).thenReturn(null);
+        Map<String,Long> diskOfferings = new HashMap();
+        getDisks().forEach((disk) -> {
+            diskOfferings.put(disk.getDiskId(), 43L);
+        });
+        diskOfferings.remove("WrongRootDisk");
+        when(importUnmanageInstanceCmd.getDataDiskToDiskOfferingList()).thenReturn(diskOfferings);
```
GabrielBrascher, line 366: I don't see how this test method is validating their PR.

I was expecting some sort of assertion or validation on the lists returned to ensure that the implemented method is indeed returning the correct order?

GabrielBrascher (commented):
@DK101010 thanks for investing some time in contributing to the project. I see that it is your first Pull Request, so ... be welcome :-)

I see that the code still will be updated to address @shwstppr; therefore, my review here may not fit after these changes.

I would like to bring some points that can be improved, with these points I would also raise some attention to the lack of Pull Request description. A PR description is a key piece of each contribution, being used as future reference if/when necessary. Can you please enhance the PR description?

DK101010:
> I would like to bring some points that can be improved, with these points I would also raise some attention to the lack of Pull Request description. A PR description is a key piece of each contribution, being used as future reference if/when necessary. Can you please enhance the PR description?

@GabrielBrascher  thanks for your welcome :).
Sure, my infos are small, I've added a corresponding issue https://github.com/apache/cloudstack/issues/4547. Issue Information and PR information have intersections, so I thought it's enough to descripted the problem in the issue. When you say it's a key piece, is it necessary to create issues?

server/src/main/java/org/apache/cloudstack/vm/UnmanagedVMsManagerImpl.java
DK101010: Jepp, but on vmware side they are called hard disks. but i agree that hard disk suggests otherwise

server/src/test/java/org/apache/cloudstack/vm/UnmanagedVMsManagerImplTest.java
DK101010: You get an exception if it not work. I oriented myself here on the existing mockito tests.

GabrielBrascher: Got it, makes sense then.
I am just too used with the _Assert_ mentality on JUnit Tests; but it looks good then.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -7274,9 +7274,7 @@ public int compare(final UnmanagedInstanceTO.Disk disk1, final UnmanagedInstance
                 }
 
                 int extractInt(UnmanagedInstanceTO.Disk disk) {
-                    String num = disk.getLabel().replaceAll("\\D", "");
-                    // return 0 if no digits found
-                    return num.isEmpty() ? 0 : Integer.parseInt(num);
+                    return disk.getPosition() + (disk.getControllerUnit() * 100);
```
shwstppr, line 7277: @DK101010 In the code above this section we set ControllerUnit only for `VirtualIDEController` and `VirtualSCSIController` device type otherwise it remains null. Should we add a null check here?

DK101010: @shwstppr Good point, But how realistic is it to say that we have a volume with a position and without a controller id.? I talked with my administration and network colleagues and other controller types like NVME and SATA have also a controller id.
Therefore I think about it to add an exception in the else condition? What do you think?

shwstppr: @DK101010 yes, exception seems appropriate

shwstppr (approved):
LGTM on changes. If we can add an exception as discussed here, https://github.com/apache/cloudstack/pull/4548/files#r545656572, that will be great

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
DK101010: @shwstppr I will add an Exception.

rhtyd (approved):

rhtyd:
@blueorangutan package

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
rhtyd: @shwstppr @DK101010 will throwing an exception cause VM to be not ingested?

rhtyd:
@blueorangutan test

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
DK101010: @rhtyd damn, your are right, that was to quick from my side, the catch block catch all exceptions :(. 

Hmm ... is it ok to remove the catch block ? 

In my opinion the ingest process should be canceled if not all volumes ingestible. What do you think @rhtyd @shwstppr ?

rhtyd: Yes, as long as that's acceptable

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -7239,7 +7239,9 @@ private static File fetchSystemVmKeyFile() {
                                     instanceDisk.setController(DiskControllerType.getType(device.getClass().getSimpleName()).toString());
                                     instanceDisk.setControllerUnit(((VirtualSCSIController) device).getBusNumber());
                                 } else {
-                                    instanceDisk.setController(DiskControllerType.none.toString());
+                                    String msg = String.format("Abort ingest process because controller of volume %s is unsupported", instanceDisk.getLabel());
```
shwstppr, line 7242:
```suggestion
                                    String msg = String.format("Controller of disk %s is unsupported", instanceDisk.getLabel());
```

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
shwstppr: I think IDE and generic SCSI (VirtualSCSIController) will cover all disk controllers. If we are getting any other type of VirtualController as a disk, an exception seems correct to me. However, the error message needs to be a bit more generic there as the same method is used for listing unmanaged isntances

shwstppr (approved):
Looks good to me except the error message which can be more generic

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
DK101010: @shwstppr Sure, we use also only IDE, SCSI controller but for me it sounds critical in case of other/new controller to ingest vm's without volumes . I think from user perspektive it is inconsistent to see that the vm is ingested but is unusable.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -7239,7 +7239,9 @@ private static File fetchSystemVmKeyFile() {
                                     instanceDisk.setController(DiskControllerType.getType(device.getClass().getSimpleName()).toString());
                                     instanceDisk.setControllerUnit(((VirtualSCSIController) device).getBusNumber());
                                 } else {
-                                    instanceDisk.setController(DiskControllerType.none.toString());
+                                    String msg = String.format("Controller of disk %s is unsupported", instanceDisk.getLabel());
+                                    s_logger.error(msg);
+                                    throw new Exception(msg);
```
rhtyd, line 7244: @DK101010 I think on failure you can decide to fail the entire VM ingestion by (a) returning an empty list and let the user of the `getUnmanageInstanceDisks` method handle the case (when disk list is empty throw an exception), or (b) in the generic exception handler, instead of Exception throw an `UnsupportedOperationException` add another handler for the specific exception. cc @shwstppr

shwstppr: @rhtyd @DK101010 same method is used when listUnmanagedINstances API is called so I'm not if we would want the complete API to raise an exception when disks for one particular VM is not found. In my opinion, we can log the error here and return the list of disks for the VM as empty. With an empty list of disks VM import with automatically fail with existing check
https://github.com/apache/cloudstack/blob/master/server/src/main/java/org/apache/cloudstack/vm/UnmanagedVMsManagerImpl.java#L942-L944

If we want to go with the above approach we can clear the `instanceDisks` list variable in the catch block

DK101010: @shwstppr @rhtyd ok, then I will modify the exception msg and clear the list in the catch block. Thanks you two for help.

DK101010: @shwstppr ahh  sorry, I had not checked that you have already changed the msg :D.

shwstppr:
@blueorangutan package

GabrielBrascher (approved):
Code LGTM. I did not test it, though.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -7274,9 +7277,7 @@ public int compare(final UnmanagedInstanceTO.Disk disk1, final UnmanagedInstance
                 }
 
                 int extractInt(UnmanagedInstanceTO.Disk disk) {
-                    String num = disk.getLabel().replaceAll("\\D", "");
-                    // return 0 if no digits found
-                    return num.isEmpty() ? 0 : Integer.parseInt(num);
+                    return disk.getPosition() + (disk.getControllerUnit() * 100);
```
shwstppr, line 7280: @DK101010 I'm not sure if this formula is giving correct position/order for the disk.
I tried listUnamanagedInstance API for a VM with 2 data disks and results were incorrect with this change
Screenshot from 2021-01-14 14-08-56

**With changes** - Root disk is showing up at second position in order
```
> list unmanagedinstances clusterid=2b35fefc-892a-4393-9ccf-30229c1a999b name=importtest
{
  "count": 1,
  "unmanagedinstance": [
    {
      "clusterid": "2b35fefc-892a-4393-9ccf-30229c1a999b",
      "cpucorepersocket": 1,
      "cpunumber": 1,
      "cpuspeed": 0,
      "disk": [
        {
          "capacity": 21474836480,
          "controller": "lsilogic",
          "controllerunit": 0,
          "datastorehost": "10.10.0.16",
          "datastorename": "4ea48dd0701d3385a5cbb8921ee1d1d4",
          "datastorepath": "/acs/primary/pr4385-t3314-vmware-65u2/pr4385-t3314-vmware-65u2-esxi-pri2",
          "datastoretype": "NFS",
          "id": "35-2000",
          "imagepath": "[4ea48dd0701d3385a5cbb8921ee1d1d4] importtest/importtest_61.vmdk",
          "label": "Hard disk 2",
          "position": 0
        },
        {
          "capacity": 2147483648,
          "controller": "ide",
          "controllerunit": 0,
          "datastorehost": "10.10.0.16",
          "datastorename": "4ea48dd0701d3385a5cbb8921ee1d1d4",
          "datastorepath": "/acs/primary/pr4385-t3314-vmware-65u2/pr4385-t3314-vmware-65u2-esxi-pri2",
          "datastoretype": "NFS",
          "id": "35-3001",
          "imagepath": "[4ea48dd0701d3385a5cbb8921ee1d1d4] importtest/importtest_62.vmdk",
          "label": "Hard disk 1",
          "position": 1
        },
        {
          "capacity": 5368709120,
          "controller": "lsilogic",
          "controllerunit": 0,
          "datastorehost": "10.10.0.16",
          "datastorename": "4ea48dd0701d3385a5cbb8921ee1d1d4",
          "datastorepath": "/acs/primary/pr4385-t3314-vmware-65u2/pr4385-t3314-vmware-65u2-esxi-pri2",
          "datastoretype": "NFS",
          "id": "35-2001",
          "imagepath": "[4ea48dd0701d3385a5cbb8921ee1d1d4] importtest/importtest_63.vmdk",
          "label": "Hard disk 3",
          "position": 1
        }
      ],
      "hostid": "fd9a2641-ce6e-456e-be78-4b1b63c2b2ac",
      "memory": 512,
      "name": "importtest",
      "nic": [
        {
          "adaptertype": "E1000",
          "id": "Network adapter 1",
          "macaddress": "02:00:44:4f:00:12",
          "networkname": "cloud.guest.2183.200.1-vSwitch1",
          "vlanid": 2183
        }
      ],
      "osdisplayname": "CentOS 4/5 or later (64-bit)",
      "osid": "centos64Guest",
      "powerstate": "PowerOn"
    }
  ]
}
```

**Without change** - Root disk at first place
```
> list unmanagedinstances clusterid=2b35fefc-892a-4393-9ccf-30229c1a999b name=importtest
{
  "count": 1,
  "unmanagedinstance": [
    {
      "clusterid": "2b35fefc-892a-4393-9ccf-30229c1a999b",
      "cpucorepersocket": 1,
      "cpunumber": 1,
      "cpuspeed": 0,
      "disk": [
        {
          "capacity": 2147483648,
          "controller": "ide",
          "controllerunit": 0,
          "datastorehost": "10.10.0.16",
          "datastorename": "4ea48dd0701d3385a5cbb8921ee1d1d4",
          "datastorepath": "/acs/primary/pr4385-t3314-vmware-65u2/pr4385-t3314-vmware-65u2-esxi-pri2",
          "datastoretype": "NFS",
          "id": "35-3001",
          "imagepath": "[4ea48dd0701d3385a5cbb8921ee1d1d4] importtest/importtest_62.vmdk",
          "label": "Hard disk 1",
          "position": 1
        },
        {
          "capacity": 21474836480,
          "controller": "lsilogic",
          "controllerunit": 0,
          "datastorehost": "10.10.0.16",
          "datastorename": "4ea48dd0701d3385a5cbb8921ee1d1d4",
          "datastorepath": "/acs/primary/pr4385-t3314-vmware-65u2/pr4385-t3314-vmware-65u2-esxi-pri2",
          "datastoretype": "NFS",
          "id": "35-2000",
          "imagepath": "[4ea48dd0701d3385a5cbb8921ee1d1d4] importtest/importtest_61.vmdk",
          "label": "Hard disk 2",
          "position": 0
        },
        {
          "capacity": 5368709120,
          "controller": "lsilogic",
          "controllerunit": 0,
          "datastorehost": "10.10.0.16",
          "datastorename": "4ea48dd0701d3385a5cbb8921ee1d1d4",
          "datastorepath": "/acs/primary/pr4385-t3314-vmware-65u2/pr4385-t3314-vmware-65u2-esxi-pri2",
          "datastoretype": "NFS",
          "id": "35-2001",
          "imagepath": "[4ea48dd0701d3385a5cbb8921ee1d1d4] importtest/importtest_63.vmdk",
          "label": "Hard disk 3",
          "position": 1
        }
      ],
      "hostid": "fd9a2641-ce6e-456e-be78-4b1b63c2b2ac",
      "memory": 512,
      "name": "importtest",
      "nic": [
        {
          "adaptertype": "E1000",
          "id": "Network adapter 1",
          "macaddress": "02:00:44:4f:00:12",
          "networkname": "cloud.guest.2183.200.1-vSwitch1",
          "vlanid": 2183
        }
      ],
      "osdisplayname": "CentOS 4/5 or later (64-bit)",
      "osid": "centos64Guest",
      "powerstate": "PowerOn"
    }
  ]
}

```

shwstppr (changes requested):
Needs changes, testing. Order of ROOT disk is coming incorrect

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
DK101010: Hi @shwstppr, the list is correct sorted by controller id and position. Works like I want. Your root disk is on controller Id 0 and position 1 and you have two disk on the same position. I'm a little bit confused. In my mind vmware puts the root disk in every case on controller id 0 and position 0.

shwstppr: Hi @DK101010, I'm not sure why two disks have same position. To get this unmanaged VM. I just created a VM in cloudstack, added two data disks to it and then cloned the vm in vcenter.

DK101010: @shwstppr ahh ok you have two controller(ide, lsilogic), But in my opinion it makes no sense that you have one disk on one controller on position one. Also should increment the controller id, when you use more the one controller. And for this reason works this feature not correct. I think you found a other bug.

DK101010: Hi @shwstppr, I talked with my team about this behavior. First, your controller ids are correct. Increments are per controller type. For this reason is my implementation not better as the old one. Perhaps we could check in the sort algorithm if device type == scsi controller, but then it isn't no longer possible to boot from other controller types. It exist no safety way to find out which disk is bootable. The only way to get sure that  CS use the right volume is that the user indicates during the ingest process via rest call which volume is the right. Therefore I think we should close this pull request. What do you think?

shwstppr: @DK101010 then is it better to add a new param in importUnamangedInstance API, something like `rootdiskid`? cc @rhtyd @PaulAngus

DK101010: @shwstppr yes, for example, but in my mind it should be optional for the user and when not set then works like now via sorting of disk names.

shwstppr:
@DK101010 closed in favour of #4591. Please re-open if needed

PR #4553, opened 2020-12-19
Fix for mapping guest OS type read from OVF to existing guest OS in C…

harikrishna-patnala:
### Description
This fix address the issue when CloudStack does reverse look up for guest OS type read from OVF to guest OS type in CloudStack during VMware template registration. After reading the OS type from OVF and while trying to map in CS, currently the first in the list of guest_os_hypervisor table with guest os name is picked and assigned to template. This cannot be true always as there are cases that multiple entries in guest_os_hypervisor table can be of different versions of same OS or it can also be different types of OS as well (in case of types like Other 32bit or otherGuest)

For example
image

"otherGuest" os name in OVF file can map to guest_os_id either to "85" or "60". But these two are completely different OS types and there chances that wrong guest OS is marked to template. 

We are trying to fix the problem using the description field of the OS in OVF file and equals ignore match with display name in guest_os table. If description field is not present or if it does not match with any of guest os display names, then last in the list of guest_os_hypervisor table is picked so that latest version of that guest OS is assigned to the template.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [x] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested registering a template with different types of OSes in OVF file.
1. OVF file containing osType: darwin18_64Guest
           - CS maps to single entry in guest_os_hypervisor table and OS type is assigned to "macOS 10.14 (64 bit)"
2. OVF file containing osType: "otherGuest' and Description: "(other 32-bit)"
           - CS maps to two entries in guest_os_hypervisor table, but based on description it assigned to "other 32-bit"
 
image

3. OVF file containing osType: "otherGuest" and no Description
           - CS maps to two entries in guest_os_hypervisor table but since no description is there CS maps to the latest guest os type which is "other 32-bit"
           MariaDB [cloud]> select * from guest_os where id in (85,60) order by id DESC;
+----+-------------+------+--------------------------------------+------------------+---------------------+---------+-----------------+
| id | category_id | name | uuid                                 | display_name     | created             | removed | is_user_defined |
+----+-------------+------+--------------------------------------+------------------+---------------------+---------+-----------------+
| 85 |           9 | NULL | 06bb4c79-1e50-11eb-bde5-1e005f013ade | SCO OpenServer 5 | 2020-11-04 03:44:52 | NULL    |               0 |
| 60 |           7 | NULL | 06b43ca5-1e50-11eb-bde5-1e005f013ade | Other (32-bit)   | 2020-11-04 03:44:52 | NULL    |               0 |
+----+-------------+------+--------------------------------------+------------------+---------------------+---------+-----------------+

### DIFF
--- a/engine/storage/src/main/java/org/apache/cloudstack/storage/image/deployasis/DeployAsIsHelperImpl.java
+++ b/engine/storage/src/main/java/org/apache/cloudstack/storage/image/deployasis/DeployAsIsHelperImpl.java
@@ -167,8 +167,26 @@ public class DeployAsIsHelperImpl implements DeployAsIsHelper {
                 hypervisor.toString(), minimumHypervisorVersion);
 
         if (CollectionUtils.isNotEmpty(guestOsMappings)) {
-            GuestOSHypervisorVO mapping = guestOsMappings.get(0);
-            return mapping.getGuestOsId();
+            Long guestOsId = null;
+            if (guestOsMappings.size() == 1) {
+                GuestOSHypervisorVO mapping = guestOsMappings.get(0);
+                guestOsId = mapping.getGuestOsId();
+            } else {
+                if (!StringUtils.isEmpty(guestOsDescription)) {
+                    for (GuestOSHypervisorVO guestOSHypervisorVO : guestOsMappings) {
+                        GuestOSVO guestOSVO = guestOSDao.findById(guestOSHypervisorVO.getGuestOsId());
+                        if (guestOsDescription.equalsIgnoreCase(guestOSVO.getDisplayName())) {
+                            guestOsId = guestOSHypervisorVO.getGuestOsId();
+                            break;
+                        }
+                    }
+                }
+                if (null == guestOsId) {
+                    GuestOSHypervisorVO mapping = guestOsMappings.get(guestOsMappings.size()-1);
+                    guestOsId = mapping.getGuestOsId();
+                }
+            }
+            return guestOsId;
         } else {
             throw new CloudRuntimeException("Did not find a guest OS with type " + guestOsType);
         }
@@ -301,7 +319,7 @@ public class DeployAsIsHelperImpl implements DeployAsIsHelper {
             }
             String propValue = null;
             try {
-                 propValue = getValueFromInformationTO(informationTO);
+                propValue = getValueFromInformationTO(informationTO);
             } catch (RuntimeException re) {
                 LOGGER.error("gson marshalling of property object fails: " + propKey,re);
             } catch (IOException e) {
--- a/utils/src/main/java/com/cloud/utils/StringUtils.java
+++ b/utils/src/main/java/com/cloud/utils/StringUtils.java
@@ -124,7 +124,7 @@ public class StringUtils {
 
     /**
      * Converts a List of tags to a comma separated list
-     * @param tags
+     * @param tagsList
      * @return String containing a comma separated list of tags
      */
 
@@ -312,5 +312,43 @@ public class StringUtils {
 
     public static String toCSVList(final List<String> csvList) {
         return join(csvList, ",");
+    }
+
+    public static int min(int x, int y, int z)
+    {
+        if (x <= y && x <= z)
+            return x;
+        if (y <= x && y <= z)
+            return y;
+        else
+            return z;
+    }
+
+    /**
+     * Calculates minimum number of edits required to convert from one string to another string.
+     * @param str1
+     *      String that needs editing
+     * @param str2
+     *      Target string that needs after editing
+     * @return minimum number of edits required to convert str1 to str2
+     */
+    public static int minimumEditDistance(String str1, String str2) {
+        int m = str1.length();
+        int n = str2.length();
+        int[][] auxiliaryArray = new int[m + 1][n + 1];
+
+        for (int i = 0; i <= m; i++) {
+            for (int j = 0; j <= n; j++) {
+                if (i == 0)
+                    auxiliaryArray[i][j] = j;
+                else if (j == 0)
+                    auxiliaryArray[i][j] = i;
+                else if (str1.charAt(i - 1) == str2.charAt(j - 1))
+                    auxiliaryArray[i][j] = auxiliaryArray[i - 1][j - 1];
+                else
+                    auxiliaryArray[i][j] = 1 + min(auxiliaryArray[i][j - 1], auxiliaryArray[i - 1][j], auxiliaryArray[i - 1][j - 1]);
+            }
+        }
+        return auxiliaryArray[m][n];
     }
 }
--- a/utils/src/main/java/com/cloud/utils/StringUtils.java
+++ b/utils/src/main/java/com/cloud/utils/StringUtils.java
@@ -316,10 +316,8 @@ public class StringUtils {
 
     /**
      * Calculates minimum number of edits required to convert from one string to another string.
-     * @param str1
-     *      String that needs editing
-     * @param str2
-     *      Target string that needs after editing
+     * @param str1      String that needs editing
+     * @param str2      Target string that is required/expected after editing
      * @return minimum number of edits required to convert str1 to str2
      */
     public static int minimumEditDistance(String str1, String str2) {
--- a/utils/src/main/java/com/cloud/utils/StringUtils.java
+++ b/utils/src/main/java/com/cloud/utils/StringUtils.java
@@ -124,7 +124,7 @@ public class StringUtils {
 
     /**
      * Converts a List of tags to a comma separated list
-     * @param tagsList
+     * @param tags
      * @return String containing a comma separated list of tags
      */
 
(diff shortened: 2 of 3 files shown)


harikrishna-patnala:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test centos7 vmware-65u3

rhtyd:

rhtyd (approved):

shwstppr (approved):
Code changes lgtm

rhtyd:
Marking it as blocker - the deploy as-is created regression in VM template registration.

utils/src/main/java/com/cloud/utils/StringUtils.java
```diff
@@ -313,4 +313,42 @@ public static String mapToString(final Map<String, String> map) {
     public static String toCSVList(final List<String> csvList) {
         return join(csvList, ",");
     }
+
+    public static int min(int x, int y, int z)
+    {
+        if (x <= y && x <= z)
+            return x;
+        if (y <= x && y <= z)
```
DaanHoogland:
```suggestion
        if (y <= z)
```

harikrishna-patnala:
@blueorangutan package

engine/storage/src/main/java/org/apache/cloudstack/storage/image/deployasis/DeployAsIsHelperImpl.java
```diff
@@ -174,8 +174,26 @@ private void handleGuestOsFromOVFDescriptor(long templateId, String guestOsType,
...
-            GuestOSHypervisorVO mapping = guestOsMappings.get(0);
-            long guestOsId = mapping.getGuestOsId();
+            Long guestOsId = null;
+            if (guestOsMappings.size() == 1) {
+                GuestOSHypervisorVO mapping = guestOsMappings.get(0);
+                guestOsId = mapping.getGuestOsId();
+            } else {
+                if (!StringUtils.isEmpty(guestOsDescription)) {
+                    int minimumEdits = Integer.MAX_VALUE;
+                    for (GuestOSHypervisorVO guestOSHypervisorVO : guestOsMappings) {
```
DaanHoogland, line 184: some small worry here. I'm not sure it applies to guestos names, but linguistically this can yield false positives easily;
- "van os"
- "vanos"
- "fan nos"
- "vamos"
- "fam. os"
with the bare eye the last one is the proper replacement of the first. Would this algorithm yield this?

utils/src/main/java/com/cloud/utils/StringUtils.java
```diff
@@ -313,4 +313,42 @@ public static String mapToString(final Map<String, String> map) {
...
+        else
+            return z;
+    }
+
+    /**
+     * Calculates minimum number of edits required to convert from one string to another string.
+     * @param str1
+     *      String that needs editing
+     * @param str2
+     *      Target string that needs after editing
```
DaanHoogland, line 322:
```suggestion
     * @param str1     String that needs editing
     * @param str2     Target string that is required/expected after editing
```

utils/src/main/java/com/cloud/utils/StringUtils.java
```diff
@@ -313,4 +313,42 @@ public static String mapToString(final Map<String, String> map) {
...
+     * @return minimum number of edits required to convert str1 to str2
+     */
+    public static int minimumEditDistance(String str1, String str2) {
```
DaanHoogland, line 325: can we have some unit tests for this, please?
I think it assumes a definition of what an edit is, that we need to make explicit as well.

DaanHoogland (commented):
I think the assuption of what is "like" another string is so dangerous that we might as well just pick the last if the strings aren't equal.

utils/src/main/java/com/cloud/utils/StringUtils.java
harikrishna-patnala: I've taken out minimum method and used min(min(x,y),z)

utils/src/main/java/com/cloud/utils/StringUtils.java
harikrishna-patnala: Added unit tests in new commit.

utils/src/test/java/com/cloud/utils/StringUtilsTest.java
```diff
@@ -258,4 +258,34 @@ public void testToCSVList() {
...
+        String str2 = "Other 32-bit";
+        int minDistance = StringUtils.minimumEditDistance(str1, str2);
+        assertEquals(minDistance, 0);
+    }
+
+    @Test
+    public void testBestMatchStringWithEditDistance() {
+        String str1 = "FreeBSD 11 (32bit)";
+        String str2 = "FreeBSD 12 (64bit)";
+        String targetString = "FreeBSD 64bit";
```
DaanHoogland, line 274: maybe not here but as extra:
```suggestion
    public void testBestFreeBSDWithEditDistance() {
        String str1 = "FreeBSD 11 (32bit)";
        String str2 = "FreeBSD 22 (64bit)";
        String targetString = "FreeBSD 11 (64bit)";
```

```suggestion
    public void testBestFedoraWithEditDistance() {
        String str1 = "Fedora 29 (32bit)";
        String str2 = "FreeBSD 32 (64bit)";
        String targetString = "FreeBSD 33 (32bit)";
```

```suggestion
    public void testBestUbuntuWithEditDistance() {
        String str1 = "ubuntu 14.04 (32bit)";
        String str2 = "ubuntu 20.10 (64bit)";
        String targetString = "ubuntu 18.04 (64bit)";
```

nvazquez (commented):
I like the proposed approach, but I'm also thinking that we could check the longest prefix match or a mix of both approaches.

Examples:

1)
Ubuntu 4.10 
Ubuntu 8.10

2)
Ubuntu 8.04
Ubuntu 8.10

In case 1) we have minDistance = 1 however by longer prefix we could have chosen case 2)

Opinions? @DaanHoogland @rhtyd  @harikrishna-patnala @andrijapanicsb

DaanHoogland:
`ubuntu 18.04 (32bit)` matching 
`ubuntu 18.04 (64bit)` or
`ubuntu 20.10 (32bit)`, should surely match the 32 bit rather than the exact version. How can we encode this?

harikrishna-patnala:
> `ubuntu 18.04 (32bit)` matching
> `ubuntu 18.04 (64bit)` or
> `ubuntu 20.10 (32bit)`, should surely match the 32 bit rather than the exact version. How can we encode this?

since we are using a generic algorithm here I think if we have to handle these kind of special cases we need to add special checks for bits. Likewise if there is some other scenario we need to handle that separately too.. Since this string comes from a description 32 bit can written in any format (32-bit, 32bit, 32 bit, 32_bit). So atleast in this case I don't see a generic solution at the moment.

utils/src/main/java/com/cloud/utils/StringUtils.java
harikrishna-patnala: Updated this section with suggested changes

DaanHoogland:
fair enough @harikrishna-patnala . I don't think we can rely on anything but an unlikely exact match.

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

harikrishna-patnala:
@DaanHoogland @rhtyd @nvazquez updated the logic to equals ignore case (removed min edit distance algorithm). I've retested the cases mentioned in the description. Please review

rhtyd:
@blueorangutan package

rhtyd (approved):

engine/storage/src/main/java/org/apache/cloudstack/storage/image/deployasis/DeployAsIsHelperImpl.java
```diff
@@ -174,8 +174,25 @@ private void handleGuestOsFromOVFDescriptor(long templateId, String guestOsType,
...
+                if (!StringUtils.isEmpty(guestOsDescription)) {
+                    for (GuestOSHypervisorVO guestOSHypervisorVO : guestOsMappings) {
+                        GuestOSVO guestOSVO = guestOSDao.findById(guestOSHypervisorVO.getGuestOsId());
+                        if (guestOsDescription.equalsIgnoreCase(guestOSVO.getDisplayName())) {
+                            guestOsId = guestOSHypervisorVO.getGuestOsId();
+                            break;
+                        }
+                    }
+                }
+                if (null == guestOsId) {
```
rhtyd, line 191: Minor nit - I think variable == null maybe more readable, but as I said that's not picking

DaanHoogland: depends on how accustomed you are in right to left reading ;)

DaanHoogland (approved):
I think we can agree this is the least bad solution (for now)

rhtyd:

rhtyd (approved):

engine/storage/src/main/java/org/apache/cloudstack/storage/image/deployasis/DeployAsIsHelperImpl.java
harikrishna-patnala: making habit of this to avoid setting null to the variable by mistake, in that case IDE also don't know if it is logically wrong "if (guestOsId = null)" :p

harikrishna-patnala:
Resolved the merge conflicts. @rhtyd @nvazquez Please do the last review

harikrishna-patnala:
@blueorangutan package

rhtyd (approved):

nvazquez (approved):
LGTM

PR #4565, opened 2021-01-04
Fix RBD primary storage host port null error

div8cn:
### Description
Ubuntu 20.04 +KVM +CEPH 15.2

Add RBD main storage through UI, it will fail when there is no host port parameter;
Because when we created the pool, we did not add the port target in the xml

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial


plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
```diff
@@ -84,7 +84,11 @@ public boolean parseDomainXML(String domXML) {
                     String authUserName = getAttrValue("auth", "username", disk);
                     String poolUuid = getAttrValue("secret", "uuid", disk);
                     String host = getAttrValue("host", "name", disk);
-                    int port = Integer.parseInt(getAttrValue("host", "port", disk));
+                    int port = 0;
+                    String _xmlPort = getAttrValue("host", "port", disk);
+                    if ( ! _xmlPort.isEmpty()) {
```
rhtyd, line 89: Should we do both null and empty check?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtStoragePoolXMLParser.java
```diff
@@ -55,7 +55,11 @@ public LibvirtStoragePoolDef parseStoragePoolXML(String poolXML) {
             String format = getAttrValue("format", "type", source);
 
             if (type.equalsIgnoreCase("rbd")) {
-                int port = Integer.parseInt(getAttrValue("host", "port", source));
+                int port = 0;
+                String _xmlPort = getAttrValue("host", "port", source);
+                if ( ! _xmlPort.isEmpty()) {
```
rhtyd, line 60: Should we do both null and empty check?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
wido: I would do that just for the sake of security

DaanHoogland: there is a lot of npe danger in `getAttrValue()`, none of teh parameters are checked. (meaning yes please)

rhtyd:
@div8cn with you comment https://github.com/apache/cloudstack/pull/4568#issuecomment-754321932 do we need this PR?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
DaanHoogland: I did an investigation of the dom code used in `getAttrValue()` and it returns an empty string if the element `"host"` does not have an attribute named `"port"`, so _xmlPort will not be null.

div8cn:
@rhtyd I think this PR is necessary,

The best practice is to let libvirt decide to use port 3300 or 6789 to communicate with CEPH.

Therefore, it is best not to enter any port when adding mon;

@wido What do you think?

wido:
I agree @div8cn , this PR is needed.

In the early days of Ceph you needed to specify the port. Now we should have libvirt/qemu/librados/librbd handle this. We shouldn't interfere.

People running Ceph monitors on alternative ports are rare, very rare. I have never encountered them recently.

So yes, just a hostname should be sufficient without a port number.

rhtyd:
@wido @DaanHoogland @GabrielBrascher @weizhouapache  should this be treated as a blocker then, or good to have?
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
GabrielBrascher: 1. @div8cn can you please change the variable name from `_xmlPort` to `xmlPort`?

According to [Java conventions](https://docs.oracle.com/javase/tutorial/java/nutsandbolts/variables.html), 

> while it's technically legal to begin your variable's name with "_", this practice is discouraged

2. Regarding the String validation. Is it possible to have a `tagNode` of length 0?

`getAttrValue()` has this piece of code:
```
        NodeList tagNode = eElement.getElementsByTagName(tag);
        if (tagNode.getLength() == 0) {
            return null;
        }
```

I would recommend, just for safety, to use something as `org.apache.commons.lang.StringUtils`; it provides String validations such as `StringUtils.isNotBlank(Str)`.

`isNotBlank` returns true only if not null, not empty, and not blank.

```
String xmlPort = getAttrValue("host", "port", disk);
if (StringUtils.isNotBlank(xmlPort)) {
     ....
}
```

GabrielBrascher:
@rhtyd it would be great to have this one on board, but I don't see reason in blocking a release because of it.

Any comments, @wido @DaanHoogland @div8cn @weizhouapache?

GabrielBrascher (approved):
LGTM based on code review.

rhtyd (approved):

PR #4569, opened 2021-01-05
Replicate `rootDiskController` behavior to `dataDiskController`.

GutoVeronezi:
### Description
The bus type to `data disk` volumes is hardcoded to `virtio` or `scsi`, when using virtio-scsi (or, based on the template type). Therefore, there is no way to specify the bus type to data disk volumes (as we have for root disks). 

This PR intends to replicate the `rootDiskController` behavior to `dataDiskController`, allowing the definition of the controller.

### Types of changes

- [x] Major
- [ ] Minor
### How Has This Been Tested?
It has been tested locally.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2545,9 +2545,10 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
                 diskBusType = getGuestDiskModel(vmSpec.getPlatformEmulator());
             }
 
-            // I'm not sure why previously certain DATADISKs were hard-coded VIRTIO and others not, however this
-            // maintains existing functionality with the exception that SCSI will override VIRTIO.
-            DiskDef.DiskBus diskBusTypeData = (diskBusType == DiskDef.DiskBus.SCSI) ? diskBusType : DiskDef.DiskBus.VIRTIO;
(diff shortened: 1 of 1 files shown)


RodrigoDLopez (approved):
Nice Enhancement @GutoVeronezi 
can you please explain to me how the local tests were carried out?

My tests consist of entering the details of the data disk (during the deployment process of an instance or via a graphical interface with a previously created instance) and then checking if the details informed are present in the dump XML of the target instance.

I performed manual tests with this branch. It is correct and working as it should.
LGTM

RodrigoDLopez:
@blueorangutan package

RodrigoDLopez:
@blueorangutan test

GutoVeronezi:
@RodrigoDLopez thanks for asking.

Actually the process is quite similar as which you described.

I made the whole process via GUI.

1. I created an instance.

2. With it stopped, I created a volume and attached to it.

3. I setted the `dataDiskController` setting as `ide` and started the instance. 

4. After that, I checked if the detail informed was in the dump XML of the instance.
    What I got was `<target dev = 'hda' bus = 'ide' />`, which was what I expected.

DaanHoogland:
@blueorangutan test

DaanHoogland (commented):
one concern; the default used to be `virtio` and now it is `null`? otherwise the code looks good

GutoVeronezi:
> one concern; the default used to be `virtio` and now it is `null`? otherwise the code looks good

@DaanHoogland 
The method `getDataDiskModelFromVMDetail` returns null if the `dataDiskController` setting is not configured.
If the method returns null, then it will keep the same behavior as before, so the default still is `virtio`.

DaanHoogland:
@GutoVeronezi you are right, i read with crooked eyes.

DaanHoogland (approved):

rhtyd (approved):

PR #4570, opened 2021-01-05
Externalize KVM Agent's option to change migration thread timeout

GutoVeronezi:
### Description
We noticed that sometimes migration in KVM fails due to a latency to move the VM process (domain) from the source host to the target. This has been noticed in Rhel7 and CentOS7. An exception can be observed due to the default (hard-coded) timeout that is extrapolated and ACS can't find the migrated VM's domain in the target host.

This PR intends to externalize that option in the KVM agent configuration. Default timeout will remain 10 seconds.

### Types of changes

- [x] Major
- [ ] Minor

### How Has This Been Tested?
It has been tested locally.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
@@ -74,9 +74,12 @@ import com.cloud.hypervisor.kvm.resource.MigrateKVMAsync;
 import com.cloud.hypervisor.kvm.resource.VifDriver;
 import com.cloud.resource.CommandWrapper;
 import com.cloud.resource.ResourceWrapper;
+import com.cloud.utils.PropertiesUtil;
 import com.cloud.utils.Ternary;
 import com.cloud.utils.exception.CloudRuntimeException;
 import com.google.common.base.Strings;
+import java.io.File;
+import org.apache.cloudstack.utils.security.KeyStoreUtils;
 
 @ResourceWrapper(handles =  MigrateCommand.class)
 public final class LibvirtMigrateCommandWrapper extends CommandWrapper<MigrateCommand, Answer, LibvirtComputingResource> {
@@ -84,6 +87,8 @@ public final class LibvirtMigrateCommandWrapper extends CommandWrapper<MigrateCo
     private static final String GRAPHICS_ELEM_END = "/graphics>";
     private static final String GRAPHICS_ELEM_START = "<graphics";
     private static final String CONTENTS_WILDCARD = "(?s).*";
+    private static final String DEST_DOMAIN_MIGRATE_RETRIEVE_TIMEOUT= "dest.domain.migrate.retrieve.timeout";
+    private static final int DEFAULT_DEST_DOMAIN_MIGREATE_RETRIEVE_TIMEOUT_IN_SECONDS = 10;
     private static final Logger s_logger = Logger.getLogger(LibvirtMigrateCommandWrapper.class);
 
     protected String createMigrationURI(final String destinationIp, final LibvirtComputingResource libvirtComputingResource) {
@@ -225,7 +230,9 @@ public final class LibvirtMigrateCommandWrapper extends CommandWrapper<MigrateCo
             }
             s_logger.info("Migration thread for " + vmName + " is done");
 
-            destDomain = migrateThread.get(10, TimeUnit.SECONDS);
+            int destDomainMigrateThreadTime = parseAgentPropertiesDestDomainMigrateThreadTime();
+
+            destDomain = migrateThread.get(destDomainMigrateThreadTime, TimeUnit.SECONDS);
 
             if (destDomain != null) {
                 deleteOrDisconnectDisksOnSourcePool(libvirtComputingResource, migrateDiskInfoList, disks);
@@ -286,6 +293,17 @@ public final class LibvirtMigrateCommandWrapper extends CommandWrapper<MigrateCo
         }
 
         return new MigrateAnswer(command, result == null, result, null);
+    }
+
+    private int parseAgentPropertiesDestDomainMigrateThreadTime() throws IOException {
+        final File agentPropertiesFile = PropertiesUtil.findConfigFile(KeyStoreUtils.AGENT_PROPSFILE);
+        if (agentPropertiesFile != null) {
+            String configValue = PropertiesUtil.loadFromFile(agentPropertiesFile).getProperty(DEST_DOMAIN_MIGRATE_RETRIEVE_TIMEOUT);
+            if(StringUtils.isNotBlank(configValue)) {
+                return Integer.parseInt(configValue);
+            }
+        }
+        return DEFAULT_DEST_DOMAIN_MIGREATE_RETRIEVE_TIMEOUT_IN_SECONDS;
     }
 
     /**


RodrigoDLopez (commented):
Hi @GutoVeronezi
Can you please explain to me the context? and how it has been tested?
Reading your description I got an idea, but I couldn't reproduce it.

GutoVeronezi:
@RodrigoDLopez 

I had two hosts with CentOS7.

I tried to migrate a VM from a host to another, but one of my hosts was with overloaded RAM, so, it would take more than 10 seconds (the hardcoded timeout) to migrate the VM.

With the externalized configuration, I could change the timeout to 20 seconds and my migration ran just fine.

DaanHoogland:
@GutoVeronezi makes sense functionally and the code looks good except for the naming, why do you use `dest` in the names of the method and the setting? isn't this a local setting of the agent we are on at this moment?

GutoVeronezi:
> @GutoVeronezi makes sense functionally and the code looks good except for the naming, why do you use `dest` in the names of the method and the setting? isn't this a local setting of the agent we are on at this moment?

@DaanHoogland 

Yes, it is local, but it refers to a migration from one host to the other, so it is verified if the VMs were correctly migrated from the source host to the `destination` host. The naming corresponds to the verification of the VM's DOM on the destination host`.

GutoVeronezi:
Can anyone review this PR?

DaanHoogland:
@wido @weizhouapache @GabrielBrascher @ravening can one of you guys look at this, please?

wido:
I honestly don't get it yet. I need to think about this for a moment.

GabrielBrascher:
@wido from my understanding, KVM with Rhel/CentOS does some kind of processing which results in Libvirt taking longer to get the domain ID when migrating the VM. Ubuntu, on the other hand, retrieves domain ID in less than 10 seconds.

Am I right, @GutoVeronezi?

GabrielBrascher (commented):
The variable `dest.domain.migrate.retrieve.timeout` is going to be placed at the `agent.properties`, right?

If so, I would recommend adding it on `agent/conf/agent.properties`. Something similar to what was done at PR #4585 where it was added commented lines as documentation as well as the variable with the default value.

wido:
> The variable `dest.domain.migrate.retrieve.timeout` is going to be placed at the `agent.properties`, right?
> 
> If so, I would recommend adding it on `agent/conf/agent.properties`. Something similar to what was done at PR #4585 where it was added commented lines as documentation as well as the variable with the default value.

In addition, shouldn't this be prefixed with 'vm' then as other migration options have the same prefix.

GutoVeronezi:
> @wido from my understanding, KVM with Rhel/CentOS does some kind of processing which results in Libvirt taking longer to get the domain ID when migrating the VM. Ubuntu, on the other hand, retrieves domain ID in less than 10 seconds.
> 
> Am I right, @GutoVeronezi?

Yes, that is the case.

GabrielBrascher:
@GutoVeronezi what do you think of @wido's suggestion?

At the [agent.properties](https://github.com/apache/cloudstack/blob/2a3849b6ca6728cd94d447662b19b3833bfa9d86/agent/conf/agent.properties#L98) there are already some parameters regarding VM migrations, such as `vm.migrate.speed`:

```
# set the vm migrate speed, by default, it will try to guess the speed of the guest network
# In MegaBytes per second
#vm.migrate.speed=0

# set target downtime at end of livemigration, the 'hiccup' for final copy. Higher numbers
# make livemigration easier, lower numbers may cause migration to never complete. Less than 1
# means hypervisor default (20ms).
#vm.migrate.downtime=0

# Busy VMs may never finish migrating, depending on environment. When its available, we will
# want to add support for autoconvergence migration flag which should fix this. Set an upper
# limit in milliseconds for how long live migration should wait, at which point VM is paused and
# migration will finish quickly.  Less than 1 means disabled.
#vm.migrate.pauseafter=0
```

Maybe adding this parameter nearby those (e.g. after `vm.migrate.pauseafter`) would help users to find it among other VM migration parameters. One option would be to change from `dest.domain.migrate.retrieve.timeout` to `vm.migrate.domain.retrieve.timeout`.

GutoVeronezi:
@GabrielBrascher @wido I have changed the config name, is there something else to do?

GabrielBrascher (approved):
Code LGTM

GabrielBrascher:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtMigrateCommandWrapper.java
```diff
@@ -288,6 +295,17 @@ Use VIR_DOMAIN_XML_SECURE (value = 1) prior to v1.0.0.
         return new MigrateAnswer(command, result == null, result, null);
     }
 
+    private int parseAgentPropertiesVmMigrateDomainRetrieveTimeout() throws IOException {
+        final File agentPropertiesFile = PropertiesUtil.findConfigFile(KeyStoreUtils.AGENT_PROPSFILE);
```
weizhouapache, line 299: @GutoVeronezi 
you can use `PropertiesStorage` instead.

GutoVeronezi: @weizhouapache I took a look into this class and I am a bit confused. It seems like a facade of `PropertiesUtils` mixed with an another purpouse. 

It seems to me that this class is badly structured, without a clear flow and with nonsense logs. I think it will add a meaningless complexity to the code instead of facility it.

In other hand, in PR https://github.com/apache/cloudstack/pull/4586 there is a proposal of a handler that will simplify this code. If it be approved, I will change this PR to use the handler.

GutoVeronezi:
@weizhouapache can you look at this?

GutoVeronezi:
Waiting for [PR 4586](https://github.com/apache/cloudstack/pull/4586/files) to be merged to use the `AgentPropertiesFileHandler`.

GabrielBrascher:
@rhtyd @GutoVeronezi my 2 cents here is: I think that PR #4586 has nothing to do with #4570.

The first one externalizes the KVM agent **storage reboot configuration** (#4586), while the second one (#4570) allows configuring **migration thread timeout**. They are completely different variables, that impact distinct components and change different behaviors.

The only thing that links both PRs is the `AgentPropertiesFileHandler`, which is basically using the same means to do a completely different thing.

I see it as a personal way of seeing. One could keep two PRs as they affect different parts and could easily be debugged and done separately. On the other hand, one could have them on a generic PR such as `Externalize some KVM variables`.

GutoVeronezi:
@DaanHoogland @GabrielBrascher @RodrigoDLopez @sureshanaparti  I'm pinging you here because you all reviewed #4585 and #4586.

#4585 introduced a handler that facilitate agent properties' file reading, which can be used here. However,  #4585 was merged and reverted, therefore I had to submit a new work (#5239), which was merged a few days ago. Now I rebased this one and applied only the real changes.

GutoVeronezi:
@blueorangutan package

GutoVeronezi:
@blueorangutan package

weizhouapache:
@blueorangutan test

RodrigoDLopez (commented):
Code LGTM.
I'll test it and then post my results as soon as possible.

GabrielBrascher (approved):
Proposal and Code LGTM

weizhouapache:
@blueorangutan package

DaanHoogland (approved):
clgtm

GutoVeronezi:
Could we re-run the tests in this one?

nvazquez:
@GutoVeronezi can you please fix the conflicts? We can start a new round of tests after that

GutoVeronezi:
> @GutoVeronezi can you please fix the conflicts? We can start a new round of tests after that

@nvazquez done, thanks!

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

PR #4575, opened 2021-01-08
Enhance log messages with host name

GabrielBrascher:
### Description

Numbers are great for computers, however, we (humans) deal better with names. Logging host's ID is good in terms of short messages; however, admins must know all hosts by their ID or look via DB/API queries to figure out which host the log message is pointing to.

This PR proposes an admin friendly approach for log messages by logging the host name instead of its ID.

An example of how it works currently log messages can be painful in large environments:

`Operation Failed vm's original host id: 334 new host id: 300 host id before state transition: 127`
Vs.
`Operation Failed vm's original host: node23.clusterA new host: node09.clusterA host before state transition: node01.clusterA`

### Types of changes

- [ ] Major
- [X] Minor


shwstppr (commented):
@GabrielBrascher can we use something like `host-name(ID: host_uuid)`? Personally I found it easier at times to debug issues with uuid in logs

@blueorangutan package

GabrielBrascher:
@shwstppr That is something that I can add. Maybe this is something that we can discuss (on the mailing list or a Github issue) ion order to find conventions that best fit CloudStack log messages.

Nowadays we do not have any convention.

RodrigoDLopez:
Hey @shwstppr what do you suggest?

I totally agree with @GabrielBrascher, normally when we are talking about a host we use their names to refer to it, I never met someone who uses UUID for this purpose.
Maybe, in some cases, the UUID turn easy to find these hosts on the database, but with a simple change on the query, we have the same result.
And for me, something like
> Operation Failed vm's original host: node23.clusterA (ID: 98c37930-599f-11eb-ae93-0242ac130002) new host: node09.clusterA (ID: bdb22ea8-599f-11eb-ae93-0242ac130002) host before state transition: node01.clusterA (ID: bdb22598-599f-11eb-ae93-0242ac130002)

it is too verbose and does not add anything to the debug factor.
Presenting only the hostname in the log makes it easy to identify, as it is the way we use it to reference it daily. And with the name, we can find the UUID if necessary.
Don't you agree?

shwstppr:
@RodrigoDLopez Hi, I don't have any issue in using just the names of host if we have consensus on that.
I liked the idea of having a discussion on ML as that will allow forming a convention project wide

ravening:
> @GabrielBrascher can we use something like `host-name(ID: host_uuid)`? Personally I found it easier at times to debug issues with uuid in logs
> 
> @blueorangutan package

But this will display uuid to users also right which doesnt make sense to them

sureshanaparti:
> > @GabrielBrascher can we use something like `host-name(ID: host_uuid)`? Personally I found it easier at times to debug issues with uuid in logs
> > @blueorangutan package
> 
> But this will display uuid to users also right which doesnt make sense to them

For logging (and debugging), I think it is better to use either ID / UUID (with/without name) as the name is not unique (multiple hosts can have the same name, across the cloudstack env).

rafaelweingartner:
What about adding all of them? I mean, ID, Name, and UUID. This would for sure help debugging. It would have the ID of the DB already there, and the UUID of the element that is used in the API. Moreover, the name to make the log entry more friendly. 

Something such as:
```
s_logger.debug(String.format("Create ClusteredDirectAgentAttache for host [id=%s, uuid=%s, name=%s] ",  host.getId(), host.getUuid(), host.getName()));
```

GabrielBrascher:
@rafaelweingartner @sureshanaparti having ID, UUID, and the name sounds like a good way. Might be a bit verbose, but it serves well all the preferences.

GabrielBrascher:
Applied changes in order to use the `toString()` method from HostVO. These changes consider the proposal from PR #4706 where `toString()` would present a "_JSON like"_ String.

- How it is today: `Host[-123-Routing]` (yep ... nowadays logs that use the toString have `-` before the ID)
- Proposal from #4706: `Host [{id: "123", name: "hostName", uuid: "1e374e50-g231-4345-d6h3-123abcdrfas143", type="Routing"}]`

The idea is that it would be good for:

1. Debug via DB filtering with the host ID
2. Run commands on API/cloudmonkey by copying the uuid presented at the log
3. Identify the host quickly as name makes logs human-friendly
4. Name also tends to be resolved to the host address; therefore, admins can jump into the host (SSH) by copying the name from the log.

Ping for discussion/review: @rafaelweingartner @sureshanaparti @shwstppr @RodrigoDLopez @DaanHoogland @rhtyd @wido @ustcweizhou @svenvogel @andrijapanicsb and others ...

weizhouapache:
@GabrielBrascher it would be good to add host name in the output.

ip is not very important as we normally ssh into server using hostname, not ip.

```
diff --git a/engine/schema/src/main/java/com/cloud/host/HostVO.java b/engine/schema/src/main/java/com/cloud/host/HostVO.java
index 1c3d7c4267..a78b940142 100644
--- a/engine/schema/src/main/java/com/cloud/host/HostVO.java
+++ b/engine/schema/src/main/java/com/cloud/host/HostVO.java
@@ -677,7 +677,7 @@ public class HostVO implements Host {

     @Override
     public String toString() {
-        return new StringBuilder("Host[").append("-").append(id).append("-").append(type).append("]").toString();
+        return new StringBuilder("Host[").append(name).append("-").append(id).append("-").append(type).append("]").toString();
     }

     public void setHypervisorType(HypervisorType hypervisorType) {
```

GabrielBrascher:
@weizhouapache Sorry if I was not clear, but PR #4706 is adding the host's name indeed, IP address would not be interesting indeed.

The idea is exactly what you mentioned. We will then log host ID, UUID, and Name (Type will also be logged the host.toString is used for non-routing hosts as well).

This PR also uses the `toString` in order to have a standard Host log. As PR #4706 already proposes an interesting approach, I added it on this proposal as well.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@GutoVeronezi this PR takes into account the Host to string added by you at PR #4706. I would appreciate any feedback on this one?

@shwstppr the changes on this PR, which takes into account #4706 (merged), should fit with your review by adding the Host **Name**, **ID**, as well as its **UUID**. Can you please review this PR with the proposed approach? Thanks!

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -4219,14 +4219,18 @@ private void orchestrateMigrateForScale(final String vmUuid, final long srcHostI
...
         if (fromHost == null) {
-            s_logger.info("Unable to find the host to migrate from: " + srcHostId);
-            throw new CloudRuntimeException("Unable to find the host to migrate from: " + srcHostId);
+            String logMessageUnableToFindHost = String.format("Unable to find host to migrate from %s.", srcHost);
+            s_logger.info(logMessageUnableToFindHost);
+            throw new CloudRuntimeException(logMessageUnableToFindHost);
         }
 
+        Host dstHost = _hostDao.findById(dstHostId);
         if (fromHost.getClusterId().longValue() != dest.getCluster().getId()) {
```
GutoVeronezi, line 4230: @GabrielBrascher As this class has already been touched, what do you think of doing a few refactors?
Like removing the explicit unboxing...

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -4339,13 +4343,13 @@ private void orchestrateMigrateForScale(final String vmUuid, final long srcHostI
                     try {
                         _agentMgr.send(srcHostId, new Commands(cleanup(vm.getInstanceName())), null);
                     } catch (final AgentUnavailableException e) {
-                        s_logger.error("AgentUnavailableException while cleanup on source host: " + srcHostId);
+                        s_logger.error(String.format("AgentUnavailableException while cleanup on source host: %s", srcHost));
```
GutoVeronezi, line 4346: @GabrielBrascher We can pass the exception as param to `s_logger.error` here.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2105,7 +2106,7 @@ private Host createHostAndAgent(final ServerResource resource, final Map<String,
                     host = findHostByGuid(firstCmd.getGuidWithoutResource());
                 }
                 if (host != null && host.getRemoved() == null) { // host already added, no need to add again
-                    s_logger.debug("Found the host " + host.getId() + " by guid: " + firstCmd.getGuid() + ", old host reconnected as new");
+                    s_logger.debug(String.format("Found host [id: %d, name: %s] by guid: %s, old host reconnected as new", host.getId(), host.getName(), firstCmd.getGuid()));
```
GutoVeronezi, line 2109: @GabrielBrascher We can log the entire object here, as it has `toString()` implemented.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2178,7 +2179,7 @@ private Host createHostAndAgentDeferred(final ServerResource resource, final Map
                     // added, no
                     // need to add
                     // again
-                    s_logger.debug("Found the host " + host.getId() + " by guid: " + firstCmd.getGuid() + ", old host reconnected as new");
+                    s_logger.debug(String.format("Found host [id: %d, name: %s] by guid %s, old host reconnected as new", host.getId(), host.getName(), firstCmd.getGuid()));
```
GutoVeronezi, line 2182: @GabrielBrascher We can log the entire object here, as it has toString() implemented.

GutoVeronezi:
@GabrielBrascher I noticed that in some changes you added `String.format` but in anothers no, is there any reason to it? Can we use `String.format` in all of them?

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
GabrielBrascher: Done!

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
GabrielBrascher: @GutoVeronezi that's a good option indeed, thanks for the attention. I've just updated it.

GabrielBrascher:
Thanks for the review, @GutoVeronezi! Updated code in order to give prefference for `String.format`; additionally,

GabrielBrascher:
THanks for the review @RodrigoDLopez @rafaelweingartner @shwstppr @weizhouapache @sureshanaparti; is there anything missing on the PR?

 All the changes aplied recently were made according to the review from @GutoVeronezi.

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -4339,13 +4345,13 @@ private void orchestrateMigrateForScale(final String vmUuid, final long srcHostI
                     try {
                         _agentMgr.send(srcHostId, new Commands(cleanup(vm.getInstanceName())), null);
                     } catch (final AgentUnavailableException e) {
-                        s_logger.error("AgentUnavailableException while cleanup on source host: " + srcHostId);
+                        s_logger.error(String.format("AgentUnavailableException while cleanup on source %s: %s", srcHost), e);
```
GutoVeronezi, line 4348: There are 2 params passed to format, but only one value.

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -4354,8 +4360,8 @@ private void orchestrateMigrateForScale(final String vmUuid, final long srcHostI
                 s_logger.info("Migration was unsuccessful.  Cleaning up: " + vm);
 
                 _alertMgr.sendAlert(alertType, fromHost.getDataCenterId(), fromHost.getPodId(),
-                        "Unable to migrate vm " + vm.getInstanceName() + " from host " + fromHost.getName() + " in zone " + dest.getDataCenter().getName() + " and pod " +
-                                dest.getPod().getName(), "Migrate Command failed.  Please check logs.");
+                        "Unable to migrate vm " + vm.getInstanceName() + " from " + fromHost + " in zone " + dest.getDataCenter().getName() + " and pod " +
```
GutoVeronezi, line 4363: We can use `String.format` here too.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2385,7 +2387,7 @@ public void deleteRoutingHost(final HostVO host, final boolean isForced, final b
                     try {
                         _vmMgr.destroy(vm.getUuid(), false);
                     } catch (final Exception e) {
-                        final String errorMsg = "There was an error Destory the vm: " + vm + " as a part of hostDelete id=" + host.getId();
+                        final String errorMsg = String.format("There was an error when destroying the vm: %s as a part of hostDelete: %s", vm, host);
```
GutoVeronezi, line 2390: Lowercase `vm`.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2400,16 +2402,17 @@ public void deleteRoutingHost(final HostVO host, final boolean isForced, final b
                     // Restart HA enabled vms
                     for (final VMInstanceVO vm : vms) {
                         if (!vm.isHaEnabled() || vm.getState() == State.Stopping) {
-                            s_logger.debug("Stopping vm: " + vm + " as a part of deleteHost id=" + host.getId());
+                            s_logger.debug("Stopping vm: " + vm + " as a part of hostDelete: " + host);
```
GutoVeronezi, line 2405: Lowercase `vm`.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2400,16 +2402,17 @@ public void deleteRoutingHost(final HostVO host, final boolean isForced, final b
...
                             try {
                                 _vmMgr.advanceStop(vm.getUuid(), false);
                             } catch (final Exception e) {
-                                final String errorMsg = "There was an error stopping the vm: " + vm + " as a part of hostDelete id=" + host.getId();
+                                final String errorMsg = String.format("There was an error stopping the vm: %s as a part of hostDelete: %s", vm, host);
```
GutoVeronezi, line 2409: Lowercase `vm`.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -903,7 +904,8 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
                 try {
                     resourceStateTransitTo(host, ResourceState.Event.DeleteHost, _nodeId);
                 } catch (final NoTransitionException e) {
-                    s_logger.debug("Cannot transmit host " + host.getId() + " to Enabled state", e);
+                    s_logger.debug("Cannot transit " + host + " to Enabled state", e);
+                    s_logger.debug(String.format("Cannot transit %s to Enabled state", host), e);
```
GutoVeronezi, line 908: The same message twice.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -1498,7 +1500,7 @@ public boolean checkAndMaintain(final long hostId) {
                 hostInMaintenance = attemptMaintain(host);
             }
         } catch (final NoTransitionException e) {
-            s_logger.debug("Cannot transmit host " + host.getId() + " to Maintenance state", e);
+            s_logger.debug(String.format("Cannot transit %s to Maintenance state", host), e);
```
GutoVeronezi, line 1503: Question / Topic for discussion:

Could exceptions be logged in `debug` instead of `warn` or `error`?

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2400,16 +2402,17 @@ public void deleteRoutingHost(final HostVO host, final boolean isForced, final b
...
                             } catch (final Exception e) {
-                                final String errorMsg = "There was an error stopping the vm: " + vm + " as a part of hostDelete id=" + host.getId();
+                                final String errorMsg = String.format("There was an error stopping the vm: %s as a part of hostDelete: %s", vm, host);
                                 s_logger.debug(errorMsg, e);
                                 throw new UnableDeleteHostException(errorMsg + "," + e.getMessage());
                             }
                         } else if (vm.isHaEnabled() && (vm.getState() == State.Running || vm.getState() == State.Starting)) {
-                            s_logger.debug("Scheduling restart for vm: " + vm + " " + vm.getState() + " on the host id=" + host.getId());
+                            s_logger.debug("Scheduling restart for vm: " + vm + " " + vm.getState() + " on host: " + host);
+                            s_logger.debug(String.format("Scheduling restart for vm: %s, state: %s on host: %s.", vm, vm.getState(), host));
```
GutoVeronezi, line 2415: The same message twice;
Lowercase `vm`;

server/src/main/java/com/cloud/resource/RollingMaintenanceManagerImpl.java
```diff
@@ -537,7 +539,7 @@ private boolean isMaintenanceScriptDefinedOnHost(Host host, List<HostSkipped> ho
                 answer = agentManager.send(host.getId(), cmd);
             } catch (AgentUnavailableException | OperationTimedoutException e) {
                 // Agent may be restarted on the scripts - continue polling until it is up
-                String msg = "Cannot send command to host: " + host.getId() + ", waiting " + pingInterval + "ms - " + e.getMessage();
+                String msg = String.format("Cannot send command to %s, waiting %sms - %s", host, pingInterval, e.getMessage());
                 s_logger.warn(msg);
```
GutoVeronezi, line 543: Pass exception as parameter.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -5165,7 +5165,7 @@ public void doInTransactionWithoutResult(TransactionStatus status) {
                     }
                 });
             } catch (Exception e) {
-                s_logger.warn("Unable to update vm disk statistics for vm: " + userVm.getId() + " from host: " + hostId, e);
+                s_logger.warn(String.format("Unable to update vm disk statistics for vm %s from %s", userVm.getInstanceName(), host), e);
```
GutoVeronezi, line 5168: Lowercase `vm`.

GutoVeronezi, line 5168:
```suggestion
                s_logger.warn(String.format("Unable to update VM disk statistics for vm %s from %s", userVm.getInstanceName(), host), e);
```

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
GutoVeronezi, line 908:
```suggestion
                    s_logger.debug(String.format("Cannot transit %s to Enabled state", host), e);
```

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
GabrielBrascher: I think that `debug` is not ideal indeed at that context, `warn` looks like a good option. `Error` might be good, but I think that it can bring unnecessary alarms for operators.

I think that the current resource state would also be nice to add on the log. I will update it to something as:
```
s_logger.warn(String.format("Cannot transit %s from %s to Maintenance state", host, host.getResourceState()), e);
```

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

DaanHoogland:
travis results are a real mess @GabrielBrascher , can you have a look?

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@DaanHoogland fixed, travis looks good now, despite one job that keeps reaching maximum logs.

rhtyd:
@GabrielBrascher can you fix the conflict?

GabrielBrascher:
@rhtyd thanks for the ping. I've fixed the conflicts.

sureshanaparti:
@blueorangutan test matrix

GabrielBrascher:
@blueorangutan package

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2481,7 +2482,7 @@ public void deleteRoutingHost(final HostVO host, final boolean isForced, final b
                 try {
                     _vmMgr.destroy(vm.getUuid(), false);
                 } catch (final Exception e) {
-                    final String errorMsg = "There was an error Destory the vm: " + vm + " as a part of hostDelete id=" + host.getId();
+                    String errorMsg = String.format("There was an error when destroying the VM: %s as a part of hostDelete: %s", vm, host);
```
GutoVeronezi, line 2485: Could you to take a look into the VM's `toString` implementation?

https://github.com/apache/cloudstack/blob/7835c0812062381c7e5dd5ab9d719297cd3187dc/engine/schema/src/main/java/com/cloud/vm/VMInstanceVO.java#L508-L510

Seems like there will be duplicated words:
`There was an error when destroying the VM: VM Instance...`

It would apply to every time we log the VM.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2607,13 +2608,13 @@ protected void connectAndRestartAgentOnHost(HostVO host, String username, String
         final com.trilead.ssh2.Connection connection = SSHCmdHelper.acquireAuthorizedConnection(
                 host.getPrivateIpAddress(), 22, username, password);
         if (connection == null) {
-            throw new CloudRuntimeException("SSH to agent is enabled, but failed to connect to host: " + host.getPrivateIpAddress());
+            throw new CloudRuntimeException(String.format("SSH to agent is enabled, but failed to connect to %s via IP address %s.", host, host.getPrivateIpAddress()));
```
GutoVeronezi, line 2611: I would rather use `[]` as var delimiter in some cases; IMHO, it is easier to recognize what is a variable and what is fixed in the final message, e.g.:   
`SSH to agent is enabled, but failed to connect to %s via IP address [%s].`.

server/src/main/java/com/cloud/resource/RollingMaintenanceManagerImpl.java
```diff
@@ -439,7 +439,7 @@ private void enableClusterIfDisabled(Cluster cluster, Set<Long> disabledClusters
      */
     private boolean isMaintenanceStageAvoided(Host host, Map<Long, String> hostsToAvoidMaintenance, List<HostSkipped> hostsSkipped) {
         if (hostsToAvoidMaintenance.containsKey(host.getId())) {
-            s_logger.debug("Host " + host.getId() + " is not being put into maintenance, skipping it");
+            s_logger.debug(String.format("%s is not being put into maintenance, skipping it", host));
```
GutoVeronezi, line 442: I think this message could be more clear, e.g.: 
`%s is in avoid maintenance list [%s], skipping its maintenance...`

server/src/main/java/com/cloud/resource/RollingMaintenanceManagerImpl.java
```diff
@@ -571,7 +571,7 @@ private void performPreFlightChecks(List<Host> hosts, int timeout, String payloa
         for (Host host : hosts) {
             Ternary<Boolean, String, Boolean> result = performStageOnHost(host, Stage.PreFlight, timeout, payload, forced);
             if (result.third() && !hostsToAvoidMaintenance.containsKey(host.getId())) {
-                s_logger.debug("Host " + host.getId() + " added to the avoid maintenance set");
+                s_logger.debug(String.format("%s added to the avoid maintenance set", host));
                 hostsToAvoidMaintenance.put(host.getId(), "Pre-flight stage set to avoid maintenance");
             }
```
GutoVeronezi, line 576: This code seems repeated, but changing parameters; We could extract it to a method.

GabrielBrascher:
Thanks for the review, @GutoVeronezi! I've updated the code addressing your comments.

server/src/main/java/com/cloud/resource/RollingMaintenanceManagerImpl.java
GabrielBrascher: Done!

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

server/src/main/java/com/cloud/capacity/CapacityManagerImpl.java
```diff
@@ -926,8 +926,11 @@ public boolean postStateTransitionEvent(StateMachine2.Transition<State, Event> t
       State oldState = transition.getCurrentState();
       State newState = transition.getToState();
       Event event = transition.getEvent();
-      s_logger.debug("VM state transitted from :" + oldState + " to " + newState + " with event: " + event + "vm's original host id: " + vm.getLastHostId() +
-              " new host id: " + vm.getHostId() + " host id before state transition: " + oldHostId);
+      Host lastHost = _hostDao.findById(vm.getLastHostId());
+      Host oldHost = _hostDao.findById(oldHostId);
+      Host newHost = _hostDao.findById(vm.getHostId());
+      s_logger.debug(String.format("VM state transited from [%s] to [%s] with event [%s]. VM's original host: %s, new host: %s, host before state transition: %s", oldState,
```
GutoVeronezi, line 932: We could log VM's info (id, name...) here.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2451,11 +2452,11 @@ public HostVO fillRoutingHostVO(final HostVO host, final StartupRoutingCommand s
     @Override
     public void deleteRoutingHost(final HostVO host, final boolean isForced, final boolean forceDestroyStorage) throws UnableDeleteHostException {
         if (host.getType() != Host.Type.Routing) {
-            throw new CloudRuntimeException("Non-Routing host gets in deleteRoutingHost, id is " + host.getId());
+            throw new CloudRuntimeException(String.format("Non-Routing host gets in deleteRoutingHost, id is %s", host.getId()));
         }
 
         if (s_logger.isDebugEnabled()) {
-            s_logger.debug("Deleting Host: " + host.getId() + " Guid:" + host.getGuid());
+            s_logger.debug(String.format("Deleting Host: %s", host));
```
GutoVeronezi, line 2459: Similar to VM's `toString` would occur here.

GutoVeronezi (approved):
Nice job @GabrielBrascher.

CLGTM, I raised one last point about `toString`.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
GabrielBrascher: Updated

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan test

nvazquez (approved):
LGTM

nvazquez:
Merging based on approvals and test results

Issue #4610, opened 2021-01-22
[UI] Cannot remove tags on host and primary storage


slavkap:
If there is only one tag for host/primary storage it's not possible to remove it. The workaround is to replace the tag with a comma ",". In the old UI, this functionality works correctly

API call from old UI is:

> /client/api?command=updateStoragePool&id=c9b09777-7340-3752-8a96-f3835f9f5c53&tags=&capacitybytes=273804165120&name=test2&response=json&_=1611312734947

API call from new UI the tag value with an empty string was not get:

> /client/api/?id=c9b09777-7340-3752-8a96-f3835f9f5c53&name=test2&command=updateStoragePool&response=json 

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
master
~~~

##### EXPECTED RESULTS

~~~
The tag should be removed
~~~

##### ACTUAL RESULTS

~~~
The tag is still set on host/primary storage
~~~

rhtyd:
Fixed in https://github.com/apache/cloudstack/pull/4611, can you test latest 4.15 and close @slavkap

slavkap:
thanks, @rhtyd, it works correctly

PR #4611, opened 2021-01-22
UI Storage Pool Tags: unable to delete last tag

Pearl1594:
### Description

This PR attempts to fix https://github.com/apache/cloudstack/issues/4610 - wherein the last tag added to a host/storage pool doesn't get deleted

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### DIFF
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -957,8 +957,8 @@ export default {
             if (param.name !== key) {
               continue
             }
-            if (input === undefined || input === null || (input === '' &&
-              !['updateStoragePool', 'updateHost'].includes(action.api))) {
+            if (!input === undefined || input === null ||
+              (input === '' && !['updateStoragePool', 'updateHost', 'updatePhysicalNetwork'].includes(action.api))) {
               if (param.type === 'boolean') {
                 params[key] = false
               }
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -957,7 +957,8 @@ export default {
             if (param.name !== key) {
               continue
             }
-            if (input === undefined || input === null || input === '') {
+            if (input === undefined || input === null || (input === '' &&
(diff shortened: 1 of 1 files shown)


ui/src/views/AutogenView.vue
```diff
@@ -957,7 +957,8 @@ export default {
             if (param.name !== key) {
               continue
             }
-            if (input === undefined || input === null || input === '') {
+            if (input === undefined || input === null || (input === '' &&
+              !['updateStoragePool', 'updateHost'].includes(action.api))) {
```
harikrishna-patnala, line 961: @Pearl1594 we have physical networks tags also, I just checked we could not delete the tag on physical network too, can you please add that here.

ui/src/views/AutogenView.vue
```diff
@@ -957,7 +957,8 @@ export default {
             if (param.name !== key) {
               continue
             }
-            if (input === undefined || input === null || input === '') {
+            if (input === undefined || input === null || (input === '' &&
```
utchoang, line 960: @Pearl1594 In my opinion it will be cleaner like this
`if (!input || (input === '' && !['updateStoragePool', 'updateHost'].includes(action.api)))` {

Pearl1594: @utchoang doing so would lead to the second condition not being evaluated when the input param is empty('') and the api is one among - update physicalnetwork/storagepool/host as the first condition evaluates to true - because (!input) = string not being undefined, null or empty.

harikrishna-patnala (approved):
Thanks @Pearl1594 , looks good to me now

ui/src/views/AutogenView.vue
utchoang: I understood! Thanks @Pearl1594

utchoang (approved):

PR #4630, opened 2021-01-29
disable hot add memory and cpu via vm settings

DK101010:
### Description

Currently hot add memory and cpu is always enabled when it supported. In some situation it is necessary to disable that features for a specific vm.

With this PR User can disable hot add memory and cpu via vm settings in the ui or via api call. the default is still enabled, therefore it should not break existing behavior.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
Manuelly tested in vmware env.

### DIFF
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
@@ -57,6 +57,7 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import org.apache.commons.lang.StringUtils;
 
 class VmwareVmImplementer {
     private static final Logger LOGGER = Logger.getLogger(VmwareVmImplementer.class);
@@ -138,6 +139,12 @@ class VmwareVmImplementer {
                     LOGGER.warn("Invalid NIC device type " + nicDeviceType + " is specified in VM details, switch to default E1000");
                     details.put(VmDetailConstants.NIC_ADAPTER, VirtualEthernetCardType.E1000.toString());
                 }
+            }
+            if(StringUtils.isEmpty(details.get(VmDetailConstants.HOT_ADD_CPU))){
+                details.put(VmDetailConstants.HOT_ADD_CPU, Boolean.TRUE.toString());
+            }
+            if(StringUtils.isEmpty(details.get(VmDetailConstants.HOT_ADD_MEMORY))){
+                details.put(VmDetailConstants.HOT_ADD_MEMORY, Boolean.TRUE.toString());
             }
         }
 
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
@@ -40,7 +40,6 @@ import com.cloud.template.VirtualMachineTemplate;
 import com.cloud.utils.exception.CloudRuntimeException;
 import com.cloud.vm.DomainRouterVO;
 import com.cloud.vm.NicProfile;
-import com.cloud.vm.VMInstanceVO;
 import com.cloud.vm.VirtualMachine;
 import com.cloud.vm.VirtualMachineProfile;
 import com.cloud.vm.VmDetailConstants;
@@ -139,10 +138,6 @@ class VmwareVmImplementer {
                     LOGGER.warn("Invalid NIC device type " + nicDeviceType + " is specified in VM details, switch to default E1000");
                     details.put(VmDetailConstants.NIC_ADAPTER, VirtualEthernetCardType.E1000.toString());
                 }
-            }
-            if(vm.getVirtualMachine() instanceof VMInstanceVO){
-                VMInstanceVO vmInstanceVO =(VMInstanceVO) vm.getVirtualMachine();
-                to.setEnableDynamicallyScaleVm(vmInstanceVO.isDynamicallyScalable());
             }
         }
 
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
@@ -40,6 +40,7 @@ import com.cloud.template.VirtualMachineTemplate;
 import com.cloud.utils.exception.CloudRuntimeException;
 import com.cloud.vm.DomainRouterVO;
 import com.cloud.vm.NicProfile;
+import com.cloud.vm.VMInstanceVO;
 import com.cloud.vm.VirtualMachine;
 import com.cloud.vm.VirtualMachineProfile;
 import com.cloud.vm.VmDetailConstants;
@@ -57,7 +58,6 @@ import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import org.apache.commons.lang.StringUtils;
 
 class VmwareVmImplementer {
     private static final Logger LOGGER = Logger.getLogger(VmwareVmImplementer.class);
@@ -140,11 +140,9 @@ class VmwareVmImplementer {
                     details.put(VmDetailConstants.NIC_ADAPTER, VirtualEthernetCardType.E1000.toString());
                 }
             }
-            if(StringUtils.isEmpty(details.get(VmDetailConstants.HOT_ADD_CPU))){
-                details.put(VmDetailConstants.HOT_ADD_CPU, Boolean.TRUE.toString());
-            }
-            if(StringUtils.isEmpty(details.get(VmDetailConstants.HOT_ADD_MEMORY))){
-                details.put(VmDetailConstants.HOT_ADD_MEMORY, Boolean.TRUE.toString());
+            if(vm.getVirtualMachine() instanceof VMInstanceVO){
+                VMInstanceVO vmInstanceVO =(VMInstanceVO) vm.getVirtualMachine();
+                to.setEnableDynamicallyScaleVm(vmInstanceVO.isDynamicallyScalable());
             }
         }
 
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -1970,7 +1970,11 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
             }
 
             if(!vmMo.isMemoryHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm()){
-                s_logger.warn("hotadd is not supported, dynamic scaling feature can not be applied " + vmInternalCSName);
+                s_logger.warn("hotadd of memory is not supported, dynamic scaling feature can not be applied " + vmInternalCSName);
+            }
+
+            if(!vmMo.isCpuHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm()){
+                s_logger.warn("hotadd of cpu is not supported, dynamic scaling feature can not be applied " + vmInternalCSName);
             }
 
             configNestedHVSupport(vmMo, vmSpec, vmConfigSpec);
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -1969,6 +1969,10 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                 vmConfigSpec.setCpuHotAddEnabled(vmMo.isCpuHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm());
             }
 
+            if(!vmMo.isMemoryHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm()){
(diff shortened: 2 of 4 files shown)


weizhouapache:
code lgtm

DaanHoogland:
@blueorangutan package

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1959,15 +1959,15 @@ protected StartAnswer execute(StartCommand cmd) {
             }
 
             // Check for hotadd settings
-            vmConfigSpec.setMemoryHotAddEnabled(vmMo.isMemoryHotAddSupported(guestOsId));
+            vmConfigSpec.setMemoryHotAddEnabled(vmMo.isMemoryHotAddSupported(guestOsId) && Boolean.parseBoolean(vmSpec.getDetails().get(VmDetailConstants.HOT_ADD_MEMORY)));
```
DaanHoogland, line 1962: wouldn't this mean you can turn it off but never back on again?

DK101010: Nope, you can switch on/off like you want.

harikrishna-patnala:
@DK101010 we already have a flag per VM to enable or disable dynamic scalability of resources on the VM.
image
May I know why we need these settings again ?

Also keeping two different flags for CPU and memory may conflict during actual VM deployment (with different combinations of true and false) since we are controlling the dynamic scalability of VM with only one flag at global/zone setting.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
DaanHoogland: @DK101010, I probably don't understand but the `vmMo` is found on the hypervisor with `vmMo = hyperHost.findVmOnHyperHost(vmInternalCSName);`, if it was already created. And if on creation `memoryHotAddEnabled` was set to false, the above will always be false on the next start, and hence all subsequent. Maybe this is not in the functional intent of this PR but it seems when you turn it off for a VM it will not be turned on again if the user switches it on again.

DK101010: Hi @DaanHoogland, At first I thought it also. But this method will be called each time when you start a vm and each time will be update the vm config spec. Btw. you can find in VmwareHelper in method setBasicVmConfig similar logic to enable/disable hot add but it will be override from this method. I find it also a little bit strange and perhaps it have refactoring potential. But currently I have not so much time and knowledge to do it. Independent from this, I tested the code to enable and disable hot add for cpu and memory and it works like expected. I checked it also in VCenter if it enabled/disabled.

DaanHoogland (approved):

DK101010:
> @DK101010 we already have a flag per VM to enable or disable dynamic scalability of resources on the VM.
> image
> May I know why we need these settings again ?
> 
> Also keeping two different flags for CPU and memory may conflict during actual VM deployment (with different combinations of true and false) since we are controlling the dynamic scalability of VM with only one flag at global/zone setting.

Hi @harikrishna-patnala, until now I don't know this feature. I had checked this, but how I can disable/enable hot add ? I can set dynamic scalability true/false but hot add cpu and memory keeps enabled.

DK101010:
@harikrishna-patnala @DaanHoogland Here is a alternative implementation to use the dynamic scalability flag. What do you think?

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1959,15 +1959,14 @@ protected StartAnswer execute(StartCommand cmd) {
             }
 
             // Check for hotadd settings
-            vmConfigSpec.setMemoryHotAddEnabled(vmMo.isMemoryHotAddSupported(guestOsId));
-
+            vmConfigSpec.setMemoryHotAddEnabled(vmMo.isMemoryHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm());
```
harikrishna-patnala, line 1962: Thanks @DK101010, this seems to be good.
I'm also thinking, if memory hot add memory and CPU are not supported by OS and vmSpec wants the VM to be dynamically scalable, may be we should atleast log it saying this is not supported by OS. Can you please do that here.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
```diff
@@ -139,6 +140,10 @@ VirtualMachineTO implement(VirtualMachineProfile vm, VirtualMachineTO to, long c
                     details.put(VmDetailConstants.NIC_ADAPTER, VirtualEthernetCardType.E1000.toString());
                 }
             }
+            if(vm.getVirtualMachine() instanceof VMInstanceVO){
+                VMInstanceVO vmInstanceVO =(VMInstanceVO) vm.getVirtualMachine();
+                to.setEnableDynamicallyScaleVm(vmInstanceVO.isDynamicallyScalable());
```
harikrishna-patnala, line 145: HypervisorGuruBase is already setting this paramter in toVirtualMachineTO(), can you please double check if this is necessary or redundant in VmwareVMImplementer.
        to.setEnableDynamicallyScaleVm(isDynamicallyScalable);

DK101010: > HypervisorGuruBase is already setting this paramter in toVirtualMachineTO(), can you please double check if this is necessary or redundant in VmwareVMImplementer.
> to.setEnableDynamicallyScaleVm(isDynamicallyScalable);

@harikrishna-patnala Hmm ... during my test I could enable/disable the flag in the fronend but in backend it keeps of false. That is the reason for my implementation in VmwareVmImplementer.java 

I have checked the HypervisorGuru and found follow line 
`Boolean isDynamicallyScalable = vmInstance.isDynamicallyScalable() && UserVmManager.EnableDynamicallyScaleVm.valueIn(vm.getDataCenterId());`

I think I understood now what do you mean with zone settings. ;) But I ask me why we need two flags for the same thing. In my opinion it is confusing and not handy for a user to enable two flags to use this feature.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
DK101010: @harikrishna-patnala sure, shouldn't be a problem. I will add this

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VmwareVmImplementer.java
DaanHoogland: bit of background guys: The VmwareVMImplementer is a worker class that is part of the VmwareGuru, and is meant to reduce the complexity by extracting the deploy and start code. It would be good to put shared code in a VmwareGuruUtilities class to prevent duplication. I can imagine that setting flags can be done on implement as well as on restart/migrate or other methods. I might be guilty of this redundancy, but take care that setting might happen twice in one scenario but twice in another ...

harikrishna-patnala: Thanks @DaanHoogland for explaining that.
Here in this case setting isDynamicallyScalable on VM has no dependency on hypervisor type that is why HypervisorGuruBase sets the dynamic scaling flag on VM while preparing TO (transfer object). Setting this flag in either VMwareGuru or VMwareVMImplementer is not required.

@DK101010 I would suggest you to please revert the change HypervisorGuruBase in which you ignored the global/zone setting. This is required because global/zone level setting actually decides whether dynamic scaling can be enabled in that management setup or not. vmInstance.isDynamicallyScalable() is not sufficient. There is another PR#4643 which fixes and all these settings. For this PR you can keep the VMwareResource changes.

DK101010: @DaanHoogland Thanks for input. Good to know for the future.

@harikrishna-patnala sure, i can revert this, but I still don't find it practical to enable 2 flags to turn a feature on ;).

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1959,15 +1959,18 @@ protected StartAnswer execute(StartCommand cmd) {
...
                 s_logger.warn("Dynamic scaling of CPU is not supported for Virtual Machines with multi-core vCPUs in case of ESXi hosts 4.1 and prior. Hence CpuHotAdd will not be"
                         + " enabled for Virtual Machine: " + vmInternalCSName);
                 vmConfigSpec.setCpuHotAddEnabled(false);
             } else {
-                vmConfigSpec.setCpuHotAddEnabled(vmMo.isCpuHotAddSupported(guestOsId));
+                vmConfigSpec.setCpuHotAddEnabled(vmMo.isCpuHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm());
+            }
+
+            if(!vmMo.isMemoryHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm()){
+                s_logger.warn("hotadd is not supported, dynamic scaling feature can not be applied " + vmInternalCSName);
```
harikrishna-patnala, line 1973: Thanks @DK101010 for making the changes. can you please add the reason being guest OS does not support hot add, something like "hotadd of memory is not supported by the guest OS, dynamic scaling feature can not be applied " + vmInternalCSName)".

and please add another log for CPU hot add.

DK101010: @harikrishna-patnala no problem, I will adapt it.

harikrishna-patnala (approved):
LGTM

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1959,15 +1959,22 @@ protected StartAnswer execute(StartCommand cmd) {
...
                 s_logger.warn("Dynamic scaling of CPU is not supported for Virtual Machines with multi-core vCPUs in case of ESXi hosts 4.1 and prior. Hence CpuHotAdd will not be"
                         + " enabled for Virtual Machine: " + vmInternalCSName);
                 vmConfigSpec.setCpuHotAddEnabled(false);
             } else {
-                vmConfigSpec.setCpuHotAddEnabled(vmMo.isCpuHotAddSupported(guestOsId));
+                vmConfigSpec.setCpuHotAddEnabled(vmMo.isCpuHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm());
+            }
+
+            if(!vmMo.isMemoryHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm()){
+                s_logger.warn("hotadd of memory is not supported, dynamic scaling feature can not be applied " + vmInternalCSName);
```
sureshanaparti, line 1973:
```suggestion
                s_logger.warn("hotadd of memory is not supported, dynamic scaling feature can not be applied to vm: " + vmInternalCSName);
```

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1959,15 +1959,22 @@ protected StartAnswer execute(StartCommand cmd) {
...
+            }
+
+            if(!vmMo.isCpuHotAddSupported(guestOsId) && vmSpec.isEnableDynamicallyScaleVm()){
+                s_logger.warn("hotadd of cpu is not supported, dynamic scaling feature can not be applied " + vmInternalCSName);
```
sureshanaparti, line 1977:
```suggestion
                s_logger.warn("hotadd of cpu is not supported, dynamic scaling feature can not be applied to vm: " + vmInternalCSName);
```

sureshanaparti (approved):
clgtm

DaanHoogland:
@borisstoyanov @vladimirpetrov any extra tests required? cc @sureshanaparti @nvazquez

rhtyd:
LGTM, need some testing/confirmation

nvazquez:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

rhtyd:
@blueorangutan centos7 vmware-67u3

rhtyd:
@blueorangutan centos7 vmware-67u3

rhtyd:
@blueorangutan help

rhtyd:

nvazquez (approved):
Code LGTM

nvazquez:
@blueorangutan package

nvazquez:

nvazquez (approved):
Tested on Vmware 6.5u2:

- Hot add unsupported guest OS (Other Linux 64 bits) -> CPU and Memory hot add flags always disabled irrespective of dynamic calling values
- Hot add supported guest OS (Debian 10 64 bits) -> CPU and memory hot add flags enabled/disabled on vCenter accordingly

Note: support compatibility checked on: https://www.vmware.com/resources/compatibility/detail.php?deviceCategory=software&testConfig=16&productid=48848&supRel=408,&deviceCategory=software&details=1&releases=408&operatingSystems=260&page=1&display_interval=10&sortColumn=Partner&sortOrder=Asc&testConfig=16

PR #4429, opened 2020-10-27
FIX s2svpn connection stuck on Pending state

ravening:
## Description

When we try to reset the site 2 site vpn connection while
the VR's are being restarted, the connection enters the
PENDING state and we cant reset the connection.

So change the state from PENDING to disconnected.

Steps to reproduce the issue

1.create a VPC with a tier (vpc-001-001 in vpc-001), create a vm
2.create a VPC with a tier (vpc-002-001 in vpc-002) with different cidr, create a vm
3.create custom gateway for both vpn
4.enable site-to-site vpn on both vpn, and add vpn connection to each other. both should be "Connected"
5.restart vpc-001 with cleanup and monitor it
6.when the first router is destroyed, go to site-to-site vpn page and reset vpn connection.
7.we will get an error "Resource [DataCenter:1] is unreachable: Unable to apply site 2 site VPN configuration, virtual router is not in the right state"
and vpn connection is stuck at Pending
8.When vpc is restarted, go to site-to-site vpn page and reset vpn connection.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

PR #4635, opened 2021-02-01
Persist vpn connection state before restarting

ravening:
### Description
If the vpn connection is in pending state then we cant restart the
vpn connection. So manually set the state to disconnected and then
try to restart the vpn connection

This pr is needed along with https://github.com/apache/cloudstack/pull/4429/files

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/network/vpn/Site2SiteVpnManagerImpl.java
+++ b/server/src/main/java/com/cloud/network/vpn/Site2SiteVpnManagerImpl.java
@@ -610,13 +610,12 @@ public class Site2SiteVpnManagerImpl extends ManagerBase implements Site2SiteVpn
         }
         _accountMgr.checkAccess(caller, null, false, conn);
 
-        if (conn.getState() == State.Pending) {
-            conn.setState(State.Disconnected);
-        }
-        if (conn.getState() == State.Connected || conn.getState() == State.Error
-            || conn.getState() == State.Disconnected || conn.getState() == State.Connecting) {
-            stopVpnConnection(id);
-        }
+        // Set vpn state to disconnected
+        conn.setState(State.Disconnected);
+        _vpnConnectionDao.persist(conn);
+
+        // Stop and start the connection again
+        stopVpnConnection(id);
         startVpnConnection(id);
         conn = _vpnConnectionDao.findById(id);
         return conn;


DaanHoogland:
@ravening did you try this while state == pending or while state == disconnected? it seems a bit brute force to me.
Also #4429 is on 4.14, would you rather point this to 4.14 as well?

ravening:
> @ravening did you try this while state == pending or while state == disconnected? it seems a bit brute force to me.
> 
> Also #4429 is on 4.14, would you rather point this to 4.14 as well?

@DaanHoogland this was when state is pending. We need to persist state in DB so that the connection is restarted. If we don't persist then check fails and connection is not restarted

The other pr is already in 4.16. so this can go-to 4.16 as well

DaanHoogland:
> > @ravening did you try this while state == pending or while state == disconnected? it seems a bit brute force to me.
> > Also #4429 is on 4.14, would you rather point this to 4.14 as well?
> 
> @DaanHoogland this was when state is pending. We need to persist state in DB so that the connection is restarted. If we don't persist then check fails and connection is not restarted
> 
> The other pr is already in 4.16. so this can go-to 4.16 as well

the other pr is also in 4.14, so if this only goes into 4.16 do we then have half a fix in the LTS versions?

ravening:
> > > @ravening did you try this while state == pending or while state == disconnected? it seems a bit brute force to me.
> > > Also #4429 is on 4.14, would you rather point this to 4.14 as well?
> > 
> > 
> > @DaanHoogland this was when state is pending. We need to persist state in DB so that the connection is restarted. If we don't persist then check fails and connection is not restarted
> > The other pr is already in 4.16. so this can go-to 4.16 as well
> 
> the other pr is also in 4.14, so if this only goes into 4.16 do we then have half a fix in the LTS versions?

good point. i will raise for 4.14 as well

rhtyd (approved):

server/src/main/java/com/cloud/network/vpn/Site2SiteVpnManagerImpl.java
```diff
@@ -561,12 +561,12 @@ public Site2SiteVpnConnection resetVpnConnection(ResetVpnConnectionCmd cmd) thro
...
-        }
-        if (conn.getState() == State.Connected || conn.getState() == State.Error || conn.getState() == State.Disconnected) {
-            stopVpnConnection(id);
-        }
+        // Set vpn state to disconnected
+        conn.setState(State.Disconnected);
+        _vpnConnectionDao.persist(conn);
+
+        // Stop and start the connection again
+        stopVpnConnection(id);
```
rhtyd, line 569: Would without checking for the current conn state cause any issue @ravening ? For example, try to stop vpn connection when VR has errors is powered off etc or in other transitional state?

sureshanaparti (approved):
CLGTM

server/src/main/java/com/cloud/network/vpn/Site2SiteVpnManagerImpl.java
ravening: @rhtyd probably no as this involves updating the state in db

server/src/main/java/com/cloud/network/vpn/Site2SiteVpnManagerImpl.java
```diff
@@ -561,12 +561,12 @@ public Site2SiteVpnConnection resetVpnConnection(ResetVpnConnectionCmd cmd) thro
...
 
-        if (conn.getState() == State.Pending) {
-            conn.setState(State.Disconnected);
-        }
-        if (conn.getState() == State.Connected || conn.getState() == State.Error || conn.getState() == State.Disconnected) {
-            stopVpnConnection(id);
-        }
+        // Set vpn state to disconnected
+        conn.setState(State.Disconnected);
+        _vpnConnectionDao.persist(conn);
```
weizhouapache, line 566: why not just persist DB only when state is pending ?

```
        if (conn.getState() == State.Pending) {
            conn.setState(State.Disconnected);
            _vpnConnectionDao.persist(conn);
        }
```

nvazquez:
Ping @ravening can you address the open comments and fix conflicts? Thanks

rhtyd:
Ping @ravening

server/src/main/java/com/cloud/network/vpn/Site2SiteVpnManagerImpl.java
ravening: @weizhouapache since we are restting the connection, it should transit to disconnect state irrecpective of the previous state. since we are stopping and starting the vpn connection, checking for pending state or any other state doesnt matter

ravening:
> Ping @ravening

@rhtyd done

rhtyd:
@blueorangutan package

rhtyd (approved):

weizhouapache (approved):
code lgtm

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

nvazquez (approved):

PR #4641, opened 2021-02-02
Speed up SSVM startup time by postponing checks

ravening:
### Description

Currently SSVM takes long time to startup.
Reduce the startup time by moving volume/template
check to post connect TimerTask to speed up ssvm startup

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/storage/download/DownloadListener.java
+++ b/server/src/main/java/com/cloud/storage/download/DownloadListener.java
@@ -96,6 +96,28 @@ public class DownloadListener implements Listener {
         }
     }
 
+    private static final class PostConnectTask extends ManagedContextTimerTask {
+        private final DownloadListener dl;
+        private final Host agent;
+        private final StartupCommand cmd;
+
+        private PostConnectTask(DownloadListener dl, Host agent, StartupCommand cmd) {
+            this.dl = dl;
+            this.agent = agent;
+            this.cmd = cmd;
+        }
+
+        @Override
+        protected void runInContext() {
+            try {
+                dl.postConnect(agent, cmd);
+            } catch (ConnectionException ex) {
+                s_logger.warn("There is an error in the postConnect process for " +
+                        agent.getId() + " due to " + ex.getMessage());
+            }
+        }
+    }
+
     public static final Logger s_logger = Logger.getLogger(DownloadListener.class.getName());
     public static final int SMALL_DELAY = 100;
     public static final long STATUS_POLL_INTERVAL = 10000L;
@@ -121,6 +143,7 @@ public class DownloadListener implements Listener {
 
     private StatusTask _statusTask;
     private TimeoutTask _timeoutTask;
+    private PostConnectTask _postConnectTask;
     private Date _lastUpdated = new Date();
     private String jobId;
 
@@ -210,6 +233,7 @@ public class DownloadListener implements Listener {
 
     public DownloadListener(DownloadMonitorImpl monitor) {
         _downloadMonitor = monitor;
+        _timer = new Timer();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/cloud/storage/download/DownloadListener.java
```diff
@@ -284,18 +308,22 @@ public void processConnect(Host agent, StartupCommand cmd, boolean forRebalance)
...
+    private void schedulePostConnectTask(Host agent, StartupCommand cmd, long delay) {
+        if (_postConnectTask != null)
+            _postConnectTask.cancel();
+
+        _postConnectTask = new PostConnectTask(this, agent, cmd);
+        _timer.schedule(_postConnectTask, delay);
+    }
+
+    private void postConnect(Host agent, StartupCommand cmd) throws ConnectionException {
+        if (cmd instanceof StartupSecondaryStorageCommand) {
```
DaanHoogland, line 326: this is only ever called with a `StartupSecondaryStorageCommand` so can we change the parameter type and skip the class check?

server/src/main/java/com/cloud/storage/download/DownloadListener.java
```diff
@@ -284,18 +308,22 @@ public void processConnect(Host agent, StartupCommand cmd, boolean forRebalance)
...
-                downloadMonitor.handleTemplateSync(agent);
-                downloadMonitor.handleVolumeSync(agent);
-            }
-        }*/
-        else if (cmd instanceof StartupSecondaryStorageCommand) {
+    }
+
+    private void schedulePostConnectTask(Host agent, StartupCommand cmd, long delay) {
+        if (_postConnectTask != null)
+            _postConnectTask.cancel();
```
DaanHoogland, line 319: is this only ever working on one dowload job? it seems we are canceling the download check for someone else here.

rhtyd:
I think it may cause some issue if we don't do this, esp in env with very large no. of templates, volumes, snapshots - pl discuss on dev/users@ @ravening

rhtyd:
Ping @ravening

ravening:
@weizhouapache you have any feedback on this pr?

rhtyd:
Pl discuss on dev@, in my opinion this should be closed as postponing check isn't something we would want. The tradeoff of speedup vs check needs to be discussed.

rhtyd:
@weizhouapache I'm just being risk-averse :) I see that checks are only postponed - but I worry if not doing checks at the time of initial connection causes any failures? For example, you allow the agent to connect and it starts doing some storage related operation, and later the post-connect thread/check runs and finds out we shouldn't have connected, or it fails but since the agent is connected, agent continues to process new commands? (a case can be the template/volume scanning logic times out, but the connect is still connected so agent gets some new operation say re-download template from a URL which is wrong/404 etc and it ends up overwriting existing files?)

weizhouapache:
> @weizhouapache I'm just being risk-averse :) I see that checks are only postponed - but I worry if not doing checks at the time of initial connection causes any failures? For example, you allow the agent to connect and it starts doing some storage related operation, and later the post-connect thread/check runs and finds out we shouldn't have connected, or it fails but since the agent is connected, agent continues to process new commands? (a case can be the template/volume scanning logic times out, but the connect is still connected so agent gets some new operation say re-download template from a URL which is wrong/404 etc and it ends up overwriting existing files?)

@rhtyd these are very good questions. 
@ravening are you ok to close this pr ? the issue this pr aims to fix is, SSVM will be Up only if all imags/volumes are checked which I think it is not a major issue.

ravening:
closing it after discussion

PR #4679, opened 2021-02-10
Disable shrinking QCOW2 volumes

ravening:
### Description
Since QCOW2 volumes cannot be resized to lower value in any case,
just disable it in backend.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
@@ -1061,6 +1061,14 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
              * This will be checked again at the hypervisor level where we can see
              * the actual disk size.
              */
+            if (currentSize > newSize) {
+                VolumeVO vol = _volsDao.findById(cmd.getEntityId());
+                if (vol != null && ImageFormat.QCOW2.equals(vol.getFormat()) && !Volume.State.Allocated.equals(volume.getState())) {
+                    String message = "Unable to shrink volumes of type QCOW2";
+                    s_logger.warn(message);
+                    throw new InvalidParameterValueException(message);
+                }
+            }
             if (currentSize > newSize && !shrinkOk) {
                 throw new InvalidParameterValueException("Going from existing size of " + currentSize + " to size of " + newSize + " would shrink the volume."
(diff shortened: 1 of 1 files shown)


weizhouapache (approved):

DaanHoogland (commented):
looks good but can you please put it in a new method with a good descriptive name?

shwstppr:
@ravening I think we can have this in 4.15.1

rhtyd:
ping @ravening ^^

rhtyd:
Ping @ravening

ravening:
> Ping @ravening

@rhtyd done

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd (approved):

slavkap:
Hi all!
Not so sure how to proceed on this. I spoke with @svenvogel and he confirmed that the SolidFire's volumes are kept in the DB in `QCOW2` format, but their actual format is RAW.  According to the code, this will affect the shrinking of So'slidfire volumes. The same is the case with our plugin (StorPool). Is it alright to revert this change until a better solution is proposed or somebody with this version of CS and Solidfire as primary storage to confirm the problem?

shwstppr:
@slavkap is it better to check format directly in `plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtResizeVolumeCommandWrapper.java' and raise an exception if resize isn't possible?

slavkap:
Yes, this sounds OK. Thanks for the answer and the suggestion, @shwstppr!  :)

svenvogel:
@shwstppr @ravening Sound interesting but its breaks our actually functionality with storepool and solidfire storages. how about revert this change until we have another fix for that. at the moment there is no backward functionality.

slavkap:
@shwstppr, do you want to add a fix (because you propose it), or could I work on a fix for this?

shwstppr:
Not sure but will be easier to test for you @slavkap. I'm happy if you can work on it else I can do it

slavkap:
Thanks, @shwstppr. I will do the fix then :)

PR #4691, opened 2021-02-15
server: delete template on storage over capacity threshold

shwstppr:
### Description

While deleting a template for a specific zone, the check should be done only for secondary storages and not for storages with the available capacity threshold.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
+++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
@@ -25,12 +25,6 @@ import java.util.concurrent.ExecutionException;
 
 import javax.inject.Inject;
 
-import com.cloud.configuration.Config;
-import com.cloud.deployasis.dao.TemplateDeployAsIsDetailsDao;
-import com.cloud.storage.dao.VMTemplateDetailsDao;
-import com.cloud.utils.db.Transaction;
-import com.cloud.utils.db.TransactionCallback;
-import com.cloud.utils.db.TransactionStatus;
 import org.apache.cloudstack.agent.directdownload.CheckUrlAnswer;
 import org.apache.cloudstack.agent.directdownload.CheckUrlCommand;
 import org.apache.cloudstack.api.command.user.iso.DeleteIsoCmd;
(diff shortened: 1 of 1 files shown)


shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test

DaanHoogland (approved):
code lgtm

rhtyd (approved):
LGTM, let's wait for @Pearl1594 's LGTM who has worked on the secondary storage management feature

Pearl1594:
@shwstppr atm deletion of templates is supported across all stores, so maybe we can omit the check for writable stores or we disable deletion of templates on read-only stores? We may also need the strictness of this check to be reduced for deletion of Isos?

shwstppr:
@Pearl1594 I've added changes for ISO deletion
I can remove the writable check but there seems another issue. If there is only a single zone with a read-only secondary store and a template is deleted from UI (or with API using zoneid param), entry from cloud.vm_template is not marked removed first. Calling the API again second time with the same parameters deletes the template completely and marks it removed in DB.

@rhtyd @DaanHoogland should we allow deleting templates, ISOs on read-only stores?

rhtyd:
Yes I think so @shwstppr - possible the read only secondary storage is more towards adding new templates than removing existing ones, @Pearl1594 is my assumption correct?

Pearl1594:
That's right @rhtyd. The idea behind read-only stores is to prevent any further resources from being downloaded onto that store.

shwstppr:
made the change to remove writable store check @Pearl1594 @rhtyd 
@blueorangutan package

Pearl1594 (approved):
LGTM

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test

PR #4704, opened 2021-02-18
Improve logs on kvmvmactivity.sh

GutoVeronezi:
### Description
This PR intends to improve logging in the file `kvmvmactivity.sh` to facilitate troubleshooting.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

### DIFF
--- a/scripts/vm/hypervisor/kvm/kvmvmactivity.sh
+++ b/scripts/vm/hypervisor/kvm/kvmvmactivity.sh
@@ -93,7 +93,7 @@ fi
 
 if [ -z "$UUIDList" ]
 then
-  echo "=====> DEAD <======"
+  echo "=====> Considering host as DEAD due to empty UUIDList <======"
   exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


GabrielBrascher (approved):

DaanHoogland (approved):
textual changes only :+1:

scripts/vm/hypervisor/kvm/kvmvmactivity.sh
```diff
@@ -93,7 +93,7 @@ fi
 
 if [ -z "$UUIDList" ]
 then
-  echo "=====> DEAD <======"
+  echo "=====> Considering host as DEAD due to empty UUIDList <======"
```
rhtyd, line 96: You may want to check if we grep/use the stdout of the script anywhere, otherwise this will cause regression.

GutoVeronezi: On KVMHAVMActivityChecker.java we verify if the stdout `contains` `DEAD`. As the keyword `DEAD` still exists in the stdout, it will keep the same behavior.
https://github.com/apache/cloudstack/blob/59fba4916b5c52ca5b3101faaa0f43498201644b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAVMActivityChecker.java#L59

GutoVeronezi:
@rhtyd can you take a look at it?

rhtyd:
4.15.1 is the priority right now for me, I'm not looking into master/4.16 specific PRs now. However, I could have been given chance to reply before merging this @DaanHoogland

rhtyd (commented):
This breaks historic convention used in the kvm specific scripts (we also have a similar kvmheartbeat.sh), but I don't see anything broken; the HA related kvm smoketests seem to have passed so no concerns over this breaking anything (in case we missed anything else, users will eventually report in future).

PR #4707, opened 2021-02-18
Improve logs on HAManagerImpl

GutoVeronezi:
### Description
This PR intends to improve logging in the class `HAManagerImpl` to facilitate troubleshooting.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

### DIFF
--- a/server/src/main/java/org/apache/cloudstack/ha/HAManagerImpl.java
+++ b/server/src/main/java/org/apache/cloudstack/ha/HAManagerImpl.java
@@ -156,9 +156,8 @@ public final class HAManagerImpl extends ManagerBase implements HAManager, Clust
             if (result) {
                 final String message = String.format("Transitioned host HA state from:%s to:%s due to event:%s for the host id:%d",
                         currentHAState, nextState, event, haConfig.getResourceId());
-                if (LOG.isTraceEnabled()) {
-                    LOG.trace(message);
-                }
+                LOG.debug(message);
+
                 if (nextState == HAConfig.HAState.Recovering || nextState == HAConfig.HAState.Fencing || nextState == HAConfig.HAState.Fenced) {
                     ActionEventUtils.onActionEvent(CallContext.current().getCallingUserId(), CallContext.current().getCallingAccountId(),
                             Domain.ROOT_DOMAIN, EventTypes.EVENT_HA_STATE_TRANSITION, message);
@@ -166,9 +165,7 @@ public final class HAManagerImpl extends ManagerBase implements HAManager, Clust
             }
             return result;
         } catch (NoTransitionException e) {
-            if (LOG.isTraceEnabled()) {
-                LOG.trace("Unable to find next HA state for current HA state: " + currentHAState + " for event: " + event + " for host" + haConfig.getResourceId());
-            }
+            LOG.warn(String.format("Unable to find next HA state for current HA state=[%s] for event=[%s] for host=[%s].", currentHAState, event, haConfig.getResourceId()), e);
         }
         return false;
     }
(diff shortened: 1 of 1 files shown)


server/src/main/java/org/apache/cloudstack/ha/HAManagerImpl.java
```diff
@@ -156,19 +156,16 @@ public boolean transitionHAState(final HAConfig.Event event, final HAConfig haCo
             if (result) {
                 final String message = String.format("Transitioned host HA state from:%s to:%s due to event:%s for the host id:%d",
                         currentHAState, nextState, event, haConfig.getResourceId());
-                if (LOG.isTraceEnabled()) {
-                    LOG.trace(message);
-                }
+                LOG.debug(message);
```
DaanHoogland, line 159: why do we need to up this level?
also why remove the isEnabled check? I'd rather move the message creation inside such a check.

server/src/main/java/org/apache/cloudstack/ha/HAManagerImpl.java
```diff
@@ -298,14 +308,10 @@ public Boolean isVMAliveOnHost(final Host host) throws Investigator.UnknownVM {
         final HAConfig haConfig = haConfigDao.findHAResource(host.getId(), HAResource.ResourceType.Host);
         if (haConfig != null) {
             if (haConfig.getState() == HAConfig.HAState.Fenced) {
-                if (LOG.isDebugEnabled()){
-                    LOG.debug("HA: Host is fenced " + host.getId());
-                }
+                LOG.debug(String.format("HA: Host [%s] is fenced.", host.getId()));
```
DaanHoogland, line 311: please put those checks back in

DaanHoogland (changes requested):

server/src/main/java/org/apache/cloudstack/ha/HAManagerImpl.java
GutoVeronezi: @DaanHoogland About removing the `isEnabled`, if you take a look at the methods (`debug`, `trace`, or others) on [Logger](https://github.com/apache/log4j/blob/trunk/src/main/java/org/apache/log4j/Logger.java) and [Category](https://github.com/apache/log4j/blob/trunk/src/main/java/org/apache/log4j/Category.java), you will notice that they already do the same validation as the `isEnabled` do, the only thing that the `if` would avoid is that in some cases the string to be logged would not be created; In an async web app, like `ACS`, it is not relevant to safe this "cost" by adding more lines to the codebase.

About changing the log level to `debug`, in this context, these messages work better as `debug`; as well, it is quite unusual to call `trace` on `ACS`, so, it is interesting to keep a pattern and use `debug`.

DaanHoogland: @GutoVeronezi you are right they do the check and this is an often returning discussion; they do the check after parameter evaluation which in this case is just the passing of the string. This is why I'd like the evaluation/construction of the String inside the `isEnabled` guarded block as well. no need to perform an expensive string contruction if you're not going to use it.

GutoVeronezi: @DaanHoogland  I understand your concern, however it is not like we are running `ACS` on an ENIAC; the cost to build a string (with `java.lang`) is barely notable.    
Besides that, it is better to read a code without 'useless' lines.

Furthermore, if we update `log4j` to `log4j 2` it would not be a problem. =)

DaanHoogland: log4j 3 has my vote, big job though. this is a pattern that finds following and this is what i want to prevent, we need very elaborate Sting construction at times. I wouldn't hassle about info or above, but debug and trace, yes. (also aggree on readible code btw, but isEnabled is very explicit. And also agree on your use of `format()`

GutoVeronezi: Well, if the real problem is only the cost of building this string, then I would ask you to take another look at the preview code, because, in this case, the string is already built when it reach the `if` statement, I just took off the `if` and changed it to `debug`.

DaanHoogland: Ah, that's why I asked you to move the string creation inside the if block as well.
```suggestion
                if (LOG.isDebugEnabled()) {
                    final String message = String.format("Transitioned host HA state from:%s to:%s due to event:%s for the host id:%d",
                        currentHAState, nextState, event, haConfig.getResourceId());
                    LOG.debug(message);
                }
```
I see now that the message is used later in the block. I don't like this pattern, but you are right.

DaanHoogland (approved):

server/src/main/java/org/apache/cloudstack/ha/HAManagerImpl.java
```diff
@@ -315,14 +321,10 @@ public Status getHostStatus(final Host host) {
...
-                if (LOG.isDebugEnabled()){
-                    LOG.debug("HA: Agent is available/suspect/checking Up " + host.getId());
-                }
+                LOG.debug(String.format("HA: Agent [%s] is available/suspect/checking Up.", host.getId()));
                 return Status.Down;
             } else if (haConfig.getState() == HAConfig.HAState.Degraded || haConfig.getState() == HAConfig.HAState.Recovering || haConfig.getState() == HAConfig.HAState.Fencing) {
-                if (LOG.isDebugEnabled()){
-                    LOG.debug("HA: Agent is disconnected " + host.getId());
-                }
+                LOG.debug(String.format("HA: Agent [%s] is disconnected. State: %s, %s.", host.getId(), haConfig.getState(), haConfig.getState().getDescription()));
```
DaanHoogland, line 327: I still really don't like this

server/src/main/java/org/apache/cloudstack/ha/HAManagerImpl.java
GabrielBrascher: I was wondering if some of the HA messages should be logged as DEBUG or INFO levels. I think that there are some quite important information to raise for ADMINs regardless of they wanting a debug, info, or trace log level. 

For example, when a host gets fenced.

What do you think @DaanHoogland @GutoVeronezi? Does that make sense?

DaanHoogland: I totally agree that debug level has been highly abused in the project. lot's of production environments use debug level by default for that reason. This should not have an impact on this PR though.

GabrielBrascher (approved):
Code LGTM

PR #4721, opened 2021-02-22
preserve root cause in exception

DaanHoogland:
### Description

This PR... preserve a root cause exception

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial


engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
```diff
@@ -1932,8 +1932,9 @@ public void copyAsync(Map<VolumeInfo, DataStore> volumeDataStoreMap, VirtualMach
             errMsg = String.format("Copy volume(s) to storage(s) [%s] and VM to host [%s] failed in StorageSystemDataMotionStrategy.copyAsync. Error message: [%s].", volumesAndStorages, formatMigrationElementsAsJsonToDisplayOnLog("vm", vmTO.getId(), srcHost.getId(), destHost.getId()), ex.getMessage());
             LOGGER.error(errMsg, ex);
 
-            throw new CloudRuntimeException(errMsg);
-        } finally {
+            throw new CloudRuntimeException(errMsg, ex);
```
sureshanaparti, line 1935: @DaanHoogland The custom error message in the CloudRuntimeException thrown, has the exception message (of root cause), which is also logged along with the stack trace. So, the root cause message is already passed to the top layer. Correct?

DaanHoogland: You are right @sureshanaparti, I missed that line above. I must have solved something I saw in a different location than where it actually occured, because I didn't see a stack trace when I created this. closing

Issue #4726, opened 2021-02-24
Check and fix all UI issue from the archived primate repository


rhtyd:
Triage and address open issues from https://github.com/apache/cloudstack-primate/issues
And ask authors to port their PRs (if not done already) https://github.com/apache/cloudstack-primate/pulls

##### ISSUE TYPE

 * Task

##### COMPONENT NAME

~~~
UI
~~~

utchoang:
1. https://github.com/apache/cloudstack-primate/issues/158
   - [x] https://github.com/apache/cloudstack/pull/4762 - When opening a dialog, the first element should be active (with the cursor ready for typing). This avoids extra clicks.
   - [x] https://github.com/apache/cloudstack/pull/4763 - Some dialogs do not have 'Cancel' buttons - for example 'Assign Instance to Another Account'. The 'x' icon on the top right corner is always an option but I think we should be consistent here.

2. https://github.com/apache/cloudstack-primate/issues/462
   - [x] https://github.com/apache/cloudstack/pull/4764 - Create zone form is not aligned in mobile view
   - [x] https://github.com/apache/cloudstack/pull/4765 - Page size changer doesn't show up on mobile mode
   - [x] https://github.com/apache/cloudstack/pull/4766 - Enter to submit a form
   - [x] https://github.com/apache/cloudstack/pull/4769 - While listing domains, on full page reload, it goes back to the root domain
   - [x] https://github.com/apache/cloudstack/pull/4767 - Account in user info card doesn't directly go to account but to the account listview

3. https://github.com/apache/cloudstack-primate/issues/465 & https://github.com/apache/cloudstack-primate/issues/615
   - [x] https://github.com/apache/cloudstack/pull/4782 - Refactor async job polling codebase-wide and show multiple async jobs

4. https://github.com/apache/cloudstack-primate/issues/719
   - [x] https://github.com/apache/cloudstack/pull/4811 - Move fetch data to created()

5. https://github.com/apache/cloudstack-primate/issues/743
   - [x] https://github.com/apache/cloudstack/pull/4820 - Edit instance - offer existing Groups

rhtyd:
Thanks for the update @utchoang

rhtyd:
@utchoang are all the issues fixed/raised, can this ticket be closed?

utchoang:
Currently, there is still the PR (https://github.com/apache/cloudstack-primate/pull/898) to consider. I will update it now.

utchoang:
[...]

PR #4735, opened 2021-03-01
ui: fix all UI issue from the archived primate repository

utchoang:
### Description

This PR for fix https://github.com/apache/cloudstack/issues/4726 

1. https://github.com/apache/cloudstack-primate/issues/158
   - [x] When opening a dialog, the first element should be active (with the cursor ready for typing). This avoids extra clicks.
   - [x] Some dialogs do not have 'Cancel' buttons - for example 'Assign Instance to Another Account'. The 'x' icon on the top right corner is always an option but I think we should be consistent here.

2. https://github.com/apache/cloudstack-primate/issues/462
   - [x] Create zone form is not aligned in mobile view
   - [x] Page size changer doesn't show up on mobile mode
   - [x] Enter to submit a form
   - [x] While listing domains, on full page reload, it goes back to the root domain
   - [x] Account in user info card doesn't directly go to account but to the account listview

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial


DaanHoogland:
I like this effort, but should every fix be in a single PR? (i am busy wit a much larger PR myslef but still don't like the principle) How about creating separate issues and merging this now?

utchoang:
@DaanHoogland I will separate it into separate PRs for easier review.

PR #4750, opened 2021-03-04
Fix VMware vSAN storage support on vSphere6.0

rhtyd:
Use moveVirtualDisk method which handled symlinks also instead of moving each disk format files specific to a volume

### Types of changes


rhtyd:
Closing this, vsan users should use vsphere 6.5 and above with 4.15.

PR #4698, opened 2021-02-17
Fix npe when migrating vm with volume

GutoVeronezi:
### Description
In the execution flow of the API `migrateVirtualMachineWithVolume`, there is a step that copies the volume's template to the target storage, if it is needed.   
In some cases when the volume is a `data` type, it (the volume) does not have a template; therefore, it is going to generate an NPE when it tries to get the template's ID (Long to long auto conversion).

This PR intends to validate if the volume being copied has a template, before trying to copy the template. 

### Types of changes
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale
- [ ] Major
- [x] Minor

#### Bug Severity
- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?
It has been tested locally in a test lab.

1. I had created a VM and attached 2 volumes to it.
2. I had migrated the VM to another `host` and the volumes to different `storage` to see if it migrates accordingly.

PR #4775, opened 2021-03-09
[Backport] #4698 Fix npe when migrating vm with volume

GabrielBrascher:
### Description

PR #4698 was merged at the master (4.16) but it fixes an issue that is also relevant for current LTS versions.

Considering that the port is relatively straight-forward and it can benefit releases 4.14.2.0 and 4.15.1.0, this PR proposes back-porting PR #4698 into 4.14 and later forward it to 4.15.

More details at  #4698.

> In the execution flow of the API migrateVirtualMachineWithVolume, there is a step that copies the volume's template to the target storage, if it is needed.
> In some cases when the volume is a data type, it (the volume) does not have a template; therefore, it is going to generate an NPE when it tries to get the template's ID (Long to long auto conversion).
> 
> This PR intends to validate if the volume being copied has a template, before trying to copy the template.

### DIFF
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/KvmNonManagedStorageDataMotionStrategy.java
@@ -212,8 +212,10 @@ public class KvmNonManagedStorageDataMotionStrategy extends StorageSystemDataMot
                 if (copyCommandAnswer != null && copyCommandAnswer.getResult()) {
                     updateTemplateReferenceIfSuccessfulCopy(srcVolumeInfo, srcStoragePool, destTemplateInfo, destDataStore);
                 }
+                return;
             }
         }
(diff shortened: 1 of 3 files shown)


weizhouapache (approved):
lgtm. 

why did not create pr for 4.14 instead of 4.16?

weizhouapache:
> @weizhouapache thanks for the review!
> 
> It was the contributor option sending the changes to the master. I have nothing against their decision. I understand and respect their point of view.
> 
> Personally, I see the benefit of porting it into 4.14.2 and 4.15.1 and thus proposed this PR.
> 
> [Here is the discussion about it with the contributor point of view.](https://github.com/apache/cloudstack/pull/4698#issuecomment-781257832)

@GabrielBrascher thanks for your explanation.
sounds good to me.

rhtyd:
@blueorangutan package

shwstppr:
@blueorangutan test

shwstppr (approved):

GabrielBrascher:
@blueorangutan package

shwstppr:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

PR #4790, opened 2021-03-10
Externalize secondary storage capacity threshold

GutoVeronezi:
### Description
On StatsCollector, it limits secondary storages capacity to 90%; therefore, operators do not have option to choose which threshold they want/need. 

This PR intends to externalize the secondary storage capacity threshold configuration on StatsCollector (`secondary.storage.capacity.threshold`), validating if it is between 0 and 100% (0.00 and 1.00).

### Types of changes

- [x] Major
- [ ] Minor

### How Has This Been Tested?
1. I enabled a zone;
2. I created a template and observed the log. It will log messages according to the configuration set.

### DIFF
--- a/server/src/main/java/com/cloud/server/StatsCollector.java
+++ b/server/src/main/java/com/cloud/server/StatsCollector.java
@@ -1389,7 +1389,7 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
             return true;
         }
 
-        s_logger.warn(String.format("Image storage [%s] has not enough capacity. Capacity: total=[%s], used=[%s], threshold=[%s%%].", imageStoreId,readableTotalCapacity, readableUsedCapacity, threshold * 100));
+        s_logger.warn(String.format("Image storage [%s] has not enough capacity. Capacity: total=[%s], used=[%s], threshold=[%s%%].", imageStoreId, readableTotalCapacity, readableUsedCapacity, threshold * 100));
         return false;
     }
 
--- a/server/src/main/java/com/cloud/server/StatsCollector.java
+++ b/server/src/main/java/com/cloud/server/StatsCollector.java
@@ -146,6 +146,7 @@ import com.cloud.vm.dao.VMInstanceDao;
 
 import static com.cloud.utils.NumbersUtil.toHumanReadableSize;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.lang3.math.NumberUtils;
 
 /**
  * Provides real time stats for various agent resources up to x seconds
@@ -1625,16 +1626,27 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
     }
 
     public double getImageStoreCapacityThreshold() {
-        double thresholdConfig = secondaryStorageCapacityThreshold.value();
+        double thresholdConfigValue = secondaryStorageCapacityThreshold.value();
+        double thresholdConfigDefaultValue = NumberUtils.toDouble(secondaryStorageCapacityThreshold.defaultValue());
+        String thresholdConfigKey = secondaryStorageCapacityThreshold.key();
 
-        if (thresholdConfig >= MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD && thresholdConfig <= MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD) {
-            return thresholdConfig;
+        if (thresholdConfigValue >= MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD && thresholdConfigValue <= MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD) {
+            return thresholdConfigValue;
         }
 
-        s_logger.warn(String.format("Invalid [%s] configuration: value set [%s] is [%s]. Assuming %s as secondary storage capacity threshold.",
-                                    secondaryStorageCapacityThreshold.key(),
-                                    thresholdConfig,
-                                    thresholdConfig < MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD ? String.format("lower than '%s'", MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD) : String.format("bigger than '%s'", MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD),
+        if (thresholdConfigValue < MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD) {
+            s_logger.warn(String.format("Invalid [%s] configuration: value set [%s] is lower than '%s'. Assuming '%s', default value, as secondary storage capacity threshold.",
+                                        thresholdConfigKey,
+                                        thresholdConfigValue,
+                                        MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD,
+                                        thresholdConfigDefaultValue));
+            return thresholdConfigDefaultValue;
+        }
+
+        s_logger.warn(String.format("Invalid [%s] configuration: value set [%s] is bigger than '%s'. Assuming '%s', top limit, as secondary storage capacity threshold.",
+                                    thresholdConfigKey,
+                                    thresholdConfigValue,
+                                    MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD,
                                     MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD));
         return MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD;
     }
--- a/server/src/main/java/com/cloud/server/StatsCollector.java
+++ b/server/src/main/java/com/cloud/server/StatsCollector.java
@@ -207,9 +207,6 @@ public class StatsCollector extends ManagerBase implements ComponentMethodInterc
     private static final String INFLUXDB_HOST_MEASUREMENT = "host_stats";
     private static final String INFLUXDB_VM_MEASUREMENT = "vm_stats";
 
-    private static final double MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD = 0.00;
-    private static final double MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD = 1.00;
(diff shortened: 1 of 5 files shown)


server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -1599,10 +1621,21 @@ public String getConfigComponentName() {
...
+        if (thresholdConfig >= MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD && thresholdConfig <= MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD) {
+            return thresholdConfig;
+        }
+
+        s_logger.warn(String.format("Invalid [%s] configuration: value set [%s] is [%s]. Assuming %s as secondary storage capacity threshold.",
+                                    secondaryStorageCapacityThreshold.key(),
+                                    thresholdConfig,
+                                    thresholdConfig < MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD ? String.format("lower than '%s'", MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD) : String.format("bigger than '%s'", MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD),
+                                    MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD));
+        return MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD;
```
DaanHoogland, line 1639: makes you wish for a bounds checker on entry, :+1:

DaanHoogland, line 1639: i think we should revert to default, i.e. `0.9`

GutoVeronezi: I thought about it too and came to a possible resolution:

- When the config is bigger than max, I think we should return max (as it is top limit);
- When the config is lower than min, then we use default;

What do you think?

DaanHoogland: sound like a plan

DaanHoogland (approved):
cltgm

GabrielBrascher (approved):
Code LGTM. I did not test it, though.

GabrielBrascher:
@blueorangutan package

GutoVeronezi:
@DaanHoogland @GabrielBrascher is there anything else to do?

rhtyd:
@blueorangutan package

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -1367,10 +1372,28 @@ public boolean imageStoreHasEnoughCapacity(DataStore imageStore) {
...
+        String readableTotalCapacity = FileUtils.byteCountToDisplaySize((long) totalCapacity);
+        String readableUsedCapacity = FileUtils.byteCountToDisplaySize((long) usedCapacity);
+
+        s_logger.debug(String.format("Verifying image storage [%s]. Capacity: total=[%s], used=[%s], threshold=[%s%%].", imageStoreId, readableTotalCapacity, readableUsedCapacity, threshold * 100));
+
+        if (usedCapacity / totalCapacity <= threshold) {
             return true;
         }
+
+        s_logger.warn(String.format("Image storage [%s] has not enough capacity. Capacity: total=[%s], used=[%s], threshold=[%s%%].", imageStoreId,readableTotalCapacity, readableUsedCapacity, threshold * 100));
```
sureshanaparti, line 1396:
```suggestion
        s_logger.warn(String.format("Image storage [%s] has not enough capacity.", imageStoreId));
```

capacity/used bytes and threshold already logged above, may not be required here.

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -219,6 +224,8 @@ public String toString() {
     private static final ConfigKey<String> statsOutputUri = new ConfigKey<String>("Advanced", String.class, "stats.output.uri", "",
             "URI to send StatsCollector statistics to. The collector is defined on the URI scheme. Example: graphite://graphite-hostaddress:port or influxdb://influxdb-hostaddress/dbname. Note that the port is optional, if not added the default port for the respective collector (graphite or influxdb) will be used. Additionally, the database name '/dbname' is  also optional; default db name is 'cloudstack'. You must create and configure the database if using influxdb.",
             true);
+    private static final ConfigKey<Double> secondaryStorageCapacityThreshold = new ConfigKey<Double>("Advanced", Double.class, "secondary.storage.capacity.threshold", "0.90",
+            "Secondary storage capacity threshold (1 = 100%).", true);
```
sureshanaparti, line 228: @GutoVeronezi can you move this config to 'CapacityManager' ? and use it wherever applicable.

also, update StorageOrchestrator class with this config, it is hardcoded here: https://github.com/apache/cloudstack/blob/a0d9acec6f3926ef31430a2be0c2737b30eb5ebb/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/StorageOrchestrator.java#L104

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -204,6 +206,9 @@ public String toString() {
     private static final String INFLUXDB_HOST_MEASUREMENT = "host_stats";
     private static final String INFLUXDB_VM_MEASUREMENT = "vm_stats";
 
+    private static final double MIN_STORAGE_SECONDARY_CAPACITY_THRESHOLD = 0.00;
```
sureshanaparti, line 209:
```suggestion
    private static final double MIN_SECONDARY_STORAGE_CAPACITY_THRESHOLD = 0.00;
```

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -204,6 +206,9 @@ public String toString() {
...
+    private static final double MAX_STORAGE_SECONDARY_CAPACITY_THRESHOLD = 1.00;
```
sureshanaparti, line 210:
```suggestion
    private static final double MAX_SECONDARY_STORAGE_CAPACITY_THRESHOLD = 1.00;
```

server/src/main/java/com/cloud/server/StatsCollector.java
sureshanaparti: @GutoVeronezi these value limits are controlled/validated in the configuration manager. you can add this new config to '_weightBasedParametersForValidation_' in configuration manager. check here:  https://github.com/apache/cloudstack/blob/a0d9acec6f3926ef31430a2be0c2737b30eb5ebb/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java#L479

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -219,6 +224,8 @@ public String toString() {
     private static final ConfigKey<String> statsOutputUri = new ConfigKey<String>("Advanced", String.class, "stats.output.uri", "",
             "URI to send StatsCollector statistics to. The collector is defined on the URI scheme. Example: graphite://graphite-hostaddress:port or influxdb://influxdb-hostaddress/dbname. Note that the port is optional, if not added the default port for the respective collector (graphite or influxdb) will be used. Additionally, the database name '/dbname' is  also optional; default db name is 'cloudstack'. You must create and configure the database if using influxdb.",
             true);
+    private static final ConfigKey<Double> secondaryStorageCapacityThreshold = new ConfigKey<Double>("Advanced", Double.class, "secondary.storage.capacity.threshold", "0.90",
```
sureshanaparti, line 227: The description for most of the percentage based configs clear indicate that values should be between 0 and 1, and same is being validated in the configuration manager. can you search code with string '_Percentage (as a value between 0 and 1)_ ' and update the config description as appropriate. Thanks.

server/src/main/java/com/cloud/server/StatsCollector.java
GutoVeronezi: Done.

server/src/main/java/com/cloud/server/StatsCollector.java
GutoVeronezi: Done.

server/src/main/java/com/cloud/server/StatsCollector.java
GutoVeronezi: Done.

server/src/main/java/com/cloud/server/StatsCollector.java
GutoVeronezi: I think it could be interesting to log here too to not depend on another log line; Also we could lose the information in some situations.

GutoVeronezi:
@sureshanaparti Thank you for this review. 

As you suggested, I added the config to the validation in `ConfigurationManagerImpl`. However, I had to change the config type from `Double` to `Float` (not that it is a problem) because the validation only works on `Float` configs. 
I tested some of they and realize that those with `Double` type are not validated. I think we need some refactors in this validation, in another PR. What do you think?

DaanHoogland:
@sureshanaparti are you satisfied?
do we need more testing? cc @borisstoyanov @vladimirpetrov

GutoVeronezi:
@sureshanaparti @DaanHoogland @GabrielBrascher is there anything else to do?

GutoVeronezi:
@sureshanaparti @DaanHoogland @GabrielBrascher is there anything else to do?

sureshanaparti:
> @sureshanaparti Thank you for the review.
> 
> As you suggested, I added the config to the validation in `ConfigurationManagerImpl`. However, I had to change the config type from `Double` to `Float` (not that it is a problem) because the validation only works on `Float` configs.
> I tested some of they and realize that those with `Double` type are not validated. I think we need some refactors in this validation, in another PR. What do you think?

@GutoVeronezi agree, these validations can be refactored in another PR.

sureshanaparti:
@blueorangutan package

sureshanaparti (approved):

sureshanaparti:
@blueorangutan test

PR #4793, opened 2021-03-11
systemvmtemplate: new template for 4.15.1

rhtyd:
Update new systemvmtemplate for 4.15.1.0; synced:
http://download.cloudstack.org/systemvm/4.15/

A new template is necessary due to many security fixes over the last year, the 4.15.0 systemvmtemplate was created about a year ago.


rhtyd:
@blueorangutan package

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
```diff
@@ -58,6 +66,175 @@ public boolean supportsRollingUpgrade() {
...
+                put(Hypervisor.HypervisorType.VMware, "4006982765846d373eb3719b2fe4d720");
+                put(Hypervisor.HypervisorType.Hyperv, "0b9514e4b6cba1f636fea2125f0f7a5f");
+                put(Hypervisor.HypervisorType.LXC, "0e9f9a7d0957c3e0a2088e41b2da2cec");
+                put(Hypervisor.HypervisorType.Ovm3, "ae3977e696b3e6c81bdcbb792d514d29");
+            }
+        };
+
+        for (final Map.Entry<Hypervisor.HypervisorType, String> hypervisorAndTemplateName : NewTemplateNameList.entrySet()) {
+            LOG.debug("Updating " + hypervisorAndTemplateName.getKey() + " System Vms");
+            try (PreparedStatement pstmt = conn.prepareStatement("select id from `cloud`.`vm_template` where name = ? and removed is null order by id desc limit 1")) {
```
weizhouapache, line 150: @rhtyd 
if an user (not root admin) register a template with the same name, will it be a security issue ?

rhtyd: Good catch - we've been copying the same code/logic; however I think it may not be a big issue as it picks the last template that was added before the upgrade, even if user may have added a template, the admin likely would have added the most recent or last template. We may be saved by the `order by id desc limit 1`. Or, do you want me to add a check for `account_idj = 2`?

rhtyd:
@blueorangutan test matrix

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
weizhouapache: @rhtyd 
is it possible that cloudstack user uses other admin account than root admin (account_id=2) ?

would it be better to use this ?
```
select id from `cloud`.`vm_template` where name = ? and removed is null and account_id in (select id from account where role_id=1 and removed is null) order by id desc limit 1;
```

rhtyd: Thanks, I'll test and include it

rhtyd: ... but what is the account is of admin role type but not role ID 1?

DaanHoogland: what would be the issue, @rhtyd? except that the account used will fail registering systemvm templates?

tools/appliance/systemvmtemplate/template.json
```diff
@@ -36,8 +36,8 @@
       "disk_size": 2500,
       "disk_interface": "virtio",
       "net_device": "virtio-net",
-      "iso_url": "https://download.cloudstack.org/systemvm/debian/debian-10.5.0-amd64-netinst.iso",
-      "iso_checksum": "0a6aee1d9aafc1ed095105c052f9fdd65ed00ea9274188c9cd0072c8e6838ab40e246d45a1e6956d74ef1b04a1fc042151762f25412e9ff0cbf49418eef7992e",
+      "iso_url": "https://cdimage.debian.org/debian-cd/10.8.0/amd64/iso-cd/debian-10.8.0-amd64-netinst.iso",
```
DaanHoogland, line 39: debian tends to move their ISOs, is this the best way, or should we keep a vopy as we used to?

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test matrix

shwstppr:
@blueorangutan package

davidjumani:
@blueorangutan package

rhtyd:
@blueorangutan package

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
rhtyd: Looks like we can use Wei's query with type=1 (instead of role_id); type1  is root admin role.

rhtyd:
@Pearl1594 to carry forward with the PR

tools/appliance/systemvmtemplate/template.json
rhtyd: It's either this or someone manually copying/updating the template on  download.cloudstack.org and then using it here. One option is to use the archive links, but Debian doesn't put the latest version/release isos there https://cdimage.debian.org/mirror/cdimage/archive/

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan test matrix

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
```diff
@@ -58,6 +66,175 @@ public boolean supportsRollingUpgrade() {
...
+                put(Hypervisor.HypervisorType.VMware, "4006982765846d373eb3719b2fe4d720");
+                put(Hypervisor.HypervisorType.Hyperv, "0b9514e4b6cba1f636fea2125f0f7a5f");
+                put(Hypervisor.HypervisorType.LXC, "0e9f9a7d0957c3e0a2088e41b2da2cec");
+                put(Hypervisor.HypervisorType.Ovm3, "ae3977e696b3e6c81bdcbb792d514d29");
+            }
+        };
+
+        for (final Map.Entry<Hypervisor.HypervisorType, String> hypervisorAndTemplateName : NewTemplateNameList.entrySet()) {
+            LOG.debug("Updating " + hypervisorAndTemplateName.getKey() + " System Vms");
+            try (PreparedStatement pstmt = conn.prepareStatement("select id from `cloud`.`vm_template` where name = ? and removed is null and account_id in (select id from account where type = 1 and removed is NULL) order by id desc limit 1")) {
```
rhtyd, line 150: @weizhouapache @DaanHoogland the security issue fixed here by checking the template belongs to an account with type=1 (root admin)

rhtyd:
Tests LGTM, thanks for fixing the issue and kicking tests @Pearl1594

Pearl1594 (approved):
LGTM.
Verified the security issue by registering a template with name "systemvm-vmware-4.15.1" as a user post root admin registering the template and then initiated upgrade. It picks up the right systemVM template, i.e., the one registered by the root admin and updates the required fields e.g. type

shwstppr (approved):

rhtyd:
Thanks Pearl for testing.

PR #4800, opened 2021-03-11
kvm: Do not set backing file format of DATADISK in vm start/migration

ustcweizhou:
### Description

We do not need to set backing file format of  DATADISK and ROOT disk which does not have a backing file.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -4266,7 +4266,8 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             Map<String, String> info = qemu.info(file);
             String backingFilePath = info.get(new String("backing_file"));
             String backingFileFormat = info.get(new String("backing_file_format"));
-            if (org.apache.commons.lang.StringUtils.isEmpty(backingFileFormat)) {
+            if (org.apache.commons.lang.StringUtils.isNotBlank(backingFilePath)
+                    && org.apache.commons.lang.StringUtils.isEmpty(backingFileFormat)) {
                 s_logger.info("Setting backing file format of " + volPath);
                 QemuImgFile backingFile = new QemuImgFile(backingFilePath);
                 Map<String, String> backingFileinfo = qemu.info(backingFile);
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -4267,7 +4267,10 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             String backingFilePath = info.get(new String("backing_file"));
             String backingFileFormat = info.get(new String("backing_file_format"));
             if (org.apache.commons.lang.StringUtils.isNotBlank(backingFilePath)
-                    && org.apache.commons.lang.StringUtils.isEmpty(backingFileFormat)) {
+                    && org.apache.commons.lang.StringUtils.isBlank(backingFileFormat)) {
+                // VMs which are created in CloudStack 4.14 and before cannot be started or migrated
+                // in latest Linux distributions due to missing backing file format
+                // Please refer to https://libvirt.org/kbase/backing_chains.html#vm-refuses-to-start-due-to-misconfigured-backing-store-format
                 s_logger.info("Setting backing file format of " + volPath);
                 QemuImgFile backingFile = new QemuImgFile(backingFilePath);
                 Map<String, String> backingFileinfo = qemu.info(backingFile);


rhtyd:
@blueorangutan package

shwstppr:
@blueorangutan test

shwstppr (approved):
LGTM

DaanHoogland:
@weizhouapache I'm confused by your description in context of the code change. If we do have a disk, but no backing format, we go ahead and rebase it as qcow2, am I right?
and if we don't have a file path or we do have a backing format, we do nothing.

weizhouapache:
> @weizhouapache I'm confused by your description in context of the code change. If we do have a disk, but no backing format, we go ahead and rebase it as qcow2, am I right?
> and if we don't have a file path or we do have a backing format, we do nothing.

@DaanHoogland I changed the title and description.

we do nothing if (1) disk does not have a backing file ; (2) disk has backing file, and also backing file format.
we set backing file format only if disk has a backing file but does not have backing file format.

DaanHoogland:
> > @weizhouapache I'm confused by your description in context of the code change. If we do have a disk, but no backing format, we go ahead and rebase it as qcow2, am I right?
> > and if we don't have a file path or we do have a backing format, we do nothing.
> 
> @DaanHoogland I changed the title and description.
> 
> we do nothing if (1) disk does not have a backing file ; (2) disk has backing file, and also backing file format.
> we set backing file format only if disk has a backing file but backing file format is missing

exactly what the code does, good text for the tin ;)

DaanHoogland (approved):

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -4266,7 +4266,8 @@ public void setBackingFileFormat(String volPath) {
             Map<String, String> info = qemu.info(file);
             String backingFilePath = info.get(new String("backing_file"));
             String backingFileFormat = info.get(new String("backing_file_format"));
-            if (org.apache.commons.lang.StringUtils.isEmpty(backingFileFormat)) {
+            if (org.apache.commons.lang.StringUtils.isNotBlank(backingFilePath)
+                    && org.apache.commons.lang.StringUtils.isEmpty(backingFileFormat)) {
```
rhtyd, line 4270: What if the backing file format is not empty?

weizhouapache: @rhtyd then it is fine, we do nothing.

GabrielBrascher: Maybe some log messages for debug/warn purposes would be interesting. Or do you think that it is something that system admins should not be bothered with?

I am thinking especially on cases such as `backingFileFormat` not empty and blank/empty `backingFilePath`, or _vice versa_.

Additionally, what do you think of replacing `StringUtils.isEmpty(backingFileFormat)` with `StringUtils.isBlank(backingFileFormat)`? In such case it would then check if `backingFileFormat` is whitespace, empty ("") or null.

GabrielBrascher (commented):
Thanks for the PR @weizhouapache.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
weizhouapache: > Maybe some log messages for debug/warn purposes would be interesting. Or do you think that it is something that system admins should not be bothered with?
> 
> I am thinking especially on cases such as `backingFileFormat` not empty and blank/empty `backingFilePath`, or _vice versa_.
> 
> Additionally, what do you think of replacing `StringUtils.isEmpty(backingFileFormat)` with `StringUtils.isBlank(backingFileFormat)`? In such case it would then check if `backingFileFormat` is whitespace, empty ("") or null.

@GabrielBrascher there is a info next to these lines
```
                s_logger.info("Setting backing file format of " + volPath);
```
I will add a line as comment to explain why we need this change.
https://libvirt.org/kbase/backing_chains.html#vm-refuses-to-start-due-to-misconfigured-backing-store-format

StringUtils.isBlank(backingFileFormat) and StringUtils.isEmpty(backingFileFormat) have no difference I think.  there are only  2 possible values: empty and 'qcow2'. I will change to isBlank

GabrielBrascher (approved):
LGTM

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

rhtyd:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -4266,7 +4266,11 @@ public void setBackingFileFormat(String volPath) {
             Map<String, String> info = qemu.info(file);
             String backingFilePath = info.get(new String("backing_file"));
```
rubieHess, line 4267: I detect that this code is problematic. According to the [Performance (PERFORMANCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#performance-performance), [Dm: Method invokes inefficient new String(String) constructor (DM_STRING_CTOR)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dm-method-invokes-inefficient-new-string-string-constructor-dm-string-ctor).
Using the java.lang.String(String) constructor wastes memory because the object so constructed will be functionally indistinguishable from the String passed as a parameter.  Just use the argument String directly.

weizhouapache: @rubieHess good point. thanks ! I will create a pr for it.

rubieHess: Happy to help.

rhtyd:
@blueorangutan test

PR #4816, opened 2021-03-15
xenserver: retrieve correct name-label for presetup store

### LINKED ISSUE #4729: XenServer, management server sends pool UUID instead of the PATH, as the SR name
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~

~~~

##### CLOUDSTACK VERSION

~~~
4.14.0.0 tested
~~~

##### CONFIGURATION

Original Cloudplatform 4.7.1 with XenServer with preSetup storage (iscis) - all this is less relevant, see down

##### OS / ENVIRONMENT

NA

##### SUMMARY

After upgrading ACP 4.7.1 to ACS 4.14.0.0, preSetup storage type can not be mounted on XS servers, as ACS DB is sending UUID (instead of the pool name) to the XS to mount it - see details below

DB upgraded to 4.14, but the same data is in 4.7.1 before the upgrade - storage_pool table:

                   id: 2
                 name: XenServer-Primary
                 **uuid: 1e189b97-974d-35b6-bc76-50bca7704ea0**
            pool_type: PreSetup
                 port: 0
       data_center_id: 2
               pod_id: 2
           cluster_id: 3
           used_bytes: 28898754560
       capacity_bytes: 214714810368
         host_address: localhost
            user_info: NULL
                 path: /XEN_PRIMARY
              created: 2021-02-24 10:38:37
              removed: NULL
          update_time: NULL
               status: Maintenance
               storage_provider_name: DefaultPrimary
                scope: CLUSTER
           hypervisor: NULL
              managed: 0
        capacity_iops: NULL

In CLoudPlatform 4.7.1. this used to work fine, "PATH" filed was sent as the SR name to the XS to mount it.

In CloudStack 4.14.0.0, mgmt server sends UUID (1e189b97-974d-35b6-bc76-50bca7704ea0) as the SR name parameter for XS when mounting it, instead of "path" (XEN_PRIMARY), so the SR mount fails:

2021-02-25 03:49:45,306 INFO  [c.c.v.VirtualMachineManagerImpl] (Work-Job-Executor-3:ctx-7a7ee26b job-118/job-119 ctx-6e3dd2e5) (logid:edf9d62a) Unable to contact resource.
com.cloud.exception.StorageUnavailableException: Resource [StoragePool:2] is unreachable: Unable to create Vol[26|vm=26|ROOT]:**There are 0 SRs with same name: 1e189b97-974d-35b6-bc76-50bca7704ea0**

The workaround was (this is customer env) to replace UUID value with the value of the "PATH" field so that XS can mount it (what consequences this might have, is unknown)

shwstppr:
### Description

Fixes #4729

As reported in the issue ACP 4.7 used a normal UUID in db for a presetup primary store on Xenserver.
Later the value has been changed to store's path with '/' removed.
Current changes try to retrieve SR's name-lable from store's path if UUID doesn't match path field for a pre-setup store.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
@@ -16,7 +16,6 @@
 // under the License.
 package com.cloud.hypervisor.xenserver.resource;
 
-import com.cloud.agent.api.to.StorageFilerTO;
 import com.cloud.storage.Storage;
 import com.xensource.xenapi.Host;
 
@@ -47,11 +46,13 @@ public class CitrixHelper {
         return "";
     }
 
-    public static String getSRNameLabel(final StorageFilerTO primaryStore) {
-        if (Storage.StoragePoolType.PreSetup.equals(primaryStore.getType()) &&
-                !primaryStore.getPath().contains(primaryStore.getUuid())) {
-            return  primaryStore.getPath().replace("/", "");
+    public static String getSRNameLabel(final String poolUuid,
+                                        final Storage.StoragePoolType poolType,
+                                        final String poolPath) {
+        if (Storage.StoragePoolType.PreSetup.equals(poolType) &&
+                !poolPath.contains(poolUuid)) {
+            return  poolPath.replaceFirst("/", "");
         }
-        return primaryStore.getUuid();
+        return poolUuid;
     }
 }
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
@@ -16,6 +16,8 @@
 // under the License.
 package com.cloud.hypervisor.xenserver.resource;
 
+import com.cloud.agent.api.to.StorageFilerTO;
+import com.cloud.storage.Storage;
 import com.xensource.xenapi.Host;
 
 /**
@@ -43,5 +45,13 @@ public class CitrixHelper {
             }
         }
         return "";
+    }
+
+    public static String getSRNameLabel(final StorageFilerTO primaryStore) {
+        if (Storage.StoragePoolType.PreSetup.equals(primaryStore.getType()) &&
+                !primaryStore.getPath().contains(primaryStore.getUuid())) {
+            return  primaryStore.getPath().replace("/", "");
+        }
+        return primaryStore.getUuid();
     }
 }
--- a/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/XenServerStorageProcessor.java
+++ b/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/XenServerStorageProcessor.java
@@ -96,6 +96,14 @@ public class XenServerStorageProcessor implements StorageProcessor {
         hypervisorResource = resource;
     }
 
+    protected String getSRNameLabel(final PrimaryDataStoreTO primaryStore) {
+        if (Storage.StoragePoolType.PreSetup.equals(primaryStore.getPoolType()) &&
+                !primaryStore.getPath().contains(primaryStore.getUuid())) {
+            return  primaryStore.getPath().replace("/", "");
+        }
+        return primaryStore.getUuid();
+    }
+
     // if the source SR needs to be attached to, do so
     // take a snapshot of the source VDI (on the source SR)
     // create an iSCSI SR based on the new back-end volume
@@ -790,7 +798,8 @@ public class XenServerStorageProcessor implements StorageProcessor {
 
         try {
             final Connection conn = hypervisorResource.getConnection();
-            final SR poolSr = hypervisorResource.getStorageRepository(conn, data.getDataStore().getUuid());
+            final PrimaryDataStoreTO primaryStore = (PrimaryDataStoreTO)data.getDataStore();
+            final SR poolSr = hypervisorResource.getStorageRepository(conn, getSRNameLabel(primaryStore));
             VDI.Record vdir = new VDI.Record();
             vdir.nameLabel = volume.getName();
             vdir.SR = poolSr;
@@ -863,7 +872,8 @@ public class XenServerStorageProcessor implements StorageProcessor {
         if (srcStore instanceof NfsTO) {
             final NfsTO nfsStore = (NfsTO) srcStore;
             try {
-                final SR primaryStoragePool = hypervisorResource.getStorageRepository(conn, destVolume.getDataStore().getUuid());
+                final PrimaryDataStoreTO destStore = (PrimaryDataStoreTO)destVolume.getDataStore();
+                final SR primaryStoragePool = hypervisorResource.getStorageRepository(conn, getSRNameLabel(destStore));
                 final String srUuid = primaryStoragePool.getUuid(conn);
                 final URI uri = new URI(nfsStore.getUrl());
                 final String volumePath = uri.getHost() + ":" + uri.getPath() + nfsStore.getPathSeparator() + srcVolume.getPath();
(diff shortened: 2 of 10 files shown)


shwstppr:
@blueorangutan package

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
```diff
@@ -44,4 +46,12 @@ public static String getPVbootloaderArgs(String guestOS) {
         }
         return "";
     }
+
+    public static String getSRNameLabel(final StorageFilerTO primaryStore) {
+        if (Storage.StoragePoolType.PreSetup.equals(primaryStore.getType()) &&
+                !primaryStore.getPath().contains(primaryStore.getUuid())) {
+            return  primaryStore.getPath().replace("/", "");
```
DaanHoogland, line 53: doesn't this replace all '/'? seems dangerous to me

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/XenServerStorageProcessor.java
```diff
@@ -96,6 +96,14 @@ public XenServerStorageProcessor(final CitrixResourceBase resource) {
         hypervisorResource = resource;
     }
 
+    protected String getSRNameLabel(final PrimaryDataStoreTO primaryStore) {
+        if (Storage.StoragePoolType.PreSetup.equals(primaryStore.getPoolType()) &&
+                !primaryStore.getPath().contains(primaryStore.getUuid())) {
+            return  primaryStore.getPath().replace("/", "");
```
DaanHoogland, line 102: as above. Make sure it is only the last one being replaced.

DaanHoogland: also why is this method double implemented? here and in CitrixHelper? <quote style='ominous'>there can be only one</quote>

DaanHoogland (changes requested):

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
shwstppr: https://github.com/apache/cloudstack/blob/4.15/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java#L340-L342

Primary store setup code uses the same. I can change this though.
For pre-setup SR path is `/SR_NAME_LABEL` so it might not be as dangerous as it looks.

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/XenServerStorageProcessor.java
shwstppr: `StorageFilerTO`, `PrimaryDataStoreTO` and `StoragePool` different classes

shwstppr:
@blueorangutan package

plugins/hypervisors/xenserver/src/main/java/org/apache/cloudstack/storage/motion/XenServerStorageMotionStrategy.java
```diff
@@ -326,7 +327,13 @@ private Answer migrateVmWithVolumesAcrossCluster(VMInstanceVO vm, VirtualMachine
                     volumeToStorageUuid.add(new Pair<>(volumeTo, iqn));
                 }
                 else {
-                    volumeToStorageUuid.add(new Pair<>(volumeTo, ((StoragePool)entry.getValue()).getUuid()));
+                    StoragePool pool = (StoragePool)entry.getValue();
+                    String srNameLabel = pool.getUuid();
+                    if (Storage.StoragePoolType.PreSetup.equals(pool.getPoolType()) &&
+                            !pool.getPath().contains(pool.getUuid())) {
+                        srNameLabel = pool.getPath().replaceFirst("/", "");
+                    }
```
DaanHoogland, line 335:
```suggestion
                    Sting srNameLabel = CitrixHelper.getSRNameLabel(pool.getUuid(), pool.getPoolType(), pool.getPath())
```
or if need be put the usitlity method elsewere, but make sure it only exists once!?

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
```diff
@@ -44,4 +45,14 @@ public static String getPVbootloaderArgs(String guestOS) {
         }
         return "";
     }
+
+    public static String getSRNameLabel(final String poolUuid,
+                                        final Storage.StoragePoolType poolType,
+                                        final String poolPath) {
+        if (Storage.StoragePoolType.PreSetup.equals(poolType) &&
+                !poolPath.contains(poolUuid)) {
```
DaanHoogland, line 53: aren't there any presetup scenarios where the path could contain the uuid?

DaanHoogland (commented):
1 comment, 1 question

davidjumani:
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

Pearl1594:
@blueorangutan package

Pearl1594:

Pearl1594:
Manually verified the fix - performing the following steps:
- Created an env with nfs stores
- Marked one store as deleted / removed from the db and re-added that store to cloudstack using presetup protocol
- updated the storage_pool table for that corresponding record and replaced uuid field which is currently set to the sr name-label to the actual uuid  of the sr as seen with xe sr-list -> at this point, we have a record similar to what it would have been if the pool  were registered in CloudPlatform 4.7 
- Successfully deployed a VM with its volume on the pool that was added / manipulated as stated above
- Migrated a volume to the storage pool in question and ACS was able to correctly map the uuid to the sr name label

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixHelper.java
shwstppr: @DaanHoogland For presetup pools, UUID in DB is set as the path of the pool (without `/`)
But in ACP it was using real UUIDs as reported in the linked issue.
So my change is just to use path (without `/`) when there is a mismatch between path and UUID and the pool is of type presetup

shwstppr:
@blueorangutan package

Pearl1594:
blueorangutan test centos7 xenserver-71

Pearl1594:

plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/XenServerStorageProcessor.java
DaanHoogland: Yes, @shwstppr but different classes implementing the exact same doesn't like right to me. It seems to me
1. the method should be a utility (easy dirty quick solution) or
2. the storage path should have it's own class that is referred from all three users of the path.

rhtyd:

rhtyd:

DaanHoogland (approved):
code looks good

rhtyd:
@Pearl1594 are you testing this, or lgtm on merging this?

Pearl1594:
@rhtyd It's already been tested (https://github.com/apache/cloudstack/pull/4816#issuecomment-806649136). It's ready for merging

Pearl1594 (approved):
LGTM

PR #4819, opened 2021-03-15
Remove meaningless password regeneration on resetSSHKeyForVirtualMachine

GutoVeronezi:
### Description
On API `resetSSHKeyForVirtualMachine`, ACS also regenerates VM password when it uses a template with `Password Enabled` as true; there is already anAPI to reset VM password, therefore, the reset SSH keys API should not reset the VM SSH password as well.

Besides running a meaningless process, the VM's password regeneration slows down the main process and may cause a confusion in operations due to password change in the VM without being explicity requested.

This PR intends to remove the meaningless password regeneration from `resetSSHKeyForVirtualMachine`. 
### Types of changes
- [ ] Major
- [x] Minor
### How Has This Been Tested?
It has been tested locally in a test lab.
1. I created VM with a template with `Password Enabled` as true.
2. I logged in VM with the generated password on creation.
3. I stopped the VM and called the API.
4. I started the VM and tried to login with the password. Before this changes, the password would be changed and I could not login to the VM because I did not have the password anymore. After this changes, I could login normally.

### DIFF
--- a/server/src/main/java/com/cloud/server/ManagementServerImpl.java
+++ b/server/src/main/java/com/cloud/server/ManagementServerImpl.java
@@ -4079,26 +4079,23 @@ public class ManagementServerImpl extends ManagerBase implements ManagementServe
     }
 
     @Override
-    public String getVMPassword(final GetVMPasswordCmd cmd) {
-        final Account caller = getCaller();
+    public String getVMPassword(GetVMPasswordCmd cmd) {
+        Account caller = getCaller();
+        long vmId = cmd.getId();
+        UserVmVO vm = _userVmDao.findById(vmId);
 
-        final UserVmVO vm = _userVmDao.findById(cmd.getId());
         if (vm == null) {
-            final InvalidParameterValueException ex = new InvalidParameterValueException("No VM with specified id found.");
-            ex.addProxyObject(cmd.getId().toString(), "vmId");
-            throw ex;
+            throw new InvalidParameterValueException(String.format("No VM found with id [%s].", vmId));
         }
 
-        // make permission check
         _accountMgr.checkAccess(caller, null, true, vm);
 
         _userVmDao.loadDetails(vm);
-        final String password = vm.getDetail("Encrypted.Password");
-        if (password == null || password.equals("")) {
-            final InvalidParameterValueException ex = new InvalidParameterValueException(
-                    "No password for VM with specified id found. " + "If VM is created from password enabled template and SSH keypair is assigned to VM then only password can be retrieved.");
-            ex.addProxyObject(vm.getUuid(), "vmId");
-            throw ex;
+        String password = vm.getDetail("Encrypted.Password");
+
+        if (StringUtils.isEmpty(password)) {
+            throw new InvalidParameterValueException(String.format("No password found for VM with id [%s]. When the VM's SSH keypair is changed, the current encrypted password is "
+              + "removed due to incosistency in the encryptation, as the new SSH keypair is different from which the password was encrypted. To get a new password, it must be reseted.", vmId));
         }
 
         return password;
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -850,22 +850,19 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         }
 
         _accountMgr.checkAccess(caller, null, true, userVm);
-        String password = null;
-        String sshPublicKey = s.getPublicKey();
-        if (template != null && template.isEnablePassword()) {
-            password = _mgr.generateRandomPassword();
-        }
 
-        boolean result = resetVMSSHKeyInternal(vmId, sshPublicKey, password);
+        String sshPublicKey = s.getPublicKey();
+
+        boolean result = resetVMSSHKeyInternal(vmId, sshPublicKey);
 
         if (!result) {
             throw new CloudRuntimeException("Failed to reset SSH Key for the virtual machine ");
         }
-        userVm.setPassword(password);
+
         return userVm;
     }
 
-    private boolean resetVMSSHKeyInternal(Long vmId, String sshPublicKey, String password) throws ResourceUnavailableException, InsufficientCapacityException {
+    private boolean resetVMSSHKeyInternal(Long vmId, String sshPublicKey) throws ResourceUnavailableException, InsufficientCapacityException {
         Long userId = CallContext.current().getCallingUserId();
         VMInstanceVO vmInstance = _vmDao.findById(vmId);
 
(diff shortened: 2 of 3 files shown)


GabrielBrascher:
@blueorangutan package

GabrielBrascher (approved):
Changes make sense and are aligned with the API documentation and expected behavior.

Code LGTM.

harikrishna-patnala:
@GutoVeronezi There is a reason why we are generating a new password upon reset ssh key. 
We save the encrypted password (using ssh public key) of VM in the database. There is an API also which gets the VM password if it is saved "getVMPassword". If we don't reset the password upon reset ssh key then the saved password is CloudStack is no more valid and "getVMPassword" API will not serve its purpose.

harikrishna-patnala:
-1 from my side for now unless if there are plans to deprecate "getVMpassword" API and not to save VM password in database.

GutoVeronezi:
@harikrishna-patnala 

This behavior does not make any sense outside of the first password/resetting password API context. I see no problem in using the SSH key public key to encrypt the VM password, but not at this point.

Let's take a look at these cases:
   - A VM is created and a random password is generated. Users logins into their VM and changes their password, then they enable/set a new SSH key to the VM. So what now? They cannot log in anymore with the password had been using so far due to 'resetSSHKeyForVirtualMachine' changing the VM password without it being requested;
   - A VM is created with an SSH key and a random encrypted password is generated. The password of the VM is saved in the database. Users login into their VM and change its password. Now VM's password is different than what was saved in the database and `getVMPassword` will not serve its purpose too;

Users changing VM's password is a normal behavior (and they should do it). the API (`getVMPassword`) is not the source of truth for the VM password. It has its purpose in a specific life-cycle in ACS, which is the life-cyle of initial passwords and reset password processes. 

What is the problem if the API method `resetSSHKeyForVirtualMachine` stops doing that password-change? It will not break anything. It will not change anything for the end-user. The `getVMPassword` will keep returning the last password randomly generated by ACS, which is probably not valid anymore as expected because the user changed his/her password already in the operating system of the VM.

DaanHoogland:
I agree with @GutoVeronezi on this, @harikrishna-patnala can you explain the ratio behind your -1. getVMpassword is only for initial pw change anyway, it should not give anybody any value after that, but is vital at that moment. I don't see why deprecating it either.

harikrishna-patnala:
I'm talking about my concern from the management server perspective. I thought of the following use case in which fetching first VM password will also not be useful if initial SSH key is lost.
1. Deploy VM with password enabled template
2. User got the password X. Management server saved the Encrypted password with Publickkey1 P1(X)
3. Reset the SSH key (dont reset the password) to Publickey2 P2
4. User forgot the password and tries to getVMPassword
5. API will return encrypted password which user cannot decrypt if initial ssh key is not there.

I got your use cases as well where you are saying when user has changed the password. I'm speaking when the cases where the management server generated password is still in use and since getVMpassword API is available.

So overall I see it here there are two sides of the issue where we have to take a stand. I'm fine with the PR if everyone agrees that getVMPassword should not be taken into consideration always and it is not the source of truth.

The reason why I'm speaking about deprecating getVMPassword because resetVMpassword makes more sense in those usecases. resetVMpassword will always gives a new password and does not have any limitations like getVMpassword which is not true in many use cases as you mentioned.

DaanHoogland:
Ok, I see your point @harikrishna-patnala. I think we should address this by means of a limitation of warranty on the API description. deprecating it is not needed. In addition to that any reset of the ssh key should than not reset but just delete the old PW. A subsequent call to the get-API could return an error message with the proper instructions if the password is in-decyferable or empty.

harikrishna-patnala:
Thanks @DaanHoogland, deleting of old password from the database upon reset of ssh key seems to be good step here satisfying all usecases.

DaanHoogland:
ok @GutoVeronezi @harikrishna-patnala @GabrielBrascher , can we do this in a new PR or do we need to add that here? I'm not sure off the top of my head what an empty PW does, but functionally this is the right way to go, i think.

GutoVeronezi:
@DaanHoogland @harikrishna-patnala @GabrielBrascher as suggested, now `resetSSHKeyForVirtualMachine` removes the encrypted password. I did a few tests with VMs and routers and catch no problem. 

Also, I did a little refactor on the `getVMPassword` API, focusing in improving logs.

GutoVeronezi:
@DaanHoogland @harikrishna-patnala @GabrielBrascher any comment about the recent changes?

harikrishna-patnala (approved):
Thanks for the changes @GutoVeronezi. LGTM

sureshanaparti (approved):

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan test

GutoVeronezi:
@harikrishna-patnala @GabrielBrascher @sureshanaparti @DaanHoogland is there anything else to do?

DaanHoogland:
@GutoVeronezi , I don't think so. @harikrishna-patnala , @GabrielBrascher are we ready to merge this?

GabrielBrascher:
@GutoVeronezi @DaanHoogland @harikrishna-patnala I am OK with this PR as it is, LGTM.

rhtyd:
Let me review a few test cases for potential regression and get back with the review in a few days. Was config drive tested as well?

@blueorangutan package

GutoVeronezi:
> Let me review a few test cases for potential regression and get back with the review in a few days. Was config drive tested as well?
> 
> @blueorangutan package

@rhtyd if by config driver you mean CloudInit; then yes, it has been tested locally in a test lab, with CloudInit enabled.

1. I created VM with a template with `Password Enabled` as true.
2. I logged in the VM with the generated password on creation..
3. I stopped the VM and called the API.
4. I started the VM and tried to login with the password. Before this changes, the password would be changed and I could not login to the VM because I did not have the password anymore. After this changes, I could login normally with both password and ssh key certificate.

rhtyd:
By config drive I mean yes cloud-init but enabling config drive by creating a new network offering @GutoVeronezi and checking with/without cloud-init if the the behaviour with this PR matches expectation on say SSH reset operation.

GutoVeronezi:
@rhtyd I created a network offering with user-data enabled and created a VM with `echo "test" > /etc/file-with-user-data` in user-data. The file was created and resetSSH worked as expected.

I do not see how this change can affect config-drive workflow. Do you have any situation?

rhtyd:
That sounds good to me. Let's merge this given tests have passed (failure intermittent in nature, not related to this PR) and Hari, Gabriel, Suresh has LGTM'd.

rhtyd:
@GutoVeronezi one tip - if you write the description in your commit message while doing git commit, it will automatically show up in the Github PR description and also helps when merging the PR (we won't need to copy/paste the PR description in the merge-commit msg).

PR #4828, opened 2021-03-16
Feat/global vm search

DK101010:
### Description

This PR is more an POC and I need some opinions.

Background:
Many of my admins have the problem that they are get an vm name with a problem but not the related project. In this situation they must check all project to find them. Currently they are using a script to get all vm's with related projects, but they want also a opportunity in the ui. 

For this reason I build up a new command class as a first step and after this I want to create a search field in the ui to see a list of vm names and the related project names. I'm continue thought me that the admin can click on the vm in this list to jump into the vm details or he can click on the project to see the project list view with all vms.

========
The api call works and I get all VM's independent of a certian project . But i have headache with some points.

1. Currently I get to much informations of the vm and I have concerns that I break the permission concept of cloudstack.
The only things that I have found is to set the ReponseView to restricted and the details != all. 

2. When I call this api call via cloudmonkey, I get the correct and valid json but I get also this error "on parsing and printing list index out of range". Not sure if it a bug of cloudmonkey or a misconfiguration of my ReponseView

3. Not sure which package the right one is to store the new command. 

  

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial


davidjumani:
@DK101010 It should be possible to find the VM across any project by passing projectid=-1 in the request. It will be simpler to just tweak the UI to do the same in your environment

DK101010:
> @DK101010 It should be possible to find the VM across any project by passing projectid=-1 in the listVirtualMachines API request. It will be simpler to just tweak the UI to do the same in your environment
> @rhtyd Would it be a good idea to add a new filter in the advance search in the UI to search across projects ?

Hi @davidjumani, I had tested your suggestion, it doesn't work . (list virtualmachines name=VM projectid=-1)

davidjumani:
@DK101010 May I know the version you are using? It works in 4.15. Also does the account making the API call have access to all the VMs ?

DK101010:
> @DK101010 May I know the version you are using? It works in 4.15. Also does the account making the API call have access to all the VMs ?

I use also the 4.15 and I have different accounts with one project in each account. Hmmm then is your suggestion not that what I want. I will see all VM that are match with my search string

davidjumani:
@DK101010 If the account doesn't have access to those VMs, then creating a new API to fetch all VMs will cause a security issue as one account should not be able to view the resources of another account unless it is an admin or explicitly granted access. It will be better to have a root admin use the list virtualmachiens to filter out the VMs across projects

DK101010:
> @DK101010 If the account doesn't have access to those VMs, then creating a new API to fetch all VMs will cause a security issue as one account should not be able to view the resources of another account unless it is an admin or explicitly granted access. It will be better to have a root admin use the list virtualmachiens to filter out the VMs across projects

@davidjumani I already use  a root admin 
image
I use the same account/user in cloudmonkey.
Or do you mean I should adapt the implementation of listVirtualMaschine ?

davidjumani:
@DK101010 As admin, It should work and display VMs in all projects if projectid=-1 and listall=true
Could you try executing this ?
```
list virtualmachines filter=name,project projectid=-1 listall=true
```

DK101010:
> @DK101010 As admin, It should work and display VMs in all projects if projectid=-1 and listall=true
> Could you try executing this as admin ?
> 
> ```
> cmk list accounts # should return admin
> cmk list virtualmachines filter=name,project projectid=-1 listall=true
> ```

with listall=true I get all machines ... nice. Btw. I have also tested this in the prod env(4.13) I works also well, without listall. OK so my commits are  unavailing :(. No matter, as next is the question how to add this to the ui. As a Filter option or in a separated search field? In my opinion, it could be confusing to use a search field in the project view to find object outside this project.

davidjumani:
@DK101010 I think it would be fine as an additional field in the search filters, visible only to root admins
Can be done by adding it to https://github.com/apache/cloudstack/blob/master/ui/src/components/view/SearchView.vue
cc @rhtyd

Screenshot from 2021-03-17 17-02-18

DK101010:
> @DK101010 I think it would be fine as an additional field in the search filters, visible only to root admins
> Can be done by adding it to https://github.com/apache/cloudstack/blob/master/ui/src/components/view/SearchView.vue
> cc @rhtyd
> 
> Screenshot from 2021-03-17 17-02-18

Hi @davidjumani, I checked this file but I have no experience with the ui and I'm a little bit overwhelmed :D. Can you give me a hint where I can find the list of search filter elements ? Are There constants, because I can not find the call to the backend.

davidjumani:
@DK101010 The list of search filtres can be found here https://github.com/apache/cloudstack/blob/master/ui/src/config/section/compute.js#L70. A new one can be added by appending it to the list
Next you'll need to add a switch in SearchView similar to https://github.com/apache/cloudstack/blob/master/ui/src/components/view/SearchView.vue#L49 and ensure that when it's checked, it passes -1 as the projectid

DK101010:
> @DK101010 The list of search filtres can be found here https://github.com/apache/cloudstack/blob/master/ui/src/config/section/compute.js#L70. A new one can be added by appending it to the list
> Next you'll need to add a switch in SearchView similar to https://github.com/apache/cloudstack/blob/master/ui/src/components/view/SearchView.vue#L49 and ensure that when it's checked, it passes -1 as the projectid

Hi @davidjumani, thanks for your hints, it was really helpful.  I'm implement it this and it works but I have trouble with some points. 

First the label is not right and I don't know why.
image

As second when I search with accountwide = true and search again with accountwide=false then looks my url like 
vm?page=1&pagesize=10&accountwide=false
For me it looks not so good because accountwide = projectid . But I don't know if it important.

Also I get from time to time an error that the checkbox object haven't a value attribute and I don't know where it comes from.

davidjumani:
@DK101010 So to fix the label, you need to add an entry in en.json for `label.accountwide`, however it would be better if it were `projectwide` instead, similar to `"label.projectwide": "Search Across Projects"`
Could you send a screenshot of the error ?

DK101010:
> @DK101010 So to fix the label, you need to add an entry in en.json for `label.accountwide`, however it would be better if it were `projectwide` instead, similar to `"label.projectwide": "Search Across Projects"`
> Could you send a screenshot of the error ?

Hi @davidjumani, here the screenshot
image

Sure, I'm rename this.

davidjumani:
@DK101010 That shouldn't be an issue

davidjumani:
@blueorangutan ui

ui/src/components/view/SearchView.vue
```diff
@@ -211,6 +214,9 @@ export default {
         } else if (item === 'tags') {
           type = 'tag'
         }
+	if (item === 'projectwide') {
```
davidjumani, line 217: @DK101010 Can you fix the indentation error

davidjumani:
@blueorangutan ui

ui/src/components/view/SearchView.vue
davidjumani:
```
/jenkins/workspace/acs-ui-builder/ui/src/components/view/SearchView.vue
  221:2  error  Mixed spaces and tabs  no-mixed-spaces-and-tabs
```

DK101010: @davidjumani sorry, github seems to be act strange with tabs and spaces. I've solved and checked with lint

DK101010:
@blueorangutan ui

davidjumani:
@DK101010 I think it would look better as a switch https://www.antdv.com/components/switch/
Otherwise LGTM

DK101010:
Unfortunately, I cannot create another zone for the test due to a lack of resources. Code LGTM, but I get this error with the switch types:
Warning: You cannot set a form field before rendering a field associated with the value. You can use `getFieldDecorator(id, options)` instead `v-decorator="[id, options]"` to register it before render.

DK101010:
@blueorangutan ui

PaulAngus:
I've just been playing with this.
I'm 60/40 that the on/off button should be a top-level thing rather than down in the filters.

image

*** But [scope creep alert] to be **_really_** useful, it would need to be available for all resources - volumes, networks, VRs...

mib1185:
@PaulAngus this PR might be super seeded by #5207 and those be obsolet, since with #5207 you will be able to see all resource from all projects 😉

PaulAngus:
> @PaulAngus this PR might be super seeded by #5207 and those be obsolet, since with #5207 you will be able to see all resource from all projects 😉

looks like #5207 is another way of doing it.
Arguably #5207 is unnecessary as the function is already available in the API (projectid=-1), It's purely a UI change to expose it in the GUI.
And an on/off switch in the UI is much more flexible too.

davidjumani:
@DK101010 The project wide switch does not reset when hitting the reset button in the searchview, can you have a look into it ?

DK101010:
> @DK101010 The project wide switch does not reset when hitting the reset button in the searchview, can you have a look into it ?

I'm a little surprised that we want to use the PR after all, but I'll look at the problem as soon as I have time.

DK101010:
Hi @davidjumani I'm fixed the reset problem but we have also a additional problem. When you click on a vm with a different project, volumes are not displayed.
image

I check this but perhaps you have an hint.

davidjumani:
@blueorangutan ui

nvazquez:
Hi @PaulAngus, there's currently one open issue (regression) to be fixed in this PR

ravening:
@DK101010 the ui functionality has to be implemented for other resources also. only vm's for all projects can be listed now but networks, disks, templates cant be searched project wide

DK101010:
> @DK101010 the ui functionality has to be implemented for other resources also. only vm's for all projects can be listed now but networks, disks, templates cant be searched project wide

I already suspected it as you closed your PR.  Can't say when I'll can do it. 

Would you like to join me?I think you have much more experience with the ui?

ravening:
> > @DK101010 the ui functionality has to be implemented for other resources also. only vm's for all projects can be listed now but networks, disks, templates cant be searched project wide
> 
> I already suspected it as you closed your PR. Can't say when I'll can do it.
> 
> Would you like to join me?I think you have much more experience with the ui?

sure

utchoang (commented):
After performing a search with the `Search Across Projects` checkbox, reopening form search, the checkbox is in an unchecked state
bugs

nvazquez:
Hi @DK101010 @utchoang can you please advise if this PR is ready or needs more work?

utchoang:
@nvazquez cc @DK101010  everything seems fine, but there is 1 problem I mentioned above but not solved yet.

DK101010:
> Hi @DK101010 @utchoang can you please advise if this PR is ready or needs more work?
@nvazquez 
Hi, japp, a couple of things still to do here. Next to this bug that @utchoang mentioned , we need to extend the global search for other resources like network and so on. 
I have currently no time for that, but @ravening wants to help to implement the logic for the other resources.  

Additional to these I found also a problem. Currently you can only use the global search in the default view. Therefore we need to check the project id or clear the id as soon the global search is in use

DaanHoogland:
@DK101010 can you look at the conflicts?

DaanHoogland:
@DK101010 I would like to release 4.18 this month  and you have conflicts. Can you have a look, or shall I move this to 4.19?

DaanHoogland:
@DK101010 I moved this to 4.19 as I know you are busy. Please move back if you think I am at err (and you can make it happen)

shwstppr:
@DK101010 can you please address the merge conflicts? Should we consider this for 4.19.0.0?

DK101010:
@shwstppr I'm sorry, but I don't have time for that at the moment. I would rather see the 4.20

shwstppr:
@DK101010 thanks for the update. Moving it to 4.20.0.0 milestone

DaanHoogland:
@DK101010 et all, closing this one as it has conflicts and is old. please update and reopen if you think it is still relevant

PR #4840, opened 2021-03-18
Remove the rule(s) validation with api names while importing a role

sureshanaparti:
### Description

This PR removes the rule(s) validation with api names while importing a role. This will be in sync with the current create role permission behavior.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

Manually tests:
- Import role with some random rule string

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/ApiServerService.java
+++ b/api/src/main/java/org/apache/cloudstack/api/ApiServerService.java
@@ -42,6 +42,4 @@ public interface ApiServerService {
     public String handleRequest(Map<String, Object[]> params, String responseType, StringBuilder auditTrailSb) throws ServerApiException;
 
     public Class<?> getCmdClass(String cmdName);
-
-    public boolean isValidApiName(String apiName);
 }
--- a/api/src/main/java/org/apache/cloudstack/api/command/admin/acl/ImportRoleCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/admin/acl/ImportRoleCmd.java
@@ -94,9 +94,6 @@ public class ImportRoleCmd extends RoleCmd {
             if (Strings.isNullOrEmpty(rule)) {
                 throw new ServerApiException(ApiErrorCode.PARAM_ERROR, "Empty rule provided in rules param");
             }
-            if (!rule.contains("*") && !_apiServer.isValidApiName(rule)) {
-                throw new ServerApiException(ApiErrorCode.PARAM_ERROR, "Invalid api name: " + rule + " provided in rules param");
-            }
             ruleDetails.put(ApiConstants.RULE, new Rule(rule));
 
             String permission = detail.get(ApiConstants.PERMISSION);
--- a/server/src/main/java/com/cloud/api/ApiServer.java
+++ b/server/src/main/java/com/cloud/api/ApiServer.java
@@ -1208,17 +1208,6 @@ public class ApiServer extends ManagerBase implements HttpRequestHandler, ApiSer
         }
     }
 
-    @Override
-    public boolean isValidApiName(String apiName) {
-        if (apiName == null || apiName.isEmpty())
-            return false;
-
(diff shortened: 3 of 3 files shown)


rhtyd (approved):

sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan test

DaanHoogland:
> This PR removes the rule(s) validation with api names while importing a role. This will be in sync with the current create role permission behavior.

Aren't we fixing the wrong thing here? I'd say add the validation on create role.

sureshanaparti:
> > This PR removes the rule(s) validation with api names while importing a role. This will be in sync with the current create role permission behavior.
> 
> Aren't we fixing the wrong thing here? I'd say add the validation on create role.

@DaanHoogland Agree with api validation part. The intent here, is to have the same behavior for rules added through import role and create role permission (where api names are not validated). Otherwise, create role permission can be extended to include the validation for consistency. Also, note that some cmds are added to the API lookup map only when the respective service/plugin is enabled (eg. cloudian connector, kubernetes service, solid fire APIs).

rhtyd:
@blueorangutan test

shwstppr (approved):
LGTM based on changes. Haven't tested

rhtyd:
@blueorangutan test

DaanHoogland:
> > > This PR removes the rule(s) validation with api names while importing a role. This will be in sync with the current create role permission behavior.
> > 
> > 
> > Aren't we fixing the wrong thing here? I'd say add the validation on create role.
> 
> @DaanHoogland Agree with api validation part. The intent here, is to have the same behavior for rules added through import role and create role permission (where api names are not validated). Otherwise, create role permission can be extended to include the validation for consistency. Also, note that some cmds are added to the API lookup map only when the respective service/plugin is enabled (eg. cloudian connector, kubernetes service, solid fire APIs).

Ok, @sureshanaparti My preference would be add the checking to the create method, but I won't block for that reason.

shwstppr:
@DaanHoogland 

> > > > This PR removes the rule(s) validation with api names while importing a role. This will be in sync with the current create role permission behavior.
> > > 
> > > 
> > > Aren't we fixing the wrong thing here? I'd say add the validation on create role.
> > 
> > 
> > @DaanHoogland Agree with api validation part. The intent here, is to have the same behavior for rules added through import role and create role permission (where api names are not validated). Otherwise, create role permission can be extended to include the validation for consistency. Also, note that some cmds are added to the API lookup map only when the respective service/plugin is enabled (eg. cloudian connector, kubernetes service, solid fire APIs).
> 
> Ok, @sureshanaparti My preference would be add the checking to the create method, but I won't block for that reason.

@DaanHoogland do you approve the changes then?

rhtyd:
ping @DaanHoogland are you LGTM with the PR?

DaanHoogland:
Said, I don't block this in favour of timeliness but:
<rant>
No, I think validation should be done both in this functionality and in the case of createRole.
I have the uneasy feeling this is the removal of a feature because it blocks a single user. I do not see how we can justify that. Is there a larger use-case of which this is part that justifies it? and can we solve that in an other way?

I think the validation should be more thorough rather than be removed (I.E. make sure a glob resolves to an API before allowing it)
</rant>

As said, I won't block. I don't have a real world example of a user/use-case to argue we should block this, so merge if you disagree.

rhtyd:
@DaanHoogland the justification is that when adding any rule, the rule string is not validated (i.e. check if the API exists in the system/plugins) which is because some APIs get activated via a global setting or on external jar/plugin. Therefore we must relax with the similar logic for the import role feature (that imports a csv with list of role rule strings, i.e. don't fail if an API doesn't exist in the system).

rhtyd:
Merging as Daan hasn't blocked, on tests and 2 lgtms.

PR #4843, opened 2021-03-19
ui: deployvm - Add option to stay on page

davidjumani:
### Description

Adds the option to stay on the deploy VM page

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

Screenshot from 2021-07-01 12-42-49


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan ui

utchoang (approved):
LGTM!

rhtyd:
-1 we need to discuss what's the best way to address this or if users would need this at all. If users need to deploy multiple VMs we can consider a different UX, or ideally users should use APIs/cmk for automation.

davidjumani:
@rhtyd This is intended for when a user might want to deploy multiple VMs with different templates, networks etc. This prevents the page from going back and keeps the form data so the user can change only what he wants

shwstppr (approved):
Not sure about the utility for a real user but the change work fine.
Useful for testing though. Maybe the option can be hidden by default and enabled on some config change.

DaanHoogland:
@rhtyd this is a good case for displaying the APIs being called during the dialog as cmk script!?

rhtyd:
@DaanHoogland sorry didn't get your comment,, what is cmk to do with the UI change?

DaanHoogland:
> @DaanHoogland sorry didn't get your comment,, what is cmk to do with the UI change?

It doesn't and it doesn't relate to this change. But this change description by @davidjumani is a good argument for having a cmk CLI description of what you do in the UI. This can be convenient for operators that are learning or setting up new environments.

rhtyd:
@davidjumani while I disagree with the change, I don't want to block if everyone agrees with the feature. Pl start a discussion on dev and users ML.

nvazquez:
Hi @davidjumani are you still working on this PR?

davidjumani:
@nvazquez It's ready but I'll send a mail on the ML to get consensus on  this PR

shwstppr:
@davidjumani just a though, will it better to tweak UI something like following to make the option bit more subtle?
Screenshot from 2021-06-25 15-38-16

davidjumani:
@blueorangutan ui

andrijapanicsb (approved):
Good change, someone will find this usable for sure.

LGTM

rhtyd:
This creates a slight confusion, I would normally want the drop down to select first and then click the button (examples in other forms such as send email in Gmail or outlook where you toggle/select between send now or later and then click the button later to send the email).

PR #4848, opened 2021-03-19
Allow host cert renewals even if client auth strictness is false

Slair1:
### Description

This PR allows the management servers to renew certificates issued to hosts, even if ca.plugin.root.auth.strictness is false.  There is already a global setting named ca.framework.cert.automatic.renewal that should be controlling the certificate renewals.  However, ca.plugin.root.auth.strictness currently circumvents that setting.

#### More Detail

The ca.plugin.root.auth.strictness setting is checked early in the client certificate checking method and exits the method without adding the certificate to the management server's in-memory certificate map.  The certificate renewal process (if enabled with ca.framework.cert.automatic.renewal) reads that in-memory certificate map to determine what certs it could renew.  Those two settings shouldn't be so closely related.

### Types of changes

- [ ] Major
- [X ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
We've deployed and tested this in our CloudStack environments.


Slair1:
Doing more testing before submitting

PR #4857, opened 2021-03-23
ui: Disable login button until redirected

davidjumani:
### Description

Disables the login button while discovering features to prevent a double login / repetitive API calls

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

#### Before :

Screenshot from 2021-03-23 15-31-48

#### After :

Screenshot from 2021-03-23 15-31-21

### DIFF
--- a/ui/src/views/auth/Login.vue
+++ b/ui/src/views/auth/Login.vue
@@ -189,8 +189,8 @@ export default {
             }
             Login(loginParams)
               .then((res) => this.loginSuccess(res))
-              .catch(err => this.requestFailed(err))
-              .finally(() => {
+              .catch(err => {
+                this.requestFailed(err)
                 state.loginBtn = false
               })
           } else if (customActiveKey === 'saml') {


davidjumani:
@blueorangutan ui

ui/src/views/auth/Login.vue
```diff
@@ -189,8 +189,8 @@ export default {
             }
             Login(loginParams)
               .then((res) => this.loginSuccess(res))
-              .catch(err => this.requestFailed(err))
-              .finally(() => {
+              .catch(err => {
+                this.requestFailed(err)
                 state.loginBtn = false
```
rhtyd, line 194: this would set loginBtn to false on error?

utchoang: @rhtyd I think it is correct because the goal is to only re-enable the button when it gets an error to perform click/submit again.

utchoang (approved):
LGTM!

ui/src/views/auth/Login.vue
davidjumani: What he said ^^

shwstppr (approved):

ui/src/views/auth/Login.vue
rhtyd: Thanks for explaining, I thought setting that to 'true' would mean enabling the button.

PR #4870, opened 2021-03-25
kvm: remove unnecessary new String

ustcweizhou:
### Description

Thanks @rubieHess  to point it out. 
see https://github.com/apache/cloudstack/pull/4800#discussion_r598318705
```
I detect that this code is problematic. According to the Performance (PERFORMANCE), Dm: Method invokes inefficient new String(String) constructor (DM_STRING_CTOR).
Using the java.lang.String(String) constructor wastes memory because the object so constructed will be functionally indistinguishable from the String passed as a parameter.  Just use the argument String directly.
```

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -4264,8 +4264,8 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
         QemuImg qemu = new QemuImg(timeout);
         try{
             Map<String, String> info = qemu.info(file);
-            String backingFilePath = info.get("backing_file");
-            String backingFileFormat = info.get("backing_file_format");
+            String backingFilePath = info.get(QemuImg.BACKING_FILE);
+            String backingFileFormat = info.get(QemuImg.BACKING_FILE_FORMAT);
             if (org.apache.commons.lang.StringUtils.isNotBlank(backingFilePath)
                     && org.apache.commons.lang.StringUtils.isBlank(backingFileFormat)) {
                 // VMs which are created in CloudStack 4.14 and before cannot be started or migrated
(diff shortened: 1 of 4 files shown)


GabrielBrascher (commented):
Thanks for this PR, @ustcweizhou. I am **+1** on removing the String constructors, they should be avoided indeed.

I would also suggest externalizing them to constants, what do you think?

In terms of performance, I think that it does not change much being a constant or not. The way you are doing the String object is created only once. For remaining iterations, the same object is returned from the String's constant pool.

However, constants help code maintainability and readability. Especially on the duplicated Strings where a developer can easily change all occurrences by simply altering the constant.

weizhouapache:
> Thanks for this PR, @ustcweizhou. I am **+1** on removing the String constructors, they should be avoided indeed.
> 
> I would also suggest externalizing them to constants, what do you think?
> 
> In terms of performance, I think that it does not change much being a constant or not. The way you are doing the String object is created only once. Java will detect the duplicated Strings in runtime and reuse them from the String's constant pool.
> 
> However, constants help code maintainability and readability. Especially on the duplicated Strings where a developer can easily change all occurrences by simply altering the constant.

@GabrielBrascher
agree. I will update this pr.

rhtyd (approved):

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland (approved):

DaanHoogland:
pl wait with merging @rhtyd , I think @weizhouapache wants to externalise those strings to constants as well.

Pearl1594:
@blueorangutan package

GabrielBrascher (approved):
Thanks for the updates, @ustcweizhou.
Code LGTM.

rhtyd:
@blueorangutan test

DaanHoogland:
great work once a gain @weizhouapache. ok to merge @rhtyd / @shwstppr ?

PR #4886, opened 2021-04-01
server: list routers by healthchecksfailed

ustcweizhou:
### Types of changes

- [ ] Major
- [X] Minor

### DIFF
--- a/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
+++ b/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
@@ -186,6 +186,7 @@ import com.cloud.hypervisor.Hypervisor.HypervisorType;
 import com.cloud.network.RouterHealthCheckResult;
 import com.cloud.network.VpcVirtualNetworkApplianceService;
 import com.cloud.network.dao.RouterHealthCheckResultDao;
+import com.cloud.network.dao.RouterHealthCheckResultVO;
 import com.cloud.network.router.VirtualNetworkApplianceManager;
 import com.cloud.network.security.SecurityGroupVMMapVO;
 import com.cloud.network.security.dao.SecurityGroupVMMapDao;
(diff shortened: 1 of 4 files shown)


DaanHoogland:
I like the functionality @weizhouapache , but by the look of the code there is still some wiring to do, is that correct?

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -1341,6 +1342,27 @@
             sb.and("networkId", sb.entity().getNetworkId(), SearchCriteria.Op.EQ);
         }
 
+        List<Long> routersWithFailures = null;
+        if (VirtualNetworkApplianceManager.RouterHealthChecksEnabled.value() && isHealthCheckFailed != null) {
```
sureshanaparti, line 1346: @ustcweizhou what if "_router.health.checks.enabled_" config is disabled and there are last health check reports ? Will the routers listed with healthchecksfailed=trur returns empty list or notifies health checks not enabled?

shwstppr:
@blueorangutan package

rhtyd:
@blueorangutan package

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
weizhouapache: @sureshanaparti if "router.health.checks.enabled" was "true" and then changed to "false", the health check results will be out of date and incorrect. In this case, do we need to consider the health check results ?


> @ustcweizhou what if "_router.health.checks.enabled_" config is disabled and there are last health check reports ? Will the routers listed with healthchecksfailed=true returns empty list or routers with last failed reports or notifies health checks are not enabled?

sureshanaparti: @ustcweizhou if the results are out of date/incorrect when queried for, i think better to perform health checks and update the results. (Note: if it is a sync call, then it can take time, so just notify the operator that results are out of date and perform fresh health checks before listing routers by health checks failed). See if any time duration can be set (through API or global config) to validate the results are outside of that duration or not.

weizhouapache: @sureshanaparti 
I think we need to make sure that the health check results are correct.
if "router.health.checks.enabled" is changed from "true" to "false", we'd better remove all existing health check results as they will not be updated any more. btw: can we perform health check via api or ui when router.health.checks.enabled" is 'false" ?

sureshanaparti: > @sureshanaparti
> I think we need to make sure that the health check results are correct.
> if "router.health.checks.enabled" is changed from "true" to "false", we'd better remove all existing health check results as they will not be updated any more. btw: can we perform health check via api or ui when router.health.checks.enabled" is 'false" ?

@weizhouapache i think, the existing health check results are still there after changing  "router.health.checks.enabled" from "true" to "false". May be this change can be checked in existing background thread and can cleanup the old health check results.

weizhouapache: @sureshanaparti yes. it is another story.
I removed the check on global setting "router.health.checks.enabled" in this pr.

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

server/src/main/java/com/cloud/api/query/QueryManagerImpl.java
```diff
@@ -1341,6 +1343,27 @@
...
+                if (isHealthCheckFailed) {
+                    sb.and("routerId", sb.entity().getId(), SearchCriteria.Op.IN);
+                } else {
+                    sb.and("routerId", sb.entity().getId(), SearchCriteria.Op.NIN);
+                }
+            } else if (isHealthCheckFailed) {
+                return new Pair<List<DomainRouterJoinVO>, Integer>(Collections.emptyList(), 0);
+            }
+        }
+
```
DaanHoogland, line 1366: @weizhouapache can you put this in a separate procudure, please?

DaanHoogland (approved):
clgtm

rhtyd (approved):

PR #4938, opened 2021-04-19
cloudian: Set cloudian.connector.enabled as not dynamic

davidjumani:
### Description

This PR sets cloudian.connector.enabled as not dynamic since cloudian requires the cloudianSsoLogin API which is added only once the MS is restarted

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### DIFF
--- a/plugins/integrations/cloudian/src/main/java/org/apache/cloudstack/cloudian/CloudianConnector.java
+++ b/plugins/integrations/cloudian/src/main/java/org/apache/cloudstack/cloudian/CloudianConnector.java
@@ -24,7 +24,7 @@ import com.cloud.utils.component.PluggableService;
 public interface CloudianConnector extends PluggableService {
 
     ConfigKey<Boolean> CloudianConnectorEnabled = new ConfigKey<>("Advanced", Boolean.class, "cloudian.connector.enabled", "false",
-            "If set to true, this enables the Cloudian Connector for CloudStack.", true);
+            "If set to true, this enables the Cloudian Connector for CloudStack.", false);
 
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

sureshanaparti (approved):
code LGTM

shwstppr (approved):

plugins/integrations/cloudian/src/main/java/org/apache/cloudstack/cloudian/CloudianConnector.java
```diff
@@ -24,7 +24,7 @@
 public interface CloudianConnector extends PluggableService {
 
     ConfigKey<Boolean> CloudianConnectorEnabled = new ConfigKey<>("Advanced", Boolean.class, "cloudian.connector.enabled", "false",
-            "If set to true, this enables the Cloudian Connector for CloudStack.", true);
+            "If set to true, this enables the Cloudian Connector for CloudStack.", false);
```
rhtyd, line 27: @davidjumani this won't work - maybe for UI but for the plugin to instantiate - you'll need to restart mgmt server, as it's used in the `configure` method of manager/class

davidjumani: Exactly why it's being set as non dynamic. The user will get a notification that the MS needs to be restarted after this change and on restart it'll pick up the necessary API

rhtyd: Ah I see, my bad I misunderstood - lgtm.

rhtyd (approved):

Issue #4844, opened 2021-03-19
UI should show IPv6 addresses of Instance


wido:
##### ISSUE TYPE
 * Enhancement Request

##### COMPONENT NAME
~~~
UI
~~~

##### CLOUDSTACK VERSION
~~~
4.15
master
~~~

##### SUMMARY
The new UI does not show IPv6 addresses of VMs and/or NICs.

##### ACTUAL RESULTS
I have taken a few screenshots from the new UI on a deployment which has IPv6 enabled Virtual Machines with VXLAN in a Shared Network.

<img width="2556" alt="Screenshot 2021-03-19 at 11 17 54" src="https://user-images.githubusercontent.com/326786/111765943-3d0a5200-88a5-11eb-9ed2-c99d48d6df10.png">

abhi005:
Hey @DaanHoogland I would like to work on this issue. As this is one of the mandatory requirements for my course work, will make sure that it gets completed.

DaanHoogland:
@abhi005 , I see @rhtyd has assigned this to @utchoang in the meanwhile so be sure to communicate the best solution amongst yourselves.

abhi005:
Cool, So just to be sure requirement is to show only IPv6 on view page and not both. correct? @DaanHoogland @rhtyd

wido:
> Cool, So just to be sure requirement is to show only IPv6 on view page and not both. correct? @DaanHoogland @rhtyd

I would say that we should display the IPv6 address of a NIC/Instance on any page where we also show the IPv4 address.

abhi005:
Ok thanks for the clarification, I have started with the local setup and will start making changes asap.

DaanHoogland:
> Cool, So just to be sure requirement is to show only IPv6 on view page and not both. correct? @DaanHoogland @rhtyd

No @abhi005 I think both should be shown if available.

abhi005:
I think showing both of them would be confusing for user and not very asthetic from UI point of view. Otherwise we can tweak UI little bit, to accomodate both the options.

DaanHoogland:
You could be right @abhi005, I'm not a UI expert (other than for making angry fruit salads). As an operator I want to be able to show both though.
what do you think @wido, @utchoang, @rhtyd ?

wido:
> I think showing both of them would be confusing for user and not very asthetic from UI point of view. Otherwise we can tweak UI little bit, to accomodate both the options. @DaanHoogland

[...]

PR #4948, opened 2021-04-26
UI: Show IPv6 address of Instance

utchoang:
### Description

Fixes https://github.com/apache/cloudstack/issues/4844

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

image

### DIFF
--- a/ui/src/components/view/DetailsTab.vue
+++ b/ui/src/components/view/DetailsTab.vue
@@ -50,6 +50,13 @@
         </div>
       </div>
     </a-list-item>
+    <a-list-item slot="renderItem" slot-scope="item" v-else-if="item === 'ip6address' && ipV6Address.length > 0">
+      <div>
+        <strong>{{ $t('label.' + String(item).toLowerCase()) }}</strong>
+        <br/>
+        <div>{{ ipV6Address }}</div>
+      </div>
+    </a-list-item>
     <HostInfo :resource="resource" v-if="$route.meta.name === 'host' && 'listHosts' in $store.getters.apis" />
     <DedicateData :resource="resource" v-if="dedicatedSectionActive" />
     <VmwareData :resource="resource" v-if="$route.meta.name === 'zone' && 'listVmwareDcs' in $store.getters.apis" />
@@ -87,6 +94,15 @@ export default {
   },
   mounted () {
     this.dedicatedSectionActive = this.dedicatedRoutes.includes(this.$route.meta.name)
+  },
+  computed: {
+    ipV6Address () {
+      if (this.resource.nic && this.resource.nic.length > 0) {
+        return this.resource.nic.filter(e => { return e.ip6address }).map(e => { return e.ip6address }).join(', ')
+      }
+
+      return null
+    }
   },
   created () {
     this.dedicatedSectionActive = this.dedicatedRoutes.includes(this.$route.meta.name)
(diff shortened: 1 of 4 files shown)


utchoang:
@blueorangutan ui

utchoang:
@blueorangutan ui

davidjumani:
@utchoang @rhtyd Would it be a good idea to include the ip6 address in the listview as well ?

utchoang:
> @utchoang @rhtyd Would it be a good idea to include the ip6 address in the listview as well ?
@davidjumani @rhtyd  I think it's fine, I'll add it to the listview

utchoang:
@blueorangutan ui

utchoang:
@blueorangutan ui

rhtyd (approved):
LGTM - didn't test, don't have env with ipv6

wido (approved):
LGTM based on code and screenshots

kricud:
If possible please show ipv6 link-local address as well. The one that is in range/block fe80::/10
If you have two instances on same network/subnet/broadcast domain and they both have fe80:xxx they will use it to communicate each other.

https://en.wikipedia.org/wiki/Link-local_address

davidjumani (approved):
LGTM based on code and screenshots

rhtyd:
@kricud are you asking the ipv6 link-local address to be shown for systemvms or user VMs? Is that nic information something that gets added internal to the VM or plugged by CloudStack? (I would image all nics/ipv4/ipv6 addresses to be returned by the API to be shown in the UI; any custom nics/setup inside the VM may not be discovered by CloudStack)

kricud:
@rhtyd In case of ipv6 you will always have link-local address even if routable ipv6 is not configured. In screenshot I do not see it and its misleading. As you wrote all ip addresses must be presented.

rhtyd:
@kricud as long as those are internal to the VM, it will be difficult for CloudStack to report those IPs (for example a custom network over L2 network, there's no IP known to CloudStack). We only see all the IPs for all the nics configured for a VM by CloudStack. As you've said, showing link-local/lo IPs won't be useful for end-users. Thanks.

wido:
> @rhtyd In case of ipv6 you will always have link-local address even if routable ipv6 is not configured. In screenshot I do not see it and its misleading. As you wrote all ip addresses must be presented.
> 
> Ignore- I mis understood its not doable and user must understand that ipv6 will have link-local.
> Configuring it manually for presentation in ui is nonsense.

I know it has been merged now, but let me add this:

The link-local address is a non-routable address which can't be used from the outside world. Nor does CloudStack know what it is allthough it could calculate the address based on the MAC.

CloudStack is aimed at routable addresses as that's what we use. Technically servers can communicate with each other over Link-Local as long as they are in the same L2 domain, but it's not the use-case which is used very much.

Most services use unicast addresses for the communication.

LL with IPv6 is mainly used for Router Advertisements, Neighbor Discovery, etc, etc.

I would not say it's very useful to display the LL addresses in the UI. In addition: The API does not return those.

rhtyd:
@wido can you create a new issue tracking the new feature asked, you're right I checked the algorithm the LL address is generated using Mac address and it's certainly possible to display it in UI as mac address is already returned in API response.

PR #4955, opened 2021-04-26
Improve logs on SecondaryStorageManagerImpl and few refactors

GutoVeronezi:
### Description
This PR intends to improve logging in the class `SecondaryStorageManagerImpl` to facilitate troubleshooting and refactor a few process.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

### DIFF
--- a/services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
+++ b/services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
@@ -170,12 +170,9 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
         ResourceStateAdapter, Configurable {
     private static final Logger s_logger = Logger.getLogger(SecondaryStorageManagerImpl.class);
 
-    private static final int DEFAULT_CAPACITY_SCAN_INTERVAL = 30000; // 30
-    // seconds
-    private static final int ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_SYNC = 180; // 3
-    // minutes
-
-    private static final int STARTUP_DELAY = 60000; // 60 seconds
+    private static final int DEFAULT_CAPACITY_SCAN_INTERVAL = 30000;
+    private static final int ACQUIRE_GLOBAL_LOCK_TIMEOUT_FOR_SYNC = 180;
+    private static final int STARTUP_DELAY = 60000;
 
     private int _mgmtPort = 8250;
 
@@ -259,7 +256,7 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
     protected long _nodeId = ManagementServerNode.getManagementServerId();
 
     private SystemVmLoadScanner<Long> _loadScanner;
-    private Map<Long, ZoneHostInfo> _zoneHostInfoMap; // map <zone id, info about running host in zone>
+    private Map<Long, ZoneHostInfo> _zoneHostInfoMap;
 
     private final GlobalLock _allocLock = GlobalLock.getInternLock(getAllocLockName());
 
@@ -307,7 +304,7 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
             List<DataStore> ssStores = _dataStoreMgr.getImageStoresByScope(new ZoneScope(zoneId));
             for (DataStore ssStore : ssStores) {
                 if (!(ssStore.getTO() instanceof NfsTO)) {
-                    continue; // only do this for Nfs
+                    continue;
                 }
                 String secUrl = ssStore.getUri();
                 SecStorageSetupCommand setupCmd = null;
@@ -321,7 +318,6 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
                 String nfsVersion = imageStoreDetailsUtil.getNfsVersion(ssStore.getId());
                 setupCmd.setNfsVersion(nfsVersion);
 
-                //template/volume file upload key
                 String postUploadKey = _configDao.getValue(Config.SSVMPSK.key());
                 setupCmd.setPostUploadKey(postUploadKey);
 
@@ -331,7 +327,6 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
                 if (answer != null && answer.getResult()) {
                     SecStorageSetupAnswer an = (SecStorageSetupAnswer)answer;
                     if (an.get_dir() != null) {
-                        // update the parent path in image_store table for this image store
                         ImageStoreVO svo = _imageStoreDao.findById(ssStore.getId());
                         svo.setParent(an.get_dir());
                         _imageStoreDao.update(ssStore.getId(), svo);
@@ -344,27 +339,7 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
                 }
             }
         }
-        /* After removing SecondaryStorage entries from host table, control should never come here!!
-        else if( cssHost.getType() == Host.Type.SecondaryStorage ) {
-            List<SecondaryStorageVmVO> alreadyRunning = _secStorageVmDao.getSecStorageVmListInStates(SecondaryStorageVm.Role.templateProcessor, zoneId, State.Running);
-            String secUrl = cssHost.getStorageUrl();
-            SecStorageSetupCommand setupCmd = new SecStorageSetupCommand(secUrl, null);
-            for ( SecondaryStorageVmVO ssVm : alreadyRunning ) {
-                HostVO host = _resourceMgr.findHostByName(ssVm.getInstanceName());
-                Answer answer = _agentMgr.easySend(host.getId(), setupCmd);
-                if (answer != null && answer.getResult()) {
-                    if (s_logger.isDebugEnabled()) {
-                        s_logger.debug("Successfully programmed secondary storage " + host.getName() + " in secondary storage VM " + ssVm.getInstanceName());
-                    }
-                } else {
-                    if (s_logger.isDebugEnabled()) {
-                        s_logger.debug("Successfully programmed secondary storage " + host.getName() + " in secondary storage VM " + ssVm.getInstanceName());
-                    }
-                    return false;
-                }
-            }
-        }
-         */
+
         return true;
     }
 
@@ -514,8 +489,7 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
         }
 
         SecondaryStorageVmVO secStorageVm = _secStorageVmDao.findById(secStorageVmId);
-        // SecondaryStorageVmVO secStorageVm =
-        // allocSecStorageVmStorage(dataCenterId, secStorageVmId);
+
         if (secStorageVm != null) {
             SubscriptionMgr.getInstance().notifySubscribers(ALERT_SUBJECT, this,
                 new SecStorageVmAlertEventArgs(SecStorageVmAlertEventArgs.SSVM_CREATED, dataCenterId, secStorageVmId, secStorageVm, null));
@@ -569,7 +543,7 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
         else {
             TrafficType defaultTrafficType = TrafficType.Public;
             List<NetworkVO> defaultNetworks = _networkDao.listByZoneAndTrafficType(dc.getId(), defaultTrafficType);
-            // api should never allow this situation to happen
+
             int networksSize = defaultNetworks.size();
             if (networksSize != 1) {
                 throw new CloudRuntimeException(String.format("Found [%s] networks of type [%s] when expect to find 1.", networksSize, defaultTrafficType));
@@ -593,7 +567,7 @@ public class SecondaryStorageManagerImpl extends ManagerBase implements Secondar
 
         TrafficType defaultTrafficType = TrafficType.Guest;
         List<NetworkVO> defaultNetworks = _networkDao.listByZoneAndTrafficType(dc.getId(), defaultTrafficType);
-        // api should never allow this situation to happen
+
         int networksSize = defaultNetworks.size();
         if (networksSize != 1) {
             throw new CloudRuntimeException(String.format("Found [%s] networks of type [%s] when expect to find 1.", networksSize, defaultTrafficType));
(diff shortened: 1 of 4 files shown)


shwstppr:
@blueorangutan package

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -328,51 +318,28 @@ public boolean generateSetupCommand(Long ssHostId) {
...
-                    if (s_logger.isDebugEnabled()) {
-                        s_logger.debug("Successfully programmed secondary storage " + host.getName() + " in secondary storage VM " + ssVm.getInstanceName());
-                    }
+
+                    s_logger.debug(debugMessage);
                 } else {
-                    if (s_logger.isDebugEnabled()) {
-                        s_logger.debug("Successfully programmed secondary storage " + host.getName() + " in secondary storage VM " + ssVm.getInstanceName());
-                    }
+                    s_logger.debug(debugMessage);
```
sureshanaparti, line 337: seems the logs msg here is not correct. method returns false when there is no answer and still logging "_Successfully programmed secondary storage..._" , check and correct log here.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -686,50 +656,41 @@ protected String connect(String ipAddress, int port) {
...
-            s_logger.trace("Assign  secondary storage vm from running pool for request from data center : " + dataCenterId);
-        }
+        s_logger.debug(String.format("Assign secondary storage VM from running pool for request from zone [%s].", dataCenterId));
 
         SecondaryStorageVmAllocator allocator = getCurrentAllocator();
         assert (allocator != null);
         List<SecondaryStorageVmVO> runningList = _secStorageVmDao.getSecStorageVmListInStates(role, dataCenterId, State.Running);
-        if (runningList != null && runningList.size() > 0) {
-            if (s_logger.isTraceEnabled()) {
-                s_logger.trace("Running secondary storage vm pool size : " + runningList.size());
```
sureshanaparti, line 699: any particular reason to move trace level logs to debug

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -1272,22 +1216,20 @@ public boolean finalizeCommandsOnStart(Commands cmds, VirtualMachineProfile prof
     public boolean finalizeStart(VirtualMachineProfile profile, long hostId, Commands cmds, ReservationContext context) {
         CheckSshAnswer answer = (CheckSshAnswer)cmds.getAnswer("checkSsh");
         if (!answer.getResult()) {
-            s_logger.warn("Unable to ssh to the VM: " + answer.getDetails());
+            s_logger.warn(String.format("Unable to connect via SSH to the VM [%s] due to [%s] ", profile.toString(), answer.getDetails()));
             return false;
         }
 
         try {
-            //get system ip and create static nat rule for the vm in case of basic networking with EIP/ELB
```
sureshanaparti, line 1280: Observed that all the comments are removed. I think these comments were added to understand the code there, and would help in debugging sometimes. You can remove them judiciously.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: About changing the log level to debug, in this context, these messages work better as debug; as well, it is quite unusual to call trace on ACS, so, it is interesting to keep a pattern and use debug.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: @sureshanaparti done. Thanks!

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: Comments were removed because they are nonsense. Most of they only describe what the next line is doing, e.q.: 
```java
//verify if var is true then print it in console
if (var) {
	System.out.println("var is true");
}
```

These types of comments do not improve debugging and goes against a clean code. However, there are comments that explain a bit of the process, but they could be resolved with an extraction of the code to a method with a corresponding name, javadoc and etc.

Also, there are `TODO` comments. These requests should be regularized in issues, not in code.

Comments should be a last resort, not the rule. If we take all the comments and sort then in a logic way, we could write a book :smile: .

If there are some points you really think that should be kept, let me know, and then we can discuss if it is important to maintain it; or, maybe, we are able to get ride of it by executing some minor refactoring.

DaanHoogland: I agree;

1. comment last resort
2. (replace by) method name first resort

in this case I only wonder about the `EIP/ELB` part of the comment, so for me it was only adding confusion.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -1251,13 +1194,12 @@ public boolean finalizeCommandsOnStart(Commands cmds, VirtualMachineProfile prof
...
         if (controlNic == null) {
             if (managementNic == null) {
-                s_logger.error("Management network doesn't exist for the secondaryStorageVm " + profile.getVirtualMachine());
+                s_logger.warn(String.format("Management network does not exist for the secondary storage %s.", profile. toString()));
                 return false;
             }
             controlNic = managementNic;
         }
 
-        // verify ssh access on management nic for system vm running on HyperV
```
DaanHoogland, line 1260: how about making this a private method name
```suggestion
        verifySshAccessOnManagementNicForSystemvmRunningOnHyperV(profile, controlNic, managementNic, cmds);
        return true;
    }

    private void verifySshAccessOnManagementNicForSystemvmRunningOnHyperV(VirtualMachineProfile profile,
    NicProfile controlNic,
    NicProfile managementNic,
    Commands cmds) {
        if(profile.getHypervisorType() == HypervisorType.Hyperv) {
            controlNic = managementNic;
        }
        CheckSshCommand check = new CheckSshCommand(profile.getInstanceName(),controlNic.getIPv4Address(), 3922);
        cmds.addCommand("checkSsh", check);
    }
```
context: see comment removal below.

DaanHoogland (commented):
code looks good to me, some discussion points added

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -382,15 +347,16 @@ public boolean generateVMSetupCommand(Long ssAHostId) {
         if (ssAHost.getType() != Host.Type.SecondaryStorageVM) {
             return false;
         }
-        SecondaryStorageVmVO secStorageVm = _secStorageVmDao.findByInstanceName(ssAHost.getName());
+        String hostName = ssAHost.getName();
+        SecondaryStorageVmVO secStorageVm = _secStorageVmDao.findByInstanceName(hostName);
         if (secStorageVm == null) {
-            s_logger.warn("secondary storage VM " + ssAHost.getName() + " doesn't exist");
+            s_logger.warn(String.format("Secondary storage VM [%s] does not exist.", hostName));
```
GabrielBrascher, line 353: I would rename the variable from `hostName` to something that fits better on the method context.
Example: `ssvmName`.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -402,15 +368,16 @@ public boolean generateVMSetupCommand(Long ssAHostId) {
         String copyPasswd = _configDao.getValue("secstorage.copy.password");
         setupCmd.setCopyPassword(copyPasswd);
         setupCmd.setCopyUserName(TemplateConstants.DEFAULT_HTTP_AUTH_USER);
+
         Answer answer = _agentMgr.easySend(ssAHostId, setupCmd);
         if (answer != null && answer.getResult()) {
             if (s_logger.isDebugEnabled()) {
-                s_logger.debug("Successfully programmed http auth into " + secStorageVm.getHostName());
+                s_logger.debug(String.format("Successfully created http auth into secondary storage VM [%s].", hostName));
```
GabrielBrascher, line 375: IMHO `programmed` can be indeed changed to something that best describes the logged operation (set the allowed internal sites, username, and password).

`Created` is an option, but I think that `set` fits better.

I would suggest changing from: `Successfully created http auth into secondary storage VM [s-123-VM]`
To: `Successfully set HTTP auth into secondary storage VM [s-123-VM]`

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: @DaanHoogland done. As the comment refers only to `if...` I thought it would be better to only extract it. Also, I extracted as protected and added unit tests to it.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: @GabrielBrascher done.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: @GabrielBrascher done.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
DaanHoogland: tests are always good.
As for the `if...` comment, maybe the lengthy name was not my best effort ever and `verifySshAccessOnManagementNicForSystemvm()` would have been better.

DaanHoogland:
> Trillian Build Failed (tid-578)

Our lab is too busy to tests these now. I'll find a timeslot when i do.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GabrielBrascher: Thanks @GutoVeronezi!

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
```diff
@@ -816,37 +773,33 @@ public boolean isZoneReady(Map<Long, ZoneHostInfo> zoneHostInfoMap, long dataCen
...
 
-            boolean useLocalStorage = false;
-            Boolean useLocal = ConfigurationManagerImpl.SystemVMUseLocalStorage.valueIn(dataCenterId);
-            if (useLocal != null) {
-                useLocalStorage = useLocal.booleanValue();
-            }
+            boolean useLocalStorage = BooleanUtils.toBoolean(ConfigurationManagerImpl.SystemVMUseLocalStorage.valueIn(dataCenterId));
             List<Pair<Long, Integer>> l = _storagePoolHostDao.getDatacenterStoragePoolHostInfo(dataCenterId, !useLocalStorage);
-            if (l != null && l.size() > 0 && l.get(0).second().intValue() > 0) {
+            if (CollectionUtils.isNotEmpty(l) && l.get(0).second() > 0) {
```
GabrielBrascher, line 796: The changes are really nice, good work. I would also suggest renaming these lists from `l` to something that actually means something.
Example:

1. From
```
List<Pair<Long, Integer>> l = _storagePoolHostDao.getDatacenterStoragePoolHostInfo(dataCenterId, !useLocalStorage);
if (CollectionUtils.isNotEmpty(l) && l.get(0).second() > 0) {
...
}
```

2. To
```
List<Pair<Long, Integer>> storagePoolHostInfos = _storagePoolHostDao.getDatacenterStoragePoolHostInfo(dataCenterId, !useLocalStorage);
if (CollectionUtils.isNotEmpty(storagePoolHostInfos) && storagePoolHostInfos.get(0).second() > 0) {
...
}
```

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: @DaanHoogland done.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
GutoVeronezi: @GabrielBrascher done.

services/secondary-storage/controller/src/main/java/org/apache/cloudstack/secondarystorage/SecondaryStorageManagerImpl.java
sureshanaparti: > Comments were removed because they are nonsense. Most of they only describe what the next line is doing, e.q.:
> 
> ```java
> //verify if var is true then print it in console
> if (var) {
> 	System.out.println("var is true");
> }
> ```
> 
> These types of comments do not improve debugging and goes against a clean code. However, there are comments that explain a bit of the process, but they could be resolved with an extraction of the code to a method with a corresponding name, javadoc and etc.
> 
> Also, there are `TODO` comments. These requests should be regularized in issues, not in code.
> 
> Comments should be a last resort, not the rule. If we take all the comments and sort then in a logic way, we could write a book 😄 .
> 
> If there are some points you really think that should be kept, let me know, and then we can discuss if it is important to maintain it; or, maybe, we are able to get ride of it by executing some minor refactoring.

@GutoVeronezi you can remove such nonsense comments, even i'm not in favour of such comments. i cannot check and advise on every comment. you can take a judicious call while refactoring them. thanks.

PR #4966, opened 2021-04-28
Refactor few process of VirtualMachineManagerImpl and improve logs

GutoVeronezi:
### Description
This PR intends to refactor a few process, like extract repeated code to methods, remove unnecessary logic... Of class `VirtualMachineManagerImpl` and improve logging to facilitate troubleshooting.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?


DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan test matrix

DaanHoogland (approved):
clgtm

rhtyd:
@GutoVeronezi you've submitted many PRs which are the same class of PRs around log improvements and refactorings and changing the resource/vo/object `toString()` that returns hand-written json and add inconsistency across other VOs/objects/... Why is that? Are the json output from log statements consumed somewhere?

Can you club all such PRs into a single PR with separate/individual commits? They make it difficult to review and test individually and take a lot of time and resources from reviewers and BO/lab.

GabrielBrascher:
Here follow my 2cents.

#### 1 - Number of PRs

@rhtyd raised an interesting point regartding the quantity of PRs.
I agree that there are quite a lot of PRs, and that this can make our life harder to track, test, and review them all. However, I understand that many of them hold their particular contexts.

I think that creating one issue to track them all would help on testing and reviewing them all. The idea would be to link any PR that fits the issue efforts on enhancing log messsages as well as keeping a track of each PR (merged, under review, draft, etc).

#### 2 - Creating an log message pattern

I have been running 4.16-SNAPSHOT already with a couple of these toString() methods. Personally I am happy to see log messages with standars and solid information. Maybe we can find better formatations than the _JSON like_. However, I find them interesting as it is a globally understood pattern for anyone at IT.

The biggest plus in adopting `toString` is the standardtization on log messages. Nowadays we see all kind of ways of logging resources. Some logs present the name, or the ID, sometimes just the UUID, or combinations of them. It is about time to promote a standard (which would be possible mostly due to `toString`).

DaanHoogland:
@rhtyd please let's make changes per PR as little as possible. smaller PRs are usually easier to review. Only when the code is very intertwined should we combine PRs into bigger ones.

rhtyd:
@DaanHoogland I've just shared my preference that is around optimising around reviewers' and lab resources time. One can get the same thing with a single PR with individual small commits that are easier to review (Github allows you to review each commit individually). The benefit of the latter is you save a lot of overhead in people's time in tracking/test, and on smoketest resources.

@GutoVeronezi what you've raised is a valid issue, however, there was no discussion around a new standard/consistent format for people to follow for new code. For example and as a suggestion, instead of `NicProfile {"id": 1, "vmId": 2, "reservationId": "3", "iPv4Address": "192.168.0.1", "broadcastUri": "..."}`, something like `NicProfile [id=1, vmId=2, reservationId=3, iPv4Address=192.168.0.1, ...]` may read better, and developers don't need to work with handwritten json. This is merely an example - I'm not asking you to change the string format.

Here something to think about:
- You can present the enhancement cleanups under a standard initiative, start discussion on dev@, i.e. what should be a new standard on improving logs, what new developers, new feature/changes should follow. For example, Gabriel's suggestion on a standard logging pattern is great feedback.
- A central utility/method that can help switch between formats of objects (json, csv, ...), if that is even feasible? Do a poc with your proposal.
- Should we log internal DB IDs in logs? Is the cleanup PRs is replacing old bad logging style in a new way?
- I've seen customers/users who pass logs through an aggregation system such as Splunk where they may have rules to make sense of things. Many old dinosaurs in the community like me are used to a pattern/style of reading logs and code i.e. have we considered what could/would break for existing users and developers, or their approach as their eyes are set looking for something they'll not adjust (written with humour - this point can be completely ignored we don't maintain backward compatibility for logs)

On the issue of context, frankly, any PR that is more than a 100 lines of changes is not going to get a thorough review for each and every line by 2xreviewers, it's simply not feasible at least for me and people I know. Instead, we use smoketests and other means (manual QA, design/architecture review, consistency/pattern of changes, ...) to validate such PRs. I would club all such PRs around the same initiative which in this particular case is around logging/refactoring improvements. Splitting the work around the same initiative in 10 different PRs camouflaging under different `contexts` will just make the process a lot slower and you're essentially in some way telling the community (of dinosaurs :D) that you want them to work hard in their "free" time and you essentially don't care of the overhead it causes them on reviewing/tracking/testing/...

rhtyd:
@blueorangutan package

DaanHoogland:
@blueorangutan test

GutoVeronezi:
Can anyone review this?

sureshanaparti:
Hi @GutoVeronezi can you fix the conflicts.

GutoVeronezi:
@sureshanaparti done, thanks!

GutoVeronezi:
Can anyone review this?

GutoVeronezi:
@GabrielBrascher done, thanks!

GabrielBrascher (approved):
Thanks for addressing the raised points, @GutoVeronezi!
Code LGTM.

DaanHoogland:
not sure about this failure
@blueorangutan test

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -572,19 +571,19 @@ protected void advanceExpunge(VMInstanceVO vm) throws ResourceUnavailableExcepti
 
         final HypervisorGuru hvGuru = _hvGuruMgr.getGuru(vm.getHypervisorType());
 
-        s_logger.debug("Cleaning up NICS");
-        final List<Command> nicExpungeCommands = hvGuru.finalizeExpungeNics(vm, profile.getNics());
```
nvazquez, line 576: Why are we removing these commands? Checked the Vmware guru may need `UnregisterNicCommand` processing

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -618,43 +607,43 @@ protected void advanceExpunge(VMInstanceVO vm) throws ResourceUnavailableExcepti
...
         guru.finalizeExpunge(vm);
-        //remove the overcommit details from the uservm details
+
         userVmDetailsDao.removeDetails(vm.getId());
 
-        // Remove deploy as-is (if any)
         userVmDeployAsIsDetailsDao.removeDetails(vm.getId());
 
-        // send hypervisor-dependent commands before removing
-        final List<Command> finalizeExpungeCommands = hvGuru.finalizeExpunge(vm);
```
nvazquez, line 628: Can you also explain why these are not needed?

GutoVeronezi:
@nvazquez I had in mind the following logic to remove these codes:

The `nicExpungeCommands` is instantied here:
https://github.com/apache/cloudstack/blob/2a4c2c250696a154ccf8a7c8fff8604d893e9258/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java#L576

There are 2 implementation for this method:
- `HypervisorGuruBase`, which returns `null`:
https://github.com/apache/cloudstack/blob/2a4c2c250696a154ccf8a7c8fff8604d893e9258/server/src/main/java/com/cloud/hypervisor/HypervisorGuruBase.java#L293-L296

- `VMWareGuru`, which returns some commands:
https://github.com/apache/cloudstack/blob/2a4c2c250696a154ccf8a7c8fff8604d893e9258/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/guru/VMwareGuru.java#L342-L358

`nicExpungeCommands` is only used in the following block:
https://github.com/apache/cloudstack/blob/2a4c2c250696a154ccf8a7c8fff8604d893e9258/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java#L628-L652

However, only `HypervisorGuruBase` implements the method `finalizeExpunge`, which returns `null`:
https://github.com/apache/cloudstack/blob/2a4c2c250696a154ccf8a7c8fff8604d893e9258/server/src/main/java/com/cloud/hypervisor/HypervisorGuruBase.java#L288-L291

Therefore, the block where `nicExpungeCommands` is used is never called, therefore it's not needed. 

If I'm missing something, please let me know.

GutoVeronezi:
As @nvazquez and me discussed offline, the block should be called, which characterize https://github.com/apache/cloudstack/pull/4966#issuecomment-897615227 as a bug. I created issue #5308 reporting this bug  and will undo the code remotion.

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

GutoVeronezi:
@DaanHoogland @GabrielBrascher @nvazquez @rhtyd is there anything else to do?

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

rhtyd:
We need test matrix on this @nvazquez
General note: such a refactoring PR worries me closer to a RC, I've seen my share of blockers which get introduced due to refactoring changes as there are so many paths which may not be covered 100% by smoketests. I suggest the RM to discuss with the author and other reviewers and take the call whether we want the risk or if there's a risk. 

@blueorangutan package

GutoVeronezi:
@rhtyd I changed `toString` implementation to use `ReflectionToStringBuilderUtils`.

rhtyd:
Rekicked Travis.

Let's do another round, merge if next round of smoketests pass.
@blueorangutan package

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -1243,15 +1216,11 @@ public void orchestrateStart(final String vmUuid, final Map<VirtualMachineProfil
                                 throw new ConcurrentOperationException("Failed to deploy VM"+ vm.getUuid());
                             }
 
-                            // Update GPU device capacity
                             final GPUDeviceTO gpuDevice = startAnswer.getVirtualMachine().getGpuDevice();
                             if (gpuDevice != null) {
                                 _resourceMgr.updateGPUDetails(destHostId, gpuDevice.getGroupDetails());
                             }
 
-                            // Remove the information on whether it was a deploy vm request.The deployvm=true information
```
rhtyd, line 1252: Some comments may be kept

rhtyd (approved):
LGTM but changes in VirtualMachineManagerImpl worries me, it may lead to some regression. So at least test matrix should be run before merging this cc @nvazquez @sureshanaparti 
Further the PR doesn't only improve logs but significantly refactor the codebase and removes some comments which I think should be kept which are not describing the immediate logic.

rhtyd:
@blueorangutan test matrix

nvazquez:
Thanks for reviewing @rhtyd I see your points - opening for discussion @GutoVeronezi @sureshanaparti @rhtyd @GabrielBrascher I think we all agree this PR will need some extensive testing for regressions on VM lifecycle operations. A good sign is that smoke tests are returning good results. Given that we are close to cutting the RC, do you think we can complete all the required testing in time for 4.16?

rhtyd:
@nvazquez General comment - closer to an RC you want to exclude any non-essential PRs you aren't confident on wrt stability and potential regressions. We can move such PRs to merging right after cutting the RC, and move them to the next major or minor release which should give enough time for community to test the changes until the next major/minor release.

nvazquez:
@rhtyd makes sense, thanks. If no objections let's move this PR to the next milestone to prevent unexpected regressions close to the RC cut date

GutoVeronezi:
Restart jenkins...

GutoVeronezi:
Restart jenkins...

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

PR #4978, opened 2021-04-30
KVM High Availability regardless of storage

GabrielBrascher:
### Description

Currently, KVM HA implementation works only if the cluster has at least one primary storage served via NFS. This is due to the NFS heartbeat script used to check if the host is healthy. This implementation adds health checks that work regardless of a storage pool. This is done via a Java client that checks Agent status via a webserver.

The additional web-server exposes a simple JSON API that returns a list of Virtual Machines that are running on that host according to Libvirt. This way, KVM HA can verify, via Libvirt, VMs status with HTTP-call to this simple webserver and determine if the host is actually down or if it is just the Java Agent which has crashed.

#### New KVM HA Helper component
The following image shows how the new KVM-HA-Helper web-service is integrated. The current NFS HeartBeat execution flow will still be used aligned with the new HA-Helper.

<p align="center">
  
</p>

#### High Availability Workflow

Proposed workflow where the HA Check takes into account both **NFS Heartbeat** and the **KVM HA Helper** checks.

**Note that** in order to simplify the diagram it is ignored the whole [HA state machine](https://cwiki.apache.org/confluence/display/CLOUDSTACK/Host+HA). However, if NFS and HA Helper fails not necessarily it is going to Recover/Fence a host as depending on the HA configurations it needs to re-check some times until it reaches a threshold of accepted failures.

<p align="center">
  
</p>

### Types of changes

- [x] Major
- [ ] Minor


packaging/systemd/cloudstack-agent-ha-helper.service
```diff
@@ -0,0 +1,36 @@
...
+
+[Unit]
+Description=CloudStack Agent HA Helper
+Documentation=http://www.cloudstack.org/
+Requires=libvirtd.service
+After=libvirtd.service
+
+[Service]
+Type=simple
+EnvironmentFile=/etc/default/cloudstack-agent-ha-helper
```
wido, line 30: This will not work on RHEL as that uses:

<pre>/etc/sysconfig</pre>

For the configuration. We need to have something similar like for the cloudstack-agent package where we have different systemd files for CentOS and Ubuntu.

plugins/hypervisors/kvm/src/main/java/com/cloud/ha/KVMInvestigator.java
```diff
@@ -85,12 +89,35 @@ public Status isAgentAlive(Host agent) {
...
+
+    /**
+     * It checks the KVM node healthy via KVM HA Agent. If the agent is healthy it returns Status.Up, otherwise it keeps the provided Status as it is.
+     */
+    private Status checkAgentStatusViaKvmHaAgent(Host agent, Status agentStatus) {
+        KvmHaAgentClient kvmHaAgentClient = new KvmHaAgentClient(agent);
+        boolean isVmsCountOnKvmMatchingWithDatabase = kvmHaAgentClient.isKvmHaAgentHealthy(agent, vmInstanceDao);
+        if(isVmsCountOnKvmMatchingWithDatabase) {
+            agentStatus = Status.Up;
+            s_logger.debug(String.format("Checking agent %s status; KVM HA Agent is Running as expected."));
```
wido, line 113: Shouldn't we pass an argument for '%s'?

Right now this won't print something properly.

plugins/hypervisors/kvm/src/main/java/com/cloud/ha/KVMInvestigator.java
```diff
@@ -85,12 +89,35 @@ public Status isAgentAlive(Host agent) {
...
+        } else {
+            s_logger.warn(String.format("Checking agent %s status. Failed to check host status via KVM HA Agent"));
```
wido, line 115: Shouldn't we pass an argument for '%s'?

Right now this won't print something properly.

GabrielBrascher: Code has been updated adrressing these `String.format` missing parameters. Thanks @wido!

div8cn:
HI, I see this PR is very good.

I have a question, if the manager network link is interrupted (such as a switch failure), but the storage network of kvm is still running normally.

Will this cause the VM to trigger HA?

I am worried that the image is "double written", resulting in damage to the image

wido:
> HI, I see this PR is very good.
> 
> I have a question, if the manager network link is interrupted (such as a switch failure), but the storage network of kvm is still running normally.
> 
> Will this cause the VM to trigger HA?
> 
> I am worried that the image is "double written", resulting in damage to the image

You should always have the Out of Band management of the hypervisors configured:

- IPMI 
- Redfish

When the mgmt server wants to perform HA it will Fence off  the Host by performing a power cycle of that host via OOB.

This will make sure no VMs are running and they can safely be started on a different host.

In addition we also have locking mechanisms:

- file locks on QCOW2 when using NFS (modern Qemu required)
- exclusive locking with RBD/Ceph on images

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -151,20 +203,33 @@ private boolean isVMActivtyOnHost(Host agent, DateTime suspectTime) throws HAChe
...
-        for (StoragePool pool : poolVolMap.keySet()) {
-            activityStatus = verifyActivityOfStorageOnHost(poolVolMap, pool, agent, suspectTime, activityStatus);
-            if (!activityStatus) {
-                LOG.warn(String.format("It seems that the storage pool [%s] does not have activity on %s.", pool.getId(), agent.toString()));
-                break;
+        boolean activityStatus = false;
+        if (isHostServedByNfsPool(agent)) {
+            HashMap<StoragePool, List<Volume>> poolVolMap = getVolumeUuidOnHost(agent);
+            for (StoragePool pool : poolVolMap.keySet()) {
+                if (Storage.StoragePoolType.NetworkFilesystem == pool.getPoolType() || Storage.StoragePoolType.ManagedNFS == pool.getPoolType()) {
```
GutoVeronezi, line 210: We could create a collection and verify if `contains` the pool type, like: 
```java
private final Set<Storage.StoragePoolType> <insertNameHere> = new HashSet<>(Arrays.asList(Storage.StoragePoolType.NetworkFilesystem, Storage.StoragePoolType.ManagedNFS));
...
if (<insertNameHere>.contains(pool.getPoolType())) {
```

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -213,6 +278,18 @@ protected boolean verifyActivityOfStorageOnHost(HashMap<StoragePool, List<Volume
         return poolVolMap;
     }
 
+    private boolean isHostServedByNfsPool(Host agent) {
+        List<StoragePoolHostVO> storagesOnHost = storagePoolHostDao.listByHostId(agent.getId());
+        for (StoragePoolHostVO storagePoolHostRef : storagesOnHost) {
+            StoragePoolVO storagePool = this.storagePool.findById(storagePoolHostRef.getPoolId());
+            if(Storage.StoragePoolType.NetworkFilesystem == storagePool.getPoolType()
+                    || Storage.StoragePoolType.ManagedNFS == storagePool.getPoolType()) {
```
GutoVeronezi, line 286: The collection cited in https://github.com/apache/cloudstack/pull/4978/files#r629568449 could be used here too.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,271 @@
...
+     *
+     *  IF: <br>
+     *  (i) KVM HA agent finds 0 running but CloudStack considers that the host has 2 or more VMs running: returns false as could not find VMs running but it expected at least
+     *    2 VMs running, fencing/recovering host would avoid downtime to VMs in this case.<br>
+     *  (ii) KVM HA agent finds 0 VM running but CloudStack considers that the host has 1 VM running: return true and log WARN messages and avoids triggering HA recovery/fencing
+     *    when it could be a inconsistency when migrating a VM.<br>
+     *  (iii) amount of listed VMs is different than expected: return true and print WARN messages so Admins can monitor and react accordingly
+     */
+    public boolean isKvmHaAgentHealthy(Host host, VMInstanceDao vmInstanceDao) {
+        int numberOfVmsOnHostAccordingToDB = listVmsOnHost(host, vmInstanceDao).size();
```
GutoVeronezi, line 149: Simple change just to let it a consistent camelcase
```java
int numberOfVmsOnHostAccordingToDb = listVmsOnHost(host, vmInstanceDao).size();
```

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,295 @@
...
+            int stoppingVms = listByHostAndStateStopping.size();
+            int migratingVms = listByHostAndStateMigrating.size();
+            int countRunningVmsOnAgent = countRunningVmsOnAgent();
+            LOGGER.trace(
+                    String.format("%s has (%d Starting) %d Running, %d Stopping, %d Migrating. Total listed via DB %d / %d (via libvirt)", agent.getName(), startingVMs, runningVMs,
+                            stoppingVms, migratingVms, listByHostAndState.size(), countRunningVmsOnAgent));
+        }
+
+        return listByHostAndState;
+    }
```
GutoVeronezi, line 160: As the method `listByHostAndStat` receive a varargs of state as parameter, we could simplify this method by joining the three first requests and filtering they if needed:

```java
List<VMInstanceVO> listByHostAndStates = vmInstanceDao.listByHostAndState(host.getId(), VirtualMachine.State.Running, VirtualMachine.State.Stopping, VirtualMachine.State.Migrating);

if (LOGGER.isTraceEnabled()) {
    List<VMInstanceVO> listByHostAndStateStarting = vmInstanceDao.listByHostAndState(host.getId(), VirtualMachine.State.Starting);
    int startingVMs = listByHostAndStateStarting.size();
    int runningVMs = listByHostAndStates.stream().filter...;
    int stoppingVms = listByHostAndStates.stream().filter...;
    int migratingVms = listByHostAndStates.stream().filter...;
    int countRunningVmsOnAgent = countRunningVmsOnAgent();
    LOGGER.trace(
            String.format("%s has (%d Starting) %d Running, %d Stopping, %d Migrating. Total listed via DB %d / %d (via libvirt)", agent.getName(), startingVMs, runningVMs,
                    stoppingVms, migratingVms, listByHostAndState.size(), countRunningVmsOnAgent));
}

return listByHostAndStates;
```

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,295 @@
...
+            LOGGER.warn(String.format("Host %s is not considered healthy and HA fencing/recovering process might be triggered.", agent.getName(), numberOfVmsOnHostAccordingToDb));
+            return false;
+        }
+        if (numberOfVmsOnHostAccordingToDb == numberOfVmsOnAgent) {
+            return true;
+        }
+        if (numberOfVmsOnAgent == 0 && numberOfVmsOnHostAccordingToDb > CAUTIOUS_MARGIN_OF_VMS_ON_HOST) {
+            // Return false as could not find VMs running but it expected at least one VM running, fencing/recovering host would avoid downtime to VMs in this case.
+            // There is cautious margin added on the conditional. This avoids fencing/recovering hosts when there is one VM migrating to a host that had zero VMs.
+            // If there are more VMs than the CAUTIOUS_MARGIN_OF_VMS_ON_HOST) the Host should be treated as not healthy and fencing/recovering process might be triggered.
```
GutoVeronezi, line 186: Javadoc already explain a little bit of the context, could we remove this comment and improve javadoc?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,295 @@
...
+                            EXPECTED_HTTP_STATUS, statusCode));
+            return null;
+        }
+
+        LOGGER.debug(String.format("Successfully executed HTTP %s request [URL: %s].", httpReq.getMethod(), url));
+        return response;
+    }
+
+    protected HttpResponse retryUntilGetsHttpResponse(String url, HttpRequestBase httpReq, HttpClient client) {
+        for (int attempt = 1; attempt < MAX_REQUEST_RETRIES + 1; attempt++) {
```
GutoVeronezi, line 259:
```suggestion
        for (int attempt = 1; attempt <= MAX_REQUEST_RETRIES; attempt++) {
```

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,295 @@
...
+        }
+        try {
+            in = response.getEntity().getContent();
+            BufferedReader streamReader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
+            jsonString = streamReader.readLine();
+        } catch (UnsupportedOperationException | IOException e) {
+            throw new CloudRuntimeException("Failed to process response", e);
+        }
+
+        return new JsonParser().parse(jsonString).getAsJsonObject();
```
GutoVeronezi, line 293:
```suggestion
        if (response == null) {
            return null;
        }

        try {
            InputStream in = response.getEntity().getContent();
            BufferedReader streamReader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
            String jsonString = streamReader.readLine();
            return new JsonParser().parse(jsonString).getAsJsonObject();
        } catch (UnsupportedOperationException | IOException e) {
            throw new CloudRuntimeException("Failed to process response", e);
        }
```

GutoVeronezi: If needed, could we improve the exception with some `response` context?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -81,7 +98,63 @@ public boolean isActive(Host r, DateTime suspectTime) throws HACheckerException
...
+        boolean isHostServedByNfsPool = isHostServedByNfsPool(r);
+        boolean isKvmHaWebserviceEnabled = isKvmHaWebserviceEnabled(r);
+
+        isHealthy = isHealthViaNfs(r);
+
+        if (!isKvmHaWebserviceEnabled) {
+            return isHealthy;
+        }
+
+        //TODO
```
sureshanaparti, line 111: an empty TODO here, may be not required.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -81,7 +98,63 @@ public boolean isActive(Host r, DateTime suspectTime) throws HACheckerException
...
+    /**
+     * Checks the host health via an web-service that retrieves Running KVM instances via Libvirt. <br>
+     * The health-check is executed on the KVM node and verifies the amount of VMs running and if the Libvirt service is running.
+     */
+    private boolean isVmActivtyOnHostViaKvmHaWebservice(Host host) {
+        KvmHaAgentClient kvmHaAgentClient = new KvmHaAgentClient(host);
+        return kvmHaAgentClient.isKvmHaAgentHealthy(host, vmInstanceDao);
+    }
+
+    //TODO
```
sureshanaparti, line 130: empty TODO comment ^^^

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
GutoVeronezi: Seems to me that this method will be called several times, should we extract this `new JsonParser()` to a constant?

plugins/hypervisors/kvm/src/test/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClientTest.java
```diff
@@ -0,0 +1,278 @@
...
+//        Mockito.doReturn(jObject).when(kvmHaAgentClient).processHttpResponseIntoJson(Mockito.any(HttpResponse.class));
+//
+//        int result = kvmHaAgentClient.countRunningVmsOnAgent();
+//        Assert.assertEquals(expectedListedVms, result);
+//    }
+//TODO
+//    @Test
+//    public void isTargetHostReachableTest() {
+//        kvmHaAgentClient.isTargetHostReachable(PRIVATE_IP_ADDRESS);
+//    }
```
GutoVeronezi, line 276: Do we need commented code?

sureshanaparti, line 276: unused code (in comments) ^^^, can be removed.

GabrielBrascher: @GutoVeronezi No, we don't. Sorry, forgot to convert for a draft. I am adding some checks to enhance the HA validation.

The idea here is to verify if the Suspect Host is reachable by other hosts on the cluster, therefore excluding network issues between mgmt and kvm agents and avoiding fencing / rebooting in such cases.

GabrielBrascher: You are right @sureshanaparti, this commit has added some WIP code, and just converted back to draft.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
GabrielBrascher: Good point, I am going to update it on the next commit :+1:

GabrielBrascher: @GutoVeronezi do you mean something like this, right?
```
public class KvmHaAgentClient {
    private static final JsonParser JSON_PARSER = new JsonParser();
    ...
    ...
    ...
    @Nullable
    protected JsonObject processHttpResponseIntoJson(HttpResponse response) {
        if (response == null) {
            return null;
        }
        try {
            InputStream in = response.getEntity().getContent();
            BufferedReader streamReader = new BufferedReader(new InputStreamReader(in, StandardCharsets.UTF_8));
            return JSON_PARSER.parse(streamReader.readLine()).getAsJsonObject();
        } catch (UnsupportedOperationException | IOException e) {
            throw new CloudRuntimeException("Failed to process response", e);
        }
    }
    ...
    ...
}
```

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHAConfig.java
```diff
@@ -53,4 +53,17 @@
...
+            "It sets the port used to communicate with the KVM HA Agent Microservice that is running on KVM nodes. Default value is 8080.",
+            true, ConfigKey.Scope.Cluster);
+
+    public static final ConfigKey<Boolean> IsKvmHaWebserviceEnabled = new ConfigKey<Boolean>("Advanced", Boolean.class, "kvm.ha.webservice.enabled", "false",
+            "The KVM HA Webservice is executed on the KVM node and checks the amount of VMs running via libvirt. It serves as a HA health-check for KVM nodes. "
+                    + "One can enable (set to 'true') or disable it ('false'). If disabled then CloudStack ignores HA validation via this agent.",
+            true, ConfigKey.Scope.Cluster);
+
+    public static final ConfigKey<Double> KvmHaAcceptedProblematicHostsRatio = new ConfigKey<Double>("Advanced", Double.class, "kvm.ha.accepted.problematic.hosts.ratio", "0.3",
+            "The ratio of problematic Hosts accepted on a Cluster. If a cluster has more than the accepted ratio, HA will not be Fence/Recover Hosts and Admins will be notified to check the cluster healthy. "
```
GabrielBrascher, line 66:
```suggestion
            "The ratio of problematic Hosts accepted on a Cluster. If a cluster has more than the accepted ratio, HA will not Fence/Recover Hosts; instead, it will notify Admins to check the cluster healthy. "
```

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

GabrielBrascher:
@rhtyd thanks for running the packaging.

I think I know why it is failing for Centos7-8: I still need to cover centos packaging for the new `agent-ha-helper`. For now it is packed only for Ubuntu. I shall update this PR in order to allow it.

This also makes sense on why all Jenkins/Travis checks are green but packaging results for centos are not good.

plugins/hypervisors/kvm/src/main/java/com/cloud/ha/KVMInvestigator.java
```diff
@@ -101,24 +115,29 @@ public Status isAgentAlive(Host agent) {
                 hostStatus = answer.getResult() ? Status.Down : Status.Up;
             }
         } catch (Exception e) {
-            s_logger.debug("Failed to send command to host: " + agent.getId());
+            s_logger.debug(String.format("Failed to send command to %s", agent));
```
GutoVeronezi, line 118: - Should an exception be logged in `debug`? 
- We could pass the exception as parameter to the log.
- Do we need a Pokemon here?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -59,29 +68,62 @@
...
+        if(isHostServedByNfsPool) {
+            isHealthy = isHealthViaNfs(host);
+        }
+
+        if (!isKvmHaWebserviceEnabled) {
+            return isHealthy;
+        }
+
+        if (kvmHaHelper.isKvmHealthyCheckViaLibvirt(host) && !isHealthy) {
+            isHealthy = true;
```
GutoVeronezi, line 112: We could just return `true` here (it is not a technical reason, just another point of view).

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -151,20 +193,34 @@ private boolean isVMActivtyOnHost(Host agent, DateTime suspectTime) throws HAChe
         if (agent.getHypervisorType() != Hypervisor.HypervisorType.KVM && agent.getHypervisorType() != Hypervisor.HypervisorType.LXC) {
```
GutoVeronezi, line 193: We could create a constant with these hypervisor types and verify if contains.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,256 @@
...
+    protected HttpResponse retryUntilGetsHttpResponse(String url, HttpRequestBase httpReq, HttpClient client) {
+        for (int attempt = 1; attempt <= MAX_REQUEST_RETRIES; attempt++) {
+            try {
+                TimeUnit.SECONDS.sleep(WAIT_FOR_REQUEST_RETRY);
+                LOGGER.debug(String.format("Retry HTTP %s request [URL: %s], attempt %d/%d.", httpReq.getMethod(), url, attempt, MAX_REQUEST_RETRIES));
+                return client.execute(httpReq);
+            } catch (IOException | InterruptedException e) {
+                String errorMessage = String.format("Failed to execute HTTP %s request retry attempt %d/%d [URL: %s] due to exception %s",
+                        httpReq.getMethod(), attempt, MAX_REQUEST_RETRIES, url, e);
+                LOGGER.error(errorMessage);
```
GutoVeronezi, line 231: We could pass the exception as parameter to the log.

plugins/hypervisors/kvm/src/main/java/com/cloud/ha/KVMInvestigator.java
GabrielBrascher: Thanks for bringing it up @GutoVeronezi, these are pertinent questions.

This is a perfect case of _Pokémon catch_ (catch them all). To be honest, I cannot even find how an Exception could be raised by `AgentManagementImpl.easySend` as the method itself handles `Exception`.

I was inclined to simply remove it; however, I avoided changing any behavior of the HeartBeat execution flow avoiding any regression issue.

I will take a double-check at this `easySend` catch.
But it is a good thing to pay attention to, if I do not change it at this PR we can always plan another PR/issue cleaning some of these catches.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
GabrielBrascher: Thanks for the suggestion, I like this idea. Done!

rhtyd:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/ha/KVMInvestigator.java
GabrielBrascher: @GutoVeronezi I decided to remove this catch.
When checking the easySend there is already enough catches. If it does not catch the exception ... I don't know what would catch it:

```
public Answer easySend(final Long hostId, final Command cmd) {
        try {
                ...
                ...
                ...
        } catch (final AgentUnavailableException e) {
            s_logger.warn(e.getMessage());
            return null;
        } catch (final OperationTimedoutException e) {
            s_logger.warn("Operation timed out: " + e.getMessage());
            return null;
        } catch (final Exception e) {
            s_logger.warn("Exception while sending", e);
            return null;
        }
```

GutoVeronezi: @GabrielBrascher indeed, I see no way to it throw an exception here

GabrielBrascher:
@blueorangutan package

nvazquez:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaHelper.java
```diff
@@ -0,0 +1,194 @@
...
+    public boolean isKvmHealthyCheckViaLibvirt(Host host) {
+        boolean isKvmHaAgentHealthy = isKvmHaAgentHealthy(host);
+
+        if (!isKvmHaAgentHealthy) {
+            if (isClusteProblematic(host) || isHostAgentReachableByNeighbour(host)) {
+                return true;
+            }
+        }
+
+        return isKvmHaAgentHealthy;
```
GutoVeronezi, line 139: We could use a ternary on return here.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -59,29 +68,63 @@
...
+    }
+
+    private boolean isHealthViaNfs(Host r) {
+        boolean isHealthy = true;
+        if (isHostServedByNfsPool(r)) {
+            isHealthy = isAgentActive(r);
+            if (!isHealthy) {
+                LOG.warn(String.format("NFS storage health check failed for %s. It seems that a storage does not have activity.", r.toString()));
+            }
+        }
```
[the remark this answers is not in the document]
GabrielBrascher: Nice! I like the idea @GutoVeronezi, I will enhance this part and ensure to cover the flows with some relevant logs.

rhtyd:
@blueorangutan package

GabrielBrascher:
@rhtyd I will check the centos packaging, it still fails to build.
Deb packages for the new service ha-helper are successfully built.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaHelper.java
GabrielBrascher: Personally, I do prefer _IFs_ than a _ternary_. At least in such cases where a ternary transforms into a huge line. But that might be just me used to the "old fashion" _IFs_ ... :thinking:.

But I have nothing against ternary, we could definitely do such a change.

1. From:
```
public boolean isKvmHealthyCheckViaLibvirt(Host host) {
    boolean isKvmHaAgentHealthy = isKvmHaAgentHealthy(host);

    if (!isKvmHaAgentHealthy) {
        if (isClusteProblematic(host) || isHostAgentReachableByNeighbour(host)) {
        return true;
        }
    }
    
    return isKvmHaAgentHealthy;
}
```

2. To:
```
public boolean isKvmHealthyCheckViaLibvirt(Host host) {
    boolean isKvmHaAgentHealthy = isKvmHaAgentHealthy(host);
    return !isKvmHaAgentHealthy && (isClusteProblematic(host) || isHostAgentReachableByNeighbour(host)) ? true : isKvmHaAgentHealthy;
}
```

OR
```
public boolean isKvmHealthyCheckViaLibvirt(Host host) {
    boolean isKvmHaAgentHealthy = isKvmHaAgentHealthy(host);
    return !isKvmHaAgentHealthy && (isClusteProblematic(host) || isHostAgentReachableByNeighbour(host))
    ? true
    : isKvmHaAgentHealthy;
}
```

GutoVeronezi: @GabrielBrascher I see no problem in keeping the `if` statement. In this case, I would only suggest you to unify the `ifs`, to avoid nesting:

```suggestion
        boolean isKvmHaAgentHealthy = isKvmHaAgentHealthy(host);

        if (!isKvmHaAgentHealthy && (isClusteProblematic(host) || isHostAgentReachableByNeighbour(host))) {
            return true;
        }

        return isKvmHaAgentHealthy;
```

From this unify came to me the idea of a ternary...

GabrielBrascher:
Fixed conflict.
As the conflict impacted some of the initial commits it was kind of tricky to rebase the branch `kvm-ha-microservice-client` to the updated `main` which led to this list of updated commits/hash.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaHelper.java
GabrielBrascher: @GutoVeronezi your suggestion makes sense. I've updated the code removing the nested IF.
Thanks for the review!

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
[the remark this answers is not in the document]
GabrielBrascher: Done, thanks!

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/ha/KVMInvestigator.java
```diff
@@ -77,57 +77,84 @@ public Status isAgentAlive(Host agent) {
...
+        for (StoragePoolVO pool : zonePools) {
+            if (pool.getPoolType() == StoragePoolType.NetworkFilesystem) {
+                return true;
             }
         }
-        if (!hasNfs) {
-            s_logger.warn(
-                    "Agent investigation was requested on host " + agent + ", but host does not support investigation because it has no NFS storage. Skipping investigation.");
-            return Status.Disconnected;
+        return false;
```
GutoVeronezi, line 112: We could use `stream` methods to simplify this implementation:

```java
return zonePools.stream().anyMatch(pool -> pool.getPoolType() == StoragePoolType.NetworkFilesystem);
``` 

Also, this code is repeated right below in `hasNfsPoolClusterWideForHost`, we could extract it to a method and add unit tests.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
```diff
@@ -213,6 +270,17 @@ protected boolean verifyActivityOfStorageOnHost(HashMap<StoragePool, List<Volume
...
 
+    private boolean isHostServedByNfsPool(Host agent) {
+        List<StoragePoolHostVO> storagesOnHost = storagePoolHostDao.listByHostId(agent.getId());
+        for (StoragePoolHostVO storagePoolHostRef : storagesOnHost) {
+            StoragePoolVO storagePool = this.storagePool.findById(storagePoolHostRef.getPoolId());
+            if (NFS_POOL_TYPE.contains(storagePool.getPoolType())) {
+                return true;
+            }
+        }
+        return false;
```
GutoVeronezi, line 281: We could use `stream().anyMatch` here.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/ha/KVMInvestigator.java
GabrielBrascher: Done!

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHostActivityChecker.java
GabrielBrascher: @GutoVeronezi here I think that it is a bit trickier than a normal loop interation with a simple if.
Such as the other one:
```
for (StoragePoolVO pool : zonePools) {
        if (pool.getPoolType() == StoragePoolType.NetworkFilesystem) {
                return true;
        }
}
```

Here the list does not store the object that is being used in the conditional, but instead just an object (`StoragePoolHostVO`) that will serve to link to the object in quesiton (`StoragePoolVO`). Thus the stream .anyMatch would not serve here.

Please correct me if I am wrong in this one.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
Most of the failures/errors do not seem related to this PR. I checked the HA failures looking for correlation to this PR but it also does not look like the case.

Maybe spin another test?

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
@blueorangutan package

nvazquez:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaHelper.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    /**
+     * Returns true if the given Host KVM-HA-Helper is reachable by another host in the same cluster.
+     */
+    protected boolean isHostAgentReachableByNeighbour(Host host) {
+        List<HostVO> neighbors = resourceManager.listHostsInClusterByStatus(host.getClusterId(), Status.Up);
+        for (HostVO neighbor : neighbors) {
+            boolean isVmActivtyOnNeighborHost = isKvmHaAgentHealthy(neighbor);
+            if (isVmActivtyOnNeighborHost) {
```
GutoVeronezi, line 114: We could invert this if to reduce indentation.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaHelper.java
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Checks if the KVM HA webservice is enabled. One can enable or disable it via global settings 'kvm.ha.webservice.enabled'.
+     */
+    public boolean isKvmHaWebserviceEnabled(Host host) {
+        boolean isKvmHaWebserviceEnabled = KVMHAConfig.IsKvmHaWebserviceEnabled.value();
+        if (!isKvmHaWebserviceEnabled) {
+            LOGGER.debug(String.format("Skipping KVM HA web-service verification for %s due to 'kvm.ha.webservice.enabled' not enabled.", host));
+            return false;
+        }
+        return true;
```
GutoVeronezi, line 147: We could remove both returns and just add a `return isKvmHaWebserviceEnabled;` at the end.

GutoVeronezi (approved):
The code LGTM, however I didn't test it.

@GabrielBrascher I raised some minor points.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaHelper.java
GabrielBrascher: @GutoVeronezi these slipped my eyes.
Thanks a lot for bringing it. I will soon update the code.

GabrielBrascher:
@blueorangutan package

GabrielBrascher:
For reference, the tests that failed are:

1. test_hostha_enable_ha_when_host_disabled
2. test_hostha_enable_ha_when_host_in_maintenance

I am still checking why they are failing and if it is related to this specific PR.

Logs:
```
testcase classname="tests.smoke.test_hostha_kvm.TestHAKVM" name="test_hostha_configure_default_driver" time="0.719"
"tests.smoke.test_hostha_kvm.TestHAKVM" name="test_hostha_enable_ha_when_host_disabled" time="1.115" "marvin.cloudstackException.CloudstackAPIException" message="Execute cmd: updatehost failed, due to: errorCode: 530, errorText:Failed to update host:2,No next resource state found for current state = Maintenance event = Disable":
  File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.6/unittest/case.py", line 622, in run
    testMethod()
  File "/marvin/tests/smoke/test_hostha_kvm.py", line 259, in test_hostha_enable_ha_when_host_disabled
    self.disableHost(self.host.id)
  File "/marvin/tests/smoke/test_hostha_kvm.py", line 609, in disableHost
    response = self.apiclient.updateHost(cmd)
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackAPI/cloudstackAPIClient.py", line 915, in updateHost
    response = self.connection.marvinRequest(command, response_type=response, method=method)
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 381, in marvinRequest
    raise e
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 376, in marvinRequest
    raise self.__lastError
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 310, in __parseAndGetResponse
    response_cls)
  File "/usr/local/lib/python3.6/site-packages/marvin/jsonHelper.py", line 155, in getResultObj
    raise cloudstackException.CloudstackAPIException(respname, errMsg)
marvin.cloudstackException.CloudstackAPIException: Execute cmd: updatehost failed, due to: errorCode: 530, errorText:Failed to update host:2,No next resource state found for current state = Maintenance event = Disable
=== TestName: test_hostha_enable_ha_when_host_disabled | Status : EXCEPTION ===

------------------------

"tests.smoke.test_hostha_kvm.TestHAKVM" name="test_hostha_enable_ha_when_host_disconected" time="14.522"
"checkForState:: expected=Ineligible, actual={haenable : True, hastate : 'Ineligible', haprovider : 'kvmhaprovider'}
]]></system-out></testcase><testcase classname="tests.smoke.test_hostha_kvm.TestHAKVM" name="test_hostha_enable_ha_when_host_in_maintenance" time="303.923" message="Job failed: {accountid : '5e5de944-fe40-11eb-9d50-1e003b000428', userid : '5e5edcd4-fe40-11eb-9d50-1e003b000428', cmd : 'org.apache.cloudstack.api.command.admin.host.PrepareForMaintenanceCmd', jobstatus : 2, jobprocstatus : 0, jobresultcode : 530, jobresulttype : 'object', jobresult : {errorcode : 530, errortext : 'Failed to prepare host for maintenance due to: Host is already in state Maintenance. Cannot recall for maintenance until resolved.'}, jobinstancetype : 'Host', jobinstanceid : '23cb0154-392a-4af3-80e2-3bf0d24bc341', created : '2021-08-16T13:19:01+0000', completed : '2021-08-16T13:19:01+0000', jobid : '946cb1c8-37cb-4110-b5ea-529da895d05e':
  File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor
    yield
  File "/usr/lib64/python3.6/unittest/case.py", line 622, in run
    testMethod()
  File "/marvin/tests/smoke/test_hostha_kvm.py", line 285, in test_hostha_enable_ha_when_host_in_maintenance
    self.setHostToMaintanance(self.host.id)
  File "/marvin/tests/smoke/test_hostha_kvm.py", line 623, in setHostToMaintanance
    response = self.apiclient.prepareHostForMaintenance(cmd)
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackAPI/cloudstackAPIClient.py", line 2435, in prepareHostForMaintenance
    response = self.connection.marvinRequest(command, response_type=response, method=method)
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 381, in marvinRequest
    raise e
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 376, in marvinRequest
    raise self.__lastError
  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 105, in __poll
    % async_response)
Exception: Job failed: {accountid : '5e5de944-fe40-11eb-9d50-1e003b000428', userid : '5e5edcd4-fe40-11eb-9d50-1e003b000428', cmd : 'org.apache.cloudstack.api.command.admin.host.PrepareForMaintenanceCmd', jobstatus : 2, jobprocstatus : 0, jobresultcode : 530, jobresulttype : 'object', jobresult : {errorcode : 530, errortext : 'Failed to prepare host for maintenance due to: Host is already in state Maintenance. Cannot recall for maintenance until resolved.'}, jobinstancetype : 'Host', jobinstanceid : '23cb0154-392a-4af3-80e2-3bf0d24bc341', created : '2021-08-16T13:19:01+0000', completed : '2021-08-16T13:19:01+0000', jobid : '946cb1c8-37cb-4110-b5ea-529da895d05e'}
=== TestName: test_hostha_enable_ha_when_host_in_maintenance | Status : EXCEPTION ===
```

DaanHoogland:
@blueorangutan test

GabrielBrascher:
Thanks for the tests @DaanHoogland. I am checking on them.
Also, I am constantly hammering the HA implementation on a staging CloudStack DC with Redfish OOBM + KVM nodes.

This is interesting. A different result but also impacting the HA tests.

#### Prior tests had:

Test | Result | Time (s) | Test File
--- | --- | --- | ---
test_hostha_enable_ha_when_host_disabled | `Error` | 1.11 | test_hostha_kvm.py
test_hostha_enable_ha_when_host_in_maintenance | `Error` | 303.92 | test_hostha_kvm.py

- case 1, disabled: it was on a "wrong" state (maintenance) which did not allow to disable.
` updatehost failed, due to: errorCode: 530, errorText:Failed to update host:2,No next resource state found for current state = Maintenance event = Disable`
- case 2, Maintenance: These tests were mainly related to a host already in maintenance that could not be replaced as in maintenance.
`Failed to prepare host for maintenance due to: Host is already in state Maintenance.`

I assume that some of the tests in the CI put a host in maintenance and this was not cleaned in time for these 2 tests that needed a host not in Maintenance


#### Recent CI tests
For the new CI spin, the test failing is a different one (I still need to check the logs):

Test | Result | Time (s) | Test File
--- | --- | --- | ---
test_hostha_kvm_host_degraded | `Failure` | 768.71 | test_hostha_kvm.py

nvazquez:
Thanks @GabrielBrascher for updating, please advise if the failure is related to the PR after your investigation

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
@GabrielBrascher does this require additional manual tests?

rhtyd:
@blueorangutan package

debian/control
```diff
@@ -56,3 +56,10 @@ Package: cloudstack-integration-tests
 Architecture: all
 Depends: ${misc:Depends}, cloudstack-marvin (= ${source:Version})
 Description: The CloudStack Marvin integration tests
+
+Package: cloudstack-agent-ha-helper
```
rhtyd, line 60: @GabrielBrascher we'll need similar for CentOS7, 8 and Suse packaging.
Can you advise why we need a separate process, can't the agent do this or a thread in libvirtcomputingresource? And it seems it may create a security issue, is the service port (8080) secured and authenticated?

scripts/vm/hypervisor/kvm/agent-ha-helper.py
```diff
@@ -0,0 +1,126 @@
...
+
+class HTTPServerV6(HTTPServer):
+    address_family = socket.AF_INET6
+
+class CloudStackAgentHAHelper(BaseHTTPRequestHandler):
+    def do_GET(self):
+        if self.path == root_path:
+            libvirt = Libvirt()
+
+            running_vms = libvirt.running_vms()
```
rhtyd, line 62: Can't agents get this information via mgmt server? Or by trying to connect on the neighbour's libvirtd process directly (over ssh/tcp/ssl)?

NuxRo (approved):
ignore this, butter fingered it

NuxRo:
Ok, a couple of questions:

1 - as Rohit asked, why can't we do the checks over SSH instead of running a separate service which might have security and other implications? Hypervisors are already expected to accept incoming SSH from the management server(s).
2 - It's not 100% clear, people will still need to rely on the old NFS HA method, right? Your work merely adds additional checks. Any way we can get rid of NFS at this point?

wido:
> Ok, a couple of questions:
> 
> 1 - as Rohit asked, why can't we do the checks over SSH instead of running a separate service which might have security and other implications? Hypervisors are already expected to accept incoming SSH from the management server(s).

No, we can't expect management servers always to SSH into the hypervisors. We for example have disabled this and the mgmt server can't SSH into the HV.

This helper is there for the case that the CloudStack Agent crashes (for what reason) and the host is disconnected. Via this helper agent we have a second way of checking if the Host is still alive and has running VMs.

If this helper does not respond either we can then fence off the host via OOB (IPMI or Redfish).

Once the Host is Fenced we can safely start the VMs on a different host.

> 2 - It's not 100% clear, people will still need to rely on the old NFS HA method, right? Your work merely adds additional checks. Any way we can get rid of NFS at this point?

This additional method allows us to also provide HA when NOT using NFS. For example when using Ceph/RBD only.

It is very, very, very much recommended to always use IPMI/Redfish with HA so that Hosts can be fenced properly.

NuxRo:
@wido Thanks for the explanation. Ok, so I'll want to do some more testing with non-nfs storage, see how it goes.

BTW, are URLs such as "http://10.0.33.2:8080/check-neighbour/10.0.34.165:8080;" supposed to always return "Down" despite them being up? 
Both my two HVs above do this, but eg http://10.0.33.2:8080 will happily report the running VMs and querying them from each other with curl also yields the expected results. I'm on Ubuntu 20.04.

NuxRo:
@GabrielBrascher Sorry, it was just a typo on my part (going for host:port**;** for some reason instead of host:port .. what a waste of time).

I'll try to get this tested without NFS and see if it works as it should.

rhtyd:
@wido @GabrielBrascher I agree we can't assume mgmt server has SSH access however;
- we should explore options to implement this without introducing a new service (my main concern is from security and upgrade point of view, a lot of people don't like non-essential services running on hypervisor)
- for example, (1) what if I the admin wants to do some maintainance etc which requires stopping of the agent - in that case could your changes cause any side-effect, (2) systemd can be configured (probably already is?) to have this service always start on boot and on-crash/on-error
- agent has a stop command answer it can tell mgmt server why it is stopping - that can be used intelligently to not cause HA led migrations (I haven't checked, probably already-is?)
- if this new service is essential, can it be secured using CA-framework generated certificates so at least the communication is validated (the simplest being server certificate was signed/created against the root CA cert)
- and a global setting/kill-switch for users who don't want/need this additional feature/service (for ex. NFS users?) and have it disabled by default

PaulAngus:
Correct me if I'm wrong, but this looks like its relying on HTTP over 8080?

This will absolutely have to be HTTPS with strong auth at both ends.

rhtyd:
> You are right, this something to be careful about.
> We've configured the service in a way that it always starts on boot and if the process/job is killed for any reason it gets restarted as well. The only way of stopping it is via systemd (e.g. systemctl stop cloudstack-hahelper.service)

Could you maybe explore systemd itself, there are ways to use dependencies and `targets` to ensure the agent is always up and unless explicitly stopped by the admin. For example, there's also restart on failure option (https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=).

> We did not implement such a way of telling that the agent has been "intentionally stopped". This would rely on Admins disabling it on the CloudStack side.
> I will need to add some information in the documentation about how to handle the cluster with this agent.

See above, most admins may not remember about this feature and I wonder is stopping an agent to do maintenance work could cause side-effects. Maybe look at my above suggestion on exploiting systemd features. Docs +1

> I can look into a way of adding CA certificates and validate the communications. For now, it has no such validation; however, it binds only with the node IP in the management network (which in theory is an isolated/secure network).

I think if adding a new service for this feature is unavoidable, we should absolutely (a) have the service use CA-framework issued certificates to serve using secured TLS/SSL certs (i.e. on https), (b) provide a default off option (which you've confirmed exists via a cluster-scope global setting), (c) have firewall-config enabled when the agent either starts (or the service/process starts?) or document on how to use this service (i.e. enable port 8080). (Probably not a good idea to expose whole of libvirtd over network, but one option may involve just exposing libvirtd over tls/ssh to other neighbour hosts https://libvirt.org/remote.html which won't require any additional services).

sureshanaparti:
@blueorangutan package

GabrielBrascher:
@nvazquez I am just running a few tests in order to ensure I am addressing the review from @rohityadavcloud and @PaulAngus.
I will be adding HTTPs + Basic authentication to enhance the security of who is able to retrieve the listed VMs from the hosts.

I hope to have it back to "ready for review" status in 5-10 days.

GabrielBrascher:
@nvazquez rebased to the current main branch.
PR is back to "ready to review"

GabrielBrascher:
@rohityadavcloud @PaulAngus I've addressed your security concerns.

By default, the HA helper service will be provided via HTTP + SSL, with Basic Auth. If one wants to keep it just HTTP, it is still possible to do it via the "insecure" mode of the script.
Service is deployed with a default configuration, but can be changed following the respective arguments:
```
    Optional arguments:
      -h, --help                Show this help message and exit
      -i, --insecure            Allows to run the HTTP server without SSL
      -p, --port PORT           Port to be used by the agent-ha-helper server
      -u, --username USERNAME   Sets the user for server authentication
      -k, --password PASSWORD   Keyword/password for server authentication
```

This requires that both ends (management and KVM agents) are configured properly. The default configuration is set to SSL + Authentication with a default username + password (obviously, admins can easily change it).

With SSL + Authentication, only the management node and configured KVM hosts are able serve and consume this API.

It is important to raise that this API **DOES NOT** allow to run Libvirt commands. It only lists the running VMs, and if the host is reachable (`Up` vs `Down`). If an attacker gets access to such API, it would be able just to collect the number of running VMs.

Also, note that the "secure mode" works only when the KVM  nodes have certificates provided (via `provisionCertificate`), otherwise the service will fail.

I will work on the whole documentation needed to make sure users would be guided, in case this implementation gets merged.

GabrielBrascher:
@blueorangutan package

nvazquez:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,346 @@
...
+
+    /**
+     * Lists VMs on host according to vm_instance DB table. The states considered for such listing are: 'Running', 'Stopping', 'Migrating'.
+     * <br>
+     * <br>
+     * Note that VMs on state 'Starting' are not common to be at the host, therefore this method does not list them.
+     * However, there is still a probability of a VM in 'Starting' state be already listed on the KVM via '$virsh list',
+     * but that's not likely and thus it is not relevant for this very context.
+     */
+    public List<VMInstanceVO> listVmsOnHost(Host host) {
```
nvazquez, line 139: Maybe this method could be outside the client with the DAO and keep `countRunningVmsOnAgent` on it?

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
```diff
@@ -0,0 +1,346 @@
...
+    private static final int WAIT_FOR_REQUEST_RETRY = 2;
+    private static final int MAX_REQUEST_RETRIES = 2;
+    private static final JsonParser JSON_PARSER = new JsonParser();
+    static final String HTTP_PROTOCOL = "http";
+    static final String HTTPS_PROTOCOL = "https";
+    private final static String APPLICATION_JSON = "application/json";
+    private final static String ACCEPT = "accept";
+
+    @Inject
+    private VMInstanceDao vmInstanceDao;
```
nvazquez, line 85: I think some of the logic could be placed on other class (maybe KvmHaHelper) as long with the DB access, and keep the client class simply to interact with the HA agent

nvazquez (commented):
Thanks @GabrielBrascher great work, I have left some comments

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHAConfig.java
```diff
@@ -53,4 +53,32 @@
     public static final ConfigKey<Long> KvmHAFenceTimeout = new ConfigKey<>("Advanced", Long.class, "kvm.ha.fence.timeout", "60",
             "The maximum length of time, in seconds, expected for a fence operation to complete.", true, ConfigKey.Scope.Cluster);
 
+    public static final ConfigKey<Integer> KvmHaWebservicePort = new ConfigKey<Integer>("Advanced", Integer.class, "kvm.ha.webservice.port", "8443",
+            "It sets the port used to communicate with the KVM HA Agent Microservice that is running on KVM nodes. Default value is 8443.",
+            true, ConfigKey.Scope.Cluster);
+
+    public static final ConfigKey<Boolean> IsKvmHaWebserviceEnabled = new ConfigKey<Boolean>("Advanced", Boolean.class, "kvm.ha.webservice.enabled", "false",
```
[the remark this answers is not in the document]
GabrielBrascher: @nvazquez maybe I need to rename or change the description to make it more clear.

This is to disable the KVM HA helper. It is an "on/off" switch that makes CloudStack check or not the added webserver on KVM nodes.

1. When it is set to `false` (**default**) it does not validate via the KVM HA Helper Client, thus it keeps the normal behavior for the CloudStack KVM HA flow.
2. When it is enabled, then it will check the web server's health.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KvmHaAgentClient.java
GabrielBrascher: That makes sense, I will re-check this part of the implementation.
Thanks for the review, @nvazquez!

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHAConfig.java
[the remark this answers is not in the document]
nvazquez: @GabrielBrascher thanks, maybe it can be simply renamed to `kvm.ha.webservice.check.enabled` or similar?

DaanHoogland:
@GabrielBrascher will this still be needed?

GabrielBrascher:
Closing this one for now.

PR #5001, opened 2021-05-06
server: NPE may cause management server to not start

rhtyd:
This NPE may happen when a VM is marked removed in the DB but not its
nics on a shared network. This can usually happen due to a failed
expunged VM or when an admin manually marks a VM as removed in DB but
does not cleanup the nics/network resources.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -2232,7 +2232,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                         VMInstanceVO vmInstance = _vmInstanceDao.findById(vmId);
 
                         // only load running vms. For stopped vms get loaded on starting
-                        if (vmInstance.getState() == State.Running) {
+                        if (vmInstance != null && vmInstance.getState() == State.Running) {
                             VmAndCountDetails vmAndCount = new VmAndCountDetails(vmId, VmIpFetchTrialMax.value());
                             vmIdCountMap.put(nicId, vmAndCount);
                         }


rhtyd:
Travis + pkgs should suffice as it's just a NPE fix
@blueorangutan package

weizhouapache (approved):
code lgtm

Pearl1594 (approved):

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -2232,7 +2232,7 @@ private void loadVmDetailsInMapForExternalDhcpIp() {
                         VMInstanceVO vmInstance = _vmInstanceDao.findById(vmId);
 
                         // only load running vms. For stopped vms get loaded on starting
-                        if (vmInstance.getState() == State.Running) {
+                        if (vmInstance != null && vmInstance.getState() == State.Running) {
```
sureshanaparti, line 2235: @rhtyd possible to remove the instance id association from the Nic if VM doesn't exist? so that this nic is not picked next time.

shwstppr (approved):

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
rhtyd: We shouldn't try to encourage people to do DB hacks and let CloudStack fix their mess, but at least let the management server come up.

sureshanaparti (approved):

weizhouapache:
quick fix for this issue:
```
update cloud.nics join vm_instance vi on nics.instance_id=vi.id set nics.removed=now() where vi.removed is not null and nics.removed is null and nics.network_id in (select networks.id from networks where networks.guest_type="Shared" and networks.removed is NULL);
```

PR #5040, opened 2021-05-26
systemvm: increase apache2 default timeout to 10min from 5mins

rhtyd:
Increases `Timeout` of apache2 to 10mins. This will potentially may be a workaround/fix for template upload that fail due to flaky network or higher timeouts that may result in 504 (gateway timeout) issue.

### Types of changes

### DIFF
--- a/systemvm/debian/opt/cloud/bin/setup/common.sh
+++ b/systemvm/debian/opt/cloud/bin/setup/common.sh
@@ -500,7 +500,7 @@ clean_ipalias_config() {
 }
 
 setup_apache2_common() {
-  sed -i 's/^Timeout.*/Timeout 900/g' /etc/apache2/apache2.conf
+  sed -i 's/^Timeout.*/Timeout 600/g' /etc/apache2/apache2.conf
   sed -i 's/^Include ports.conf.*/# CS: Done by Python CsApp config\n#Include ports.conf/g' /etc/apache2/apache2.conf
   # Disable listing of http://SSVM-IP/icons folder for security issue. see article http://www.i-lateral.com/tutorials/disabling-the-icons-folder-on-an-ubuntu-web-server/
   [ -f /etc/apache2/mods-available/alias.conf ] && sed -i s/"Options Indexes MultiViews"/"Options -Indexes MultiViews"/ /etc/apache2/mods-available/alias.conf
--- a/systemvm/debian/opt/cloud/bin/setup/common.sh
+++ b/systemvm/debian/opt/cloud/bin/setup/common.sh
@@ -500,6 +500,7 @@ clean_ipalias_config() {
 }
 
 setup_apache2_common() {
+  sed -i 's/^Timeout.*/Timeout 900/g' /etc/apache2/apache2.conf
   sed -i 's/^Include ports.conf.*/# CS: Done by Python CsApp config\n#Include ports.conf/g' /etc/apache2/apache2.conf
   # Disable listing of http://SSVM-IP/icons folder for security issue. see article http://www.i-lateral.com/tutorials/disabling-the-icons-folder-on-an-ubuntu-web-server/
   [ -f /etc/apache2/mods-available/alias.conf ] && sed -i s/"Options Indexes MultiViews"/"Options -Indexes MultiViews"/ /etc/apache2/mods-available/alias.conf


rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

DaanHoogland (approved):
clgtm

rhtyd:
@blueorangutan test

wido (approved):
LGTM

GabrielBrascher (approved):
Change LGTM.

andrijapanicsb:
Hopefully, I don't sound wrong - but do we all have an understanding of the Apache Timeout and why it should NOT be increased to some high values? I understand the ACS code is LGTM, but is the change itself LGTM itself, will it cause an easier denial of service on httpd (which is wide open to public / on a public IP range for SSVM), etc?

@rhtyd  I know where this change comes from, but until we verify that this indeed solves that specific user's problem (let's wait for that first...), I don't see how this makes sense. Let's discuss offline, please.

From Apache site:

_The TimeOut directive defines the length of time Apache httpd will wait for I/O in various circumstances:

    When reading data from the client, the length of time to wait for a TCP packet to arrive if the read buffer is empty.

    For initial data on a new connection, this directive doesn't take effect until after any configured AcceptFilter has passed the new connection to the server.
    When writing data to the client, the length of time to wait for an acknowledgement of a packet if the send buffer is full.
    In mod_cgi and mod_cgid, the length of time to wait for any individual block of output from a CGI script.
    In mod_ext_filter, the length of time to wait for output from a filtering process.
    In mod_proxy, the default timeout value if ProxyTimeout is not configured._

rhtyd:
@andrijapanicsb the way template/iso upload is implemented is that SSL termination and cors is done by `apache2` which handovers the uploaded data to SSVM agents' internal port 8210: https://github.com/apache/cloudstack/blob/4.15/systemvm/debian/opt/cloud/bin/setup/secstorage.sh#L62

The apache2 Timeout option can be tuned for client/browser is sending data and max. time it can take. A simple test you can do (I've asked @shwstppr to do to confirm) is to try and upload a template that takes more than 5 mins to upload (we can use browser-based latency, in Chrome -> dev tools -> network tab -> throttle) and see if the timeout option makes any difference.

shwstppr:
@rhtyd Without changing timeout value I was able to upload a template in ~30mins
Started around 1:33pm with 75kbps throttling
Screenshot from 2021-05-28 13-33-48
Removed throttling at 2pm
Screenshot from 2021-05-28 14-00-48
Template got successfully uploaded at 2:02pm
Screenshot from 2021-05-28 14-02-30
Screenshot from 2021-05-28 14-02-33
Probably the timeout affects when no data is received in a span of 5min but certainly not affecting total upload time/response.

andrijapanicsb:
Exactly my point @shwstppr  - "timeout affects when no data is received in a span of 5min" -that is what apache timeout is, in my understanding. I don't think this increase is needed and would propose closing this PR, unless we get to some other conclusion by anyone - because this seems like a flaky user's connection from laptop/PC to public IP/SSVM/Apache

rhtyd:
@andrijapanicsb I think we should increase timeout to 10mins at least (600s) to allow for flaky connections - to make the upload feature more resilient to timeout based failures.

rhtyd:
@andrijapanicsb I'll wait till Monday EOD for your response. You may read why we need to increase it here:
https://kinsta.com/blog/504-gateway-timeout/#configure-apache-or-nginx-settings-properly

rhtyd:
@blueorangutan package

shwstppr:
@blueorangutan test

andrijapanicsb:
Until we prove that this would solve the issue we are internally trying to solve, I don't see how increasing the timeout (which by definition makes httpd more prone to DoS attacks) makes sense. 

Let's confirm with the user, that they tested and it solves the issue for their case, otherwise let's please move this to 4.16 (if/when we confirm it solves anything).

rhtyd:
Milestone moved. 
@andrijapanicsb 
Simply changing timeout won't' cause a dos attack; in fact the ssvm's web server hasn't been tested for that. There is a way to check upload timeout, which is by lowering the default Timeout to 1s in apache2.conf and then try to upload a template. As far as the internal issue is concerned we've concluded that all upload template cases were `consistently` failing took more than 5mins to upload (the default timeout value in apache2.conf is 300s or 5mins).

rhtyd:
@blueorangutan test

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

rhtyd:
Closing this, let's revisit this again in future if somebody shares a potential gateway/504 error on uploading template/iso.
Potential issues in such cases are;
- cloud agent (ssvm) is not able to write to secondary storage (i..e handle browser iso/template upload request) due to poor iops or other resource issues
- cpu/mem/resources are not available/free on the ssvm
- network/switch glitches

PR #5060, opened 2021-06-01
ui: Adding option to choose between name and displayname

davidjumani:
### Description
Adds an option to choose between name and displayname

### Types of changes

- [ ] Major
- [x] Minor

Screenshot from 2021-06-01 15-43-17
Screenshot from 2021-06-01 15-43-24

### DIFF
--- a/ui/src/config/section/compute.js
+++ b/ui/src/config/section/compute.js
@@ -38,7 +38,7 @@ export default {
         return filters
       },
       columns: () => {
-        const fields = ['displayname', 'name', 'state', 'ipaddress']
+        const fields = ['name', 'state', 'ipaddress']
         const metricsFields = ['cpunumber', 'cpuused', 'cputotal',
           {
             memoryused: (record) => {
@@ -47,6 +47,10 @@ export default {
           },
           'memorytotal', 'networkread', 'networkwrite', 'diskkbsread', 'diskkbswrite', 'diskiopstotal'
         ]
+
+        if (store.getters.displayname) {
+          fields.splice(0, 1, 'displayname')
+        }
 
         if (store.getters.metrics) {
           fields.push(...metricsFields)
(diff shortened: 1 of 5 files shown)


rhtyd:
-0 we need to ask what others thing. I like that unnecessary column can be toggled; but don't like the new toggle button (the nav bar is already crowded)

DaanHoogland:
+0 but i agree with @rhtyd , is it possible to make the column header a drop down?

davidjumani:
@DaanHoogland Will try but not sure that is natively supported. @utchoang Any idea how to have a table header cell with a dropdown ?

utchoang:
@davidjumani 
I think you can try to use a custom header with the slot by image (1) and add 1 drop down similar to the header filter by image (2). Then catch the event and reload the table?

davidjumani:
@DaanHoogland @rhtyd Would it be better to have an option to let the user select which columns to be displayed ? And any idea where to place it ?
General idea :

Screenshot from 2021-06-22 10-00-47

DaanHoogland:
@davidjumani do we want to let the user configure this (in a cookie) on individual basis, and/or do we want to let the operator set restrictions?
your ui design looks good at first sight but a setting of defaults could work as well.

davidjumani:
> @davidjumani do we want to let the user configure this (in a cookie) on individual basis, and/or do we want to let the operator set restrictions?
> your ui design looks good at first sight but a setting of defaults could work as well.

Yes, on a per user basis saved in the browser storage

rhtyd:
I like the approach to allow users to select the columns they want to see, but it's still not necesary. Maybe useful for smaller screens such as mobiles. In my opinion this may be closed.

rhtyd:
-1 let's discuss in dev@, the name and displayname may be used different cases by operators/providers; the list of columns are only huge for root admins and not normal users who are only shown a subset of the list of columns.

nvazquez:
@davidjumani I'm removing it from the 4.16 milestone for now. Please discuss it in dev@ and add it back to the milestone if needed

rhtyd:
I'm closing this @davidjumani kindly discuss on dev@ first and reopen in future if we can gather consensus. Thanks.

PR #5085, opened 2021-06-08
Root disk size should be listed in GB at listServiceOffering

GabrielBrascher:
### Description

When listing service offerings it is expected to list their Root disk size in GB (see documentation [[1]](http://cloudstack.apache.org/api/apidocs-4.15/apis/listServiceOfferings.html)); however, it is returned in Bytes.
Additionally, during deployment of a VM or creation of a Service Offering the root disk size is always passed as a value in GB [[2]](https://cloudstack.apache.org/api/apidocs-4.15/apis/deployVirtualMachine.html).

This PR aims to fix this by converting the root disk size at the ServiceOffering response from Bytes to GB.

[1] http://cloudstack.apache.org/api/apidocs-4.15/apis/listServiceOfferings.html
[2] https://cloudstack.apache.org/api/apidocs-4.15/apis/deployVirtualMachine.html

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

N/A

### How Has This Been Tested?

List Service offerings via CloudMonkey (or any other way of calling API). Example: `list serviceofferings filter=rootdisksize`.

#### 1. Currently
~~~
list serviceofferings filter=rootdisksize
rootdisksize
3.221225472e+10
3.221225472e+10
3.221225472e+10
3.221225472e+10
1.073741824e+11
~~~

#### 2. With the fix
~~~
list serviceofferings filter=rootdisksize
rootdisksize
30
30
30
30
100
~~~

### DIFF
--- a/server/src/main/java/com/cloud/api/query/dao/ServiceOfferingJoinDaoImpl.java
+++ b/server/src/main/java/com/cloud/api/query/dao/ServiceOfferingJoinDaoImpl.java
@@ -45,6 +45,13 @@ public class ServiceOfferingJoinDaoImpl extends GenericDaoBase<ServiceOfferingJo
 
     private SearchBuilder<ServiceOfferingJoinVO> sofIdSearch;
 
+    /**
+     * Constant used to convert GB into Bytes (or the other way around).
+     * GB   *  MB  *  KB  = Bytes //
+     * 1024 * 1024 * 1024 = 1073741824
+     */
+    private static final long GB_TO_BYTES = 1073741824;
+
     protected ServiceOfferingJoinDaoImpl() {
 
         sofIdSearch = createSearchBuilder();
@@ -123,7 +130,8 @@ public class ServiceOfferingJoinDaoImpl extends GenericDaoBase<ServiceOfferingJo
             }
         }
 
-        offeringResponse.setRootDiskSize(offering.getRootDiskSize());
+        long rootDiskSizeInGb = (long) offering.getRootDiskSize() / GB_TO_BYTES;
+        offeringResponse.setRootDiskSize(rootDiskSizeInGb);
 
         return offeringResponse;
     }
--- a/ui/src/components/view/DetailsTab.vue
+++ b/ui/src/components/view/DetailsTab.vue
@@ -35,7 +35,7 @@
         </div>
         <div v-else-if="$route.meta.name === 'computeoffering' && item === 'rootdisksize'">
           <div>
-            {{ parseFloat( resource.rootdisksize / (1024.0 * 1024.0 * 1024.0)).toFixed(1) }} GB
+            {{ resource.rootdisksize }} GB
           </div>
         </div>
         <div v-else-if="['name', 'type'].includes(item)">


shwstppr:
@GabrielBrascher this will require change in UI as well. Somewhere here, https://github.com/apache/cloudstack/blob/4.15/ui/src/components/view/DetailsTab.vue#L36-L40

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan test

ui/src/components/view/DetailsTab.vue
```diff
@@ -35,7 +35,7 @@
         </div>
         <div v-else-if="$route.meta.name === 'computeoffering' && item === 'rootdisksize'">
           <div>
-            {{ parseFloat( resource.rootdisksize / (1024.0 * 1024.0 * 1024.0)).toFixed(1) }} GB
+            {{ resource.rootdisksize }} GB
```
GabrielBrascher, line 38: @shwstppr thanks for pointing out the change necessary to keep UI aligned with API.
I believe this will fix it, please let me know if I am missing anything.

GabrielBrascher:
@blueorangutan ui

GabrielBrascher:
It looks like the UI change has been successfully applied.

See http://qa.cloudstack.cloud:8080/client/pr/5085/#/computeoffering/7becb40b-315a-413f-a89a-5a8b8c6b01b0

image

One can compare with how it was (e.g. at this PR http://qa.cloudstack.cloud:8080/client/pr/5080/#/computeoffering/7becb40b-315a-413f-a89a-5a8b8c6b01b0)

image

rhtyd (approved):

Pearl1594 (approved):
LGTM

server/src/main/java/com/cloud/api/query/dao/ServiceOfferingJoinDaoImpl.java
```diff
@@ -45,6 +45,13 @@
 
     private SearchBuilder<ServiceOfferingJoinVO> sofIdSearch;
 
+    /**
+     * Constant used to convert GB into Bytes (or the other way around).
+     * GB   *  MB  *  KB  = Bytes //
+     * 1024 * 1024 * 1024 = 1073741824
+     */
+    private static final long GB_TO_BYTES = 1073741824;
```
sureshanaparti, line 53: multiple constants are defined, in various classes, for the value "1024 * 1024 * 1024", and it is hardcoded in some cases. may be (later), a constant can be defined in common package( eg. cloud utils), and use it across. (Such common util functions has to be documented, so that the developers can use these)

Also, some get disk size calls return size in Bytes, and others return in GB. I think, it is better to explicitly mention (in variables / methods) it indicating Bytes or GB for non-default unit (i.e. If Bytes is the default unit, then methods which return in 'GB' should have it mentioned). If there is no such default unit, then variables / methods should indicate the size units explicitly. This can be a huge refactoring work (so can be skipped for now) and at least, have to make sure new code indicates proper size unit.

GabrielBrascher: @sureshanaparti I totally agree with you. We had many issues in the past and it still haunts us. If we simply added a proper name and documentation it would help a lot many bugs that we face and also make new features and extensions easier on that matter.

Regarding the constant. You have a good point; we could create constants/enums for such conversions as many times we store in Bytes but make it available to users in GB or MB.

PR #5119, opened 2021-06-16
Externalize tls version and security protocols configuration on mail sending

GutoVeronezi:
### Description

- Some servers (such as [Microsoft](https://docs.microsoft.com/en-us/microsoft-365/compliance/prepare-tls-1.2-in-office-365?view=o365-worldwide) exchange) are deprecating TLSv1 and TLSv1.1. On alert's and project's mailing settings, ACS uses the protocol TLSv1; therefore, operators do not have options to choose which version they want to use.

- ACS has configurations (`\*.smtp.useAuth`) to inform if it will use secure SMTP authentication when sending emails; However, this configuration only refers to the use of SSL. Operators have no means to choose if they want to use StartTLS option.

This PR intends to externalize the protocol's and StartTLS settings on alert's and project's mail sending. Alert and project (also quota) send email through the class `SMTPMailSender`, inserted in PR [4954](https://github.com/apache/cloudstack/pull/4954), which will handle these configurations.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?
It was tested in a test lab.

I set SMTP global settings to project and alert and enabled StartTLS (global setting); Then I allowed `Less secure app access` on my Google account and set the following settings:
```
server address: smtp.gmail.com
username: <gmail address>
password: <gmail password>
port (TLS): 587
```
- To alert, I started mgmt;
- To project, I enabled global setting `project.invite.required` and added a user via email;

In Gmail, we have an option to `show original` message, which provides the TLS version and the cipher. So I changed the configuration a few times and sent some alerts and invitations to verify if both versions, in configuration and Gmail, were the same.

### DIFF
--- a/engine/components-api/src/main/java/com/cloud/alert/AlertManager.java
+++ b/engine/components-api/src/main/java/com/cloud/alert/AlertManager.java
@@ -35,6 +35,12 @@ public interface AlertManager extends Manager, AlertService {
         "Alert", "0.75", "Percentage (as a value between 0 and 1) of allocated storage utilization above which alerts will be sent about low storage available.", true,
         ConfigKey.Scope.Cluster, null);
 
+    public static final ConfigKey<Boolean> AlertSmtpUseStartTLS = new ConfigKey<Boolean>("Advanced", Boolean.class, "alert.smtp.useStartTLS", "false",
+            "If set to true and if we enable security via alert.smtp.useAuth, this will enable StartTLS to secure the conection.", true);
+
+    public static final ConfigKey<String> AlertSmtpEnabledSecurityProtocols = new ConfigKey<String>("Advanced", String.class, "alert.smtp.enabledSecurityProtocols", "",
+            "White-space separated security protocols; ex: \"TLSv1 TLSv1.1\". Supported protocols: SSLv2Hello, SSLv3, TLSv1, TLSv1.1 and TLSv1.2", true);
+
     void clearAlert(AlertType alertType, long dataCenterId, long podId);
 
     void recalculateCapacity();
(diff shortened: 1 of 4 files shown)


rhtyd:
@blueorangutan package

DaanHoogland:
@GutoVeronezi It makes sense to make this configuratble, but should we allow it to be different for alert and project? seems like it is a true global setting.

GutoVeronezi:
> @GutoVeronezi It makes sense to make this configuratble, but should we allow it to be different for alert and project? seems like it is a true global setting.

@DaanHoogland I am following the way mail settings are implemented today; each scope (quota, alert and project) has its own settings. Although, I don't see any plausible reason to keep settings to each scope (it seems that an operator could use different providers to send mail is a remote probability).

I think that if we turn these global, we should turn those too, to maintain the same pattern.

DaanHoogland:
> I think that if we turn these global, we should turn those too, to maintain the same pattern.

makes sense.

GutoVeronezi:
Can anyone review this?

RodrigoDLopez (approved):
Nice work @GutoVeronezi 
The Code LGTM, i will try run some manual tests. But i'm not sure if i can do this in my env.

rhtyd:
@blueorangutan package

DaanHoogland (approved):
clgtm

GutoVeronezi:
@DaanHoogland @RodrigoDLopez is there anything else to do?

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan package

nvazquez (approved):
LGTM

nvazquez:
@GutoVeronezi as the PR is only exposing config keys for existing properties I think a round of tests will be enough as there are no changes in the functionality

nvazquez:
@blueorangutan test

GabrielBrascher (approved):
Code LGTM

nvazquez:
Merging based on approvals and test results as failures are intermitent issues

PR #5141, opened 2021-06-22
Add selection for podman/docker

Niklas-Koch:
### Description

This PR gives you the option to choose between podman and docker in the docker.sh script.
I came up with this idea because Red Hat and the most Red Hat-Distribution does not allow docker anymore. Instead they use podman. 
Now you can choose between both options while entering a -p for podman or a -d for docker after the ./docker.sh command.

### Types of changes

### How Has This Been Tested?

I tested it on our testenvironment in my company.
I also tested both options, including the case that you do not choose any option.

### DIFF
--- a/ui/docker.sh
+++ b/ui/docker.sh
@@ -35,5 +35,6 @@ do
         case "$OPTION" in
                 p)podman build -t cloudstack-ui ${LABEL_DATE} ${LABEL_GIT_REV} ${LABEL_GIT_TAG};;
                 d)docker build -t cloudstack-ui ${LABEL_DATE} ${LABEL_GIT_REV} ${LABEL_GIT_TAG};;
+                *)echo "Please choose between -p for podman or -d for docker.";;
         esac
 done
--- a/ui/docker.sh
+++ b/ui/docker.sh
@@ -30,4 +30,10 @@ LABEL_DATE="--label \"org.opencontainers.image.created=${DATE}\""
 GIT_REV="$(git rev-parse HEAD)"
 LABEL_GIT_REV="--label \"org.opencontainers.image.revision=${GIT_REV}\""
 
-docker build -t cloudstack-ui ${LABEL_DATE} ${LABEL_GIT_REV} ${LABEL_GIT_TAG} .
+while getopts $OPTIONS OPTION
+do
+        case "$OPTION" in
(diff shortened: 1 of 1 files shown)


weizhouapache:
@Niklas-Koch 
for backwards compatibility, can we use docker by default ?

svenvogel:
> @Niklas-Koch
> for backwards compatibility, can we use docker by default ?

Niklas-Koch:
@weizhouapache 
Good idea, 
So I changed the file, that docker is the default value now, to hold the backwards compatibility. 
Now you have the possibility to use the script without any options and it will use docker as standard. But you also can you the script with an option like -d for docker (it will use docker the same way as without using the option) or -p for podman. So we have the opportunity to use podman. That means we can use the podman, which is getting more and more used and at the same time, we maintained the backwards compatibility.
I think it is now looking good to me.
Can you look through it?

ui/docker.sh
```diff
@@ -29,5 +29,20 @@ DATE="$(date --iso-8601=seconds)"
...
+while getopts $OPTIONS OPTION
+do
+        case "$OPTION" in
+                p)USE_PODMAN=1;;
+                d)USE_DOCKER=1;;
+                *)echo "Please choose between -p for podman or -d for docker.";;
+        esac
+done
+if [[ $USE_PODMAN -eq 1 ]]; then
+        podman build -t cloudstack-ui ${LABEL_DATE} ${LABEL_GIT_REV} ${LABEL_GIT_TAG}
```
j-porsche, line 44: @Niklas-Koch You missed the point at the end of the line 47
```suggestion
        podman build -t cloudstack-ui ${LABEL_DATE} ${LABEL_GIT_REV} ${LABEL_GIT_TAG} .
```

ui/docker.sh
```diff
@@ -29,5 +29,20 @@ DATE="$(date --iso-8601=seconds)"
...
+        exit
+fi
+docker build -t cloudstack-ui ${LABEL_DATE} ${LABEL_GIT_REV} ${LABEL_GIT_TAG}
```
j-porsche, line 47: @Niklas-Koch You missed the point at the end of the line 47

```suggestion
docker build -t cloudstack-ui ${LABEL_DATE} ${LABEL_GIT_REV} ${LABEL_GIT_TAG} .
```

DaanHoogland:
@Niklas-Koch @j-porsche closing this as #5217 supercedes it. please re-open if I'm mistaking

PR #5144, opened 2021-06-23
configdrive: fix some failures in tests/component/test_configdrive.py

weizhouapache:
### Description

This PR fixes a series of bug when test test_configdrive.py

(1) detachAnd Attach ConfigDrive ISO on kvm, in some cases configdrive ISO is updated, for example, plug a new nic, or migrate vm.

(2) use "copy" instead of "mv" when create configdrive ISO in SSVM, as configdrive ISO will disappear in VM if "mv" is used.

(3) fix two exceptions when enable/disable static nat on vms with configdrive ISO. (public IP in configdrive ISO is updated in these two cases)

(4) fix two issues in setup/bindir/cloud-set-guest-sshkey-configdrive.in 

(5) fix few bugs in test/integration/component/test_configdrive.py

(6) build and use centos55 with both sshkey and configdrive support for testing.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

tested on with kvm.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -303,6 +303,8 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     protected String _agentHooksVmOnStopScript = "libvirt-vm-state-change.groovy";
     protected String _agentHooksVmOnStopMethod = "onStop";
 
+    private static final String CONFIG_DRIVE_ISO_DISK_LABEL = "hdd";
+    private static final int CONFIG_DRIVE_ISO_DEVICE_ID = 4;
 
     protected File _qemuSocketsPath;
     private final String _qemuGuestAgentSocketName = "org.qemu.guest_agent.0";
@@ -2754,6 +2756,32 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
 
     protected KVMStoragePoolManager getPoolManager() {
         return _storagePoolMgr;
+    }
+
+    public void detachAndAttachConfigDriveISO(final Connect conn, final String vmName) {
+        // detach and re-attach configdrive ISO
+        List<DiskDef> disks = getDisks(conn, vmName);
+        DiskDef configdrive = null;
+        for (DiskDef disk : disks) {
+            if (disk.getDeviceType() == DiskDef.DeviceType.CDROM && disk.getDiskLabel() == CONFIG_DRIVE_ISO_DISK_LABEL) {
+                configdrive = disk;
+            }
+        }
+        if (configdrive != null) {
+            try {
+                String result = attachOrDetachISO(conn, vmName, configdrive.getDiskPath(), false, CONFIG_DRIVE_ISO_DEVICE_ID);
+                if (result != null) {
+                    s_logger.warn("Detach ConfigDrive ISO with result: " + result);
+                }
+                result = attachOrDetachISO(conn, vmName, configdrive.getDiskPath(), true, CONFIG_DRIVE_ISO_DEVICE_ID);
+                if (result != null) {
+                    s_logger.warn("Attach ConfigDrive ISO with result: " + result);
+                }
+            } catch (final LibvirtException | InternalErrorException | URISyntaxException e) {
+                final String msg = "Detach and attach ConfigDrive ISO failed due to " + e.toString();
+                s_logger.warn(msg, e);
+            }
+        }
     }
 
     public synchronized String attachOrDetachISO(final Connect conn, final String vmName, String isoPath, final boolean isAttach, final Integer diskSeq) throws LibvirtException, URISyntaxException,
--- a/scripts/storage/secondary/createvolume.sh
+++ b/scripts/storage/secondary/createvolume.sh
@@ -135,7 +135,8 @@ create_from_file() {
   local tmpltname=$3
 
   [ -n "$verbose" ] && echo "Moving to $tmpltfs/$tmpltname...could take a while" >&2
-  mv $tmpltimg /$tmpltfs/$tmpltname
+  cp -rf $tmpltimg /$tmpltfs/$tmpltname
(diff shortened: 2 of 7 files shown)


rhtyd:
Should we consider that for 4.15.1?

rhtyd:
@blueorangutan package

weizhouapache:
> Should we consider that for 4.15.1?

@rhtyd 
these are bug fixes for some failures in tests/component/test_configdrive.py
(the testing is part of my POC of configdrive feature for VR agent)

They are not critical issues I think. we can move it to 4.15.2.0 or 4.16.0.0.

The testing still fails in testing of isolated networks and vpcs for now (shared networks works). Very few users use configdrive for vms on isolated networks and vpcs.

rhtyd:
@weizhouapache We know several users who use config drive feature esp for L2 networks and in VR-less networks, I'm planing to cut RC3 on Monday. Would this PR be ready before that?

scripts/storage/secondary/createvolume.sh
```diff
@@ -135,7 +135,8 @@ create_from_file() {
   local tmpltname=$3
 
   [ -n "$verbose" ] && echo "Moving to $tmpltfs/$tmpltname...could take a while" >&2
-  mv $tmpltimg /$tmpltfs/$tmpltname
+  cp -rf $tmpltimg /$tmpltfs/$tmpltname
```
rhtyd, line 138: mv maybe faster?

rhtyd:
I guess the question is - is the feature broken or just the tests?

weizhouapache:
> @weizhouapache We know several users who use config drive feature esp for L2 networks and in VR-less networks, I'm planing to cut RC3 on Monday. Would this PR be ready before that?

@rhtyd it is difficult. 

> I guess the question is - is the feature broken or just the tests?

@rhtyd the feature is not broken.
however, there are several test failures in tests/component/test_configdrive.py, which means bugs in some scenarios.

what I have fixed in this pr for now
(1) cannot enable static nat for vms with configdrive, as apply userdata fails.
(2) vms are stuck at expunging, as disable static nat fails.
(3) configdrive ISO is not recognized in vm when add second nic to the vm.

scripts/storage/secondary/createvolume.sh
weizhouapache: @rhtyd y, mv is faster, however for configdrive ISO, if it is attached to a vm, it will not be recognized in vm after "mv"
there is no issue if "cp" is used.

I am thinking about (1) not change this script, but copy configdrive ISO in java instead of calling this script (createvolume.sh); (2) detach and re-attach configdrive ISO when it is changed.

this is a draft PR, still WIP.

weizhouapache:
> @weizhouapache We know several users who use config drive feature esp for L2 networks and in VR-less networks, I'm planing to cut RC3 on Monday. Would this PR be ready before that?

@rhtyd I do not have an exact date when all issues are fixed and tests pass. it seems very difficult to be ready before next monday. it is better to move it to 4.15.2.0

rhtyd:
cc @Pearl1594 @weizhouapache does it break any basic config drive feature if tested manually, say config drive with ssh support and user data on L2 and isolated networks (without VR)?

weizhouapache:
> cc @Pearl1594 @weizhouapache does it break any basic config drive feature if tested manually, say config drive with ssh support and user data on L2 and isolated networks (without VR)?

@rhtyd from what I have tested, there are no issues with following vm actions
(1) deploy
(2) stop/start
(3) reset password
(4) reset sshkeypair
(5) update userdata
maybe it is good enough for L2 networks.

What might not work (some are confirmed, some are not)
(1) enable static nat, disable static nat
(2) plug nic, unplug nic, update default nic
(3) in vpc

rhtyd:
Thanks for confirming @weizhouapache I'll move it to 4.15.2

rhtyd:
@weizhouapache is this ready for review (PR is in draft)

weizhouapache:
> @weizhouapache is this ready for review (PR is in draft)

@rhtyd @DaanHoogland 
updated this pr to fix some bugs (see description).

please note I only tested with kvm, not xenserver.

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

DaanHoogland:
@blueorangutan test centos7 xenserver-72

DaanHoogland:
@blueorangutan test centos7 xenserver-74

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2756,6 +2758,32 @@ protected KVMStoragePoolManager getPoolManager() {
...
+    public void detachAndAttachConfigDriveISO(final Connect conn, final String vmName) {
+        // detach and re-attach configdrive ISO
+        List<DiskDef> disks = getDisks(conn, vmName);
+        DiskDef configdrive = null;
+        for (DiskDef disk : disks) {
+            if (disk.getDeviceType() == DiskDef.DeviceType.CDROM && disk.getDiskLabel() == CONFIG_DRIVE_ISO_DISK_LABEL) {
+                configdrive = disk;
+            }
+        }
+        if (configdrive != null) {
```
GutoVeronezi, line 2770: We could invert the logic here to reduce block complexity.

rhtyd:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
DaanHoogland: I like your thinking @GutoVeronezi , but in this case that would leave a return in the middle of the method. I'd rather extract the block into an extra method.

DaanHoogland (approved):
cltgm, going to test this (starting) today

rhtyd (approved):
+1 based on Daan's remarks/testing

PR #5149, opened 2021-06-23
Refactor and improvements for method com.cloud.hypervisor.kvm.resource.LibvirtComputingResource.createVMFromSpec()

SadiJr:
### Description

This PR goal is to refactor the method createVMFromSpec(final VirtualMachineTO vmTO) in class com.cloud.hypervisor.kvm.resource.LibvirtComputingResource and improve its logs, javadocs, while adding unit tests for all of the code that is extracted and modified.
### Types of changes

### How Has This Been Tested?
All unit tests have been tested, and I've checked that the results on the main branch are the same of this refactor.


DaanHoogland (commented):
great improvement in readability. I am missing how you verified that the "results are the same", as you say. Did you test all kinds of configurations i.e. with config drive and in different kinds of networks (L2, VPC). Can you expand on that please?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2239,211 +2305,303 @@ public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
...
-            } else {
-                if (_uefiProperties.getProperty(GuestDef.GUEST_NVRAM_TEMPLATE_LEGACY) != null) {
-                    guest.setNvramTemplate(_uefiProperties.getProperty(GuestDef.GUEST_NVRAM_TEMPLATE_LEGACY));
-                }
-            }
+     // Add extra configuration to User VM Domain XML before starting
+        if (MapUtils.isNotEmpty(extraConfig) && VirtualMachine.Type.User.equals(vmTO.getType())) {
+            s_logger.debug(String.format("Appending extra configuration data [%s] to guest VM [%s] domain XML.", extraConfig, vmTO.toString()));
+            addExtraConfigComponent(extraConfig, vm);
+        }
```
GutoVeronezi, line 2364: We can extract this code to a method and turn the comment into a javadoc.

SadiJr:
@DaanHoogland I literally debug the existing tests in main and checked the VM XML settings for each test. And I repeated this process on the branch where I made the changes to compare the results.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
SadiJr: Done. Thanks for the tip.

DaanHoogland:
> @DaanHoogland I literally debug the existing tests in main and checked the VM XML settings for each test. And I repeated this process on the branch where I made the changes to compare the results.

tnx, code looks good, but need testing.

GutoVeronezi (approved):
CLGTM

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2239,211 +2305,309 @@ public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
...
 
-             This change is due to the overcommit feature in 4.2
-             */
-            if (vmTO.getMinSpeed() != null) {
-                ctd.setShares(vmTO.getCpus() * vmTO.getMinSpeed());
-            } else {
-                ctd.setShares(vmTO.getCpus() * vmTO.getSpeed());
-            }
+    /**
+     * Add the VNC port passwd here, get the passwd from the vmInstance.
```
GabrielBrascher, line 2426: I think that this Javadoc does not entirely match with what is done (please correct me if I am wrong)

`Add the VNC port passwd here, get the passwd from the vmInstance.`

Suggestion :arrow_down:

`Given a VM (VirtualMachineTO), it creates a Libvirt Graphic Definition with the VM's password and VNC address.`
_OR simply:_
`Creates a Libvirt Graphic Definition with the VM's password and VNC address.`

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2239,211 +2305,309 @@ public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
...
+     */
+    protected GraphicDef createGraphicDef(VirtualMachineTO vmTO) {
+        return new GraphicDef(VNC, (short)0, true, vmTO.getVncAddr(), vmTO.getVncPassword(), null);
+    }
 
-            setQuotaAndPeriod(vmTO, ctd);
+    /**
+     * Add a VirtIO channel for the Qemu Guest Agent tools.
```
GabrielBrascher, line 2433: `VirtIO` -> `Virtio`

See:
- https://wiki.libvirt.org/page/Virtio
- https://libvirt.org/formatdomain.html#virtio-transitional-devices

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2239,211 +2305,309 @@ public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
...
+    protected ChannelDef createChannelDef(VirtualMachineTO vmTO) {
+        File virtIoChannel = Paths.get(_qemuSocketsPath.getPath(), vmTO.getName() + "." + _qemuGuestAgentSocketName).toFile();
+        return new ChannelDef(_qemuGuestAgentSocketName, ChannelDef.ChannelType.UNIX, virtIoChannel);
+    }
 
-            vm.addComp(ctd);
-        }
+    /**
+     * If we're using virtio scsi, then we need to add a virtual scsi controller.
+     */
```
GabrielBrascher, line 2442: I would suggest changing it to:
```
    /**
     * Creates Virtio SCSI controller.
     */
```
     
I am not sure if it is relevant to mention `If we're using virtio scsi, then we need to add a virtual scsi controller`. The method logic is restricted to creating a SCSI XML definition.

The logic that defines if it does create (or not) such definition is:
     ~~~
        if (busT == DiskDef.DiskBus.SCSI) {
            devices.addDevice(createSCSIDef(vcpus));
        }
     ~~~
     
But if you think it still relevant, I would then suggest adding the respective SCSI validation to this method.

In such case, here follows a suggestion (if you think it is better) :-)
```
    /**
     * Creates Virtio SCSI controller. <br>
     * The respective Virtio SCSI XML definition is generated only if the VM's Disk Bus is of ISCSI.
     */
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2239,211 +2305,309 @@ public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
...
+        ClockDef clock = new ClockDef();
+        if (org.apache.commons.lang.StringUtils.startsWith(vmTO.getOs(), WINDOWS)) {
             clock.setClockOffset(ClockDef.ClockOffset.LOCALTIME);
-            clock.setTimer("hypervclock", null, null);
-        } else if (vmTO.getType() != VirtualMachine.Type.User || isGuestPVEnabled(vmTO.getOs())) {
-            if (_hypervisorLibvirtVersion >= 9 * 1000 + 10) {
-                clock.setTimer("kvmclock", null, null, _noKvmClock);
-            }
+            clock.setTimer(HYPERVCLOCK, null, null);
+        } else if ((vmTO.getType() != VirtualMachine.Type.User || isGuestPVEnabled(vmTO.getOs())) && _hypervisorLibvirtVersion >= MIN_LIBVIRT_VERSION_FOR_GUEST_CPU_MODE) {
```
GabrielBrascher, line 2468: This if is quite big, maybe extract it to a method?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2239,211 +2305,309 @@ public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
...
+        return AARCH64.equals(_guestCpuArch);
+    }
 
-        // If we're using virtio scsi, then we need to add a virtual scsi controller
-        if (busT == DiskDef.DiskBus.SCSI) {
-            final SCSIDef sd = new SCSIDef((short)0, 0, 0, 9, 0, vcpus);
-            devices.addDevice(sd);
-        }
+    /**
+     * Create a guest definition from a VM specification.
```
GabrielBrascher, line 2564: Some documentations have the verb as Imperative (e.g. **create**), some Indicative (e.g. it **creates**).
This is a small observation, but maybe you can normalize them. (I would go to the indicative form; but feel free to choose)

GabrielBrascher (commented):
@SadiJr some adjustments are necessary. I've left some comments, please take a look at them.

I want to be clear that I am surprised (in a good way) by the quality of this refactoring.

Kudos for 1) documentation, 2) code extraction to concise methods, and 3) addition of JUnit tests covering some key parts of the execution flow.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
SadiJr: I can't understand what would be the use of extracting this if for a method, considering that it is not used anywhere else. Could you explain your point?

SadiJr:
@GabrielBrascher Thanks for the suggestions. Most have been implemented, although there is a suggestion that I don't quite understand how useful it is.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
GabrielBrascher: That's a good question @SadiJr. In fact, some code extractions have little to do with duplicated code but with making cleaner code and also helping with unit test coverage.

For instance, there are cases where conditionals can be replaced by polymorphism. In some cases, it is also possible to replace a conditional with map/set. As an example:
```
if (var == value1 || var == value2 || ... var == valueN) {
    <body>
}
```
:arrow_down: 
```
/**
 * This JavaDoc might help. But if the variable name is self-explanatory this Javadoc might be unnecessary.
 */
final static Set<Type> GOOD_VAR_NAME = new HashSet<>(Arrays.asList(value1, value2, ..., valueN));
....
protected void method() {
   ....
   if (GOOD_VAR_NAME.contains(var)) {
       <body>
   }
}
```

Therefore, the goal of such extraction would be to:
1. keep code clean
2. add documentation (or not)
3. allow easy JUnit test that covers such if-conditional

You can proceed with the code as it is or look for some alternatives that might improve this IF. I would not block or judge if you keep it as it is. What matters the most is to always try to keep such care when developing/reviewing.

SadiJr: Now I get your point. However, while I don't see any problem making this change, I also don't see the need to make it, so I'll keep the code as is. But thanks for the clarification, it was a great explanation.

api/src/main/java/com/cloud/agent/api/to/VirtualMachineTO.java
```diff
@@ -413,4 +413,9 @@ public DeployAsIsInfoTO getDeployAsIsInfo() {
     public void setDeployAsIsInfo(DeployAsIsInfoTO deployAsIsInfo) {
         this.deployAsIsInfo = deployAsIsInfo;
     }
+
+    @Override
+    public String toString() {
+        return String.format("VM {\"id\": %s, \"name\": \"%s\", \"uuid\": \"%s\", \"type\": \"%s\"}", id, name, uuid, type);
```
GabrielBrascher, line 419: A "cosmetic" observation/question.

This way it will print: `VM {"id": 123, "name": my VM, "uuid": 73ec..uuid, "type": User}`

Wouldn't the idea be printing the following?
`VM {id: "123", name: "my VM", uuid: "73ec..uuid", type: "User"}`

If the second option is desired then it would need to change to:
`return String.format("VM {id: \"%s\", name: \"%s\", uuid: \"%s\", type: \"%s\"}", id, name, uuid, type);`

SadiJr: Good observation, done.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2239,215 +2309,314 @@ public LibvirtVMDef createVMFromSpec(final VirtualMachineTO vmTO) {
...
-             A 4.0.X/4.1.X management server doesn't send the correct JSON
-             command for getMinSpeed, it only sends a 'speed' field.
+    protected WatchDogDef createWatchDogDef() {
+        return new WatchDogDef(_watchDogAction, _watchDogModel);
+    }
 
-             So if getMinSpeed() returns null we fall back to getSpeed().
+    /*
+     * Adds an explicit USB devices for ARM64
+     */
```
DaanHoogland, line 2418: this comment doesn't add much to the contents of the method. Is it really needed?

DaanHoogland (commented):
clgtm and good volume of unit tests added. I do wonder if the continued support for LXC is realistic, though.

SadiJr:
@DaanHoogland Thanks for the review. 

I agree with your opinion about the LXC support, but I think is better to discuss this in an issue or an PR specific for it.

GabrielBrascher (approved):
Code LGTM

nvazquez:
Looks really nice @SadiJr. 
@davidjumani this PR will require some effort to test on KVM ensuring there are no regressions, let's try to find some time for testing it

davidjumani:
@blueorangutan package

davidjumani (commented):
CLGTM. Will test and report

davidjumani:
@blueorangutan test keepEnv

davidjumani (approved):
Tested against main, similar xmls created

nvazquez (approved):
LGTM

PR #5206, opened 2021-07-14
UI: only display host information, if they are relevant

mib1185:
### Description

This PR improves the HostView, so that only relevant information is shown.
This also fixes a small bug, when no `hosttags` or `oscategoryid` exists, that wrong (_maybe cached?_) values were shown, instead of no values.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### DIFF
--- a/ui/src/views/infra/HostInfo.vue
+++ b/ui/src/views/infra/HostInfo.vue
@@ -43,7 +43,7 @@
           </div>
         </div>
       </a-list-item>
-      <a-list-item>
+      <a-list-item v-if="host.hosttags">
         <div>
           <strong>{{ $t('label.hosttags') }}</strong>
           <div>
@@ -51,7 +51,7 @@
           </div>
         </div>
       </a-list-item>
-      <a-list-item>
+      <a-list-item v-if="host.oscategoryid">
         <div>
           <strong>{{ $t('label.oscategoryid') }}</strong>
           <div>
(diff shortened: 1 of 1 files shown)


mib1185:
@blueorangutan ui

DaanHoogland:
@mib1185 can you add test info/screen shots before/after?
cltgm

ui/src/views/infra/HostInfo.vue
```diff
@@ -83,15 +83,15 @@
           </div>
         </div>
       </a-list-item>
-      <a-list-item v-if="host.hostha">
+      <a-list-item v-if="host.hostha && host.hostha.haenable">
```
davidjumani, line 86: Don't think this should be changed so that the user can explicitly see whether ha is enabled or not

mib1185: the `host.hostha.haenable` state is already  shown above, so the `host.hostha.hastate` should only be relevant, if `haenable` is True

mib1185:
> @mib1185 can you add test info/screen shots before/after?
> cltgm

@DaanHoogland I'm sorry, but I'm on vacation the next three weeks and those do not have access to our environment

ui/src/views/infra/HostInfo.vue
davidjumani: Ah yes, my bad

davidjumani (approved):

utchoang (approved):

nvazquez (approved):

Issue #3121, opened 2019-01-07
Paging isn't fully implemented in WebUI


khmarochos:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
WebUI
~~~

##### CLOUDSTACK VERSION
~~~
4.11.2, maybe all prior versions too
~~~

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### SUMMARY
When one has more that `N` (where `N` equals to `default.page.size`) subdomains, it becomes impossible to see them all in the list of domains, as the paging feature isn't fully implemented in the WebUI frontend application.

##### STEPS TO REPRODUCE
Just create over `N` subdomains and you'll see that:
![image](https://user-images.githubusercontent.com/5197757/50765562-ee051300-127e-11e9-9a2a-68f19e05c064.png)
![image](https://user-images.githubusercontent.com/5197757/50765585-02e1a680-127f-11e9-9dcb-878f4c97f988.png)

##### EXPECTED RESULTS
There should be 999 subdomains (from A00000001 to A00000999)

##### ACTUAL RESULTS
Only 500 domains are shown (500 is the value of `default.page.size` now).

andrijapanicsb:
@shwstppr is this something you could pick up?

shwstppr:
@andrijapanicsb will try to have a look at it during weekend or next week

rhtyd:
I think we've addressed this in Primate cc @davidjumani - can you comment?

davidjumani:
@rhtyd Nope, just the default page, but in some places, we've added the option to search. Have to look into this

nvazquez:
@shwstppr @davidjumani can you confirm if this is fixed please?

shwstppr:
@nvazquez I feel this is still not completely fixed. At many places in UI where we show select element it doesn't allow filtering with keyboard input.
Also, many forms (such as AttachVolume) allow selecting items such as instance, domain using select. An env can have 1000s of instances, domains, templates or other resources and we are trying to list them all at once which might be an issue for some.

PR #5276, opened 2021-08-04
ui: Fetch all records in list* API and allow searching in dropdowns

davidjumani:
### Description

Fixes https://github.com/apache/cloudstack/issues/3121, also #5338 

Ensures that all records are fetched for any list* API when page is not part of the parameters passed

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### How Has This Been Tested?
TODO


davidjumani:
Please ignore the console.log and the pagesize set to 10. Was done deliberately so that anyone reviewing can see it in action rather than creating 1000 records for any item

davidjumani:
@blueorangutan ui

rhtyd:
Wouldn't it make the UI slow?

davidjumani:
@rhtyd Hardly, will set the pagesize to 500 (default) so that most calls won't need to be repeated

ui/src/api/index.js
```diff
@@ -29,14 +29,73 @@ export function api (command, args = {}, method = 'GET', data = {}) {
...
+
+  const pagesize = 10
+  let page = 1
+  let items = []
+  let done = false
+  let response = null
+
+  while (!done) {
+    args.page = page
+    args.pagesize = pagesize
```
shwstppr, line 53: @davidjumani do we use `default.ui.page.size` global setting in the UI?

And just want to know if I have 1000 VMs in my env and I open AttachVolume form will this call listVirtualMachines API 100 times?

utchoang (approved):
It seems ok as long as the number of API calls is not too many times it will slow down the system

ui/src/api/index.js
davidjumani: 10 is a placeholder now since `default.ui.page.size` is generally 500 and it would be difficult to see it in the test env. Will change it to `default.ui.page.size` after people test and get back

davidjumani:
@blueorangutan ui

davidjumani:
@blueorangutan ui

davidjumani:
Closing to separate into distinct PRs

PR #5310, opened 2021-08-12
Support docker builds for ubuntu 20.04

tetra12:
### Description

This PR

Updates debian build rules to support custom build flow with Docker and use of the latest Node and Maven versions.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

deps

 

### How Has This Been Tested?

1. Checked out the latest CloudStack master
2. Built a docker builder image for Ubuntu 20.04 and published it [here](https://hub.docker.com/r/tampler/cs-deb)
3. Created the `deb` files for Ubuntu 20.04 with the following command: `docker run -v $PWD/cloudstack:/mnt/build/cloudstack tampler/cs-deb:2004 /usr/local/bin/docker-entrypoint.sh`
4. Uploaded *.deb binaries and installed on a cloud server
5. Launched CloudStack mgmt server and GUI
screen gui

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -2,8 +2,7 @@ Source: cloudstack
 Section: libs
 Priority: extra
 Maintainer: Wido den Hollander <wido@widodh.nl>
-Build-Depends: debhelper (>= 9), openjdk-11-jdk | java11-sdk | java11-jdk | zulu-11, genisoimage,
- python-mysql.connector | python3-mysql.connector, maven (>= 3) | maven3, python (>= 2.7), python3 (>= 3), nodejs (>= 12), lsb-release, dh-systemd, python-setuptools
+Build-Depends: debhelper (>= 9), openjdk-11-jdk | java11-sdk | java11-jdk | zulu-11, genisoimage, python3 (>= 3), lsb-release, dh-systemd, python-setuptools
 Standards-Version: 3.8.1
 Homepage: http://www.cloudstack.org/
 
--- a/debian/rules
+++ b/debian/rules
@@ -96,7 +96,7 @@ override_dh_auto_install:
 	# cloudstack-ui
 	mkdir $(DESTDIR)/$(SYSCONFDIR)/$(PACKAGE)/ui
 	mkdir -p $(DESTDIR)/usr/share/$(PACKAGE)-ui
-	cd ui && npm install && npm run build && cd ..
+	cd ui && npm config set unsafe-perm true && npm install && npm run build && cd ..
 	cp -r ui/dist/config.json $(DESTDIR)/$(SYSCONFDIR)/$(PACKAGE)/ui/
 	cp -r ui/dist/* $(DESTDIR)/usr/share/$(PACKAGE)-ui/
 	rm -f $(DESTDIR)/usr/share/$(PACKAGE)-ui/config.json


tetra12:
This PR allows to use the latest NodeJS, Maven and Python2 deps, which are NOT present in the 20.04 (focal) apt repo.

That's why I had to remove some dependencies from rules, since the packager won't detect custom installation for Node, Maven and Python.

I assume this will not be an issue for those, who wanna build `*.deb` locally, since build flow will yield an error downstream if any deps are missing.

Docker image simplifies building CloudStack and gives a single immutable image, which can used locally, for testing, CI, CD, etc

To build `deb` files for Ubuntu 20.04, run this:
`docker run -v $PWD/cloudstack:/mnt/build/cloudstack tampler/cs-deb:2004 /usr/local/bin/docker-entrypoint.sh`

RodrigoDLopez:
Hey @tampler
To build my projects, I use that project I mentioned in Slack earlier, created and maintained by @khos2ow (https://github.com/khos2ow/cloudstack-deb-builder).
I made a  [PR](https://github.com/khos2ow/cloudstack-deb-builder/pull/23) in this repository with some changes that allowed to generate packages using Ubuntu 20.04 (Focal) as a base.

I haven't published the generated image, but it is possible to generate a local image and use it to compile your packages easily.

Here is a simple screenshot to show the successful execution of the instance docker. 
image

GabrielBrascher (commented):
Thanks for opening a PR @tetra12. I appreciate the detailed description.

Considering @RodrigoDLopez's tests, it looks like this might be more of an issue with the docker container than with CloudStack itself.

Is there anything that we are missing here? Can you please elaborate a bit more on why you've done these changes?
Thanks!

tetra12:
Hello @GabrielBrascher 

My Dockerfile and the one from @RodrigoDLopez are derived from the same source and clearly both do the job providing a containerized build env for Ubuntu 20.04.

The biggest difference is that my Dockerfile uses custom(the latest!) versions of Node, Maven and Python, which are downloaded as binaries. Downloading the binaries serves three purposes:

1. Easily change the version of the package during dev/CI/test builds
2. Minimize the docker build runtime by having several stages, which are separately and efficiently cached and reused during fails/restarts 
3. Minimize the image size ( no caching, no extra deps, just download-extract-use)
 
The change in CS build flow was required since the current CS packager looks into only Ubuntu 20.04 apt packages and thinks that custom deps are NOT installed and fails on dependency checking. 

Plus, I have published my ready-to-use container [here](https://hub.docker.com/r/tampler/cs-deb). I can contribute with adding other building images, including CentOS 7/8, Ubuntu 18.04 and Ubuntu 21.04

debian/control
```diff
@@ -2,8 +2,7 @@ Source: cloudstack
 Section: libs
 Priority: extra
 Maintainer: Wido den Hollander <wido@widodh.nl>
-Build-Depends: debhelper (>= 9), openjdk-11-jdk | java11-sdk | java11-jdk | zulu-11, genisoimage,
- python-mysql.connector | python3-mysql.connector, maven (>= 3) | maven3, python (>= 2.7), python3 (>= 3), nodejs (>= 12), lsb-release, dh-systemd, python-setuptools
```
rhtyd, line 6: Removing some of the dependencies such as maven3, nodejs is not correct - these are build dependencies.

tetra12: Hello @rhtyd I disagree with your statement. My change doesn't affect the code base - only the build flow, which is currently over constrained. 

Over constrained means that you forced it to use deps versions, shipped with the distribution (Ubuntu 20.04 apt in this case).

Some packages are no longer available in 21.04, as mentioned #5350 . For some deps, like Maven and Node you don't get the latest version easily, which I easily do [here](https://github.com/tetra12/cloudstack-deb-builder/blob/7be9703ee9d031743c32be4c97ac124616593593/ubuntu2004/Dockerfile#L57).

I was able to build ACS and make it up&running with this script for Ubuntu 21.04 and a customized `qemu` and `libvirt`, cloned from git from master. Thus I guarantee that this works fine for Ubuntu 20.04+

If you don't see a value in supporting the latest versions of Ubuntu, Maven, Node, Qemu and Libvirt, feel free to close this PR

rhtyd: Your change breaks packaging

nvazquez:
Needs to be revisited as it may cause regressions as stated in the review comments

tetra12:
Closing due to inactivity

PR #5324, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
According to [Build times out because no output was received](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received), we should carefully use travis_wait, as it may make the build unstable and extend the build time.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -186,11 +186,11 @@ env:
              component/test_vpc_routers
              component/test_vpn_users"
 
-before_install: travis_wait 30 ./tools/travis/before_install.sh
+before_install: ./tools/travis/before_install.sh
 install: ./tools/travis/install.sh
-before_script: travis_wait 30 ./tools/travis/before_script.sh
+before_script: ./tools/travis/before_script.sh
 script:
-  - travis_wait 40 ./tools/travis/script.sh ${TESTS}
+  - ./tools/travis/script.sh ${TESTS}
 after_success: ./tools/travis/after_success.sh
 after_failure: ./tools/travis/after_failure.sh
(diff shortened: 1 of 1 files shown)


DaanHoogland:
interesting @YunLemon , I see the travis run on this PR succeeded in one go.
@rhtyd @nvazquez  I think we should give this a try. Travis is haunting us a lot and certainly the mvn commands are not without output so wouldn't need the travis_wait. The longer tests may be silent, but also not persisting much, so it wouldn't help much.

nvazquez:
Thanks @YunLemon, interesting PR. I've noticed the `travis_wait` function has been used since the integration with Travis. Besides the documentation link in the PR description I have no other reference of this function to analyze its removal @rhtyd any ideas?

rhtyd:
I think it was shear luck, travis_wait is used when/if builds take time and there are no logs so travis won't fail/kill such jobs (producing no outputs). Let me rekick travis to see what happens.

DaanHoogland:
I think https://app.travis-ci.com/github/apache/cloudstack/jobs/532528584 proves @rhtyd 's point @YunLemon . See the last line in the output.

rhtyd:
If you/we search we've had a discussion somewhere about this, I don't think removing the travis_wait benefits; instead removing it may cause failures if/when deploy datacenter or any smoketests takes time to run (i.e. does not output logs) - travis may kill the job.

rhtyd:
Ping @YunLemon - can you review and make more changes or close this?

Issue #5332, opened 2021-08-18
[UI] Unable to add ESX/ESXi host from UI, Username / Password inputs are missing on Add Host dialog.


sureshanaparti:
Unable to add ESX/ESXi host from UI, Username / Password inputs are missing on Add Host dialog.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
main
~~~

##### OS / ENVIRONMENT

CloudStack, with VMware hypervisor

##### SUMMARY

Unable to add ESX/ESXi host from UI.

![ACS_AddHost_VMware](https://user-images.githubusercontent.com/12028987/129894297-25791d6f-9052-45be-bbb8-0859dcaa728b.png)

##### STEPS TO REPRODUCE

- Go to Infrastructure -> Hosts
- Click of 'Add Host' (Select Cluster with 'VMware')
- Unable to specify username / password for the ESX/ESXi

##### EXPECTED RESULTS

Should able to add ESX/ESXi host from UI.

##### ACTUAL RESULTS

Unable to add ESX/ESXi host from UI. Fails with error: "Empty or null value provided for API arg: password"

rhtyd:
Fixed in https://github.com/apache/cloudstack/pull/5337

PR #5136, opened 2021-06-22
apiserver : Ensure required parameters are not empty

davidjumani:
### Description

Addon to https://github.com/apache/cloudstack/pull/5135
Ensures that any required parameter is not empty / null

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

```
(localhost) 🐱 > create domain 
💩 Missing required parameters:  name
(localhost) 🐱 > create domain name= 
🙈 Error: (HTTP 431, error code 4350) Empty or null value provided for API arg: name
```

PR #5337, opened 2021-08-19
Bypass empty string check for username and password

Pearl1594:
### Description

Fixes: https://github.com/apache/cloudstack/issues/5332
This PR attempts to bypass the empty string check added in ParamProcessorValidator - for username and password, so that in case of vmware, it doesn't fail during addition of hosts (where username and password is obtained from cluster details)

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
Added a host to a VMware cluster via UI without any issues.

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/admin/host/AddHostCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/admin/host/AddHostCmd.java
@@ -54,7 +54,7 @@ public class AddHostCmd extends BaseCmd {
     @Parameter(name = ApiConstants.CLUSTER_NAME, type = CommandType.STRING, description = "the cluster name for the host")
     private String clusterName;
 
-    @Parameter(name = ApiConstants.PASSWORD, type = CommandType.STRING, required = true, description = "the password for the host")
+    @Parameter(name = ApiConstants.PASSWORD, type = CommandType.STRING, description = "the password for the host")
     private String password;
 
     @Parameter(name = ApiConstants.POD_ID, type = CommandType.UUID, entityType = PodResponse.class, required = true, description = "the Pod ID for the host")
@@ -63,7 +63,7 @@ public class AddHostCmd extends BaseCmd {
     @Parameter(name = ApiConstants.URL, type = CommandType.STRING, required = true, description = "the host URL")
     private String url;
 
-    @Parameter(name = ApiConstants.USERNAME, type = CommandType.STRING, required = true, description = "the username for the host")
+    @Parameter(name = ApiConstants.USERNAME, type = CommandType.STRING, description = "the username for the host")
     private String username;
 
     @Parameter(name = ApiConstants.ZONE_ID, type = CommandType.UUID, entityType = ZoneResponse.class, required = true, description = "the Zone ID for the host")
--- a/api/src/main/java/org/apache/cloudstack/api/command/admin/host/AddHostCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/admin/host/AddHostCmd.java
@@ -54,7 +54,10 @@ public class AddHostCmd extends BaseCmd {
     @Parameter(name = ApiConstants.CLUSTER_NAME, type = CommandType.STRING, description = "the cluster name for the host")
     private String clusterName;
 
-    @Parameter(name = ApiConstants.PASSWORD, type = CommandType.STRING, description = "the password for the host")
+    @Parameter(name = ApiConstants.USERNAME, type = CommandType.STRING, description = "the username for the host; required to be passed for hypervisors other than VMWare")
+    private String username;
+
+    @Parameter(name = ApiConstants.PASSWORD, type = CommandType.STRING, description = "the password for the host; required to be passed for hypervisors other than VMWare")
     private String password;
 
     @Parameter(name = ApiConstants.POD_ID, type = CommandType.UUID, entityType = PodResponse.class, required = true, description = "the Pod ID for the host")
@@ -62,9 +65,6 @@ public class AddHostCmd extends BaseCmd {
 
     @Parameter(name = ApiConstants.URL, type = CommandType.STRING, required = true, description = "the host URL")
     private String url;
-
-    @Parameter(name = ApiConstants.USERNAME, type = CommandType.STRING, description = "the username for the host")
-    private String username;
 
     @Parameter(name = ApiConstants.ZONE_ID, type = CommandType.UUID, entityType = ZoneResponse.class, required = true, description = "the Zone ID for the host")
     private Long zoneId;
--- a/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
+++ b/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
@@ -24,6 +24,7 @@ import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Date;
 import java.util.HashMap;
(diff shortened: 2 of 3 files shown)


rhtyd (approved):

rhtyd:
I think we had fixed the issue in Trillian to work-around that @Pearl1594 cc @nvazquez

harikrishna-patnala (approved):
LGTM. the old UI behaves the same as this PR.

rhtyd:
@blueorangutan package

Pearl1594:
@blueorangutan test

server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
```diff
@@ -70,6 +71,7 @@
     private static final Logger s_logger = Logger.getLogger(ParamProcessWorker.class.getName());
     public final DateFormat inputFormat = new SimpleDateFormat("yyyy-MM-dd");
     public final DateFormat newInputFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+    public static final List<String> avoidList = new ArrayList<>(Arrays.asList("username", "password"));
```
ravening, line 74: How about using ApiConstants.USERNAME and PASSWORD ?

Pearl1594: Done - Thanks @ravening

Pearl1594:
@blueorangutan package

nvazquez (approved):

weizhouapache:
@Pearl1594 
this has small code change, but might have big impact.
can we skip the username/password check only in some apis ?

Pearl1594:
We could do that @weizhouapache  - However, my understanding is, if a null / empty uname / password is provided as part of any API args - the corresponding APIs logic would ideally validate the correctness of the input or the authorization would fails if it's empty - which is the assumption that we had even before.

weizhouapache:
> We could do that @weizhouapache - However, my understanding is, if a null / empty uname / password is provided as part of any API args - the corresponding APIs logic would ideally validate the correctness of the input or the authorization would fails if it's empty - which is the assumption that we had even before this check was added by - #5136.

@Pearl1594 
If you can make sure that all required username/password have been verified, this pr is good to me.

Otherwise, I suggest you change the fields from required to optional in the specific API (add host ?) and check if they are null in codes.

weizhouapache:
> We could do that @weizhouapache - However, my understanding is, if a null / empty uname / password is provided as part of any API args - the corresponding APIs logic would ideally validate the correctness of the input or the authorization would fails if it's empty - which is the assumption that we had even before this check was added by - #5136.

#5136 is good I think, it fixes some potential missing validations, and causes few regressions (for example the issue you mentioned in PR description). in my opinion, it's better to fix the regression (eg, optional username/password in specific api), instead of rolling back to old behavior (skip validation on username/password in all apis).

Pearl1594:
> > We could do that @weizhouapache - However, my understanding is, if a null / empty uname / password is provided as part of any API args - the corresponding APIs logic would ideally validate the correctness of the input or the authorization would fails if it's empty - which is the assumption that we had even before this check was added by - #5136.
> 
> @Pearl1594
> If you can make sure that all required username/password have been verified, this pr is good to me.
> 
> Otherwise, I suggest you change the fields from required to optional in the specific API (add host ?) and check if they are null in codes.

Will do @weizhouapache  - Thanks.

weizhouapache:
@Pearl1594 
should we check if username/password are null when hypervisor is not vmware ?

api/src/main/java/org/apache/cloudstack/api/command/admin/host/AddHostCmd.java
```diff
@@ -54,7 +54,7 @@
     @Parameter(name = ApiConstants.CLUSTER_NAME, type = CommandType.STRING, description = "the cluster name for the host")
     private String clusterName;
 
-    @Parameter(name = ApiConstants.PASSWORD, type = CommandType.STRING, required = true, description = "the password for the host")
+    @Parameter(name = ApiConstants.PASSWORD, type = CommandType.STRING, description = "the password for the host")
```
weizhouapache, line 57: @Pearl1594 
should we update the description if they are required by kvm/xenserver or other hypervisors ?

weizhouapache (approved):
code lgtm
thanks for fixing @Pearl1594

weizhouapache:
@blueorangutan package

weizhouapache:
@Pearl1594 @rhtyd should we fix it in 4.15.2.0 ?

Pearl1594:
@weizhouapache I believe this issue is only seen on main branch

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -687,6 +688,10 @@ public Discoverer getMatchingDiscover(final Hypervisor.HypervisorType hypervisor
         if ((clusterName != null || clusterId != null) && podId == null) {
             throw new InvalidParameterValueException("Can't specify cluster without specifying the pod");
         }
+        if (!hypervisorType.equalsIgnoreCase(HypervisorType.VMware.toString()) &&
+                (Strings.isNullOrEmpty(username) || Strings.isNullOrEmpty(password))) {
+            throw new InvalidParameterValueException("Username and Password need to be provided.");
+        }
```
DaanHoogland, line 694: maybe excessive but also a tat safer:
```suggestion
        if (!HypervisorType.VMware.toString().equalsIgnoreCase(hypervisorType) &&
                (Strings.isNullOrEmpty(username) || Strings.isNullOrEmpty(password))) {
            throw new InvalidParameterValueException("Username and Password need to be provided.");
        }
```

DaanHoogland (approved):
clgtm

rhtyd:
@blueorangutan package

weizhouapache:
> @weizhouapache I believe this issue is only seen on main branch

@Pearl1594 ok, got it. thanks for confirm

Pearl1594:
@blueorangutan test

DaanHoogland:
must be unlucky on that snapshot test
@blueorangutan test

shwstppr (commented):
@Pearl1594 sorry but I don't understand point of this change. We are removing `required` tag for API params but making them required again in API handling.

Maybe the intention is to check `username`, `password` is not empty when they are passed but I don't think the check correspond to that. Sorry if I'm missing something.

weizhouapache:
> @Pearl1594 sorry but I don't understand point of this change. We are removing `required` tag for API params but making them required again in API handling.
> 
> Maybe the intention is to check `username`, `password` is not empty when they are passed but I don't think the check correspond to that. Sorry if I'm missing something.

username and password can be empty when add a vmware host.

shwstppr (approved):

weizhouapache:
@Pearl1594 is it possible that username and password are empty in cluster details ?

Pearl1594:
> @Pearl1594 is it possible that username and password are empty in cluster details ?

@weizhouapache even if that's the case - there's a check in place to handle it
https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java#L701-L706

sureshanaparti (approved):

weizhouapache:
> > @Pearl1594 is it possible that username and password are empty in cluster details ?
> 
> @weizhouapache even if that's the case - there's a check in place to handle it
> https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java#L701-L706

@Pearl1594 ok, that's good.

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -687,6 +688,10 @@ public Discoverer getMatchingDiscover(final Hypervisor.HypervisorType hypervisor
         if ((clusterName != null || clusterId != null) && podId == null) {
             throw new InvalidParameterValueException("Can't specify cluster without specifying the pod");
         }
+        if (!HypervisorType.VMware.toString().equalsIgnoreCase(hypervisorType) &&
+                (Strings.isNullOrEmpty(username) || Strings.isNullOrEmpty(password))) {
```
GutoVeronezi, line 692: We could import `org.apache.commons.lang3.StringUtils` here and use `StringUtils.isAnyEmpty(username, password)`.

GabrielBrascher: I agree with @GutoVeronezi.

Nowadays we already have some discussions about adopting `org.apache.commons.lang3.StringUtils` vs `com.cloud.utils.StringUtils` "_facade_" (@DaanHoogland :slightly_smiling_face:); with the `google.common.base.Strings` adds a third variable to this "equation".

PR #5216, opened 2021-07-16
Clone a Virtual Machine

atrocitytheme:
### Description/Report

This PR adds a Clone VM feature at the API level (as part of GSoC project #4818), which enables the creation of a fully-cloned virtual machine with ROOT / DATA disks, and the same system configuration as the original VM. (Currently Supported for KVM hypervisor only)

Steps involved: 
1. Creation of temporary snapshots (during the clone VM operation) for both ROOT
2. Creation of template from the ROOT disk snapshot
3. Create a new VM from the template created in step 2
4. Automatic assignment of new network resources for the new Cloned VM
5. Creation of temporary snapshots for data disk
6. Create data disk Volume from the snapshots created in step 5
7. Newly created DATA disk created in step 6 is attached to the new VM
8. Cleanup of temporary resources (snapshots) and error handling of the clone VM process

Note: The template created in step 2 cannot be cleaned up as the newly created clone VM uses this template

Feature included: 
- A new CloneVmCmd API interface to use this feature
- A clone button in the compute instance page supporting the related functionalities

### Documentation
- Documentation PR: https://github.com/apache/cloudstack-documentation/pull/236

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### Screenshots (if appropriate):
Clone Scene

### How Has This Been Tested?

This has been manually tested with a mbx KVM setup on a local machine and mbx kvm setup on GCP

- This feature has been tested on a local Linux system with KVM support (5.3.0-64-generic Ubuntu) and Google Cloud instance (4.9.0-15-amd64 Debian)
- Try with cloudmonkey if the management server is running with default configurations.
```
cloneVirtualMachine virtualmachineid=<target_vm_id>
```
will start the cloning process, it'll create a new cloned VM and start it (with all copied data available), network Ip will be assigned to DB instantly and the actual VM will get this IP after a while
- When secondary system VM agents are not available, the cloning process will fail and clean the previously cloned resources
[...]

PR #5358, opened 2021-08-23
Fix potential NullPointerException in findStoragePool (VolumeOrchestrator)

atrocitytheme:
### Description
The recent changes of the main branch enable the global setting to select preferred storage pool and apply its implementation in the VolumeOrchestrator.java in the findStoragePool method
```
Optional<StoragePool> storagePool = getPreferredStoragePool(poolList, vm);
```
But the usage of it has a potential risk of null pointer exception when the vm parameter in findStoragePool is null, current implementation of getPreferredStoragePool doesn't handle such a condition and directly assumes that VM is never null.
```
String accountStoragePoolUuid = StorageManager.PreferredStoragePool.valueIn(vm.getAccountId());
```
But in one of the implementations in the same class where the findStoragePool is called, it passes vm variable explicitly as null 
*interface*
```
public VolumeInfo createVolumeFromSnapshot(Volume volume, Snapshot snapshot, UserVm vm) throws StorageUnavailableException;
```
When the VM is creating a volume from a snapshot whose VM or pod is not specified, a null pointer exception might happen due to this null passing of the "VM" variable around line 467 in the VolumeOrchestrator.java
```
while ((pool = findStoragePool(dskCh, dc, pod.first(), null, null, null, poolsToAvoid)) != null) {
                    break;
                }
```
This PR adds a null handler and returns the default pool when the vm is null
### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
@@ -311,7 +311,10 @@ public class VolumeOrchestrator extends ManagerBase implements VolumeOrchestrati
     }
 
     private Optional<StoragePool> getPreferredStoragePool(List<StoragePool> poolList, VirtualMachine vm) {
-        String accountStoragePoolUuid = StorageManager.PreferredStoragePool.valueIn(vm.getAccountId());
+        String accountStoragePoolUuid = null;
+        if (vm != null) {
+            accountStoragePoolUuid = StorageManager.PreferredStoragePool.valueIn(vm.getAccountId());
+        }
         Optional<StoragePool> storagePool = getMatchingStoragePool(accountStoragePoolUuid, poolList);
 
         if (storagePool.isPresent()) {
@@ -349,9 +352,6 @@ public class VolumeOrchestrator extends ManagerBase implements VolumeOrchestrati
 
             final List<StoragePool> poolList = allocator.allocateToPool(dskCh, profile, plan, avoidList, StoragePoolAllocator.RETURN_UPTO_ALL);
             if (poolList != null && !poolList.isEmpty()) {
-                if (vm == null) {
-                    return (StoragePool) dataStoreMgr.getDataStore(poolList.get(0).getId(), DataStoreRole.Primary);
-                }
                 // Check if the preferred storage pool can be used. If yes, use it.
                 Optional<StoragePool> storagePool = getPreferredStoragePool(poolList, vm);
 
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java
@@ -349,6 +349,9 @@ public class VolumeOrchestrator extends ManagerBase implements VolumeOrchestrati
 
             final List<StoragePool> poolList = allocator.allocateToPool(dskCh, profile, plan, avoidList, StoragePoolAllocator.RETURN_UPTO_ALL);
             if (poolList != null && !poolList.isEmpty()) {
+                if (vm == null) {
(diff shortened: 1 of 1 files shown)


nvazquez (commented):
I think it makes sense to move the check to the `getPreferredStoragePool() method`:

````
private Optional<StoragePool> getPreferredStoragePool(List<StoragePool> poolList, VirtualMachine vm) {
   String accountStoragePoolUuid = null;
   if (vm != null) {
      accountStoragePoolUuid = StorageManager.PreferredStoragePool.valueIn(vm.getAccountId());
   }
...
````

Then the pool ID specified in the global setting (if set) will be picked instead of the first in the list of pools.

**Comment by atrocitytheme** [comment]
> I think it makes sense to move the check to the `getPreferredStoragePool() method`:
> 
> ```
> private Optional<StoragePool> getPreferredStoragePool(List<StoragePool> poolList, VirtualMachine vm) {
>    String accountStoragePoolUuid = null;
>    if (vm != null) {
>       accountStoragePoolUuid = StorageManager.PreferredStoragePool.valueIn(vm.getAccountId());
>    }
> ...
> ```
> 
> Then the pool ID specified in the global setting (if set) will be picked instead of the first in the list of pools.

Makes sense, just moved the checker to getPreferredStorge method

**Comment by atrocitytheme** [comment]
@blueorangutan package

**Comment by atrocitytheme** [comment]
@blueorangutan test

File: engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/VolumeOrchestrator.java [hunk 1]
```diff
@@ -311,7 +311,10 @@ public VolumeVO allocateDuplicateVolumeVO(Volume oldVol, Long templateId) {
     }
 
     private Optional<StoragePool> getPreferredStoragePool(List<StoragePool> poolList, VirtualMachine vm) {
-        String accountStoragePoolUuid = StorageManager.PreferredStoragePool.valueIn(vm.getAccountId());
+        String accountStoragePoolUuid = null;
+        if (vm != null) {
+            accountStoragePoolUuid = StorageManager.PreferredStoragePool.valueIn(vm.getAccountId());
```
rhtyd, line 316: cc @nvazquez @davidjumani @sureshanaparti @weizhouapache - this needs validation, can you check what caused the regression and whether this fix actually fixes the regression or could potentially cause another one?

atrocitytheme: The issue can be caused when a temporary snapshot (which is used only to copy a new VM or volumes) was created, I had this null pointer exception when implementing Clone a Virtual Machine https://github.com/apache/cloudstack/pull/5216/files, and this exception can be fixed after this implementation.

DaanHoogland:
@atrocitytheme please check the travis failure. might just be related to your fix.

atrocitytheme:
> @atrocitytheme please check the travis failure. might just be related to your fix.

Can anyone run that specific Travis test again? test_affinity_groups_ caught an exception and seems it's not related to this change

shwstppr:
@ravening can you please review this

nvazquez:
The test that failed has run twice, I'll restart that Jenkins job:
````
=== TestName: test_02_deploy_vm_anti_affinity_group_fail_on_not_enough_hosts | Status : SUCCESS ===
=== TestName: test_02_deploy_vm_anti_affinity_group_fail_on_not_enough_hosts | Status : EXCEPTION ===
````

ravening:
LGTM

nvazquez (approved):
Code LGTM - also Travis back to green after restarting the job

rhtyd:
@blueorangutan package

sureshanaparti (approved):
clgtm

rhtyd:
@blueorangutan test

weizhouapache (approved):
code lgtm

Issue #5352, opened 2021-08-20
Strange error message when deleting account


vladimirpetrov:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.16 latest main
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

When an account is deleted (assuming you have enough rights), there is one proper message that it has been deleted and another one 'Unable to find account by id X', which should not be there. Note - the account is successfully deleted.

##### STEPS TO REPRODUCE

~~~
Delete an account.
~~~

##### EXPECTED RESULTS

~~~
There should be only one message about the success of the operation.
~~~

##### ACTUAL RESULTS

~~~
There is an unnecessary error message 'Unable to find account by id X' even though the operation was successfull.
~~~
![image](https://user-images.githubusercontent.com/12384665/130242468-05b6aa62-1039-4373-8de9-5587d7b5e547.png)

rhtyd:
Fixed in https://github.com/apache/cloudstack/pull/5360

PR #5360, opened 2021-08-24
ui: Go back for delete actions before querying async job

davidjumani:
### Description

Fixes https://github.com/apache/cloudstack/issues/5352
Navigates back in history before querying the result of an async job for any delete action
Querying the result leads to a page refresh which can cause an error if it is a delete action and the page has not been changed

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -972,6 +972,9 @@ export default {
             break
           }
         }
+        if ((action.icon === 'delete' || ['archiveEvents', 'archiveAlerts', 'unmanageVirtualMachine'].includes(action.api)) && this.dataView) {
+          this.$router.go(-1)
+        }
         if (['addLdapConfiguration', 'deleteLdapConfiguration'].includes(action.api)) {
           this.$store.dispatch('UpdateConfiguration')
         }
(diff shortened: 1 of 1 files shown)


davidjumani:
@blueorangutan ui

utchoang (approved):
LGTM!

ui/src/views/AutogenView.vue
```diff
@@ -972,6 +972,9 @@ export default {
             break
           }
         }
+        if ((action.icon === 'delete' || ['archiveEvents', 'archiveAlerts', 'unmanageVirtualMachine'].includes(action.api)) && this.dataView) {
```
rhtyd, line 975: Could this cause regression that you go back to previous screen and the action fails for some reason?

ui/src/views/AutogenView.vue
```diff
@@ -972,6 +972,9 @@ export default {
...
+          this.$router.go(-1)
```
shwstppr, line 976: @davidjumani is it possible to check if previous page even exist? Like I can login with a VM details page redirect and then delete that VM. In that case will this take me back to VM list?

ui/src/views/AutogenView.vue
davidjumani: Fixed

ui/src/views/AutogenView.vue
davidjumani: Will redirect to the dashboard. Right now vue doesn't support getting the history list so instead we need to use hooks for every navigation (to ensure we go up instead of to the dashboard) which will slow down the UI

davidjumani:
@blueorangutan ui

shwstppr (approved):
LGTM. Tested deleting and deploying resource, works fine.

utchoang (approved):

Issue #4063, opened 2020-05-08
Improved Error message: Installing on Hosts that don't meet requirements.


rvalle:
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
Host Agent 
~~~

##### CLOUDSTACK VERSION

~~~
4.13.1
~~~

##### CONFIGURATION

Typical testing environment / 3 node minimal cluster.

##### OS / ENVIRONMENT

Centos 7 / KVM

##### SUMMARY

I have been lost for days trying to setup a ACS test cluster. One of my mistakes was to try to add a host that did not meet the RAM requirement. See #4054 for details.

This improvement is about providing better error messages, and facilitate ACS adoption.

##### STEPS TO REPRODUCE

~~~
Install an ACS on a test cluster, 3 nodes 1GB ram each. 
use KVM with nested virtualization enabled.
Follow the quick guide or use Ansible to install the compontents.
~~~

##### EXPECTED RESULTS

A clear ERROR or WARNING should be received stating that the Hosts do not meet the requirements. Something like:

~~~

SSH command output:Starting to configure your system:
Configure SElinux ...         [OK]
Configure Network ...         [OK]
Configure Libvirt ...         [OK]
Configure Firewall ...        [OK]
Configure Nfs ...             [OK]
Configure cloudAgent ...      [OK]
Host meets Requirements... [ERROR]
CloudStack Agent setup finalized with ERROR.

~~~

##### ACTUAL RESULTS

The logs suggests that everything is OK, but system is not functional: 

~~~

SSH command output:Starting to configure your system:
Configure SElinux ...         [OK]
Configure Network ...         [OK]
Configure Libvirt ...         [OK]
Configure Firewall ...        [OK]
Configure Nfs ...             [OK]
Configure cloudAgent ...      [OK]
CloudStack Agent setup is done!
~~~

davidjumani:
Can add a check for the min ram requirements

PR #5369, opened 2021-08-25
kvm: Add check if host meets the minimum requirements

davidjumani:
### Description

Fixes https://github.com/apache/cloudstack/issues/4063

Adds a check whether the kvm host meets the minimum system requirements based on http://docs.cloudstack.apache.org/en/latest/installguide/overview/_requirements.html#host-hypervisor-system-requirements

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

```
Starting to configure your system:
Configure Host ...            [Failed]
Configuration failed for service Host
Try to restore your system:
Restore Host ...              [OK]
```
```
# tail /var/log/cloudstack/agent/setup.log

DEBUG:root:execute:awk '/MemTotal/ { printf "%.3f \n", $2/1024 }' /proc/meminfo
DEBUG:root:CloudStack requires a minimum of 4096 MB and more than 9000 MB memory since 9000 MB is reserved
```

### DIFF
--- a/python/lib/cloudutils/serviceConfig.py
+++ b/python/lib/cloudutils/serviceConfig.py
@@ -505,11 +505,8 @@ class hostConfig(serviceCfgBase):
             cfo = configFileOps("/etc/cloudstack/agent/agent.properties", self)
             reservedMemory = float(cfo.getEntry("host.reserved.mem.mb").strip() or 1024)
             totalMemory = float(bash("awk '/MemTotal/ { printf \"%.3f \\n\", $2/1024 }' /proc/meminfo").getStdout().strip())
-            if totalMemory < reservedMemory or totalMemory < 4096 :
+            if totalMemory < reservedMemory :
                 raise CloudRuntimeException("CloudStack requires a minimum of 4096 MB and more than %d MB memory since %d MB is reserved" %(reservedMemory, reservedMemory))
-
-            if not bash('uname -a | grep x86_64').isSuccess() :
-                raise CloudRuntimeException("Unsupported CPU Architecture. CloudStack requires x86_64 CPU Architecture")
 
             if int(bash('ip a | grep "^\\w" | grep -iv "^lo" | wc -l').getStdout()) < 1 :
                 raise CloudRuntimeException("CloudStack requires at least a single NIC")
--- a/python/lib/cloudutils/serviceConfig.py
+++ b/python/lib/cloudutils/serviceConfig.py
@@ -511,7 +511,7 @@ class hostConfig(serviceCfgBase):
             if not bash('uname -a | grep x86_64').isSuccess() :
                 raise CloudRuntimeException("Unsupported CPU Architecture. CloudStack requires x86_64 CPU Architecture")
 
-            if int(bash('ifconfig | grep "^\\w" | grep -iv "^lo" | wc -l').getStdout()) < 1 :
+            if int(bash('ip a | grep "^\\w" | grep -iv "^lo" | wc -l').getStdout()) < 1 :
                 raise CloudRuntimeException("CloudStack requires at least a single NIC")
 
             return True
--- a/python/lib/cloudutils/serviceConfig.py
+++ b/python/lib/cloudutils/serviceConfig.py
@@ -495,6 +495,30 @@ class nfsConfig(serviceCfgBase):
             logging.debug(formatExceptionInfo())
             return False
 
+class hostConfig(serviceCfgBase):
+    def __init__(self, syscfg):
+        super(hostConfig, self).__init__(syscfg)
+        self.serviceName = "Host"
+
+    def config(self):
+        try:
+            cfo = configFileOps("/etc/cloudstack/agent/agent.properties", self)
+            reservedMemory = float(cfo.getEntry("host.reserved.mem.mb").strip() or 1024)
+            totalMemory = float(bash("awk '/MemTotal/ { printf \"%.3f \\n\", $2/1024 }' /proc/meminfo").getStdout().strip())
+            if totalMemory < reservedMemory or totalMemory < 4 :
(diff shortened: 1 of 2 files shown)


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

rhtyd:
If it's not critical then let's leave if for main only.

python/lib/cloudutils/serviceConfig.py
```diff
@@ -495,6 +495,30 @@ def config(self):
...
+        super(hostConfig, self).__init__(syscfg)
+        self.serviceName = "Host"
+
+    def config(self):
+        try:
+            cfo = configFileOps("/etc/cloudstack/agent/agent.properties", self)
+            reservedMemory = float(cfo.getEntry("host.reserved.mem.mb").strip() or 1024)
+            totalMemory = float(bash("awk '/MemTotal/ { printf \"%.3f \\n\", $2/1024 }' /proc/meminfo").getStdout().strip())
+            if totalMemory < reservedMemory or totalMemory < 4096 :
+                raise CloudRuntimeException("CloudStack requires a minimum of 4096 MB and more than %d MB memory since %d MB is reserved" %(reservedMemory, reservedMemory))
```
rhtyd, line 509: @davidjumani this is not necessary, I can run/use KVM hosts with 2GB RAM and lower the host reserved mem property to say 256/512 mb.

davidjumani: By default 1 GB is reserved
https://github.com/apache/cloudstack/blob/main/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java#L1038

Should I then update the docs since it says a min of 4GB is required?
http://docs.cloudstack.apache.org/en/latest/installguide/overview/_requirements.html#host-hypervisor-system-requirements

nvazquez (approved):

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan test

rhtyd:
@blueorangutan package

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

python/lib/cloudutils/serviceConfig.py
```diff
@@ -495,6 +495,30 @@ def config(self):
...
+
+            if not bash('uname -a | grep x86_64').isSuccess() :
```
rhtyd, line 511: This is not acceptable, we also support aarch64 now

python/lib/cloudutils/serviceConfig.py
rhtyd: The docs are right for recommendation but this code will break usage on people's test env. I would suggest log as a warning. What do others think @nvazquez @sureshanaparti @DaanHoogland...

rhtyd:
Ping @davidjumani

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

NuxRo:
@blueorangutan test

davidjumani:
Test results don't matter as it only affects adding a host

davidjumani:
@blueorangutan package

rhtyd (approved):
LGTM but this needs testing/smoketests across all distros we support KVM agent on

python/lib/cloudutils/serviceConfig.py
```diff
@@ -495,6 +495,27 @@ def config(self):
...
+
+    def config(self):
+        try:
+            cfo = configFileOps("/etc/cloudstack/agent/agent.properties", self)
+            reservedMemory = float(cfo.getEntry("host.reserved.mem.mb").strip() or 1024)
+            totalMemory = float(bash("awk '/MemTotal/ { printf \"%.3f \\n\", $2/1024 }' /proc/meminfo").getStdout().strip())
+            if totalMemory < reservedMemory :
+                raise CloudRuntimeException("CloudStack requires more than %d MB memory since %d MB is reserved" %(reservedMemory, reservedMemory))
+
+            if int(bash('ip a | grep "^\\w" | grep -iv "^lo" | wc -l').getStdout()) < 1 :
```
rhtyd, line 511: It may fail for env where hosts are connected on wifi? And will this command work on all distros?

davidjumani: It will work since it checks for any nic. No explicit regex such as starts with 'e', etc. IP is present by default on all supported distros

davidjumani:
@blueorangutan test

python/lib/cloudutils/serviceConfig.py
rhtyd: So just to confirm it counts any nics and doesn't ignore any nics, we shouldn't assume nics name will start with "e" though.
You can instead change the logic to count physical nics, for ex use something from:
lshw -class network
Or,
find /sys/class/net -type l -not -lname '*virtual*' -printf '%f\n'

davidjumani: Yes, it counts all nics excluding loopback
In some minimal instances lshw is not present, so using IP -a since that is pretty much always there

rhtyd:
@davidjumani can you confirm if this was tested with the final changes on all supported Linux/KVM distros?

davidjumani:
Tested by adding the following hosts :
- Centos 7
- Rocky 8
- Suse 15
- Ubuntu 18
- Ubuntu 20

```
Starting to configure your system:
Configure Host ...            [OK]    <<<<
Configure SElinux ...         [OK]
Configure Network ...         [OK]
Configure Libvirt ...         [OK]
Configure Firewall ...        [OK]
Configure Nfs ...             [OK]
Configure cloudAgent ...      [OK]
CloudStack Agent setup is done!

```

rhtyd (approved):

PR #5385, opened 2021-08-27
engine/schema: Use same upgrade path as 4.15.1-4.16.0 as for 4.15.2

rhtyd:
Introduce same Db upgrade path from both 4.15.1/4.15.2 to 4.16.0


rhtyd:
Did a manual test, the upgrade seems to kick in. I'm merging this on discretion as upgrade tests are done as part of release/pre-RC efforts.

Issue #5391, opened 2021-08-31
Scale VM custom compute offering not working


kricud:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.15.1
~~~

##### STEPS TO REPRODUCE

~~~
Press on Scale VM select Custom Compute offering try to move sliders for CPU and RAM not working. 
Enter values by hand and press ok presents error.
~~~
![Screenshot 2021-08-31 094315](https://user-images.githubusercontent.com/30048481/131455496-f78aad6b-bcbb-44c5-8c36-670faad35339.png)
![Screenshot 2021-08-31 094333](https://user-images.githubusercontent.com/30048481/131455502-ad4df404-9fbc-48a6-8126-660159a91c67.png)

##### EXPECTED RESULTS

~~~
Scale VM select Custom Compute offering works
~~~

##### ACTUAL RESULTS

~~~
Scale VM select Custom Compute offering not working
~~~

nvazquez:
Thanks @utchoang, closing as PR is merged

PR #5392, opened 2021-08-31
UI - Scale VM - Fix compute offering selection not working

utchoang:
### Description

Fixes https://github.com/apache/cloudstack/issues/5391

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/ui/src/views/compute/wizard/ComputeSelection.vue
+++ b/ui/src/views/compute/wizard/ComputeSelection.vue
@@ -175,10 +175,8 @@ export default {
   },
   computed: {
     colContraned () {
-      if (this.isConstrained) {
-        if (this.maxCpu && isNaN(this.maxCpu)) {
-          return 12
-        }
+      if (this.isConstrained && this.maxCpu && !isNaN(this.maxCpu)) {
+        return 12
       }
 
       return 8
--- a/ui/src/views/compute/wizard/ComputeSelection.vue
+++ b/ui/src/views/compute/wizard/ComputeSelection.vue
@@ -192,7 +192,6 @@ export default {
     }
   },
   mounted () {
-    console.log(this.isCustomizedIOps)
     if (this.isCustomized) {
       this.fillValue()
     }
--- a/ui/src/views/compute/wizard/ComputeSelection.vue
+++ b/ui/src/views/compute/wizard/ComputeSelection.vue
@@ -25,7 +25,7 @@
             :validate-status="errors.cpu.status"
             :help="errors.cpu.message">
             <a-row :gutter="12">
-              <a-col :md="10" :lg="10" v-show="isConstrained">
+              <a-col :md="10" :lg="10" v-show="isConstrained && maxCpu && isNaN(maxCpu)">
                 <a-slider
                   :min="minCpu"
                   :max="maxCpu"
@@ -61,7 +61,7 @@
             :validate-status="errors.memory.status"
             :help="errors.memory.message">
             <a-row :gutter="12">
-              <a-col :md="10" :lg="10" v-show="isConstrained">
+              <a-col :md="10" :lg="10" v-show="isConstrained && maxMemory && !isNaN(maxMemory)">
                 <a-slider
                   :min="minMemory"
                   :max="maxMemory"
@@ -175,7 +175,13 @@ export default {
   },
   computed: {
     colContraned () {
-      return this.isConstrained ? 12 : 8
+      if (this.isConstrained) {
+        if (this.maxCpu && isNaN(this.maxCpu)) {
+          return 12
+        }
+      }
+
+      return 8
     }
   },
   watch: {
@@ -186,6 +192,7 @@ export default {
     }
   },
   mounted () {
+    console.log(this.isCustomizedIOps)
     if (this.isCustomized) {
       this.fillValue()
     }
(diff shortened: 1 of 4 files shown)


utchoang:
@blueorangutan ui

rhtyd:
@blueorangutan ui

davidjumani:
@utchoang I think another issue is the cpu speed slider is not being shown. Can you look into that?

kricud:
@utchoang There is such field in legacy UI. It’s needed as its mandatory in offering. It would be grate that this field is prefilled, and admin can control default/prefilled value. Now for every new user trying to use this ticket is created with question what I must enter in CPU (in MHz) field.

Screenshot 2021-08-31 152904

utchoang:
@davidjumani In my opinion, it would make more sense to not display the slider here anymore. What do you think?

davidjumani:
@utchoang Yes, I think the slider is pointless unless it is a constrained offering (show it only for values within a range of cpu / memory and let the speed be a regular field) cc @rhtyd

utchoang:
@blueorangutan ui

utchoang:
@blueorangutan ui

ui/src/views/compute/wizard/ComputeSelection.vue
```diff
@@ -25,7 +25,7 @@
             :validate-status="errors.cpu.status"
             :help="errors.cpu.message">
             <a-row :gutter="12">
-              <a-col :md="10" :lg="10" v-show="isConstrained">
+              <a-col :md="10" :lg="10" v-show="isConstrained && maxCpu && isNaN(maxCpu)">
```
shwstppr, line 28: @utchoang is this show condition correct here. For memory you have used reversed for isNaN which probably makes sense

utchoang: @shwstppr Yes. I'm wrong. Thanks

utchoang:
@blueorangutan ui

utchoang:
@blueorangutan ui

kricud:
@utchoang 
In Legacy UI if you want to use custom offering you need to enter/fill 3 values 
CPU (in MHz), # of CPU Cores, Memory (in MB) they are mandatory.

In New UI you can enter only 2 values CPU Cores, Memory (in MB).
If you want to create service offering manually or dynamically you need to specify 
CPU (in MHz) in both cases its mandatory field.

New UI is missing CPU (in MHz).

My two cents:
New UI sliders in their existing form are pointless:
For RAM slider you would expect increments in 1024MB.
For CPU Cores you would expect averaged maximum of host vCPU.
For CPU MHz you would expect all host MHz options.
I would suggest remove sliders if only things they do is increase number from 1 to something.

New UI:
Screenshot 2021-09-01 122614

Legacy UI:
Screenshot 2021-09-01 122712

Static offering:
Screenshot 2021-09-01 123235

shwstppr:
@kricud Is **Custom compute offering - A** a constrained custom offering or an unconstrained offering
In my test for this PR, UI seems to be showing correct fields for both constrained and unconstrained compute offerings.
No sliders for unconstrained custom offering
Screenshot from 2021-09-01 15-25-36
Sliders for constrained custom offering
Screenshot from 2021-09-01 15-25-53

Regarding max value for sliders, they are based on the max value set in the compute offering at the time of the creation. I don't think we need to involve host CPU/memory here.
Although the step for slider can be tweaked based on the min-max if it is not done already.

kricud:
@shwstppr @utchoang My fail - operating with constrained offerings and raging on keyboard why its not doing what I’m expecting. Apologies for BS typing.

kricud:
@shwstppr @utchoang Now it all came together- in our case its constrained offering that has no constrains.
It was created on 4.11.3 after migration to 4.15.1 its all bugging as it has missing mandatory values.

Name	Custom compute offering - A
ID	5adb6fde-60e6-4bd1-b8ab-d48dae8f5fa0
Description	Custom compute offering - A
Storage Type	shared
Provisioning Type	thin
Write-cache Type	
# of CPU Cores	
CPU (in MHz)	
Memory (in MB)	
Network Rate (Mb/s)	1000
Custom IOPS	
Min IOPS	N/A
Max IOPS	N/A
Hypervisor Snapshot Reserve	N/A
Disk Read Rate (BPS)	
Disk Write Rate (BPS)	
Disk Read Rate (IOPS)	
Disk Write Rate (IOPS)	
Offer HA	Yes
CPU Cap	No
Volatile	No
Deployment planner	
Planner mode	
GPU	
vGPU type	
Storage Tags	
Host Tag	
Domain	ROOT
Zone	
Created	08 Dec 2020 13:27:46

Apologies once more.

shwstppr:
@kricud No problem.
I think the concept of constrained-unconstrained custom compute offerings got introduced in 4.14. Maybe there could be an issue with how custom compute offerings created prior to that were handled. Will try to check with a 4.11 to 4.14/4.15 upgrade.

shwstppr (approved):
LGTM. Checked different users of `ComputeSelection.vue` and ScaleVM form

nvazquez (approved):
LGTM

Issue #5440, opened 2021-09-13
Dashboard/project UI issues as user account role


rhtyd:
All issues reproduced under user-account role (not admin!):

1. Deploy 2 Vms, volumes in a user account
2. Create a project under a user account -> the project tab doesn't refresh; next try to add an account (say admin) -> the account tab doesn't refreshes. Finally deploy 1 VM, volumes in that account.
3. Go to project dashboard, you'll see 1 VM; now switch to default view -> you'll see dashboard does not refresh/change

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.15, 4.16/main
~~~

rhtyd:
cc @Pearl1594 @shwstppr @davidjumani @nvazquez @sureshanaparti @weizhouapache

nvazquez:
@davidjumani @utchoang I could reproduce the issues, simply clicking Compute -> Instances as a user and then Projects makes the projects list to not being refreshed, however I can see the notification at the top. Any idea on this issues?

Pearl1594:
@rhtyd @nvazquez regarding step #2 i.e., On creation of a project - the project tab doesn't auto-refresh, I was able to reproduce the issue, but on clearing the cache - surprisingly things worked as it should i.e., Project tab auto refreshes on creation of a new project

PR #5447, opened 2021-09-14
ui: Refresh Usage dashboard when swapping between Project and Default view

Pearl1594:
### Description

This PR fixes https://github.com/apache/cloudstack/issues/5440 - where we notice that when logged in with a User role, and on swapping between project and default view, the resource count details on the Usage dashboard isn't properly updated

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
Followed the steps mentioned in https://github.com/apache/cloudstack/issues/5440 - and observed the usage dashboard properly updated.

### DIFF
--- a/ui/src/views/dashboard/UsageDashboard.vue
+++ b/ui/src/views/dashboard/UsageDashboard.vue
@@ -137,7 +137,7 @@ export default {
     this.$store.watch(
       (state, getters) => getters.project,
       (newValue, oldValue) => {
-        if (newValue && newValue.id) {
+        if (newValue && newValue !== oldValue) {
           this.fetchData()
         }
       }


Pearl1594:
@blueorangutan ui

rhtyd (approved):
LGTM but instead of comparing objects should we compare the ID of new and old values?

Pearl1594:
> LGTM but instead of comparing objects should we compare the ID of new and old values?

@rhtyd There isn't an id (project ID) for default view - which was the reason for not refreshing in the earlier case.

rhtyd:
Ah okay that makes sense, maybe we can add a dummy ID of 0 for default view? But this change also LGTM if it works.

ui/src/views/dashboard/UsageDashboard.vue
```diff
@@ -137,7 +137,7 @@ export default {
     this.$store.watch(
       (state, getters) => getters.project,
       (newValue, oldValue) => {
-        if (newValue && newValue.id) {
+        if (newValue && newValue !== oldValue) {
```
shwstppr, line 140: @Pearl1594 why not compare ids itself? This condition might be as good as `if (newValue) {`

Pearl1594: there is no id for default view @shwstppr

nvazquez (approved):
LGTM thanks @Pearl1594

utchoang (approved):

PR #5501, opened 2021-09-23
server: check service offering (storage) tags when reallocate a ROOT disk

### LINKED ISSUE #4406: Retry logic - ignores tags
NOT burning issue, as it's hardly reproducible - but a proper code analysis might prove bug (non) existance.

I've seen this in master/PR #4307 (VMware , but probably a general issue) - an issue with storage tags not being honored during VM deployment retry.

By using tags, I've made sure a VM will be deployed on a specific storage pool.

Due to VM failing to start on host X on storage X _(whatever the root cause - i.e vApp properties missing in the VM deployment API call and no default value set - so VM fails to start on VMware side)_ - **ACS will retry the action for a few times, and eventually will try to deploy a VM on different datastore than the one choosen by usage of storage tags** - I've seen a constant VM "reallocate" task to other hosts until many hosts were tried and failed to start a VM - then we cleaned-up that VM, and deployed VM from scratch on a different datastore.

i.e. storage tags are NOT doing the work during retry logic - as it seems (or does it...)

Perhaps also @alexandremattioli saw similar issue with host tags in 4.14 or so (please update here if that's the case...)

The way to reproduce this is a bit problematic - I can only think about the edge case that I saw (which btw, will be stated as a limitation in 4.15 Release notes):

- can be reproduced by clean ACS 4.14 install, registering a template that has some vApp properties (I can provide some) with no default value set for at least 1 vApp property. Then upgrade to master (PR 4307) and try to deploy VM from that template (some vApp property has NO default value) - on **VVols** storage (new in ACS 415/PR #4307), while still having an old fashioned NFS pool as well (not targeted by the storage tag).
- I'm not sure what other way would be a good way to cause some failure on VMware side (when starting a VM)

Worth testing also with 2 storage pools, but where the targetd one (by storage tags) is over the allocated threashold and can't be used - see then if other pools will be used (despite NOT having the required storage tag)

weizhouapache:
### Description

This PR fixes #4406  (probably)

In some cases, the disk offering of ROOT disk is not same as service offering of VM.
for example
(1) create a vm but not start it, then scale vm to another service offering
(2) reinstall vm, then scale vm to another service offering

These two issues are fixed by this PR.
The issue might exist in some other scenarios, so it would be better to always update ROOT disk offering tags to the tags of vm service offering in DeploymentPlanningManagerImpl.java.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
+++ b/engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
@@ -61,6 +61,7 @@ public class VolumeDaoImpl extends GenericDaoBase<VolumeVO, Long> implements Vol
     protected final GenericSearchBuilder<VolumeVO, Long> ActiveTemplateSearch;
     protected final SearchBuilder<VolumeVO> InstanceStatesSearch;
     protected final SearchBuilder<VolumeVO> AllFieldsSearch;
+    protected final SearchBuilder<VolumeVO> RootDiskStateSearch;
     protected GenericSearchBuilder<VolumeVO, Long> CountByAccount;
     protected GenericSearchBuilder<VolumeVO, SumCount> primaryStorageSearch;
     protected GenericSearchBuilder<VolumeVO, SumCount> primaryStorageSearch2;
@@ -218,6 +219,15 @@ public class VolumeDaoImpl extends GenericDaoBase<VolumeVO, Long> implements Vol
     }
 
     @Override
+    public List<VolumeVO> findReadyAndAllocatedRootVolumesByInstance(long instanceId) {
+        SearchCriteria<VolumeVO> sc = RootDiskStateSearch.create();
+        sc.setParameters("instanceId", instanceId);
+        sc.setParameters("state", Volume.State.Ready, State.Allocated);
+        sc.setParameters("vType", Volume.Type.ROOT);
+        return listBy(sc);
+    }
+
+    @Override
     public List<VolumeVO> findByPod(long podId) {
         SearchCriteria<VolumeVO> sc = AllFieldsSearch.create();
         sc.setParameters("pod", podId);
@@ -364,6 +374,12 @@ public class VolumeDaoImpl extends GenericDaoBase<VolumeVO, Long> implements Vol
         AllFieldsSearch.and("updatedCount", AllFieldsSearch.entity().getUpdatedCount(), Op.EQ);
         AllFieldsSearch.and("name", AllFieldsSearch.entity().getName(), Op.EQ);
         AllFieldsSearch.done();
+
+        RootDiskStateSearch = createSearchBuilder();
+        RootDiskStateSearch.and("state", RootDiskStateSearch.entity().getState(), Op.IN);
+        RootDiskStateSearch.and("vType", RootDiskStateSearch.entity().getVolumeType(), Op.EQ);
+        RootDiskStateSearch.and("instanceId", RootDiskStateSearch.entity().getInstanceId(), Op.EQ);
+        RootDiskStateSearch.done();
 
         DetachedAccountIdSearch = createSearchBuilder();
         DetachedAccountIdSearch.and("accountId", DetachedAccountIdSearch.entity().getAccountId(), Op.EQ);
--- a/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
+++ b/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
@@ -1558,7 +1558,8 @@ StateListener<State, VirtualMachine.Event, VirtualMachine> {
 
             DiskOfferingVO diskOffering = _diskOfferingDao.findById(toBeCreated.getDiskOfferingId());
 
-            if (vmProfile.getTemplate().getFormat() == Storage.ImageFormat.ISO && vmProfile.getServiceOffering().getTagsArray().length != 0) {
+            if ((vmProfile.getTemplate().getFormat() == Storage.ImageFormat.ISO || toBeCreated.getVolumeType() == Volume.Type.ROOT)
+                    && vmProfile.getServiceOffering().getTagsArray().length != 0) {
                 diskOffering.setTagsArray(Arrays.asList(vmProfile.getServiceOffering().getTagsArray()));
             }
 
--- a/engine/schema/src/main/java/com/cloud/storage/dao/VolumeDao.java
+++ b/engine/schema/src/main/java/com/cloud/storage/dao/VolumeDao.java
@@ -94,6 +94,8 @@ public interface VolumeDao extends GenericDao<VolumeVO, Long>, StateDao<Volume.S
 
     List<VolumeVO> findReadyRootVolumesByInstance(long instanceId);
 
+    List<VolumeVO> findReadyAndAllocatedRootVolumesByInstance(long instanceId);
+
     List<Long> listPoolIdsByVolumeCount(long dcId, Long podId, Long clusterId, long accountId);
 
     List<Long> listZoneWidePoolIdsByVolumeCount(long dcId, long accountId);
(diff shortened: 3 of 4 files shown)


davidjumani:
@blueorangutan package

server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
```diff
@@ -1558,7 +1558,8 @@ protected boolean hostCanAccessSPool(Host host, StoragePool pool) {
 
             DiskOfferingVO diskOffering = _diskOfferingDao.findById(toBeCreated.getDiskOfferingId());
 
-            if (vmProfile.getTemplate().getFormat() == Storage.ImageFormat.ISO && vmProfile.getServiceOffering().getTagsArray().length != 0) {
+            if ((vmProfile.getTemplate().getFormat() == Storage.ImageFormat.ISO || toBeCreated.getVolumeType() == Volume.Type.ROOT)
+                    && vmProfile.getServiceOffering().getTagsArray().length != 0) {
```
sureshanaparti, line 1562: For ROOT volume from template, the disk / service offering storage tags are the same. Is it required to explicitly set the disk offering tags?

weizhouapache: @sureshanaparti 
yes, you are right. it is not required. this is why I converted this PR to draft today.

I thought I have reproduced the issue two days before. but today I am not able to reproduce.
I am still investigating the issue #4406 
When I found some clue I will update this PR

weizhouapache: @sureshanaparti @davidjumani 
found two scenarios that root disk offering is not same as vm service offering.
there might be more scenarios (for example, attach data disk as root disk ?).

please review

engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
```diff
@@ -217,6 +218,15 @@ public VolumeVO findByPoolIdName(long poolId, String name) {
         return listBy(sc);
     }
 
+    @Override
+    public List<VolumeVO> findReadyAndAllocatedRootVolumesByInstance(long instanceId) {
```
DaanHoogland, line 222: nitpick: `findReadyAndAllocatedRootVolumeByInstance` (singular) we are only looking for one volume, can we encounter more?

DaanHoogland (approved):
clgtm

DaanHoogland:
@blueorangutan package

engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
shwstppr: @DaanHoogland not sure but don't we have multi-disk OVA in VMware? Is only a single volume considered ROOT in that case?

weizhouapache: @DaanHoogland for kvm, there is only 1 root disk. 
for other hypervisors eg vmware, I have no idea. 
so I'd like to use similar code as findReadyRootVolumesByInstance which returns a list of root volumes and processed in other java files.

weizhouapache:
@blueorangutan test

engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
nvazquez: Yes, it is possible to have multiple ROOT disks for deploy-as-is templates on Vmware. However, I'm not seeing a problem as the method is returning a list of volumes

server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
sureshanaparti: @weizhouapache first scenario - when VM is deployed from ISO, what is the other scenario ?

Does attach data disk as root disk supported?

engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
```diff
@@ -365,6 +375,12 @@ public VolumeDaoImpl() {
         AllFieldsSearch.and("name", AllFieldsSearch.entity().getName(), Op.EQ);
         AllFieldsSearch.done();
 
+        RootDiskStateSearch = createSearchBuilder();
```
sureshanaparti, line 378: May be you can re-use _AllFieldsSearch_, by changing Op.EQ to Op.IN here: https://github.com/apache/cloudstack/blob/67bc74de56357eeefdec9cccb6314d73dc757f5d/engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java#L362

weizhouapache: @sureshanaparti I was thinking about it. but at the end I decided to add a new searchbuilder.
no idea which way is suggested.

server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
weizhouapache: @sureshanaparti 
as I said in the description

```
In some cases, the disk offering of ROOT disk is not same as service offering of VM.
for example
(1) create a vm but not start it, then scale vm to another service offering
(2) reinstall vm, then scale vm to another service offering

These two issues are fixed by this PR.
The issue might exist in some other scenarios, so it would be better to always update ROOT disk offering tags to the tags of vm service offering in DeploymentPlanningManagerImpl.java.
```

engine/schema/src/main/java/com/cloud/storage/dao/VolumeDaoImpl.java
sureshanaparti: @weizhouapache suggested to re-use _AllFieldsSearch_ , but it is ok with new searchbuilder as well.

server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
sureshanaparti: @weizhouapache in both these scenarios, does the ROOT disk offering in volumes table updated with the new service offering?

weizhouapache: @sureshanaparti 
only the root volumes in Ready state are updated (root volumes are Allocated in these two scenarios).
the issues are fixed in this PR.

sureshanaparti (approved):
clgtm

borisstoyanov (approved):
```
(localcloud) 🐱 > scale virtualmachine id=23fa735f-a26b-4490-ad2a-af6f9b81bd7b serviceofferingid=3db58368-afd9-477f-8d2e-729232b338eb
{
  "accountid": "5becde6e-21bc-11ec-890a-1e001f000425",
  "cmd": "org.apache.cloudstack.api.command.admin.vm.ScaleVMCmdByAdmin",
  "completed": "2021-10-01T09:10:32+0000",
  "created": "2021-10-01T09:10:32+0000",
  "jobid": "20185515-fd03-4771-b78b-7cd76a4c6784",
  "jobprocstatus": 0,
  "jobresult": {
    "errorcode": 431,
    "errortext": "Unable to upgrade virtual machine; the current service offering  should have tags as subset of the new service offering tags. Current service offering tags: [tag1]; new service offering tags: [tag2]"
  },
  "jobresultcode": 530,
  "jobresulttype": "object",
  "jobstatus": 2,
  "userid": "5bee7260-21bc-11ec-890a-1e001f000425"
```
LGTM, it does consider storage tags now.

sureshanaparti:
@weizhouapache can you re-target to 'main' please.

sureshanaparti:
@blueorangutan package

weizhouapache:
@blueorangutan package

sureshanaparti:
@blueorangutan test

sureshanaparti:
@blueorangutan test

rhtyd:
@blueorangutan test

nvazquez (approved):

PR #5504, opened 2021-09-23
Fix permission issue during Diagnostic service garbage collection

Pearl1594:
### Description

This PR attempts to fix the permission issue observed during unmounting of secondary store(s) during the diagnostic service run (on VMware) and the Diagnostic service garbage collector:
```
2021-09-23 05:17:58,206 DEBUG [o.a.c.s.NfsMountManager] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Executing while with timeout : 10000
2021-09-23 05:17:58,263 DEBUG [o.a.c.s.NfsMountManager] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Execution is successful.
2021-09-23 05:17:58,267 DEBUG [o.a.c.s.NfsMountManager] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Executing: sudo chmod 1777 /var/cloudstack/mnt/32989157261914.7b4126ed
2021-09-23 05:17:58,275 DEBUG [o.a.c.s.NfsMountManager] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Executing while with timeout : 10000
2021-09-23 05:17:58,292 DEBUG [o.a.c.s.NfsMountManager] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Execution is successful.
2021-09-23 05:17:58,300 DEBUG [o.a.c.d.DiagnosticsHelper] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Executing: /bin/bash -c umount /var/cloudstack/mnt/32989157261914.7b4126ed
2021-09-23 05:17:58,304 DEBUG [o.a.c.d.DiagnosticsHelper] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Executing while with timeout : 3600000
2021-09-23 05:17:58,305 DEBUG [o.a.c.d.DiagnosticsHelper] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) Exit value is 1
2021-09-23 05:17:58,306 DEBUG [o.a.c.d.DiagnosticsHelper] (BackgroundTaskPollManager-1:ctx-d5fcff87) (logid:779879f6) umount: /var/cloudstack/mnt/32989157261914.7b4126ed: umount failed: Operation not permitted

```
This can be solved in one of 2 ways, 
1. Do not unmount the secondary store across successive runs of the diagnostic service clean up job or collection of diagnostic data(on vmware)
2. Fix the issue of unmount (by running the umount command using sudo) and clean up the mount path so that during successive runs a new mount path can be created - https://github.com/shapeblue/cloudstack/commit/a9b460869b8f244f4b3db3cc3973f3419dffed83

Current solution follows Option 1 as it seems to be a simpler fix, and the storage unmount happens during Cloudstack Management service restarts

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
Tested on KVM and VMware env - reduced the GC interval for diagnostic service - 'diagnostics.data.gc.interval' and observed  no more failures and on restart of MS, the store is unmounted and path is deleted

### DIFF
--- a/server/src/main/java/org/apache/cloudstack/diagnostics/DiagnosticsServiceImpl.java
+++ b/server/src/main/java/org/apache/cloudstack/diagnostics/DiagnosticsServiceImpl.java
@@ -18,7 +18,6 @@
 package org.apache.cloudstack.diagnostics;
 
 import static org.apache.cloudstack.diagnostics.DiagnosticsHelper.getTimeDifference;
-import static org.apache.cloudstack.diagnostics.DiagnosticsHelper.umountSecondaryStorage;
 import static org.apache.cloudstack.diagnostics.fileprocessor.DiagnosticsFilesList.RouterDefaultSupportedFiles;
 import static org.apache.cloudstack.diagnostics.fileprocessor.DiagnosticsFilesList.SystemVMDefaultSupportedFiles;
 
@@ -352,8 +351,6 @@ public class DiagnosticsServiceImpl extends ManagerBase implements PluggableServ
             String msg = String.format("Exception caught during scp from %s to secondary store %s: ", vmSshIp, dataDirectoryInSecondaryStore);
             LOGGER.error(msg, e);
             return new Pair<>(false, msg);
-        } finally {
-            umountSecondaryStorage(mountPoint);
         }
 
         return new Pair<>(success, "File copied to secondary storage successfully");
@@ -481,17 +478,11 @@ public class DiagnosticsServiceImpl extends ManagerBase implements PluggableServ
 
         private void cleanupOldDiagnosticFiles(DataStore store) {
             String mountPoint = null;
-            try {
-                mountPoint = serviceImpl.mountManager.getMountPoint(store.getUri(), null);
-                if (StringUtils.isNotBlank(mountPoint)) {
-                    File directory = new File(mountPoint + File.separator + DIAGNOSTICS_DIRECTORY);
-                    if (directory.isDirectory()) {
-                        deleteOldDiagnosticsFiles(directory, store.getName());
-                    }
-                }
-            } finally {
-                if (StringUtils.isNotBlank(mountPoint)) {
-                    umountSecondaryStorage(mountPoint);
+            mountPoint = serviceImpl.mountManager.getMountPoint(store.getUri(), null);
+            if (StringUtils.isNotBlank(mountPoint)) {
+                File directory = new File(mountPoint + File.separator + DIAGNOSTICS_DIRECTORY);
+                if (directory.isDirectory()) {
+                    deleteOldDiagnosticsFiles(directory, store.getName());
                 }
             }
         }


weizhouapache (approved):
code lgtm

it is ok to keep secondary storage mounted on mgt server.

Pearl1594:
@blueorangutan package

rhtyd (approved):

rhtyd:
@blueorangutan test

server/src/main/java/org/apache/cloudstack/diagnostics/DiagnosticsServiceImpl.java
```diff
@@ -481,17 +478,11 @@ public Long getDelay() {
...
-                    }
-                }
-            } finally {
-                if (StringUtils.isNotBlank(mountPoint)) {
-                    umountSecondaryStorage(mountPoint);
+            mountPoint = serviceImpl.mountManager.getMountPoint(store.getUri(), null);
+            if (StringUtils.isNotBlank(mountPoint)) {
+                File directory = new File(mountPoint + File.separator + DIAGNOSTICS_DIRECTORY);
+                if (directory.isDirectory()) {
+                    deleteOldDiagnosticsFiles(directory, store.getName());
```
sureshanaparti, line 485: Does the secondary storage stay mounted on hypervisors other than VMware, after cleanup of diagnostics files? If so, it is better to fix the permission issue (on VMware), and unmount the secondary storage from mgmt server when not needed.

nvazquez (approved):
LGTM

server/src/main/java/org/apache/cloudstack/diagnostics/DiagnosticsServiceImpl.java
sureshanaparti: @Pearl1594 any update on this ^^^ , possible to fix this permission issue for VMware only?

Pearl1594: Sorry @sureshanaparti I seemed to have missed your earlier comment. Currently what's happening is that due to the permission issue, we see the secondary stores remain mounted on all the hypervisors. We could fix the permission issue using this solution - https://github.com/shapeblue/cloudstack/commit/a9b460869b8f244f4b3db3cc3973f3419dffed83. However, I am of the opinion that we probably don't need to mount and un-mount every time the garbage collection thread runs. As fixing the permission issue wouldn't solve the issue entirely, we'd also need to cleanup the mount path (at /var/cloudstack/mnt) and remove that mount path from the static list of storageMounts maintained, so that in successive runs are successful.

sureshanaparti: OK @Pearl1594 , Got it. I think, keeping secondary store mounted always on all the hypervisors is also not good. Only SSVM should have the access to it always.

DaanHoogland: I would agree if this was on hosts @sureshanaparti , but it is on the MS and
1. I would hate using sudo once more in the system
2. I would prefer to keep as much uniformity over hypervisor types as possible, so rather not have a vmware specific solution.
what do you think?

DaanHoogland (approved):

rhtyd:
@blueorangutan package

server/src/main/java/org/apache/cloudstack/diagnostics/DiagnosticsServiceImpl.java
sureshanaparti: > I would agree if this was on hosts @sureshanaparti , but it is on the MS and
> 
> 1. I would hate using sudo once more in the system
> 2. I would prefer to keep as much uniformity over hypervisor types as possible, so rather not have a vmware specific solution.
>    what do you think?

agree @DaanHoogland

rhtyd:
@blueorangutan test

nvazquez:
@blueorangutan test

nvazquez (approved):

PR #5517, opened 2021-09-28
Fix #3448 quota calculation for monthly tariffs

### LINKED ISSUE #3448: Quota calculation is incorrect
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Quota
~~~

##### CLOUDSTACK VERSION

~~~
master branch
~~~

##### SUMMARY

https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java#L101
The default value of `usage.stats.job.aggregation.range` is 1440.
_aggregationDuration = Integer.parseInt(aggregationRange);
https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java#L142
The value of `aggregationRatio` calculated is `1440/(30 * 24 * 60)`.

All updateQuota* functions have the bug below.
https://github.com/apache/cloudstack/blob/1d05fead49f5c856257a741b07122f5633d2e359/framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java#L436
This line would be one-day-cost not one-hour-cost.
So the `aggregationRatio` should be fixed with `1/(30 * 24)`.

jairov4:
### Description

The quota calculation has been wrong for monthly tariffs (like Compute-Month, GB-Month, etc) for so much time but even when #3448 was reported it was closed without evaluation.

The wrong calculation was the price per hour, it was taking the monthly tariff and multiplying by the aggregation ratio which has no sense at all.

The right price per hour is the monthly tariff divided by the standard number of hours in a month.

Fixes #3448.

### Types of changes

- [X] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
Partial execution of the component with a test vector.
Tracing the values and compared with spreadsheet.

### DIFF
--- a/framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
+++ b/framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
@@ -139,16 +139,15 @@ public class QuotaManagerImpl extends ManagerBase implements QuotaManager {
         }
         s_logger.info("Getting pending quota records for account=" + account.getAccountName());
         for (UsageVO usageRecord : usageRecords.first()) {
-            BigDecimal aggregationRatio = new BigDecimal(_aggregationDuration).divide(s_minutesInMonth, 8, RoundingMode.HALF_EVEN);
             switch (usageRecord.getUsageType()) {
             case QuotaTypes.RUNNING_VM:
-                List<QuotaUsageVO> lq = updateQuotaRunningVMUsage(usageRecord, aggregationRatio);
+                List<QuotaUsageVO> lq = updateQuotaRunningVMUsage(usageRecord);
                 if (!lq.isEmpty()) {
                     quotaListForAccount.addAll(lq);
                 }
                 break;
             case QuotaTypes.ALLOCATED_VM:
-                QuotaUsageVO qu = updateQuotaAllocatedVMUsage(usageRecord, aggregationRatio);
+                QuotaUsageVO qu = updateQuotaAllocatedVMUsage(usageRecord);
                 if (qu != null) {
                     quotaListForAccount.add(qu);
                 }
@@ -159,7 +158,7 @@ public class QuotaManagerImpl extends ManagerBase implements QuotaManager {
             case QuotaTypes.VOLUME:
             case QuotaTypes.VM_SNAPSHOT:
             case QuotaTypes.BACKUP:
-                qu = updateQuotaDiskUsage(usageRecord, aggregationRatio, usageRecord.getUsageType());
+                qu = updateQuotaDiskUsage(usageRecord, usageRecord.getUsageType());
                 if (qu != null) {
                     quotaListForAccount.add(qu);
                 }
@@ -170,7 +169,7 @@ public class QuotaManagerImpl extends ManagerBase implements QuotaManager {
             case QuotaTypes.NETWORK_OFFERING:
             case QuotaTypes.SECURITY_GROUP:
             case QuotaTypes.VPN_USERS:
-                qu = updateQuotaRaw(usageRecord, aggregationRatio, usageRecord.getUsageType());
+                qu = updateQuotaRaw(usageRecord, usageRecord.getUsageType());
                 if (qu != null) {
                     quotaListForAccount.add(qu);
                 }
@@ -333,7 +332,7 @@ public class QuotaManagerImpl extends ManagerBase implements QuotaManager {
         return true;
     }
 
-    public QuotaUsageVO updateQuotaDiskUsage(UsageVO usageRecord, final BigDecimal aggregationRatio, final int quotaType) {
+    public QuotaUsageVO updateQuotaDiskUsage(UsageVO usageRecord, final int quotaType) {
         QuotaUsageVO quota_usage = null;
         QuotaTariffVO tariff = _quotaTariffDao.findTariffPlanByUsageType(quotaType, usageRecord.getEndDate());
         if (tariff != null && tariff.getCurrencyValue().compareTo(BigDecimal.ZERO) != 0) {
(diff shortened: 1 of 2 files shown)


DaanHoogland:
Your code looks alright @jairov4 but only if the aggregation is always hourly. I think there is another problem with the code and that is that the onehour vars are not named correctly; `BigDecimal onehourcostpercpu, onehourcostper100mhz, onehourcostper1mb, onehourcostforvmusage;`. I think these should have been called 'oneaggregation*'.
Have you tested with other agrregation durations?
Care to comment @abhinandanprateek , @rhtyd , @rafaelweingartner ? (not sure who else to ask)

jairov4:
@DaanHoogland yes, this calculation is independent of the aggregation ratio because the raw usage variable is the number of hours used and we only need the hour cost. As you can see in `src/main/java/com/cloud/usage/parser/VMInstanceUsageParser.java`
image

jairov4:
btw: should I rebase it to 4.15?

DaanHoogland:
You can rebase it @jairov4 , if you want it on that branch. It is unlikely an new 4.15 version will come out. if you do it will be merged forward to 4.16 as well.

rhtyd:
@blueorangutan package

rohityadavcloud:
@blueorangutan package

sureshanaparti:
@blueorangutan test

framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
```diff
@@ -340,7 +340,7 @@ public QuotaUsageVO updateQuotaDiskUsage(UsageVO usageRecord, final BigDecimal a
             BigDecimal quotaUsgage;
             BigDecimal onehourcostpergb;
             BigDecimal noofgbinuse;
-            onehourcostpergb = tariff.getCurrencyValue().multiply(aggregationRatio);
+            onehourcostpergb = tariff.getCurrencyValue().divide(s_hoursInMonth, 8, RoundingMode.HALF_DOWN);
```
DaanHoogland, line 343: I think you need to remove the parameter aggregationRotio` on this change as it is no longer used is it?

DaanHoogland (commented):
genrally good to go, but some instances of the parameter `aggregationRotio` need to be removed.

framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
```diff
@@ -368,7 +368,7 @@ public QuotaUsageVO updateQuotaDiskUsage(UsageVO usageRecord, final BigDecimal a
         QuotaTariffVO tariff = _quotaTariffDao.findTariffPlanByUsageType(QuotaTypes.CPU_NUMBER, usageRecord.getEndDate());
         if (tariff != null && tariff.getCurrencyValue().compareTo(BigDecimal.ZERO) != 0 && serviceoffering.getCpu() != null) {
             BigDecimal cpu = new BigDecimal(serviceoffering.getCpu());
-            onehourcostpercpu = tariff.getCurrencyValue().multiply(aggregationRatio);
```
sureshanaparti, line 371: Hi @jairov4 The `aggregationRatio` considers the aggregation range specified in the global setting 'usage.stats.job.aggregation.range', which would be skipped when it is removed. Please check.

GutoVeronezi: @sureshanaparti, actually, considering the global setting `usage.stats.job.aggregation.range` in the calculation turns it wrong. The global setting `usage.stats.job.aggregation.range` only defines the range between each record, but the raw usage is always inserted in hours.

The `aggregationRatio` is defined by the `aggregationDuration` (`usage.stats.job.aggregation.range`) divided by `minutesInMonth` (43200). If the setting `usage.stats.job.aggregation.range` has `1440` (1 day) as value, the cost per hour will be `0.033333333`. Otherwise, if it has `60` (1 hour) as value, the cost per hour will be `0.001388889`. 

Therefore, the lower the global setting `usage.stats.job.aggregation.range` is, the lower the cost per hour will be, which is incorrect, because the raw usage is always in hour.

jairov4: @sureshanaparti considering the `aggregationRatio` here is actually causing the missbehavior. The raw usage already comes in hours as I pointed in the image above so it is independent of that ratio.

DaanHoogland (changes requested):
@jairov4 , please remove the parameter `aggregationRatio` from methods that no longer use it.

sureshanaparti:
> @jairov4 , please remove the parameter `aggregationRatio` from methods that no longer use it.

Hi @jairov4 Can you address the outstanding comments please. Thanks.

jairov4:
@sureshanaparti  done!

sureshanaparti:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland (approved):
cltgm

sureshanaparti (approved):
code LGTM

rohityadavcloud (approved):

DaanHoogland:
@jairov4 looks good no regressions. The test_quota.py integration test is not run in either the smoke tests or in travis, So I'll run thos manually to make triple sure.

DaanHoogland:
I ran the quota tests and got all to succeed except for one that requires more than a single day of data to accumulate. recarding this as success and merging because of succesfull testing and multiple LTGM.

PR #5552, opened 2021-10-05
KVM Agent disconnect hook feature

mlsorensen:
### Description

This PR adds a way for KVM agent CommandWrappers to register code to run in case of an Agent disconnect event.  When KVM agents lose communication to the management cluster, outstanding work sent to the agent is considered failed, however the work may continue on and run to completion. This is not a big issue for things like live migration where a VM will eventually move to another host and then CloudStack will sync up with the new state, but it becomes more important for long running tasks like volume copy, where we may want to attempt to stop a copy or do some cleanup.

This change doesn't introduce much in the way of behavioral changes, it just adds the framework for CommandWrappers to utilize.  It does add simple "example" that provides cancellation of VM live migration upon agent disconnect.

The functionality is simple, we provide LibvirtComputingResource with a list to store DisconnectHooks, and then it implements the `disconnected()` method to call these hooks in the event of a loss of connectivity. These DisconnectHooks are basically just Java Thread types with a little extra metadata, which are executed in parallel with a best-effort timeout. 

The CommandWrappers are responsible for providing the implementation for their own DisconnectHooks (one or many), registering them prior to doing their work, and un-registering them when work is complete. The implementer of a DisconnectHook is responsible for dealing with any race conditions - in the live migration example it is understood that it's possible for the hook to be called immediately after live migration has succeeded but before the CommandWrapper can de-register the hook. Therefore cancellation of live migration is treated as best-effort and any errors are logged and ignored.

### Alternatives Considered
There are probably better designs for this - perhaps the agent could queue results and the management server could hold onto commands and wait for reconnect to get the results. However, that would be a larger effort, and we would still need to handle agent service stop. This DisconnectHook implementation also works for stopping of the KVM agent service and gives CommandWrappers a way to respond to that.

The DisconnectHooks themselves could be Runnables instead of Threads. I chose Threads simply because it's more straightforward to run them in parallel, however I'm open to suggestions and any changes requested as well.

### Types of changes

- [ ] Major
- [X] Minor

### How Has This Been Tested?
This is a feature I've been using in production for years, and would like to fold it back into upstream

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/disconnecthook/DisconnectHook.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/disconnecthook/DisconnectHook.java
@@ -18,7 +18,7 @@
 //
 package com.cloud.hypervisor.kvm.resource.disconnecthook;
 
-/*
+/**
 DisconnectHooks can be used to cleanup/cancel long running commands when
 connection to the management server is interrupted (which results in job
 failure). Agent CommandWrappers can register a hook with the
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/disconnecthook/DisconnectHook.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/disconnecthook/DisconnectHook.java
@@ -1,0 +1,59 @@
+//
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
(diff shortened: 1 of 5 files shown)


weizhouapache:
Great. thanks @mlsorensen for sharing !
Are you going to add more hooks ?

mlsorensen:
Yes, possibly! Wanted to feel out support for the feature before putting a lot of effort in, but I can imagine creating at least a short list of commands that would benefit most and working on those.

nvazquez:
Nice, thanks for sharing @mlsorensen!

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/disconnecthook/DisconnectHook.java
```diff
@@ -0,0 +1,59 @@
...
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+package com.cloud.hypervisor.kvm.resource.disconnecthook;
+
+/*
```
DaanHoogland, line 21:
```suggestion
/**
```

DaanHoogland (commented):
nice idea @mlsorensen , I understand this are hard coded hooks/callbacks , not configurable!?
This will probably solve a lot of old standing issues. I wonder if this could be generalised to solve some of the vmware storage issues that occur on disconnect from the vcenters as well.

weizhouapache:
@blueorangutan package

PR #5575, opened 2021-10-13
Fix storage cleanup corner case preventing VM deletion

nvazquez:
### Description

This PR prevents a corner case for the storage cleanup task observed in Vmware in which:
- There are storage pool migrations for the ROOT disk of a VM
- After the migrations, the volumes table creates new volume records, marking the ones to be deleted with state = 'Destroy'
- The storage cleanup task will attempt to expunge the volumes in 'Destroy' state, but if there are underlying storage issues then this task is not completed
- Once the storage issues are solved and the current volume is on the same storage pool as one of the volumes to be removed then the VM can be completely removed 

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial


nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan package

PR #5593, opened 2021-10-20
[UI] Fixed RBD storage connection bug when there are multiple '/', '+' characters in 'RADOS Secret' in Add Primary Storage

dhslove:
This PR fixes an issue where the RBD storage connection fails when there are multiple '/', '+' characters in the RADOS Secret in the Add Primary Storage UI.

The current code uses replace() to only change the first position where that character is found.

To solve this problem, we modify the code to use a regular expression as a parameter to replace() to achieve the same result as replaceAll .

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial


DaanHoogland:
@nvazquez should this minor fix be considered for the next RC?

PR #5647, opened 2021-11-01
assume a property is one when it isn't a number

DaanHoogland:
### Description

This PR handles an OVA that uses a property replacement for a disksize value

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/core/src/main/java/com/cloud/storage/template/OVAProcessor.java
+++ b/core/src/main/java/com/cloud/storage/template/OVAProcessor.java
@@ -246,7 +246,17 @@ public class OVAProcessor extends AdapterBase implements Processor {
             NodeList diskElements = new OVFHelper().getElementsByTagNameAndPrefix(ovfDoc, "Disk", "ovf");
             for (int i = 0; i < diskElements.getLength(); i++) {
                 Element disk = (Element)diskElements.item(i);
-                long diskSize = Long.parseLong(disk.getAttribute("ovf:capacity"));
+                String diskSizeValue = disk.getAttribute("ovf:capacity");
+                long diskSize = 1;
+                try {
+                    diskSize = Long.parseLong(diskSizeValue);
+                } catch (NumberFormatException e) {
+                    // ASSUMEably the diskSize contains a property for replacement
+                    LOGGER.warn(String.format("the disksize for disk %s is not a valid number: %s", disk.getAttribute("diskId"), diskSizeValue));
+                    // TODO parse the property to get any value can not be done at registration time
+                    //  and will have to be done at deploytime, so for orchestration purposes
+                    //  we now assume, a value of one
+                }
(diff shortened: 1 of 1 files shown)


core/src/main/java/com/cloud/storage/template/OVAProcessor.java
```diff
@@ -246,7 +246,17 @@ public long getTemplateVirtualSize(String templatePath, String templateName) thr
             NodeList diskElements = new OVFHelper().getElementsByTagNameAndPrefix(ovfDoc, "Disk", "ovf");
             for (int i = 0; i < diskElements.getLength(); i++) {
                 Element disk = (Element)diskElements.item(i);
-                long diskSize = Long.parseLong(disk.getAttribute("ovf:capacity"));
+                String diskSizeValue = disk.getAttribute("ovf:capacity");
+                long diskSize = 1;
```
DaanHoogland, line 250: not sure here if we should 'some' or 0 (zero)

core/src/main/java/com/cloud/storage/template/OVAProcessor.java
```diff
@@ -246,7 +246,17 @@ public long getTemplateVirtualSize(String templatePath, String templateName) thr
...
+                try {
+                    diskSize = Long.parseLong(diskSizeValue);
+                } catch (NumberFormatException e) {
+                    // ASSUMEably the diskSize contains a property for replacement
+                    LOGGER.warn(String.format("the disksize for disk %s is not a valid number: %s", disk.getAttribute("diskId"), diskSizeValue));
+                    // TODO parse the property to get any value can not be done at registration time
+                    //  and will have to be done at deploytime, so for orchestration purposes
+                    //  we now assume, a value of one
```
DaanHoogland, line 258: more like a design brainstorm than a TODO

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test centos7 vmware-70u1

rohityadavcloud (approved):

core/src/main/java/com/cloud/storage/template/OVAProcessor.java
harikrishna-patnala: @DaanHoogland any specific reason for completely not supporting those kinds of templates instead of assuming the size to be 1 ? I think this approach will have some effect on the usage records of the volume also.

DaanHoogland: @harikrishna-patnala this is only the template and there will be only a "real" disksize on/after deployment. This will only fix registration btw, and assumes deployment is going on as usual, probably for bad reason.

harikrishna-patnala (approved):

PR #5661, opened 2021-11-03
linstor-volume-plugin: Only create diskless assignments on nodes

rp-:
With diskful set to true, linstor will fail if it cannot create local
storage for the resource. Which in turn will make it impossible to have a
setup with just compute nodes on cloudstack.

### Description

This is a small bugfix that will enable to have cloudstack-agents without storage to work with Linstor volume plugin.
Before the plugin always tried to create a diskful assignment on nodes on connect, but if the node doesn't have
linstor storage, this would fail.
This was intentionally done to improve performance, as having a local disk makes write operations faster,
but there are other ways to configure diskful nodes within linstor(auto diskful, manual toggle disk).

### Types of changes

#### Bug Severity

- [x] BLOCKER (for compute node setups, with linstor)
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Local cluster setup, but the change is only within the linstor plugin, so should not have any impact on other components.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LinstorStorageAdaptor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LinstorStorageAdaptor.java
@@ -258,7 +258,6 @@ public class LinstorStorageAdaptor implements StorageAdaptor {
             final String rscName = getLinstorRscName(volumePath);
 
             ResourceMakeAvailable rma = new ResourceMakeAvailable();
-            rma.setDiskful(true);
             ApiCallRcList answers = api.resourceMakeAvailableOnNode(rscName, localNodeName, rma);
             checkLinstorAnswersThrow(answers);
 


rp-:
@rhtyd It would be great if this could be included in the next RC

rhtyd (approved):

rhtyd:
Changes are strictly in the Linstor plugin which we don't have any means to test. We can merge this based on author/maintainers testing and smoketests from Travis and packaging pass. No need to run actual smoketests. 
@blueorangutan package

rhtyd:
@blueorangutan package

rhtyd:
@blueorangutan package

nvazquez (approved):
LGTM, packaging and Travis pass
@rp- please confirm if manual tests pass and we will be able to merge

weizhouapache (approved):
I do not have enough knowledge to judge if this works.
but I am 100% sure this does not impact other cloudstack components ;-D

PR #5675, opened 2021-11-10
server bug fix: remove network details when network is removed

weizhouapache:
### Description

This PR removes network details when network is removed.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -3116,6 +3116,8 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
                                 if (networkAccount != null) {
                                     _networkAccountDao.remove(networkAccount.getId());
                                 }
+                                // Remove network details
+                                networkDetailsDao.removeDetails(networkFinal.getId());
                             }
 
                             final NetworkOffering ntwkOff = _entityMgr.findById(NetworkOffering.class, networkFinal.getNetworkOfferingId());
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -3116,7 +3116,7 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
                                 if (networkAccount != null) {
                                     _networkAccountDao.remove(networkAccount.getId());
                                 }
-                                // Remove network details
+
                                 networkDetailsDao.removeDetails(networkFinal.getId());
                             }
 


engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -3116,6 +3116,8 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
                                 if (networkAccount != null) {
                                     _networkAccountDao.remove(networkAccount.getId());
                                 }
+                                // Remove network details
```
DaanHoogland, line 3119: I don't think this comment is needed, given the name of the method invoked

DaanHoogland (approved):
clgtm

rhtyd:
@blueorangutan package

weizhouapache:
@blueorangutan package

GutoVeronezi (approved):
Code LTGM, I did not test it though.

rhtyd:
@blueorangutan package

sureshanaparti:
@blueorangutan package

rhtyd:

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
```diff
@@ -3116,6 +3116,8 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
                                 if (networkAccount != null) {
                                     _networkAccountDao.remove(networkAccount.getId());
                                 }
+
+                                networkDetailsDao.removeDetails(networkFinal.getId());
```
sureshanaparti, line 3120: The foreign key should serve the purpose here. But, when a network is destroyed, the removed date is set, and the record is not removed from the table (to keep/track the history at later stage when needed). The same is the case with most of the resources, where the foreign key is set with 'ON DELETE CASCADE' but unused. May be, a cleanup (enabled/disabled) for all such removed entires from the DB, have to be introduced at regular intervals (that can be set using global config).

`CONSTRAINT `fk_network_details__network_id` FOREIGN KEY `fk_network_details__network_id`(`network_id`) REFERENCES `networks`(`id`) ON DELETE CASCADE`

sureshanaparti (approved):

engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
weizhouapache: > 

@sureshanaparti 
yes, you are absolutely right. `ON DELETE CASCADE` does not work as expected because cloudstack updates `removed` field instead deleting the record from DB.

weizhouapache:
travis failures are not related to this pr

```
Tests ha enable/disable feature at cluster and zone level ... === TestName: test_ha_configure_enabledisable_across_clusterzones | Status : EXCEPTION ===
ERROR
```

and
```
Tests that an user is unable to remove, modify tags created by admin but should access ... === TestName: test_31_user_cant_remove_update_admin_tags | Status : EXCEPTION ===
ERROR
```

weizhouapache:
Merged based on 3 approvals and trillian test result.

PR #5683, opened 2021-11-12
CCC21 Hackathon: quick & dirty 2FA Google Authentication integration

rhtyd:
- Hardcodes secretkey and QR code for user (can be generated with extension)
- No per-user 2FA check and enforcement (check is not enforced when 2FA code is not entered ;)
- User scans the code, log into portal using username, password and code

I'll close this next week. This is meant only as a PoC/hack for the 2FA feature, ideally we should:
- Have per user-account enforcement and checks
- Allow a pluggable framework to allow people support multiple 2-FA plugins, for ex. Google Authenticator, SMS, email OTP ...
- The QR code should be dynamically created uniquely per user-account

How it works now:
1. User scans his 2FA QR code from users page to their Google Authenticator app:
Screenshot from 2021-11-12 15-46-16

2. Upon login they enter their username, password, 2FA code, domain:

Screenshot from 2021-11-12 15-40-55

3. Upon failure:
Screenshot from 2021-11-12 15-52-38

4. Upon success: (login as normal)
Screenshot from 2021-11-12 15-42-27

### DIFF
--- a/ui/src/components/view/InfoCard.vue
+++ b/ui/src/components/view/InfoCard.vue
@@ -47,10 +47,15 @@
                   {{ name }}
                 </h4>
               </div>
-              <span v-if="resource.username">
-                <img src="https://www.google.com/chart?chs=200x200&chld=M%%7C0&cht=qr&chl=otpauth://totp/Apache%20CloudStack%3Atest%40gmail.com?secret=7t4gabg72liipmq7n43lt3cw66fel4iz&issuer=Apache%20CloudStack"/>
-              </span>
             </slot>
+          </div>
+          <div>
+            <br/>
+            <a-divider/>
+            <p><b>2FA QR</b></p>
+            <span v-if="resource.username">
+              <img src="https://www.google.com/chart?chs=200x200&chld=M%%7C0&cht=qr&chl=otpauth://totp/Apache%20CloudStack%3Atest%40gmail.com?secret=7t4gabg72liipmq7n43lt3cw66fel4iz&issuer=Apache%20CloudStack"/>
+            </span>
           </div>
           <slot name="actions">
             <div class="tags">
--- a/ui/src/components/view/InfoCard.vue
+++ b/ui/src/components/view/InfoCard.vue
@@ -47,6 +47,9 @@
                   {{ name }}
                 </h4>
               </div>
+              <span v-if="resource.username">
+                <img src="https://www.google.com/chart?chs=200x200&chld=M%%7C0&cht=qr&chl=otpauth://totp/Apache%20CloudStack%3Atest%40gmail.com?secret=7t4gabg72liipmq7n43lt3cw66fel4iz&issuer=Apache%20CloudStack"/>
+              </span>
             </slot>
           </div>
           <slot name="actions">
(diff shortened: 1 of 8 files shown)


rhtyd:
(let's at least ensure this doesn't break unit tests)
@blueorangutan package

ui/src/components/view/InfoCard.vue
```diff
@@ -49,6 +49,14 @@
               </div>
             </slot>
           </div>
+          <div>
+            <br/>
+            <a-divider/>
+            <p><b>2FA QR</b></p>
+            <span v-if="resource.username">
+              <img src="https://www.google.com/chart?chs=200x200&chld=M%%7C0&cht=qr&chl=otpauth://totp/Apache%20CloudStack%3Atest%40gmail.com?secret=7t4gabg72liipmq7n43lt3cw66fel4iz&issuer=Apache%20CloudStack"/>
```
rhtyd, line 57: TODO: This is harcoded, needs to be generated by backend API and shown only one time to the user (with maybe some backup codes)

DaanHoogland (commented):
Looks promissing @rhtyd .
As you said in comments enforcing is missing.
I'm missing the linking to/creation of a user account here. Do we use google identities, or just the 2fa with our own credentials (and should we)?

rhtyd:
Thanks @DaanHoogland - yes by enforcing I meant that when 2FA is not provided in login it doesn't check 2FA code. In prod, we want to enforce checks at backend for all user-accounts which has 2FA enabled.

This doesn't require any integration/identities with Google;
- As a quick hack the QR code is generated by Google charts, in prod. the CloudStack UI or backend would generate that
- Google Authenticator is just the app to show us 2FA codes, other phone apps that do 2FA code by scanning QR code can work too (https://www.maketecheasier.com/google-authenticator-alternatives/). I've only used google authenticator to do the checks.

DaanHoogland:
I was thinking of the third party identity provider use-case we have, but this isgreat @rhtyd let's test
@blueorangutan test keepEnv

ui/src/components/view/InfoCard.vue
weizhouapache: @rhtyd  nice work !

I have a question, if user has not logged in yet, how to get this QR code ?

weizhouapache:
@rhtyd 
good work !

we could consider 2FA by sending code via email or SMS.

ui/src/components/view/InfoCard.vue
rhtyd: The workflow would be for the user to first log-in, then click a button to enable 2FA where they'll be shown this QR code. For this hack, I've taken shortcuts (for ex. the static QR code, instead of a randomly generated one per account).

rhtyd:
Closing this for now - will propose a proper design doc around pluggable 2FA feature.

ravening:
> Closing this for now - will propose a proper design doc around pluggable 2FA feature.

@rhtyd hello Rohit. is there any progress on this? will it be reopened again sometime in future?

rhtyd:
Hi @ravening, yes I think there'll be some progress on it soon. I've started a discussion thread on dev ML and shared this design doc: https://cwiki.apache.org/confluence/display/CLOUDSTACK/2FA+Framework+and+Plugins
Since this was just a PoC for fun, I closed it.

PR #5687, opened 2021-11-15
db: Avoid db locks if account resource limit is -1

ravening:
### Description
If account resource limit is -1 then there is no need to lock the
account along with its domain entries to check for resource count
as it wont matter anyway.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
+++ b/server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
@@ -447,9 +447,8 @@ public class ResourceLimitManagerImpl extends ManagerBase implements ResourceLim
         }
     }
 
-    private void checkAccountResourceLimit(final Account account, final Project project, final ResourceType type, long numResources) throws ResourceAllocationException {
-        // Check account limits
-        long accountResourceLimit = findCorrectResourceLimitForAccount(account, type);
+    private void checkAccountResourceLimit(final Account account, final Project project, final ResourceType type, long numResources,
+                                            long accountResourceLimit) throws ResourceAllocationException {
         long currentResourceCount = _resourceCountDao.getResourceCount(account.getId(), ResourceOwnerType.Account, type);
         long requestedResourceCount = currentResourceCount + numResources;
 
@@ -521,13 +520,21 @@ public class ResourceLimitManagerImpl extends ManagerBase implements ResourceLim
         }
 
         final Project projectFinal = project;
+
+        // Check account limits. If it's unlimited then don't lock the db rows
+        long accountResourceLimit = findCorrectResourceLimitForAccount(account, type);
+        if (Resource.RESOURCE_UNLIMITED == accountResourceLimit) {
+            s_logger.info((project == null ? "Account Name = " + account.getAccountName() : "Project Name = " + project.getName()) +
+                    " has unlimited resource count. Skipping capacity check");
+            return;
+        }
         Transaction.execute(new TransactionCallbackWithExceptionNoReturn<ResourceAllocationException>() {
             @Override
             public void doInTransactionWithoutResult(TransactionStatus status) throws ResourceAllocationException {
                 // Lock all rows first so nobody else can read it
                 lockAccountAndOwnerDomainRows(account.getId(), type);
                 // Check account limits
-                checkAccountResourceLimit(account, projectFinal, type, numResources);
+                checkAccountResourceLimit(account, projectFinal, type, numResources, accountResourceLimit);
                 // check all domains in the account's domain hierarchy
                 checkDomainResourceLimit(account, projectFinal, type, numResources);
             }


server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
```diff
@@ -521,13 +520,21 @@ public void checkResourceLimit(final Account account, final ResourceType type, l
         }
 
         final Project projectFinal = project;
+
+        // Check account limits. If it's unlimited then don't lock the db rows
```
DaanHoogland, line 524: This comment says no more than the code below, we might as well drop it.

DaanHoogland (dismissed):
clgtm

DaanHoogland:
@blueorangutan package

server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
ravening: ok

server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
```diff
@@ -521,13 +520,21 @@ public void checkResourceLimit(final Account account, final ResourceType type, l
...
+        long accountResourceLimit = findCorrectResourceLimitForAccount(account, type);
+        if (Resource.RESOURCE_UNLIMITED == accountResourceLimit) {
```
weizhouapache, line 526: @ravening 
if account resource limit is -1, it does not mean the account can use resources without limitation. the domain resource limit should be checked as well.

ravening: @weizhouapache if the account resource limit is -1 then the resource count are not incremented for this account and so the resource count wont be incremented for domain as well. so why do we need to do extra check?

if domain has two accounts: A with limit as -1 and B with limit as 100, then account A can use unlimited resources. so the used account resource value will not be added to domain resource count

weizhouapache: > @weizhouapache if the account resource limit is -1 then the resource count are not incremented for this account and so the resource count wont be incremented for domain as well. so why do we need to do extra check?
> 
@ravening 
Even if the account can use unlimited resource, the domain resource limit is still effective.
why not ignore the domain resource check and not updated the resource count (of account and domain) ?

> if domain has two accounts: A with limit as -1 and B with limit as 100, then account A can use unlimited resources. so the used account resource value will not be added to domain resource count

why not ?

ravening: @weizhouapache 

lets say a domain has limit of 100 cpu count and lets the account under this domain has cpu limit as -1.
Technically we can create as many vm's in this account without any cpu limit restrictions as the limit for this account is -1. So I can create 100 vm's each with 10 cpu cores thus making a total of 1000 cpu count . but if i increase the domain resource limit then i will get error as soon as 100 is reached.
So in that case whats the use of having -1 as resource limit for account?

weizhouapache: @ravening 
-1 means the accounts can use resource as much as they can, but the domain resource can not exceed the domain's limitataion.
when check account resource limit, the domain resource limit is also checked.
when update account resource count, the domain resource count is updated as well.

ravening: @weizhouapache but whats the point in updating the account resource when account limit is -1? as you said above, -1 means unlimited resources. then updating the domain resource limit doesnt make any sense.

if we update the domain resource count then -1 no longer has its meaning

weizhouapache: @ravening 
you know there are two level resource limitation: domain and account
if account limit is -1, then account level limit is skipped

domain resource count is the total resource of all accounts in the domain, including the accounts which have unlimit resource (to be clear, resource limit is -1).

ravening: @weizhouapache yes there are two isolation levels but can you answer the below question?

Domain has three account A, B and C
Domain resource limit is 100

A : resource limit is -1 and current resource count is 200
B: resource limit is 50 and current resource count is 30
C : resource limit is 50 and current resource count is 40


So you are saying that the total domain count should be 270 (200 + 40 + 30) and not 70 (30 + 40) ?

ravening: @weizhouapache 
Also technically speaking, I dont think there will be a case where account resource limit is -1 and domain resource limit is not -1.
if thats the case then the below check will always return false 
https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java#L439

So even though the account resource limit is -1 then it cant exceed domain resource limit, thus making "-1" value meaningless

weizhouapache: > @weizhouapache yes there are two isolation levels but can you answer the below question?
> 
> Domain has three account A, B and C Domain resource limit is 100
> 
> A : resource limit is -1 and current resource count is 200 B: resource limit is 50 and current resource count is 30 C : resource limit is 50 and current resource count is 40
> 
> So you are saying that the total domain count should be 270 (200 + 40 + 30) and not 70 (30 + 40) ?

@ravening yes, the total resource count of domain is 270.

weizhouapache: @ravening 
you may have noticed that there are two checks.
 https://github.com/apache/cloudstack/blob/4.16.0.0/server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java#L529-L532

```
                // Check account limits
                checkAccountResourceLimit(account, projectFinal, type, numResources);
                // check all domains in the account's domain hierarchy
                checkDomainResourceLimit(account, projectFinal, type, numResources);
```

these ensure that the resouce will not exceed the account and domain limitation..
the 1st check will always pass if account resource limit is -1, but 2nd check will not be skipped.
It makes sense to me.

You may need to discuss your colleagues if they expect an account with resource limit =-1 in a domain can exceed the domain resource limit.

ravening: @weizhouapache if you logically think, if account resource limit is -1 then there is no point in having domain resource limit as not -1.
Our team is fine with this. this should be same for everyone i guess. 

My point is, if the domain resource limit is not -1 then there is no point in having account resource limit as -1 since the second check you mentioned above will always fail as soon as the limit exceeds

weizhouapache: @ravening 
I think I do not need to explain you again. You know what I mean.
Assume the domain resource limit is 20, account resource limit is 80. what is the max resource the account can use ? it is 20, not 80. -1 just means the account can use all resources of the domain.
there is no special account in a domain, who can use real-unlimited resource (ignore the domain resource limit) and domain resource count does not take the account resource count into calculation. It is very strange. This change is unacceptable in my opinion.
If you want to have such account, add it to `ROOT` domain or update the domain resource limit to -1.

DaanHoogland: @ravening , I think @weizhouapache is right; If you have unlimited resources that would be still limited to the hierarchy above you, mainly your domain/project or what ever you are in. There is also a posibility that it would be limited by any domain above it, and in the end by the capacity of the cloud of course. A domain limit would otherwise not be useful at all.
say Dom1 has a limit of 10 and has two subdomains Dom2 and Dom3 which each have a limit of 6. Now if Dom2 uses all 6, Dom3 has an effective limit of 4 until Dom2 releases some. I think the same should be true for unlimited.
I thought `findCorrectResourceLimitForAccount()` already took care of that, but it seems it doesn't.  I am also sure we can find more bugs in this respect somewhere ;)

ravening: @DaanHoogland if thats the case then whats the use of using `-1` as value for resource limit? Also whats the use of having the check like `accountResourceLimit == -1 then skip something` ?

Having -1 on account resource limit and limiting it on the domain level, defeats the whole purpose of `-1` right?

If that is the case then -1 shouldnt be allowed at all on account level

weizhouapache: @ravening 
you'd better describe the problem you are facing. we can work it out together.

DaanHoogland: no it doesn't @ravening 
say Dom1 has a limit of 10
acc1 has a limit of 4
acc2 has a limit of -1
now acc1 can get 4 resources unless acc2 takes 8, but still together they can't take more than 10.
in this case the -1 is a convenience of not having to copy the domain limit, but also when changing the domain limit, you don't have to adjust acc2 for it.

ravening: @DaanHoogland @weizhouapache now i got your point. i will make the code change to check for -1 on both account and domain level

rhtyd:
@blueorangutan package

server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
ravening: @DaanHoogland @weizhouapache im closing this pr as this wont give much improvement in vm deployment time

Issue #5719, opened 2021-11-25
Cannot scale VM with deleted template


alexandru-bagu:
CS version 4.16.0

Error occurs here:
```
java.lang.NullPointerException
        at com.cloud.vm.UserVmManagerImpl.checkIfDynamicScalingCanBeEnabled(UserVmManagerImpl.java:4138)
```

sureshanaparti:
Hi @alexandru-bagu Are you able to delete the template when a VM exists / using that template?

harikrishna-patnala:
There is force delete option for templates @sureshanaparti. This looks like a genuine issue I could reproduce it.

PR #5721, opened 2021-11-26
Fix NPE on scale VM operation after the corresponding template is del…

harikrishna-patnala:
### Description

This PR fixes issue #5719 NPE during scale VM operation on a VM for which the corresponding template is already deleted.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

1. Created a VM with a template
2. Force deleted the template 
3. Dynamically scaled VM created from that template -> successful

### DIFF
--- a/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -3808,7 +3808,7 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         if (currentServiceOffering.isDynamic() && !newServiceOffering.isDynamic()) {
             removeCustomOfferingDetails(vmId);
         }
-        VMTemplateVO template = _templateDao.findById(vmForUpdate.getTemplateId());
+        VMTemplateVO template = _templateDao.findByIdIncludingRemoved(vmForUpdate.getTemplateId());
         boolean dynamicScalingEnabled = _userVmMgr.checkIfDynamicScalingCanBeEnabled(vmForUpdate, newServiceOffering, template, vmForUpdate.getDataCenterId());
         vmForUpdate.setDynamicallyScalable(dynamicScalingEnabled);
         return _vmDao.update(vmId, vmForUpdate);


shwstppr:
@blueorangutan package

weizhouapache (approved):
code lgtm

shwstppr (approved):
LGTM

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
```diff
@@ -3808,7 +3808,7 @@ public boolean upgradeVmDb(final long vmId, final ServiceOffering newServiceOffe
         if (currentServiceOffering.isDynamic() && !newServiceOffering.isDynamic()) {
             removeCustomOfferingDetails(vmId);
         }
-        VMTemplateVO template = _templateDao.findById(vmForUpdate.getTemplateId());
+        VMTemplateVO template = _templateDao.findByIdIncludingRemoved(vmForUpdate.getTemplateId());
```
sureshanaparti, line 3811: does a VM exists after its template got deleted?

sureshanaparti: -1, this may not be the right fix, it checks to update a VM whose template is deleted.

weizhouapache: > does a VM exists after its template got deleted?

@sureshanaparti 
yes, template can be forced deleted (from secondary storage), but the base image of template still exist on primary storage, vms are not impacted (reinstall might not work)

sureshanaparti: got it @weizhouapache , but is it right thing to consider the dynamic scaling setting from the deleted template?

weizhouapache: @sureshanaparti
I think it should be ok. If template is configured to support dynamic scaling, the VM image should support it as well because the vm image is created from the template.

harikrishna-patnala: yes, agree with @weizhouapache

harikrishna-patnala: @sureshanaparti would you like to review this again

sureshanaparti: @weizhouapache @harikrishna-patnala If VM existence (and it's scaling) is allowed after template deletion, better to skip dynamically scalable flag in that template (and consider as 'true'). If one wants to scale the VM when the flag is false, it is not possible to update the flag on the deleted template. What's your thoughts?

rhtyd:
@sureshanaparti @harikrishna-patnala @weizhouapache @shwstppr - what's the consensus here, are we looking for more changes, or is this PR acceptable/reviewable?
@blueorangutan package

shwstppr:
@rhtyd looks good to me. We use the same method at several other places as well

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
harikrishna-patnala: @sureshanaparti, as @shwstppr mentioned we use deleted template properties in other places also. So this procedure looks good to me and we cannot consider true because it can happen that the deleted template was not dynamically scalable. So better use the template properties as it is

sureshanaparti: lgtm, discussed offline with @harikrishna-patnala

sureshanaparti (approved):

sureshanaparti:
Minor change, merged based on the approvals and manual tests.

PR #5722, opened 2021-11-26
server, storage: Add new allocation algorithm for vm deployment

ravening:
### Description
Add a new parameter `unorderedfirstfitleastconsumed` for
`vm.allocation.algorithm` which computes the suitable hosts list
in parallel threads and does not care about priority
but still ensures that other criterias are met

Currently all vm allocation alogirhm computes the suitable host list
sequentially and also the list for sutiable storage pools.
If we have more than 200 hosts and storage pools then these checks will
be done twice and takes really long time.
If we dont care about the priority but still want all other criterias to be
met then use the new allocation algorithm which computes the host list
in parallel which is way faster.

### Types of changes

- [X] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial


weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

svenvogel:
@ravening how is the state of this PR?

Issue #4956, opened 2021-04-27
ui: Simplify code


davidjumani:
### Optional chaining
The optional chaining `?.` is a safe way to access nested object properties, even if an intermediate property doesn’t exist.
`(error.response && error.response.headers && error.response.headers['x-description']) || error.message`
using optional chaining it reduces to 
`error.response?.headers?.['x-description'] || error.message`

Ref : https://javascript.info/optional-chaining

This can also be used to invoke methods on objects, etc
This improves readability as well as simplifies the existing code

### Nullish coalescing

Nullish coalescing is an improvement on the || operator where it only checks if an object is null or undefined.
Using the || operator returns false if the object is zero or an empty string which can cause issues in the code when the value is zero but incorrectly treated as false

```
let height = 0;

alert(height || 100); // 100
alert(height ?? 100); // 0
```

Ref: https://javascript.info/nullish-coalescing-operator

### Object shorthand

Provides a simpler way when creating an object who's keys have the same name as the variables passed-in as properties
```
let cat = 'Miaow';
let dog = 'Woof';
let bird = 'Peet peet';

let someObject = {
  cat,
  dog,
  bird
}
```

Ref https://alligator.io/js/object-property-shorthand-es6/

##### ISSUE TYPE

 * Enhancement Request

nvazquez:
Hi @davidjumani @utchoang - can you please confirm if PR #5167 which has been merged fixes this issue?

davidjumani:
@nvazquez This is a code wide change, the PR only fixes it in one place

nvazquez:
Thanks @davidjumani, please advise if/when there are more PRs about this issue. Perhaps we can also track them on the issue description using a task list

Yashikaone:
Hello, I am Yashika Sarkar and I am new to open source but since ACS projects(2021) interests me I was going through the issues and I came across title **"Optioinal chainging**", I am not sure if it's an error but shouldn't it be optional chaining?
#4956
Thank you.

rohityadavcloud:
cc @borisstoyanov

Yashikaone:
@utchoang  Hello, I have come across certain lines of codes where optional chaining may be used . 
Please let me know if it is okay to modify them or not 
These are the screenshots-
![Screenshot (68)](https://user-images.githubusercontent.com/65132044/151562718-a99f7aed-cc35-43df-812f-73e414385822.png) last line
[...]

PR #5725, opened 2021-11-27
simplify code using object shorthand in storage views

Damans227:
### Description

- This PR is related to the issue: `ui: Simplify code` #4956
- It uses [object shorthand notation from Javascript - ES6 ](https://alligator.io/js/object-property-shorthand-es6/)to simplify the `.vue` files in storage view. 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial


Damans227:
@davidjumani @utchoang  Ptal, thx!

utchoang:
@davidjumani In my opinion, the shorthand object does not apply to the data() object, but only to the objects that are used in the handling functions of the method, mounted, created,...? What do you think?

sureshanaparti:
@blueorangutan ui

ui/src/views/storage/AttachVolume.vue
```diff
@@ -59,9 +59,11 @@ export default {
     }
   },
   data () {
+    let virtualmachines = [];
```
utchoang, line 62: Please remove the ";" character

utchoang (changes requested):

ui/src/views/storage/AttachVolume.vue
utchoang: Apply for the same changes

davidjumani:
While using object shorthand is a good idea, creating variables with empty / null / static values just to be returned a couple of lines down can actually reduce the readability. @Damans227 Can you look at other areas where doing so will make reading the code easier or simplify it ? Thanks

Damans227:
> While using object shorthand is a good idea, creating variables with empty / null / static values just to be returned a couple of lines down can actually reduce the readability. @Damans227 Can you look at other areas where doing so will make reading the code easier or simplify it ? Thanks

That makes sense. I will go ahead and close this PR.

Issue #5747, opened 2021-12-03
[Vmware][Deploy-as-is] DB exception when the template/appliance has empty default values


nvazquez:
Cannot deploy VM from a template when:
- 'Read Settings from OVA' = true
- The template has one or more properties without a default value (null values)

Example (missing the `ovf:value` property):
```
<Property ovf:key="SCHProxyPort" ovf:type="string" ovf:userConfigurable="true">
        <Label>Smart Call Home HTTP Proxy port number</Label>
        <Description>HTTP proxy server port number (Defaults to port 80 if omitted.)</Description>
</Property>
```

Leads to:
![image](https://user-images.githubusercontent.com/5295080/144592239-4174efd3-3a01-4e23-ac62-98336f5ca0df.png)

##### ISSUE TYPE
 * Enhancement Request

##### COMPONENT NAME
~~~
VMware
~~~

##### CLOUDSTACK VERSION
~~~
4.16 (main)
~~~

##### STEPS TO REPRODUCE
~~~
- Register a template missing default values for properties as described above using 'Read Settings from OVA' = true
- Deploy VM
~~~

##### EXPECTED RESULTS
~~~
Deployment succeeds
~~~

##### ACTUAL RESULTS
~~~
Exception (picture above)
~~~

rohityadavcloud:
Fixed in https://github.com/apache/cloudstack/pull/5748

PR #5748, opened 2021-12-03
Prevent null values on Vmware appliances details that are missing a default value

nvazquez:
### Description

This PR allows deploying VMware appliances missing default values on some properties
Fixes: #5747 

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?
- Register template missing default values on some properties using 'Read Settings from OVA' = true
- Deploy VM

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -4399,6 +4399,8 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                             value = DBEncryptionUtil.encrypt(value);
                         }
                     }
+                } else if (value == null) {
+                    value = "";
                 }
                 if (s_logger.isTraceEnabled()) {
                     s_logger.trace(String.format("setting property '%s' as '%s' with value '%s'", key, detailKey, value));


nvazquez:
@blueorangutan package

sureshanaparti:
@nvazquez can the detail be considered as null when it doesn't exist in the table, or use empty instead of null. this way, the details table is inline with other details table.

nvazquez:
Thanks @sureshanaparti, have refactored the PR
@blueorangutan package

nvazquez:
@blueorangutan package

nvazquez:

nvazquez:

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4399,6 +4399,8 @@ private void persistVMDeployAsIsProperties(UserVmVO vm, Map<String, String> user
                             value = DBEncryptionUtil.encrypt(value);
                         }
                     }
+                } else if (value == null) {
+                    value = "";
```
DaanHoogland, line 4403: @nvazquez does this require special handling in the vmware resource/- guru? I mean to make sure vmware get the right type of value.

sureshanaparti: how about not adding the key/value to map `userVmOVFPropertiesMap` when the value is null or empty?

rhtyd (approved):

shwstppr (approved):
Looks good. Though it might make sense to not store the detail altogether as empty string may end up as a used value during OVA deployment

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
nvazquez: This could cause the deployments to fail as Vmware will expect some keys even with empty values

sureshanaparti: @nvazquez Ok, In case VMware expects keys with empty values. Have you tested without the keys?

sureshanaparti: > @nvazquez Ok, In case VMware expects keys with empty values. Have you tested without the keys?

@nvazquez Have you tested without the keys (instead of keys with empty values) ?

nvazquez: Sorry @sureshanaparti I was on leave, no. Have not tested without the keys, will do

rohityadavcloud (approved):

PR #5802, opened 2021-12-22
kvm: don't always force scsi controller for aarch64 VMs

rohityadavcloud:
This would allow use of virtio disk controller with Ceph, etc or as
defined in the VM's root disk controller setting, rather than always
enforce SCSI.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

On ARM64 setup, this would now allow use of virtio disk controller for nfs and ceph. With Ceph, when virtio is used rbd performance increases drastically.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -3713,10 +3713,6 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             return null;
         }
 
-        if (_guestCpuArch != null && _guestCpuArch.equals("aarch64")) {
-            return DiskDef.DiskBus.SCSI;
-        }
-
         String rootDiskController = details.get(VmDetailConstants.ROOT_DISK_CONTROLLER);
         if (StringUtils.isNotBlank(rootDiskController)) {
             s_logger.debug("Passed custom disk controller for ROOT disk " + rootDiskController);
@@ -3750,10 +3746,6 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     }
 
     private DiskDef.DiskBus getGuestDiskModel(final String platformEmulator, boolean isUefiEnabled) {
-        if (_guestCpuArch != null && _guestCpuArch.equals("aarch64")) {
-            return DiskDef.DiskBus.SCSI;
-        }
-
         if (platformEmulator == null) {
             return DiskDef.DiskBus.IDE;
         } else if (platformEmulator.startsWith("Other PV Virtio-SCSI")) {
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -3756,6 +3756,8 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
             return DiskDef.DiskBus.VIRTIO;
         } else if (isUefiEnabled && org.apache.commons.lang3.StringUtils.startsWithAny(platformEmulator, "Windows", "Other")) {
             return DiskDef.DiskBus.SATA;
+        } else if (_guestCpuArch != null && _guestCpuArch.equals("aarch64")) {
+            return DiskDef.DiskBus.SCSI;
         } else {
             return DiskDef.DiskBus.IDE;
         }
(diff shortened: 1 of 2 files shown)


plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -3713,10 +3713,6 @@ public boolean isCentosHost() {
             return null;
         }
 
-        if (_guestCpuArch != null && _guestCpuArch.equals("aarch64")) {
```
harikrishna-patnala, line 3716: Instead of removing, is it better to move this block to the default else case, because the default case always returns IDE ? I'm thinking in terms of backward compatibility

sureshanaparti:
Following unit test failed
`LibvirtComputingResourceTest.testCreateDevicesWithSCSIDisk:432->assertNodeExists:847`

@rhtyd Can you update the unit test. Thanks.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
rhtyd: This my own hardcoded check for experimental arm64 guest CPU/platform, this is only hit for users using arm64 hosts such as Ampere or Raspberry Pi4 - so there's no backward compatibility to take into consideration.

rhtyd: ... on second thought, I've moved the code to fallback as default for aarch64 when other controller's can't be found/guessed.

rhtyd:
Thanks @sureshanaparti done (I've tested this in my ceph/rpi lab, works)
@blueorangutan package

plugins/hypervisors/kvm/src/test/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResourceTest.java
```diff
@@ -416,25 +416,6 @@ public void testCreateDevicesDef() {
         verifyDevices(devicesDef, to);
     }
 
-    @Test
```
weizhouapache, line 419: @rhtyd 
code lgtm

is it necessary to remove this unit test ?

rhtyd: Yes, as it was checking that for aarch64 scsi would always return; now it's not the case.

weizhouapache: good. just tested it can be fixed by adding ```to.setPlatformEmulator("");```

harikrishna-patnala (approved):

rhtyd:
@blueorangutan test

RohitYadavCloud:
@blueorangutan test

plugins/hypervisors/kvm/src/test/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResourceTest.java
weizhouapache: @RohitYadavCloud 
I just had a look at the test again.  I think we should not remove the test.
we could apply the following change 
```
-        libvirtComputingResourceSpy._guestCpuArch = "aarch64";
+        to.setPlatformEmulator("Other PV Virtio-SCSI");
```
unit test will pass

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@blueorangutan test

plugins/hypervisors/kvm/src/test/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResourceTest.java
GabrielBrascher: Agree with @weizhouapache on keeping the test with the needed adjustments.

rohityadavcloud:
@blueorangutan package

GabrielBrascher (approved):
Code LGTM

rohityadavcloud:
@blueorangutan test

weizhouapache (approved):

DaanHoogland:
@rohityadavcloud is further testing needed for this?

rohityadavcloud:
@DaanHoogland No, I've manually test the changes with RPi4 and smoketests with x86/kvm has passed.

PR #5815, opened 2021-12-29
API: Fix listSSHKeyPairs API when listing all resources (listall=true & projectid=-1)

Pearl1594:
### Description

This PR fixes the issue noticed when listing all ssh key pairs including ones owned by projects. Exception:
```
(primateqa) 🐱 > list sshkeypairs listall=true projectid=-1
🙈 Error: (HTTP 530, error code 4250) DB Exception on: com.mysql.cj.jdbc.ClientPreparedStatement: SELECT ssh_keypairs.id, ssh_keypairs.uuid, ssh_keypairs.account_id, ssh_keypairs.domain_id, ssh_keypairs.keypair_name, ssh_keypairs.fingerprint, ssh_keypairs.public_key FROM ssh_keypairs  INNER JOIN account ON ssh_keypairs.account_id=account.id  WH ORDER BY ssh_keypairs.id DESC  LIMIT 0, 500
```

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

#### Post Fix:
```
(localcloud) 🐱 > list sshkeypairs listall=true page=1 pagesize=20 filter=name,
{
  "count": 1,
  "sshkeypair": [
    {
      "name": "ssh"
    }
  ]
}
(localcloud) 🐱 > list sshkeypairs listall=true projectid=-1 page=1 pagesize=20 filter=name,
{
  "count": 2,
  "sshkeypair": [
    {
      "name": "test"
    },
    {
      "name": "ssh"
    }
  ]
}
```

### DIFF
--- a/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
+++ b/framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
@@ -41,6 +41,7 @@ import java.util.Date;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.TimeZone;
 import java.util.UUID;
@@ -1293,7 +1294,8 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
             fromIndex += onClause.length();
         }
 
-        str.delete(str.length() - 4, str.length());
+        int diff =  str.toString().trim().toLowerCase(Locale.ROOT).endsWith("where") ? 6 : 4;
+        str.delete(str.length() - diff, str.length());
 
         for (JoinBuilder<SearchCriteria<?>> join : joins) {
             if (join.getT().getJoins() != null) {


framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1293,7 +1294,8 @@ protected void addJoins(StringBuilder str, Collection<JoinBuilder<SearchCriteria
             fromIndex += onClause.length();
         }
 
-        str.delete(str.length() - 4, str.length());
+        int diff =  str.toString().trim().toLowerCase(Locale.ROOT).endsWith("where") ? 6 : 4;
+        str.delete(str.length() - diff, str.length());
```
rohityadavcloud, line 1298: @Pearl1594 is there any unit test to check/cover this, could this cause a regression? Maybe explore a more localised fix that's limited to ssh-key based DaoImpl?

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1293,7 +1294,8 @@ protected void addJoins(StringBuilder str, Collection<JoinBuilder<SearchCriteria
             fromIndex += onClause.length();
         }
 
-        str.delete(str.length() - 4, str.length());
+        int diff =  str.toString().trim().toLowerCase(Locale.ROOT).endsWith("where") ? 6 : 4;
```
DaanHoogland, line 1297: @Pearl1594 I'd leave out the trim() it is not clear why the 6 or 4 like this (or add a comment, but I prefer the code to explain itself)

sureshanaparti: +1, add comment if code is not self-explanatory.

weizhouapache:
@Pearl1594 
I suggest to append 'WHERE' only when it is required, like code changes below
```
@@ -1262,10 +1262,11 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
 
     @DB()
     protected void addJoins(StringBuilder str, Collection<JoinBuilder<SearchCriteria<?>>> joins) {
+        boolean hasWhereClause = true;
         int fromIndex = str.lastIndexOf("WHERE");
         if (fromIndex == -1) {
             fromIndex = str.length();
-            str.append(" WHERE ");
+            hasWhereClause = false;
         } else {
             str.append(" AND ");
         }
@@ -1287,13 +1288,20 @@ public abstract class GenericDaoBase<T, ID extends Serializable> extends Compone
             .append(" ");
             str.insert(fromIndex, onClause);
             String whereClause = join.getT().getWhereClause();
-            if ((whereClause != null) && !"".equals(whereClause)) {
+            if (!StringUtils.isEmpty(whereClause)) {
+                if (!hasWhereClause) {
+                    str.append(" WHERE ");
+                    hasWhereClause = true;
+                }
                 str.append(" (").append(whereClause).append(") AND");
             }
             fromIndex += onClause.length();
         }
 
-        str.delete(str.length() - 4, str.length());
+        if (hasWhereClause) {
+            // Remove ' AND' OR 'AND ' at the end of line
+            str.delete(str.length() - 4, str.length());
+        }
 
         for (JoinBuilder<SearchCriteria<?>> join : joins) {
             if (join.getT().getJoins() != null) {
```

rohityadavcloud:
This looks like a bugfix, should it be considered for 4.16.1 @sureshanaparti @nvazquez @DaanHoogland @weizhouapache

nvazquez:
Thanks @rohityadavcloud I agree, @sureshanaparti please include it on 4.16.1 if necessary

sureshanaparti:
@Pearl1594 Is this PR good to go in 4.16.1? If so, please address outstanding comment and rebase with 4.16. Thanks.

PR #5832, opened 2022-01-06
Fix 'endpointe.url' global settings configuration typo

GabrielBrascher:
### Description

This PR proposes updating the `endpointe.url` global settings configuration name to `endpoint.url`.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial


GabrielBrascher:
@blueorangutan package

weizhouapache (approved):
code LGTM

sureshanaparti (approved):
changes lgtm

DaanHoogland:
@GabrielBrascher you are probably absolutely right wanting to change this, but ...
1. it is high risk; did you search teh whole system for occuences of the string "pointe"?
2. it is an backwards incompatible change; global settings are user facing, and thus basically API.
that all said, I'm not loosing sleep on this one, ... if upgrade tests are successful!!!

GabrielBrascher:
@DaanHoogland I took a look at the repository and it looks like all occurrences have been addressed in this PR.

Regarding backward-incompatible, maybe this should join in the set of changes that depend on a "5.0" then.

I think that we need to start tracking/documenting all bugs/enhancements that would require a 5.0.
Otherwise, we get multiple fixes & enhancements locked due to the endless `4.X.Y.Z`.

rohityadavcloud:
LGTM but we need to update docs too @GabrielBrascher 
cc @davidjumani @Pearl1594 pl check if that breaks CKS or anything?

rohityadavcloud:
Backward compatibility is not an issue for global settings as long as there's an upgrade path which I see you've added @GabrielBrascher (I also don't think there'll be a 5.0 ever with major breaks)

GabrielBrascher:
Thanks for the review @rohityadavcloud.
I believe that **5.0** would be welcome, as long as it offers a stable upgrade process (via documentation and automation) I don't see issues.

But I agree that we would need to work hard to keep eventual Backward compatibility issues well handled in a way that would not make users avoid upgrading from **4.X** to a **5.0**.

Or we go the way Intel did with their processors. Brand new CPUs that still support instructions of decades ago xD

DaanHoogland:
> Backward compatibility is not an issue for global settings as long as there's an upgrade path which I see you've added @GabrielBrascher (I also don't think there'll be a 5.0 ever with major breaks)

yes it is, 3rd party scripts do depend on naming of global settings, @rohityadavcloud .

GabrielBrascher:
@DaanHoogland that's true in case of someone automated global settings changes during or after deployment of a Zone.

shwstppr (approved):
LGTM. CKS uses ConfigKey.value so looks okay to me

server/src/main/java/com/cloud/network/as/AutoScaleManagerImpl.java
```diff
@@ -327,7 +327,7 @@ protected AutoScaleVmProfileVO checkValidityAndPersist(AutoScaleVmProfileVO vmPr
         }
 
         if (csUrl == null || csUrl.contains("localhost")) {
-            throw new InvalidParameterValueException("Global setting endpointe.url has to be set to the Management Server's API end point");
+            throw new InvalidParameterValueException("Global setting endpoint.url has to be set to the Management Server's API end point");
```
GutoVeronezi, line 330: @GabrielBrascher, instead of only fixing the typo, we could use `ApiServiceConfiguration.ApiServletPath.key()` wherever it's referenced.

rohityadavcloud:
@DaanHoogland don't entirely agree, the listConfiguration API is privileged, not available for normal users but only root admin. Most users should consume the mgmt server endpoint URL as provided by the cloud provider. The intent of this global setting is largely for internal sub-system consumption, for example, it's currently only used by CKS or autoscaler manager (nobody uses that, not sure if it even works with netscaler now). That said I'll leave with @sureshanaparti, you, @GabrielBrascher and others to discuss and reach concensus.

server/src/main/java/com/cloud/network/as/AutoScaleManagerImpl.java
GabrielBrascher: @GutoVeronezi I've updated `AutoScaleManagerImpl.java` and `LoadBalancingRulesManagerImpl.java` to be aligned with your suggestions.
Thanks for the review!

GutoVeronezi (approved):
LGTM

GabrielBrascher:
@rohityadavcloud thanks for the feedback.
@DaanHoogland do you think it would be OK to have this global setting changed for 4.17? Or do you think it should be moved to a "5.0"?

sureshanaparti:
> @rohityadavcloud thanks for the feedback.
> @DaanHoogland do you think it would be OK to have this global setting changed for 4.17? Or do you think it should be moved to a "5.0"?

@GabrielBrascher @DaanHoogland I think, 4.17 is fine. Some global settings were changed in 4.16 as well (as part of inclusivity)

DaanHoogland:
> @rohityadavcloud thanks for the feedback. @DaanHoogland do you think it would be OK to have this global setting changed for 4.17? Or do you think it should be moved to a "5.0"?

@GabrielBrascher I'm not allowing myself being accused of being consequent or adherent to principies, however if we get a complaint of someone that changes tis configuration in an automated way or on a regular basis for some failover/regular maintenance they are going to ask us to change it back, for that reason and I would side with them. Chances are small, ....

GabrielBrascher:
@DaanHoogland I appreciate the fact that you are being cautious with any sort of backward compatibility issues.
This is something that deserves such attention, so it is really good that you raised it.
If this goes in 4.17, we might need to make sure that this is announced then, so users/admins are aware of such change.

Checking at the [Releases](https://cwiki.apache.org/confluence/display/CLOUDSTACK/Releases) in CWIKI, there is the following description regarding backward compatibility:
> The community strives to ensure **backward API compatibility** within each major version (i.e.: code written against the CloudStack 4.0.0-incubating API should work with all future 4.y.z versions)

It looks that our main effort is to keep the API backward compatibility within each version; however, it does not state covering configurations or other parts. I assume that as far as the upgrade process covers the changes, this should be OK.

However, I agree that this could lead to scripts breaking (especially on deployment, where such configuration would probably be changed). So I assume that we might need to open it for a public discussion and eventually do a few changes on documentation and cover such cases.

api/src/main/java/org/apache/cloudstack/config/ApiServiceConfiguration.java
```diff
@@ -21,7 +21,7 @@
 
 public class ApiServiceConfiguration implements Configurable {
     public static final ConfigKey<String> ManagementServerAddresses = new ConfigKey<String>("Advanced", String.class, "host", "localhost", "The ip address of management server. This can also accept comma separated addresses.", true);
-    public static final ConfigKey<String> ApiServletPath = new ConfigKey<String>("Advanced", String.class, "endpointe.url", "http://localhost:8080/client/api",
+    public static final ConfigKey<String> ApiServletPath = new ConfigKey<String>("Advanced", String.class, "endpoint.url", "http://localhost:8080/client/api",
```
sureshanaparti, line 24: this change may not be required, as it is handled in schema upgrade.

DaanHoogland: I think, if we go ahead with this, it must be in both the schema change and in this object, @sureshanaparti. Otherwise both will end up in the table.

sureshanaparti: > I think, if we go ahead with this, it must be in both the schema change and in this object, @sureshanaparti. Otherwise both will end up in the table.

ok @DaanHoogland

DaanHoogland:
@GabrielBrascher , I do not want to add to the rule base for this, let;s merge and deal with fallout as it comes/appropriate.

RodrigoDLopez (approved):
LGTM. +1

rohityadavcloud:
@GabrielBrascher why was this tagged to 5.0? We don't promise backward compatibility for global settings, in the past in rare occasions we've renamed global settings. As long as your upgrade path will rename the global setting for an existing env upon upgrade, this LGTM for main/4.17.

GabrielBrascher:
@rohityadavcloud thanks for the ping. I think I tagged as 5.0 when I saw there could be a discussion about possible backward compatibility, in order to avoid merging in 4.17 before addressing the raised concerns. 

I believe that this should be OK to be merged in 4.17, but I don't want to  overlook any review/reviewer.

rohityadavcloud:
Thanks for your consideration @GabrielBrascher I've not read all the comments above in much detail, but you've my support. I think this should be OK to merge for 4.17, let me fix the milestone and ask others on this PR - any objection to Gabriel's comment (OK to merge in 4.17)?

DaanHoogland:
merge! merge! merge!

rohityadavcloud:
@blueorangutan package

nvazquez:
@blueorangutan test matrix

nvazquez:
Merging based on approvals and test results - Vmware tests failing also on the health checks PR, not related to this PR

PR #5854, opened 2022-01-12
api,api-discovery,ui: improve listAPIs and login

### LINKED ISSUE #5266: CloudStack UI login is slow
The CloudStack UI in 4.15+ uses API auto-discovery to generated dynamic UI views, actions, paths etc. It uses `listApis` to find list of allowed APIs for the user/role which is generated on per user/account basis than per-role basis. Is there a way to make this faster, for example (1) reduce the size of list API response (based on what is actually used/consumed by the UI, for example in the json response, the params part is used but the response part may not be used), (2) pre-cache and host is statically the response based on Roles in the system than accounts, since every account has a role, (3) ship in-built json maps with the UI for default roles?, (4) run the API discovery after login screen/dashboard shows (change in view would make the UI load faster for the users)?

##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.15
~~~

shwstppr:
### Description

Partially fixes #5266

- Allows calling listAPIs API to not return anything except API name in the response.
- Refactored retrieving account and role details  during access check for each available API.
- Refactored checkAccess caller to query only enabled checkers.
- In UI, during login make two lisApis call. First to retrieve only API names to show appropriate UI and second in background to retrieve complete API data. 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/acl/APIChecker.java
+++ b/api/src/main/java/org/apache/cloudstack/acl/APIChecker.java
@@ -23,11 +23,13 @@ import com.cloud.utils.component.Adapter;
 
 // APIChecker checks the ownership and access control to API requests
 public interface APIChecker extends Adapter {
+    boolean isEnabled();
+
     // Interface for checking access for a role using apiname
     // If true, apiChecker has checked the operation
     // If false, apiChecker is unable to handle the operation or not implemented
     // On exception, checkAccess failed don't allow
     boolean checkAccess(User user, String apiCommandName) throws PermissionDeniedException;
 
-    boolean checkAccess(Account account, User user, String apiCommandName, Role role) throws PermissionDeniedException;
+    boolean checkAccessWithoutEnabledCheck(Account account, User user, String apiCommandName, Role role) throws PermissionDeniedException;
 }
--- a/api/src/main/java/org/apache/cloudstack/acl/APIChecker.java
+++ b/api/src/main/java/org/apache/cloudstack/acl/APIChecker.java
@@ -17,6 +17,7 @@
 package org.apache.cloudstack.acl;
 
 import com.cloud.exception.PermissionDeniedException;
+import com.cloud.user.Account;
 import com.cloud.user.User;
 import com.cloud.utils.component.Adapter;
 
@@ -27,4 +28,6 @@ public interface APIChecker extends Adapter {
     // If false, apiChecker is unable to handle the operation or not implemented
     // On exception, checkAccess failed don't allow
     boolean checkAccess(User user, String apiCommandName) throws PermissionDeniedException;
+
+    boolean checkAccess(Account account, User user, String apiCommandName, Role role) throws PermissionDeniedException;
 }
--- a/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
+++ b/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
@@ -25,12 +25,11 @@ import java.util.Set;
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
-import com.cloud.exception.UnavailableCommandException;
+import org.apache.cloudstack.api.APICommand;
 import org.apache.log4j.Logger;
 
-import org.apache.cloudstack.api.APICommand;
-
 import com.cloud.exception.PermissionDeniedException;
+import com.cloud.exception.UnavailableCommandException;
 import com.cloud.user.Account;
 import com.cloud.user.AccountService;
 import com.cloud.user.User;
@@ -80,10 +79,19 @@ public class StaticRoleBasedAPIAccessChecker extends AdapterBase implements APIA
             throw new PermissionDeniedException("The account id=" + user.getAccountId() + "for user id=" + user.getId() + "is null");
         }
 
-        RoleType roleType = accountService.getRoleType(account);
+        final Role accountRole = roleService.findRole(account.getRoleId());
+        return checkAccess(account, user, commandName, accountRole);
+    }
+
+    @Override
+    public boolean checkAccess(Account account, User user, String commandName, Role accountRole) throws PermissionDeniedException {
+        if (isDisabled()) {
+            return true;
+        }
+        RoleType roleType = accountRole.getRoleType();
         boolean isAllowed =
-            commandsPropertiesOverrides.contains(commandName) ? commandsPropertiesRoleBasedApisMap.get(roleType).contains(commandName) : annotationRoleBasedApisMap.get(
-                roleType).contains(commandName);
+                commandsPropertiesOverrides.contains(commandName) ? commandsPropertiesRoleBasedApisMap.get(roleType).contains(commandName) : annotationRoleBasedApisMap.get(
+                        roleType).contains(commandName);
 
         if (isAllowed) {
             return true;
--- a/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
+++ b/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
@@ -64,13 +64,14 @@ public class StaticRoleBasedAPIAccessChecker extends AdapterBase implements APIA
         }
     }
 
-    public boolean isDisabled() {
-        return roleService.isEnabled();
+    @Override
+    public boolean isEnabled() {
+        return !roleService.isEnabled();
     }
 
     @Override
     public boolean checkAccess(User user, String commandName) throws PermissionDeniedException {
-        if (isDisabled()) {
+        if (!isEnabled()) {
             return true;
         }
 
@@ -80,14 +81,11 @@ public class StaticRoleBasedAPIAccessChecker extends AdapterBase implements APIA
         }
 
         final Role accountRole = roleService.findRole(account.getRoleId());
-        return checkAccess(account, user, commandName, accountRole);
+        return checkAccessWithoutEnabledCheck(account, user, commandName, accountRole);
     }
 
     @Override
-    public boolean checkAccess(Account account, User user, String commandName, Role accountRole) throws PermissionDeniedException {
-        if (isDisabled()) {
-            return true;
-        }
+    public boolean checkAccessWithoutEnabledCheck(Account account, User user, String commandName, Role accountRole) throws PermissionDeniedException {
         RoleType roleType = accountRole.getRoleType();
         boolean isAllowed =
                 commandsPropertiesOverrides.contains(commandName) ? commandsPropertiesRoleBasedApisMap.get(roleType).contains(commandName) : annotationRoleBasedApisMap.get(
--- a/plugins/api/discovery/src/test/java/org/apache/cloudstack/discovery/ApiDiscoveryTest.java
+++ b/plugins/api/discovery/src/test/java/org/apache/cloudstack/discovery/ApiDiscoveryTest.java
@@ -85,6 +85,7 @@ public class ApiDiscoveryTest {
         s_discoveryService._apiAccessCheckers = mock(List.class);
         s_discoveryService._services = mock(List.class);
 
+        when(s_apiChecker.isEnabled()).thenReturn(true);
         when(s_apiChecker.checkAccess(any(User.class), anyString())).thenReturn(true);
         when(s_pluggableService.getCommands()).thenReturn(new ArrayList<Class<?>>());
         when(s_discoveryService._apiAccessCheckers.iterator()).thenReturn(Arrays.asList(s_apiChecker).iterator());
@@ -143,7 +144,7 @@ public class ApiDiscoveryTest {
         if (responses != null) {
             assertTrue("No. of response items > 2", responses.getCount().intValue() == 2);
             for (ApiDiscoveryResponse response : responses.getResponses()) {
-                assertTrue("API response param is not empty", response.getApiResponse().isEmpty());
+                assertTrue("API response param is not empty", response.getApiResponse() == null);
             }
         }
     }
--- a/plugins/api/discovery/src/test/java/org/apache/cloudstack/discovery/ApiDiscoveryTest.java
+++ b/plugins/api/discovery/src/test/java/org/apache/cloudstack/discovery/ApiDiscoveryTest.java
@@ -16,10 +16,27 @@
 // under the License.
 package org.apache.cloudstack.discovery;
 
-import com.cloud.user.User;
-import com.cloud.user.UserVO;
-import com.cloud.utils.component.PluggableService;
+import static org.junit.Assert.assertEquals;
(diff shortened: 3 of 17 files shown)


shwstppr:
@rohityadavcloud @weizhouapache will changes in this PR make sense? In a local mbx env I wasn't able to see huge difference though,

```
[root@qa1-cs415-kvm-7447a3fa-kvm-mgmt1 ~]# time cmk list apis > test.txt

real	0m1.981s
user	0m0.176s
sys	0m0.018s
[root@qa1-cs415-kvm-7447a3fa-kvm-mgmt1 ~]# time cmk list apis trim=true > test.txt

real	0m1.755s
user	0m0.106s
sys	0m0.017s
[root@qa1-cs415-kvm-7447a3fa-kvm-mgmt1 ~]# time cmk list apis > test.txt

real	0m2.011s
user	0m0.164s
sys	0m0.014s
[root@qa1-cs415-kvm-7447a3fa-kvm-mgmt1 ~]# time cmk list apis trim=true > test.txt

real	0m1.737s
user	0m0.102s
sys	0m0.017s
[root@qa1-cs415-kvm-7447a3fa-kvm-mgmt1 ~]# time cmk list apis > test.txt

real	0m2.053s
user	0m0.158s
sys	0m0.020s
[root@qa1-cs415-kvm-7447a3fa-kvm-mgmt1 ~]# time cmk list apis trim=true > test.txt

real	0m1.729s
user	0m0.111s
sys	0m0.013s
```

shwstppr:
@blueorangutan package

sureshanaparti:
@blueorangutan package

weizhouapache:
@shwstppr 
that's exactly what I mean.
other than this, you can pass the value of "isEnabled()" to the method.
each time when you call "isEnabled", it connects to cloudstack database. therefore there are more than 2000 db queries which are not needed.

rohityadavcloud:
@shwstppr Another way is to cache the entire listApis response cache for each role type in the system (or roles in use by account); we can try to keep this for most common/"hot" roles the algorithm is simply:
- when a list API is called -> check if an API cache exists for a role
- If not, create the response and put the list API response cached somewhere
- misc: keep a max/threadhold on this in-memory cache so this cache doesn't take say more than 100MB ram etc.
- complex/adv level: let this be served by a different/static end point where this cache is stored somewhere in memory/disk and let UI load diffs automatically (for ex. list APIs that are allowed by all + diff. APIs specific t roles)

sureshanaparti:
@shwstppr Is it possible to re-use the filter parameter (and consider empty value as trim true) ?

shwstppr:
@sureshanaparti as far as I know `filter` is a cmk/cloudmonkey param.
```
2022-01-13 06:51:28,225 DEBUG [c.c.a.ApiServlet] (qtp1603198149-21:ctx-9756f46e) (logid:4d9e7b6a) ===START===  172.20.0.1 -- GET  listall=true&page=1&pagesize=20&showIcon=true&command=listVirtualMachinesMetrics&response=json&filter=name
2022-01-13 06:51:28,233 DEBUG [c.c.a.ApiServer] (qtp1603198149-21:ctx-9756f46e ctx-2429a961) (logid:4d9e7b6a) CIDRs from which account 'Acct[f9e9afab-721d-11ec-98bd-525400ae2f99-admin] -- Account {"id": 2, "name": "admin", "uuid": "f9e9afab-721d-11ec-98bd-525400ae2f99"}' is allowed to perform API calls: 0.0.0.0/0,::/0
2022-01-13 06:51:28,239 WARN  [c.c.a.d.ParamGenericValidationWorker] (qtp1603198149-21:ctx-9756f46e ctx-2429a961) (logid:4d9e7b6a) Received unknown parameters for command listVirtualMachinesMetrics. Unknown parameters : filter
2022-01-13 06:51:28,277 DEBUG [c.c.a.ApiServlet] (qtp1603198149-21:ctx-9756f46e ctx-2429a961) (logid:4d9e7b6a) ===END===  172.20.0.1 -- GET  listall=true&page=1&pagesize=20&showIcon=true&command=listVirtualMachinesMetrics&response=json&filter=name
```
Considering empty value of `trim` as true would change default behaviour of the API that some people may not like

sureshanaparti:
> @sureshanaparti as far as I know `filter` is a cmk/cloudmonkey param.
> 
> ```
> 2022-01-13 06:51:28,225 DEBUG [c.c.a.ApiServlet] (qtp1603198149-21:ctx-9756f46e) (logid:4d9e7b6a) ===START===  172.20.0.1 -- GET  listall=true&page=1&pagesize=20&showIcon=true&command=listVirtualMachinesMetrics&response=json&filter=name
> 2022-01-13 06:51:28,233 DEBUG [c.c.a.ApiServer] (qtp1603198149-21:ctx-9756f46e ctx-2429a961) (logid:4d9e7b6a) CIDRs from which account 'Acct[f9e9afab-721d-11ec-98bd-525400ae2f99-admin] -- Account {"id": 2, "name": "admin", "uuid": "f9e9afab-721d-11ec-98bd-525400ae2f99"}' is allowed to perform API calls: 0.0.0.0/0,::/0
> 2022-01-13 06:51:28,239 WARN  [c.c.a.d.ParamGenericValidationWorker] (qtp1603198149-21:ctx-9756f46e ctx-2429a961) (logid:4d9e7b6a) Received unknown parameters for command listVirtualMachinesMetrics. Unknown parameters : filter
> 2022-01-13 06:51:28,277 DEBUG [c.c.a.ApiServlet] (qtp1603198149-21:ctx-9756f46e ctx-2429a961) (logid:4d9e7b6a) ===END===  172.20.0.1 -- GET  listall=true&page=1&pagesize=20&showIcon=true&command=listVirtualMachinesMetrics&response=json&filter=name
> ```
> 
> Considering empty value of `trim` as true would change default behaviour of the API that some people may not like

ok, got it @shwstppr

api/src/main/java/org/apache/cloudstack/acl/APIChecker.java
```diff
@@ -27,4 +28,6 @@
     // If false, apiChecker is unable to handle the operation or not implemented
     // On exception, checkAccess failed don't allow
     boolean checkAccess(User user, String apiCommandName) throws PermissionDeniedException;
+
+    boolean checkAccess(Account account, User user, String apiCommandName, Role role) throws PermissionDeniedException;
```
weizhouapache, line 32:
```suggestion
    boolean checkAccess(Account account, User user, String apiCommandName, Role role, Boolean isDynamicApiCheckerEnabled) throws PermissionDeniedException;
```

weizhouapache: @shwstppr 
as said, I suggest to add `isDynamicApiCheckerEnabled` in `checkAccess` method.

shwstppr: @weizhouapache made some changes to query only enabled APICheckers for checkAccess.
Also, realized my earlier tests were a bit skewed as I should be comparing vanilla 4.16 branch with this PR build instead of only trim parameter usage of the API. Will test and report them
cc @rohityadavcloud

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

sureshanaparti:
@blueorangutan test

shwstppr:
@rohityadavcloud @weizhouapache in my testing I found one reason for login taking considerable time is the size of listApis response. Even if remove `response` param it is still large. Therefore, I've refactored API to return just the names (as UI at first needs available APIs to show appropriate view) and in the UI make two calls first to return just the name of API and second to get the complete response. I've seen decent improvement in login time with this. I've tested different views and haven't seen any UI issue yet.
However, I'm a bit concerned if it can break UI is some cases like user instantly trying some UI action that depends on AutoGen while second API call is still in progress.
Also, I'm not sure about if caching based on roles in the backend can bring a lot of improvement. A complete listApis response just for a User role can be **1.3mb** or higher

rohityadavcloud:
That won't work @shwstppr - we use the api cache in browser to setup/configure many things (so we at least need api list with request ; but need to check if UI code does any checks for response). My suggestion was to simply cache the entire API response `per role` in memory (even though large), if/when somebody requests that that is sent without creating/processing that for individual account.

DaanHoogland:
> That won't work @shwstppr - we use the api cache in browser to setup/configure many things (so we at least need api list with request ; but need to check if UI code does any checks for response). My suggestion was to simply cache the entire API response `per role` in memory (even though large), if/when somebody requests that that is sent without creating/processing that for individual account.

Good idea, but we'd need to make it a background process and enable a refresh on it as well. What info can we give a user before all is in? can we make a smallest set of allowed data and start displaying that before we display the more restricted data.

@shwstppr code looks good so far. please let me know if this is to be tested/merged.

shwstppr:
> 
> @shwstppr code looks good so far. please let me know if this is to be tested/merged.

@DaanHoogland yes, you can test it

@blueorangutan package

shwstppr:
@blueorangutan package

DaanHoogland:
@blueorangutan test keepEnv

weizhouapache (approved):
code lgtm

in my testing, the api will be processed in less than 1 seconds (instead of around 2 seconds before). 
`trim=true` will reduces the size of api response and speeds up the api discovery. it helps.

plugins/api/rate-limit/src/main/java/org/apache/cloudstack/ratelimit/ApiRateLimitServiceImpl.java
```diff
@@ -139,24 +139,33 @@ public boolean resetApiLimit(Long accountId) {
...
             return true;
         }
-        Long accountId = user.getAccountId();
-        Account account = _accountService.getAccount(accountId);
+        Account account = _accountService.getAccount(user.getAccountId());
+        return checkAccessWithoutEnabledCheck(account, user, apiCommandName, null);
+    }
+
+    @Override
+    public boolean checkAccessWithoutEnabledCheck(Account account, User user, String apiCommandName, Role role) throws PermissionDeniedException {
```
DaanHoogland, line 158: i'd say, pass accountId instead of account and leave the accountName out of the trace message.

plugins/api/discovery/src/test/java/org/apache/cloudstack/discovery/ApiDiscoveryTest.java
```diff
@@ -16,10 +16,27 @@
 // under the License.
 package org.apache.cloudstack.discovery;
 
-import com.cloud.user.User;
-import com.cloud.user.UserVO;
-import com.cloud.utils.component.PluggableService;
+import static org.junit.Assert.assertEquals;
```
DaanHoogland, line 19: can you undo the import reorganisation, please?

DaanHoogland (approved):
code looks generally good, minor remarks
Was going to test but Wei beat me to it.

plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
```diff
@@ -80,10 +80,16 @@ public boolean checkAccess(User user, String commandName) throws PermissionDenie
...
+    }
+
+    @Override
+    public boolean checkAccessWithoutEnabledCheck(Account account, User user, String commandName, Role accountRole) throws PermissionDeniedException {
+        RoleType roleType = accountRole.getRoleType();
         boolean isAllowed =
-            commandsPropertiesOverrides.contains(commandName) ? commandsPropertiesRoleBasedApisMap.get(roleType).contains(commandName) : annotationRoleBasedApisMap.get(
-                roleType).contains(commandName);
+                commandsPropertiesOverrides.contains(commandName) ? commandsPropertiesRoleBasedApisMap.get(roleType).contains(commandName) : annotationRoleBasedApisMap.get(
+                        roleType).contains(commandName);
```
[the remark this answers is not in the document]
shwstppr: IDE did that tbh

plugins/api/discovery/src/test/java/org/apache/cloudstack/discovery/ApiDiscoveryTest.java
shwstppr: this will be hard with my IDE settings and my preference of alphabetical imports. I'm absolutely if you want to change this and push into PR branch

plugins/api/rate-limit/src/main/java/org/apache/cloudstack/ratelimit/ApiRateLimitServiceImpl.java
shwstppr: Can do but don't see particular benefit of that and will have to tinker the code (trace bit) that's working fine.

shwstppr:
@weizhouapache @DaanHoogland this might need a round of UI testing from someone (other than me) to check:
- there is a visible difference in login time
- it doesn't add any regression or breaks UI just after login in any case (URI redirect, different user role)

DaanHoogland:
> @weizhouapache @DaanHoogland this might need a round of UI testing from someone (other than me) to check:
> 
>     * there is a visible difference in login time

no noticable delay.

>     * it doesn't add any regression or breaks UI just after login in any case (URI redirect, different user role)

nothing found.

shwstppr:
@DaanHoogland @weizhouapache @rohityadavcloud @sureshanaparti cc @borisstoyanov As I've mentioned earlier my approach can break UI in some scenarios. Below is an example:
- Login with redirection `?redirect=%2Faction%2FdeployVirtualMachine`
- Before 2nd listApis call is completed (Discovering features notification still showing), click on create network. This will result in empty form
Screenshot from 2022-01-21 15-55-26

I've already added a change to hide view action buttons till feature discovery is complete but there could be still many place in UI which may show similar error when login is done with redirection (or the listApis call with complete response is still running).

I've also tried role based caching of API lists in the backend but that hardly makes any difference in UI login.
Screenshot from 2022-01-21 15-49-47

Therefore, at this point I'm of the opinion of closing this PR (or creating a new one with only APIChecker improvements)

DaanHoogland:
ok, @shwstppr (cc @sureshanaparti ) should we still consider this for 4.16?

rohityadavcloud:
@shwstppr @DaanHoogland @sureshanaparti I'm of the opinion that if there's scope of work in both dev+testing, let's move this to 4.17, @shwstppr is this ready ?

DaanHoogland:
@shwstppr , do you mean we shouldn't merge this, or that we should merge this biut not close the issue?

shwstppr:
@rohityadavcloud @DaanHoogland @sureshanaparti I'm moving this to 4.17

shwstppr:
@DaanHoogland let's revisit this 4.17.0

nvazquez:
Hi @shwstppr can you please fix the conflicts?

shwstppr:
Closed this PR as it doesn't solve the problem correctly and will break UI in some cases cc @nvazquez

Issue #4211, opened 2020-07-15
VM HA does not work when server is stopped gracefully


weizhouapache:
When we stop a server (hypervisor) gracefully, vms on it will NOT be started on other servers even HA is enabled.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME
~~~
VM HA
~~~

##### CLOUDSTACK VERSION

~~~
4.14/4.15
~~~

##### CONFIGURATION

~~~
advanced zone
~~~

##### OS / ENVIRONMENT

~~~
KVM
~~~

##### STEPS TO REPRODUCE

~~~
1. create a service offering with HA enabled
2. deploy a vm
3. go to the server where vm is running on, and stop it by "poweroff" command

~~~

##### EXPECTED RESULTS

~~~
vm will be started on other servers.

~~~

##### ACTUAL RESULTS

~~~
vm is NOT started on other servers.

only when the server is Up again, cloudstack detects the vm is stopped out-of-band and starts it.
~~~

weizhouapache:
@rhtyd @DaanHoogland @PaulAngus @svenvogel have you faced this issue ? is this an expected behavior ?

DaanHoogland:
This doesn't sound familiar @weizhouapache . To recap,
- a hosts is HA enabled? (genuine question , here)
- a vm with a ha enabled offering is created on it.
- the server is powered off. Is by means of an remote (ipmi) action or by logging on to the console and issueing a halt command?

I would not expect this behaviour as a user but it might be a side effect of the host-ha, hence my questions.

GabrielBrascher:
@weizhouapache adding to @DaanHoogland's questions:
- what is the primary storage used on the tests?

weizhouapache:
> @weizhouapache adding to @DaanHoogland's questions:
> 
> * what is the primary storage used on the tests?

@GabrielBrascher we use nfs as zone-wide storage pool.

weizhouapache:
> This doesn't sound familiar @weizhouapache . To recap,
> 
> * a hosts is HA enabled? (genuine question , here)
> * a vm with a ha enabled offering is created on it.
> * the server is powered off. Is by means of an remote (ipmi) action or by logging on to the console and issueing a halt command?
> 
> I would not expect this behaviour as a user but it might be a side effect of the host-ha, hence my questions.

@DaanHoogland 
1. do you mean OOBM based ha ? no, we do not use it.
2. all vms are ha-enabled.

[...]

PR #5862, opened 2022-01-14
KVM HA Enhancements for Ceph RBD Primary Storage

### LINKED ISSUE #9728: [Status] KVM Host-HA using CEPH RBD
Hi there, 

I would like to Draw attention to  #5862 it looks like most of the work is done to implement this new feature? Could someone look into it to make that available in an upcoming version or give some feedback of what is pending to do? 
@weizhouapache

dhslove:
### Description

This PR includes the features to provide high availability using Ceph RBD-based Primary Storage.

The KVM Host HA feature uses NFS Primary Storage. However, an additional storage consideration in KVM is Ceph RBD. This PR considers the following two to provide HA using RBD.

1. Must use the multi-monitor server feature of Ceph RBD.
2. The RBD-based HA function should be provided in the same way as the NFS-based HA.

This PR includes the feature to enter multiple monitor addresses separated by commas in the RADOS Monitor item when registering RBD Primary Storage.

And it includes adding code to provide HA based on RBD Primary Storage.

To support RBD-based HA, add RBDStoragePool class to KVMHABase and use it to perform heartbeat check and activity check in the same way as NFS. Similar to how NFS writes hb and ac files to a KVMHA directory, RADOS commands are used to put hb and ac files into the RBD pool registered as primary storage.

To perform this heartbeat and activity check, kvmheartbeat_rbd.sh and kvmvmactivity_rbd.sh files were added. This is to clearly distinguish roles from the same shell file for nfs.

This PR is necessary to support agent-free HA based on Ceph RBD like NFS, even if the related PR #4978 is merged.

The execution result of the function can be checked in the attached image.

### Types of changes

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

RBD Multiple Monitor Support - Add RBD Primary Storage
rbd-primary-add

RBD Multiple Monitor Support - RBD Primary Storage List
primarystorage-list

RBD Multiple Monitor Support - KVM Host Secret List
secret-list

RBD Multiple Monitor Support - VM Disk XML(Multi RBD Monitor)
disk-xml

RBD Based KVM Host HA - One of Hosts Off
host-list

RBD Based KVM Host HA - Host Suspect State
host-status

RBD Based KVM Host HA - Host Down State
host-down1


sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan test

wido:
This looks very interesting. I'll review this and get back. Give me a few days

wido:
May I ask why you are not just looking up if there is a Watcher on RBD images?

dhslove:
> 

Thanks your review. Wido.

The reason is 
1. To use the same method as NFS (modified time check), 
2. RBD Watcher continues to exist when the KVM host hangs (Watcher's inaccuracy in some cases).

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -243,6 +257,10 @@ public DataStore initialize(Map<String, Object> dsInfos) {
             if (port == -1) {
                 port = 0;
             }
+            if (multi) {
+                storageHost = storageHost + (hostPath.substring(0, hostPath.lastIndexOf("/")).replaceAll("/", ","));
+                hostPath = hostPath.substring(hostPath.lastIndexOf("/")+1);
```
GutoVeronezi, line 262:
```suggestion
                hostPath = hostPath.substring(hostPath.lastIndexOf("/") + 1);
```

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -162,10 +168,18 @@ public DataStore initialize(Map<String, Object> dsInfos) {
...
+                String uriHost = uri.getHost();
                 String uriPath = uri.getPath();
                 if (uriPath == null) {
                     throw new InvalidParameterValueException("host or path is null, should be rbd://hostname/pool");
                 }
+                if (multi) {
+                    String multiHost = uriHost + (uriPath.substring(0, uriPath.lastIndexOf("/")).replaceAll("/", ","));
+                    String[] hostArr = multiHost.split(",");
+                    if (hostArr.length > 5) {
+                        throw new InvalidParameterValueException("RADOS monitor can support up to 5.");
```
GutoVeronezi, line 180:
```suggestion
                        throw new InvalidParameterValueException("RADOS monitor can support up to 5 hosts.");
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFenceCommandWrapper.java
```diff
@@ -47,20 +48,25 @@ public Answer execute(final FenceCommand command, final LibvirtComputingResource
...
          * We can only safely fence off hosts when we use NFS
          * On NFS primary storage pools hosts continuesly write
          * a heartbeat. Disable Fencing Off for hosts without NFS
          */
-        if (pools.size() == 0) {
+        if (nfspools.size() == 0) {
             String logline = "No NFS storage pools found. No way to safely fence " + command.getVmName() + " on host " + command.getHostGuid();
             s_logger.warn(logline);
             return new FenceAnswer(command, false, logline);
+        }else if (rbdpools.size() == 0) {
```
GutoVeronezi, line 63:
```suggestion
        } else if (rbdpools.size() == 0) {
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFenceCommandWrapper.java
```diff
@@ -47,20 +48,25 @@ public Answer execute(final FenceCommand command, final LibvirtComputingResource
...
+            String logline = "No RBD storage pools found. No way to safely fence " + command.getVmName() + " on host " + command.getHostGuid();
```
GutoVeronezi, line 64: We could use `String.format` here.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtCheckVMActivityOnStoragePoolCommandWrapper.java
```diff
@@ -43,9 +44,13 @@ public Answer execute(final CheckVMActivityOnStoragePoolCommand command, final L
...
         final KVMHAMonitor monitor = libvirtComputingResource.getMonitor();
         final StorageFilerTO pool = command.getPool();
-        if (Storage.StoragePoolType.NetworkFilesystem == pool.getType()){
+        if (Storage.StoragePoolType.NetworkFilesystem == pool.getType() || Storage.StoragePoolType.RBD == pool.getType()){
             final NfsStoragePool nfspool = monitor.getStoragePool(pool.getUuid());
-            final KVMHAVMActivityChecker ha = new KVMHAVMActivityChecker(nfspool, command.getHost().getPrivateNetwork().getIp(), command.getVolumeList(), libvirtComputingResource.getVmActivityCheckPath(), command.getSuspectTimeInSeconds());
+            final RbdStoragePool rbdpool = monitor.getRbdStoragePool(pool.getUuid());
+            String vmActivityCheckPath = "";
+            if (Storage.StoragePoolType.NetworkFilesystem == pool.getType())    vmActivityCheckPath = libvirtComputingResource.getVmActivityCheckPath();
+            else if (Storage.StoragePoolType.RBD == pool.getType())    vmActivityCheckPath = libvirtComputingResource.getVmActivityCheckPathRbd();
```
GutoVeronezi, line 52: According to our [coding conventions](https://cwiki.apache.org/confluence/display/CLOUDSTACK/Coding+conventions),  blocks must follow the following layout:

```java
if (condition) {
    statements;
} else {
    statements;
}
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -181,6 +206,26 @@ protected String runScriptRetry(String cmdString, OutputInterpreter interpreter)
...
+                if (host.matches(ipRegex)) {
+                    sourceHostIP += host + ",";
+                } else {
+                    InetAddress addr = InetAddress.getByName(host);
+                    sourceHostIP += addr.getHostAddress() + ",";
+                }
+            }
+            return sourceHostIP;
+        } catch (UnknownHostException e) {
+            s_logger.debug("Failed to get connection: " + e.getMessage());
```
GutoVeronezi, line 224: We could pass the exception as parameter.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,221 @@
...
+                Set<String> removedPools = new HashSet<>();
+                for (String uuid : nfsstoragePool.keySet()) {
+                    NfsStoragePool primaryStoragePool = nfsstoragePool.get(uuid);
+                    StoragePool storage;
+                    try {
+                        Connect conn = LibvirtConnection.getConnection();
+                        storage = conn.storagePoolLookupByUUIDString(uuid);
+                        if (storage == null || storage.getInfo().state != StoragePoolState.VIR_STORAGE_POOL_RUNNING) {
+                            if (storage == null) {
+                                s_logger.debug(String.format("Libvirt storage pool [%s] not found, removing from HA list.", uuid));
```
GutoVeronezi, line 129:
```suggestion
                                s_logger.debug(String.format("Libvirt storage pool [%s] was not found, removing it from HA list.", uuid));
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,221 @@
...
+                            } else {
+                                s_logger.debug(String.format("Libvirt storage pool [%s] found, but not running, removing from HA list.", uuid));
```
GutoVeronezi, line 131:
```suggestion
                                s_logger.debug(String.format("Libvirt storage pool [%s] was found, but it is not running, removing it from HA list.", uuid));
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,221 @@
...
-                        continue;
                     }
 
-                    s_logger.debug(String.format("Found NFS storage pool [%s] in libvirt, continuing.", uuid));
+                    String result = null;
+                    for (int i = 1; i <= _heartBeatUpdateMaxTries; i++) {
+                        Script cmd = createHeartBeatCommand(primaryStoragePool, hostPrivateIp, true);
+                        result = cmd.execute();
+
+                        s_logger.debug(String.format("The command (%s), to the pool [%s], has the result [%s].", cmd.toString(), uuid, result));
```
GutoVeronezi, line 156:
```suggestion
                        s_logger.debug(String.format("The command [%s], to the pool [%s], had the result [%s].", cmd.toString(), uuid, result));
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,221 @@
...
-                StoragePool storage;
-                try {
-                    Connect conn = LibvirtConnection.getConnection();
-                    storage = conn.storagePoolLookupByUUIDString(uuid);
-                    if (storage == null || storage.getInfo().state != StoragePoolState.VIR_STORAGE_POOL_RUNNING) {
-                        if (storage == null) {
-                            s_logger.debug(String.format("Libvirt storage pool [%s] not found, removing from HA list.", uuid));
-                        } else {
-                            s_logger.debug(String.format("Libvirt storage pool [%s] found, but not running, removing from HA list.", uuid));
+        if(nfsstoragePool != null && !nfsstoragePool.isEmpty()) {
```
GutoVeronezi, line 118:
```suggestion
        if (nfsstoragePool != null && !nfsstoragePool.isEmpty()) {
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,221 @@
...
+                if (!removedPools.isEmpty()) {
+                    for (String uuid : removedPools) {
+                        removeStoragePool(uuid);
+                    }
+                }
+            }
+        }
 
-                    s_logger.debug(String.format("The command (%s), to the pool [%s], has the result [%s].", cmd.toString(), uuid, result));
+        if (rbdstoragePool != null && !rbdstoragePool.isEmpty()) {
```
GutoVeronezi, line 186: This code and the one above look quite similar, we could extract it to a method

dhslove: @GutoVeronezi  OK, I made method 'run

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
dhslove: I modified the code as you requested.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
dhslove: @GutoVeronezi     I modified the code as you requested.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
dhslove: @GutoVeronezi   I modified the code as you requested.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtCheckVMActivityOnStoragePoolCommandWrapper.java
dhslove: @GutoVeronezi   I modified the code as you requested.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFenceCommandWrapper.java
dhslove: @GutoVeronezi    I modified the code as you requested.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtFenceCommandWrapper.java
dhslove: @GutoVeronezi   I modified the code as you requested.

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
dhslove: @GutoVeronezi   I modified the code as you requested.

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
dhslove: @GutoVeronezi   I modified the code as you requested.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
DaanHoogland: I agree with @GutoVeronezi and you could do more than just the `runHeartbeatToPool()`. The rest of the block can be parameterised with the synchronisation object, `nfsstoragePool` and `rbdstoragePool` respectively. `obj.getClass()` to get the type and instantiate the and get the object in the loop, or add the type ars parameter as well.

no blocker, just a suggestion!

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,189 @@
...
+    private Set<String> runHeartbeatToPool(NfsStoragePool nfsStoragePool, RbdStoragePool rbdStoragePool, String uuid, Set<String> removedPools) {
+        StoragePool storage;
+        try {
+            Connect conn = LibvirtConnection.getConnection();
+            storage = conn.storagePoolLookupByUUIDString(uuid);
+            if (storage == null || storage.getInfo().state != StoragePoolState.VIR_STORAGE_POOL_RUNNING) {
+                if (storage == null) {
+                    s_logger.debug(String.format("Libvirt storage pool [%s] not found, removing from HA list.", uuid));
+                } else {
+                    s_logger.debug(String.format("Libvirt storage pool [%s] was found, but it is not running, removing it from HA list.", uuid));
```
DaanHoogland, line 164: not as strict on logging as @GutoVeronezi but one suggestion (making it only once)
```suggestion
                    s_logger.debug(String.format("Libvirt storage pool [%s] was found, but it is not running, removing it from the HA list.", uuid));
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -181,6 +206,26 @@ protected String runScriptRetry(String cmdString, OutputInterpreter interpreter)
         return result;
     }
 
+    public static String getRbdMonIpAddress(String sourceHost) {
+        try {
+            String[] hostArr = sourceHost.split(",");
+            String sourceHostIP = "";
+            for (String host : hostArr) {
+                String ipRegex = "(\\d{1,2}|1\\d\\d|2[0-4]\\d|25[0-5])\\.(\\d{1,2}|1\\d\\d|2[0-4]\\d|25[0-5])\\.(\\d{1,2}|1\\d\\d|2[0-4]\\d|25[0-5])\\.(\\d{1,2}|1\\d\\d|2[0-4]\\d|25[0-5])";
```
wido, line 214: Does this support IPv6?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -181,6 +206,26 @@ protected String runScriptRetry(String cmdString, OutputInterpreter interpreter)
...
+                if (host.matches(ipRegex)) {
+                    sourceHostIP += host + ",";
+                } else {
+                    InetAddress addr = InetAddress.getByName(host);
```
wido, line 218: Does this support IPv6?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAChecker.java
```diff
@@ -71,6 +73,30 @@ public Boolean checkingHeartBeat() {
             }
         }
 
+        for (RbdStoragePool pool : rbdStoragePools) {
+            Script cmd = new Script(s_heartBeatPathRbd, heartBeatCheckerTimeout, s_logger);
+            cmd.add("-i", getRbdMonIpAddress(pool._poolSourceHost));
+            cmd.add("-p", pool._poolMountSourcePath);
+            cmd.add("-n", pool._poolAuthUserName);
+            cmd.add("-s", pool._poolAuthSecret);
```
wido, line 81: Should we not use the UUID of the host? IPs can change, UUIDs should not

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.sh
```diff
@@ -0,0 +1,151 @@
...
+  exit 2
+fi
+
+keyringFile="/etc/ceph/keyring"
+confFile="/etc/ceph/ceph.conf"
+
+create_cephConf() {
+#Creating Ceph keyring and conf for executing rados commands
+  if [ ! -f $keyringFile ]; then
+    echo -e "[client.$PoolAuthUserName]\n key=$PoolAuthSecret" > $keyringFile
```
wido, line 84: We should never write to files in /etc/ceph from a script like this.

CloudStack hosts can work without files in /etc/ceph

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.sh
```diff
@@ -0,0 +1,151 @@
...
+  fi
+  
+  if [ ! -f $confFile ]; then
+    confContents="[global]\n mon host = "
+    for ip in $(echo $SourceHostIP | sed 's/,/ /g'); do
+        confContents+="[v2:${ip}:3300/0,v1:${ip}:6789/0], "
+    done
+    echo -e "$confContents" | sed 's/, $//' > $confFile
```
wido, line 92: Same here. We should not write to /etc/ceph/ceph.conf in any circumstance.

Not needed as well

wido (commented):
I am not convinced that this is the best route forward.

RBD already supports watchers and exclusive locking on images by default. If a host dies you should see within 30 sec or so if it is no longer watching a RBD image and also has not refreshed it's lock anymore.

This external script we are calling also does not support IPv6 (InetAddress) and would rely on external tooling.

Imho we should have each Host maybe just 'watch' a RBD image. Other hosts can check if there still is a watcher on that image. If not we can assume the Host has died.

Overall I think we should not merge this and rely on the things Ceph already has like exclusive locking and watchers on RBD images.

dhslove:
@wido 

Based on your comments, I changed the heartbeat/activity check method to the rbd watcher check method.
In this commit, unnecessary code has been removed to support IPv4/6, and code changes have been applied to prevent duplicate code.
Code review please.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -57,6 +58,28 @@ public NfsStoragePool(String poolUUID, String poolIp, String poolSourcePath, Str
...
+        String _poolMountSourcePath;
+        String _mountDestPath;
+        PoolType _type;
+        String _poolAuthUserName;
+        String _poolAuthSecret;
+        String _poolSourceHost;
+
+        public RbdStoragePool(String poolUUID, String poolIp, String poolSourcePath, String mountDestPath, PoolType type, String poolAuthUserName, String poolAuthSecret, String poolSourceHost) {
+            _poolUUID = poolUUID;
+            _poolIp = poolIp;
```
wido, line 73: I would not call this 'poolIP'. The correct naming is the mon_host as it can contain an IP-address of a hostname.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAChecker.java
```diff
@@ -71,6 +73,29 @@ public Boolean checkingHeartBeat() {
             }
         }
 
+        for (RbdStoragePool pool : rbdStoragePools) {
+            Script cmd = new Script(s_heartBeatPathRbd, heartBeatCheckerTimeout, s_logger);
+            cmd.add("-i", pool._poolSourceHost);
+            cmd.add("-p", pool._poolMountSourcePath);
+            cmd.add("-n", pool._poolAuthUserName);
+            cmd.add("-s", pool._poolAuthSecret);
+            cmd.add("-h", hostIp);
```
wido, line 82: I know we use this with NFS, but isn't it better to use the UUID of the host here?

Each host has a unique UUID and that seems safer to use then the Private IPv4 address of the host.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAChecker.java
```diff
@@ -71,6 +73,29 @@ public Boolean checkingHeartBeat() {
...
+            cmd.add("-r");
+            OutputInterpreter.OneLineParser parser = new OutputInterpreter.OneLineParser();
+            String result = cmd.execute(parser);
+            String parsedLine = parser.getLine();
+
+            s_logger.debug(String.format("Checking heart beat with KVMHAChecker [{command=\"%s\", result: \"%s\", log: \"%s\", pool: \"%s\"}].", cmd.toString(), result, parsedLine,
+                    pool._poolIp));
+
+            if (result == null && parsedLine.contains("DEAD")) {
+                s_logger.warn(String.format("Checking heart beat with KVMHAChecker command [%s] returned [%s]. [%s]. It may cause a shutdown of host IP [%s].", cmd.toString(),
```
wido, line 92: Then also use the UUID here.

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -139,7 +139,13 @@ public DataStore initialize(Map<String, Object> dsInfos) {
         PrimaryDataStoreParameters parameters = new PrimaryDataStoreParameters();
 
         URI uri = null;
+        boolean multi = false;
         try {
+            String urlType = url.substring(0, 3);
```
wido, line 144: Should we not do this split after we have used the URI class and then just split the 'host' part?

Now you are doing manual string parsing on a string which we will then run through a library.

plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java
```diff
@@ -162,10 +168,18 @@ public DataStore initialize(Map<String, Object> dsInfos) {
                     throw new InvalidParameterValueException("host or path is null, should be sharedmountpoint://localhost/path");
                 }
             } else if (uri.getScheme().equalsIgnoreCase("rbd")) {
+                String uriHost = uri.getHost();
                 String uriPath = uri.getPath();
                 if (uriPath == null) {
                     throw new InvalidParameterValueException("host or path is null, should be rbd://hostname/pool");
                 }
+                if (multi) {
```
wido, line 176: We can probably just split uriHost here.

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.sh
```diff
@@ -0,0 +1,136 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+help() {
+  printf "Usage: $0 
+                    -p rbd pool name
+                    -n pool auth username
+                    -s pool auth secret
+                    -h host
+                    -i source host ip
```
wido, line 25: Again, I suggest using the UUID of the Host

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.sh
```diff
@@ -0,0 +1,136 @@
...
+                    -r cretae/read hb watcher
```
wido, line 26: Typo

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.sh
```diff
@@ -0,0 +1,136 @@
...
+     help
+     ;;
+  esac
+done
+
+if [ -z "$PoolName" ]; then
+  exit 2
+fi
+
+keyringFile="/etc/cloudstack/agent/keyring"
```
wido, line 73: This will break if multiple Ceph clusters are attached to the same host.

This is technically possible and also done (we do this for example).

At least you need a unique keyfile per Ceph cluster, but I would even say that we should not write this credentials to a file anyway.

If we do so at least write it to a temporary directory in /run somewhere and make sure the chmod permissions are set properly so no other user can read it.

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.sh
wido: This still stands. We should not write to /etc, never.

IF we need such a file, use /run/cloudstack/agent (for example) to make sure the files are gone after a reboot.

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.sh
```diff
@@ -0,0 +1,136 @@
...
+  else
+    echo "=====> Considering host as DEAD due to [hb-$HostIP] watcher that the host is seeing is not running. <======"
+  fi
+  delete_cephKeyring
+  exit 0
+elif [ "$cflag" == "1" ]; then
+  /usr/bin/logger -t heartbeat "kvmheartbeat_rbd.sh reboots the system because there is no heartbeat watcher."
+  sync &
+  sleep 5
+  echo b > /proc/sysrq-trigger
```
wido, line 129: Do we really need to kill the host? Since Ceph/RBD lives in userspace we could also 'just' kill all the VMs which are using this storage pool.

No need to reboot the host completely.

With NFS you have something hanging in kernel space and causes I/O to lock up. This is not the case with RBD.

scripts/vm/hypervisor/kvm/kvmvmactivity_rbd.sh
```diff
@@ -0,0 +1,107 @@
...
+     ;;
+  esac
+done
+
+if [ -z "$PoolName" ]; then
+  exit 2
+fi
+
+#Creating Ceph keyring for executing rbd commands
+keyringFile="/etc/cloudstack/agent/keyring.bin"
```
wido, line 68: See my earlier comment about these keyring files.

wido (commented):
First: Thanks for the changes!

Overall it's an improvement, but I don't think we are there yet.

I also doubt if Bash is the proper language to do this checking. Python has great librbd bindings which even make checking the RBD watcher easier: https://docs.ceph.com/en/pacific/rbd/api/librbdpy/

And also: https://github.com/ceph/ceph/blob/master/src/pybind/rbd/rbd.pyx

<pre>watchers_list()</pre>

Also my comment in general:

- We should probably not write the RBD secrets to a file. We can just pass them as arguments
- Keep in mind multiple Ceph/RBD pools can be in use in a CloudStack setup
- Is Bash really the best language here?
- We should probably use the host UUID and not Private IPv4 address

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -32,9 +32,10 @@
     private static final Logger s_logger = Logger.getLogger(KVMHABase.class);
     private long _timeout = 60000; /* 1 minutes */
     protected static String s_heartBeatPath;
+    protected static String s_heartBeatPathRbd;
     protected long _heartBeatUpdateTimeout = 60000;
     protected long _heartBeatUpdateFreq = 60000;
-    protected long _heartBeatUpdateMaxTries = 5;
+    protected long _heartBeatUpdateMaxTries = 10;
```
ravening, line 38: Instead of changing the existing value, is it better to use new variable for ceph?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,189 @@
...
-    private static synchronized void configureHeartBeatPath(String scriptPath) {
+    private static synchronized void configureHeartBeatPath(String scriptPath, String scriptPathRbd) {
         KVMHABase.s_heartBeatPath = scriptPath;
+        KVMHABase.s_heartBeatPathRbd = scriptPathRbd;
     }
 
     public void addStoragePool(NfsStoragePool pool) {
-        synchronized (storagePool) {
-            storagePool.put(pool._poolUUID, pool);
+        synchronized (nfsstoragePool) {
```
ravening, line 64: Since you are already using `ConcurrentHashMap`, I dont see the need of synchronized

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,189 @@
...
-                            s_logger.debug(String.format("Libvirt storage pool [%s] not found, removing from HA list.", uuid));
-                        } else {
-                            s_logger.debug(String.format("Libvirt storage pool [%s] found, but not running, removing from HA list.", uuid));
-                        }
-
-                        removedPools.add(uuid);
-                        continue;
-                    }
+    public NfsStoragePool getStoragePool(String uuid) {
+        synchronized (nfsstoragePool) {
```
ravening, line 108: `synchronized` is not needed here. same case below also.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +35,189 @@
...
+            if (nfsStoragePool != null) {
+                cmd = createHeartBeatCommand(nfsStoragePool, null, false);
+                result = cmd.execute();
+            } else if (rbdStoragePool != null) {
+                cmd = createRbdHeartBeatCommand(rbdStoragePool, null, false);
+                result = cmd.execute();
             }
         }
 
+        return removedPools;
```
ravening, line 220: You are not doing anything with the returned set in line 125. So `return` is not needed

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAVMActivityChecker.java
```diff
@@ -27,36 +28,50 @@
...
     }
 
     @Override
     public Boolean checkingHeartBeat() {
         Script cmd = new Script(vmActivityCheckPath, activityScriptTimeout.getStandardSeconds(), LOG);
-        cmd.add("-i", nfsStoragePool._poolIp);
-        cmd.add("-p", nfsStoragePool._poolMountSourcePath);
-        cmd.add("-m", nfsStoragePool._mountDestPath);
+        String poolIp = "";
+        if (poolType == StoragePoolType.NetworkFilesystem) {
```
ravening, line 53: Rather than adding `if/else` extract them according the pool type. In future, if we support more pool type, they can goto their own functions else code will look messy

dhslove:
@wido @ravening 

Thank you for your comments. We took your comments and modified the code.

1. Use Python, librbd.
2. Use Ceph Watcher for hb, ac work.
3. Key file and secret are created and used temporarily, and temporary files are created in the /run/cloudstack/agent directory.

We didn't apply your comment to the code to use host uuid as the name of the hb image. In order for the HA program to use the Host uuid, the management side code needs to be changed, but we think that using the existing code is a better way.

dhslove:
@weizhouapache 

I checked the issue(#4211) you requested and added handling of RBD to KVMInvestigator.java.

And the fundamental problem of this issue(#4211) is that HA does not work normally when the host is in Off state. To solve this, if the host is already off in the recovering stage, it is reasonable to treat it as recover failure and then switch to the fencing stage. I solved this problem by modifying the source of KVMHAProvider.java.

weizhouapache:
@dhslove good feature.

multiple ceph rbd monitors are supported in #6792 , can you update this PR to fix the conflicts ?
can you please also address Wido's outstanding comments ?

dhslove:
@weizhouapache Sorry for the late reply. A source conflict was fixed. Please check.

wido (changes requested):

scripts/vm/hypervisor/kvm/kvmheartbeat_rbd.py
```diff
@@ -0,0 +1,115 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
dhslove: @wido I have applied your request.

scripts/vm/hypervisor/kvm/kvmvmactivity_rbd.py
```diff
@@ -0,0 +1,103 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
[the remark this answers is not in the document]
dhslove: @wido I have applied your request.

wido (approved):
Code looks good

DaanHoogland:
@GutoVeronezi @ravening I resolved all outdated comments. cc @weizhouapache. Do you giys approve?

weizhouapache:
@blueorangutan package

weizhouapache:
@dhslove 
can you solve the compilation error ?

```
[ERROR] COMPILATION ERROR : 
[ERROR] /home/travis/build/apache/cloudstack/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java:[241,16] error: cannot find symbol
  symbol:   variable multi
  location: class CloudStackPrimaryDataStoreLifeCycleImpl
[ERROR] /home/travis/build/apache/cloudstack/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java:[241,15] error: illegal parenthesized expression
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project cloud-plugin-storage-volume-default: Compilation failure: Compilation failure: 
[ERROR] /home/travis/build/apache/cloudstack/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java:[241,16] error: cannot find symbol
[ERROR]   symbol:   variable multi
[ERROR]   location: class CloudStackPrimaryDataStoreLifeCycleImpl
[ERROR] /home/travis/build/apache/cloudstack/plugins/storage/volume/default/src/main/java/org/apache/cloudstack/storage/datastore/lifecycle/CloudStackPrimaryDataStoreLifeCycleImpl.java:[241,15] error: illegal parenthesized expression
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
[ERROR] 
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <args> -rf :cloud-plugin-storage-volume-default
The command "./tools/travis/install.sh" failed and exited with 1 during .
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -57,6 +58,28 @@ public NfsStoragePool(String poolUUID, String poolIp, String poolSourcePath, Str
         }
     }
 
+    public static class RbdStoragePool {
+        String _poolUUID;
```
GutoVeronezi, line 62: IMHO, using strict camel case (non consecutive capitals/treat acronyms as normal words) would improve readability; with this approach, the variable name would be `poolUuid`. Also, this approach is already being applied in the class name: it is `RbdStoragePool` instead of  `RBDStoragePool`.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHABase.java
```diff
@@ -57,6 +58,28 @@ public NfsStoragePool(String poolUUID, String poolIp, String poolSourcePath, Str
...
+        String _monHost;
+        String _poolMountSourcePath;
+        String _mountDestPath;
+        PoolType _type;
+        String _poolAuthUserName;
+        String _poolAuthSecret;
+        String _poolSourceHost;
+
+        public RbdStoragePool(String poolUUID, String monHost, String poolSourcePath, String mountDestPath, PoolType type, String poolAuthUserName, String poolAuthSecret, String poolSourceHost) {
```
GutoVeronezi, line 71: The same remark about applying camel case to acronyms.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAChecker.java
```diff
@@ -71,6 +78,46 @@ public Boolean checkingHeartBeat() {
...
+            rbdpools._monHost));
+
+            if (process != null && parsedLine.contains("DEAD")) {
+                s_logger.warn(String.format("Checking heart beat with KVMHAChecker command [%s] returned. [%s]. It may cause a shutdown of host IP [%s].", processBuilder.command().toString(),
+                        parsedLine, hostIp));
+            } else {
+                validResult = true;
+            }
+        }
+
```
GutoVeronezi, line 120: This block could be extracted to a method, to better organize the code.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAChecker.java
```diff
@@ -71,6 +78,46 @@ public Boolean checkingHeartBeat() {
...
+            processBuilder.command().add("r");
+            Process process = null;
+            String parsedLine = "";
+            try {
+                process = processBuilder.start();
+                BufferedReader bfr = new BufferedReader(new InputStreamReader(process.getInputStream()));
+                parsedLine = bfr.readLine();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
```
GutoVeronezi, line 108: We should use the class `com.cloud.utils.script.Script` here (same as it is being doing to NFS), as it already does all the command handling.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +38,168 @@
...
     private final boolean rebootHostAndAlertManagementOnHeartbeatTimeout;
 
     private final String hostPrivateIp;
 
-    public KVMHAMonitor(NfsStoragePool pool, String host, String scriptPath) {
+    public KVMHAMonitor(NfsStoragePool pool, RbdStoragePool rbdpool, String host, String scriptPath, String scriptPathRbd) {
         if (pool != null) {
-            storagePool.put(pool._poolUUID, pool);
+            nfsstoragePool.put(pool._poolUUID, pool);
+        }else if (rbdpool != null) {
```
GutoVeronezi, line 52:
```suggestion
        } else if (rbdpool != null) {
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -35,121 +38,168 @@
...
+                s_logger.debug(String.format("The command [%s], to the pool [%s], had the result [%s].", cmd.toString(), uuid, result));
+            } else if (rbdStoragePool != null) {
+                ProcessBuilder processBuilder = createRbdHeartBeatCommand(rbdStoragePool, hostPrivateIp, true);
+                try {
+                    process = processBuilder.start();
+                    BufferedReader bfr = new BufferedReader(new InputStreamReader(process.getInputStream()));
+                    result = bfr.readLine();
+                } catch (IOException e) {
+                    e.printStackTrace();
                 }
```
GutoVeronezi, line 171: Same here about using `com.cloud.utils.script.Script`.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAMonitor.java
```diff
@@ -169,6 +219,27 @@ private Script createHeartBeatCommand(NfsStoragePool primaryStoragePool, String
...
+        processBuilder.command().add(primaryStoragePool._poolAuthUserName);
+        processBuilder.command().add("-s");
+        processBuilder.command().add(primaryStoragePool._poolAuthSecret);
+
+        if (hostValidation) {
+            processBuilder.command().add("-v");
+            processBuilder.command().add(hostPrivateIp);
+        }
+
+        return processBuilder;
```
GutoVeronezi, line 240: Same here about using `com.cloud.utils.script.Script`.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/KVMHAVMActivityChecker.java
```diff
@@ -18,48 +18,91 @@
...
+            command = processBuilder.command().toString();
+            Process process = null;
+
+            try {
+                process = processBuilder.start();
+                BufferedReader bfr = new BufferedReader(new InputStreamReader(process.getInputStream()));
+                parsedLine = bfr.readLine();
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
```
GutoVeronezi, line 99: Same here about using `com.cloud.utils.script.Script`.

plugins/hypervisors/kvm/src/main/java/org/apache/cloudstack/kvm/ha/KVMHAProvider.java
```diff
@@ -88,8 +99,13 @@ public boolean recover(Host r) throws HARecoveryException {
     public boolean fence(Host r) throws HAFenceException {
         try {
             if (outOfBandManagementService.isOutOfBandManagementEnabled(r)){
-                final OutOfBandManagementResponse resp = outOfBandManagementService.executePowerOperation(r, PowerOperation.OFF, null);
-                return resp.getSuccess();
+                final OutOfBandManagement oobm = outOfBandManagementDao.findByHost(r.getId());
+                if (oobm.getPowerState() == PowerState.Unknown){
```
GutoVeronezi, line 103:
```suggestion
                if (oobm.getPowerState() == PowerState.Unknown) {
```

GutoVeronezi (commented):
@dhslove, ACS already uses java bindings for librados in `com.ceph.rados.Rados`; could not the heartbeat be implemented through Java instead of creating Python scripts? cc: @wido

shwstppr:
@dhslove  can you please address the merge conflicts? Should we consider this for 4.19.0.0?

weizhouapache:
moved to 4.21.0.0 milestone

harikrishna-patnala:
@dhslove can you please advice if this PR is still valid and can we target this to 4.22 ?

DaanHoogland:
> @dhslove can you please advice if this PR is still valid and can we target this to 4.22 ?

@wido can you make this ^ call? (/me volunteering to do the rebasing)

dhslove:
@harikrishna-patnala Sorry for the late reply. I think it would be better to close this PR and implement this feature in a new PR rather than maintain it. I currently have the tested code for this feature, so I think I can contribute it to the next CloudStack version. Please let me know your thoughts.

DaanHoogland:
> @harikrishna-patnala Sorry for the late reply. I think it would be better to close this PR and implement this feature in a new PR rather than maintain it. I currently have the tested code for this feature, so I think I can contribute it to the next CloudStack version. Please let me know your thoughts.

thanks @dhslove , that would be great, go ahead.

DaanHoogland:
tbc

weizhouapache:
> tbc

maybe remove the milestone and project too ?

calvix:
Hello, any plans to continue working on this?

weizhouapache:
> Hello, any plans to continue working on this?

@calvix 
I will test it for ACS 4.24.0

PR #5902, opened 2022-01-26
Allow users to view reserved System VM IPs, if they're already allocated to user

SadiJr:
### Description

The default behavior of ACS is to try to use all available public IPs, including those reserved for system VMs; by default when a public IP is reserved to System VM, they are used as preferred to be allocated to system VMs, but not restricted to these type of VMs. This behavior can be changed by changing the value of the global setting `system.vm.public.ip.reservation.mode.strictness` from false to true. However, if an IP reserved for system VMs is already in use for a user VR, the parameter change will make the IP to be hidden from the user, which can lead to all sorts of confusion. Therefore, this PR is intended to address these situations.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

It was tested in a local lab:
1. I allocated a range of 10 IPs to System VMs with `system.vm.public.ip.reservation.mode.strictness` as false;
2. I allocated all the other IPs to another network;
3. I created a new network and started a new VM with this network;
4. Then I changed the `system.vm.public.ip.reservation.mode.strictness` to true;
5. Before the changes, in the VR details (UI), there is no IP information;
6. With this PR, the user can see correctly the IP.

Also, I added new unit tests.


server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -304,7 +304,7 @@
 
     static Boolean rulesContinueOnErrFlag = true;
 
-    private static final ConfigKey<Boolean> SystemVmPublicIpReservationModeStrictness = new ConfigKey<Boolean>("Advanced",
+    public static ConfigKey<Boolean> SystemVmPublicIpReservationModeStrictness = new ConfigKey<Boolean>("Advanced",
```
DaanHoogland, line 307: shouldn't this remain final? we don't want a global setting to get another config key in runtime. This has become more important as the var is now public.

SadiJr: @DaanHoogland thanks for the suggestion.

I made these changes so that I could perform the unit tests. Unfortunately it was the only way I could test different configuration values. If you have any suggestions on this, I would be very grateful.

DaanHoogland: I don't undestand why you would have to do that. The `IpAddressManager` you added for your test is already a mock and you can let it return anything you want. if need be you can add a `IsSVMReservationStrict()` method for convenience.

SadiJr: Done, thanks

RodrigoDLopez (approved):
LGTM

nvazquez (approved):

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan test

GutoVeronezi (approved):
LGTM

nvazquez:
@blueorangutan test

server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
DaanHoogland: > Done, thanks

I still see it as public, not private static @SadiJr .

server/src/main/java/com/cloud/server/ManagementServerImpl.java
```diff
@@ -2424,7 +2425,9 @@ private void setParameters(SearchCriteria<IPAddressVO> sc, final ListPublicIpAdd
             sc.setParameters("state", state);
         }
 
-        sc.setParameters( "forsystemvms", false);
+        if (sourceNetworkId == null && IpAddressManagerImpl.SystemVmPublicIpReservationModeStrictness.value()) {
+            sc.setParameters("forsystemvms", false);
+        }
```
weizhouapache, line 2430: @SadiJr 
I have two questions
(1) if the global setting is `true`,  can the IP (in the reserved ip range  but used by uservm) listed by api ?
(2) if the global setting is `false`, can domain admin and normal users see the public ips used by systemvms ?

SadiJr: @weizhouapache Sorry for the delay in responding:

1 - If the configuration is 'true', the user will not be able to use the reserved IP range. This scenario will only be true if the setting is 'false', the user creates a VM that will use one of the reserved IPs, and then the setting is again set to 'true'. In this scenario, the reserved IPs for the system VMs will not be listed by the API.

2 - No, you will only see the IPs from that range that are available, and those that are used by the VMs themselves, but the IPs used by System VMs are not listed.

weizhouapache: > 1 - If the configuration is 'true', the user will not be able to use the reserved IP range. This scenario will only be true if the setting is 'false', the user creates a VM that will use one of the reserved IPs, and then the setting is again set to 'true'. In this scenario, the reserved IPs for the system VMs will not be listed by the API.
> 
> 2 - No, you will only see the IPs from that range that are available, and those that are used by the VMs themselves, but the IPs used by System VMs are not listed.

@SadiJr 
thanks for reply.
2 looks good to me.
however, for 1, as you said , "In this scenario, the reserved IPs for the system VMs will not be listed by the API." so this issue is not really fixed, right ?
would it be better to use
`IpAddressManagerImpl.SystemVmPublicIpReservationModeStrictness.value() || isAllocated` ?

weizhouapache: @SadiJr 
no problem. thanks for your reply.
The 4.17.0.0 code freeze date is very close. If you want this PR to be included, please update.

server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
SadiJr: Done, thanks for the suggestion and sorry for the delay in answer.

server/src/main/java/com/cloud/server/ManagementServerImpl.java
SadiJr: Done, thanks for the suggestion and sorry for the delay in answer.

weizhouapache:
@blueorangutan package

DaanHoogland (approved):

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

server/src/main/java/com/cloud/server/ManagementServerImpl.java
weizhouapache: @SadiJr 
thanks for update. I will review this PR

weizhouapache:
@SadiJr 
I have tested this PR. unforunately the issue you mentioned in description is not fixed.

I have made some changes to your code. see https://github.com/weizhouapache/cloudstack/commit/18137e105cefbc163a1bf8cad4961c30239762dd
what I have tested are

1. global setting `system.vm.public.ip.reservation.mode.strictness` is set to `false`

- 1.1 Free IPs in the range can be listed (OK to me)
- 1.2 Free IPs in the range can be associated to isolated networks (OK to me)
- 1.3 Allocated IPs (in step 1.2) are listed in the "Public IP Addresses" tab in the network details page. (OK to me)

2. global setting `system.vm.public.ip.reservation.mode.strictness` is set to `true`

- 2.1 Free IPs in the range can NOT be listed (OK to me)
- 2.2 Free IPs in the range can NOT be associated to isolated networks by `cmk` (OK to me)
- 2.3 Allocated IPs (in step 1.2) are listed in the "Public IP Addresses" tab in the network details page. (OK to me)

other than that, I change the global configuration to dynamic, so it is not required to restart management server when you change the value.

Could you please test it ? If it works, please advise if you add the commit to your PR or I create a new PR which including all changes.

SadiJr:
@weizhouapache thanks for yours tests and your help. I will implement your suggestions and test them. If all goes well, I'll add a new commit with your changes.

SadiJr:
@weizhouapache 

I did some more tests, and everything is working good. I also checked that, with the setting as `true`, when releasing a reserved IP from the system VMs (but in use by a user VM), it is correctly excluded from the listing and it is also not possible for the user to reallocate it via cmk . Also, I created a new commit with your changes. 

Again, thank you very much for your testing and your help.

weizhouapache:
@blueorangutan test

weizhouapache:
@SadiJr 
the travis smoke test `test_portable_publicip` failed, it seems to be related to this PR.

```
Currently running test: smoke/test_portable_publicip
==== Marvin Init Started ====
=== Marvin Parse Config Successful ===
=== Marvin Setting TestData Successful===
==== Log Folder Path: /tmp/MarvinLogs/Apr_18_2022_23_17_26_IICLVA All logs will be available here ====
=== Marvin Init Logging Successful===
==== Marvin Init Successful ====
=== TestName: test_createPortablePublicIPAcquire | Status : EXCEPTION ===
=== TestName: test_createPortablePublicIPRange | Status : EXCEPTION ===
=== Final results are now copied to: /tmp/MarvinLogs/test_portable_publicip_UW5HUH ===
real	0m9.150s
user	0m0.841s
sys	0m0.068s
```

weizhouapache (approved):

PR #5905, opened 2022-01-27
Fix internet dependent test from 'UriUtilsParametrizedTest' class

paulazomig:
### Description

This PR address the `UriUtilsParametrizedTest` need for Internet connection. The unit tests of the class `UriUtilsParametrizedTest` will fail when they are executed without a connection to the Internet. That happens because of the method `getByName` from `InetAdress` class (that, given the host's name, determines its IP address). Therefore, if one executes the build in a network that does not allow Internet access, or if some DNS configuration is not resolving the configure DNS name in the tests, the tests fail. The goal of the test in that class is not to test DNS, but the validation of the accepted formats. Therefore, we should not depend on DNS, Internet connection, and others.

This PR proposes to solve that problem by mocking the object for the `InetAdress` class. That way, when the `getByName` method is called, the mocked object is returned and the tests can proceed normally without needing a healthy and working DNS or Internet connection to use some Internet DNS system. 

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity
- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
I ran the unit tests from the `UriUtilsParametrizedTest` class, before this PR, and without an Internet connection, the tests failed. Afterwards, I ran the tests again (with this PR), without an Internet connection, and they all passed as expected.

### DIFF
--- a/utils/src/test/java/com/cloud/utils/UriUtilsParametrizedTest.java
+++ b/utils/src/test/java/com/cloud/utils/UriUtilsParametrizedTest.java
@@ -19,6 +19,7 @@
 
 package com.cloud.utils;
 
+import java.net.InetAddress;
 import java.util.Collection;
 import java.util.LinkedList;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


DaanHoogland (approved):
code looks good. tnx @paulazomig

joseflauzino:
@blueorangutan package

DaanHoogland:
As this is only test code and travis passed, I don't think we need further testing

GutoVeronezi:
Based in the tests results, @DaanHoogland's [comment](https://github.com/apache/cloudstack/pull/5905#issuecomment-1024869527) and 2 approvals, I'll merge this one.

Issue #5934, opened 2022-02-04
vmware: interface for additional public IP in VR is not unplugged when public IP is released


weizhouapache:
step to reproduce the issue

(1) create isolated network
(2) add new public ip range
(3) acquire public ip in new ip range
(4) create port forwarding rule. a new NIC is added to the network VR
(5) releae the public ip.

expected result (this is actual result in kvm/xenserver env):
new NIC is removed in network VR

actual result (in vmware env):
IP is unassigned, but NIC is not removed

related code:

(vmware)

https://github.com/apache/cloudstack/blob/194bf6ead96a8fbc13028081974a56a821b131ab/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java#L1102-L1105

(kvm)

https://github.com/apache/cloudstack/blob/194bf6ead96a8fbc13028081974a56a821b131ab/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java#L490-L496

(xenserver)

https://github.com/apache/cloudstack/blob/194bf6ead96a8fbc13028081974a56a821b131ab/plugins/hypervisors/xenserver/src/main/java/com/cloud/hypervisor/xenserver/resource/CitrixResourceBase.java#L590-L596

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR
~~~

##### CLOUDSTACK VERSION

~~~
4.16/main
~~~

##### STEPS TO REPRODUCE

~~~

~~~

##### EXPECTED RESULTS

~~~

~~~

##### ACTUAL RESULTS

~~~

~~~

PR #5985, opened 2022-02-11
[VMware] Support for removal of NIC on IP disassociation on the VR

Pearl1594:
### Description

This PR fixes https://github.com/apache/cloudstack/issues/5934

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

(1) create isolated network
(2) add new public IP range
(3) acquire public IP in new IP range
(4) create firewall rule. a new NIC is added to the network VR
(5) release the public IP. - if it is the only IP on the NIC, the nic will be removed

Test 1:
Initial output of `ip a` with default NIC:
image

Acquired an IP in another Public IP range and added a firewall rule: New nic was added
image

On deletion of the firewall rule on the NIC, the nic is removed:
image

Test2:

Add another Public IP range in the same VLAN as the one before 
Acquire & add firewall rule for 2 IPs belonging to 2 different Public IP ranges but the same VLAN, such that there are 2 IPs on a NIC (check eth3)
image

Remove the rule from one of the IPs, and verify that the IP is disassociated from the VRs NIC, but the NIC isn't removed:
image

### DIFF
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -1597,7 +1597,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
             String routerName = cmd.getAccessDetail(NetworkElementCommand.ROUTER_NAME);
 
             VirtualMachineMO vmMo = hyperHost.findVmOnHyperHost(routerName);
-            // command may sometimes be redirect to a wrong host, we relax
+            // command may sometimes be redirected to a wrong host, we relax
             // the check and will try to find it within datacenter
             if (vmMo == null) {
                 if (hyperHost instanceof HostMO) {
@@ -1611,14 +1611,17 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                 s_logger.error(msg);
                 throw new Exception(msg);
             }
-
+            final String lastIp = cmd.getAccessDetail(NetworkElementCommand.NETWORK_PUB_LAST_IP);
             if (ips.length == 1 && !ips[0].isAdd()) {
-                VirtualDevice nic = getVirtualDevice(vmMo, ips[0]);
+                Pair<VirtualDevice, Integer> nicInfo = getVirtualDevice(vmMo, ips[0]);
 
-                if (nic == null) {
+                if (nicInfo.first() == null && lastIp.equalsIgnoreCase("true")) {
+                    if (nicInfo.second() == 2) {
+                        return new ExecutionResult(false, "Unable to remove eth2 in network VR because it is the public NIC of source NAT");
+                    }
                     return new ExecutionResult(false, "Couldn't find NIC");
                 }
-                configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.REMOVE, "unplugNicCommand");
+                configureNicDevice(vmMo, nicInfo.first(), VirtualDeviceConfigSpecOperation.REMOVE, "unplugNicCommand");
             }
         } catch (Throwable e) {
             s_logger.error("Unexpected exception: " + e.toString() + " will shortcut rest of IPAssoc commands", e);
@@ -1627,7 +1630,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
         return new ExecutionResult(true, null);
     }
 
-    private VirtualDevice getVirtualDevice(VirtualMachineMO vmMo, IpAddressTO ip) throws Exception {
+    private Pair<VirtualDevice, Integer> getVirtualDevice(VirtualMachineMO vmMo, IpAddressTO ip) throws Exception {
         NicTO nicTO = ip.getNicTO();
         URI broadcastUri = BroadcastDomainType.fromString(ip.getBroadcastUri());
         if (BroadcastDomainType.getSchemeValue(broadcastUri) != BroadcastDomainType.Vlan) {
@@ -1638,11 +1641,16 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
         String publicNetworkName = HypervisorHostHelper.getPublicNetworkNamePrefix(vlanId);
         Pair<Integer, VirtualDevice> publicNicInfo = vmMo.getNicDeviceIndex(publicNetworkName);
 
+        if (publicNicInfo.first() == 2) {
+            s_logger.debug("Do not remove eth2 in network VR because it is the public NIC of source NAT.");
+            return new Pair<>(null, publicNicInfo.first());
+        }
+
         if (s_logger.isDebugEnabled()) {
             s_logger.debug(String.format("Find public NIC index, public network name: %s , index: %s", publicNetworkName, publicNicInfo.first()));
         }
 
-        return findVirtualNicDevice(vmMo, nicTO.getMac());
+        return new Pair<>(findVirtualNicDevice(vmMo, nicTO.getMac()), publicNicInfo.first());
     }
 
     private void configureNicDevice(VirtualMachineMO vmMo, VirtualDevice nic, VirtualDeviceConfigSpecOperation operation, String commandName) throws Exception {
--- a/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
+++ b/plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
@@ -1329,7 +1329,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                     nicTo.getMac(), deviceNumber + 1, true, true);
         }
 
-        configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.ADD);
+        configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.ADD, "PlugNicCommand");
     }
 
     private ReplugNicAnswer execute(ReplugNicCommand cmd) {
@@ -1390,7 +1390,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                 VmwareHelper.updateNicDevice(nic, networkInfo.first(), networkInfo.second());
             }
 
-            configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.EDIT);
+            configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.EDIT, "ReplugNicCommand");
 
             return new ReplugNicAnswer(cmd, true, "success");
         } catch (Exception e) {
@@ -1431,7 +1431,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
             if (nic == null) {
                 return new UnPlugNicAnswer(cmd, true, "success");
             }
-            configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.REMOVE);
+            configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.REMOVE, "unplugNicCommand");
 
             return new UnPlugNicAnswer(cmd, true, "success");
         } catch (Exception e) {
@@ -1478,7 +1478,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
                 device.setBacking(dataCenterMo.getDvPortBackingInfo(networkInfo));
             }
 
-            configureNicDevice(vmMo, device, VirtualDeviceConfigSpecOperation.EDIT);
+            configureNicDevice(vmMo, device, VirtualDeviceConfigSpecOperation.EDIT, "plugPublicNic");
         } catch (Exception e) {
 
             // restore allocation mask in case of exceptions
@@ -1598,7 +1598,7 @@ public class VmwareResource implements StoragePoolResource, ServerResource, Vmwa
 
             VirtualMachineMO vmMo = hyperHost.findVmOnHyperHost(routerName);
             // command may sometimes be redirect to a wrong host, we relax
-            // the check and will try to find it within cluster
+            // the check and will try to find it within datacenter
             if (vmMo == null) {
                 if (hyperHost instanceof HostMO) {
                     final DatacenterMO dcMo = new DatacenterMO(context, hyperHost.getHyperHostDatacenter());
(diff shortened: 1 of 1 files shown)


Pearl1594:
@blueorangutan package

Pearl1594:

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1625,69 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+            VmwareHypervisorHost hyperHost = getHyperHost(context);
+            IpAddressTO[] ips = cmd.getIpAddresses();
+            String routerName = cmd.getAccessDetail(NetworkElementCommand.ROUTER_NAME);
+
+            VirtualMachineMO vmMo = hyperHost.findVmOnHyperHost(routerName);
+            // command may sometimes be redirect to a wrong host, we relax
+            // the check and will try to find it within cluster
+            if (vmMo == null) {
+                if (hyperHost instanceof HostMO) {
+                    ClusterMO clusterMo = new ClusterMO(hyperHost.getContext(), ((HostMO) hyperHost).getParentMor());
```
nvazquez, line 1640: What about datacenter instead of cluster search?

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1625,69 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+                }
+            }
+
+            if (vmMo == null) {
+                String msg = String.format("Router %s no longer exists to execute IPAssoc command ", routerName);
+                s_logger.error(msg);
+                throw new Exception(msg);
+            }
+
+            for (IpAddressTO ip : ips) {
```
nvazquez, line 1651: I think this for loop is not needed since we ensure the logic is executed only when it has one item as is not add. Could be refactored to pre-checks and logic to reduce indentation

weizhouapache:
code works well. @Pearl1594

Pearl1594:
@blueorangutan package

Pearl1594:
@nvazquez I've addressed the comments. @weizhouapache I've rebased the branch off 4.16.

Pearl1594:

weizhouapache:
@Pearl1594 @nvazquez @sureshanaparti 
this PR needs to be tested/approved. 
However 4.16.1.0 is close to be released, let's test/approve/merge this after 4.16.1.0 release, is that ok to you ?

Pearl1594:
Sure @weizhouapache I am fine with moving to 4.17, I rebased it to 4.16, per a request made earlier : ) 
@DaanHoogland I am not sure if there is a hypervisor agnostic way to solve this, as each hypervisor has it's own way of handling hot unplugging of NICs.

weizhouapache:
> agree @weizhouapache in addition I wonder why this apparently generic feature requires such a vmware specific implementation. Isn't a more hypevisor independent implementation required? I understand that it already works for Xen and KVM, so is there a way to fix it so it works for all, instead of adding ahypervisor specific code?

@DaanHoogland 
I do not know if there is way to do it.
The command is sent to host or vcenter and the behaviors are different for hypervisors.
Please note, this is not a feature. it is a part of IpAssocCommand processing.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,72 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+                    final DatacenterMO dcMo = new DatacenterMO(context, hyperHost.getHyperHostDatacenter());
+                    vmMo = dcMo.findVm(routerName);
+                }
+            }
+
+            if (vmMo == null) {
+                String msg = String.format("Router %s no longer exists to execute IPAssoc command ", routerName);
+                s_logger.error(msg);
+                throw new Exception(msg);
+            }
```
DaanHoogland, line 1613: can you factor out this bit?

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,72 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+                    s_logger.debug(String.format("Find public NIC index, public network name: %s , index: %s", publicNetworkName, publicNicInfo.first()));
+                }
+
+                VirtualDevice nic = findVirtualNicDevice(vmMo, nicTO.getMac());
+
+                if (nic == null) {
+                    return new ExecutionResult(false, "Couldn't find NIC");
+                }
+                configureNicDevice(vmMo, nic, VirtualDeviceConfigSpecOperation.REMOVE);
+            }
```
DaanHoogland, line 1637: can you factor out this bit? (maybe if even smaller/nested chunks)

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,72 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+    private void configureNicDevice(VirtualMachineMO vmMo, VirtualDevice nic, VirtualDeviceConfigSpecOperation operation) throws Exception {
+        VirtualMachineConfigSpec vmConfigSpec = new VirtualMachineConfigSpec();
+        VirtualDeviceConfigSpec deviceConfigSpec = new VirtualDeviceConfigSpec();
+        deviceConfigSpec.setDevice(nic);
+        deviceConfigSpec.setOperation(operation);
+
+
+        vmConfigSpec.getDeviceChange().add(deviceConfigSpec);
+        if (!vmMo.configureVm(vmConfigSpec)) {
+            throw new Exception("Failed to configure devices when running unplugNicCommand");
```
DaanHoogland, line 1654: this is not always true I think,
```suggestion
            throw new Exception("Failed to configure devices when running a (un/re)PlugNicCommand");
```
maybe even pass it as parameter?
good extraction otherwise.

DaanHoogland:
ok, got it @weizhouapache @Pearl1594 . It doesn't feel ideal, but let's move on.

weizhouapache:
> ok, got it @weizhouapache @Pearl1594 . It doesn't feel ideal, but let's move on.

@DaanHoogland 
there are many (hundreds of) commands which are processed in hypervisor plugins.
if you have an idea how to improve the processes and remove the hypervisor specific codes, it would be great.

DaanHoogland:
> > ok, got it @weizhouapache @Pearl1594 . It doesn't feel ideal, but let's move on.
> 
> @DaanHoogland there are many (hundreds of) commands which are processed in hypervisor plugins. if you have an idea how to improve the processes and remove the hypervisor specific codes, it would be great.

always on the lookout @weizhouapache ;)

weizhouapache:
@Pearl1594 
in my testing, if public ips are used for port forwarding or load balancer, everything seems ok.
If public ips are used for static nat, the interface will be removed even if there is still public ip in use (for example, pf or lb or nat). This issue happens with eth2 as well which is the interface for source nat.

I think we should use similar code like
```
            final String lastIp = cmd.getAccessDetail(NetworkElementCommand.NETWORK_PUB_LAST_IP);
            for (final IpAddressTO ip : ips) {
                if (!StringUtils.equalsIgnoreCase(lastIp, "true") || ip.isAdd()) {
                    continue;
                }
```
and
```
        if (publicNicInfo.first() == 2) {
            s_logger.debug("Do not remove eth2 in network VR because it is the public NIC of source NAT.");
            continue;
        }
```

Pearl1594:
Thanks for the review/testing @weizhouapache . I'll fix the issue.

Pearl1594:
@weizhouapache Could you please help verify if the fix addresses your comment. Thanks.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,83 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+                s_logger.error(msg);
+                throw new Exception(msg);
+            }
+            final String lastIp = cmd.getAccessDetail(NetworkElementCommand.NETWORK_PUB_LAST_IP);
+            if (ips.length == 1 && !ips[0].isAdd()) {
+                Pair<VirtualDevice, Integer> nicInfo = getVirtualDevice(vmMo, ips[0]);
+
+                if (nicInfo.first() == null && lastIp.equalsIgnoreCase("true")) {
+                    if (nicInfo.second() == 2) {
+                        return new ExecutionResult(false, "Unable to remove eth2 in network VR because it is the public NIC of source NAT");
```
weizhouapache, line 1620: @Pearl1594 
for eth2, we can return a sucess result.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,83 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+                }
+            }
+
+            if (vmMo == null) {
+                String msg = String.format("Router %s no longer exists to execute IPAssoc command ", routerName);
+                s_logger.error(msg);
+                throw new Exception(msg);
+            }
+            final String lastIp = cmd.getAccessDetail(NetworkElementCommand.NETWORK_PUB_LAST_IP);
+            if (ips.length == 1 && !ips[0].isAdd()) {
```
weizhouapache, line 1615: @Pearl1594 
sometimes there are multiple public Ip addresses in the IpAssocCommand
maybe it is better to loop all ip addresses, like kvm/xenserver plugins do.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,83 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+        URI broadcastUri = BroadcastDomainType.fromString(ip.getBroadcastUri());
+        if (BroadcastDomainType.getSchemeValue(broadcastUri) != BroadcastDomainType.Vlan) {
+            throw new InternalErrorException(String.format("Unable to assign a public IP to a VIF on network %s", ip.getBroadcastUri()));
+        }
+        String vlanId = BroadcastDomainType.getValue(broadcastUri);
+
+        String publicNetworkName = HypervisorHostHelper.getPublicNetworkNamePrefix(vlanId);
+        Pair<Integer, VirtualDevice> publicNicInfo = vmMo.getNicDeviceIndex(publicNetworkName);
+
+        if (publicNicInfo.first() == 2) {
```
weizhouapache, line 1644: this is not needed ,because it is already handled in line 1619/1620

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
Pearl1594: @weizhouapache we only unplug the nic if it is there is only 1 IP on the NIC / last IP right? In that case is it required to loop through all the IPs? sorry, I probably am missing something..

weizhouapache: @Pearl1594
from what I have tested, it is good to use `if (ips.length == 1 && !ips[0].isAdd()) {`
I do not know if there are special case that, there are multiple ip address in IpAssocCommand and we need to remove one of them.

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,83 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+            if (ips.length == 1 && !ips[0].isAdd()) {
+                Pair<VirtualDevice, Integer> nicInfo = getVirtualDevice(vmMo, ips[0]);
+
+                if (nicInfo.first() == null && lastIp.equalsIgnoreCase("true")) {
+                    if (nicInfo.second() == 2) {
+                        return new ExecutionResult(false, "Unable to remove eth2 in network VR because it is the public NIC of source NAT");
+                    }
+                    return new ExecutionResult(false, "Couldn't find NIC");
+                }
+                configureNicDevice(vmMo, nicInfo.first(), VirtualDeviceConfigSpecOperation.REMOVE, "unplugNicCommand");
```
weizhouapache, line 1624: @Pearl1594 
nic can be unplugged only if `lastIp.equalsIgnoreCase("true")`

https://github.com/apache/cloudstack/blob/593f1b698836749856f38f706b30b23a660d26a3/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java#L2129-L2136

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,83 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+            }
+            final String lastIp = cmd.getAccessDetail(NetworkElementCommand.NETWORK_PUB_LAST_IP);
+            if (ips.length == 1 && !ips[0].isAdd()) {
+                Pair<VirtualDevice, Integer> nicInfo = getVirtualDevice(vmMo, ips[0]);
+
+                if (nicInfo.first() == null && lastIp.equalsIgnoreCase("true")) {
+                    if (nicInfo.second() == 2) {
+                        return new ExecutionResult(false, "Unable to remove eth2 in network VR because it is the public NIC of source NAT");
+                    }
+                    return new ExecutionResult(false, "Couldn't find NIC");
```
weizhouapache, line 1622: return a success if no need to unplug the nic ?

weizhouapache:
@Pearl1594 
the following code works in my tesing
```
            final String lastIp = cmd.getAccessDetail(NetworkElementCommand.NETWORK_PUB_LAST_IP);
            if (ips.length == 1 && !ips[0].isAdd()) {
                Pair<VirtualDevice, Integer> nicInfo = getVirtualDevice(vmMo, ips[0]);

                if (nicInfo.first() != null && nicInfo.second() != 2 && lastIp.equalsIgnoreCase("true")) {
                    configureNicDevice(vmMo, nicInfo.first(), VirtualDeviceConfigSpecOperation.REMOVE, "unplugNicCommand");
                }
            }
```

plugins/hypervisors/vmware/src/main/java/com/cloud/hypervisor/vmware/resource/VmwareResource.java
```diff
@@ -1622,6 +1589,81 @@ private ExecutionResult prepareNetworkElementCommand(IpAssocCommand cmd) {
...
+                String msg = String.format("Router %s no longer exists to execute IPAssoc command ", routerName);
+                s_logger.error(msg);
+                throw new Exception(msg);
+            }
+            final String lastIp = cmd.getAccessDetail(NetworkElementCommand.NETWORK_PUB_LAST_IP);
+            for (IpAddressTO ip : ips) {
+                if (ip.isAdd() || lastIp.equalsIgnoreCase("false")) {
+                    continue;
+                }
+                Pair<VirtualDevice, Integer> nicInfo = getVirtualDevice(vmMo, ips[0]);
```
weizhouapache, line 1619: @Pearl1594 
use ip instead of `ips[0]` in line 1619 ?

weizhouapache (approved):
did lots of testing, all seem fine.

weizhouapache:
@blueorangutan package

nvazquez:

nvazquez (approved):
LGTM

Issue #5913, opened 2022-01-31
Support setting boot parameter via template settings


Pearl1594:
##### ISSUE TYPE

 * Enhancement Request

##### COMPONENT NAME

~~~
template, VM
~~~

##### CLOUDSTACK VERSION

~~~
ACS 4.15.2
~~~

##### SUMMARY

When a template is configured with UEFI setting - UEFI=secure and a VM is deployed with the template, the boot parameters passed via the template settings does not take effect on the first boot. However, powering off the VM and restarting it, will bring up the instance using UEFI i.e., the template settings take effect

##### STEPS TO REPRODUCE

~~~
1. Register a template as deploy-as-is and specify UEFI=SECURE in the template settings
2. Deploy a VM using the above registered template - VM doesn't boot using uefi
3. Stop & start the VM - the boot parameters are considered

~~~

##### EXPECTED RESULTS

~~~
Consider template settings on the initial boot, or if not possible, prompt the user (by means of logs) that the specific setting wasn't considered
~~~

##### ACTUAL RESULTS

~~~

~~~

DaanHoogland:
@Pearl1594 this warrants a discussion on the functionality; if it is a deploy-as-is template, the contents of the template should be honoured. I.E. it should not be altered during initial instantiation. Altering it after that no longer has anything to do with the template and is ok, afaiac.
We should either forbid altering parameters on a deploy as is template or make very explicit what is going on (and maybe apply it immediately.
what you encountered is a strange situation but is as designed, i think.

nvazquez:
I agree with @DaanHoogland, however I think it is a bug as per design boot parameters should not be considered on deploy-as-is VMs IMO. @borisstoyanov @rohityadavcloud @andrijapanicsb what's your input on this?

andrijapanicsb:
with deployasis, everything should be honoured from the template, NOT from the template settings

kricud:
IMHO- template settings should be honored.
How I will explain to end users/customers that this template is for BIOS and this one is for UEFI and if you do not do correct selection, it will not boot.
I think it’s a bug that deployment wizard is not obeying settings from template.
If its not "Read VM settings from OVA" with this all settings should come from OVA.

Create #5932 same problem different description.

kricud:
@andrijapanicsb @DaanHoogland @nvazquez
[...]

PR #5990, opened 2022-02-14
Template boot type

davidjumani:
### Description

Fixes https://github.com/apache/cloudstack/issues/5913
Boots the VM based on the template UEFI setting

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

Check the smoke tests!


davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2211,6 +2214,31 @@ else if (details != null && !details.isEmpty()) {
...
+        if (details == null || details.isEmpty()) {
+            return;
+        }
+        String bootMode = details.get(ApiConstants.BootType.UEFI.toString());
+        if (bootMode == null) {
+            return;
+        }
+        if (template.isDeployAsIs()) {
+            String msg = "Deploy-as-is templates can not have UEFI parameter set. Settings are read directly from the template";
+            throw new InvalidParameterValueException(msg);
```
nvazquez, line 2227: I think we shouldn't fail at this point, but ignore the settings (return at this point in the method)

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4336,6 +4336,18 @@ public UserVmVO doInTransaction(TransactionStatus status) throws InsufficientCap
                 vm.setUserVmType(type);
                 _vmDao.persist(vm);
                 for (String key : customParameters.keySet()) {
+                    // BIOS was explicitly passed as the boot type, so honour it
+                    if (key.equalsIgnoreCase(ApiConstants.BootType.BIOS.toString())) {
```
nvazquez, line 4340: What about the case when BIOS is set and the template is deploy as is?

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
davidjumani: This might lead the user to wonder why the setting was not applied

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
davidjumani: Will respect the template (deploy as is) settings

davidjumani:
@blueorangutan test

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
nvazquez: Yes, sorry, was thinking about the deployment but this is updating the settings +1

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2211,6 +2214,31 @@ else if (details != null && !details.isEmpty()) {
...
+    void validateDetails(VMTemplateVO template, Map<String, String> details) {
+        if (details == null || details.isEmpty()) {
+            return;
+        }
+        String bootMode = details.get(ApiConstants.BootType.UEFI.toString());
+        if (bootMode == null) {
+            return;
+        }
+        if (template.isDeployAsIs()) {
+            String msg = "Deploy-as-is templates can not have UEFI parameter set. Settings are read directly from the template";
```
GutoVeronezi, line 2226: We can add some context info here, like the template id, name and so on.

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
```diff
@@ -2211,6 +2214,31 @@ else if (details != null && !details.isEmpty()) {
...
+        }
+        try {
+            String mode = bootMode.trim().toUpperCase();
+            ApiConstants.BootMode.valueOf(mode);
+            details.put(ApiConstants.BootType.UEFI.toString(), mode);
+            return;
+        } catch (IllegalArgumentException e) {
+            String msg = String.format("Invalid %s: %s specified. Valid values are: %s",
+                    ApiConstants.BOOT_MODE, bootMode, Arrays.toString(ApiConstants.BootMode.values()));
+            s_logger.error(msg);
```
GutoVeronezi, line 2237: We could pass the exception as parameter of  `s_logger.error`.

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan package

davidjumani:
@blueorangutan test

server/src/main/java/com/cloud/template/TemplateManagerImpl.java
davidjumani: The exception is not thrown but logged instead. So explicitly throwing it in the following line

nvazquez (approved):
LGTM - test failures are not related to this PR

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4336,6 +4336,18 @@ public UserVmVO doInTransaction(TransactionStatus status) throws InsufficientCap
                 vm.setUserVmType(type);
                 _vmDao.persist(vm);
                 for (String key : customParameters.keySet()) {
+                    // BIOS was explicitly passed as the boot type, so honour it
```
harikrishna-patnala, line 4339: what if template details has UEFI -> LEGACY and deploy VM passes UEFI -> SECURE ? As per my understanding, we are considering template details, since we are not deleting from vm.details if boot type is UEFI. (vn.details is loaded with template details before this).

If precedence goes to template details this is fine, but it might be confusing to the user if he/she provides SECURE mode during VM deploy but it is not honoured since template has other detail.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4344,11 +4356,6 @@ public UserVmVO doInTransaction(TransactionStatus status) throws InsufficientCap
                     } else {
                         vm.setDetail(key, customParameters.get(key));
                     }
-
-                    if (key.equalsIgnoreCase(ApiConstants.BootType.UEFI.toString())) {
```
harikrishna-patnala, line 4348: I think we need to add this too, if template details do not have any data, we need to set the deploy VM param into VM details.

davidjumani: This is done in line 4345 without any explicit need for an if condition

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
davidjumani: The template boot mode will be overwritten by the boot mode passed in the deployvm command in line 4345
Precedence is as follows : deploy-as-is > deployvm > VM details > template details

harikrishna-patnala: got it, thanks David

harikrishna-patnala (approved):

PR #6031, opened 2022-02-22
Update VM priority (cpu_chares) when live scaling it

BryanMLima:
### Description

This PR addresses an issue when live scaling a VM with the KVM hypervisor. When the VM has the CPU cores or CPU speed increased, the cpu_shares (priority) of the VM is not updated accordingly. To overcome that, it was needed to manually restart the VM or using the command: `virsh schedinfo --domain <vm_internal_name> --live  cpu_shares=<cpu_shares_value>`. This PR changes this to automatically update the cpu_shares when the CPU cores or speed is increased. It is worth mention that the current behavior caused steal time when VMs would compete for the CPU after a dynamic scale process.

Steps to reproduce:
- Host
    - 6 vCPUs X 2.39Ghz
    - 16GB RAM

- Custom service offering
    - 1 to 6 vCPUs
    - 2.35Ghz

- VM (deploy config)
    - 1 vCPU
    - 1024MB RAM

The cpu_shares of a VM is calculated as follow:
`cpu_shares = number of vCPUs X vCPU frequency in Mhz`

Using the command `virsh schedinfo --domain <internal_vm_name>` the cpu_shares displayed was 2350, which is expected. However, when live scaling the VM to 4 vCPUs, which should return 9400 shares, it was the initial 2350 shares. If the VM is restarted, however, the 9400 shares is used, as it should without the need to restart the VM.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

Using the same configuration and steps I used to describe the bug. Then, when live scaling a VM, it returns the correct cpu_shares without the need to restart it. Therefore, when the VM with the 2350 shares is live scaled to 4 vCPUs, the cpu_shares when using the command mentioned above, returned the correct value of 9400.

Moreover, I created unit tests for the affected methods.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -77,6 +77,9 @@ import org.libvirt.DomainSnapshot;
 import org.libvirt.LibvirtException;
 import org.libvirt.MemoryStatistic;
 import org.libvirt.Network;
+import org.libvirt.SchedParameter;
+import org.libvirt.SchedUlongParameter;
+import org.libvirt.VcpuInfo;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
@@ -186,7 +189,6 @@ import com.cloud.vm.VirtualMachine.PowerState;
 import com.cloud.vm.VmDetailConstants;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.cloudstack.utils.bytescale.ByteScaleUtils;
-import org.libvirt.VcpuInfo;
 
 /**
  * LibvirtComputingResource execute requests on the computing/routing host using
@@ -4620,5 +4622,38 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     public static long countDomainRunningVcpus(Domain dm) throws LibvirtException {
         VcpuInfo vcpus[] = dm.getVcpusInfo();
         return Arrays.stream(vcpus).filter(vcpu -> vcpu.state.equals(VcpuInfo.VcpuState.VIR_VCPU_RUNNING)).count();
+    }
+
+    /**
+     * Retrieves the cpu_shares (priority) of the running VM <br/>
+     * @param dm domain of the VM.
+     * @return the value of cpu_shares of the running VM.
+     * @throws org.libvirt.LibvirtException
+     **/
+    public static Integer getCpuShares(Domain dm) throws LibvirtException {
+        int cpu_shares = 0;
+        for (SchedParameter c : dm.getSchedulerParameters()) {
+            if (c.field.equals("cpu_shares")) {
+                cpu_shares = Integer.parseInt(c.getValueAsString());
+                return cpu_shares;
+            }
+        }
+        s_logger.warn(String.format("Could not get cpu_shares of domain: [%s]. Returning default value of 0. ", dm.getName()));
+        return cpu_shares;
+    }
+
+    /**
+     * Set the cpu_shares (priority) of the running VM <br/>
+     * @param dm domain of the VM.
+     * @param cpuShares new priority of the running VM.
+     * @throws org.libvirt.LibvirtException
+     **/
+    public static void setCpuShares(Domain dm, Integer cpuShares) throws LibvirtException {
+        SchedUlongParameter[] params = new SchedUlongParameter[1];
+        params[0] = new SchedUlongParameter();
+        params[0].field = "cpu_shares";
+        params[0].value = cpuShares;
+
+        dm.setSchedulerParameters(params);
     }
 }
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
@@ -39,8 +39,10 @@ public class LibvirtScaleVmCommandWrapper extends CommandWrapper<ScaleVmCommand,
 
         long newMemory = ByteScaleUtils.bytesToKib(vmSpec.getMaxRam());
         int newVcpus = vmSpec.getCpus();
+        int newCpuSpeed = vmSpec.getMinSpeed() != null ? vmSpec.getMinSpeed() : vmSpec.getSpeed();
+        int newCpuShares = newVcpus * newCpuSpeed;
         String vmDefinition = vmSpec.toString();
-        String scalingDetails = String.format("%s memory to [%s KiB] and CPU cores to [%s]", vmDefinition, newMemory, newVcpus);
+        String scalingDetails = String.format("%s memory to [%s KiB], CPU cores to [%s] and cpu_shares to [%s]", vmDefinition, newMemory, newVcpus, newCpuShares);
 
         try {
             LibvirtUtilitiesHelper libvirtUtilitiesHelper = libvirtComputingResource.getLibvirtUtilitiesHelper();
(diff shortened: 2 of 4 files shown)


GutoVeronezi (approved):
CLGTM, tested manually.

I've pointed just some writing fixes.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -4621,4 +4623,37 @@ public static long countDomainRunningVcpus(Domain dm) throws LibvirtException {
...
+    public static Integer getCpuShares(Domain dm) throws LibvirtException {
+        int cpu_shares = 0;
+        for (SchedParameter c : dm.getSchedulerParameters()) {
+            if (c.field.equals("cpu_shares")) {
+                cpu_shares = Integer.parseInt(c.getValueAsString());
+                return cpu_shares;
+            }
+        }
+        s_logger.warn(String.format("Could not get cpu_shares of domain: [%s]. Returning default value of 0. ", dm.getName()));
+        return cpu_shares;
```
DaanHoogland, line 4642: not vital but as we are not aggregating no need for a var:
```suggestion
        for (SchedParameter c : dm.getSchedulerParameters()) {
            if (c.field.equals("cpu_shares")) {
                return Integer.parseInt(c.getValueAsString());
            }
        }
        s_logger.warn(String.format("Could not get cpu_shares of domain: [%s]. Returning default value of 0. ", dm.getName()));
        return o;
```

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
```diff
@@ -68,6 +71,23 @@ public Answer execute(ScaleVmCommand command, LibvirtComputingResource libvirtCo
         }
     }
 
+    /**
+     * Set the cpu_shares (priority) of the running VM. This is necessary because the priority is only calculated when deploying the VM.
+     * When the number of cores and/or speed of the CPU is changed the cpu_shares is only updated if the VM is restarted or manually, using the command schedinfo.
```
DaanHoogland, line 76: I don't understand this sentence; is it updated manually or on restart? Restart seems clear, but can the update be forced manually or is the restart manually?
Also, the sentence below sugests that this is no longer true. I think this needs a re-formulation. Something like "To prevent that the cpu_shares value is only updated in circumstances A and B, this method ....."

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
```diff
@@ -68,6 +71,23 @@ public Answer execute(ScaleVmCommand command, LibvirtComputingResource libvirtCo
...
+     * To correct this behaviour when live scaling, this command changes the cpu_shares of a running VM.
+     * @param dm domain of the VM.
+     * @param newCpuShares new priority of the running VM.
+     * @throws org.libvirt.LibvirtException
+     **/
+    protected void updateCpuShares(Domain dm, int newCpuShares) throws LibvirtException {
+        int oldCpuShares = LibvirtComputingResource.getCpuShares(dm);
+
+        if (oldCpuShares < newCpuShares) {
```
DaanHoogland, line 85: only up, never down? is this like `nice` for users or can the root user force something?

DaanHoogland (commented):
cltgm but some remarks and questions.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
BryanMLima: It can be manually updated through the command mentioned in the PR description: `virsh schedinfo --domain <vm_internal_name> --live cpu_shares=<cpu_shares_value>` or restarting the VM. But I understand it is a bit confusing the way it is written, I will rewrite it.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
```diff
@@ -73,8 +73,7 @@ public Answer execute(ScaleVmCommand command, LibvirtComputingResource libvirtCo
 
     /**
      * Sets the cpu_shares (priority) of the running VM. This is necessary because the priority is only calculated when deploying the VM.
-     * When the number of cores and/or speed of the CPU is changed the cpu_shares is only updated if the VM is restarted or manually, using the command schedinfo.
-     * To correct this behaviour when live scaling, this command changes the cpu_shares of a running VM.
+     * To prevent the cpu_shares to be manually updated by using the command virsh schedinfo or restarting the VM. This method updates the cpu_shares of a running VM on the fly.
```
DaanHoogland, line 76: :+1:

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
BryanMLima: In ACS, the live scale process does not allow to reduce the resource of a running VM. Therefore, it can only occur upwards and the cpu_shares is calculated as follow:
`cpu_shares = number of vCPUs X vCPU frequency in Mhz`.
This `if` (the conditional) is to check if the vCPUs or CPU speed is higher than the current compute offering, as it makes no sense to decrease the priority of a VM that is having its resources increased. In other words, this `if` is to check if it is a memory scale only, which should not have its priority changed in the host.

DaanHoogland (approved):

sureshanaparti (approved):
code LGTM

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtScaleVmCommandWrapper.java
weizhouapache: it looks the new speed >= current speed, new cpu core >= current cpu core
so new cpu shares is always equal or bigger than old cpu shares.
the check may be unnecessary but looks ok.

weizhouapache (approved):
code lgtm

BryanMLima:
@blueorangutan package

BryanMLima:
The error was caused while trying to download a Template, so it had no relation to the code itself.
image

weizhouapache:

weizhouapache:

weizhouapache:

weizhouapache:

weizhouapache:

weizhouapache:

weizhouapache:
@blueorangutan package

weizhouapache:

BryanMLima:
The recurrent errors `test_01_migrate_VM_and_root_volume` and `test_02_migrate_VM_with_two_data_disks` seems to be related to the Ubuntu OS. I tested both situations on my lab, and the command ran smoothly.  Moreover, my PR is not related to these commands. @DaanHoogland could you look into this? The test on CentOS didn't shown any errors either.

Pearl1594:
Ran the above failed tests on a 4.16.1 env - and the same behavior was observed. Test failures as mentioned aren't related to this PR - the specific issue is:
```
 Failed to migrated vm VM instance {id: "14", name: "i-9-14-VM", uuid: "a1ba2a16-2e59-43c9-bd43-d55309f51a0b", type="User"} along with its volumes. Can't find strategy to move data. Source Host: ref-trl-2732-k-Mu20-pearl-dsilva-kvm1, Destination Host: ref-trl-2732-k-Mu20-pearl-dsilva-kvm2, Volume UUIDs: 0b96a192-0ecc-4de3-84f6-6090230cd4a0

```

weizhouapache:
thanks @Pearl1594 for investigation.

let's merge this PR @nvazquez

nvazquez:
Merging based on approvals and tests results - failing tests are only for ubuntu 20 which are being tracked on #6114

PR #6048, opened 2022-02-28
Refactor account type

JoaoJandre:
### Description

This PR aims to refactor the Account class by changing both the account type and the account state to de facto enums to facilitate future ACS development. Moreover, the use of well-documented and clear enum classes is a good practice in Java, and as such, we should try to adopt such good practices to reduce our technical debt.

### Types of changes

- [X] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
This was tested in a local lab, since the refactoring affects a lot of classes, all the basic operations were tested:
Creating new VMs;
Creating new networks;
Creating new volumes, snapshots and VM snapshots;
Creating new accounts;
Creating new roles;

### DIFF
--- a/api/src/main/java/com/cloud/user/Account.java
+++ b/api/src/main/java/com/cloud/user/Account.java
@@ -16,35 +16,55 @@
 // under the License.
 package com.cloud.user;
 
-import java.util.Date;
-
 import org.apache.cloudstack.acl.ControlledEntity;
 import org.apache.cloudstack.api.Identity;
 import org.apache.cloudstack.api.InternalIdentity;
 
+import java.util.Date;
+import java.util.HashMap;
+
 public interface Account extends ControlledEntity, InternalIdentity, Identity {
 
+    /**
+     *  Account states.
+     * */
+    enum State {
+        DISABLED, ENABLED, LOCKED;
 
-    public enum State {
-        disabled, enabled, locked
+        /**
+         * The toString method was overridden to maintain consistency in the DB, as the GenericDaoBase uses toString in the enum value to make the sql statements
+         * and previously the enum was in lowercase.
+         * */
+        @Override
+        public String toString(){
+            return super.toString().toLowerCase();
+        }
     }
 
-    public static final short ACCOUNT_TYPE_NORMAL = 0;
-    public static final short ACCOUNT_TYPE_ADMIN = 1;
-    public static final short ACCOUNT_TYPE_DOMAIN_ADMIN = 2;
-    public static final short ACCOUNT_TYPE_RESOURCE_DOMAIN_ADMIN = 3;
-    public static final short ACCOUNT_TYPE_READ_ONLY_ADMIN = 4;
-    public static final short ACCOUNT_TYPE_PROJECT = 5;
+    /**
+     * Account types.
+     * */
+    enum Type {
+        NORMAL, ADMIN, DOMAIN_ADMIN, RESOURCE_DOMAIN_ADMIN, READ_ONLY_ADMIN, PROJECT;
 
-    public static final String ACCOUNT_STATE_DISABLED = "disabled";
-    public static final String ACCOUNT_STATE_ENABLED = "enabled";
-    public static final String ACCOUNT_STATE_LOCKED = "locked";
+        private static HashMap<Integer,Type> ACCOUNT_TYPE_MAP = new HashMap<>();
+
+        static {
+            for (Type t: Type.values()) {
+                ACCOUNT_TYPE_MAP.put(t.ordinal(),t);
+            }
+        }
+
+        public static Type getFromValue(Integer type){
+            return ACCOUNT_TYPE_MAP.get(type);
+        }
+    }
 
     public static final long ACCOUNT_ID_SYSTEM = 1;
 
     public String getAccountName();
 
-    public short getType();
+    public Type getType();
 
     public Long getRoleId();
 
--- a/api/src/main/java/org/apache/cloudstack/acl/RoleType.java
+++ b/api/src/main/java/org/apache/cloudstack/acl/RoleType.java
@@ -20,20 +20,32 @@ import org.apache.commons.lang3.StringUtils;
 
 import com.cloud.user.Account;
 import com.google.common.base.Enums;
+import org.apache.log4j.Logger;
+
+import java.util.HashMap;
 
 // Enum for default roles in CloudStack
 public enum RoleType {
-    Admin(1L, Account.ACCOUNT_TYPE_ADMIN, 1),
-    ResourceAdmin(2L, Account.ACCOUNT_TYPE_RESOURCE_DOMAIN_ADMIN, 2),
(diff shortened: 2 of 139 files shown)


api/src/test/java/com/cloud/user/AccountTypeTest.java
```diff
@@ -0,0 +1,36 @@
+package com.cloud.user;
```
DaanHoogland, line 1: license

api/src/main/java/com/cloud/user/Account.java
```diff
@@ -16,35 +16,55 @@
...
-    public static final short ACCOUNT_TYPE_ADMIN = 1;
-    public static final short ACCOUNT_TYPE_DOMAIN_ADMIN = 2;
-    public static final short ACCOUNT_TYPE_RESOURCE_DOMAIN_ADMIN = 3;
-    public static final short ACCOUNT_TYPE_READ_ONLY_ADMIN = 4;
-    public static final short ACCOUNT_TYPE_PROJECT = 5;
+    /**
+     * Account types.
+     * */
+    enum Type {
+        NORMAL, ADMIN, DOMAIN_ADMIN, RESOURCE_DOMAIN_ADMIN, READ_ONLY_ADMIN, PROJECT;
```
DaanHoogland, line 48: should we hardcode these to their ordinals? I'm a bit worried regressions will occur in the future.

DaanHoogland (commented):
good cleanup! tnx @JoaoJandre. I'm not sure if I missed it (amongst the 139 files changed) but do we need a upgrade path for this?

api/src/main/java/com/cloud/user/Account.java
JoaoJandre: I opted to hardcode them to ordinals because the `prepareAttribute` method in the `GenericDaoBase` uses the `ordinal()` method to convert enums to integers when they are annotated as `EnumType.ORDINAL`; this ensures compatibility to the current DB schema. Moreover, in the AccountTypeTest there is a unit test to ensure that the order is maintained.
I hope to extend this refactoring in the future so that both the API and the DB schema use the string value of the enums instead of the integer value, thus making the application more intuitive.
To answer your question, since the `Account.Type` uses `EnumType.ORDINAL`, we don't need a upgrade path to change anything in the DB.

api/src/main/java/com/cloud/user/Account.java
```diff
@@ -16,35 +16,55 @@
...
+
+        private static HashMap<Integer,Type> ACCOUNT_TYPE_MAP = new HashMap<>();
```
ravening, line 50: Better to use `private static Map<Integer, Type>` rather than `HashMap`

api/src/main/java/org/apache/cloudstack/acl/RoleType.java
```diff
@@ -20,20 +20,32 @@
...
-    Admin(1L, Account.ACCOUNT_TYPE_ADMIN, 1),
-    ResourceAdmin(2L, Account.ACCOUNT_TYPE_RESOURCE_DOMAIN_ADMIN, 2),
-    DomainAdmin(3L, Account.ACCOUNT_TYPE_DOMAIN_ADMIN, 4),
-    User(4L, Account.ACCOUNT_TYPE_NORMAL, 8),
-    Unknown(-1L, (short) -1, 0);
+    Admin(1L, Account.Type.ADMIN, 1),
+    ResourceAdmin(2L, Account.Type.RESOURCE_DOMAIN_ADMIN, 2),
+    DomainAdmin(3L, Account.Type.DOMAIN_ADMIN, 4),
+    User(4L, Account.Type.NORMAL, 8),
+    Unknown(-1L,  null, 0);
```
ravening, line 33: Better to use `Type.UNKNOWN` rather than `null` to prevent unknown NPE

api/src/main/java/org/apache/cloudstack/acl/RoleType.java
```diff
@@ -68,26 +80,15 @@ public static RoleType fromMask(int mask) {
...
-                break;
-            case Account.ACCOUNT_TYPE_RESOURCE_DOMAIN_ADMIN:
-                roleType = RoleType.ResourceAdmin;
-                break;
-            case Account.ACCOUNT_TYPE_NORMAL:
-                roleType = RoleType.User;
-                break;
+    public static RoleType getByAccountType(final Account.Type accountType) {
+        RoleType t = ACCOUNT_TYPE_MAP.get(accountType);
+        if (t == null) {
```
ravening, line 85: if you use `Type.UNKNOWN` rather than null then this can be simplified to

`return (t == UNKNOWN) ? RoleType.UNKNOWN : t`

JoaoJandre: I prefer not to use contracted ifs, I believe it worsens the legibility for no gain whatsoever.

ravening: its not about `if` here but its about not using `null`

JoaoJandre: The problem is, if a `Account.Type` is passed that is not in the map, `ACCOUNT_TYPE_MAP.get(accountType)` will return null nonetheless, for example, passing Account.Type.PROJECT will return null. So checking `if (t == null)` is still needed.

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan test

nvazquez:
Hi @JoaoJandre can you please check the smoke tests?

JoaoJandre:
Sure @nvazquez, I just checked them and found the error, in the `findUserAccountByApiKey(String apiKey)` the `State.valueOf()`method was throwing an error because the state retrieved from the DB was in lowercase. I created a new method to be used instead of `valueOf()` since we cannot override it. It seems that the only place it was being used was in that method, so that was the only place I changed.

nvazquez:
Thanks @JoaoJandre - will kick a new round of tests
@blueorangutan package

nvazquez:
@blueorangutan test

DaanHoogland (approved):
clgtm

nvazquez (approved):
LGTM

GutoVeronezi:
@blueorangutan package

GutoVeronezi (approved):
LGTM, nice work @JoaoJandre

GutoVeronezi:
Merging it based on the approvals and test results.

rohityadavcloud:
@nvazquez were we waiting for @borisstoyanov who's assigned on this to manually test this PR?

nvazquez:
Ideally yes @rohityadavcloud - but not an issue in this case since tests have passed and changes were purely cleanup

PR #6057, opened 2022-03-04
server: mark volume snapshots as Destroyed if it does not exist on primary and secondary storage when delete a volume

### LINKED ISSUE #5433: KVM problem with volume snapshots kept only on the primary storage (non-managed storages)
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
volume snapshots 
~~~

##### CLOUDSTACK VERSION

~~~
from 4.11.3.0 to main
~~~

##### SUMMARY

When `snapshot.backup.to.secondary` is set to `false` the snapshots are kept only on the primary storage. When you take more than one snapshot of a volume and delete this volume the snapshots are still active in the UI/DB but deleted from the storage. The problem with different primary storages is:
- for the current NFS implementation is that the snapshots are internal, and they are deleted when you delete the volume - active in the UI/DB
- for Ceph - the snapshots are deleted from the pool, but active in DB/UI
- probably with the new Linstor driver the problem will be the same
- for StorPool the snapshots are not deleted, but a random record is deleted from the `snapshot_store_ref` table and the reference to the snapshot's path is lost
- with this PR - [#5297](https://github.com/apache/cloudstack/pull/5297) - the problem will be the same as the StorPool primary storage

##### STEPS TO REPRODUCE

~~~
- set `snapshot.backup.to.secondary` to `false`
- take a few snapshots of a volume
- delete the volume

The snapshots are visible in the UI, active in DB table `snapshots` and one of them is deleted in `snapshot_store_ref` DB table, because of this code in VolumeServiceImpl.deleteVolumeCallback:


                SnapshotDataStoreVO snapStoreVo = _snapshotStoreDao.findByVolume(vo.getId(), DataStoreRole.Primary);

                if (snapStoreVo != null) {
                    long storagePoolId = snapStoreVo.getDataStoreId();
                    StoragePoolVO storagePoolVO = storagePoolDao.findById(storagePoolId);

                    if (storagePoolVO.isManaged()) {
                        DataStore primaryDataStore = dataStoreMgr.getPrimaryDataStore(storagePoolId);
                        Map<String, String> mapCapabilities = primaryDataStore.getDriver().getCapabilities();

                        String value = mapCapabilities.get(DataStoreCapabilities.STORAGE_SYSTEM_SNAPSHOT.toString());
                        Boolean supportsStorageSystemSnapshots = new Boolean(value);

                        if (!supportsStorageSystemSnapshots) {
                            _snapshotStoreDao.remove(snapStoreVo.getId());
                        }
                    } else {
                        _snapshotStoreDao.remove(snapStoreVo.getId());
                    }
                }



 ~~~

##### EXPECTED RESULTS

~~~
Probably for all non-managed primary storages we should not delete the snapshots. For NFS this PR - #5297 will fix the problem with the internal snapshots. 

~~~

##### ACTUAL RESULTS

~~~
Snapshots are deleted from the storage, but active in the UI/DB as a garbage 
~~~

weizhouapache:
when delete a volume in QCOW2 format, if volume snapshot does not exist on primary and secondary storage, mark the snapshot as Destroyed.

This fixes #5433 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
+++ b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
@@ -122,6 +122,7 @@ import com.cloud.storage.VolumeVO;
 import com.cloud.storage.dao.VMTemplatePoolDao;
 import com.cloud.storage.dao.VolumeDao;
 import com.cloud.storage.dao.VolumeDetailsDao;
+import com.cloud.storage.snapshot.SnapshotApiService;
 import com.cloud.storage.snapshot.SnapshotManager;
 import com.cloud.storage.template.TemplateProp;
 import com.cloud.user.AccountManager;
@@ -194,6 +195,8 @@ public class VolumeServiceImpl implements VolumeService {
     private StorageManager _storageMgr;
     @Inject
     private AnnotationDao annotationDao;
+    @Inject
+    private SnapshotApiService snapshotApiService;
 
     private final static String SNAPSHOT_ID = "SNAPSHOT_ID";
 
@@ -468,6 +471,7 @@ public class VolumeServiceImpl implements VolumeService {
                         _snapshotStoreDao.remove(snapStoreVo.getId());
                     }
                 }
+                snapshotApiService.markVolumeSnapshotsAsDestroyed(vo);
             } else {
                 vo.processEvent(Event.OperationFailed);
                 apiResult.setResult(result.getResult());
--- a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
+++ b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
@@ -448,9 +448,9 @@ public class VolumeServiceImpl implements VolumeService {
                     volDao.remove(vo.getId());
                 }
 
-                SnapshotDataStoreVO snapStoreVo = _snapshotStoreDao.findByVolume(vo.getId(), DataStoreRole.Primary);
+                List<SnapshotDataStoreVO> snapStoreVOs = _snapshotStoreDao.listAllByVolumeAndDataStore(vo.getId(), DataStoreRole.Primary);
 
-                if (snapStoreVo != null) {
+                for (SnapshotDataStoreVO snapStoreVo : snapStoreVOs) {
                     long storagePoolId = snapStoreVo.getDataStoreId();
                     StoragePoolVO storagePoolVO = storagePoolDao.findById(storagePoolId);
 
--- a/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
@@ -1583,25 +1583,10 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
                 s_logger.warn("Failed to expunge volume: " + volumeId);
                 return null;
             }
-            markVolumeSnapshotsAsDestroyed(volume);
             removeVolume(volume.getId());
         }
 
         return volume;
-    }
-
-    private void markVolumeSnapshotsAsDestroyed(VolumeVO volume) {
-        if (!Storage.ImageFormat.QCOW2.equals(volume.getFormat())) {
-            return;
-        }
-        List<SnapshotVO> snapshots = _snapshotDao.listByVolumeId(volume.getId());
-        for (SnapshotVO snapshot: snapshots) {
-            List<SnapshotDataStoreVO> snapshotDataStoreVOs = _snapshotDataStoreDao.findBySnapshotId(snapshot.getId());
-            if (CollectionUtils.isEmpty(snapshotDataStoreVOs)) {
-                snapshot.setState(Snapshot.State.Destroyed);
-                _snapshotDao.update(snapshot.getId(), snapshot);
-            }
-        }
     }
 
     @Override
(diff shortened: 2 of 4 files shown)


weizhouapache:
@blueorangutan package

nvazquez:
@weizhouapache can you please target the PR to the main branch?
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez:
@blueorangutan test

nvazquez (approved):
LGTM

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1583,12 +1583,27 @@ public Volume destroyVolume(long volumeId, Account caller, boolean expunge, bool
...
             }
+            markVolumeSnapshotsAsDestroyed(volume);
             removeVolume(volume.getId());
         }
 
         return volume;
     }
 
+    private void markVolumeSnapshotsAsDestroyed(VolumeVO volume) {
+        if (!Storage.ImageFormat.QCOW2.equals(volume.getFormat())) {
```
slavkap, line 1594: If you have a VM with Ceph's volumes, for example, and `snapshot.backup.to.secondary` is disabled when you expunge the VM, all snapshots remain active in UI, but there are no records in the `snapshot_store_ref` table.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1583,12 +1583,27 @@ public Volume destroyVolume(long volumeId, Account caller, boolean expunge, bool
                 s_logger.warn("Failed to expunge volume: " + volumeId);
                 return null;
             }
+            markVolumeSnapshotsAsDestroyed(volume);
```
slavkap, line 1586: @weizhouapache, this is not applied for the ROOT volumes. If the `snapshot.backup.to.secondary` is disabled and you have snapshots on the ROOT volume when you expunge the VM the snapshots will be in a `BackedUp` state in the DB table `snapshots`, but with no records in the `snapshot_store_ref` table, it is removed.

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
```diff
@@ -448,9 +448,9 @@ public Void deleteVolumeCallback(AsyncCallbackDispatcher<VolumeServiceImpl, Comm
                     volDao.remove(vo.getId());
                 }
 
-                SnapshotDataStoreVO snapStoreVo = _snapshotStoreDao.findByVolume(vo.getId(), DataStoreRole.Primary);
+                List<SnapshotDataStoreVO> snapStoreVOs = _snapshotStoreDao.listAllByVolumeAndDataStore(vo.getId(), DataStoreRole.Primary);
 
-                if (snapStoreVo != null) {
+                for (SnapshotDataStoreVO snapStoreVo : snapStoreVOs) {
```
slavkap, line 453: It's only my opinion, but we should handle this in each storage plugin. Each storage has to decide whether it wants to keep the snapshots on the primary, like they handle deleting or keeping the snapshots when a volume is deleted.

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
weizhouapache: @slavkap 
I was not quite sure if the snapshots still work and removing this check will not lead to other unexpected issues.
If we can make sure it is not a problem to mark volume snapshot as destroyed if snapshot does not exist in primary and secondary storage, we can remove this check.

slavkap: Yes, it's better to remove this check because if we don't have snapshots on primary and secondary storage, the snapshot has to be marked as destroyed. I will test it

weizhouapache: @slavkap 
good.
I have removed this check, and fixed the issue in vm expunge/destroy.

nvazquez:
@blueorangutan package

weizhouapache:
@blueorangutan test

DaanHoogland (approved):
clgtm

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
```diff
@@ -468,6 +471,7 @@ public Void deleteVolumeCallback(AsyncCallbackDispatcher<VolumeServiceImpl, Comm
                         _snapshotStoreDao.remove(snapStoreVo.getId());
```
slavkap, line 471: Hi @weizhouapache, isn't it better to keep the info in the DB for the snapshots and update the State as `Destroyed`?

slavkap (approved):
code LGTM - tested but left a comment

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/volume/VolumeServiceImpl.java
weizhouapache: @slavkap 
is there any benifit ?
the records will be cleaned up by storage garbage collector.

weizhouapache: @slavkap 
thanks for testing

slavkap: @weizhouapache, to keep it as history, nothing more

weizhouapache: @slavkap ah ok, thanks for reply.
the records will be finally cleaned up by storage garbage collector (by default run once every day). They will be kept for only 1 day.
The behaviour is not changed by this PR. It was introduced by 336df84f1787de962a67d0a34551f9027303040e and ccb68e13a74342d4798aae11fe5c57890ff6c0ae

nvazquez:
Merging based on approvals and test results

PR #6081, opened 2022-03-09
[UI] Dont show project view menu when user doesn't have permission

BryanMLima:
### Description

This PR improves clarity to the user concerning the project list view component. The current behavior disables the project view menu when an Account does not have permission to list projects. However, that can be a bit confusing for users, if they have no permission to list projects, then, the expected behavior would be to hide the component that the user does not have access to (that is what happens with other components). Therefore, this PR aims to hide the project view when the user does not have the permission list projects.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

**Before:**
image

**After:**
image

### How Has This Been Tested?
This was tested in a local lab.

Login with an Account that had the permission to the api `listProjects`:
It displayed the Project Menu view as expected.

Login with an Account that didn't have the permission to the api `listProjects`:
The Project View Menu is not shown, as the **After** image in the Screenshots section.

### DIFF
--- a/ui/src/components/header/ProjectMenu.vue
+++ b/ui/src/components/header/ProjectMenu.vue
@@ -65,8 +65,7 @@ export default {
   data () {
     return {
       projects: [],
-      loading: false,
-      showProjectMenu: true
+      loading: false
     }
   },
   created () {
@@ -75,7 +74,6 @@ export default {
   methods: {
     fetchData () {
       if (this.isDisabled()) {
-        this.showProjectMenu = false
         return
       }
       var page = 1
--- a/ui/src/components/header/ProjectMenu.vue
+++ b/ui/src/components/header/ProjectMenu.vue
@@ -18,7 +18,7 @@
 <template>
   <span class="header-notice-opener">
     <a-select
-      v-if="showProjectMenu"
+      v-if="!isDisabled()"
       class="project-select"
       :defaultValue="$t('label.default.view')"
       :loading="loading"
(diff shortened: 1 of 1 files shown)


GabrielBrascher (approved):
LGTM

nvazquez:
@blueorangutan ui

ui/src/components/header/ProjectMenu.vue
```diff
@@ -18,11 +18,11 @@
 <template>
   <span class="header-notice-opener">
     <a-select
+      v-if="showProjectMenu"
```
utchoang, line 21: I think just like this instead of creating a new variable
```suggestion
      v-if="!isDisabled()"
```

utchoang (changes requested):

ui/src/components/header/ProjectMenu.vue
davidjumani: +1

BryanMLima:
When comparing a feature like live scaling a VM, it makes sense to disabled the button when the user has the permission to do so, but the API cannot be executed considering some external factors, like the VM is not dynamically scalable, for example. In this case, showing the user that he can live scale a VM makes sense, even though he can't in that specific case.
However, when considering the API `listProjects`, the side menu does not show the Project section, so it makes sense to hide the Project Menu on the top bar as well. Moreover, if the user has the permission to an API, then it should show at all times, even though the API can't be executed at that time, like the live scale mentioned above. In this case, it should be displayed as disabled. However, if the user doesn't have the permission to the API, it should not be displayed at all.

nvazquez:
Thanks for the explanation @BryanMLima LGTM

nvazquez (approved):

PR #4339, opened 2020-09-23
Allow domain admin to change domain and account settings

davidjumani:
## Description
Allow domain admin to change domain and account settings

Fixes https://github.com/apache/cloudstack/issues/3752 && https://github.com/apache/cloudstack/issues/5494

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

## How Has This Been Tested?

Screenshot from 2021-09-21 15-40-53

PR #6101, opened 2022-03-11
server: add default hypervisor to supported hypervisor list of VRs

### LINKED ISSUE #6084: router service offering tags are not honored
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Router Deployment
~~~

##### CLOUDSTACK VERSION

~~~
4.16
~~~

##### CONFIGURATION

Having both VMware and Xen clusters
custom offering for router with host tag: xen 
Have a XenServer/xcp-ng host tagged with tag: xen
system.vm.default.hypervisor = XenServer
router.service.offering = [uuid-of-custom-offering]

##### SUMMARY

When a user has 2 types of hypervisors, and he wants to create VRs on one of the types (say Xen), he would create a host tag in the custom router offering and tag a xen host/cluster. Then they would change the system.vm.default.hypervisor to XenServer. 

##### Steps

~~~
- set the configuration described
- deploy a vmware template in a newly created network
- observe the failure
~~~

##### EXPECTED RESULTS

~~~
deployment successful, router on xen
~~~

##### ACTUAL RESULTS

~~~
deployment fails, cloudstack tried to deploy an ova to xen
~~~
[vmware.log](https://github.com/apache/cloudstack/files/8221143/vmware.log)

weizhouapache:
### Description

This PR fixes #6084 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java
+++ b/server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java
@@ -573,6 +573,10 @@ public class NetworkHelperImpl implements NetworkHelper {
             } else {
                 hypervisors.add(dest.getCluster().getHypervisorType());
             }
+            final HypervisorType defaults = _resourceMgr.getDefaultHypervisor(dest.getDataCenter().getId());
+            if (defaults != HypervisorType.None && !hypervisors.contains(defaults)) {
+                hypervisors.add(defaults);
+            }
         } else {
             final HypervisorType defaults = _resourceMgr.getDefaultHypervisor(dest.getDataCenter().getId());
             if (defaults != HypervisorType.None) {


weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

nvazquez (approved):
Code LGTM

nvazquez:
Thanks @weizhouapache - I think it could be better if it is a zone-wide setting since it could contain different hypervisors clusters. What do you think @weizhouapache @borisstoyanov?

rohityadavcloud:
Should accounts be allowed to do this? We generally don't expose any implementation/hardware settings to end accounts or domains.

weizhouapache:
> Should accounts be allowed to do this? We generally don't expose any implementation/hardware settings to end accounts or domains.

@rohityadavcloud 
account-level setting can only be set by root admin in the past. 
but with #4339, domain admin can change them as well.

PaulAngus:
Hi Wei,
Could you explain why we would want this.  on the face of it, it doesn't seem a good idea.
One of the biggest points of a Cloud Management Platform is to abstract the underlying hardware & hypervisor away from the users.

weizhouapache:
> Thanks @weizhouapache - I think it could be better if it is a zone-wide setting since it could contain different hypervisors clusters. What do you think @weizhouapache @borisstoyanov?

@nvazquez if it is a zone-wide setting,  all networks in the zone will be impacted

weizhouapache:
> Hi Wei, Could you explain why we would want this. on the face of it, it doesn't seem a good idea. One of the biggest points of a Cloud Management Platform is to abstract the underlying hardware & hypervisor away from the users.

Hi @PaulAngus 
we have a customer issue. the customer has two clusters with different hypervisors. They want to deploy user VMs in the cluster A with hypervisor A but want the VR to be deployed in the cluster B with hypervisor B.
since many years ago, cloudstack tries to deploy network VR using same deployment destination as user vm, the VRs are created with the same hypervisor as user vm, they cannot be started to another cluster because the hypervisor is different.

The workaround is, deploy another user vm C in cluster B, after network VR is started in cluster B, deploy the user vm in Cluster A. The PR aims to provide a resolution.

As said before, The scope of the new setting can be discussed. The higher level means bigger impact (e.g. Zone), the lower level means smaller impact (e.g. Network Offering or Account).

CloudStack has some settings which impact the vm/vr allocation on hardware, some are visible to users (e.g. affinity group,dedication), some are not (e.g. host tags, storage tags). the account setting (not limited to this new setting) are not visible to users, but yes to root admins and domain admins.

rohityadavcloud:
@weizhouapache could we use tags/labels with some way to link/specify them in-network offerings be used as a means of specifying preferred hypervisor for router instead? So for ex. users using Network Offering A will get VR deployed in Hypervisor X...

weizhouapache:
> @weizhouapache could we use tags/labels with some way to link/specify them in-network offerings be used as a means of specifying preferred hypervisor for router instead? So for ex. users using Network Offering A will get VR deployed in Hypervisor X...

@rohityadavcloud 
yes, that's an option, similar as what I posted in https://github.com/apache/cloudstack/pull/6101#issuecomment-1068951583

PR #6143, opened 2022-03-21
api: Remove redundant API parameters

Pearl1594:
### Description

This PR removes the redundant api parameters listed under listVirtualMachines API. The parameters are already defined in `ListVMsCmdByAdmin` class
```
(localcloud) 🐱 > list virtualmachines -h
listVirtualMachines: List the virtual machines owned by the account.
API Params               Type     Description
==========               ====     ===========
account                  string   list resources by account. Must be used w
                                  ith the domainId parameter.
...
hostid                   uuid     the host ID
hostid                   uuid     the host ID
...
podid                    uuid     the pod ID
podid                    uuid     the pod ID
...
storageid                uuid     the storage ID where vm's volumes belong 
                                  to
storageid                uuid     the storage ID where vm's volumes belong 
                                  to
...
```

### Types of changes

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
@@ -38,14 +38,11 @@ import org.apache.cloudstack.api.BaseListTaggedResourcesCmd;
 import org.apache.cloudstack.api.Parameter;
 import org.apache.cloudstack.api.ResponseObject.ResponseView;
 import org.apache.cloudstack.api.response.BackupOfferingResponse;
-import org.apache.cloudstack.api.response.HostResponse;
 import org.apache.cloudstack.api.response.InstanceGroupResponse;
 import org.apache.cloudstack.api.response.IsoVmResponse;
 import org.apache.cloudstack.api.response.ListResponse;
 import org.apache.cloudstack.api.response.NetworkResponse;
-import org.apache.cloudstack.api.response.PodResponse;
 import org.apache.cloudstack.api.response.ServiceOfferingResponse;
-import org.apache.cloudstack.api.response.StoragePoolResponse;
 import org.apache.cloudstack.api.response.TemplateResponse;
 import org.apache.cloudstack.api.response.UserVmResponse;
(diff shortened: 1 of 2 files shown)


weizhouapache (approved):
code lgtm

weizhouapache:
@blueorangutan package

Pearl1594:
@blueorangutan package

GabrielBrascher (commented):
Thanks for the PR, @Pearl1594.

I would like to double-check if those parameters should be aimed only at Admins.
I ask this because it is not clear if that's the goal of this PR as well, or if it is aiming just at removing the redundant/duplicated parameters.

If the goal is to remove duplicated parameters but keep the current API behavior (Users & Admins), then I believe it should instead be removed at the `ListVMsCmdAdmin` which extends `ListVMsCmd`.

Pearl1594:
Thanks for reviewing @GabrielBrascher , the following params: pod_id, storage_id, host_id, seem to be used only applicable only for root admins - https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/api/query/QueryManagerImpl.java#L1004-L1011 , hence I took the following approach. It removes the redundant params and also ensures to behave the same way as before.

DaanHoogland:
@Pearl1594 shouldn't a domain- or resource admin be allowed to list by host, pod, zone ?

GabrielBrascher:
Sorry @Pearl1594, I raised the question regarding  "Admins & Users".
However, my concerns are regarding what @DaanHoogland mentioned, Admins & Domain Admins.

Pearl1594:
It seems that only listing zones is allowed by domainAdmins - listing storage pools, hosts, clusters and pods, are restricted to root admins only. So, this change shouldn't effect any behavior.

Pearl1594:
@blueorangutan package

weizhouapache (approved):
code lgtm

Pearl1594:
@blueorangutan test

GabrielBrascher (approved):
LGTM

nvazquez (approved):
Code LGTM, checked the UI codebase and could not find any of the params being passed to the listVirtualMachines API

nvazquez:
Have tested different roles and as advised by @Pearl1594 the removed parameters only apply for ROOT admin

PR #6146, opened 2022-03-21
configDrive: Fix failure to delete (unstarted) VM

Pearl1594:
### Description

This PR fixes the issue observed when trying to expunge a VM created using config drive userdata service (with startVM=false) and the env has no secondary stores.  

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
1. On an env with no secondary store, set the global setting: `vm.configdrive.primarypool.enabled` to true
2. Deploy a VM with a network offering with configDrive userdata service, but do not start it
3. Attempt deleting(expunge) the VM - the VM deletes successfully without NPE 

Performed the above test with  `vm.configdrive.primarypool.enabled` set to false and was also able to expunge the VM

### DIFF
--- a/server/src/main/java/com/cloud/network/element/ConfigDriveNetworkElement.java
+++ b/server/src/main/java/com/cloud/network/element/ConfigDriveNetworkElement.java
@@ -497,7 +497,7 @@ public class ConfigDriveNetworkElement extends AdapterBase implements NetworkEle
             agentId = dest.getHost().getId();
         }
         if (!VirtualMachineManager.VmConfigDriveOnPrimaryPool.valueIn(dest.getDataCenter().getId()) &&
-                !VirtualMachineManager.VmConfigDriveForceHostCacheUse.valueIn(dest.getDataCenter().getId())) {
+                !VirtualMachineManager.VmConfigDriveForceHostCacheUse.valueIn(dest.getDataCenter().getId()) && dataStore != null) {
             agentId = findAgentIdForImageStore(dataStore);
         }
         return agentId;
@@ -505,6 +505,7 @@ public class ConfigDriveNetworkElement extends AdapterBase implements NetworkEle
 
     private Location getConfigDriveLocation(long vmId) {
         final UserVmDetailVO vmDetailConfigDriveLocation = _userVmDetailsDao.findDetail(vmId, VmDetailConstants.CONFIG_DRIVE_LOCATION);
+        UserVmVO virtualMachine = _userVmDao.findById(vmId);
         if (vmDetailConfigDriveLocation != null) {
             if (Location.HOST.toString().equalsIgnoreCase(vmDetailConfigDriveLocation.getValue())) {
                 return Location.HOST;
@@ -513,6 +514,9 @@ public class ConfigDriveNetworkElement extends AdapterBase implements NetworkEle
             } else {
                 return Location.SECONDARY;
             }
+        }
+        if (virtualMachine.getState() == VirtualMachine.State.Stopped && virtualMachine.getLastHostId() == null && VirtualMachineManager.VmConfigDriveOnPrimaryPool.valueIn(virtualMachine.getDataCenterId())) {
+            return Location.PRIMARY;
         }
         return Location.SECONDARY;
     }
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/cloud/network/element/ConfigDriveNetworkElement.java
```diff
@@ -639,7 +639,10 @@ private boolean deleteConfigDriveIso(final VirtualMachine vm) throws ResourceUna
 
         if (location == Location.SECONDARY) {
             dataStore = _dataStoreMgr.getImageStoreWithFreeCapacity(vm.getDataCenterId());
-            agentId = findAgentIdForImageStore(dataStore);
+            if (!VirtualMachineManager.VmConfigDriveOnPrimaryPool.valueIn(vm.getDataCenterId()) &&
+                    !VirtualMachineManager.VmConfigDriveForceHostCacheUse.valueIn(vm.getDataCenterId()) && dataStore != null) {
+                agentId = findAgentIdForImageStore(dataStore);
+            }
```
sureshanaparti, line 645: @Pearl1594 Any reason, why config drive deletion should depend on these settings _vm.configdrive.primarypool.enabled / vm.configdrive.use.host.cache.on.unsupported.pool_  (which are dynamic) ?

server/src/main/java/com/cloud/network/element/ConfigDriveNetworkElement.java
```diff
@@ -639,7 +639,10 @@ private boolean deleteConfigDriveIso(final VirtualMachine vm) throws ResourceUna
 
         if (location == Location.SECONDARY) {
             dataStore = _dataStoreMgr.getImageStoreWithFreeCapacity(vm.getDataCenterId());
-            agentId = findAgentIdForImageStore(dataStore);
+            if (!VirtualMachineManager.VmConfigDriveOnPrimaryPool.valueIn(vm.getDataCenterId()) &&
+                    !VirtualMachineManager.VmConfigDriveForceHostCacheUse.valueIn(vm.getDataCenterId()) && dataStore != null) {
+                agentId = findAgentIdForImageStore(dataStore);
```
weizhouapache, line 644: @Pearl1594 will `agentId` be null if the conditions do not match ?

Pearl1594: yes @weizhouapache it could be null - and it's handled by https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/network/element/ConfigDriveNetworkElement.java#L651 - however, this exception, doesn't prevent the deletion of the VM

server/src/main/java/com/cloud/network/element/ConfigDriveNetworkElement.java
Pearl1594: @sureshanaparti if case the config drive iso for the vm is on primary store or image cache, then there is no need to search for image store agent id ..

sureshanaparti: But, these settings doesn't indicate the VM config drive on primary store or host cache. What if these are changed after VM creation?

sureshanaparti: > But, these settings doesn't indicate the VM config drive on primary store or host cache. What if these are changed after VM creation? Only when deploying a VM, this settings help to choose the config drive location.

weizhouapache: agree with @sureshanaparti 
it is better to check what do exist on primary, secondary or host.

Pearl1594: got it @sureshanaparti @weizhouapache - I'll handle it Thanks.

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

weizhouapache:
@blueorangutan test

server/src/main/java/com/cloud/network/element/ConfigDriveNetworkElement.java
```diff
@@ -630,6 +630,10 @@ private DataStore getDatastoreForConfigDriveIso(DiskTO disk, VirtualMachineProfi
     private boolean deleteConfigDriveIso(final VirtualMachine vm) throws ResourceUnavailableException {
         Long hostId  = (vm.getHostId() != null) ? vm.getHostId() : vm.getLastHostId();
         Location location = getConfigDriveLocation(vm.getId());
```
weizhouapache, line 632: @Pearl1594 
move `Location location = getConfigDriveLocation(vm.getId());` to line 637 (just before `if (location == Location.HOST) {`) ?

weizhouapache (approved):
tested configdrive ISO on primary (local host and shared NFS storage) and secondary storage.
looks good.

left a minor comment.

nvazquez (approved):

Issue #6127, opened 2022-03-16
SSO fails with error "Expired session, missing signature, or missing apiKey"


kohrar:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Authentication, SAML2
~~~

##### CLOUDSTACK VERSION

~~~
4.16.1
~~~

##### CONFIGURATION

SAML2 authentication is enabled and configured to Microsoft Azure

##### OS / ENVIRONMENT

Rocky Linux 8.5

##### SUMMARY

CloudStack SSO works the first time on a clean browser session. However, after the user attempts to log out or when the CloudStack session expires, CloudStack will no longer allow the user to authenticate via SSO any more.

What happens on subsequent SSO sign-in attempts is the user gets sent to the SAML2 server which redirects them back with the token. Everything is good up to here. However, CloudStack for some reason invalidates the session and the user is redirected back to the login page but this time with the 'Single Sign-On' option disabled.

The management-server.log shows the SSO working initially.
```
2022-03-16 12:00:24,504 DEBUG [c.c.a.ApiServlet] (qtp365590665-1149:ctx-a6c7c3f1) (logid:e18c4373) ===START===  ###.###.96.250 -- GET  command=samlSso&idpid=https://sts.windows.net/c609a0ec-a5e3-4631-9686-#########/
2022-03-16 12:00:24,513 DEBUG [o.a.c.a.c.SAML2LoginAPIAuthenticatorCmd] (qtp365590665-1149:ctx-a6c7c3f1) (logid:e18c4373) Sending SAMLRequest id=idj8da8n95krcr1bco9r7h13a2008cfrv4
2022-03-16 12:00:24,560 DEBUG [c.c.a.ApiServlet] (qtp365590665-1149:ctx-a6c7c3f1) (logid:e18c4373) ===END===  ###.###.96.250 -- GET  command=samlSso&idpid=https://sts.windows.net/c609a0ec-a5e3-4631-9686-#########/
2022-03-16 12:00:25,340 DEBUG [c.c.a.ApiServlet] (qtp365590665-1149:ctx-43bf32d6) (logid:d2efb5a3) ===START===  ###.###.96.250 -- POST  command=samlSso
[...]

PR #6149, opened 2022-03-21
Update SAML2 auth sessionkey cookie path

kohrar:
### Description

This change will set the sessionkey under the /client/api path. This commit should prevent duplicate sessionkey cookies from being set on both /client (incorrect) and /client/api (correct). Prior to this commit, the /client version was being set while the /client/api version remained unchanged with an invalid sessionkey. As a result, subsequent requests after the SAML2 authentication would immediately fail with an invalid session and results in the user being logged out. 

The sessionkey is now set explicitly for the /client/api path which should fix this issue, regardless of the SSO URL and path that's being used.

Fixes: #6127

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

I've built 4.16.1.0 with this change and SAML2 auth is now working as expected again.

### DIFF
--- a/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
+++ b/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
@@ -280,7 +280,7 @@ public class SAMLUtils {
             resp.addCookie(new Cookie("timezone", URLEncoder.encode(timezone, HttpUtils.UTF_8)));
         }
         resp.addCookie(new Cookie("userfullname", URLEncoder.encode(loginResponse.getFirstName() + " " + loginResponse.getLastName(), HttpUtils.UTF_8).replace("+", "%20")));
-        resp.addHeader("SET-COOKIE", String.format("%s=%s;HttpOnly", ApiConstants.SESSIONKEY, loginResponse.getSessionKey()));
+        resp.addHeader("SET-COOKIE", String.format("%s=%s;HttpOnly;Path=/client/api", ApiConstants.SESSIONKEY, loginResponse.getSessionKey()));
     }
 
     /**


plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
```diff
@@ -280,7 +280,7 @@ public static void setupSamlUserCookies(final LoginCmdResponse loginResponse, fi
             resp.addCookie(new Cookie("timezone", URLEncoder.encode(timezone, HttpUtils.UTF_8)));
         }
         resp.addCookie(new Cookie("userfullname", URLEncoder.encode(loginResponse.getFirstName() + " " + loginResponse.getLastName(), HttpUtils.UTF_8).replace("+", "%20")));
-        resp.addHeader("SET-COOKIE", String.format("%s=%s;HttpOnly", ApiConstants.SESSIONKEY, loginResponse.getSessionKey()));
+        resp.addHeader("SET-COOKIE", String.format("%s=%s;HttpOnly;Path=/client/api", ApiConstants.SESSIONKEY, loginResponse.getSessionKey()));
```
rohityadavcloud, line 283: we might want to take the path from the configurable setting? cc @Pearl1594 to check if it makes sense to read the variable defined in server.properties

nvazquez:
Hi @kohrar is this PR still valid after the UI fixes?

kohrar:
I would think so. The UI issue is a separate issue. The original issue was that the session key cookie was set with a different path by the browser, resulting in the valid session key being ignored by the invalid key set previously.

The UI issue is merely an interface issue with the login page which can be worked around by refreshing the login page. The cookie issue here can't be worked around unless cookies are cleared completely.

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

rohityadavcloud:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

nvazquez (approved):

plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
rohityadavcloud: @kohrar can you check/adapt the same kind of code as in:
https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/api/ApiServlet.java#L224

rohityadavcloud: Nevermind, my comment- Based on https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/api/ApiServlet.java#L224 your changes LGTM (the Path /client/api is more restrictive than /client but should work and ensure all APIs function. The only issue is when the path isn't /client/api which is not a case I think think of as all APIs calls are handled at /client/api)

rohityadavcloud (approved):

Issue #6155, opened 2022-03-22
updateAccount don't change "type" column in DB


edikevich:
##### ISSUE TYPE
 * Bug Report

##### CLOUDSTACK VERSION
~~~
4.16.0
~~~

##### CONFIGURATION
XCP-NG 8.2
Advanced Zone

##### OS / ENVIRONMENT
Oracle Linux 8.4 - Management server, MYSQL 8.0.26
XCP-NG 8.2 - hypervisor

##### SUMMARY
updateAccount via CloudMonkey (also maybe via API) don't change "type" column in DB 

##### STEPS TO REPRODUCE
Create new account with User role
Change account role via CloudMonkey to Root Admin
Role is changed succesfuly but user can't login with new privileges

##### EXPECTED RESULTS
~~~
Account's role is changed succesfuly. User can login with new privileges
~~~

##### ACTUAL RESULTS
~~~
Account's role is changed succesfuly. User can't login with new privileges. Many errors on login page about this account dont have privileges to list  instances, network and etc. After manualy change "type" column in DB from 0 to 1 user can loggin with new privileges. no errors in management-server.log
~~~

PR #6156, opened 2022-03-23
api: Update account type when updating account role

Pearl1594:
### Description

This PR fixes: #6155  - updates account type to a value corresponding to the role.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
Updated role and attempted to login - successfully able to login with new rolw

### DIFF
--- a/server/src/main/java/com/cloud/user/AccountManagerImpl.java
+++ b/server/src/main/java/com/cloud/user/AccountManagerImpl.java
@@ -1181,6 +1181,18 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
         return _userAccountDao.findById(userId);
     }
 
+    private boolean isValidRoleChange(Account account, Role role) {
+        Long currentAccRoleId = account.getRoleId();
+        Role currentRole = roleService.findRole(currentAccRoleId);
+
+        if (role.getRoleType().ordinal() < currentRole.getRoleType().ordinal() && ((account.getType() == Account.Type.NORMAL && role.getRoleType().getAccountType().ordinal() > Account.Type.NORMAL.ordinal()) ||
+                account.getType().ordinal() > Account.Type.NORMAL.ordinal() && role.getRoleType().getAccountType().ordinal() < account.getType().ordinal() && role.getRoleType().getAccountType().ordinal() > 0)) {
+            throw new PermissionDeniedException(String.format("Unable to update account role to %s as you are " +
+                    "attempting to escalate the account %s to account type %s which has higher privileges", role.getName(), account.getAccountName(), role.getRoleType().getAccountType().name()));
+        }
+        return true;
+    }
+
     /**
      * if there is any permission under the requested role that is not permitted for the caller, refuse
      */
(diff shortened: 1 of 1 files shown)


Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan test

Pearl1594:
@blueorangutan test

weizhouapache (approved):
code lgtm

rohityadavcloud:
Prior to the roles feature, was updating of account type supported? The roles feature allowed admins to map a set of allowed/denies apis/rules for an account; maybe we should restrict changing roles for an account matching the role/account type? (it depends if changing the type of an account is an allowed feature - this must be tested for both domain admin and root admin cases, i.e. anybody/whoever can create or update accounts)

rohityadavcloud:
cc @Pearl1594 pl see my comment above.

Pearl1594:
@rohityadavcloud I don't think there was support to change account type - However, I am not sure if that was left out as mere oversight or by design. However, if we do allow updating user role, wouldn't it be required that the account type be mapped to the corresponding role's account type - for it to function correctly. In terms of security, the rules associated with the role would govern what the user in concern is allowed (or not allowed) to access.

rohityadavcloud:
@Pearl1594 I think the update API shouldn't be allowed to switch an account to a role that is incompatible with the account type. For example, if it's allowed to update a normal user account's role to a root admin role, that shouldn't be allowed. That means the UI/client should filter and allow only roles of matching types, and/or in the backend this must be checked. (this is assuming that updateAccount doesn't explicitly allows updating the account type)

Pearl1594:
I understood @rohityadavcloud . So, would you say that an account's role can be changed to another role with the same account type and with that of lower privileges or we do not want to allow the latter case as well?

rohityadavcloud:
Yes @Pearl1594 both UI should filter and show roles matching account type wrt role type. I think otherwise, we have a situation where a normal user account can have root admin role ACLs. We should put check to guard against any potential privilege escalation. I think there was some work put in by @DaanHoogland for add account API - so maybe investigate if it's already there and can be (re)used.

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan package

Pearl1594:
@blueorangutan test

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -1897,7 +1897,17 @@ public AccountVO updateAccount(UpdateAccountCmd cmd) {
                         "in the domain '" + domainId + "'.");
             }
 
+            Role role = roleService.findRole(roleId);
+            Long currentAccRoleId = account.getRoleId();
+            Role currentRole = roleService.findRole(currentAccRoleId);
+
+            if (role.getRoleType().ordinal() < currentRole.getRoleType().ordinal() && ((account.getType() == Account.Type.NORMAL && role.getRoleType().getAccountType().ordinal() > Account.Type.NORMAL.ordinal()) ||
+                    account.getType().ordinal() > Account.Type.NORMAL.ordinal() && role.getRoleType().getAccountType().ordinal() < account.getType().ordinal() && role.getRoleType().getAccountType().ordinal() > 0)) {
```
DaanHoogland, line 1905: This condition is hard to read,
Can you extract it to a method like `isRoleEscalation(...)` and indent according to the logic of it, please?
It would make sense to extract the entire block as well, i.e. `checkForRoleEscalation(...)`.

DaanHoogland (changes requested):
I think the logic looks ok, but it is a bit hard to read.

Pearl1594:
@blueorangutan ui

Pearl1594:
@blueorangutan package

nvazquez:
@blueorangutan test

nvazquez (approved):

PR #6164, opened 2022-03-25
Mount disabled storage pool on host reboot

ravening:
### Description
Add a global setting so that disabled pools will be mounted
again on host reboot

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

### How Has This Been Tested?

1. Disable the storage pool
2. Set the global setting `mount.disabled.storage.pool` to true
3. The pool will still be mounted in the host.
4. Reboot the host.
5. The pool will still be mounted
6. Change the setting to false
7. Reboot the host
8. The disabled pool will not be mounted

### DIFF
--- a/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
+++ b/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
@@ -159,9 +159,9 @@ public interface StorageManager extends StorageService {
             "mount.disabled.storage.pool",
             "Storage",
             "false",
-            "Mount all disabled storage pools after node reboot",
+            "Mount all zone-wide or cluster-wide disabled storage pools after node reboot",
             true,
-            ConfigKey.Scope.Zone,
+            ConfigKey.Scope.Cluster,
             null);
 
     /**
--- a/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
+++ b/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
@@ -155,6 +155,15 @@ public interface StorageManager extends StorageService {
     ConfigKey<String> PreferredStoragePool = new ConfigKey<String>(String.class, "preferred.storage.pool", "Advanced", "",
             "The UUID of preferred storage pool for allocation.", true, ConfigKey.Scope.Account, null);
 
+    ConfigKey<Boolean> MountDisabledStoragePool = new ConfigKey<>(Boolean.class,
+            "mount.disabled.storage.pool",
+            "Storage",
+            "false",
+            "Mount disabled storage pool after node reboot",
+            true,
+            ConfigKey.Scope.Cluster,
+            null);
+
     /**
      * Returns a comma separated list of tags for the specified storage pool
      * @param poolId
--- a/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
+++ b/engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
@@ -159,9 +159,9 @@ public interface StorageManager extends StorageService {
             "mount.disabled.storage.pool",
             "Storage",
             "false",
-            "Mount disabled storage pool after node reboot",
+            "Mount all disabled storage pools after node reboot",
             true,
-            ConfigKey.Scope.Cluster,
+            ConfigKey.Scope.Zone,
             null);
 
     /**
--- a/server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
+++ b/server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
@@ -45,7 +45,6 @@ import com.cloud.storage.ScopeType;
 import com.cloud.storage.Storage.StoragePoolType;
 import com.cloud.storage.StorageManagerImpl;
 import com.cloud.storage.StoragePoolHostVO;
-import com.cloud.storage.StoragePoolStatus;
 
 public class StoragePoolMonitor implements Listener {
     private static final Logger s_logger = Logger.getLogger(StoragePoolMonitor.class);
--- a/server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
+++ b/server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
@@ -108,9 +108,14 @@ public class StoragePoolMonitor implements Listener {
                 List<StoragePoolVO> zoneStoragePoolsByAnyHypervisor = _poolDao.findZoneWideStoragePoolsByHypervisor(host.getDataCenterId(), HypervisorType.Any);
                 pools.addAll(zoneStoragePoolsByAnyHypervisor);
 
-                // get the disabled pools list if global setting is true.
+                // get the zone wide disabled pools list if global setting is true.
                 if (StorageManager.MountDisabledStoragePool.value()) {
                     pools.addAll(_poolDao.findDisabledPoolsByScope(host.getDataCenterId(), null, null, ScopeType.ZONE));
+                }
+
+                // get the cluster wide disabled pool list
+                if (StorageManager.MountDisabledStoragePool.valueIn(host.getClusterId())) {
+                    pools.addAll(_poolDao.findDisabledPoolsByScope(host.getDataCenterId(), host.getPodId(), host.getClusterId(), ScopeType.CLUSTER));
                 }
 
                 for (StoragePoolVO pool : pools) {
--- a/server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
+++ b/server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
@@ -20,6 +20,7 @@ import java.util.List;
 
 import javax.inject.Inject;
 
+import com.cloud.storage.StorageManager;
 import org.apache.cloudstack.engine.subsystem.api.storage.DataStoreProvider;
 import org.apache.cloudstack.engine.subsystem.api.storage.DataStoreProviderManager;
 import org.apache.cloudstack.engine.subsystem.api.storage.HypervisorHostListener;
@@ -108,10 +109,12 @@ public class StoragePoolMonitor implements Listener {
                 List<StoragePoolVO> zoneStoragePoolsByAnyHypervisor = _poolDao.findZoneWideStoragePoolsByHypervisor(host.getDataCenterId(), HypervisorType.Any);
                 pools.addAll(zoneStoragePoolsByAnyHypervisor);
 
+                // get the disabled pools list if global setting is true.
+                if (StorageManager.MountDisabledStoragePool.value()) {
+                    pools.addAll(_poolDao.findDisabledPoolsByScope(host.getDataCenterId(), null, null, ScopeType.ZONE));
+                }
+
                 for (StoragePoolVO pool : pools) {
-                    if (pool.getStatus() != StoragePoolStatus.Up) {
-                        continue;
-                    }
                     if (!pool.isShared()) {
                         continue;
                     }
--- a/server/src/main/java/com/cloud/storage/StorageManagerImpl.java
+++ b/server/src/main/java/com/cloud/storage/StorageManagerImpl.java
@@ -3340,7 +3340,8 @@ public class StorageManagerImpl extends ManagerBase implements StorageManager, C
                 MaxDataMigrationWaitTime,
                 DiskProvisioningStrictness,
                 PreferredStoragePool,
-                SecStorageVMAutoScaleDown
+                SecStorageVMAutoScaleDown,
+                MountDisabledStoragePool
         };
     }
 


nvazquez:
@blueorangutan package

engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
```diff
@@ -155,6 +155,15 @@
     ConfigKey<String> PreferredStoragePool = new ConfigKey<String>(String.class, "preferred.storage.pool", "Advanced", "",
             "The UUID of preferred storage pool for allocation.", true, ConfigKey.Scope.Account, null);
 
+    ConfigKey<Boolean> MountDisabledStoragePool = new ConfigKey<>(Boolean.class,
+            "mount.disabled.storage.pool",
+            "Storage",
+            "false",
+            "Mount disabled storage pool after node reboot",
+            true,
+            ConfigKey.Scope.Cluster,
```
nvazquez, line 164: Shouldn't this scope be set to StoragePool? If not, please update the configuration description mentioning that it impacts all the storage pools in the cluster

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
```diff
@@ -108,10 +108,12 @@ public void processConnect(Host host, StartupCommand cmd, boolean forRebalance)
                 List<StoragePoolVO> zoneStoragePoolsByAnyHypervisor = _poolDao.findZoneWideStoragePoolsByHypervisor(host.getDataCenterId(), HypervisorType.Any);
                 pools.addAll(zoneStoragePoolsByAnyHypervisor);
 
+                // get the disabled pools list if global setting is true.
+                if (StorageManager.MountDisabledStoragePool.value()) {
+                    pools.addAll(_poolDao.findDisabledPoolsByScope(host.getDataCenterId(), null, null, ScopeType.ZONE));
```
nvazquez, line 113: Cluster-wide storage pools are not considered?

engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
ravening: This should be zone wide as vm can be migrated across cluster. I will update the description

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
ravening: no. only zone wide is considered as a single pool should be available for all clusters

nvazquez:
@blueorangutan package

engine/components-api/src/main/java/com/cloud/storage/StorageManager.java
```diff
@@ -155,6 +155,15 @@
     ConfigKey<String> PreferredStoragePool = new ConfigKey<String>(String.class, "preferred.storage.pool", "Advanced", "",
             "The UUID of preferred storage pool for allocation.", true, ConfigKey.Scope.Account, null);
 
+    ConfigKey<Boolean> MountDisabledStoragePool = new ConfigKey<>(Boolean.class,
+            "mount.disabled.storage.pool",
+            "Storage",
+            "false",
+            "Mount all disabled storage pools after node reboot",
```
nvazquez, line 162:
```suggestion
            "Mount all disabled zode-wide storage pools after node reboot",
```

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
```diff
@@ -108,10 +108,12 @@ public void processConnect(Host host, StartupCommand cmd, boolean forRebalance)
...
+                }
+
                 for (StoragePoolVO pool : pools) {
-                    if (pool.getStatus() != StoragePoolStatus.Up) {
```
nvazquez, line 112: Maybe can revert this and extend the condition to `if (pool.getStatus() != StoragePoolStatus.Up && pool.getStatus() != StoragePoolStatus.Disabled)` ?

nvazquez (commented):
Thanks @ravening have left some minor comments after the refactor

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
weizhouapache, line 113: @ravening 
how about cluster wide storage pools ?

nvazquez: @weizhouapache same comment from me here :) https://github.com/apache/cloudstack/pull/6164/files#r835471732

weizhouapache: ah, ok thanks @nvazquez 

@ravening any reason why cluster-wide pools are not supported ? can you extend this PR to support cluster-wide pools ?

ravening: @ravening since VM's can be migrated across clusters, I want to support only zone wide storage pools. If disabled pool is only mounted in cluster A but if we want to migrate vm to cluster B then its not possible.

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
ravening: all of the above functions are only returning pools in Up state. so this check is not needed

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
weizhouapache: @ravening 
ok. I am not quite clear what you want to fix/achieve. can you explain a bit more ?

anyway, the support of cluster-wide pools is an improvement of this pr. it means not only zone-wide pools , but also cluster-wide pools are mounted. any issue ?

ravening: @weizhouapache when we set the status of the storage pool as disabled, it will not be used for vm allocation but it will be still mounted/accessible for the vm's. When the node reboots, all the disabled pools will not be mounted and as a result the volume of the vm will not be accessible.

My understanding of the cluster wide pool is that it will be mounted on the hosts within the same cluster right? The same pool will not be mounted on the nodes of different cluster.


Lets say we have three clusters. A, B and C.
I want the disabled pool to be mounted on all three clusters rather than just on A or B or C. Thats why I am searching for zone wide pools

weizhouapache: thanks @ravening 

if node reboots, you cannot acess the disabled pools (and the volumes on the pools) because they are not mounted on node. is it the issue you want to fix ?

zone-wide pools will be mounted on all hosts in the zone. cluster-wide pools will be mounted on all hosts in specific cluster. I understand you use only zone-wide pools, but many other users use cluster-wide pools. why not mount cluster-wide pools as well ?

ravening: @weizhouapache  you are saying that, I need to make an extra call like this `pools.addAll(_poolDao.findDisabledPoolsByScope(host.getDataCenterId(), host.getClusterId(), null, ScopeType.CLUSTER));` ?

weizhouapache: > @weizhouapache you are saying that, I need to make an extra call like this `pools.addAll(_poolDao.findDisabledPoolsByScope(host.getDataCenterId(), host.getClusterId(), null, ScopeType.CLUSTER));` ?

@ravening 
yes exactly. it would be good to have it.
please also consider change it to a cluster level setting ?

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
nvazquez: Actually not only Up state since you have added the Disabled pools as well, actually my suggestion above is not correct, should be:

````
if (pool.getStatus() != StoragePoolStatus.Up || (StorageManager.MountDisabledStoragePool.value() && pool.getStatus() != StoragePoolStatus.Disabled && pool.getStatus() != StoragePoolStatus.Up)) {
    continue;
}
````

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
ravening: @weizhouapache @nvazquez added cluster wide support. please review again

server/src/main/java/com/cloud/storage/listener/StoragePoolMonitor.java
weizhouapache: @nvazquez 
this might not be needed, because `findZoneWideStoragePoolsByHypervisor` has the condition `sc.and(sc.entity().getStatus(), Op.EQ, Status.Up);`

weizhouapache:
@blueorangutan package

nvazquez (approved):
Code LGTM

weizhouapache:
@blueorangutan test

weizhouapache (approved):
code lgtm

wido:
I noticed this PR a bit late. Great PR and good catch, however.... (There is always a however!)

The API: https://cloudstack.apache.org/api/apidocs-4.16/apis/updateStoragePool.html

<pre>
enabled | false to disable the pool for allocation of new volumes, true to enable it back.
</pre>

The **enabled** flag only tells if the storage pool should be considered for the allocation of new volumes. It is not intended to be use for not being mounted on hosts. Thats where we have the **maintenance** mode of a storage pool for.

The global setting (mount.disabled.storage.pool defaults to false) which we introduce actually keeps the bug in tact but we can influence the bug by manually changing a config setting.

I would say that we should not add this config value and always 'mount' disabled storage pools as that's how it was designed.

nvazquez:
Hi @wido thanks for your input - @ravening @weizhouapache what's your opinion?

ravening:
@wido @nvazquez Yes we always wanted the disabled pools to be mounted by default even after the node reboot. I didnt want to change the existing behaviour as it might break the functionality for others. Thats the reason, I added the global setting.

If everyone agrees that the disabled pools should be mounted by default then I can remove the global setting and always try to mount it

wido:
@ravening and @nvazquez The API docs are clear. Disabling a pool only avoids it for allocation of new volumes. If you don't want the pool to be mounted you can put it in maintenance mode.

The bug right now is that a disabled storage pool is not mounted and thus prevents VMs from being started on a hypervisor/host after it was rebooted. That's a true bug.

I would say we need to get rid of the global setting and stick to the workings intended and also documented in the API.

ravening:
> @ravening and @nvazquez The API docs are clear. Disabling a pool only avoids it for allocation of new volumes. If you don't want the pool to be mounted you can put it in maintenance mode.
> 
> 
> 
> The bug right now is that a disabled storage pool is not mounted and thus prevents VMs from being started on a hypervisor/host after it was rebooted. That's a true bug.
> 
> 
> 
> I would say we need to get rid of the global setting and stick to the workings intended and also documented in the API.

@wido we learnt in a hard way that api docs are always not reliable. So I didn't want to change any code without altering the behavior existing in master

I know it's actually a bug but as mentioned earlier, if I get a majority vote to remove global setting then I will do so

wido:
> > @ravening and @nvazquez The API docs are clear. Disabling a pool only avoids it for allocation of new volumes. If you don't want the pool to be mounted you can put it in maintenance mode.
> > The bug right now is that a disabled storage pool is not mounted and thus prevents VMs from being started on a hypervisor/host after it was rebooted. That's a true bug.
> > I would say we need to get rid of the global setting and stick to the workings intended and also documented in the API.
> 
> @wido we learnt in a hard way that api docs are always not reliable. So I didn't want to change any code without altering the behavior existing in master
> 
> I know it's actually a bug but as mentioned earlier, if I get a majority vote to remove global setting then I will do so

We actually learned this the hard way earlier this week. An operator disabled a (Ceph) storage pool as we wanted new volumes to be allocated on a different Ceph cluster. After a hypervisor reboot VMs were stuck in the 'Starting' state and it took a while to figure out how and what.

If we keep the current behavior then what's the difference between Disabled and Maintenance?

ravening:
@weizhouapache @nvazquez @rohityadavcloud @DaanHoogland  are you guys ok with removing the global setting and making the default behaviour as mounting the disabled pools all the time?

rohityadavcloud:
I would want to keep the global setting but keep the default behaviour to be what is either generally accepted or what was previously happening (i.e. for backward compatibility). Since the PR has been merged @ravening do you mind raising a new PR (if necessary?) or at least start a thread on dev@ as many people may miss your msg.

PR #6165, opened 2022-03-25
SAML: replace first number with random alphabet if request ID starts with a number

### LINKED ISSUE #5548: SAML2 authentication with MS Azure AD intermittently fails with error AADSTS7500529
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
SAML/SSO
~~~

##### CLOUDSTACK VERSION

~~~
4.15.2
~~~

##### CONFIGURATION
Enabled SAML plugin and configured it to authenticate against Azure AD.

##### STEPS TO REPRODUCE
SAML authentication fails intermittently depending on the SAML request ID that is generated. Depending on your luck, you might get a random request ID that begins with something between A-Z. In such cases, the authentication request goes through as expected. On the chance that the generated number begins with 0-9, the authentication request fails with an error similar to:
> AADSTS7500529: The value '692rv91k6dgmdas33vr3b2keahr4lqjv' is not a valid SAML ID. The ID must not begin with a number.

As per Azure AD's documentation, an acceptable request ID must not start with a number. 
> ID must not begin with a number, so a common strategy is to prepend a string like "id" to the string representation of a GUID. For example, id6c1c178c166d486687be4aaf5e482730 is a valid ID.
> https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/active-directory/develop/single-sign-on-saml-protocol.md

The CloudStack implementation generates the random ID string with the following method in SAMLUtils and has no ability to prefix the SAML ID string as recommended by Azure AD's documentation.
```
100 public class SAMLUtils {
101     public static final Logger s_logger = Logger.getLogger(SAMLUtils.class);
102
103     public static String generateSecureRandomId() {
104         return new BigInteger(160, new SecureRandom()).toString(32);
105     }
```

##### EXPECTED RESULTS
Successful authentication with SAML and Azure AD.

##### ACTUAL RESULTS
Intermittent authentication failure with SAML and Azure AD.

weizhouapache:
### Description

This PR fixes #5548 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
+++ b/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
@@ -100,8 +100,11 @@ import com.cloud.utils.HttpUtils;
 public class SAMLUtils {
     public static final Logger s_logger = Logger.getLogger(SAMLUtils.class);
 
+    static final String charset = "abcdefghijklmnopqrstuvwxyz";
+
     public static String generateSecureRandomId() {
-        return new BigInteger(160, new SecureRandom()).toString(32);
+        return new BigInteger(160, new SecureRandom()).toString(32).replaceFirst("^[0-9]",
+                String.valueOf(charset.charAt(new SecureRandom().nextInt(charset.length()))));
     }
 
     public static String getValueFromAttributeStatements(final List<AttributeStatement> attributeStatements, final String attributeKey) {
--- a/plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
+++ b/plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
@@ -22,6 +22,7 @@ package org.apache.cloudstack;
 import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.security.PublicKey;
+import java.util.regex.Pattern;
 
 import org.apache.cloudstack.saml.SAMLUtils;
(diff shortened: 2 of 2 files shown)


nvazquez:
@blueorangutan package

plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
```diff
@@ -100,8 +100,11 @@
 public class SAMLUtils {
     public static final Logger s_logger = Logger.getLogger(SAMLUtils.class);
 
+    static final String charset = "abcdefghijklmnopqrstuvwxyz";
```
rohityadavcloud, line 103: should we also have A-Z (i.e. a-z in upper case too) @weizhouapache ?

nvazquez:
@blueorangutan test

plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
weizhouapache: @rohityadavcloud 
I have checked hundreds of request IDs, all are composed by digits (0 to 9) or lower case alphabet (a to z).

rohityadavcloud (approved):

rohityadavcloud:
@blueorangutan test

nvazquez (approved):
LGTM

kohrar:
That looks great. Thank-you for implementing the fix, @weizhouapache.

PR #6175, opened 2022-03-28
KVM: Enhance CPU speed detection on hosts

nvazquez:
### Description

This PR does two things:

1. In a recent commit the way host CPU speed is detected was changed to use `lscpu`. This is a great enhancement, but it changed the detected speed from being the host's max turbo frequency to being the base clock. I agree this is the ideal speed to use since the max turbo frequency usually is not achieved by all cores at once. To ensure the fallback of using `cpufreq` files matches this behavior, we change the fallback to look at the `base_frequency` file rather than `cpuinfo_max_freq`.

2. Introduce agent.properties setting `host.cpu.manual.speed.mhz` to allow the admin to set the frequency in case all else fails.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

### DIFF
--- a/agent/conf/agent.properties
+++ b/agent/conf/agent.properties
@@ -285,3 +285,6 @@ iscsi.session.cleanup.enabled=false
 
 # Enable manually setting CPU's topology on KVM's VM.
 # enable.manually.setting.cpu.topology.on.kvm.vm=true
+
+# host.cpu.manual.speed.mhz = int
+# Manually set the host CPU MHz, in cases where CPU scaling support detected value is wrong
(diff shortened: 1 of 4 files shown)


weizhouapache (approved):
tested ok

Pearl1594:
@blueorangutan package

weizhouapache:
@blueorangutan test

weizhouapache:

agent/conf/agent.properties
```diff
@@ -285,3 +285,6 @@ iscsi.session.cleanup.enabled=false
 
 # Enable manually setting CPU's topology on KVM's VM.
 # enable.manually.setting.cpu.topology.on.kvm.vm=true
+
+# host.cpu.manual.speed.mhz = int
+# Manually set the host CPU MHz, in cases where CPU scaling support detected value is wrong
```
DaanHoogland, line 290: should we adhere to the convention of putting the explaining comment before the commented default value?

DaanHoogland:
```suggestion
# Manually set the host CPU MHz, in cases where CPU scaling support detected value is wrong
# host.cpu.manual.speed.mhz=0
```

weizhouapache (approved):

DaanHoogland (approved):

PR #6245, opened 2022-04-11
Fix VM stats inconsistencies

JoaoJandre:
### Description

This PR aims to fix various inconsistencies in the front and back-end concerning the VM statistics. I mainly enhanced the descriptions, adding explanations and/or context, but also corrected some errors in the units of measurements (there were cases of metrics being represented in bytes when the correct one should be KiB, for example).
Also, many inconsistencies were tested and found in the way stats were being collected for the KVM hypervisor. Namely, when generating the `VmMetricsResponse`, the network usage was being overwritten with the disk usage, so neither disk nor network usage was correct. Another issue was that memory usage wasn't being properly collected.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
This was tested in a local lab, by creating VMs and comparing their internal stats with the ones gathered by CloudStack.


ui/public/locales/pl.json
```diff
@@ -623,8 +623,8 @@
 "label.diskiopswriterate": "Disk Write Rate (IOPS)",
 "label.diskioread": "Disk Read (IO)",
 "label.diskiowrite": "Disk Write (IO)",
-"label.diskkbsread": "Disk Read (Bytes)",
-"label.diskkbswrite": "Disk Write (Bytes)",
+"label.diskkbsread": "Disk Read",
+"label.diskkbswrite": "Disk Write",
```
DaanHoogland, line 627: should these contain a `(KiB)` as well?

plugins/metrics/src/main/java/org/apache/cloudstack/response/VmMetricsResponse.java
```diff
@@ -36,7 +36,7 @@ public class VmMetricsResponse extends UserVmResponse {
     private String cpuTotal;
 
     @SerializedName("memorytotal")
-    @Param(description = "the total memory capacity in GiB")
+    @Param(description = "the total memory capacity in GB")
```
DaanHoogland, line 39: is this `1000^3` and not `1024^3`?

DaanHoogland (approved):

ui/public/locales/pl.json
JoaoJandre: They shouldn't, because these labels are used as column titles in instances tab when the metrics toggle is on. And to maintain the same pattern with the other columns, the units of measurements should be alongside the values, not the title.

DaanHoogland: @JoaoJandre but they are included in the other languages for `label.diskkbsread` and `label.diskkbswrite`, and not in the polish json. Why is that?

rohityadavcloud (approved):

rohityadavcloud:
@blueorangutan package

nvazquez:
@blueorangutan test

weizhouapache (approved):
code lgtm

PR #6357, opened 2022-05-04
Add logs to the listLoadBalancerRuleInstances API

joseflauzino:
### Description

Currently, the listLoadBalancerRuleInstances API has no logs, which makes troubleshooting harder.
This PR aims to add several logs with context on the processing performed by this API. This will allow cloud operators to better analyze possible errors or inconsistencies that may occur when using the API.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

I performed different calls to the listLoadBalancerRuleInstances API and could notice that the processes were logged correctly.


server/src/main/java/com/cloud/network/lb/LoadBalancingRulesManagerImpl.java
```diff
@@ -2231,11 +2235,21 @@ public Pair<List<? extends UserVm>, List<String>> listLoadBalancerInstances(List
                 continue;
             }
 
+            String userVmAsString = ReflectionToStringBuilderUtils.reflectOnlySelectedFields(userVm, "uuid", "name");
```
GutoVeronezi, line 2238: As `UserVmVO` extends `VMInstanceVO` and `VMInstanceVO` has an implementation of `toString`, we could use it instead of the `ReflectionToStringBuilderUtils.reflectOnlySelectedFields`.

server/src/main/java/com/cloud/network/lb/LoadBalancingRulesManagerImpl.java
```diff
@@ -2189,13 +2190,16 @@ public Pair<List<? extends UserVm>, List<String>> listLoadBalancerInstances(List
...
+            s_logger.debug(String.format("The [%s] parameter was not passed. Using the default value [%s].", ApiConstants.APPLIED, Boolean.TRUE));
             applied = Boolean.TRUE;
         }
 
         LoadBalancerVO loadBalancer = _lbDao.findById(loadBalancerId);
         if (loadBalancer == null) {
+            s_logger.warn(String.format("Unable to find the load balancer with ID [%s].", cmd.getId()));
             return null;
         }
+        String loadBalancerAsString = ReflectionToStringBuilderUtils.reflectOnlySelectedFields(loadBalancer, "uuid", "name");
```
GutoVeronezi, line 2202: We could implement `toString` of `LoadBalancerVO`. This way we can call `toString` instead of repeating the same code wherever it is used.

joseflauzino:
@blueorangutan package

api/src/main/java/org/apache/cloudstack/api/command/user/loadbalancer/ListLoadBalancerRuleInstancesCmd.java
```diff
@@ -97,7 +97,10 @@ public String getCommandName() {
     public void execute() {
         Pair<List<? extends UserVm>, List<String>> vmServiceMap =  _lbService.listLoadBalancerInstances(this);
         List<? extends UserVm> result = vmServiceMap.first();
+        s_logger.debug(String.format("A total of [%s] user VMs were obtained when listing the load balancer instances: [%s].", result.size(), result.toString()));
```
DaanHoogland, line 100:
```suggestion
        if (s_logger.isDebugEnabled()) {
            s_logger.debug(String.format("A total of [%s] user VMs were obtained when listing the load balancer instances: [%s].", result.size(), result.toString()));
        }
```

api/src/main/java/org/apache/cloudstack/api/command/user/loadbalancer/ListLoadBalancerRuleInstancesCmd.java
```diff
@@ -97,7 +97,10 @@ public String getCommandName() {
...
+
         List<String> serviceStates  = vmServiceMap.second();
+        s_logger.debug(String.format("A total of [%s] service states were obtained when listing the load balancer instances: [%s].", serviceStates.size(), serviceStates.toString()));
```
DaanHoogland, line 103:
```suggestion
        if (s_logger.isDebugEnabled()) {
            s_logger.debug(String.format("A total of [%s] service states were obtained when listing the load balancer instances: [%s].", serviceStates.size(), serviceStates.toString()));
        }
```

server/src/main/java/com/cloud/network/lb/LoadBalancingRulesManagerImpl.java
```diff
@@ -2189,13 +2189,16 @@ public Pair<List<? extends UserVm>, List<String>> listLoadBalancerInstances(List
         Boolean applied = cmd.isApplied();
 
         if (applied == null) {
+            s_logger.debug(String.format("The [%s] parameter was not passed. Using the default value [%s].", ApiConstants.APPLIED, Boolean.TRUE));
```
DaanHoogland, line 2192:
```suggestion
            if (s_logger.isDebugEnabled()) {
                s_logger.debug(String.format("The [%s] parameter was not passed. Using the default value [%s].", ApiConstants.APPLIED, Boolean.TRUE));
            }
```

server/src/main/java/com/cloud/network/lb/LoadBalancingRulesManagerImpl.java
```diff
@@ -2231,11 +2234,21 @@ public Pair<List<? extends UserVm>, List<String>> listLoadBalancerInstances(List
...
             }
 
+            String userVmAsString = userVm.toString();
+
             boolean isApplied = appliedInstanceIdList.contains(userVm.getId());
-            if ((isApplied && applied) || (!isApplied && !applied)) {
-                loadBalancerInstances.add(userVm);
-                serviceStates.add(vmServiceState.get(userVm.getId()));
+            String isAppliedMsg = isApplied ? "is applied" : "is not applied";
+            s_logger.debug(String.format("The user VM [%s] %s to a rule of the load balancer [%s].", userVmAsString, isAppliedMsg, loadBalancerAsString));
```
DaanHoogland, line 2241:
```suggestion
            if (s_logger.isDebugEnabled()) {
                s_logger.debug(String.format("The user VM [%s] %s to a rule of the load balancer [%s].", userVmAsString, isAppliedMsg, loadBalancerAsString));
            }
```

server/src/main/java/com/cloud/network/lb/LoadBalancingRulesManagerImpl.java
```diff
@@ -2231,11 +2234,21 @@ public Pair<List<? extends UserVm>, List<String>> listLoadBalancerInstances(List
...
+
+            if (isApplied != applied) {
+                s_logger.debug(String.format("Skipping add service state from the user VM [%s] to the service state list. This happens because the VM %s to the load "
+                        + "balancer rule and the [%s] parameter was passed as [%s].", userVmAsString, isAppliedMsg, ApiConstants.APPLIED, applied));
```
DaanHoogland, line 2245:
```suggestion
                if (s_logger.isDebugEnabled()) {
                    s_logger.debug(String.format("Skipping add service state from the user VM [%s] to the service state list. This happens because the VM %s to the load "
                        + "balancer rule and the [%s] parameter was passed as [%s].", userVmAsString, isAppliedMsg, ApiConstants.APPLIED, applied));
               }
```

server/src/main/java/com/cloud/network/lb/LoadBalancingRulesManagerImpl.java
```diff
@@ -2231,11 +2234,21 @@ public Pair<List<? extends UserVm>, List<String>> listLoadBalancerInstances(List
...
+                continue;
             }
+            loadBalancerInstances.add(userVm);
+            String serviceState = vmServiceState.get(userVm.getId());
+            s_logger.debug(String.format("Add the service state [%s] from the user VM [%s] to the service state list.", serviceState, userVmAsString));
```
DaanHoogland, line 2250:
```suggestion
            if (s_logger.isDebugEnabled()) {
                s_logger.debug(String.format("Add the service state [%s] from the user VM [%s] to the service state list.", serviceState, userVmAsString));
            }
```

DaanHoogland (commented):
Aren´t some of those debug statements supposed to be trace-level instead?

joseflauzino:
@blueorangutan package

joseflauzino:
Hi @DaanHoogland, sorry for the delay in responding.

Regarding your suggested code changes:
In these particular code snippets, I have a preference for not using the `isDebugEnabled()` check.

The main reason is that this check can provide a significant performance improvement only when the message construction is very costly, which I don't think is the case. In these snippets, the messages are built by the concatenation of strings (no need for expensive calculations or processing).

Also, if the logger is enabled at the debug level, it incurs the cost of evaluating whether the logger is enabled or not twice: once in `isDebugEnabled()` and once in `debug()`. Since it is very common for ACS users to leave the logger enabled for the debug level  (because of the useful messages for troubleshoting), I don't think it is worth doing this check (especially with simple strings).

DaanHoogland:
@joseflauzino , I don´t think your argument holds, in the long run we should avoid debug statements in production environments. If you think your logging is valid in production you should give it a higher level.
@blueorangutan test

engine/schema/src/main/java/com/cloud/network/dao/LoadBalancerVO.java
```diff
@@ -136,4 +136,9 @@ public Scheme getScheme() {
     public String getCidrList() {
         return cidrList;
     }
+
+    @Override
+    public String toString() {
+        return String.format("Load Balancer {uuid: \"%s\", name: \"%s\"}", getUuid(), name);
```
GutoVeronezi, line 142: We can use `ReflectionToStringBuilderUtils` here.

GutoVeronezi (approved):
CLGTM, just pointed an adjust.

engine/schema/src/main/java/com/cloud/network/dao/LoadBalancerVO.java
joseflauzino: @GutoVeronezi 
I think `ReflectionToStringBuilderUtils` will not work properly here, since `LoadBalancerVO` doesn't have the `uuid` attribute (it's inherited from `FirewallRuleVO`). So I've used `String.format()` getting the `uuid` from the getter method.

DaanHoogland:
> @joseflauzino , I don´t think your argument holds, in the long run we should avoid debug statements in production environments. If you think your logging is valid in production you should give it a higher level.

@joseflauzino please review your loging statements and make sure anything you want in present in production is at least at INFO level. You are right that this is not perfectly implemented this way, but that does not mean we shouldn´t improve on it.

joseflauzino:
@DaanHoogland 
Most of the log messages added/improved in this PR are really useful for debugging purposes. There are only two exceptions, which are presented as warning and error. Considering the characteristics of all the messages involved, I think the log levels are appropriate.

engine/schema/src/main/java/com/cloud/network/dao/LoadBalancerVO.java
GutoVeronezi: `ReflectionToStringBuilderUtils` uses `ReflectionUtils.getAllFields`, which will also retrieve fields of the super types too; therefore it will work properly.

RodrigoDLopez (approved):

DaanHoogland:
ok @joseflauzino but where ever there is a string operation done in the parameter passing I would still like the `if (log.isDebugEnabled())` guard around it. Other wise fine.

engine/schema/src/main/java/com/cloud/network/dao/LoadBalancerVO.java
joseflauzino: @GutoVeronezi done!

joseflauzino:
@blueorangutan package

api/src/main/java/org/apache/cloudstack/api/command/user/loadbalancer/ListLoadBalancerRuleInstancesCmd.java
DaanHoogland, line 100:
```suggestion
        if (result != null && s_logger.isDebugEnabled()) {
            s_logger.debug(String.format("A total of [%s] user VMs were obtained when listing the load balancer instances: [%s].", result.size(), result.toString()));
        }
```

DaanHoogland (changes requested):

GutoVeronezi:
@joseflauzino, do you plan on continuing this PR? Is it still necessary?

joseflauzino:
Hi @GutoVeronezi! Sorry for the delay.
These logs are very useful when troubleshooting problems related to the mentioned API.
As we discussed in another channel, it might be better to implement this kind of log using log4j2 to avoid adding conditional structures.
You also mentioned that you could add these logs to another PR that is upgrading log4j to log4j2 (or something like that).

Please feel absolutely free to do so. I'd really appreciate it! :)

shwstppr:
@joseflauzino can you please address the review comments?

DaanHoogland:
@joseflauzino are you going to address any of my comments or apply any of my suggestions?

GutoVeronezi:
@DaanHoogland @shwstppr I talked with @joseflauzino offline and I'll handle this PR later; not necessary for 4.19.

GutoVeronezi:
@DaanHoogland @shwstppr @joseflauzino I will close this PR as its changes is being addressed along with some other refactoring in #8094

PR #6371, opened 2022-05-10
agent: enable ssl only for kvm agent (not in system vms)

weizhouapache:
### Description

This PR fixes the issue that template cannot be downloaded from https site in 4.17.0.0 RC2.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [x] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/agent/src/main/java/com/cloud/agent/AgentShell.java
+++ b/agent/src/main/java/com/cloud/agent/AgentShell.java
@@ -27,7 +27,6 @@ import com.cloud.utils.PropertiesUtil;
 import com.cloud.utils.backoff.BackoffAlgorithm;
 import com.cloud.utils.backoff.impl.ConstantTimeBackoff;
 import com.cloud.utils.exception.CloudRuntimeException;
-import org.apache.cloudstack.utils.security.KeyStoreUtils;
 import org.apache.commons.daemon.Daemon;
 import org.apache.commons.daemon.DaemonContext;
 import org.apache.commons.daemon.DaemonInitException;
@@ -375,7 +374,6 @@ public class AgentShell implements IAgentShell, Daemon {
 
         loadProperties();
         parseCommand(args);
-        enableSSL();
 
         if (s_logger.isDebugEnabled()) {
             List<String> properties = Collections.list((Enumeration<String>)_properties.propertyNames());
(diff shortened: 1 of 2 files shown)


weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan package

nvazquez:
@blueorangutan test

agent/src/main/java/com/cloud/agent/AgentShell.java
```diff
@@ -399,12 +401,17 @@ public void init(String[] args) throws ConfigurationException {
...
         final File agentFile = PropertiesUtil.findConfigFile("agent.properties");
         if (agentFile == null) {
             s_logger.info("Failed to find agent.properties file");
             return;
         }
+        final String resource = getProperty(null, "resource");
+        if (!LIBVIRT_COMPUTING_RESOURCE.equalsIgnoreCase(resource)) {
+            s_logger.info("This is not a cloudstack kvm agent, ignoring");
+            return;
+        }
```
rohityadavcloud, line 414: @nvazquez @weizhouapache Could this cause an issue with (a) direct download, (b) ssl enabled cpvm and ssvm. Should we attempt to fix the issue by importing or referencing the system keystore too where java/ca certs are stored and imported (for x1/letsencrypt) @Pearl1594 ?

weizhouapache: @rohityadavcloud @nvazquez @Pearl1594 
this part of code was introduced in #6200 which aims to fix an issue with direct download on KVM.
with this pr, SSL is still enabled on KVM hosts, but not in system VMs where Java process load key store real hostile.keystone (same as the behavior without #6200 )

weizhouapache:
@blueorangutan ubuntu20 kvm-ubuntu20

nvazquez (approved):

nvazquez:
Verified HTTPS templates working on KVM env with this fix

agent/src/main/java/com/cloud/agent/AgentShell.java
```diff
@@ -399,12 +401,17 @@ public void init(String[] args) throws ConfigurationException {
...
 
-    private void enableSSL() {
+    private void enableSSLForKvmAgent() {
         final File agentFile = PropertiesUtil.findConfigFile("agent.properties");
         if (agentFile == null) {
             s_logger.info("Failed to find agent.properties file");
             return;
         }
+        final String resource = getProperty(null, "resource");
+        if (!LIBVIRT_COMPUTING_RESOURCE.equalsIgnoreCase(resource)) {
```
rohityadavcloud, line 411: The use of `LIBVIRT_COMPUTING_RESOURCE` or `com.cloud.hypervisor.kvm.resource.LibvirtComputingResource` may be misleading (for ex. what if this isn't the resource name? for example, we do support lxc etc?).

Can you simply add a check if the agent is either is in systemvm (for ex. /etc/cloudstack-release would exist) or if we're on a KVM host (check for /dev/kvm exists or libvirt running)? (Depending on what we're trying to achieve here). The other option usually is to check the resource class via some method.

What do you think @weizhouapache @nvazquez ?

agent/src/main/java/com/cloud/agent/AgentShell.java
rohityadavcloud: Makes sense if this was relatively newer code.

agent/src/main/java/com/cloud/agent/AgentShell.java
weizhouapache: @rohityadavcloud 
I will move all these codes to `LibvirtComputingResource` which makes more sense.
I will verify the issue fixed by #6200 as well.

weizhouapache:
@blueorangutan package

agent/src/main/java/com/cloud/agent/AgentShell.java
nvazquez: +1

nvazquez:
@blueorangutan test

slavkap (approved):
tested - LGTM

Pearl1594 (approved):
LGTM

nvazquez (approved):
LGTM, manually tested

nvazquez:
@blueorangutan test

PR #6398, opened 2022-05-19
github-action: Prevent bo ui build on every PR irrespective of UI changes

Pearl1594:
### Description

This PR attempts to fix the issue noticed during the run of the github action that automatically triggers ui build on a PR. However, irrespective of the fact that the PR has ui changes, it seems to always trigger this action.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

### DIFF
--- a/.github/workflows/coverage-check.yml
+++ b/.github/workflows/coverage-check.yml
@@ -33,8 +33,6 @@ jobs:
         id: ui-changes
         uses: tj-actions/changed-files@a59f800cbb60ed483623848e31be67659a2940f8
         with:
-          sha: "${{ github.pull_request.sha }}"
-          base_sha: "${{ github.sha }}"
           files: |
             ui/*
 


.github/workflows/coverage-check.yml
```diff
@@ -33,8 +33,6 @@ jobs:
         id: ui-changes
         uses: tj-actions/changed-files@a59f800cbb60ed483623848e31be67659a2940f8
         with:
-          sha: "${{ github.pull_request.sha }}"
-          base_sha: "${{ github.sha }}"
```
rohityadavcloud, line 37: Due to the way `pull_request_target` works, when explicit sha, base_sha aren't passed I saw it fails to diff well. Maybe the values aren't right, something else should be passed?

PR #6253, opened 2022-04-12
Extract the IO_URING configuration into the agent.properties

slavkap:
### Description

This PR fixes #6252 
When using advanced virtualization, the IO Driver is not supported. The admins will decide if they want to enable/disable this configuration from `agent.properties` file. It's enabled by default

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

PR #6399, opened 2022-05-19
[KVM] Enable IOURING only when it is available on the host

nvazquez:
### Description

This PR enables IO_URING only when it is available on the hosts, as being discussed on https://github.com/apache/cloudstack/pull/6253#issuecomment-1131216926

### Types of changes

- [x] Major
- [ ] Minor

#### Bug Severity

- [x] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -54,13 +54,6 @@ public class AgentProperties{
      */
     public static final Property<Boolean> ENABLE_MANUALLY_SETTING_CPU_TOPOLOGY_ON_KVM_VM = new Property<Boolean>("enable.manually.setting.cpu.topology.on.kvm.vm", true);
 
-    /**
-     * Enable manually IO driver on KVM's VM. If it is not manually enabled CloudStack can detect if its available<br>
-     * Data type: boolean.<br>
-     * Default value: null.
-     */
-    public static final Property<Boolean> ENABLE_IO_URING = new Property<>("enable.io.uring", null);
-
     public static class Property <T>{
         private final String name;
         private final T defaultValue;
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -57,9 +57,9 @@ public class AgentProperties{
     /**
      * Enable manually IO driver on KVM's VM. <br>
      * Data type: boolean.<br>
-     * Default value: true.
+     * Default value: false.
      */
-    public static final Property<Boolean> ENABLE_IO_URING = new Property<Boolean>("enable.io.uring", true);
+    public static final Property<Boolean> ENABLE_IO_URING = new Property<>("enable.io.uring", false);
 
     public static class Property <T>{
         private final String name;
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -55,11 +55,11 @@ public class AgentProperties{
     public static final Property<Boolean> ENABLE_MANUALLY_SETTING_CPU_TOPOLOGY_ON_KVM_VM = new Property<Boolean>("enable.manually.setting.cpu.topology.on.kvm.vm", true);
 
     /**
-     * Enable manually IO driver on KVM's VM. To be applied it also requires qemu supports io_uring <br>
+     * Enable manually IO driver on KVM's VM. If it is not manually enabled CloudStack can detect if its available<br>
      * Data type: boolean.<br>
-     * Default value: true.
+     * Default value: null.
      */
-    public static final Property<Boolean> ENABLE_IO_URING = new Property<>("enable.io.uring", true);
+    public static final Property<Boolean> ENABLE_IO_URING = new Property<>("enable.io.uring", null);
 
     public static class Property <T>{
         private final String name;
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -54,6 +54,13 @@ public class AgentProperties{
      */
     public static final Property<Boolean> ENABLE_MANUALLY_SETTING_CPU_TOPOLOGY_ON_KVM_VM = new Property<Boolean>("enable.manually.setting.cpu.topology.on.kvm.vm", true);
 
+    /**
+     * Enable manually IO driver on KVM's VM. To be applied it also requires qemu supports io_uring <br>
+     * Data type: boolean.<br>
+     * Default value: true.
+     */
+    public static final Property<Boolean> ENABLE_IO_URING = new Property<>("enable.io.uring", true);
+
     public static class Property <T>{
         private final String name;
         private final T defaultValue;
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -54,13 +54,6 @@ public class AgentProperties{
      */
     public static final Property<Boolean> ENABLE_MANUALLY_SETTING_CPU_TOPOLOGY_ON_KVM_VM = new Property<Boolean>("enable.manually.setting.cpu.topology.on.kvm.vm", true);
 
-    /**
-     * Enable manually IO driver on KVM's VM. <br>
-     * Data type: boolean.<br>
-     * Default value: false.
-     */
-    public static final Property<Boolean> ENABLE_IO_URING = new Property<>("enable.io.uring", false);
-
     public static class Property <T>{
         private final String name;
         private final T defaultValue;
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -820,6 +820,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
         }
 
         // Enable/disable IO driver for Qemu (in case it is not set CloudStack can also detect if its supported by qemu)
+        // Do not remove - switching it to AgentProperties.Property may require accepting null values for the properties default value
         String enableIoUringConfig = (String) params.get(ENABLE_IO_URING_PROPERTY);
         enableIoUring = isIoUringEnabled(enableIoUringConfig);
 
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2950,6 +2950,23 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     }
 
     /**
+     * Check if IO_URING is available on the host
+     */
+    protected boolean isIoUringEnabled() {
+        s_logger.debug("Checking if iouring is enabled on the host");
+        String hostOsKey = "Host.OS";
+        Map<String, String> versionStrings = getVersionStrings();
+        if (MapUtils.isEmpty(versionStrings) || !versionStrings.containsKey(hostOsKey)) {
+            return false;
+        }
+        String os = versionStrings.get(hostOsKey);
+        String qemuBinary = os.equalsIgnoreCase("ubuntu") ? "/usr/bin/qemu-system-x86_64" : "/usr/libexec/qemu-kvm";
+        String command = String.format("ldd %s | grep -i uring", qemuBinary);
+        String result = executeBashScript(command);
+        return StringUtils.isNotBlank(result) && result.contains("liburing");
+    }
+
+    /**
      * Set Disk IO Driver, if supported by the Libvirt/Qemu version.
      * IO Driver works for:
      * (i) Qemu >= 5.0;
@@ -2958,7 +2975,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     protected void setDiskIoDriver(DiskDef disk) {
         if (getHypervisorLibvirtVersion() >= HYPERVISOR_LIBVIRT_VERSION_SUPPORTS_IO_URING
                 && getHypervisorQemuVersion() >= HYPERVISOR_QEMU_VERSION_SUPPORTS_IO_URING
-                && AgentPropertiesFileHandler.getPropertyValue(AgentProperties.ENABLE_IO_URING)) {
+                && isIoUringEnabled()) {
             disk.setIoDriver(DiskDef.IoDriver.IOURING);
         }
     }
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2954,23 +2954,18 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
      */
     protected boolean isIoUringSupportedByQemu() {
         s_logger.debug("Checking if iouring is supported");
-        String hostOsKey = "Host.OS";
-        Map<String, String> versionStrings = getVersionStrings();
-        if (MapUtils.isEmpty(versionStrings) || !versionStrings.containsKey(hostOsKey)) {
-            return false;
-        }
-        String os = versionStrings.get(hostOsKey);
-        if (org.apache.commons.lang3.StringUtils.isBlank(os)) {
-            return false;
-        }
-        if (os.equalsIgnoreCase("ubuntu")) {
-            // Enabled by default on ubuntu
-            return true;
-        }
-        String qemuBinary = os.toLowerCase().contains("suse") ? "/usr/bin/qemu-kvm" : "/usr/libexec/qemu-kvm";
-        String command = String.format("ldd %s | grep -Eqe '[[:space:]]liburing\\.so'", qemuBinary);
+        String command = getIoUringCheckCommand();
         int exitValue = executeBashScriptAndRetrieveExitValue(command);
         return exitValue == 0;
+    }
+
+    protected String getIoUringCheckCommand() {
+        String qemuPath = "/usr/bin/qemu-system-x86_64";
+        File file = new File(qemuPath);
+        if (!file.exists()) {
+            qemuPath = "/usr/libexec/qemu-kvm";
+        }
+        return String.format("ldd %s | grep -Eqe '[[:space:]]liburing\\.so'", qemuPath);
     }
 
     /**
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2964,7 +2964,7 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     }
 
     protected String getIoUringCheckCommand() {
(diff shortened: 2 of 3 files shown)


weizhouapache (approved):
code lgtm

GabrielBrascher:
@nvazquez maybe we should disable IO_URING to Rhel environments only, as these are the affected ones.
Ubuntu environments normally have the correct features mapped by Libvirt/Qemu.

Unfortunately, as we have been seeing in multiple KVM implementations, RedHat has a different route and tends to not keep up the pace with most of the new features that Qemu/Libvirt present. In this regard, there is no perfect world and we will always be compromising one side or the other (as many other features that had to be blocked due to Rhel).

I see that global settings are indeed a good way of disabling such features, although we can easily get lost with the growth of multiple features that need to be enabled/disabled accordingly to CentOS vs Ubuntu.

We might need better options to proceed with this.

With that said, your approach looks fine with me, as a quick fix.
A note in the upgrade process, raising how to proceed to enable in case the KVM is running in Ubuntu (or how to proceed to disable in the case of Rhel) could be interesting.

GabrielBrascher (approved):
Thanks for the PR @nvazquez! LGTM

A documentation point raising this might be good, in order to avoid users with Ubuntu and the specific Libvirt/Qemu setup missing this feature.

Although I do not believe this is going to be a good way for handling the multiple differences between CentOS & Ubuntu in the medium/long term ... it is what it is :)

For another time & future implementations: we might need to revisit some parts where we check for the Libvirt version as well and run different checks.

rohityadavcloud (approved):

rohityadavcloud:
@nvazquez cc @GabrielBrascher @wido what about Suse? If we can reliably detect the distro then the default behaviour (which should be documented somewhere) can be: if <detect distro> is Ubuntu, then enable io_uring, otherwise, disable it by default for other distros, and use the agent.properties always at the override.

wido:
io_uring brings a massive performance improvement to end-users. It's RHEL and thus CentOS/Alma/Rocky which have this problem. Ubuntu does not have this issue.

I do understand the problem here where CloudStack doesn't work out of the box, but if CloudStack becomes a project where you have a super long list of tunables/settings to get the optimal performance out of the system you will see users having endless tuning to do.

In the future io_uring should really be enabled by default or at least we should enable it if supported by the Qemu version. Otherwise users won't get the performance they want/expect.

Proxmox nowadays enables io_uring by default for all VMs being created. It's just such a massive performance boost.

GabrielBrascher:
@rohityadavcloud your idea would be great indeed in order to present the best features to each distro.

It is a shame that not all OS distros support some great features that Libvirt/Qemu present.
We've been always retrieving the Libvirt / Qemu versions in the codebase to decide which way to proceed when adding KVM features. I tend to follow Libvirt versions and release notes. Unfortunately, this has not been the best way in this particular case.

For us, as developers, it will always be a struggle to map all the different environments when developing a feature.

GabrielBrascher:
@andrijapanicsb I totally understand it. I am sorry in the case I was not clear.
I am +1 on proceeding with the current route and merging this, I do understand that this impact a relevant amount of our users.

But I wanted to take the opportunity to raise the concerns.
We've been seeing this happening other times, and it will stay happening many other times in the future.

Supporting a wide set of technologies has been a great strength of CloudStack, I do not want to take this out.
But we might need to improve the way we design some features in order to explore the best of each environment that CloudStack will be running into.

Otherwise, inevitably, people will start blaming CloudStack / KVM for "poor performance" and walk away to other options that suit best them. While they could actually be having the best for their environment.

Right now, I don't have any option on the table to fix it. But it is something good to keep in mind (for another time).

rohityadavcloud:
@nvazquez based on this, possibly the best compromise is to have io_uring enabled by default for Ubuntu based KVM hosts, but disabled otherwise. Can you amend the PR with this logic and propose a doc PR to document the default behaviour by distro? Thanks.

slavkap:
Hi all, we could try by parsing the stderr (but I know it's ugly) and enable/disable according to the output:

Ubuntu
```
sudo /usr/bin/qemu-system-x86_64 -drive aio=io_uring
qemu-system-x86_64: Device needs media, but drive is empty
```

Rocky
```
 /usr/libexec/qemu-kvm -drive aio=io_uring 
qemu-kvm: -drive aio=io_uring: invalid aio option
```

nvazquez:
Have discussed with @slavkap and @GabrielBrascher about a possible fix to detect if iouring is enabled by a call to qemu-kvm with the aio parameter. Exploring this option and will be updating the PR

weizhouapache:
> Have discussed with @slavkap and @GabrielBrascher about a possible fix to detect if iouring is enabled by a call to qemu-kvm with the aio parameter. Exploring this option and will be updating the PR

good. thanks @slavkap @nvazquez and @GabrielBrascher

NuxRo:
> Hi all, we could try by parsing the stderr (but I know it's ugly) and enable/disable according to the output:
> 
> Ubuntu
> 
> ```
> sudo /usr/bin/qemu-system-x86_64 -drive aio=io_uring
> qemu-system-x86_64: Device needs media, but drive is empty
> ```
> 
> Rocky
> 
> ```
>  /usr/libexec/qemu-kvm -drive aio=io_uring 
> qemu-kvm: -drive aio=io_uring: invalid aio option
> ```

I like this, it could work.
I've just tested on EL7 and we get the same error as on Rocky.

```
~]# /usr/libexec/qemu-kvm -drive aio=io_uring 
qemu-kvm: -drive aio=io_uring: invalid aio option
```

Did I mention io_uring breaks installs and upgrades on EL7, as well, which is a huge user base? :)

weizhouapache:
> 
> I like this, it could work. I've just tested on EL7 and we get the same error as on Rocky.
> 
> ```
> ~]# /usr/libexec/qemu-kvm -drive aio=io_uring 
> qemu-kvm: -drive aio=io_uring: invalid aio option
> ```
> 
> Did I mention io_uring breaks installs and upgrades on EL7, as well, which is a huge user base? :)

@nuxro EL7 might not be impacted, as io_uring is enabled only if libvirt >=6.0.3 and qemu>=5.0.0

NuxRo:
> > I like this, it could work. I've just tested on EL7 and we get the same error as on Rocky.
> > ```
> > ~]# /usr/libexec/qemu-kvm -drive aio=io_uring 
> > qemu-kvm: -drive aio=io_uring: invalid aio option
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Did I mention io_uring breaks installs and upgrades on EL7, as well, which is a huge user base? :)
> 
> @NuxRo EL7 might not be impacted, as io_uring is enabled only if libvirt >=6.0.3 and qemu>=5.0.0

Indeed, double checked this now.

slavkap:
A colleague told me another option to check if `liburing.so.2` is linked

On Ubuntu

```
ldd /usr/bin/qemu-system-x86_64 | grep -i uring
liburing.so.2 => /lib/x86_64-linux-gnu/liburing.so.2 (0x00007f4effa7f000)
```

On Rocky

```
ldd /usr/libexec/qemu-kvm | grep -i uring
echo $?
1
```

nvazquez:
@blueorangutan package

nvazquez:
Hi all - have applied the suggestions above to check whether iouring is enabled, please review

agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
```diff
@@ -54,13 +54,6 @@
      */
     public static final Property<Boolean> ENABLE_MANUALLY_SETTING_CPU_TOPOLOGY_ON_KVM_VM = new Property<Boolean>("enable.manually.setting.cpu.topology.on.kvm.vm", true);
 
-    /**
-     * Enable manually IO driver on KVM's VM. <br>
-     * Data type: boolean.<br>
-     * Default value: false.
-     */
-    public static final Property<Boolean> ENABLE_IO_URING = new Property<>("enable.io.uring", false);
-
```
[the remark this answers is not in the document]
NuxRo: @nvazquez +1 for optional properties entry, people might want to disable it, it's new tech, might misbehave etc.
agent.properties should have the last word on whether it's enabled or not, imho, if it's not too much bother.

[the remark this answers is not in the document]
nvazquez: Done thanks @slavkap @NuxRo

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2949,6 +2949,23 @@ private KVMPhysicalDisk getPhysicalDiskPrimaryStore(PrimaryDataStoreTO primaryDa
...
+     */
+    protected boolean isIoUringSupportedByQemu() {
+        s_logger.debug("Checking if iouring is supported");
+        String hostOsKey = "Host.OS";
+        Map<String, String> versionStrings = getVersionStrings();
+        if (MapUtils.isEmpty(versionStrings) || !versionStrings.containsKey(hostOsKey)) {
+            return false;
+        }
+        String os = versionStrings.get(hostOsKey);
+        String qemuBinary = os.equalsIgnoreCase("ubuntu") ? "/usr/bin/qemu-system-x86_64" : "/usr/libexec/qemu-kvm";
```
weizhouapache, line 2963: @nvazquez 
I think we can get the qemu emulator in other ways like
(1) use `virsh capabilities` to get the emulator for x86_64
(2) check if the files exist. e.g. if `/usr/bin/qemu-system-x86_64` exist, then use it, otherwise use `/usr/libexec/qemu-kvm`

(1) looks better.

nvazquez:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
nvazquez: @weizhouapache I have been testing on ubuntu20 and it seems like the check does not pass, so I have refactored it to be enabled by default on ubuntu: cc @wido @GabrielBrascher @slavkap 
````
# /usr/share/cloudstack-common/scripts/vm/hypervisor/versions.sh
Host.OS=Ubuntu
Host.OS.Version=20.04
Host.OS.Kernel.Version=5.9.0-050900-generic
# ldd /usr/bin/qemu-system-x86_64 | grep -i uring
# /usr/bin/qemu-system-x86_64 -drive aio=uring
Unable to init server: Could not connect: Connection refused
qemu-system-x86_64: -drive aio=uring: invalid aio option

Comment (nvazquez) [reply plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java]: This was regarding the qemu and libvirtd versions (< 5.0 and <6.3.0) - not an issue

Comment (nvazquez) [reply plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java]: Refactored @weizhouapache

**Comment by nvazquez** [comment]
@blueorangutan package

**Comment by weizhouapache** [comment]
@blueorangutan package

**Comment by nvazquez** [comment]
@blueorangutan test

**Comment by nvazquez** [comment]
@blueorangutan test

**Comment by nvazquez** [comment]
@blueorangutan test

**Comment by davidjumani** [comment]
@blueorangutan test

**Review by shwstppr — APPROVED** [review]
Tested with following hosts:

- Ubuntu 21.10. Was able to verify `<driver name='qemu' type='qcow2' cache='none' io='io_uring'/>`
- CentOS 7. VMs working fine and io-uring wasn't set for the VMs
- RockyLinux-8. Not seeing io-uring issue during VM deployment. Though deployment fails with issue fixed in #6244 . Should that be backported for 4.16 branch @nvazquez?
-

**Review by GabrielBrascher — APPROVED** [review]
LGTM

File: plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java [hunk 3]
```diff
@@ -2949,6 +2949,33 @@ private KVMPhysicalDisk getPhysicalDiskPrimaryStore(PrimaryDataStoreTO primaryDa
...
+        int exitValue = executeBashScriptAndRetrieveExitValue(command);
+        return exitValue == 0;
+    }
+
+    protected String getIoUringCheckCommand() {
+        String[] qemuPaths = {"/usr/bin/qemu-system-x86_64", "/usr/libexec/qemu-kvm", "/usr/bin/qemu-kvm" };
+        for (String qemuPath : qemuPaths) {
+            File file = new File(qemuPath);
+            if (file.exists()) {
+                String cmd = String.format("ldd %s | grep -Eqe '[[:space:]]liburing\\.so'", qemuPath);
```
rohityadavcloud, line 2971: is this reliable @nvazquez cc others?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2958,11 +2985,18 @@ private KVMPhysicalDisk getPhysicalDiskPrimaryStore(PrimaryDataStoreTO primaryDa
...
+                && isIoUringEnabled()) {
             disk.setIoDriver(DiskDef.IoDriver.IOURING);
         }
     }
 
+    /**
+     * IO_URING supported if it is supported by qemu AND the property 'enable.io.uring' is set to true
+     */
+    private boolean isIoUringEnabled() {
+        return isIoUringSupportedByQemu() && AgentPropertiesFileHandler.getPropertyValue(AgentProperties.ENABLE_IO_URING);
```
rohityadavcloud, line 2997: Should this be `propertyValue || if-io-uring-is-supported()` ?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
NuxRo: @rohityadavcloud LGTM

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
NuxRo: IMHO if "ldd qemu" does not reveal uring associations, but "enable.io.uring=true" is enabled in the agent.properties, then Cloudstack should still attempt to use uring. This is to cover custom (maybe statically linked etc) qemu-kvm installations.

nvazquez:
Thanks all!
@NuxRo I have modified the condition to honour the property in case its added, also as @wido @GabrielBrascher pointed out have simplified it for ubuntu and enabled it as well, otherwise check the ldd command
@shwstppr I have created the backport PR for the VNC issue: https://github.com/apache/cloudstack/pull/6402

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2958,11 +2985,27 @@ private KVMPhysicalDisk getPhysicalDiskPrimaryStore(PrimaryDataStoreTO primaryDa
...
             disk.setIoDriver(DiskDef.IoDriver.IOURING);
         }
     }
 
+    /**
+     * IO_URING supported if the property 'enable.io.uring' is set to true OR it is supported by qemu
+     */
+    private boolean isIoUringEnabled() {
+        Boolean propertyValue = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.ENABLE_IO_URING);
+        return BooleanUtils.isTrue(propertyValue) || isBaseOsUbuntu() || isIoUringSupportedByQemu();
```
weizhouapache, line 2998: In my opiniom, the logic is
If enable.io.uring is true, enable io uring.
If enable.io.uring is false, disable io uring.
If it is not set, determine by ldd

nvazquez: Thanks @weizhouapache, I have applied this - only difference I added: when it is not set then enable for ubuntu or determine ldd as per @wido @GabrielBrascher comments this wasn't an issue for Ubuntu (also this check is performed when meeting the supported qemu and libvirt versions)

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

DaanHoogland (approved):

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -817,6 +819,11 @@ public boolean configure(final String name, final Map<String, Object> params) th
             directDownloadTemporaryDownloadPath = getDefaultDirectDownloadTemporaryPath();
         }
 
+        // Enable/disable IO driver for Qemu (in case it is not set CloudStack can also detect if its supported by qemu)
+        // Do not remove - switching it to AgentProperties.Property may require accepting null values for the properties default value
+        String enableIoUringConfig = (String) params.get(ENABLE_IO_URING_PROPERTY);
+        enableIoUring = isIoUringEnabled(enableIoUringConfig);
```
slavkap, line 825:
```suggestion
        enableIoUring = isIoUringEnabled(enableIoUringConfig);
```
needs to be moved after this part of the code 
 https://github.com/apache/cloudstack/blob/363a2cff8231a31c3c9def329805910e43358b1d/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java#L1126-L1133

otherwise it will be always `false`

nvazquez: Thanks @slavkap, done

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -1132,6 +1134,11 @@ public boolean configure(final String name, final Map<String, Object> params) th
             s_logger.trace("Ignoring libvirt error.", e);
         }
 
+        // Enable/disable IO driver for Qemu (in case it is not set CloudStack can also detect if its supported by qemu)
+        // Do not remove - switching it to AgentProperties.Property may require accepting null values for the properties default value
+        String enableIoUringConfig = (String) params.get(ENABLE_IO_URING_PROPERTY);
+        enableIoUring = isIoUringEnabled(enableIoUringConfig);
+
```
slavkap, line 1141: one small suggestion if you want to log if the IO uring is enabled/disabled

slavkap (approved):
code LGTM! Rebased on main and tested on Rocky - libvirt/qemu versions

> Compiled against library: libvirt 8.0.0
> Using library: libvirt 8.0.0
> Using API: QEMU 8.0.0
> Running hypervisor: QEMU 6.2.0

and with set/not set `enable.io.uring` in the agent properties

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
andrijapanicsb: Yes please, let's do the proper logging.

andrijapanicsb:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
nvazquez: Done, thanks

PR #5984, opened 2022-02-10
Persistence of VM stats

joseflauzino:
### Description

This PR addresses issue #5935.

In summary, this PR makes ACS persist VM stats in the database to provide more consistency and make it possible to obtain historical data.

For more information, please refer to the spec contained in issue #5935.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [X] Major
- [ ] Minor

#### New API
- Getting all available stats:
sem-filtro

- Filtering by start and end date:
com filtro

#### UI changes
- Metrics disabled. It shows all VMs:

- Metrics enabled. It shows only running VMs and their current stats:

- Metrics enabled and filtering by stopped VMs. It forces the display of stopped VMs and their latest stats collected (not reflecting their current stats):

### How Has This Been Tested?

I added unit tests to several methods.

Also, in a local lab, I performed the following tests:
- I tested different combinations of the global settings `vm.stats.increment.metrics`, `vm.stats.max.retention.time`, and `vm.stats.interval`. All tests resulted in the expected behavior, as described in section [2.2](https://github.com/apache/cloudstack/issues/5935#2-proposed-changes--22-configuration-proposal) of the spec.
- I tested the listVirtualMachinesMetrics API with different combinations of its new `accumulate` parameter with the global configuration `vm.stats.increment.metrics`, and everything worked as expected. See sections [2.2](https://github.com/apache/cloudstack/issues/5935#2-proposed-changes--22-configuration-proposal) and [3.3](https://github.com/apache/cloudstack/issues/5935#3-work-items--33-api) of the spec for more details.
- I tested the UI modifications mentioned in section [3.4](https://github.com/apache/cloudstack/issues/5935#3-work-items--34-ui) of the spec. The obtained behavior was as follows:
    - When clicking on the "Metrics" button only running VMs and their respective stats were listed;
    - Then it was possible to "force" the listing of stopped VMs by changing the filter from "All" to "Stopped".
- I tested the new listVirtualMachinesUsageHistory API by filtering by different time periods and VMs. All the filtering described in section [2.4](https://github.com/apache/cloudstack/issues/5935#2-proposed-changes--24-new-api-proposal) of the spec was successfully tested.

Issue #5935, opened 2022-02-04
Persistence of VM stats


joseflauzino:
##### ISSUE TYPE

 * Enhancement Request

##### COMPONENT NAME

~~~
StatsCollector
~~~

##### CLOUDSTACK VERSION

~~~
4.17
~~~

##### SUMMARY

This spec changes the way Apache CloudStack collects and stores the VM stats to make the data more consistent and provide historical data.

------

# Table of Contents

1.  [Problem description](#1-problem-description)
    1.1. [Current collecting/storing data workflows and possible configurations](#1-problem-description--11-current-collectingstoring-data-workflows-and-possible-configurations)
    1.2. [Current data cleaning workflow](#1-problem-description--12-current-data-cleaning-workflow)
    1.3. [Current API](#1-problem-description--13-current-api)
2. [Proposed changes](#2-proposed-changes)
    2.1. [Proposed collecting/storing data workflow](#2-proposed-changes--21-proposed-collecting/storing-data-workflow)
    2.2. [Configuration proposal](#2-proposed-changes--22-configuration-proposal)
    2.3. [Data cleaning proposal](#2-proposed-changes--23-data-cleaning-proposal)
    2.4. [New API proposal](#2-proposed-changes--24-new-api-proposal)
    2.5. [UI adjustment proposal](#2-proposed-changes--25-ui-adjustment-proposal)
3. [Work items](#3-work-items)
    3.1. [Database tables](#3-work-items--31-database-tables)
    3.2. [Global configurations](#3-work-items--32-global-configurations)
    3.3. [API](#3-work-items--33-api)
    3.4. [UI](#3-work-items--34-ui)
4. [Future works](#4-future-works)
------

<h1 id="1-problem-description">1. Problem description</h1>

In Apache CloudStack (ACS), VM stats are collected by Management Servers. Currently, each Management Server collects the data independently and stores it only in (primary) memory. This model of collecting and storing VM stats results in some limitations, numbered as follows:

1.  When restarting a Management Server (or when it crashes), the VMs stats data is lost (since there is no data persistence);

2.  When the cloud is composed of multiple Management Servers, each one of them can show different data about the VMs, as there is no centralization or synchronization of the data collected by different Management Servers;

[...]

PR #6401, opened 2022-05-20
VM stats history visualization

joseflauzino:
### Description

PR #5984 recently implemented the `listVirtualMachinesUsageHistory` API, which provides historical data of VM stats. However, the current ACS UI only shows VM stats from the most recent data collection.

This PR implements new features in the UI which allow ACS users to view the historical stats of each VM individually.

### Types of changes

- [x] Major
- [ ] Minor

### Screenshots:

**Charts:**

CPU

Memory

Network

Disk

**Possible Filtering:**
filters

### How Has This Been Tested?

I went to Compute > Instances > `any-vm-instance` > Statistics and was able to visualize all the statistics charts for the selected VM.
Also, I filtered by different period combinations and the data was presented correctly.


rohityadavcloud:
Looks fantastic, looking forward to reviewing/testing this.

nvazquez:
Wow - looks very nice @joseflauzino thanks!

DaanHoogland:
@blueorangutan test keepEnv

DaanHoogland:
@joseflauzino, two minor points
1. I build and deployed the code from this PR but don´t get any data from this weekend, dispite deploying some VMs in the nested env on saterday: This is because the default value for `vm.stats.max.retention.time` is 1. Does it make sense to set it for instance to 5, so there is always something to see?
2. when refreshing the default tab is displayed again and after re-opening the statistics tab the configurations (`all data`or a `start date`/`end date` pair) are reset/forgotten.

in general tests good and looks good,
nice thanks

DaanHoogland (approved):
clgtm
tests ok

joseflauzino:
@DaanHoogland thanks for your review and testing.

>     1. I build and deployed the code from this PR but don´t get any data from this weekend, dispite deploying some VMs in the nested env on saterday: This is because the default value for `vm.stats.max.retention.time` is 1. Does it make sense to set it for instance to 5, so there is always something to see?

Both `vm.stats.max.retention.time` and `vm.stats.interval` were set to 1 min by default on purpose (in PR #5984, not in the current PR). The reason is that it is common to find clouds with a considerable amount of VM instances (e.g. hundreds or even over 1k). So the amount of data generated even in a short 5 min period can be very large and the impact can also be large if it is not something planned by the cloud operators. So I think it is more prudent that we leave the default values of these parameters in a way that generates the minimum amount of data possible. It is then up to the cloud administrators/operators to decide whether or not to increase the amount of stats data to be kept in the database.


>     2. when refreshing the default tab is displayed again and after re-opening the statistics tab the configurations (`all data`or a `start date`/`end date` pair) are reset/forgotten.

This was also an intentional detail. The idea is to have the user open the filtering modal and click the OK button whenever they want to get updated data for the selected period. The reason is that during our tests there were several times when users clicked the refresh button and it looked like the data was not updated. But what happened was that the selected period was in the past, so the data would always be the same. The way it is now is that the user will always check the selected period before getting updated data. My intention is to leave it this way for now and according to the feedback we get from users we could think of other ways to improve the user experience.

DaanHoogland:
@joseflauzino
Ad 1. Ok makes sense
Ad 2. Makes sense as well, can we then in this situation change the text on the button or add a note next to it instructing like wise?

joseflauzino:
@DaanHoogland I have just committed a change that applies your suggestion related to the submit button text.

DaanHoogland:
> @DaanHoogland I have just committed a change that applies your suggestion related to the submit button text.

looks ok, I can imagine the novice won´t be clear on this, but letś go for it!

ui/public/locales/en.json
```diff
@@ -2880,6 +2904,8 @@
 "message.copy.iso.confirm": "Please confirm that you wish to copy your ISO to",
 "message.copy.template": "Copy template <b id=\"copy_template_name_text\">XXX</b> from zone <b id=\"copy_template_source_zone_text\"></b> to",
 "message.copy.template.confirm": "Are you sure you want to copy template?",
+"message.cpu.usage.info.multi.cpu": "The percentage of CPU usage can go over 100% when a VM has more than 1 vCPU. This happens because the usage of all the vCPUs of the VM are accumulated.",
```
GutoVeronezi, line 2907: @joseflauzino, actually it depends on how the hypervisor account the stats. As far as I know, the way VMware accounts it, the CPU will never pass 100%. I think a more accurately message would be:

```suggestion
"message.cpu.usage.info.multi.cpu": "The CPU usage percentage can exceed 100% if the VM has more than 1 vCPU or when CPU Cap is not enabled. This behavior happens according to how the hypervisor being used accounts the stats.",
```

ui/src/components/view/StatsTab.vue
```diff
@@ -0,0 +1,663 @@
...
+    /**
+     * Calculates the memory percentage.
+     * @param isUsed "true" if the memory used percentage should be returned, "false" if the free memory percentage should be returned.
+     * @param memoryTotalInKB the memory total (in KB).
+     * @param memoryFreeInKB the memory free (in KB).
+     * @returns the percentage of used/free memory.
+     */
+    calculateMemoryPercentage (isUsed, memoryTotalInKB, memoryFreeInKB) {
+      var percentage = -1
+      if (memoryTotalInKB != null && memoryFreeInKB != null) {
```
GutoVeronezi, line 525: We could invert this `if` to reduce indentation.

ui/src/components/view/StatsTab.vue
```diff
@@ -0,0 +1,663 @@
...
+        if (isUsed) {
+          percentage = parseFloat(100.0 * (memoryTotalInKB - memoryFreeInKB) / memoryTotalInKB).toFixed(2)
+          return percentage
```
GutoVeronezi, line 528:
```suggestion
          return parseFloat(100.0 * (memoryTotalInKB - memoryFreeInKB) / memoryTotalInKB).toFixed(2)
```

RodrigoDLopez (approved):
lgtm

GabrielBrascher (approved):
Nice PR, thanks @joseflauzino!
Overall LGTM (I haven't tested though)

rohityadavcloud:
@joseflauzino can you fix the conflicts.

Asking @Pearl1594 @vladimirpetrov @borisstoyanov - if we can have a round of review/testing this PR. Thanks. cc @shwstppr

rohityadavcloud:
@joseflauzino can you fix the conflicts.

rohityadavcloud:
ping @joseflauzino can you fix the conflicts.

harikrishna-patnala:
pasting the same question which is on the other PR:

Hi @joseflauzino - does the new API return the exact same API output format/keys as the deprecated listVirtualMachinesMetrics with the PR https://github.com/apache/cloudstack/pull/5984?

GutoVeronezi:
> pasting the same question which is on the other PR:
> 
> Hi @joseflauzino - does the new API return the exact same API output format/keys as the deprecated listVirtualMachinesMetrics with the PR #5984?

Hi, @harikrishna-patnala

No, the APIs have different outputs. `listVirtualMachinesUsageHistory` retrieves just some VM information (`name`, `id`, `displayname`) and the stats history (an array with several stats objects, according to the period filtered and the configurations `vm.stats.interval` and `vm.stats.max.retention.time`), while `listVirtualMachinesMetrics` retrieves several VM fields and just the most recent stats data.

GutoVeronezi:
@harikrishna-patnala, as this PR only touches on the UI, the deprecation of the API `listVirtualMachinesMetrics` is related to another PR (#5984), which is already merged, and the deprecation or not of the API does not affect this PR, the discussion about the deprecation of the API `listVirtualMachinesMetrics` would not fit here. I think we should create an _issue_/ML to discuss the situation and move on from there.

harikrishna-patnala:
agreed @GutoVeronezi I've raised it because @joseflauzino raised this PR and both are related. I'll create a separate issue for this.

rohityadavcloud:
I think @joseflauzino the author must comment if there is/was a good reason to deprecate the API in his backend PR/feature. If there is then no need to track it in another issue/PR if no further work is needed.

joseflauzino:
Hi all. Sorry for not answering, I'm away for these days for health reasons. But I will try to answer your questions as much as possible.

@harikrishna-patnala and @rohityadavcloud, all the details regarding the changes made to PR #5984 are contained in spec #5935. Please refer to sections [1.3](https://github.com/apache/cloudstack/issues/5935#1-problem-description--13-current-api) and [3.3](https://github.com/apache/cloudstack/issues/5935#3-work-items--33-api) for more details on the changes regarding the `listVirtualMachinesMetrics` API. If needed, we can discuss this in another issue, as @GutoVeronezi  commented.

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test keepEnv

DaanHoogland:
@shwstppr are you still testing this or is it ready for merge?

shwstppr:
@DaanHoogland yes, still testing. Hoping to post my review by tomorrow or day after.

shwstppr (changes requested):
@joseflauzino great work on this functionality. Tested a few things, and it looks good.
But one issue I found is that when I use local timezone, the date in the period selector is shown in local timezone but on charts they are still in UTC
Screenshot from 2022-08-03 13-19-45

I was able to fix it with,
```
diff --git a/ui/src/components/view/StatsTab.vue b/ui/src/components/view/StatsTab.vue
index 55b0fdb854..aa6cb5c1d2 100644
--- a/ui/src/components/view/StatsTab.vue
+++ b/ui/src/components/view/StatsTab.vue
@@ -398,7 +398,8 @@ export default {
       const diskIopsLine = { label: 'IOPS', backgroundColor: blueInRgba, borderColor: blue, data: [], pointRadius: chartPointRadius }
 
       for (const element of vm[0].stats) {
-        const currentLabel = element.timestamp.split('T')[0] + ' ' + element.timestamp.split('T')[1].split('-')[0]
+        var ts = this.$toLocalDate(element.timestamp)
+        const currentLabel = ts.split('T')[0] + ' ' + ts.split('T')[1].split('-')[0]
         this.chartLabels.push(currentLabel)
 
         cpuLine.data.push({ timestamp: currentLabel, stat: element.cpuused.split('%')[0] })
diff --git a/ui/src/utils/plugins.js b/ui/src/utils/plugins.js
index 63a0e609c5..35a9564691 100644
--- a/ui/src/utils/plugins.js
+++ b/ui/src/utils/plugins.js
@@ -305,6 +305,18 @@ export const toLocaleDatePlugin = {
       dateWithOffset = dateWithOffset.substring(0, dateWithOffset.length - 4)
       return dateWithOffset
     }
+
+    app.config.globalProperties.$toLocalDate = function (date) {
+      var timezoneOffset = this.$store.getters.timezoneoffset
+      if (this.$store.getters.usebrowsertimezone) {
+        // Since GMT+530 is returned as -330 (mins to GMT)
+        timezoneOffset = new Date().getTimezoneOffset() / -60
+      }
+      var milliseconds = Date.parse(date)
+      // e.g. "Tue, 08 Jun 2010 19:13:49 GMT", "Tue, 25 May 2010 12:07:01 UTC"
+      var dateWithOffset = new Date(milliseconds + (timezoneOffset * 60 * 60 * 1000))
+      return dateWithOffset.toISOString()
+    }
   }
 }
```

Few improvements, 
- default time period could be used the current time minus stats retention time. This is just my opinion. It will give better UX as user will see some stats when view is opened first.
- Something for future maybe, functionality could be extended to other system vms and VRs as well. Or maybe other resources too.

rohityadavcloud:
Overall, LGTM but before this can be merged a few concerns if you can advise on @joseflauzino:
- How does the backend API/DB handle as the data grows - my concern is if this would always grow with time and create issues we perhaps hit like with the usage server (I understand they are not in this PR but the feature)
- Is this framework expandable to other resources such as volumes, systemvms and virtual routers?

joseflauzino:
Thanks for the tests and review, @shwstppr .

I'll test your suggestions soon.

About the possible improvements you commented on:

> default time period could be used the current time minus stats retention time. This is just my opinion. It will give better UX as user will see some stats when view is opened first.

Depending on the data retention settings and the amount of VM instances in the cloud, using the current time minus the retention time can bring in a massive amount of data. So the delay in loading the VM stats could be very large when accessing this view. The idea of filtering by the last hour of available data when accessing this view is safer for this initial version. Maybe in the future, we can think of something more elaborate based on feedback from ACS users.

> Something for future maybe, functionality could be extended to other system vms and VRs as well. Or maybe other resources too.

I think so too! In the future it would be nice to have this for system VMs and other cloud resources.

joseflauzino:
@rohityadavcloud 

> How does the backend API/DB handle as the data grows - my concern is if this would always grow with time and create issues we perhaps hit like with the usage server (I understand they are not in this PR but the feature)

The amount of data that is kept in the database is controlled by the cloud administrators. This can be done through the global configuration `vm.stats.max.retention.time`, which was created in PR #5984 (please, in spec #5935, see sections [2.2](https://github.com/apache/cloudstack/issues/5935#2-proposed-changes--22-configuration-proposal) and [2.3](https://github.com/apache/cloudstack/issues/5935#2-proposed-changes--23-data-cleaning-proposal) for details).

> Is this framework expandable to other resources such as volumes, systemvms and virtual routers?

At the code level we would need to analyze how much it would be reusable. But for sure, at the very least, the idea (including the design and some logics of the code) can be extended to the other resources/elements.

shwstppr:
@blueorangutan package

shwstppr (changes requested):
@joseflauzino found a couple of issues while testing this time,

1. While filtering stats, API is called with hidden fields. eg: First filter using start, end dates Then select All data. Check API call, it still contains start/enddate param even though they are hidden in UI.
2. Time in filter UI is shown in the local timezone and the same is used in the API call without any timezone. Does `listVirtualMachinesUsageHistory` API convert it to server time? I guess not and it won't list stats correctly in many cases

joseflauzino:
@shwstppr 

> 1. While filtering stats, API is called with hidden fields. eg: First filter using start, end dates Then select All data. Check API call, it still contains start/enddate param even though they are hidden in UI.

Good point. I fixed this bug now.

> 2. Time in filter UI is shown in the local timezone and the same is used in the API call without any timezone. Does listVirtualMachinesUsageHistory API convert it to server time? I guess not and it won't list stats correctly in many cases

The timestamps come in server timezone, but the `listVirtualMachinesUsageHistory` API does not do a conversion or anything like that. In fact, the API receives the date fields without timezone. But the reason for this goes beyond the API itself:

Currently, the backend only supports dates in the `yyy-MM-dd` or `yyy-MM-dd HH:mm:ss` format, so there is no way to pass dates with timezone to the APIs (the related code is in the ParamProcessWorker class).
I think it would be interesting to implement support for timezone date objects in the ACS APIs. However, this is beyond the scope of this PR, so my suggestion is to map this to another issue.

shwstppr:
@joseflauzino sorry if I didn't explain myself better. I'm not saying API should accept timezone but maybe we can call API with timezone that the server expects. My issue currently is,

- In UI, I'm using IST timezone (UTC+0530)
- Server is storing stats with UTC timezone, API returns the times with the same.
- I deploy a VM at 1000hr IST. Server starts collecting/storing stats for the VM but they are stored with time starting 0430hr UTC
- Now in UI at 1100hr IST I want to see stats. UI allows me to filter stats between 1000-1100hr IST. When I make this selection I see nothing because API is called with something like `startDate=2022-08-09+10:00:00&endDate=2022-08-09+11:00:00`.
- As the server is behind by 0530h. I'll never see stats for the current time. But if I call API with all available data it will show the stats.

joseflauzino:
@shwstppr 

> ... maybe we can call API with timezone that the server expects.

To do that, before calling the API, we would have to convert the client timezone (from the browser) to the management server timezone, right?

In this case, my question is: in the UI context (on the client), how can we find out the management server timezone before calling the API?

Is there a ready-made mechanism to do this? If there is, please let me know and then I will implement this timezone conversion.

If not, I think we can address this in another issue and then implement this kind of timezone conversion in this and other views that do similar filtering (API calls passing date objects as parameters).

shwstppr:
@joseflauzino maybe the reverse of how we convert server time into local time, https://github.com/apache/cloudstack/blob/main/ui/src/utils/plugins.js#L294-L301

rohityadavcloud:
ping @joseflauzino are you able to address the comment - I think we're very close to merging this.

shwstppr:
@blueorangutan package

shwstppr (approved):
LGTM

rohityadavcloud:
Good stuff @joseflauzino !

PR #6410, opened 2022-05-24
Add check for update only name and/or description of disk offering

BryanMLima:
### Description

While trying to update a disk offering through the UI, the tag fields are always sent, even when they are not changed. This causes a bug when active volumes are using a disk offering in storage pools with tags. As ACS would interpret this as an update on the tags as well. This PR aims to fix this problem, only sending the tags if they are different from the current tags, allowing the user to update the name and/or description of disk offerings, as it is already allowed through CloudMonkey.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### How Has This Been Tested?

On a local lab, I tried updating the disk offering with the following use cases:
- updating only the name
- updating only the description
- updating only the name and description fields

All of the tests worked as expected, the UI only sent the the fields that were changed.

Then, while trying to update the name and/or description fields, and changing the tags as well, the API was sending the tags parameters, as expected.

### DIFF
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -1363,6 +1363,15 @@ export default {
           delete values.tags
         }
 
+        if (['updateServiceOffering'].includes(action.api)) {
+          if (values.hosttags === this.resource.hosttags) {
+            delete values.hosttags
+          }
+          if (values.storagetags === this.resource.storagetags) {
+            delete values.tags
+          }
+        }
+
         for (const key in values) {
           const input = values[key]
           for (const param of action.params) {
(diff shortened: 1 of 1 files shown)


utchoang (approved):
LGTM!

GabrielBrascher (approved):
LGTM

shwstppr:
@BryanMLima  Similar issue is with updateServiceOffering for host and storage tags, is it possible to add check for that as well?

shwstppr (approved):

Issue #6088, opened 2022-03-10
UI: Add zone wizard missing provider for primary storage


nvazquez:
##### ISSUE TYPE
 * Improvement Request

##### COMPONENT NAME
~~~
UI
~~~

##### CLOUDSTACK VERSION
~~~
4.16, main
~~~

##### SUMMARY
When adding a primary storage through the Add zone wizard, the different storage providers are not displayed. This ends up on a non-successful zone creation which needs manual addition of the storage pools for different providers

![image](https://user-images.githubusercontent.com/5295080/157691647-4049165a-e1d6-4de8-ab18-95e9b64691fd.png)

##### STEPS TO REPRODUCE
~~~
Add zone
Complete the steps until Primary storage
Try registering for example NFS storage pool for a provider such as PowerFlex, SolidFire, etc
~~~

##### EXPECTED RESULTS
~~~
Add zone succeeds
~~~

##### ACTUAL RESULTS
~~~
Add zone fails, needs manual registration of storage pools
~~~

rohityadavcloud:
@nvazquez would this be useful for 4.17.1.0 cc @utchoang @shwstppr ?

PR #6429, opened 2022-06-01
UI: add a provider for primary storage when adding zone wizard

utchoang:
### Description

This PR fixes https://github.com/apache/cloudstack/issues/6088

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

image

### DIFF
--- a/ui/src/views/infra/zone/StaticInputsForm.vue
+++ b/ui/src/views/infra/zone/StaticInputsForm.vue
@@ -142,6 +142,9 @@ export default {
         const fieldsChanged = toRaw(changedFields)
         this.$emit('fieldsChanged', fieldsChanged)
       }
+    },
+    'prefillContent.primaryStorageProtocol' (val) {
+      this.form.primaryStorageProtocol = val
     }
   },
   methods: {
--- a/ui/src/views/infra/zone/StaticInputsForm.vue
+++ b/ui/src/views/infra/zone/StaticInputsForm.vue
@@ -145,6 +145,11 @@ export default {
     },
     'prefillContent.primaryStorageProtocol' (val) {
       this.form.primaryStorageProtocol = val
+    },
+    'prefillContent.provider' (val) {
+      if (['SolidFire', 'PowerFlex'].includes(val)) {
+        this.form.primaryStorageProtocol = 'custom'
+      }
     }
   },
   methods: {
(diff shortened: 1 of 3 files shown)


slavkap:
Hi @utchoang, I think adding a `custom` protocol is needed for SolidFire and PowerFlex, and because the protocol parameter is set as mandatory

sureshanaparti:
> Hi @utchoang, I think adding a `custom` protocol is needed for SolidFire and PowerFlex, and because the protocol parameter is set as mandatory

@utchoang Correct, protocol is the required input at UI, for adding primary storage. This is imp to pick the correct storage provider, url and related parameter inputs. If `custom` protocol for PowerFlex is confusing, this can be updated to something meaningful, eg. PreSetup or PowerFlex (now or can be re-visited later ?), but ensure that all the provider details are properly picked and passed to the API call. Also, it's better to keep adding primary storage options in sync, at zone wizard and add primary storage dialog. Check if any documentation update needed for these changes. Thanks.

utchoang:
@slavkap @sureshanaparti Thank you for your comments. I fixed it, please review it again.

slavkap:
Hi @utchoang, I think it will be more user-friendly if the `custom` protocol option is shown with the rest of the protocol options. Because we fill in the form sequentially, the provider, in this case, should be above the protocol, or the `custom` option should be in the list of protocols. Another option when you choose PowerFlex/SolidFire, the `custom` protocol could be automatically set.
Also, I think that it's not a good idea to limit the `custom` option only to SolidFire/PowerFlex (e.g. Datera plugin also is using this option, probably there are other storage plugins even such not in CS upstream but have integrations with CS)

sureshanaparti:
@utchoang Sometimes 'Protocol' is cleared when 'Provider' is selected, please check.

slavkap (changes requested):
Hi @utchoang, here https://github.com/apache/cloudstack/blob/d815b9cf490ab27c9e341114d82415bde0805291/ui/src/views/infra/zone/ZoneWizardLaunchZone.vue#L1327
if it's SharedMountPoint, the `server` should be set to `server = 'localhost'` because it's not provided by default

slavkap (approved):
Thanks, @utchoang, for the PR! 
code LGTM tested it with SharedMoutPoint and StorPool

shwstppr (approved):

PR #6435, opened 2022-06-02
VPC VR: fix /etc/network/interfaces is misconfigured

weizhouapache:
### Description

This PR fixes that the /etc/network/interfaces is misconfigured in VPC VR startup, which causes VPC VR to be stuck for 1 minute and continue working.

The issue is because `ifup eth0` and `ifdown eth0` do not work when `/etc/network/interfaces` is not correct in VPC VR.

severity: to be discussed.
This requires a new systemvm template.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/systemvm/debian/opt/cloud/bin/setup/init.sh
+++ b/systemvm/debian/opt/cloud/bin/setup/init.sh
@@ -188,6 +188,7 @@ setup_interface_sshd() {
     setup_sshd $ETH1_IP "eth1"
 
   elif [ "$TYPE" == "vpcrouter" ]; then
+    init_interfaces "eth0"
     setup_interface "0" $ETH0_IP $ETH0_MASK $GW
     setup_sshd $ETH0_IP "eth0"
 
--- a/systemvm/debian/opt/cloud/bin/setup/vpcrouter.sh
+++ b/systemvm/debian/opt/cloud/bin/setup/vpcrouter.sh
@@ -25,11 +25,6 @@ setup_vpcrouter() {
     grep -q $NAME /etc/hosts || echo "127.0.0.1 $NAME" >> /etc/hosts;
   fi
 
-    cat > /etc/network/interfaces << EOF
-auto lo eth0
-iface lo inet loopback
-EOF
-
   echo $NAME > /etc/hostname
   echo 'AVAHI_DAEMON_DETECT_LOCAL=0' > /etc/default/avahi-daemon
   hostnamectl set-hostname $NAME


weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan package

systemvm/debian/opt/cloud/bin/setup/init.sh
```diff
@@ -188,6 +188,7 @@ setup_interface_sshd() {
     setup_sshd $ETH1_IP "eth1"
 
   elif [ "$TYPE" == "vpcrouter" ]; then
+    init_interfaces "eth0"
```
rohityadavcloud, line 191: why not move this entirely to the appliance/type specific script cc @weizhouapache @Pearl1594 ? This would avoid requiring new systemvmtemplate anytime the init (or early-config?) script.

Pearl1594: With the new approach the IP on the interfaces need to be available right at the time of VM boot up, so that we can perform the scp operation , so, I am not sure moving this to a separate script would eliminated the need to build a new systemVM template.

rohityadavcloud: I suppose the question is - is there a way to refactor the script/sequence in a way that avoids building a new template everytime the first-run scripts (such as init.sh or cloud-early-config) are changed? This would improve the maintainability of the systemvmtemplates. @Pearl1594 @weizhouapache

weizhouapache: @rohityadavcloud 
in my opinion, it might be possible. 
`init.sh` needs to  (1) get cmdline; (2) configure link-local interface; (3) setup ssh. 
The rest (init and configure public and guest nics, etc) can be moved to other scripts and executed after `scp`. 
is it right ? @Pearl1594

Pearl1594: Yup, if the control IP is set and ssh comes up fine, then yes I guess it could work.

rohityadavcloud:
ping @weizhouapache any update on this?

shwstppr:
Based on discussion with @weizhouapache moving this to the next milestone

shwstppr (commented):
As already discussed in earlier conversations this would need a new systemvm template with a change in `init.sh`.
And some form of testing after that

systemvm/debian/opt/cloud/bin/setup/init.sh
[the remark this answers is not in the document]
weizhouapache: @shwstppr it should be ok.

If only eth0 is passed, it is same as
```
cat > /etc/network/interfaces << EOF
auto lo eth0
iface lo inet loopback
EOF
```

shwstppr (approved):

DaanHoogland (approved):
clgtm

weizhouapache:
> @weizhouapache @shwstppr does this require a new systemvmtemplate or are the changes patches and run after the patching process? @blueorangutan package

@rohityadavcloud 
yes, it requires a new systemvm template

Actually, any change with init.sh and bootstrap.sh requires a new systemvm template

weizhouapache:
Prior to this change
image

with this change
image

PR #6442, opened 2022-06-07
Possibility to choose the source NAT IP address on a isolated network or VPC

DaanHoogland:
### Description

This PR intends to allow users to choose an IP address for the VR of their network from the pool of available addresses.

### Types of changes

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial


DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

weizhouapache:
@DaanHoogland 
is this ready for review ?

DaanHoogland:
> @DaanHoogland is this ready for review ?

I need to do some functional testing and discuss with @alexandremattioli if this is a good intermediat state to work from.

DaanHoogland:
@blueorangutan test

DaanHoogland:
@weizhouapache please review?

api/src/main/java/com/cloud/network/Network.java
```diff
@@ -268,6 +268,7 @@ public static class Capability {
         public static final Capability LoadBalancingSupportedIps = new Capability("LoadBalancingSupportedIps");
         public static final Capability AllowDnsSuffixModification = new Capability("AllowDnsSuffixModification");
         public static final Capability RedundantRouter = new Capability("RedundantRouter");
+        public static final Capability SelectSnatIpAllowed = new Capability("SelectSnatIpAllowed");
```
weizhouapache, line 271: The capability name seems inconsistent with other capabilities.
Maybe better to use AllowSpecificSourceNatIp or just SpecifySourceNatIp

api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
```diff
@@ -931,6 +931,7 @@ public class ApiConstants {
     public static final String LOGIN = "login";
     public static final String LOGOUT = "logout";
     public static final String LIST_IDPS = "listIdps";
+    public static final String IS_SELECTION_OF_STATIC_NAT_ALLOWED = "selectsnatipallowed";
```
weizhouapache, line 934: I think it is Source Nat IP, not Static Nat Ip, right ?

api/src/main/java/org/apache/cloudstack/api/response/NetworkOfferingResponse.java
```diff
@@ -143,6 +143,10 @@ public class NetworkOfferingResponse extends BaseResponseWithAnnotations {
     @Param(description = "the internet protocol of the network offering")
     private String internetProtocol;
 
+    @SerializedName(ApiConstants.IS_SELECTION_OF_STATIC_NAT_ALLOWED)
```
weizhouapache, line 146: Better to use SourceNat instead of Snat

engine/schema/src/main/java/com/cloud/offerings/NetworkOfferingVO.java
```diff
@@ -165,6 +165,17 @@ public String getDisplayText() {
     @Column(name="service_package_id")
     String servicePackageUuid = null;
 
+    @Column(name = "select_snat_address_allowed")
```
weizhouapache, line 168: Can we use network offering details instead of a new column ?
well, it means lots of changes maybe

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -1289,15 +1289,12 @@ private void checkSharedNetworkCidrOverlap(Long zoneId, long physicalNetworkId,
         }
     }
 
-    private void validateRouterIps(String routerIp, String routerIpv6, String startIp, String endIp, String gateway,
-                                   String netmask, String startIpv6, String endIpv6, String ip6Cidr) {
+    private void validateSharedRouterIPv4(String routerIp, String startIp, String endIp, String gateway, String netmask) {
```
weizhouapache, line 1292: SharedNetwork instead of only Shared?

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -1462,12 +1480,12 @@ public Network createGuestNetwork(CreateNetworkCmd cmd) throws InsufficientCapac
             }
         }
 
-        if (ntwkOff.getGuestType() != GuestType.Shared && (!StringUtils.isAllBlank(routerIp, routerIpv6))) {
-            throw new InvalidParameterValueException("Router IP can be specified only for Shared networks");
+        if (ntwkOff.getGuestType() == GuestType.L2 && (!StringUtils.isAllBlank(cmd.getRouterIPv4(), cmd.getRouterIPv6()))) {
```
weizhouapache, line 1483: Is there a validation for isolated networks source 
nat support and capabilities here ?

api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
alexandremattioli: Correct

api/src/main/java/com/cloud/network/Network.java
alexandremattioli: +1 on SpecifySourceNatIp

api/src/main/java/org/apache/cloudstack/api/response/NetworkOfferingResponse.java
alexandremattioli: +1

engine/schema/src/main/java/com/cloud/offerings/NetworkOfferingVO.java
DaanHoogland: this would be extra work and this functionality is rather generic. It is not so that we want to allow this with one network plugin and not with the other.
So unless there is a good argument against it, I would like to keep it as is.

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
DaanHoogland: There isn´t one, but I think supported services is handled in another location. in my tests I've seen no problems so far. Also this change is only a small fix, allowing the router ip to be specified, which would already happen if you don´t start and isolated network and then allocate an new IP.

Maybe there is another check you are missing?

weizhouapache:
@DaanHoogland 
should this be in 4.18.0.0 milestone ?

DaanHoogland:
> @DaanHoogland should this be in 4.18.0.0 milestone ?

I will try to get it in, it doesn´t have high prio though, so only if it is proven correct before RC1

DaanHoogland:
@blueorangutan package

engine/schema/src/main/java/com/cloud/offerings/NetworkOfferingVO.java
weizhouapache: ok to me. @DaanHoogland 
can you rename the column name as well ?

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
weizhouapache: @DaanHoogland 
I was just wondering if there are checks to verify the source nat is supported by the network (or offering) and the specified IP is available. I did not find the checks.
I will have a look again. If you know the position of checks, it would be great.

DaanHoogland: I'll try to break it by creating an offering without source nat and see what I can break.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

weizhouapache:
@DaanHoogland 
Do you have time to update this PR before 4.18.0.0 cut ?

DaanHoogland:
> @DaanHoogland Do you have time to update this PR before 4.18.0.0 cut ?

@weizhouapache , I am investigating similar errors as the tungsten PR. that one has priority of course. I do not think there is anything to update anymore, but if so, I will try.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test keepEnv

DaanHoogland:
@blueorangutan package

rohityadavcloud:
Just FYI - there is a workaround that may even be implemented just in the UI. When creating a network in the UI, ask in the same form or as a next step the IP that the user wants, and acquire that public IP. The first IP acquired in an isolated network for ex. becomes the SNAT IP :) I've been using this approach for years ;)

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -1492,3 +1492,120 @@ CREATE TABLE IF NOT EXISTS `cloud`.`console_session` (
...
+        `cloud`.`data_center` AS `zone` ON FIND_IN_SET(`zone`.`id`, `zone_details`.`value`)
+            LEFT JOIN
+        `cloud`.`network_offering_details` AS `offering_details` ON `offering_details`.`network_offering_id` = `network_offerings`.`id` AND `offering_details`.`name` = 'internetProtocol'
+    GROUP BY
+        `network_offerings`.`id`;
+
+-- cloud.vpc_offering_view source
+
+CREATE OR REPLACE
+ALGORITHM = UNDEFINED VIEW `vpc_offering_view` AS
```
GutoVeronezi, line 1570:
```suggestion
DROP VIEW IF EXISTS `cloud`.`vpc_offering_view`;
CREATE VIEW `cloud`.`vpc_offering_view` AS
```

NuxRo:
@blueorangutan package

weizhouapache:
+1

> Just FYI - there is a UI-only workaround that may even be implemented (so just a UI feature). When creating a network in the UI, ask in the same form or as a next step the IP that the user wants, and acquire that public IP. The first IP acquired in an isolated network for ex. becomes the SNAT IP :) I've been using this approach for years ;)

DaanHoogland:
@blueorangutan package

GutoVeronezi:
> Just FYI - there is a UI-only workaround that may even be implemented (so just a UI feature). When creating a network in the UI, ask in the same form or as a next step the IP that the user wants, and acquire that public IP. The first IP acquired in an isolated network for ex. becomes the SNAT IP :) I've been using this approach for years ;)

It is good to have a workaround right away; however, I think we still should provide a clear and solid method for operators to define the source NAT of the network/VPC, as the workaround is only possible for not persistent networks and if we create VPCs without starting it.

I think that a better solution here would be to behave as the "add VM to network process": if no IP is provided, one is selected at random; if the IP is provided, validations are executed, and if everything is fine, the IP is used. We could receive the parameter via the `createVPC`/`createNetwork` APIs and acquire the IP for the VPC/network (if it is available) directly in the code; in the UI, we could implement the IP selection for the `Create VPC` and `Create Network` forms.

---

@DaanHoogland, is there any reason to consider `selecting the source NAT IP` as a capacity of the `source NAT` service? In my opinion, it is not related to the service supporting the functionality itself, but to the offering supporting the service; if the offering supports the `source NAT` service, then the source NAT IP can be informed, otherwise it cannot. For instance, with the current design, one could define a VPC offering without the `select the source NAT IP` capability, create the VPC without starting it, and still select the source NAT IP through the workaround @rohityadavcloud mentioned.

GutoVeronezi:
> @GutoVeronezi ,
> 
> > @DaanHoogland, is there any reason to consider `selecting the source NAT IP` as a capacity of the `source NAT` service?
> 
> In short because the offering having such a attribute when source NAT is not enabled makes no sense.
> 
> > In my opinion, it is not related to the service supporting the functionality itself, but to the offering supporting the service; if the offering supports the `source NAT` service, then the source NAT IP can be informed, otherwise it cannot. For instance, with the current design, one could define a VPC offering without the `select the source NAT IP` capability, create the VPC without starting it, and still select the source NAT IP through the workaround @rohityadavcloud mentioned.
> 
> Yes, one could. So do you want to prevent this, and how? delecting IPs must happen and the first one will always be the source NAT one, whether it is chosen or randomly assigned. I a, still trying to define the end functionality, but I think this is a good intermediate. Does that make sense?

@DaanHoogland, my point was that we would not need an attribute like `selecting the source NAT IP` in the offering; we could just validate if the offering supports the `SourceNat` service; if the offering supports the `SourceNat` service, then the source NAT IP could be informed while creating the network/VPC, otherwise, it could not. In my opinion, it is a simplier design to implement and the operators would not need to create new offerings to use the feature. What do you think?

weizhouapache:
> > @GutoVeronezi ,
> > > @DaanHoogland, is there any reason to consider `selecting the source NAT IP` as a capacity of the `source NAT` service?
> > 
> > 
> > In short because the offering having such a attribute when source NAT is not enabled makes no sense.
> > > In my opinion, it is not related to the service supporting the functionality itself, but to the offering supporting the service; if the offering supports the `source NAT` service, then the source NAT IP can be informed, otherwise it cannot. For instance, with the current design, one could define a VPC offering without the `select the source NAT IP` capability, create the VPC without starting it, and still select the source NAT IP through the workaround @rohityadavcloud mentioned.
> > 
> > 
> > Yes, one could. So do you want to prevent this, and how? delecting IPs must happen and the first one will always be the source NAT one, whether it is chosen or randomly assigned. I a, still trying to define the end functionality, but I think this is a good intermediate. Does that make sense?
> 
> @DaanHoogland, my point was that we would not need an attribute like `selecting the source NAT IP` in the offering; we could just validate if the offering supports the `SourceNat` service; if the offering supports the `SourceNat` service, then the source NAT IP could be informed while creating the network/VPC, otherwise, it could not. In my opinion, it is a simplier design to implement and the operators would not need to create new offerings to use the feature. What do you think?

I agree with @GutoVeronezi 
it may be unnecessary to add a capability to enable/disable SourceNat IP selection. 

As @rohityadavcloud said, currently users can do it by
- create network or VPC (but not start it)
- acquire a specific public IP
- create vm
The acquired Public IP will be the Source NAT IP.

DaanHoogland:
So what you are saying, @weizhouapache @GutoVeronezi , is that if source NAT is enabled on an offering, the user automatically has the possibility to choose a source NAT IP?

alexandremattioli:
Exactly that @DaanHoogland   If the SourceNAT service is enabled, then the end user should be able to change the IP.

DaanHoogland:
> @DaanHoogland, my point was that we would not need an attribute like `selecting the source NAT IP` in the offering; we could just validate if the offering supports the `SourceNat` service; if the offering supports the `SourceNat` service, then the source NAT IP could be informed while creating the network/VPC, otherwise, it could not. In my opinion, it is a simplier design to implement and the operators would not need to create new offerings to use the feature. What do you think?

makes sense, but I would not make it obligatory, if that is what you mean, just make it an optional parameter that is accepted and validated if source NAT is enabled.

GutoVeronezi:
> > @DaanHoogland, my point was that we would not need an attribute like `selecting the source NAT IP` in the offering; we could just validate if the offering supports the `SourceNat` service; if the offering supports the `SourceNat` service, then the source NAT IP could be informed while creating the network/VPC, otherwise, it could not. In my opinion, it is a simplier design to implement and the operators would not need to create new offerings to use the feature. What do you think?
> 
> makes sense, but I would not make it obligatory, if that is what you mean, just make it an optional parameter that is accepted and validated if source NAT is enabled.

Yes, it could behave as the "add VM to network process": if no IP is provided, one is selected at random; if the IP is provided, validations are executed, and if everything is fine, the IP is used.

DaanHoogland:
@blueorangutan package

DaanHoogland:
I am taking ipv6 addresses out of scope for this. Also I am slightly refactorring the interface tto allow for a UUID, though i think a UUID for an IP address makes no sense as the IP itself is already a unique identifyer. #legacy

DaanHoogland:
scope hase somewhat creeped, so I changed the title

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

api/src/main/java/org/apache/cloudstack/api/command/user/network/CreateNetworkCmd.java
```diff
@@ -183,6 +183,20 @@ public class CreateNetworkCmd extends BaseCmd implements UserCmd {
...
+                    " If an address is given and it cannot be acquired, an error will be returned and the network won´t be implemented," +
+                    " even if so requested.",
+            since = "4.19")
+    private String sourceNatIP;
+
+    @Parameter(name = ApiConstants.SOURCE_NAT_IP_ID,
+            type = CommandType.UUID,
+            description = "IPV4 address to be assigned to the public interface of the network router.",
+            since = "4.19")
+    private Long sourceNatIpUuid;
```
harikrishna-patnala, line 198: is this parameter in use ? I don't see any usage of this.

api/src/main/java/org/apache/cloudstack/api/command/user/network/UpdateNetworkCmd.java
```diff
@@ -104,6 +104,12 @@ public class UpdateNetworkCmd extends BaseAsyncCustomIdCmd implements UserCmd {
     @Parameter(name = ApiConstants.IP6_DNS2, type = CommandType.STRING, description = "the second IPv6 DNS for the network. Empty string will update the second IPv6 DNS with the value from the zone", since = "4.18.0")
     private String ip6Dns2;
 
+    @Parameter(name = ApiConstants.SOURCE_NAT_IP, type = CommandType.STRING, description = "IPV4 address to be assigned to the public interface of the network router. This address must already be acquired for this network", since = "4.19")
+    private String sourceNatIP;
+
+    @Parameter(name = ApiConstants.SOURCE_NAT_IP_ID, type = CommandType.UUID, description = "IPV4 address to be assigned to the piublic interface of the network router.", since = "4.19")
+    private Long sourceNatIpUuid;
```
harikrishna-patnala, line 111: same here, can you check if this is an used param and delete it. I think same is there in other commands as well

server/src/test/java/com/cloud/network/NetworkServiceImplTest.java
```diff
@@ -770,4 +773,144 @@ public void validateIfServiceOfferingIsActiveAndSystemVmTypeIsDomainRouterTestMu
...
+            Assert.assertTrue(e.getMessage().contains("Router IPv4 IP provided is of incorrect format"));
+            passing = true;
+        }
+        Assert.assertTrue(passing);
+    }
+
+    @Test
+    public void checkAndDontSetSourceNatIp() {
+        CreateNetworkCmd cmd = new CreateNetworkCmd();
+        String srcNatIp = "10.100.1000.1";
```
harikrishna-patnala, line 879: There is no usage for this ! can we delete this

harikrishna-patnala: String srcNatIp = "10.100.1000.1";
I meant about this. may be yes, you set it for readability. please ignore if that is the case.

DaanHoogland:
@blueorangutan package

harikrishna-patnala (approved):
thanks for addressing review comments @DaanHoogland 

Code LGTM

rohityadavcloud (approved):
LGTM didn't test it.

api/src/main/java/org/apache/cloudstack/api/response/IPAddressResponse.java
```diff
@@ -159,10 +159,9 @@ public class IPAddressResponse extends BaseResponseWithAnnotations implements Co
     @Param(description="the name of the Network where ip belongs to")
     private String networkName;
 
-    /*
-        @SerializedName(ApiConstants.JOB_ID) @Param(description="shows the current pending asynchronous job ID. This tag is not returned if no current pending jobs are acting on the volume")
-        private IdentityProxy jobId = new IdentityProxy("async_job");
-    */
+    @SerializedName(ApiConstants.HAS_RULES)
+    @Param(description="whether the ip address has Frewall-/PortForwarding-/LoadBalancing-rules defined")
```
GutoVeronezi, line 163:
```suggestion
    @Param(description="whether the ip address has Firewall/PortForwarding/LoadBalancing rules defined")
```

server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java
```diff
@@ -6207,34 +6209,32 @@ void validateLoadBalancerServiceCapabilities(final Map<Capability, String> lbSer
...
-                    if (!enabled && !disabled) {
-                        throw new InvalidParameterValueException("Unknown specified value for " + Capability.RedundantRouter.getName());
-                    }
-                } else {
-                    throw new InvalidParameterValueException("Only " + Capability.SupportedSourceNatTypes.getName() + " and " + Capability.RedundantRouter
-                            + " capabilities can be sepcified for source nat service");
+        if (MapUtils.isNotEmpty(sourceNatServiceCapabilityMap) && (sourceNatServiceCapabilityMap.size() > 2 || ! sourceNatCapabilitiesContainValidValues(sourceNatServiceCapabilityMap))) {
+            throw new InvalidParameterValueException("Only " + Capability.SupportedSourceNatTypes.getName()
+                    + ", " + Capability.RedundantRouter
+                    + " capabilities can be sepcified for source nat service");
```
GutoVeronezi, line 6215:
```suggestion
                    + " capabilities can be specified for source nat service");
```

server/src/test/java/com/cloud/configuration/ConfigurationManagerTest.java
```diff
@@ -547,16 +548,68 @@ public void searchForNetworkOfferingsTest() {
...
+    @Test
+    public void validateInvalidCapabilityForSourceNatServiceCapablitiesTest() {
+        Map<Capability, String> sourceNatServiceCapabilityMap = new HashMap<>();
+        sourceNatServiceCapabilityMap.put(Capability.ElasticIp, "perDomain");
+
+        boolean caught = false;
+        try {
+            configurationMgr.validateSourceNatServiceCapablities(sourceNatServiceCapabilityMap);
+        } catch (InvalidParameterValueException e) {
+            Assert.assertTrue(e.getMessage(), e.getMessage().contains("Only SupportedSourceNatTypes, Network.Capability[name=RedundantRouter] capabilities can be sepcified for source nat service"));
```
GutoVeronezi, line 597:
```suggestion
            Assert.assertTrue(e.getMessage(), e.getMessage().contains("Only SupportedSourceNatTypes, Network.Capability[name=RedundantRouter] capabilities can be specified for source nat service"));
```

GutoVeronezi (commented):
@DaanHoogland, could we improve the PR description with more details?

Also, if we are going to allow updating the source NAT IP of a network/VPC, some tests will be needed to guarantee the changes are being propagated properly.

DaanHoogland:
> @DaanHoogland, could we improve the PR description with more details?
> 
> Also, if we are going to allow updating the source NAT IP of a network/VPC, some tests will be needed to guarantee the changes are being propagated properly.

@GutoVeronezi They will not be propagted until the user restarts the network. I will ad more unit test and think of an appropriate integration test.

kiranchavala:
@blueorangutan package

.github/workflows/ci.yml
```diff
@@ -79,7 +79,8 @@ jobs:
                   smoke/test_metrics_api
                   smoke/test_migration
                   smoke/test_multipleips_per_nic
-                  smoke/test_nested_virtualization",
+                  smoke/test_nested_virtualization
+                  smoke/test_set_source_nat",
```
harikrishna-patnala, line 83: test_set_source_nat.py smoke test file is missing !

api/src/main/java/org/apache/cloudstack/api/response/IPAddressResponse.java
```diff
@@ -159,10 +159,9 @@ public class IPAddressResponse extends BaseResponseWithAnnotations implements Co
     @Param(description="the name of the Network where ip belongs to")
     private String networkName;
 
-    /*
-        @SerializedName(ApiConstants.JOB_ID) @Param(description="shows the current pending asynchronous job ID. This tag is not returned if no current pending jobs are acting on the volume")
-        private IdentityProxy jobId = new IdentityProxy("async_job");
-    */
+    @SerializedName(ApiConstants.HAS_RULES)
+    @Param(description="whether the ip address has Firewall/PortForwarding/LoadBalancing rules defined")
```
harikrishna-patnala, line 163: please help me understand where this is used or how is it helpful. Also I see we are currently setting this flag only based on firewall rules (not portforwarding/loadbalancing rules), do we need to update the comment !

ipResponse.setHasRules(firewallRulesDao.countRulesByIpId(ipAddr.getId()) > 0);

DaanHoogland: This is used to determine if the "make source nat"button should be activated.

(I think it is base on all rules, independent of type. I'll have a look to see if I missed something.

.github/workflows/ci.yml
DaanHoogland: oops

api/src/main/java/org/apache/cloudstack/api/response/IPAddressResponse.java
DaanHoogland: @harikrishna-patnala firewall rules contains a rule type and will store all rules (also for portforwarding and loadbalancing) This should work.

kiranchavala:
@blueorangutan package

kiranchavala (approved):
LGTM, tested the feature and all the scenarios 

[Configurable SNAT IP for VR (test cases).xlsx](https://github.com/apache/cloudstack/files/11430562/Configurable.SNAT.IP.for.VR.test.cases.xlsx)

kiranchavala:
@blueorangutan package

DaanHoogland:
@weizhouapache @GutoVeronezi @alexandremattioli , I am assuming your comments have all been addressed?

weizhouapache:
> @weizhouapache @GutoVeronezi @alexandremattioli , I am assuming your comments have all been addressed?

@DaanHoogland 
lgtm

DaanHoogland:
@blueorangutan test

weizhouapache:
@DaanHoogland 
I tested this feature, overall it looks nice !

two issues
- update network with a public IP did nothing, if the IP is not acquired before. Would be good to display an error message ?
- the network router is destroyed and then recreated, which caused 43 seconds downtime (static nat). this can be improved I think.

DaanHoogland:
> * update network with a public IP did nothing, if the IP is not acquired before. Would be good to display an error message ?

that seems a bug. It should throw an exception in this case.

ui/src/views/network/CreateIsolatedNetworkForm.vue
```diff
@@ -291,6 +291,36 @@
...
+            ref="account"
+            name="account"
+            v-if="accountVisible">
+            <template #label>
+              <tooltip-label :title="$t('label.account')" :tooltip="apiParams.account.description"/>
+            </template>
+            <a-input
+             v-model:value="form.account"
+              :placeholder="apiParams.account.description"/>
+          </a-form-item>
```
BryanMLima, line 323: @DaanHoogland, this code seems duplicated, considering that the network domain and account fields are already present in the form. Is this on purpose?

While working on PR #8919, there is no option to create a network for a project, so I was thinking of adding the `Ownership` component to this view as well and removing this duplicate code.

DaanHoogland: Sounds great @BryanMLima , I think this was not intentional. I just wonder about this dialog containing a "network domain" twice, while it should be a "cloudstack domain" and a real network domain. So a bit of semantic hacking needed ;) Thanks.

Issue #6427, opened 2022-05-31
Broken SAML SSO plugin with malformed URL when authenticating against Google IdP


Luis-3M:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
SAML SSO plugin
~~~

##### CLOUDSTACK VERSION
~~~
LATEST -> https://github.com/apache/cloudstack/blame/main/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java#L153
~~~

##### CONFIGURATION
SAML SSO plugin activated.

##### OS / ENVIRONMENT
N/A

##### SUMMARY
Google's IdP URL already contains a query param, thus this line of code will create a malformed redirect URL.
https://github.com/apache/cloudstack/blame/main/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java#L153

Example of google's IdP URL: `https://accounts.google.com/o/saml2/idp?idpid=XXXXXXXXX`

##### STEPS TO REPRODUCE
- Enable and configure SAML SSO plugin2 using Google as the IdP.
- Access the UI and attempt to login via Single Sign-On

You should now face an Invalid Request error.

##### EXPECTED RESULTS
Users should be redirected to Google's login portal thus completing the authentication flow.

##### ACTUAL RESULTS
![Screenshot 2022-05-31 at 11 08 55](https://user-images.githubusercontent.com/26753403/171154930-8a54162c-6552-458f-9e53-0ea2bb28ddb3.png)

~~~
Malformed URL: 
https://accounts.google.com/o/saml2/idp?idpid=<ID>?SAMLRequest=<SAMLRequest>
~~~

~~~
Expected URL:
https://accounts.google.com/o/saml2/idp?idpid=<ID>&SAMLRequest=<SAMLRequest>
~~~

Luis-3M:
Hi @rohityadavcloud ,
Relevant fields of our IdP Metadata are the following:

`<md:SingleSignOnService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect" Location="https://accounts.google.com/o/saml2/idp?idpid=XXXXXXX"/>`

`<md:SingleSignOnService Binding="urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST" Location="https://accounts.google.com/o/saml2/idp?idpid=XXXXXXX"/>`

[...]

PR #6451, opened 2022-06-11
fix google idp saml redirect url

neogismm:
Fixes #6427 

This PR fixes the redirect URL for SAML authentication while using Google as IdP.

Note: I could not test my fix because choosing google as idp for a user requires the SSO url which also contains the unique IDP id created by google. 

Example:
SSO URL: https://accounts.google.com/o/saml2/idp?dpid={IDP-ID}
Entity ID: https://accounts.google.com/o/saml2?idpid={IDP-ID}

To set up a custom google saml app requires a google workspace account which was not possible for me to create at the moment.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

PR #6457, opened 2022-06-14
Fix SAML SSO plugin redirect URL

Luis-3M:
### Description

This PR fixes the issue #6427 -> SAML request must be appended to an IdP URL as a query param with an ampersand, if the URL already contains a question mark, as opposed to always assume that IdP URLs don't have any query params.
Google's IdP URL for instance looks like this: `https://accounts.google.com/o/saml2/idp?idpid=<ID>`, therefore the expected redirect URL would be `https://accounts.google.com/o/saml2/idp?idpid=<ID>&SAMLRequest=<SAMLRequest>`

This code change is backwards compatible with the current behaviour.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

Screenshot 2022-06-14 at 14 02 28

Screenshot 2022-06-14 at 14 01 48

### DIFF
--- a/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
+++ b/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
@@ -150,7 +150,7 @@ public class SAMLUtils {
             if (spMetadata.getKeyPair() != null) {
                 privateKey = spMetadata.getKeyPair().getPrivate();
             }
-            redirectUrl = idpMetadata.getSsoUrl() + "?" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm);
+            redirectUrl = idpMetadata.getSsoUrl().contains("?") ? idpMetadata.getSsoUrl() + "&" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm) : idpMetadata.getSsoUrl() + "?" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm);
         } catch (ConfigurationException | FactoryConfigurationError | MarshallingException | IOException | NoSuchAlgorithmException | InvalidKeyException | java.security.SignatureException e) {
             s_logger.error("SAML AuthnRequest message building error: " + e.getMessage());
         }
--- a/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
+++ b/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
@@ -150,7 +150,8 @@ public class SAMLUtils {
             if (spMetadata.getKeyPair() != null) {
                 privateKey = spMetadata.getKeyPair().getPrivate();
             }
-            redirectUrl = idpMetadata.getSsoUrl().contains("?") ? idpMetadata.getSsoUrl() + "&" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm) : idpMetadata.getSsoUrl() + "?" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm);
+            String appendOperator = idpMetadata.getSsoUrl().contains("?") ? "&" : "?";
+            redirectUrl = idpMetadata.getSsoUrl() + appendOperator + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm);
         } catch (ConfigurationException | FactoryConfigurationError | MarshallingException | IOException | NoSuchAlgorithmException | InvalidKeyException | java.security.SignatureException e) {
             s_logger.error("SAML AuthnRequest message building error: " + e.getMessage());
         }
--- a/plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
+++ b/plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
@@ -19,20 +19,22 @@
 
 package org.apache.cloudstack;
 
+import junit.framework.TestCase;
+import org.apache.cloudstack.saml.SAML2AuthManager;
+import org.apache.cloudstack.saml.SAMLProviderMetadata;
+import org.apache.cloudstack.saml.SAMLUtils;
+import org.apache.cloudstack.utils.security.CertUtils;
+import org.junit.Test;
+import org.opensaml.saml2.core.AuthnRequest;
+import org.opensaml.saml2.core.LogoutRequest;
+
+import java.net.URI;
 import java.security.KeyPair;
 import java.security.PrivateKey;
 import java.security.PublicKey;
 import java.util.regex.Pattern;
 
-import org.apache.cloudstack.saml.SAML2AuthManager;
-import org.apache.cloudstack.saml.SAMLUtils;
-import org.apache.cloudstack.utils.security.CertUtils;
-import org.junit.Test;
-import org.opensaml.DefaultBootstrap;
-import org.opensaml.saml2.core.AuthnRequest;
-import org.opensaml.saml2.core.LogoutRequest;
-
-import junit.framework.TestCase;
+import static org.assertj.core.api.Assertions.assertThat;
 
 public class SAMLUtilsTest extends TestCase {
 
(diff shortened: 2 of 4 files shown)


nvazquez:
@Luis-3M @neogismm it seems we have 2 PRs solving the same issue, can you work on unifying them? (#6451)

Luis-3M:
@nvazquez Yep happy to close mine if @neogismm keeps #6451 open 👍

neogismm:
Hi @Luis-3M thanks for the PR. I was stuck on how to write the tests for the same and looking at your code helps me to learn.

Luis-3M:
No worries at all @neogismm - thank you.

Luis-3M:
@nvazquez we will keep this one open.

plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/saml/SAMLUtils.java
```diff
@@ -150,7 +150,7 @@ public static String buildAuthnRequestUrl(final String authnId, final SAMLProvid
             if (spMetadata.getKeyPair() != null) {
                 privateKey = spMetadata.getKeyPair().getPrivate();
             }
-            redirectUrl = idpMetadata.getSsoUrl() + "?" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm);
+            redirectUrl = idpMetadata.getSsoUrl().contains("?") ? idpMetadata.getSsoUrl() + "&" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm) : idpMetadata.getSsoUrl() + "?" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm);
```
DaanHoogland, line 153: for readibility :
```suggestion
            String appendOperator = idpMetadata.getSsoUrl().contains("?") ? "&" : "?";
            redirectUrl = idpMetadata.getSsoUrl()
                 + appendOperator
                 + SAMLUtils.generateSAMLRequestSignature("SAMLRequest="
                     + SAMLUtils.encodeSAMLRequest(authnRequest), privateKey, signatureAlgorithm);
```

Luis-3M: Addressed https://github.com/apache/cloudstack/pull/6457/commits/749dca23b7fe86c4dbf811c1afe14ef5f444f413

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
```diff
@@ -60,6 +62,32 @@ public void testBuildAuthnRequestObject() throws Exception {
...
+    public void testBuildAuthnRequestUrlWithQueryParam() throws Exception {
+        String consumerUrl = "http://someurl.com";
+        String idpUrl = "http://idp.domain.example?idpid=CX1298373";
+        String spId = "cloudstack";
+        String authnId = SAMLUtils.generateSecureRandomId();
+        DefaultBootstrap.bootstrap();
+        AuthnRequest req = SAMLUtils.buildAuthnRequestObject(authnId, spId, idpUrl, consumerUrl);
+        String appendOperator = idpUrl.contains("?") ? "&" : "?";
+        String redirectUrl = idpUrl + appendOperator + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(req), null, SAML2AuthManager.SAMLSignatureAlgorithm.value());
+        assertEquals(redirectUrl, idpUrl + "&" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(req), null, SAML2AuthManager.SAMLSignatureAlgorithm.value()));
```
DaanHoogland, line 88: same here, this does not seem to test and ACS code, even thought it calls `SAMLUtils.buildAuthnRequestObject(..)`, it seems to take its output at face value.

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
```diff
@@ -60,6 +62,32 @@ public void testBuildAuthnRequestObject() throws Exception {
...
+    public void testBuildAuthnRequestUrlWithoutQueryParam() throws Exception {
+        String consumerUrl = "http://someurl.com";
+        String idpUrl = "http://idp.domain.example";
+        String spId = "cloudstack";
+        String authnId = SAMLUtils.generateSecureRandomId();
+        DefaultBootstrap.bootstrap();
+        AuthnRequest req = SAMLUtils.buildAuthnRequestObject(authnId, spId, idpUrl, consumerUrl);
+        String appendOperator = idpUrl.contains("?") ? "&" : "?";
+        String redirectUrl = idpUrl + appendOperator + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(req), null, SAML2AuthManager.SAMLSignatureAlgorithm.value());
+        assertEquals(redirectUrl, idpUrl + "?" + SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(req), null, SAML2AuthManager.SAMLSignatureAlgorithm.value()));
```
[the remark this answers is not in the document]
Luis-3M: @DaanHoogland These two tests are not much different from `testBuildAuthnRequestObject`, except that each one does an assertion on the two possible results of the redirectUrl, based on the code change added in this PR.
Thoughts on how should we implement these tests ?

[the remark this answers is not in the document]
DaanHoogland: Not sure @Luis-3M , but now they are not testing anything in the production code except maybe consistence of the result of `SAMLUtils.generateSAMLRequestSignature("SAMLRequest=" + SAMLUtils.encodeSAMLRequest(req), null, SAML2AuthManager.SAMLSignatureAlgorithm.value())`, as you are calling it twice and coparing the results of both calls concatenated to something that can be read from this code to be the same. If this function breaks for some reason this test is still going to pass.
`idpUrl` is the same and `appendOperator` is predictable from what you make `idpUrl`.
Depending on what you would test you would hard code the expected result and let the production code produce the same for comparison. I am not sure what the exact format of an SML url would look like and I suppose there would be some part that is unpredictable (e.g. a one time token) so whether and how to test this. I would remove this test if we cannot hardcode anything, as it is giving us false security this way.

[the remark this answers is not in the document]
Luis-3M: > I suppose there would be some part that is unpredictable (e.g. a one time token) so whether and how to test this
True we can not predict the value of the SAMLRequest.

Based on the original issue, we know that the result of the `buildAuthnRequestUrl` method must have at least one query param: `SAMLRequest`. The original code was simply appending this same query param with `?`, therefore if the IDP URL already contains at least one query param, then `SAMLRequest` would actually be missing as seen here:

image

Think that we should at least test the existence of this and other parameters already part of the IDP URL, if any, as well as some basic URL components.

Let me know your thoughts @DaanHoogland .

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
```diff
@@ -60,6 +64,68 @@ public void testBuildAuthnRequestObject() throws Exception {
...
+        SAMLProviderMetadata idpMetadata = new SAMLProviderMetadata();
+        idpMetadata.setSsoUrl(idpUrl);
+        idpMetadata.setEntityId(idpId);
+
+        URI redirectUrl = new URI(SAMLUtils.buildAuthnRequestUrl(authnId, spMetadata, idpMetadata, SAML2AuthManager.SAMLSignatureAlgorithm.value()));
+        assertThat(redirectUrl).hasScheme(urlScheme);
+        assertEquals(urlScheme, redirectUrl.getScheme());
+        assertThat(redirectUrl).hasHost(idpDomain);
+        assertEquals(idpDomain, redirectUrl.getHost());
+        assertThat(redirectUrl).hasParameter("SAMLRequest");
```
DaanHoogland, line 94:
```suggestion
        assertThat(redirectUrl).hasScheme(urlScheme).hasHost(idpDomain).hasParameter("SAMLRequest");
        assertEquals(urlScheme, redirectUrl.getScheme());
        assertEquals(idpDomain, redirectUrl.getHost());
```

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
```diff
@@ -60,6 +64,68 @@ public void testBuildAuthnRequestObject() throws Exception {
...
+        idpMetadata.setSsoUrl(idpUrl);
+        idpMetadata.setEntityId(idpId);
+
+        URI redirectUrl = new URI(SAMLUtils.buildAuthnRequestUrl(authnId, spMetadata, idpMetadata, SAML2AuthManager.SAMLSignatureAlgorithm.value()));
+        assertThat(redirectUrl).hasScheme(urlScheme);
+        assertEquals(urlScheme, redirectUrl.getScheme());
+        assertThat(redirectUrl).hasHost(idpDomain);
+        assertEquals(idpDomain, redirectUrl.getHost());
+        assertThat(redirectUrl).hasParameter("idpid");
+        assertThat(redirectUrl).hasParameter("SAMLRequest");
```
DaanHoogland, line 126:
```suggestion
        assertThat(redirectUrl).hasScheme(urlScheme).hasHost(idpDomain).hasParameter("idpid").hasParameter("SAMLRequest");
        assertEquals(urlScheme, redirectUrl.getScheme());
        assertEquals(idpDomain, redirectUrl.getHost());
```

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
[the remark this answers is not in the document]
DaanHoogland: your tests look much cleaner now @Luis-3M , and they seem to actually test the return of the production method, thanks a lot. We can always improve later if we feel we should assert something else.

DaanHoogland (approved):

harikrishna-patnala:
@Luis-3M Code LGTM. If you can please address one code smell as suggested by Daan, that will look pefect.

harikrishna-patnala (approved):

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
Luis-3M: Addressed https://github.com/apache/cloudstack/pull/6457/commits/2deaefb8d7120b9ee48adebdd18b6fd28c7969b3

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
Luis-3M: Addressed https://github.com/apache/cloudstack/pull/6457/commits/2deaefb8d7120b9ee48adebdd18b6fd28c7969b3

Luis-3M:
Thanks for your input @DaanHoogland @harikrishna-patnala .
Suggested changes have been addressed.

plugins/user-authenticators/saml2/src/test/java/org/apache/cloudstack/SAMLUtilsTest.java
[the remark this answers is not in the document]
Luis-3M: No worries at all @DaanHoogland . Thank you for your input on this.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test keepEnv

nvazquez (approved):

PR #6505, opened 2022-06-27
Gives the possibility to redirect to external links when the property is defined

RodrigoDLopez:
### Description

There is currently no way in ACS to redirect users to external links. To resolve the situation, one new property has been added to the `plugin` component inside `config.json` settings file so that redirects to external links can be defined. The `isExternalLink` property define if the component will be shown as an external link in the top corner of the page, or if it will be treated as a plugin. If the property is omitted the component will be treated as a plugin.

```
"plugins": [
  {
    "name": "some testing",
    "icon": "cloud",
    "path": "https://cloudstack.apache.org/api/apidocs-4.17/"
  },
  {
    "name": "Cloudstack",
    "icon": "myIcons/cloudstach.png",
    "path": "https://cloudstack.apache.org/",
    "isExternalLink": "true"
  },
  {
    "name": "",
    "icon": "",
    "path": "",
    "isExternalLink": ""
  },
  {
    "name": "Apache",
    "icon": "https://www.apache.org/icons/apache.png",
    "path": "https://apache.org/",
    "isExternalLink": "true"
  }
],
```

The following rules govern the redirection feature for external links:
    - If the property is not defined, the redirect button will not be displayed to the user;
    - The `path` attribute of the `plugin` component is mandatory in this context, and elements of this property that do not have the `path` attribute or are empty will be disregarded. The `name` and `icon` attributes are optional, if an icon is not defined a default icon will be applied;
    - When the `name` attribute is not set, the `path` will be displayed instead.

Once the `name`, `path`, `icon` and `isExternalLink` attributes are informed, two possible behaviors can be observed:

1. Only one defined element: in this context, a button will be displayed, which, when clicked, will redirect the user to the `path` defined in the `plugin` component;

image1

2. More than one element is defined: a dropdown containing the configured `paths` will be displayed.

image2

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?
Changing the configuration file to get one or a list of external links


ui/public/config.json
```diff
@@ -59,5 +67,119 @@
   "basicZoneEnabled": true,
   "multipleServer": false,
   "allowSettingTheme": true,
-  "docHelpMappings": {}
+  "docHelpMappings": {
```
shwstppr, line 70: @RodrigoDLopez doc help mappings are auto-generated while building UI and shouldn't be a part of this PR

DaanHoogland: @shwstppr can you explain how to deal with having such an extended list of documentation links, please? (I am not saying that these are appropriate or not and should stay in or not, just curious)

shwstppr: @DaanHoogland This list is auto-generated when building UI https://github.com/apache/cloudstack/blob/main/ui/prebuild.sh#L19-L47.
It is cleared and set to empty once `npm run build` finishes. At times when the build is aborted or fails the list doesn't get cleared. It must've been added by author by mistake.

RodrigoDLopez: oh. i didn't see that.
thanks for the review.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test keepEnv

shwstppr (commented):
-0

Tested functionality with QA server, it works.
I'm not sure about its use-case. But I'm a dev and other users can pitch in. Right now UI is just showing icon/menu for links without much context.
@RodrigoDLopez it might be a good idea to add some documentation for the users of this feature

DaanHoogland:
> @DaanHoogland this is purely a UI PR, not sure why tests were needed for this :)

I was going to look into editting the `config,json` and would like to do that in an isolated env, @shwstppr

DaanHoogland (approved):
tried and tested

ui/public/config.json
```diff
@@ -18,6 +18,14 @@
     "404": "assets/404.png",
     "500": "assets/500.png"
   },
+  "externalLinksIcon": "",
+  "externalLinks": [
```
rohityadavcloud, line 22: `plugins` feature in config.json does exactly the same I think cc @utchoang can you review and check this?

utchoang (approved):
Tested locally! Not sure if this is appropriate but I think it's fine. LGTM!

rohityadavcloud (changes requested):
`plugins` feature in config.json does exactly the same - let's review if this is the same feature or something different, does it make sense to expand that instead?

RodrigoDLopez:
> `plugins` feature in config.json does exactly the same - let's review if this is the same feature or something different, does it make sense to expand that instead?

I did not catch what you mean, which feature in the config.json do the same as proposed by this PR?

RodrigoDLopez:
@rohityadavcloud, @shwstppr 
is there anything else missing on this one, or we are good to go?

GutoVeronezi:
@RodrigoDLopez, I think what @rohityadavcloud means is that in file `/ui/public/config.json` there is a property to define third party plugins in the UI. It does not work exactly the same; however, it is similar. Here is a simple example of how to use the plugins to achieve something similiar to what you proposed: 
 
- In `config.json` I added the following: 
```json 
  ... 
  "plugins": [ 
            {"name": "some testing", "icon": "cloud", "path": "https://cloudstack.apache.org/api/apidocs-4.17/"} 
  ], 
  ... 
``` 
 
Then, `/ui/src/config/router.js` will handle this property, creating a new menu for each entry that will address it in an `IFrame`: 
 
``` 
  const plugins = Vue.prototype.$config.plugins 
  if (plugins && plugins.length > 0) { 
    plugins.map(plugin => { 
      routerMap[0].children.push({ 
        path: '/plugins/' + plugin.name, 
        name: plugin.name, 
        component: IFramePlugin, 
        meta: { title: plugin.name, icon: plugin.icon, path: plugin.path } 
      }) 
    }) 
  } 
 
  return routerMap 
} 
``` 
 
The result is: 
 
image 
 
The [official documentation] has more examples on how to use it. 
 
I like the result that your proposal generates (a single icon or list of icons redirecting the user). In some cases it can be better aesthetically (and functionally) for the user than a new menu opening the page inside ACS. 
 
With that said, I think we could extend the `plugins` functionallity to be able to generate the kind of result you proposed instead of creating a new set of properties. What do you think? 
 
 
[official documentation]: https://docs.cloudstack.apache.org/en/latest/developersguide/plugins.html#third-party-ui-plugins

RodrigoDLopez:
Thanks, @GutoVeronezi 

Based on your suggestion I added a new attribute called `isExternalLink` to `plugin`; This way, the default plugin behavior is kept as creating a new menu and, when identifying the object with the new attribute `isExternalLink` as `true`, we will create a component (a single button or a list of links), at the top-right corner of the page, to redirect users to the link configured. 
 
Also, I noticed that if an empty component is inserted (without link or name), `router.js` will create a new empty entry in the left menu, which will do nothing. Therefore, I made a small change to avoid this behavior and only create the menu when at least the name and link are provided.

RodrigoDLopez:
@DaanHoogland, @rohityadavcloud, @utchoang and @shwstppr 
Could you, please, check out this new proposal?

RodrigoDLopez:
@DaanHoogland, @rohityadavcloud, @utchoang and @shwstppr
is there any suggestion or request on this or are we ready to go?

ui/public/config.json
```diff
@@ -18,6 +18,14 @@
...
+    {
+      "title": "Example",
+      "link": "",
+      "icon": ""
+    }
+  ],
```
shwstppr, line 28: @RodrigoDLopez is this needed now with your new changes?

RodrigoDLopez: no, it's not. Thanks

DaanHoogland:
I think the only hesitation was by @rohityadavcloud . I'm good to go (or do you have reservations still @shwstppr ?)

shwstppr (approved):
Tested locally
Screenshot from 2022-12-22 17-28-17

PR #6506, opened 2022-06-27
externalizes used protocol to upload isos, volumes and templates

RodrigoDLopez:
### Description

CloudStack has the functionality to upload volumes, templates, and local ISOs. To execute the process, ACS generates a link to the secondary storage (using its public IPs), where the resource must be sent. The *links* are always generated with the HTTPS protocol and the environment has to be configured to handle HTTPS, requiring us to have SSL certificates. However, for environments that do not need to use certificates (such as staging ones), via the UI, we need a workaround (get the URL generated by the API for generating upload links, open it in a new tab, and accept an insecure connection with the application).  

The ACS was changed so that it also handles HTTP to upload volumes/images/ISOs according to the global configuration `use.https.to.upload`; then operators can choose whether to use the HTTPS or HTTP protocol to generate upload links. By default (true) HTTPS will be used.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?
In order to test the behavior, the value of the `use.https.to.upload` configuration was changed to `FALSE` and then checked if the upload of ISOs, volumes, and templates was carried out using the HTTP protocol.

### DIFF
--- a/api/src/main/java/com/cloud/storage/VolumeApiService.java
+++ b/api/src/main/java/com/cloud/storage/VolumeApiService.java
@@ -46,6 +46,14 @@ public interface VolumeApiService {
             , true // not sure if this is to be dynamic
             , ConfigKey.Scope.Global);
 
+    ConfigKey<Boolean> UseHttpsToUpload = new ConfigKey<Boolean>("Advanced",
+            Boolean.class,
+            "use.https.to.upload",
+            "true",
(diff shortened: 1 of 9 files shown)


DaanHoogland:
@blueorangutan package

sureshanaparti:
@blueorangutan test

DaanHoogland (approved):
code looks good
are any integrations tests available/possible for this?

JoaoJandre (approved):
CLGTM

PR #6510, opened 2022-06-27
Fix API deleteTrafficType not filtering physical network

GutoVeronezi:
### Description

While deleting a traffic type, ACS validates if there is any VM related to it. However, if we have several physical networks containing a traffic type, ACS does not filter the physical network to do the validation. For instance, if we have two (2) physical networks containing the traffic type Guest, the first one having VMs related, and the second not having VMs related, if we try to remove the second traffic type, ACS give us the message `The Traffic Type is not deletable because there are existing networks with this traffic type:Guest`.

The API `deleteTrafficType` was designed to filter the physical network where the traffic type is, however, due to a typo this filtering was not been applied correctly. This PR intends to fix this typo to honor the API behavior.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

In an advanced zone I created 4 physical networks, one for each traffic type (Public, Guest, Management, Storage). I instantiated some VMs so they get guest IPs. In the Public physical network I added a Guest traffic type. I tried to remove the new Guest traffic type from Public physical network, which did not have any VMs related to it, and, before the changes, I was getting the message `The Traffic Type is not deletable because there are existing networks with this traffic type:Guest`. After the changes, I could remove successfully the new Guest traffic type via API deleteTrafficType. I also tried to remove the Guest traffic type which had VMs related to it, however, as expected, I received the `The Traffic Type is not deletable...` message.

I also created a unit test to validate the data retrieving.

### DIFF
--- a/engine/schema/src/main/java/com/cloud/network/dao/NetworkDaoImpl.java
+++ b/engine/schema/src/main/java/com/cloud/network/dao/NetworkDaoImpl.java
@@ -572,7 +572,7 @@ public class NetworkDaoImpl extends GenericDaoBase<NetworkVO, Long>implements Ne
     public List<NetworkVO> listByPhysicalNetworkTrafficType(final long physicalNetworkId, final TrafficType trafficType) {
         final SearchCriteria<NetworkVO> sc = AllFieldsSearch.create();
         sc.setParameters("trafficType", trafficType);
-        sc.setParameters("physicalNetworkId", physicalNetworkId);
+        sc.setParameters("physicalNetwork", physicalNetworkId);
         return listBy(sc);
     }
 
--- a/engine/schema/src/test/java/com/cloud/network/dao/NetworkDaoImplTest.java
+++ b/engine/schema/src/test/java/com/cloud/network/dao/NetworkDaoImplTest.java
@@ -1,0 +1,47 @@
+package com.cloud.network.dao;
+
+import com.cloud.network.Networks;
+import com.cloud.utils.db.SearchBuilder;
+import com.cloud.utils.db.SearchCriteria;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+import java.util.List;
+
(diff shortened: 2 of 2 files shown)


engine/schema/src/test/java/com/cloud/network/dao/NetworkDaoImplTest.java
```diff
@@ -0,0 +1,47 @@
+package com.cloud.network.dao;
```
DaanHoogland, line 1: forgot a license :zap:

GutoVeronezi: Done, thanks @DaanHoogland

engine/schema/src/main/java/com/cloud/network/dao/NetworkDaoImpl.java
```diff
@@ -572,7 +572,7 @@ public List<NetworkVO> listByPhysicalNetwork(final long physicalNetworkId) {
     public List<NetworkVO> listByPhysicalNetworkTrafficType(final long physicalNetworkId, final TrafficType trafficType) {
         final SearchCriteria<NetworkVO> sc = AllFieldsSearch.create();
         sc.setParameters("trafficType", trafficType);
-        sc.setParameters("physicalNetworkId", physicalNetworkId);
+        sc.setParameters("physicalNetwork", physicalNetworkId);
```
DaanHoogland, line 575: this is strange, the field in both `NetworkVO` and in `PhysicalNetworkTrafficType` is called `physicalNetworkId`. I would say the new value is the typo. Do you know why this should work this way @GutoVeronezi ?

DaanHoogland: yes, you are right, very confusing, but I shouldn´t rely on naming conventions ;)

DaanHoogland (approved):

sureshanaparti (approved):
code LGTM

sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan test

rohityadavcloud (approved):

PR #6552, opened 2022-07-15
removed the use of SharedMountPoint storage type for the StorPool plugin

### LINKED ISSUE #6455: Bug : SharedMountPoint not working after StorPool plugin support
After adding support for StorPool in CloudStack version 4.17, Default Primary Storage of SharedMountPoint Type is no longer created. This Bug is related and caused by #6007 

This is an issue caused by the StoragePoolType of the StorPoolStorageAdaptor.

Currently, the Pool Type of StorPoolStorageAdaptor is set to SharedMountPoint.
See the following code.

StorPoolStorageAdaptor : 
~~~
@StorageAdaptorInfo(storagePoolType=StoragePoolType.SharedMountPoint)
public class StorPoolStorageAdaptor implements StorageAdaptor {
    public static void SP_LOG(String fmt, Object... args) {
    ......
~~~

For this reason, even if it is a Default SharedMountPoint type of storage other than StorPool, LibvirtStorageAdaptor is not selected by the following code, but StorPoolStorageAdaptor is selected and an error occurs.

KVMStoragePoolManager :
~~~
    private StorageAdaptor getStorageAdaptor(StoragePoolType type) {
        // type can be null: LibVirtComputingResource:3238
        if (type == null) {
            return _storageMapper.get("libvirt");
        }
        StorageAdaptor adaptor = _storageMapper.get(type.toString());
        if (adaptor == null) {
            // LibvirtStorageAdaptor is selected by default
            adaptor = _storageMapper.get("libvirt");
        }
        return adaptor;
    }
~~~

In my opinion, it is better to create a separate Pool Type for StorPoolStorageAdaptor like the LINSTOR Plugin.

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI, API, Plugin
~~~

##### CLOUDSTACK VERSION

~~~
4.17.0
~~~

##### CONFIGURATION

3 Hosts, GFS2 Clustered File System, Shared Mount Point : /mnt/gfs2

##### OS / ENVIRONMENT

CentOS 8.5, CloudStack 4.17.0

##### STEPS TO REPRODUCE

~~~
1. AddPrimaryStorage 
2. Select SharedMountPoint Protocol
3. Input Mount Directory
4. Select DefaultPrimary Provider
5. Submit 
~~~

##### EXPECTED RESULTS

Create Shared Mount Point Primary Storage

##### ACTUAL RESULTS

SharedMountPoint Primary Storage is recorded in DB, but the pool is not actually created because an appropriate StorageAdaptor is not selected.

slavkap:
### Description

This PR fixes #6455 

The default storage adaptor - `LibvirtStorageAdaptor` - is used by different storage types and doesn't use the annotation `@StorageAdaptorInfo`. In this case, a storage plugin that wants to adopt one of the predefined storage pool types will override the default behaviour. If fixing the issue in general (for new storage plugins or current ones that want to reuse the existing storage pool types) would affect all volume/snapshot/VM cases. This will lead to the need of extensive testing for each storage plugin for which we don't have the resources to do it. That's why this patch fixes the old behaviour for the SharedMountPoint by adding a new storage pool type for the StorPool plugin.

### Types of changes

- [] Breaking change (fix or feature that would cause existing functionality to change)

#### Bug Severity

- [ ] BLOCKER
- [X] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
Tested with the upgrade from CS 4.17.0.0 to 4.17.1.0
Manually tested test cases with created instances before the upgrade
executed smoke tests for the StorPool plugin

### DIFF
--- a/api/src/main/java/com/cloud/storage/Storage.java
+++ b/api/src/main/java/com/cloud/storage/Storage.java
@@ -149,7 +149,7 @@ public class Storage {
         ManagedNFS(true, false),
         Linstor(true, true),
         DatastoreCluster(true, true), // for VMware, to abstract pool of clusters
-        StorPool(true, false);
+        StorPool(true, true);
 
         private final boolean shared;
         private final boolean overprovisioning;
--- a/api/src/main/java/com/cloud/storage/Storage.java
+++ b/api/src/main/java/com/cloud/storage/Storage.java
@@ -148,7 +148,8 @@ public class Storage {
         PowerFlex(true, true), // Dell EMC PowerFlex/ScaleIO (formerly VxFlexOS)
         ManagedNFS(true, false),
         Linstor(true, true),
-        DatastoreCluster(true, true); // for VMware, to abstract pool of clusters
+        DatastoreCluster(true, true), // for VMware, to abstract pool of clusters
+        StorPool(true, false);
 
         private final boolean shared;
         private final boolean overprovisioning;
--- a/engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
+++ b/engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
@@ -16,17 +16,29 @@
 // under the License.
 package com.cloud.upgrade.dao;
 
-import com.cloud.upgrade.SystemVmTemplateRegistration;
-import com.cloud.utils.exception.CloudRuntimeException;
-import org.apache.log4j.Logger;
-
 import java.io.InputStream;
 import java.sql.Connection;
+import java.util.List;
+
+import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
+import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDaoImpl;
+import org.apache.cloudstack.storage.datastore.db.StoragePoolVO;
+import org.apache.log4j.Logger;
+
+import com.cloud.storage.Storage.StoragePoolType;
+import com.cloud.storage.VolumeVO;
+import com.cloud.storage.dao.VolumeDao;
+import com.cloud.storage.dao.VolumeDaoImpl;
+import com.cloud.upgrade.SystemVmTemplateRegistration;
+import com.cloud.utils.exception.CloudRuntimeException;
 
 public class Upgrade41700to41710 implements DbUpgrade, DbUpgradeSystemVmTemplate {
 
     final static Logger LOG = Logger.getLogger(Upgrade41610to41700.class);
     private SystemVmTemplateRegistration systemVmTemplateRegistration;
+
+    private PrimaryDataStoreDao storageDao;
+    private VolumeDao volumeDao;
 
     @Override
     public String[] getUpgradableVersionRange() {
@@ -56,6 +68,7 @@ public class Upgrade41700to41710 implements DbUpgrade, DbUpgradeSystemVmTemplate
 
     @Override
     public void performDataMigration(Connection conn) {
+        updateStorageType();
     }
 
     @Override
@@ -81,6 +94,27 @@ public class Upgrade41700to41710 implements DbUpgrade, DbUpgradeSystemVmTemplate
             systemVmTemplateRegistration.updateSystemVmTemplates(conn);
         } catch (Exception e) {
             throw new CloudRuntimeException("Failed to find / register SystemVM template(s)");
+        }
+    }
+
+    private void updateStorageType() {
+        storageDao = new PrimaryDataStoreDaoImpl();
+        List<StoragePoolVO> storPoolPools = storageDao.findPoolsByProvider("StorPool");
+        for (StoragePoolVO storagePoolVO : storPoolPools) {
+            if (StoragePoolType.SharedMountPoint == storagePoolVO.getPoolType()) {
+                storagePoolVO.setPoolType(StoragePoolType.StorPool);
+                storageDao.update(storagePoolVO.getId(), storagePoolVO);
+            }
+            updateStorPoolVolumesToStorPoolType(storagePoolVO.getId());
+        }
+    }
+
+    private void updateStorPoolVolumesToStorPoolType(long storagePoolId) {
+        volumeDao = new VolumeDaoImpl();
+        List<VolumeVO> volumes = volumeDao.findByPoolId(storagePoolId, null);
+        for (VolumeVO volumeVO : volumes) {
+            volumeVO.setPoolType(StoragePoolType.StorPool);
+            volumeDao.update(volumeVO.getId(), volumeVO);
         }
     }
 }
--- a/engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
+++ b/engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
@@ -68,7 +68,7 @@ public class Upgrade41700to41710 implements DbUpgrade, DbUpgradeSystemVmTemplate
 
     @Override
     public void performDataMigration(Connection conn) {
-        updateStorageType();
+        updateStorPoolStorageType();
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
```diff
@@ -83,4 +96,25 @@ public void updateSystemVmTemplates(Connection conn) {
             throw new CloudRuntimeException("Failed to find / register SystemVM template(s)");
         }
     }
+
+    private void updateStorageType() {
```
sureshanaparti, line 100:
```suggestion
    private void updateStorPoolStorageType() {
```

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
```diff
@@ -83,4 +96,25 @@ public void updateSystemVmTemplates(Connection conn) {
...
+        storageDao = new PrimaryDataStoreDaoImpl();
+        List<StoragePoolVO> storPoolPools = storageDao.findPoolsByProvider("StorPool");
```
sureshanaparti, line 102: Can you use the same string constant defined for StorPool provider ?

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
```diff
@@ -83,4 +96,25 @@ public void updateSystemVmTemplates(Connection conn) {
...
+        for (StoragePoolVO storagePoolVO : storPoolPools) {
+            if (StoragePoolType.SharedMountPoint == storagePoolVO.getPoolType()) {
+                storagePoolVO.setPoolType(StoragePoolType.StorPool);
+                storageDao.update(storagePoolVO.getId(), storagePoolVO);
+            }
+            updateStorPoolVolumesToStorPoolType(storagePoolVO.getId());
+        }
+    }
+
+    private void updateStorPoolVolumesToStorPoolType(long storagePoolId) {
```
sureshanaparti, line 112:
```suggestion
    private void updateStorageTypeForStorPoolVolumes(long storagePoolId) {
```

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
```diff
@@ -83,4 +96,25 @@ public void updateSystemVmTemplates(Connection conn) {
...
+
+    private void updateStorageType() {
+        storageDao = new PrimaryDataStoreDaoImpl();
+        List<StoragePoolVO> storPoolPools = storageDao.findPoolsByProvider("StorPool");
+        for (StoragePoolVO storagePoolVO : storPoolPools) {
+            if (StoragePoolType.SharedMountPoint == storagePoolVO.getPoolType()) {
+                storagePoolVO.setPoolType(StoragePoolType.StorPool);
+                storageDao.update(storagePoolVO.getId(), storagePoolVO);
+            }
+            updateStorPoolVolumesToStorPoolType(storagePoolVO.getId());
```
sureshanaparti, line 108: better move this call to if above

plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/driver/StorPoolPrimaryDataStoreDriver.java
```diff
@@ -233,7 +233,6 @@ public void createAsync(DataStore dataStore, DataObject data, AsyncCompletionCal
 
                     VolumeVO volume = volumeDao.findById(vinfo.getId());
                     volume.setPoolId(dataStore.getId());
-                    volume.setPoolType(StoragePoolType.SharedMountPoint);
```
sureshanaparti, line 236: I think, you can keep pool type as StoragePoolType.StorPool

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
slavkap: I don't think there will be cases with StorPool primary storage with a different storage type than the `SharedMountPoint`. But just in case, if there are any, I want to update all the volumes of the pool with the new storage type

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41700to41710.java
sureshanaparti: > Hi, @sureshanaparti. Thanks for the code review!
> I don't want to add a dependency for the StorPool plugin to use one string. Also, I didn't check it, but I think there will be a circular dependency problem here if I add the plugin as a dependency

fine, ensure the provider string is correct/the same as in the StorPool plugin.

api/src/main/java/com/cloud/storage/Storage.java
```diff
@@ -148,7 +148,8 @@ public static enum StoragePoolType {
         PowerFlex(true, true), // Dell EMC PowerFlex/ScaleIO (formerly VxFlexOS)
         ManagedNFS(true, false),
         Linstor(true, true),
-        DatastoreCluster(true, true); // for VMware, to abstract pool of clusters
+        DatastoreCluster(true, true), // for VMware, to abstract pool of clusters
+        StorPool(true, false);
```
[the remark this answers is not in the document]
slavkap: I forgot to change it to true. Thanks!

sureshanaparti:
@blueorangutan package

sureshanaparti (approved):
code LGTM

sureshanaparti:
@blueorangutan test matrix

nvazquez (approved):
Code LGTM

rohityadavcloud:
@blueorangutan package

shwstppr:
@slavkap does this need more changes or can it be marked ready for review?

slavkap:
@shwstppr, I need to execute a few more tests in our lab. For that reason, it's still marked as a draft if there is a need to change something

slavkap:
Hi @dhslove, @mlsorensen! If you have time, is it possible to test this PR with SharedMountPoint?

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

shwstppr:
@slavkap are you done with tests? Smoke tests look good. Should we merge?

slavkap:
@shwstppr, yes, it's ready, but isn't it better for users with `SharedMountPoint` to test it?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2868,7 +2869,8 @@ public int compare(final DiskTO arg0, final DiskTO arg1) {
                     && (pool.getType() == StoragePoolType.NetworkFilesystem
                     || pool.getType() == StoragePoolType.SharedMountPoint
                     || pool.getType() == StoragePoolType.Filesystem
-                    || pool.getType() == StoragePoolType.Gluster)) {
+                    || pool.getType() == StoragePoolType.Gluster
+                    || pool.getType() == StoragePoolType.StorPool)) {
```
GutoVeronezi, line 2873: We could create a `Set` with the types allowed and verify if it "contains", to reduce the `if` complexity.

dhslove:
> @shwstppr, yes, it's ready, but isn't it better for users with `SharedMountPoint` to test it?

@slavkap I can test during next week. I will share the result after testing.

shwstppr (approved):
LGTM. Tested with a KVM centos7 env. Was able to add a SharedMountPoint primary storage and deploy VM, volumes using it.

rohityadavcloud:
ping @dhslove any update on your testing?
I think this can be merged, looks like @shwstppr has tested this.

dhslove:
> ping @dhslove any update on your testing? I think this can be merged, looks like @shwstppr has tested this.

@slavkap @rohityadavcloud LGTM, I tested with CentOS 8 KVM environment. I checked create Shared Mount Point Storage. Also I checked working fine with already existing Shared Mount Point Storage. It works fine.

slavkap:
Thank you, everyone, for the reviews and tests!

PR #6640, opened 2022-08-15
Add usermode interface option to Libvirt Domain XML builder

mlsorensen:
### Description

This PR provides constructors and the associated changes to use LibvirtVMDef for creating user mode network interfaces.

While this isn't used directly in the CloudStack KVM agent today, it could be used in the future for e.g. pod networking/management networks without needing to assign a pod IP. The VIF driver used by the CloudStack Agent is also pluggable, so this allows plugin code to create user mode network interfaces as well.

Note that the user mode network already exists in the GuestNetType enum, but wasn't usable prior to this change.

Also included unit test to ensure we continue to create the expected XML.

Additionally, this uncovered a null pointer on `_networkRateKBps` and this PR fixes it. The decision to add bandwidth throttling assumes this field is not null and simply checks for `> 0`.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
Unit tests. Also created a test VIF driver and configured CloudStack Agent with it to test it out manually.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
@@ -1188,7 +1188,7 @@ public class LibvirtVMDef {
         private String _ipAddr;
         private String _scriptPath;
         private NicModel _model;
-        private Integer _networkRateKBps;
+        private int _networkRateKBps;
         private String _virtualPortType;
         private String _virtualPortInterfaceId;
         private int _vlanTag = -1;
@@ -1199,9 +1199,25 @@ public class LibvirtVMDef {
         private String _dpdkSourcePort;
         private String _dpdkExtraLines;
         private String _interfaceMode;
+        private String _userIp4Network;
+        private Integer _userIp4Prefix;
 
         public void defBridgeNet(String brName, String targetBrName, String macAddr, NicModel model) {
             defBridgeNet(brName, targetBrName, macAddr, model, 0);
+        }
+
+        public void defUserNet(NicModel model, String macAddr, String ip4Network, Integer ip4Prefix) {
+            _netType = GuestNetType.USER;
+            _macAddr = macAddr;
+            _userIp4Network = ip4Network;
+            _userIp4Prefix = ip4Prefix;
+            _model = model;
+        }
+
+        public void defUserNet(NicModel model, String macAddr) {
+            _netType = GuestNetType.USER;
+            _macAddr = macAddr;
+            _model = model;
         }
 
         public void defBridgeNet(String brName, String targetBrName, String macAddr, NicModel model, Integer networkRateKBps) {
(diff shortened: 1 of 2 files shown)


DaanHoogland:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -1421,13 +1438,18 @@ public String getContent() {
...
+            if (_netType != GuestNetType.VHOSTUSER && _netType != GuestNetType.USER) {
                 netBuilder.append("<link state='" + (_linkStateUp ? "up" : "down") +"'/>\n");
             }
 
             if (_slot  != null) {
                 netBuilder.append(String.format("<address type='pci' domain='0x0000' bus='0x00' slot='0x%02x' function='0x0'/>\n", _slot));
             }
+
+            if (StringUtils.isNotBlank(_userIp4Network) && _userIp4Prefix != null) {
+                netBuilder.append(String.format("<ip family='ipv4' address='%s' prefix='%s'/>\n", _userIp4Network, _userIp4Prefix));
```
DaanHoogland, line 1450: sonar advice , i think it makes sense, even though we are pretty sure we run on linux ;)
```suggestion
                netBuilder.append(String.format("<ip family='ipv4' address='%s' prefix='%s'/>%n", _userIp4Network, _userIp4Prefix));
```

DaanHoogland (approved):
clgtm

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
mlsorensen: Good catch @DaanHoogland - my only reservation is that it would be different from all of the other XML formatting strings in this `getContent()` method.

borisstoyanov:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
DaanHoogland: It is not serious @mlsorensen , just a trivial warning and i am just mentioning it without a request for change. We only run this on Linux AFAIK. If you want to bother go ahead, we could also create a ticket to change all occurrences of `\n` to `%n` in `String.format()` occurrences. But I know I can be over-pedantic in these matters so I tend to not push for these formalities.

rohityadavcloud:
@blueorangutan test

rohityadavcloud (approved):

rohityadavcloud:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
mlsorensen: I'd probably opt to leave it as-is, rather than potentially inserting one random carriage return or something different on one line of the file compared to all of the other lines. 

I don't think a theoretical libvirt server would care/require a carriage or alternate whitespace on Windows or other OS anyway.

borisstoyanov (approved):
LGTM, based on code and smoketests

Issue #6652, opened 2022-08-17
KVM - Upgrade to main fails if the storage's UUID doesn't follow the UUID format


slavkap:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Primary storage, DB
~~~

##### CLOUDSTACK VERSION

~~~
4.17.0.0 to main
~~~

##### SUMMARY
If the UUID of the primary storage doesn't meet the format of the UUID, the upgrade of CloudStack fails. This probably affects the storage plug-ins: Datera/Nexenta/Cloudbyte/StorPool and Local storage if the UUID is set in the `agent.properties` file. This is caused by change #5382 

##### STEPS TO REPRODUCE

~~~
On version 4.16.1.0 or lower
Set UUID to a storage that doesn't meet the format `%-%-%-%`
Upgrade to version 4.17.0.0 or above
~~~

##### EXPECTED RESULTS

~~~
successful upgrade
~~~

##### ACTUAL RESULTS

~~~
java.lang.StringIndexOutOfBoundsException: begin 0, end 16, length 4

or 

java.lang.NumberFormatException: For input string: "SP"
~~~

nvazquez:
@rohityadavcloud @shwstppr it looks like the original issue https://github.com/apache/cloudstack/pull/5382 can be restricted to Vmware only, a quick fix would be extending the SQL query to consider Vmware results only. I've created the PR: https://github.com/apache/cloudstack/pull/6654 on top of 4.17.0.1, would this need a 4.17.0.2 tag?

rohityadavcloud:
nope - we can only attempt to fix this in 4.17.1.0 or 4.18.0.0 - users who can't upgrade need to build their own from the PR or wait for the next official release

shwstppr:
Fixed with #6654 
@slavkap if possible, please test latest night and re-open if needed

PR #5382, opened 2021-08-27
fix mismatching between db uuids and custom attributes uuids

DK101010:
### Description
During the datastore cluster creation, cloudstack could not
recognize the existing primary storage and create a new one because
uuid format not equal.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested in my own Vmware env.

PR #6654, opened 2022-08-17
Fix SQL query for uuid wrong format

nvazquez:
### Description

Fixes: #6652 

### Types of changes

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41610to41700.java
+++ b/engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41610to41700.java
@@ -61,7 +61,7 @@ public class Upgrade41610to41700 implements DbUpgrade, DbUpgradeSystemVmTemplate
 
     @Override
     public void performDataMigration(Connection conn) {
-        fixWrongPoolUuid(conn);
+        fixWrongDatastoreClusterPoolUuid(conn);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


rohityadavcloud (approved):

rohityadavcloud:
@nvazquez the question is - could it break on vmware too?
@blueorangutan package

nvazquez:
@rohityadavcloud the original fix was for Vmware, in some cases datastore clusters seem to get the UUID in a wrong format, but needs checking for other cases in Vmware in which the same thing can happen cc.@DK101010. In my testing of the original fix I was using Vmware but didn't get pools in the wrong uuid format, and this logic was not executed

nvazquez:
Looks like in the case of Vmware + SolidFire it may be using an IQN as the UUID and this could also affect it
image

shwstppr:
@nvazquez cc @DK101010 is it possible to filter storage pools based on `pool_type` as #5382 mentions the issue is with datastore cluster on VMware?

DK101010:
Hi @shwstppr @nvazquez  
yes it is possible, pool_type = "DatastoreCluster". 

We discussed at the time in which format uuids should be stored in the database and that it should always be the same format. That's why we created the sql back then.

Therefore, the code for the storage plugin and local storage should be adapted and not the uuids in the database should be used with the wrong format or am I missing something?

nvazquez:
Thanks @DK101010 - I've updated the PR to restrict the SQL to datastore cluster only. 
@blueorangutan package

slavkap (approved):
Code LGTM

shwstppr:
@blueorangutan package

shwstppr:
> Hi @shwstppr @nvazquez yes it is possible, pool_type = "DatastoreCluster".
> 
> We discussed at the time in which format uuids should be stored in the database and that it should always be the same format. That's why we created the sql back then.
> 
> Therefore, the code for the storage plugin and local storage should be adapted and not the uuids in the database should be used with the wrong format or am I missing something?

@DK101010 I guess it is not just the code for the storage plugins and local storage. Like @slavkap pointed out in her issue it could be manually set or could be already set and used in production env where upgrades would fail. Using pool_type filter would be safer for now. Can you please review the changes?

shwstppr (approved):
Code LGTM

DaanHoogland:
would this be particular to the concept of "DataCluster" or could there be implementations of the concept that handle things differenty @nvazquez @shwstppr ? (cc @harikrishna-patnala )

Reason for my question: it seems vmware specific but I don´t see a check for that in the code.

DaanHoogland:
> @DaanHoogland yes it seems particular to the Vmware implementation but its the only supported for the type 'DatastoreCluster'

clear, i´d say it is maybe a bit early to amend this for a hypothetical future enhancement, just wondering. CLTGM

DK101010:
I am a little confused. 

I developed this because all storage (in VMware) uuids were stored with separators. But catastore cluster was stored without separator. 
It was said that all uuids are stored with separator in the database and that therefore the datastore cluster uuids are wrong.
Now it sounds like there is no structure to the storage uuids.

@shwstppr
if DatastoreClusters is only supported for VMware then code looks good. 
If not, then I would add **hypervisor= "VMware"**.

PR #6491, opened 2022-06-23
Add live migration of system VMs (KVM)

BryanMLima:
### Description

Currently, there is no way to migrate running system VMs, using the KVM hypervisor, with volumes; this means, we cannot move their volumes to a new storage pool in runtime. This PR aims to add this feature by altering the behavior of the `migrateVirtualMachineWithVolume` API to support system VMs. Furthermore, the `listVolumes` API was changed to also return system VMs root volumes, to provide easy access to the system VMs volume's UUID to the root admins. For this, a new parameter called `listsystemvms` was added to the `listVolumesCmd`. Moreover, it was added unit test for `MigrateVirtualMachineWithVolumeCmd` class.

Furthermore, the changes made by this feature required some tweaking in the UI as the system VM's volumes were now available. When these volumes were accessed, the path led to `/vm/<vm-uuid>`. However, ACS has a different path to VRs and system VMs resulting in a 404 page. This bug was fixed with the new changes; moreover, it was added a Volumes tab to system VMs and VRs as well.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### Screenshots (if appropriate):

VR volume's tab:  
image

System volume's tab:
image

### How Has This Been Tested?
This was tested on a local lab through CloudMonkey, using the command `migrate virtualmachinewithvolume` with the UUID of a system VM and migrating its volume to another storage pool. The command ran as expected. Furthermore, it was added unit test to verify the API behavior.

For the UI, I tried to access the volumes of system VMs and VRs as well as the volumes of user VMs to check if nothing was broken. I did not find any errors.

PR #6685, opened 2022-08-29
Return volume UUID in the `listRouters` and `listSystemVms` APIs

JoaoJandre:
### Description

This PR extends the `listRouters` and `listSystemVms` APIs to make them return the VM's root volume ID. So that in the future the UI may be extended to use the `findStoragePoolsForMigration` to migrate VRs and System VMs.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
This was tested in a local lab, by calling the APIs through Cloudmonkey and checking if the VM's root volume ID was being correctly returned.
Some unit tests were also added.

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/response/DomainRouterResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/DomainRouterResponse.java
@@ -237,6 +237,10 @@ public class DomainRouterResponse extends BaseResponseWithAnnotations implements
     @Param(description = "the version of the code / software in the router")
     private String softwareVersion;
 
+    @SerializedName("rootvolumeid")
+    @Param(description = "the router's ROOT volume ID")
+    private String rootVolumeID;
+
     public DomainRouterResponse() {
         nics = new LinkedHashSet<NicResponse>();
     }
@@ -501,5 +505,13 @@ public class DomainRouterResponse extends BaseResponseWithAnnotations implements
 
     public void setSoftwareVersion(String softwareVersion) {
         this.softwareVersion = softwareVersion;
+    }
+
+    public String getRootVolumeID() {
+        return rootVolumeID;
+    }
+
+    public void setRootVolumeID(String rootVolumeID) {
+        this.rootVolumeID = rootVolumeID;
     }
 }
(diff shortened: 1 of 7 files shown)


DaanHoogland:
@JoaoJandre in itself the code looks good but I have a problem with the design; a join dao is supposed to get data from a view and not to manippulate or augment it. This is more in place in the service serving the API command. What made you decide to implement it like this? (in the `..JoinDaoImpl` classes)

JoaoJandre:
> @JoaoJandre in itself the code looks good but I have a problem with the design; a join dao is supposed to get data from a view and not to manippulate or augment it. This is more in place in the service serving the API command. What made you decide to implement it like this? (in the `..JoinDaoImpl` classes)

@DaanHoogland the response is already being built in the `DomainRouterJoinDaoImpl`, and there are already other similar procedures being made in the same method, such as retrieving the `podname` (lines 137 through 139).

I understand that ideally this should be done elsewhere, but in that case this whole response builder method would have to be refactored, which would be out of scope of this PR.

GutoVeronezi (approved):
CLGTM

GutoVeronezi:
@JoaoJandre, can you take a look at the Travis errors?

@blueorangutan package

shwstppr:
@JoaoJandre why not instead allow listing volumes for routers and system VMs for admin?

JoaoJandre:
> @JoaoJandre why not instead allow listing volumes for routers and system VMs for admin?

I'm sorry @shwstppr , I didn't understand your suggestion. Could you please elaborate?

shwstppr:
> > @JoaoJandre why not instead allow listing volumes for routers and system VMs for admin?
> 
> I'm sorry @shwstppr , I didn't understand your suggestion. Could you please elaborate?

@JoaoJandre I meant why not allow `listVolumes` API to list router and systemvm volumes with a new param `systemvmid` or something. Because with the current approach today we need volume's uuid, then we may need to add volume name or something to show in UI, then volume's current storage pool to assist migration in UI

JoaoJandre:
I was looking through the open PRs and found PR #6491, which already implements what you suggested. With PR #6491 a ROOT admin will be able to list system VM volumes by passing the parameter `listall` as `true`. Therefore, there is no need for my PR anymore. 

@shwstppr, when thinking about your suggestion another idea came to me; as system VM volumes will be listed via API `listVolumes`, why not handle system VMs in `listVirtualMachines`? This way, we could make `listVirtualMachines` a complete API, handling all types of VMs, and `listRouters` and `listSystemVms` would not be necessary anymore.

PR #6733, opened 2022-09-12
Refactor TestHttp.testHttpclient to avoid the Exception Suppression

### LINKED ISSUE #6673: Refactor `TestHttp.testHttpclient` to avoid the Exception Suppression
##### ISSUE TYPE
* Improvement Request
##### COMPONENT NAME
unit-test
##### OS / ENVIRONMENT
N/A
##### SUMMARY
When I am running the test for [testHttpclient](https://github.com/apache/cloudstack/blob/ad0ae8397460c243e34d4017639d19a379c9e995/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/TestHttp.java#L44) I noticed that the test case is using the Try-Catch block with `e.printStackTrace()` to suppress some of the Exceptions in the function.

```java
public void testHttpclient(String templateUrl) {
        try {
            ...
            final HttpGet getMethod = new HttpGet(templateUrl);
            response = client.execute(getMethod);
            final HttpEntity entity = response.getEntity();

            output = new BufferedOutputStream(new FileOutputStream(localFile));
            entity.writeTo(output);
        } catch (final ClientProtocolException e) {
            // TODO Auto-generated catch block
            e.printStackTrace();
        } catch (final IOException e) {
            // TODO Auto-generated catch block
            e.printStackTrace();
        } finally {
            try {
                if (output != null) {
                    output.close();
                }
            } catch (final IOException e) {
                // TODO Auto-generated catch block
                e.printStackTrace();
            }
...
}
```

These three catch blocks will only print the the Exception information and will not raise any failure in the testing process. This behavior will lead the developer harder to notice the issue in the CI\CD reports.

I suggest **removing** the try-catch structure and adding some additional assertion(if possible). This will improve the readability of the test cases, and avoid missing the test failure. Even if the exception does not raise by the component that we are testing here, we can still know that we need to fix the outside conditions first to make sure the test case running under stable and consistent conditions.

After refactoring
```java
public void testHttpclient(String templateUrl) throws ClientProtocolException IOException{
        ...
        try {
            HttpResponse response = client.execute(method);
            length = Long.parseLong(response.getFirstHeader("Content-Length").getValue());
            System.out.println(response.getFirstHeader("Content-Length").getValue());
            final File localFile = new File("/tmp/test");
            if (!localFile.exists()) {
                localFile.createNewFile();
            }

            final HttpGet getMethod = new HttpGet(templateUrl);
            response = client.execute(getMethod);
            final HttpEntity entity = response.getEntity();

            output = new BufferedOutputStream(new FileOutputStream(localFile));
            entity.writeTo(output);
        } finally {
                if (output != null) {
                    output.close();
                }
       ...
}
```

###### Operational impact

The testing refactoring won’t affect any part of the production code behavior. Also, only the original cases are replaced with unit tests.

Codegass:
### Description

FIX #6673

This PR removes the catch block with the `e.printStackTrace()`, which suppressed the Exceptions to be thrown in the CI\CD's final report.

These catch blocks will only print the the Exception information and will not raise any failure in the testing process. This behavior will lead the developer harder to notice the issue in the CI\CD reports.

I suggest removing the try-catch structure and adding some additional assertion(if possible). This will improve the readability of the test cases, and avoid missing the test failure. Even if the exception does not raise by the component that we are testing here, we can still know that we need to fix the outside conditions first to make sure the test case running under stable and consistent conditions.
### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

Local cluster environment (KVM, UBUNTU 18.04) runs the test suite.

### DIFF
--- a/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/TestHttp.java
+++ b/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/TestHttp.java
@@ -41,7 +41,7 @@ import junit.framework.Assert;
 public class TestHttp extends AbstractTestNGSpringContextTests {
     @Test
     @Parameters("template-url")
-    public void testHttpclient(String templateUrl) {
+    public void testHttpclient(String templateUrl) throws IOException {
         final HttpHead method = new HttpHead(templateUrl);
         final DefaultHttpClient client = new DefaultHttpClient();
 
@@ -62,20 +62,9 @@ public class TestHttp extends AbstractTestNGSpringContextTests {
 
             output = new BufferedOutputStream(new FileOutputStream(localFile));
             entity.writeTo(output);
-        } catch (final ClientProtocolException e) {
-            // TODO Auto-generated catch block
-            e.printStackTrace();
-        } catch (final IOException e) {
-            // TODO Auto-generated catch block
-            e.printStackTrace();
         } finally {
-            try {
-                if (output != null) {
-                    output.close();
-                }
-            } catch (final IOException e) {
-                // TODO Auto-generated catch block
-                e.printStackTrace();
+            if (output != null) {
+                output.close();
             }
         }
 
--- a/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/TestHttp.java
+++ b/engine/storage/integration-test/src/test/java/org/apache/cloudstack/storage/test/TestHttp.java
@@ -26,7 +26,6 @@ import java.io.OutputStream;
 
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpResponse;
-import org.apache.http.client.ClientProtocolException;
 import org.apache.http.client.methods.HttpGet;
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@Codegass is this test actually run? as it is in a package called `org.apache.cloudstack.storage.test;` and in a project called `engine/storage/integration-test` I doubt that it is and a cleanup would require more action.

Codegass:
> @Codegass is this test actually run? as it is in a package called `org.apache.cloudstack.storage.test;` and in a project called `engine/storage/integration-test` I doubt that it is and a cleanup would require more action.

@DaanHoogland Hi Daan, Thank you for the notice. You are correct, I just run the overall test suite and missed the single test of itself. Please allow me to dig a little bit.

rohityadavcloud:
@blueorangutan package

rohityadavcloud (approved):

rohityadavcloud:
On second look, the changes are in (unit) test code and Travis has passed.

PR #6735, opened 2022-09-13
Fix of template size when setup CS with StorPool as a primary storage

slavkap:
### Description

In the newer versions of CS, the template size is null until the Secondary storage VM isn't created. This leads to NullPointerException, and CS cannot create a volume for the VM. We'll now save the real size of the template, and we'll get it from the template_spool_ref table

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
Fresh install
2 Zones with 5 agents on KVM

### DIFF
--- a/plugins/storage/volume/storpool/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/StorPoolDownloadTemplateCommandWrapper.java
+++ b/plugins/storage/volume/storpool/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/StorPoolDownloadTemplateCommandWrapper.java
@@ -25,6 +25,7 @@ import java.io.File;
 import java.util.List;
 
 import org.apache.cloudstack.storage.command.CopyCmdAnswer;
+import org.apache.cloudstack.storage.to.TemplateObjectTO;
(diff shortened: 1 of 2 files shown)


rohityadavcloud:
@blueorangutan package

plugins/storage/volume/storpool/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/StorPoolDownloadTemplateCommandWrapper.java
```diff
@@ -53,7 +54,7 @@ public CopyCmdAnswer execute(final StorPoolDownloadTemplateCommand cmd, final Li
         String dstPath = null;
         KVMStoragePool secondaryPool = null;
         DataTO src = cmd.getSourceTO();
-        DataTO dst = cmd.getDestinationTO();
+        TemplateObjectTO dst = (TemplateObjectTO) cmd.getDestinationTO();
```
nvazquez, line 57: Maybe ask if the object is instance of TemplateObjectTO, otherwise fail and log it?

PR #6742, opened 2022-09-15
Ignore calls to PowerFlex for host revocation when host is null

### LINKED ISSUE #6739: VM deletion can fail on volume access revocation if VM's last host is removed
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
API, Storage
~~~

##### CLOUDSTACK VERSION
~~~
main
~~~

##### CONFIGURATION
PowerFlex, Datera storage

##### SUMMARY
If a VM is stopped, then the last host it was running on is removed from cloudstack, the VM can fail to delete. This is because deletion attempts to revoke host access to the VM's volumes and various storage plugins will throw errors if the provided host is null.

Looking through the storage plugin implementations, it looks like this would affect Datera and PowerFlex storage types only.

Generally this seems to be due to passing a null host object to the storage provider's revokeAccess() method. Looked into a holistic fix but there are several places where revokeAccess() can be invoked or could be invoked in the future. Additionally, a storage plugin could choose to handle a null host as a "revoke all" command.

##### STEPS TO REPRODUCE
1. Create VM with root on Datera or PowerFlex storage 
2. Start/stop the VM
3. remove the host the VM ran on from CloudStack
4. attempt to delete the VM

##### EXPECTED RESULTS
VM should be deleted

##### ACTUAL RESULTS
Error is thrown during deletion

mlsorensen:
Signed-off-by: Marcus Sorensen <mls@apple.com>

### Description

This PR Fixes #6739 (for PowerFlex/ScaleIO only, Datera still needs to be addressed), which can occur if the last host the VM ran on is deleted from CloudStack. At the point the VM is deleted, cloudstack attempts to make a final call to revoke access to volumes, passing the last host the VM ran on. If this host is gone, we get an error and are unable to delete the VM.

It's possible that there may be a more holistic fix to this by identifying all of the places where `revokeAccess()` is called and checking for null host. It's possible other storage plugins don't even need host information to revoke access to volumes and may need this call to revoke. Therefore I'm only applying this fix to the ScaleIOPrimaryDataStoreDriver to skip revoking access when there is no host to revoke access for, and this should protect us as well when a new part of the code tries to use `revokeAccess()` in the future.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?
Tested in a local branch against PowerFlex primary storage, and with internal testing environments. Repeated steps outlined in #6739. We also had quite a few VMs that would not delete before this change was put into place, which cleaned up successfully afterward.

### DIFF
--- a/plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/driver/ScaleIOPrimaryDataStoreDriver.java
+++ b/plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/driver/ScaleIOPrimaryDataStoreDriver.java
@@ -189,6 +189,11 @@ public class ScaleIOPrimaryDataStoreDriver implements PrimaryDataStoreDriver {
 
     @Override
     public void revokeAccess(DataObject dataObject, Host host, DataStore dataStore) {
+        if (host == null) {
+            LOGGER.warn("Declining to revoke access to PowerFlex volume when a host is not provided");
+            return;
+        }
+
         try {
             if (DataObjectType.VOLUME.equals(dataObject.getType())) {
                 final VolumeVO volume = volumeDao.findById(dataObject.getId());
--- a/plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/driver/ScaleIOPrimaryDataStoreDriver.java
+++ b/plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/driver/ScaleIOPrimaryDataStoreDriver.java
@@ -190,7 +190,7 @@ public class ScaleIOPrimaryDataStoreDriver implements PrimaryDataStoreDriver {
     @Override
     public void revokeAccess(DataObject dataObject, Host host, DataStore dataStore) {
         if (host == null) {
-            LOGGER.warn("Declining to revoke access to PowerFlex volume when a host is not provided");
+            LOGGER.info("Declining to revoke access to PowerFlex volume when a host is not provided");
             return;
         }
 


rohityadavcloud (approved):

rohityadavcloud:
@blueorangutan package

plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/driver/ScaleIOPrimaryDataStoreDriver.java
```diff
@@ -189,6 +189,11 @@ public boolean grantAccess(DataObject dataObject, Host host, DataStore dataStore
 
     @Override
     public void revokeAccess(DataObject dataObject, Host host, DataStore dataStore) {
+        if (host == null) {
+            LOGGER.warn("Declining to revoke access to PowerFlex volume when a host is not provided");
```
DaanHoogland, line 193: could be `info`, right? or is this indicative of a possible (user) error?

rohityadavcloud:
@blueorangutan test

plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/driver/ScaleIOPrimaryDataStoreDriver.java
mlsorensen: You're right @DaanHoogland - it could probably be info. My thinking was that yes, it could potentially mean that the caller did something wrong, but not necessarily.

rohityadavcloud:
@blueorangutan package

shwstppr (approved):

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@blueorangutan test

PR #6750, opened 2022-09-19
Allow download of system vm templates

GaOrtiga:
### Description

Currently, ACS does not allow the user to download System VM Templates,  even though it may be usefull as it can speed up the registration process of the template for production once the homologation is done beforehand. This PR changes this, allowing the user to download said VM Templates

### Types of changes

- [ ] Major
- [X] Minor

### How Has This Been Tested?

I built the code with the respective changes in a local lab. After that, I copied the Uuid from one of the System VM templates. Then, using CloudMonkey, I utilized the command 'extract template' with the copied Uuid. CloudMonkey then generated a link for the download of the template, confirming that the alteration was successful.

### DIFF
--- a/server/src/main/java/com/cloud/template/TemplateManagerImpl.java
+++ b/server/src/main/java/com/cloud/template/TemplateManagerImpl.java
@@ -520,9 +520,8 @@ public class TemplateManagerImpl extends ManagerBase implements TemplateManager,
             throw new InvalidParameterValueException("Unable to find " + desc + " with id " + templateId);
         }
 
-        if (template.getTemplateType() == Storage.TemplateType.SYSTEM) {
-            throw new InvalidParameterValueException("Unable to extract the " + desc + " " + template.getName() + " as it is a default System template");
-        } else if (template.getTemplateType() == Storage.TemplateType.PERHOST) {
+
+        if (template.getTemplateType() == Storage.TemplateType.PERHOST) {
             throw new InvalidParameterValueException("Unable to extract the " + desc + " " + template.getName() + " as it resides on host and not on SSVM");
         }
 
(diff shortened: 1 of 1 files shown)


rohityadavcloud:
Thanks for the PR - can you explain the use case as the systemvmtemplate aren't for public/general consumption for users? The systemvmtemplates are also publicly available so if anybody wants they can register using such links.

GutoVeronezi (approved):

GaOrtiga:
@rohityadavcloud 
Access to the official repositories may be slow at time; therefore, being able to perform a direct download can speed up this process in some use cases. Also since access to these templates is already public, there should be no problem in allowing the downloads to be performed from within the platform itself.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland (approved):
code looks good and I'm fine with it functionaly

DaanHoogland:
> Thanks for the PR - can you explain the use case as the systemvmtemplate aren't for public/general consumption for users? The systemvmtemplates are also publicly available so if anybody wants they can register using such links.

is this a :-1: , @rohityadavcloud ?

rohityadavcloud:
@DaanHoogland no just a question, really trying to understand why we're making this public? (historically this has been private/hidden to users, or at least the download button isn't it?)

DaanHoogland:
@rohityadavcloud it is a valid question but I think @GaOrtiga 's argument makes sense. Do we merge?

SadiJr (approved):
CLGTM, and I agree with @DaanHoogland and @GaOrtiga arguments

DaanHoogland:
@rohityadavcloud I need to test but i think that the system vm template will not be visible to the normal user so no download button either. Only root admins should be able to download it this way. (not sure about the API)
We can verify. Is there any other blocking concern you have?

GaOrtiga:
@DaanHoogland you are correct, the download button will not be visible in the UI as shown below:
image (1)

Furthermore, even through CloudMonkey download will only be available to Root Admins, as shown below:
Running the command with a Root Admin account:
downloadRootAdmin


Running the command with a user-level account:
downloaduser


With this being said, I still do not understand why allowing operators to download said templates is a concern. These images are publicly available, and operators are the ones registering them in ACS. Therefore, it makes sense to allow them to download directly from ACS if they wish so.

rohityadavcloud:
Makes sense, LGTM, let's merge this.

PR #6839, opened 2022-10-20
Fixed NPE on volume creation from snapshot

JoaoJandre:
### Description

This PR fixes a NPE that occurs when trying to create a volume from a snapshot. This happens because when trying to log the `virtualMachineProfile` in the `AbstractStoragePoolAllocator#logStartOfSearch` method, its `toString()` method will throw a NPE since the `_vm` attribute is null.

Some `toString()`s were also created in classes that had objects being logged during the volume creation process, as they were being logged as memory addresses since they had no `toString()` implementation.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
This was tested in a local lab, before this patch every time I tried to create a volume from a snapshot, a NPE was thrown. After the patch the volume creation from snapshot was fixed.

### DIFF
--- a/engine/components-api/src/main/java/com/cloud/vm/VirtualMachineProfileImpl.java
+++ b/engine/components-api/src/main/java/com/cloud/vm/VirtualMachineProfileImpl.java
@@ -86,7 +86,10 @@ public class VirtualMachineProfileImpl implements VirtualMachineProfile {
 
     @Override
     public String toString() {
-        return _vm.toString();
+        if (_vm != null) {
+            return _vm.toString();
+        }
+        return null;
     }
 
     @Override
--- a/engine/components-api/src/main/java/com/cloud/vm/VirtualMachineProfileImpl.java
+++ b/engine/components-api/src/main/java/com/cloud/vm/VirtualMachineProfileImpl.java
@@ -89,7 +89,7 @@ public class VirtualMachineProfileImpl implements VirtualMachineProfile {
         if (_vm != null) {
             return _vm.toString();
         }
-        return null;
+        return "";
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


engine/storage/src/main/java/org/apache/cloudstack/storage/allocator/ClusterScopeStoragePoolAllocator.java
```diff
@@ -107,7 +107,7 @@ protected List<StoragePool> select(DiskProfile dskCh, VirtualMachineProfile vmPr
             }
         }
 
-        s_logger.debug(String.format("ClusterScopeStoragePoolAllocator is returning [%s] suitable storage pools [%s].", suitablePools.size(), suitablePools));
+        logEndOfSearch(suitablePools);
```
weizhouapache, line 110: code lgtm

however, I like the old style more than the new method `logEndOfSearch`

engine/components-api/src/main/java/com/cloud/vm/VirtualMachineProfileImpl.java
```diff
@@ -86,7 +86,10 @@ public VirtualMachineProfileImpl(VirtualMachine.Type type) {
 
     @Override
     public String toString() {
-        return _vm.toString();
+        if (_vm != null) {
+            return _vm.toString();
+        }
+        return null;
```
DaanHoogland, line 92:
```suggestion
        return "";
```

engine/storage/src/main/java/org/apache/cloudstack/storage/allocator/ClusterScopeStoragePoolAllocator.java
weizhouapache: I know it. I just think it is unnecessary. it will not bring any benefit I think.
on the contrary, if I get log "ClusterScopeStoragePoolAllocator is returning xxx" and grep in source code, the result will be empty.
ps: this does not mean -1 to this pr.

DaanHoogland (approved):

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

PR #5375, opened 2021-08-26
vr: ipsec/l2tp vpn secret with no ID selectors

weizhouapache:
### Description

according to https://wiki.strongswan.org/projects/strongswan/wiki/Ipsecsecrets , the ID selector is optional.

Not sure if this PR fixes #4281

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

Issue #4281, opened 2020-08-24
Problem with L2TP remote VPN


humito36:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME
L2TP VPN

##### CLOUDSTACK VERSION

4.14

##### CONFIGURATION
VPC network system vm version: 4.14.0

##### OS / ENVIRONMENT
Mac OS, Windows 10.

##### SUMMARY
I have got problem with remote VPN feature in VPC. Probably script generate bad config file at VPC VM:
/etc/ipsec.d/ipsec.any.secrets 

PUBLIC IP %any : PSK "PRESHAREDKEY"
When i edit it to:
 : PSK "PRESHAREDKEY and do ipsec rereadsecrets via shell at vac it starts to work.

rhtyd:
@humito36 are you able to reproduce it if you restart VPC with cleanup?

humito36:
Yes The problem persists i tried few times

Pozdrawiam,

Tomasz Humiczewski
Członek Zarządu

tomasz.humiczewski@blackice.com.pl
kom: +48 533 599 910
tel. +48 12 634 99 90

BlackIce Networks Sp. z o. o., ul. Marcika 2, 30-443, Kraków
NIP: 676 245 98 98, REGON: 122709468 KRS: 0000439317 Sąd Rejonowy dla Krakowa – Śródmieścia w Krakowie Wydział XI Gospodarczy
This e-mail message may contain confidential or legally privileged information and is intended only for the use of the intended recipient(s).
Any unauthorized disclosure, dissemination, distribution, copying or the taking of any action in reliance on the information herein is prohibited.
E-mails are not secure and cannot be guaranteed to be error free as they can be intercepted, amended, or contain viruses.
Anyone who communicates with us by e-mail is deemed to have accepted these risks.
BlackIce Networks is not responsible for errors or omissions in this message and denies any responsibility for any damage arising from the use of e-mail.
Any opinion and other statement contained in this message and any attachment are solely those of the author and do not necessarily represent those

________________________________
Od: Rohit Yadav <notifications@github.com>
Wysłane: Friday, August 28, 2020 11:34:07 AM
Do: apache/cloudstack <cloudstack@noreply.github.com>
DW: Tomasz Humiczewski <tomasz.humiczewski@blackice.com.pl>; Mention <mention@noreply.github.com>
Temat: Re: [apache/cloudstack] Problem with L2TP remote VPN (#4281)


[...]

PR #6907, opened 2022-11-17
Fix C2S VPN in parallel to S2S VPN

GutoVeronezi:
### Description

PR #5375, introduced in version 4.15.2.0, removed parameter `%any` of VPNs client-to-site (C2S) IPSec secrets:
- structure before PR #5375:
  `<IP> %any : PSK "<PSK>"`
- structure after PR #5375:
  `<IP> : PSK "<PSK>"`

Because of that, when a VPN site-so-site (S2S) is created in parallel to a VPN C2S in the same network, the C2S will not handle any IP (`%any`) anymore and, as the network is being tunneled to the other VPN, the connection will be handled by the final peer. This way, when a VPN S2S is created in parallel to a VPN C2S in the same network, it is only possible to connect to the C2S with the S2S PSK.

As ACS is only able to implement a single C2S per network (ACS allows setting more than one IP of the network as VPN, however, only the first will be implemented) and every S2S has its own secret file, the secrets structure of C2S was changed to contain only the PSK:
```
: PSK "<PSK>"
```
 By doing that, StrongSwan will handle correctly C2S connections from any IP and still will use the correct PSK for S2S.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

I created 3 networks (A, B, and C) and a VM in each network. In networks B and C I enabled the VPN. From VM in network A, I configured both VPNs and successfully connected to them with their respective PSK. Then, from B to C I established a S2S VPN. Without this change, when I was trying to connect again to C2S B with B PSK, the connection was refused; however, when tried to connect to C2S B with C PSK, the connection was successful. After the changes, I tried to connect again to C2S B with B PSK and the connection was successful.

### DIFF
--- a/systemvm/debian/opt/cloud/bin/configure.py
+++ b/systemvm/debian/opt/cloud/bin/configure.py
@@ -999,7 +999,7 @@ class CsRemoteAccessVpn(CsDataBag):
 
         secret = CsFile(vpnsecretfilte)
         secret.empty()
-        secret.addeq("%s : PSK \"%s\"" % (left, psk))
+        secret.addeq(": PSK \"%s\"" % (psk))
         secret.commit()
 
         xl2tpdconf = CsFile(xl2tpdconffile)


systemvm/debian/opt/cloud/bin/configure.py
```diff
@@ -1002,7 +1002,7 @@ def configure_l2tpIpsec(self, left, obj):
 
         secret = CsFile(vpnsecretfilte)
         secret.empty()
-        secret.addeq("%s : PSK \"%s\"" % (left, psk))
+        secret.addeq(": PSK \"%s\"" % (psk))
```
rohityadavcloud, line 1005: @GutoVeronezi LGTM, but should we revert to old ? I don't know what the syntax really means 
```        secret.addeq("%s %%any : PSK \"%s\"" % (left, psk))
```

rohityadavcloud (approved):
LGTM based on test results shared; this would need syntactic review from a strongswan/ipsec/vpn expert

systemvm/debian/opt/cloud/bin/configure.py
rohityadavcloud: Update, from https://github.com/apache/cloudstack/issues/4281#issue-684586236 it seems the user's fix correlates with the syntax used.

rohityadavcloud:
Packaging result: :heavy_check_mark: el7 :heavy_check_mark: el8 :heavy_check_mark: debian :heavy_check_mark: suse15. SL-JID 4624

rohityadavcloud:
(manually fired smoketest)

weizhouapache:
@GutoVeronezi 
as you said, PR #5375 changed the structure from 
`<IP> %any : PSK "<PSK>" `
to
`<IP> : PSK "<PSK>" `
With this PR, the structure is changed to
` : PSK "<PSK>" `


I have few questions
- Have you ever tested the result by reverting #5375 ?
- Since public IP is removed from the configuration in this PR, might it be potential security issue ?

GutoVeronezi:
@weizhouapache, yes, I have tested reverting #5375, and it works as well. The problem was that, by removing `%any` and enabling S2S, the request would be redirected to the final peer because the source (`right`) was not being handled by the VPN C2S. Removing the destination (`left`) makes the C2S handle the connection with the PSK, independent of the source.

Since only a single VPN C2S is configured for each network/VPC, I do not see how it could be a security issue.

---

@rohityadavcloud, and @weizhouapache, since the user's problem (https://github.com/apache/cloudstack/issues/4281#issue-684586236) was observed in `4.14.0`, with another version of StrongSwan, and I could not reproduce it, the change was made in order to honor their comment. However, if we can confirm that it was only a problem with the StrongSwan version and #5375 change was not necessary, I think we could revert #5375.

weizhouapache:
> @weizhouapache, yes, I have tested reverting #5375, and it works as well. The problem was that, by removing `%any` and enabling S2S, the request would be redirected to the final peer because the source (`right`) was not being handled by the VPN C2S. Removing the destination (`left`) makes the C2S handle the connection with the PSK, independent of the source.
> 
> Since only a single VPN C2S is configured for each network/VPC, I do not see how it could be a security issue.
> 
> @rohityadavcloud, and @weizhouapache, since the user's problem ([#4281 (comment)](https://github.com/apache/cloudstack/issues/4281#issue-684586236)) was observed in `4.14.0`, with another version of StrongSwan, and I could not reproduce it, the change was made in order to honor their comment. However, if we can confirm that it was only a problem with the StrongSwan version and #5375 change was not necessary, I think we could revert #5375.

@GutoVeronezi 
Thanks for your explanation.
I am ok with this pr or reverting #5375. We need to make sure both #6907 and #4281 are fixed.

rohityadavcloud:
@blueorangutan test

DaanHoogland (approved):
not tested but looks good. btween reverting and applying this, this seems to be the simpler configuration which has my preference.

rohityadavcloud:
@GutoVeronezi can you cherry-pick or rebase your PR to 4.17 branch and edit PR base branch to 4.17?

rohityadavcloud:
Ping @weizhouapache @harikrishna-patnala can you help review this?

weizhouapache (approved):
code lgtm

not tested it yet

harikrishna-patnala (approved):
Code LGTM, 

@GutoVeronezi can you update the structure in the description to reflect the latest change. Thanks

GutoVeronezi:
Sorry, @harikrishna-patnala, I did not catch what you meant. There were no changes in the code, the branch was only rebased to 4.17.

harikrishna-patnala:
@GutoVeronezi I'm just asking about the structure mentioned in the description.
structure after the change:
<IP> : PSK "<PSK>"

I think now it is only
 : PSK "<PSK>"

right!

rohityadavcloud:
Thanks @GutoVeronezi I’ll kick tests
@blueorangutan package

rohityadavcloud:
@blueorangutan test

harikrishna-patnala:
> @harikrishna-patnala, that part of the description is an introduction to the problem, explaining why it was happening. The changes I am doing in this PR is in the third paragraph. I made a little change to make it more explicit, thanks.

Thanks @GutoVeronezi for the change

PR #6922, opened 2022-11-24
Display all volumes to root admin

CharlesQueiroz:
### Description

Currently only volume of User vm's are displayed in api.
Display all volumes to root admins so that its easy to see
rather than doing some sql queries

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial


harikrishna-patnala:
@blueorangutan package

BryanMLima:
@CharlesQueiroz  this feature is already available through the parameter `listSystemVms` of the the `listVolumes` API. With the parameter set to `true` it will list the system VMs' volumes of the VMs created by the caller account. If the account is a Root Admin, then, it can use the parameter `listAll` to list all volumes, including those created by the `system` account.

PR #6942, opened 2022-12-02
ssvm: synchorise when adding iptables and routing rules

rohityadavcloud:
Use synchronised in NfsSecondaryStorageResource when adding iptables and routing rules.

### Types of changes

### DIFF
--- a/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
+++ b/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
@@ -2242,7 +2242,7 @@ public class NfsSecondaryStorageResource extends ServerResourceBase implements S
 
         addRouteToInternalIpOrCidr(_localgw, _eth1ip, _eth1mask, destCidr);
 
-        return msg;
+        return null;
     }
 
     private Answer execute(SecStorageFirewallCfgCommand cmd) {
--- a/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
+++ b/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
@@ -2238,13 +2238,6 @@ public class NfsSecondaryStorageResource extends ServerResourceBase implements S
         if (result != null) {
             s_logger.warn("Error in allowing outgoing to " + destCidr + ", err=" + result);
             return "Error in allowing outgoing to " + destCidr + ", err=" + result;
-            if (result != null) {
-                msg = "Error in allowing outgoing to " + destCidr + ", err=" + result;
-                s_logger.warn(msg);
-            } else {
-                msg = null;
-                break;
-            }
         }
 
         addRouteToInternalIpOrCidr(_localgw, _eth1ip, _eth1mask, destCidr);
--- a/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
+++ b/services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
@@ -2225,7 +2225,7 @@ public class NfsSecondaryStorageResource extends ServerResourceBase implements S
         return null;
     }
 
-    public String allowOutgoingOnPrivate(String destCidr) {
+    public synchronized String allowOutgoingOnPrivate(String destCidr) {
         if (!_inSystemVM) {
             return null;
         }
@@ -2234,15 +2234,21 @@ public class NfsSecondaryStorageResource extends ServerResourceBase implements S
         command.add("-c");
         command.add("iptables -I OUTPUT -o " + intf + " -d " + destCidr + " -p tcp -m state --state NEW -m tcp  -j ACCEPT");
 
-        String result = command.execute();
-        if (result != null) {
-            s_logger.warn("Error in allowing outgoing to " + destCidr + ", err=" + result);
-            return "Error in allowing outgoing to " + destCidr + ", err=" + result;
+        String msg = null;
+        for (int retry = 3; retry > 0; retry--) {
+            String result = command.execute();
+            if (result != null) {
+                msg = "Error in allowing outgoing to " + destCidr + ", err=" + result;
+                s_logger.warn(msg);
+            } else {
+                msg = null;
+                break;
+            }
         }
 
         addRouteToInternalIpOrCidr(_localgw, _eth1ip, _eth1mask, destCidr);
 
-        return null;
+        return msg;
     }
 
     private Answer execute(SecStorageFirewallCfgCommand cmd) {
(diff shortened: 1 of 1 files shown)


rohityadavcloud:
This is hard to test/reproduce. I think this code is called by multiple threads which can cause xlock/iptables locking issue where the command fails and when it does ip routing rules aren't added. By putting synchronized on the method and logging/trying three times seem to be working in a rare-case env.

rohityadavcloud:
cc @andrijapanicsb

andrijapanicsb:
thx @rohityadavcloud 

For the record: this has been impossible to reproduce in my test env (based on the instructions given by the relevant third party) - the problem is when doing multiple volume snapshots in multiple clusters, at the exact same time - ACS tries to add iptables rule inside SSVM to allow outgoing traffic (to vCenter I think) - and since there are multiple concurrent attempts to add iptable rule - some of those would fail (since iptable has an internal locking mechanism where it creates a lock while an rule is being added, so another rule/thread can not run) - so some of the snapshot jobs would fail as a consequence of not being able to add iptable rule.

services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
```diff
@@ -2234,15 +2234,21 @@ public String allowOutgoingOnPrivate(String destCidr) {
         command.add("-c");
         command.add("iptables -I OUTPUT -o " + intf + " -d " + destCidr + " -p tcp -m state --state NEW -m tcp  -j ACCEPT");
 
-        String result = command.execute();
-        if (result != null) {
-            s_logger.warn("Error in allowing outgoing to " + destCidr + ", err=" + result);
-            return "Error in allowing outgoing to " + destCidr + ", err=" + result;
+        String msg = null;
+        for (int retry = 3; retry > 0; retry--) {
+            String result = command.execute();
```
weizhouapache, line 2239: @rohityadavcloud 
I wonder if the exit code should be checked instead of the output.

rohityadavcloud:
@blueorangutan ui

services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
rohityadavcloud: I think on error the result is non-empty which is what the code is earlier, I added just this retry-logic around it

rohityadavcloud:
@blueorangutan test

services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
weizhouapache: @rohityadavcloud 
understood, your code is good.

I just wondered if it works as you expected. Have you faced any issue with it ? if you know how to reproduce the issue, we could verify if it is fixed by your PR.

I did a small test with the iptables command . the output is always empty, exit code is 0 or 2.

1. iptables command fails
```
[root@mgmt1 ~]# output=$(bash -c 'iptables -I OUTPUT -o eth1 -d 300.300.300.300/24 -p tcp -m state --state NEW -m tcp  -j ACCEPT')
iptables v1.4.21: host/network `300.300.300.300' not found
Try `iptables -h' or 'iptables --help' for more information.
[root@mgmt1 ~]# echo $?
2
[root@mgmt1 ~]# echo $output

[root@mgmt1 ~]#
```
2. iptables command succeeds
```
[root@mgmt1 ~]# output=$(bash -c 'iptables -I OUTPUT -o eth1 -d 100.100.100.100/24 -p tcp -m state --state NEW -m tcp  -j ACCEPT')
[root@mgmt1 ~]# echo $?
0
[root@mgmt1 ~]# echo $output

[root@mgmt1 ~]#
```

rohityadavcloud, line 2239:
```suggestion
        String result = command.execute();
        if (result != null) {
            s_logger.warn("Error in allowing outgoing to " + destCidr + ", err=" + result);
            return "Error in allowing outgoing to " + destCidr + ", err=" + result;
```

services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
```diff
@@ -2242,7 +2242,7 @@ public String allowOutgoingOnPrivate(String destCidr) {
 
         addRouteToInternalIpOrCidr(_localgw, _eth1ip, _eth1mask, destCidr);
 
-        return null;
+        return msg;
```
rohityadavcloud, line 2245:
```suggestion
        return null;
```

rohityadavcloud:
This is extremely rare case, reverted other changes except the synchornised fix.
@blueorangutan package

rohityadavcloud:
@blueorangutan package

shwstppr (commented):
LGTM
Maven build failure in GH actions not related to the FR

shwstppr (approved):

weizhouapache (approved):
small change, make sense.

weizhouapache:
@blueorangutan test

DaanHoogland (approved):

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@blueorangutan package

weizhouapache:
@blueorangutan package

rohityadavcloud:
@blueorangutan test matrix

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

DaanHoogland:
@blueorangutan test matrix

services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
```diff
@@ -2225,7 +2225,7 @@ private String deleteLocalFile(String fullPath) {
         return null;
     }
 
-    public String allowOutgoingOnPrivate(String destCidr) {
+    public synchronized String allowOutgoingOnPrivate(String destCidr) {
```
DaanHoogland, line 2228: @rohityadavcloud could you add a javadoc with the description and background?

DaanHoogland:
@andrijapanicsb have you verified that this solves the problem?

services/secondary-storage/server/src/main/java/org/apache/cloudstack/storage/resource/NfsSecondaryStorageResource.java
DaanHoogland, line 2228:
```suggestion
    /**
    * allow *only one* setting of an outgoing destination at a time
    *
    * @destCidr the destination network that will be allowed for outgoing traffic.
    * @return any error message that might be helpful or <null> on success or when called anywhere but in the router VM.
    */
    public synchronized String allowOutgoingOnPrivate(String destCidr) {
```

andrijapanicsb: Could not be reproduced (the bug), so the solution as well could not be verified. The customer has confirmed (by testing in their env) that the "synchronized" seems to solve the issue on their end.

DaanHoogland:
@blueorangutan test matrix

PR #6896, opened 2022-11-13
start transition of CI according to apache infra reqs

DaanHoogland:
### Description
This PR transitions from travis to gha. it is trying to follow https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-travis-ci-to-github-actions

An alternative approach would be to use: https://github.com/marketplace/actions/run-travis-yml (parallel PR?)

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

so far only in the original branch on the PR repo. It can be replicated to your own.

PR #6956, opened 2022-12-07
Migrate Travis to Github Actions: Simulator based CI

rohityadavcloud:
ASF Infra has advised Travis will be deprecated and removed by Dec '22 end. This PR migrates our PR CI from Travis to Github Actions.

All workflow/jobs run on default spec:
2-core CPU (x86_64)
7 GB of RAM
14 GB of SSD space

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources


rohityadavcloud:
Referencing https://github.com/apache/cloudstack/pull/6896 but this is to try a different Github Actions approach

rohityadavcloud:
Appears I’m getting close, now to figure out the best way to pass the data, fix Marvin tests dependency issues.

rohityadavcloud:
Request for review - @DaanHoogland @shwstppr @nvazquez @weizhouapache @harikrishna-patnala  @davidjumani 

Finally it seems to be working for following:
- On commit push on PR cancels old jobs
- RAT checks
- UI Build (currently failing the unit test, which I confirmed is not related to this PR but happening on main)
- Simulator CI runs alright (most are passing now, those failing need to be investigated - either issue in the test or ACS code, or something in Github actions env)
- TBD: Unlike Travis, Github actions job timeout is 6hrs; should we keep these many jobs or have fewer?

Screenshot 2022-12-09 at 12 17 15 PM

ui/package.json
```diff
@@ -89,7 +89,6 @@
     "eslint-plugin-vue": "^7.0.0",
     "less": "^3.0.4",
     "less-loader": "^5.0.0",
-    "node-sass": "^4.0.0",
```
rohityadavcloud, line 92: cc @shwstppr @davidjumani @utchoang - pl check if we need node-sass? Removing this unblocks us to use any node version (I've tested both node v16 and v18). I did some quick build and run checks and I don't see any issues.

pom.xml
```diff
@@ -1024,6 +1024,7 @@
                             <exclude>debian/rules</exclude>
                             <exclude>debian/source/format</exclude>
                             <exclude>dist/console-proxy/js/jquery.js</exclude>
+                            <exclude>engine/schema/dist/**</exclude>
```
rohityadavcloud, line 1027: Added this to exclude, the dist isn't in the source but gets created during maven build.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,251 @@
...
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        tests: [ "smoke/test_certauthority_root",
```
rohityadavcloud, line 33: Copied the tests matrix list from Daan's PR, thanks @DaanHoogland

.github/workflows/build.yml
```diff
@@ -0,0 +1,66 @@
...
+          python3 --version
+          free -m
+          nproc
+          git status
+
+      - name: Noredist Build
+        run: |
+          git clone https://github.com/shapeblue/cloudstack-nonoss.git nonoss && cd nonoss && bash -x install-non-oss.sh && cd ..
+          rm -fr nonoss
+          mvn -B -P developer,systemvm -Dsimulator -Dnoredist clean install
```
rohityadavcloud, line 66: Separate noredist build gives 10-15 mins speed up for CI jobs which skip unit tests.

rohityadavcloud:
// kicking pkging only as there are changes in ui/package.json; otherwise all changes in .github for CI purpose
@blueorangutan package

rohityadavcloud:
@blueorangutan ui

All tests passing except the one with (intermittent) failure
Screenshot 2022-12-09 at 9 15 41 PM

rohityadavcloud:
@blueorangutan package

DaanHoogland:
@rohityadavcloud I restarted the failing job several times, but it seems to be consistent in failing. It is the tagging tests that fail.

DaanHoogland (approved):
looks good, not sure if we will now have the test_tags.py consistently failing, but I say deal with that later.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,280 @@
...
+      - name: Setup MySQL Server
+        run: |
+          # https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#mysql
+          sudo apt-get install -y mysql-server
+          sudo systemctl start mysql
+          sudo mysql -uroot -proot -e "ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY ''; FLUSH PRIVILEGES;"
+          sudo systemctl restart mysql
+          sudo mysql -uroot -e "SELECT VERSION();"
+
+      - name: Build with Maven
```
assignUser, line 237: This step builds on each of the matrix jobs (17x) for 10-15minutes so ~ 3 hours of runner time for something that IIUC (I am not a java person :D)  could be done in an initial job, the built package uploaded as artifacts and redistributed to the matrix jobs where it only needs to be installed. (this would also allow you to de-duplicate installation of some of the build dependencies I guess but those don't take as long) 

You can use `needs: <job_name>` to create a dependency between the build job and the matrix jobs and https://github.com/actions/upload-artifact to share the artifacts.

assignUser: with an over all runtime of 1d 10h 26m 24s any savings are important...

assignUser:
@DaanHoogland see my comment above :)

GutoVeronezi (commented):
Overall, code looks good.

Waiting for the reply on @assignUser reviews before approving.

.github/workflows/ci.yml
rohityadavcloud: Thanks @assignUser I thought about that, I'll explore. In fact because of the sheer amount of tests and duration I'm also thinking about:
- Running on tests based on type of changes (based on filesystem paths/wildcards)
- Context builds for example - not running these java builds&tests say if the changes are in UI
However, if Travis CI will be removed for asf repos by end of Dec, these optimisations can happen as a separate iteration/PR

DaanHoogland: so you are saying move (just) these three lines to another yml, right? can you point to an example for the plumbing @assignUser ?
I think we can leave this for an improvement btw, if the job runs we should start from there (I know, we will slack, but the build times will keep reminding us)

rohityadavcloud:
Thanks all - any thoughts on consolidating the component tests together (as Github Actions timeout is 6hrs, compared to an hour for Travis). What we can do is to have 5-6 runners per PR (than 20+) with pipelining and passing artifacts b/w jobs:

Check type of changes, do we need to build UI or java code -> Do mvn -Dnoredist -Dsimulator build with unit tests -> if it passes -> run smoketests (2-4 runners) and integration tests (2-4 runners)

.github/workflows/ci.yml
assignUser: @DaanHoogland  I think @rohityadavcloud is on it :D but yes you would add another `job:` with those build steps + artifact upload and dependency installation and have the other jobs depend on it with `needs: build-job`

Very much +1 on context builds, sounds good!

assignUser:
@rohityadavcloud that sounds good, do your tests use multiple cores by default? the gh runners have only 2 but still :D Feel free to ping me here or on asf slack if you are stuck with the implementation details!

DaanHoogland:
> Thanks all - any thoughts on consolidating the component tests together (as Github Actions timeout is 6hrs, compared to an hour for Travis). What we can do is to have 5-6 runners per PR (than 20+) with pipelining and passing artifacts b/w jobs:
> 
> Check type of changes, do we need to build UI or java code -> Do mvn -Dnoredist -Dsimulator build with unit tests -> if it passes -> run smoketests (2-4 runners) and integration tests (2-4 runners)

I don´t like the idea of consolidating if that lengthens the duration of the total run of checks. Our feedback loop will be effectively a day when we exceed 5 hours. Even with 2 hours some of us will only return the next day. Let's run more in parallel rather than less. especially if we separate the build part out.

weizhouapache (approved):
code lgtm

rohityadavcloud:
Logged optimisations/suggests at https://github.com/apache/cloudstack/issues/7010

PR #6875, opened 2022-11-08
Fix flaky tests in SMTPMailSenderTest.java and ReflectionToStringBuilderUtilsTest.java

yannizhou05:
### Description

The tests below in `ReflectionToStringBuilderUtilsTest.java` and `SMTPMailSenderTest.java` were found flaky because of the ordering of HashSet and result Arrays. The orders of elements in the HashSet and result Arrays are not the same every time the test being called. In this case, LinkedHashSet was applied and Arrays were sorted to ensure the orders do not cause the flakiness. This code change is trying to fix the [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html) mentioned above, because they sometimes fail (as the picture showed) and sometimes pass. The failure could be reproduced by the commands below by using the tool of [NonDex](https://github.com/TestingResearchIllinois/NonDex). The code change is to make sure the tests will always pass in this case.

- `org.apache.cloudstack.utils.mailing.SMTPMailSenderTest.setMailRecipientsTest`
- `org.apache.cloudstack.utils.reflectiontostringbuilderutils.ReflectionToStringBuilderUtilsTest.validateGetNonSelectedFieldsObjectIsACollectionAndValidSelectedFieldsMustReturnNonSelectedFields`
- `org.apache.cloudstack.utils.reflectiontostringbuilderutils.ReflectionToStringBuilderUtilsTest.validateGetNonSelectedFieldsObjectIsNotACollectionAndValidSelectedFieldsMustReturnNonSelectedFields`

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

The test failures could be reproduced by

1. `mvn install -pl utils -am -DskipTests`
2.  run tests with NonDex 
    `mvn -pl utils edu.illinois:nondex-maven-plugin:2.1:nondex -Dtest=org.apache.cloudstack.utils.mailing.SMTPMailSenderTest#setMailRecipientsTest`

PR #6967, opened 2022-12-10
Fixed flaky test in ResetVMUserDataCmdTest.java

tt0suzy:
### Description

Similar as [6875](https://github.com/apache/cloudstack/pull/6875).

The test `org.apache.cloudstack.api.command.test.ResetVMUserDataCmdTest.testUserdataDetails`, is a [flaky tests](https://docs.gitlab.com/ee/development/testing_guide/flaky_tests.html). It can pass mvn test but when run using the tool [NonDex](https://github.com/TestingResearchIllinois/NonDex), it fails. NonDex is a tool that will introduce non-determinism in certain java collections.

The cause of failure is because the test is comparing the string format of two hashmaps, however, as per [Oracle's documentation](https://docs.oracle.com/javase/7/docs/api/java/util/HashMap.html):

> The HashMap class is roughly equivalent to Hashtable, except that it is unsynchronized and permits nulls. This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time.

So the result of `toString()` is non-deterministic. Use hashmap's own equal methods is more reasonable.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

image

### How Has This Been Tested?

Steps to reproduce the failure:

Run the following command in cloudstack:

 1. First, build the module:
` mvn -pl api -am -DskipTests`
 2. Then run the test using NonDex
`mvn -pl api edu.illinois:nondex-maven-plugin:2.1:nondex -Dtest=org.apache.cloudstack.api.command.test.ResetVMUserDataCmdTest#testUserdataDetails`
The result will be saved under the module folder in .nondex

### DIFF
--- a/api/src/test/java/org/apache/cloudstack/api/command/test/ResetVMUserDataCmdTest.java
+++ b/api/src/test/java/org/apache/cloudstack/api/command/test/ResetVMUserDataCmdTest.java
@@ -132,7 +132,7 @@ public class ResetVMUserDataCmdTest {
         Map<String, String> result = cmd.getUserdataDetails();
 
         values1.putAll(values2);
-        Assert.assertEquals(values1.toString(), result.toString());
+        Assert.assertEquals(values1, result);
     }
 
 }


harikrishna-patnala (approved):
Tested the unit test locally and worked.

LGTM

rohityadavcloud (approved):

rohityadavcloud:
@blueorangutan package

PR #6925, opened 2022-11-28
Add and improve logs in snapshot scheduling

stephankruggg:
### Description

This PR adds and improves logs in the snapshot scheduling process in order to facilitate troubleshooting, as current logs are scarce.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [X] Minor

### How Has This Been Tested?

I applied the changes in a local lab and checked if the messages were being correctly logged.

PR #6973, opened 2022-12-12
conditional logging for prio

DaanHoogland:
### Description

This PR intents to guard against string evaluation in parameters to low level logging without sacrificing readability.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### DIFF
--- a/utils/src/main/java/com/cloud/utils/LogUtils.java
+++ b/utils/src/main/java/com/cloud/utils/LogUtils.java
@@ -27,6 +27,7 @@ import java.util.Set;
 import org.apache.log4j.Appender;
 import org.apache.log4j.FileAppender;
 import org.apache.log4j.Logger;
+import org.apache.log4j.Priority;
 import org.apache.log4j.xml.DOMConfigurator;
 
 public class LogUtils {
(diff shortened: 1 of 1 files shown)


GutoVeronezi:
@DaanHoogland 

I understand your point in avoiding building a String for low-level logging and not using it; however, I think this proposal is not the best option for handling the situation. As you pointed out in the comment https://github.com/apache/cloudstack/pull/6925#issuecomment-1346446696 and from conversations we had along the time, Log4j2 would solve our issues of string evaluation when it is not necessary. Unfortunately, the community could not proceed with the upgrade from Log4j to Log4j2, as it would demand a huge work effort, and also due to the lack of time and having other priorities; however, now we managed to put efforts into this huge change. We separated the whole project into two phases: 
1. Upgrade from Log4j to Log4j2;
2. Normalize the use of the Log4j2 syntax (as Log4j2 has back compatibility with Log4j, we could separate the scopes);

We are already working on the first phase; however, as it is a huge change, it is going to take some time until releasing the first draft. As soon as we get the first draft done, we are going to open a well-detailed PR explaining the changes and the effects they will have on ACS.

DaanHoogland:
@GutoVeronezi I appreciate the work.
Does this mean you do not approve of this change as a short term, intermediate step?

GutoVeronezi:
@DaanHoogland, IMHO, I do not think it worth the effort of adding an intermediate step; however, the changes are not mutually exclusive. Therefore, it is not a -1 at all, it is a -0.

DaanHoogland:
clearly voted away

Issue #6888, opened 2022-11-09
Slow mysql query encountered when browsing the Network Tab


s-seitz:
When using a broader IP Range (e.g multiple /16), a particular query is getting slow:

~~~
SELECT COUNT(user_ip_address.public_ip_address) FROM user_ip_address  INNER JOIN vlan ON user_ip_address.vlan_db_id=vlan.id WHERE user_ip_address.data_center_id  = 1 AND  (vlan.vlan_type = 'DirectAttached' );
~~~

##### ISSUE TYPE
 * Improvement Request

##### COMPONENT NAME
API, Mysql Backend
-->

##### CLOUDSTACK VERSION
4.17.1.0

##### CONFIGURATION
Advanced Networking, VLAN isolation

##### OS / ENVIRONMENT
Ubuntu 22.04 LTS

##### SUMMARY
When using a broader IP Range (e.g multiple /16), a particular query is getting slow:

~~~
SELECT COUNT(user_ip_address.public_ip_address) FROM user_ip_address  INNER JOIN vlan ON user_ip_address.vlan_db_id=vlan.id WHERE vlan.vlan_type = 'DirectAttached' AND user_ip_address.data_center_id  = 1  AND user_ip_address.state != 'Free'  AND user_ip_address.removed IS NULL;
~~~

This could be resolved by adding a (missing) key:
```
ALTER TABLE `user_ip_address` ADD KEY `fk_user_ip_address__state` (`state`);
```

##### STEPS TO REPRODUCE
Having a few /16 IPv4 Ranges assigned,
Enable mysql slowlog with a timeout of e.g. 2 seconds, and watch it happen.

##### EXPECTED RESULTS
Having the result faster.

##### ACTUAL RESULTS
Waiting for the result for about 2 seconds. (At least for my setup)

rohityadavcloud:
Is this critical/blocker issue @DaanHoogland ?

DaanHoogland:
@rohityadavcloud , for anybody using /16 nets it is a blocker and I don´t know a workaround. I would say it is critical. Any reason it would not be?

DaanHoogland:
@s-seitz I am assuming you tested this to resolve your issue?
If not please try with PR #7028

PR #7028, opened 2022-12-27
add index to speed up querying IPs in the network-tab

DaanHoogland:
### Description

This PR...

Fixes: #6888

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
@@ -869,3 +869,17 @@ WHERE   usage_unit = 'Policy-Month';
 
 -- delete configuration task.cleanup.retry.interval #6910
 DELETE FROM `cloud`.`configuration` WHERE name='task.cleanup.retry.interval';
+
+--- #6888 add index to speed up querying IPs in the network-tab
+DROP PROCEDURE IF EXISTS `cloud`.`IDEMPOTENT_ADD_KEY`;
+
+CREATE PROCEDURE `cloud`.`IDEMPOTENT_ADD_KEY` (
+		IN in_index_name VARCHAR(200)
+    , IN in_table_name VARCHAR(200)
+    , IN in_key_definition VARCHAR(1000)
+)
+BEGIN
+
+    DECLARE CONTINUE HANDLER FOR 1061 BEGIN END; SET @ddl = CONCAT('ALTER TABLE ', in_tablename); SET @ddl = CONCAT(@ddl, ' ', ' ADD KEY ') ; SET @ddl = CONCAT(@ddl, ' ', in_index_name); SET @ddl = CONCAT(@ddl, ' ', in_index_definition); PREPARE stmt FROM @ddl; EXECUTE stmt; DEALLOCATE PREPARE stmt; END;
+
+CALL `cloud`.`IDEMPOTENT_ADD_KEY`('i_user_ip_address_state','user_ip_address', '(state)');
--- a/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
@@ -880,6 +880,6 @@ CREATE PROCEDURE `cloud`.`IDEMPOTENT_ADD_KEY` (
 )
 BEGIN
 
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -869,3 +869,6 @@ WHERE   usage_unit = 'Policy-Month';
 
 -- delete configuration task.cleanup.retry.interval #6910
 DELETE FROM `cloud`.`configuration` WHERE name='task.cleanup.retry.interval';
+
+--- #6888 add index to speed up querying IPs in the network-tab
+ALTER TABLE `user_ip_address` ADD KEY `fk_user_ip_address__state` (`state`);
```
weizhouapache, line 874: @DaanHoogland 
`state` is not a foreign key.
It would be good to use `i_user_ip_address__state`

might be better to create a procedure similar as IDEMPOTENT_ADD_FOREIGN_KEY ?

DaanHoogland: The procedure should already exist, so calling like
```suggestion
CALL `cloud`.`IDEMPOTENT_CREATE_UNIQUE_INDEX`('uk_user_ip_address_state','user_ip_address', '(state)');
```

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -869,3 +869,6 @@ WHERE   usage_unit = 'Policy-Month';
 
 -- delete configuration task.cleanup.retry.interval #6910
 DELETE FROM `cloud`.`configuration` WHERE name='task.cleanup.retry.interval';
+
+--- #6888 add index to speed up querying IPs in the network-tab
+CALL `cloud`.`IDEMPOTENT_CREATE_UNIQUE_INDEX`('uk_user_ip_address_state','user_ip_address', '(state)');
```
weizhouapache, line 874: if we add two shared  networks with same ip range but different vlans, will it be broken ?
they  have the same ip addresses and state (Free by default).

DaanHoogland: :face_with_spiral_eyes: right, it shouldn´t be unique. I start over really thinking this through a bit.

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -869,3 +869,17 @@ WHERE   usage_unit = 'Policy-Month';
...
+DROP PROCEDURE IF EXISTS `cloud`.`IDEMPOTENT_ADD_KEY`;
+
+CREATE PROCEDURE `cloud`.`IDEMPOTENT_ADD_KEY` (
+		IN in_index_name VARCHAR(200)
+    , IN in_table_name VARCHAR(200)
+    , IN in_key_definition VARCHAR(1000)
+)
+BEGIN
+
+    DECLARE CONTINUE HANDLER FOR 1061 BEGIN END; SET @ddl = CONCAT('ALTER TABLE ', in_tablename); SET @ddl = CONCAT(@ddl, ' ', ' ADD KEY ') ; SET @ddl = CONCAT(@ddl, ' ', in_index_name); SET @ddl = CONCAT(@ddl, ' ', in_index_definition); PREPARE stmt FROM @ddl; EXECUTE stmt; DEALLOCATE PREPARE stmt; END;
```
weizhouapache, line 883:
```suggestion
    DECLARE CONTINUE HANDLER FOR 1061 BEGIN END; SET @ddl = CONCAT('ALTER TABLE ', in_table_name); SET @ddl = CONCAT(@ddl, ' ', ' ADD KEY ') ; SET @ddl = CONCAT(@ddl, ' ', in_index_name); SET @ddl = CONCAT(@ddl, ' ', in_key_definition); PREPARE stmt FROM @ddl; EXECUTE stmt; DEALLOCATE PREPARE stmt; END;
```

with this change, the CALL works. @DaanHoogland

DaanHoogland:
@blueorangutan package

weizhouapache:
@blueorangutan test

DaanHoogland:
@blueorangutan help

DaanHoogland:
@blueorangutan test ubuntu22 kvm-alma8

DaanHoogland:
@blueorangutan test rocky8 kvm-ubuntu20

rohityadavcloud (approved):
Didn't check/verify, but generally we know from past that tables with specific index help to speed up list APIs (same goes for using views instead of tables)

PR #7032, opened 2022-12-28
Allow users to inform timezones on APIs that have the date parameter

JoaoJandre:
Although ACS stores all dates in the DB with the UTC timezone, all APIs that have parameters of type date are handled according to the server's timezone. In this case, if the server's timezone is in UTC-3, for instance, and user is in UTC-2 and inform `yyyy-MM-dd 09:00:00` in the parameter, the Date object will be `yyyy-MM-dd 09:00:00` (UTC-3) and ACS will convert it to `yyyy-MM-dd 12:00:00` while querying the database. This way, if the final users want to retrieve data from date `09:00:00`, they must know the server's timezone in order to inform the correspondent date (in this example, `yyyy-MM-dd 06:00:00`).

This situation is very clear when calling the `listvirtualmachinesusagehistory` API. If the data was collected at `yyyy-MM-dd 12:00:00` (UTC), final user must know that the server is in UTC-3 and inform `yyyy-MM-dd 09:00:00`. Via UI, either using the local timezone or not, the informed timestamp is always converted to UTC. Then, the server would consider this converted timestamp as UTC-3, and would convert it to UTC again, adding 3 hours to the date and returning nothing. Example: user is in UTC-2 and informs `yyyy-MM-dd 10:00:00`; the UI converts it to UTC (`yyyy-MM-dd 12:00:00`); ACS interprets it as UTC-3 and converts again to UTC (`yyyy-MM-dd 15:00:00`).

To handle this situation, an extension was made to allow all APIs that have parameters of type date to also receive them in the format with timezone `yyyy-MM-dd'T'HH:mm:ssZ`, for example, `2022-12-20T14:33:42+0100`. This change only extends the date parameter, previous behavior still works normally. Thus the user will be able to send requests with the chosen timezone, and the ACS will carry out the conversion to the timezone it operates.

Furthermore, the VM stats listing in the UI has been updated to always send the date and time with the timezone of the user's machine.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

After setting the management server's timezone to UTC+1, and setting `vm.stats.max.retention.time` to 5, I did the following tests: 
| # | Test | Result | 
| ------ | ------ | ------ |
| 1 | I called the `listvirtualmachinesusagehistory` API with the `startdate` as 5 minutes prior to the tests time, passing my timezone (UTC-3) | The stats were correctly recovered | 
| 2 | I called the API with the same timestamp, but informing UTC-4 as the timezone | No records were returned|
| 3 | I called the API with the `startdate` as 5 minutes prior to the tests time, informing the date and time using the UTC+1 timezone, but without informing the timezone, e.g. `2022-12-28 15:10:00` | The stats were correctly recovered |
| 4 | I called the API with the `startdate` as 5 minutes prior to the tests time, informing the date and time using the UTC-3 timezone, but without informing the timezone | No stats were recovered |
| 5 | I called the listEvents API with the `startDate` in the `YYYY-MM-DD HH:mm:ss` format | the events were properly listed |

### DIFF
--- a/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
+++ b/server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
@@ -63,8 +63,10 @@ import com.cloud.utils.exception.CloudRuntimeException;
 public class ParamProcessWorker implements DispatchWorker {
 
     private static final Logger s_logger = Logger.getLogger(ParamProcessWorker.class.getName());
-    public final DateFormat inputFormat = new SimpleDateFormat("yyyy-MM-dd");
-    public final DateFormat newInputFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+    private final String inputFormatString = "yyyy-MM-dd";
+    private final String newInputFormatString = "yyyy-MM-dd HH:mm:ss";
+    public final DateFormat inputFormat = new SimpleDateFormat(inputFormatString);
+    public final DateFormat newInputFormat = new SimpleDateFormat(newInputFormatString);
 
     @Inject
     protected AccountManager _accountMgr;
@@ -335,13 +337,20 @@ public class ParamProcessWorker implements DispatchWorker {
             case DATE:
                 // This piece of code is for maintaining backward compatibility
                 // and support both the date formats(Bug 9724)
-                final boolean isObjInNewDateFormat = isObjInNewDateFormat(paramObj.toString());
-                if (isObjInNewDateFormat) {
+                try {
+                    field.set(cmdObj, DateUtil.parseTZDateString(paramObj.toString()));
+                    break;
+                } catch (ParseException parseException) {
+                    s_logger.debug(String.format("Could not parse date [%s] with timezone parser, trying to parse without timezone.", paramObj));
+                }
+                if (isObjInNewDateFormat(paramObj.toString())) {
+                    s_logger.debug(String.format("Parsing date [%s] using the [%s] format.", paramObj, newInputFormatString));
                     final DateFormat newFormat = newInputFormat;
                     synchronized (newFormat) {
                         field.set(cmdObj, newFormat.parse(paramObj.toString()));
                     }
                 } else {
+                    s_logger.debug(String.format("Parsing date [%s] using the [%s] format.", paramObj, inputFormatString));
                     final DateFormat format = inputFormat;
                     synchronized (format) {
                         Date date = format.parse(paramObj.toString());
@@ -427,9 +436,6 @@ public class ParamProcessWorker implements DispatchWorker {
                         field.set(cmdObj, paramObj.toString());
                     }
                 }
-                break;
-            case TZDATE:
-                field.set(cmdObj, DateUtil.parseTZDateString(paramObj.toString()));
                 break;
             case MAP:
             default:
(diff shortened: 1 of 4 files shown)


server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
```diff
@@ -63,8 +63,10 @@
 public class ParamProcessWorker implements DispatchWorker {
 
     private static final Logger s_logger = Logger.getLogger(ParamProcessWorker.class.getName());
-    public final DateFormat inputFormat = new SimpleDateFormat("yyyy-MM-dd");
-    public final DateFormat newInputFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+    private final String inputFormatString = "yyyy-MM-dd";
+    private final String newInputFormatString = "yyyy-MM-dd HH:mm:ss";
+    public final DateFormat inputFormat = new SimpleDateFormat(inputFormatString);
+    public final DateFormat newInputFormat = new SimpleDateFormat(newInputFormatString);
```
DaanHoogland, line 69:
```suggestion
    private final static String inputFormatString = "yyyy-MM-dd";
    private final static String newInputFormatString = "yyyy-MM-dd HH:mm:ss";
    public final static DateFormat inputFormat = new SimpleDateFormat(inputFormatString);
    public final static DateFormat newInputFormat = new SimpleDateFormat(newInputFormatString);
```

server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
```diff
@@ -335,13 +337,20 @@ private void setFieldValue(final Field field, final BaseCmd cmdObj, final Object
...
                 // This piece of code is for maintaining backward compatibility
                 // and support both the date formats(Bug 9724)
-                final boolean isObjInNewDateFormat = isObjInNewDateFormat(paramObj.toString());
-                if (isObjInNewDateFormat) {
+                try {
+                    field.set(cmdObj, DateUtil.parseTZDateString(paramObj.toString()));
+                    break;
+                } catch (ParseException parseException) {
+                    s_logger.debug(String.format("Could not parse date [%s] with timezone parser, trying to parse without timezone.", paramObj));
+                }
```
DaanHoogland, line 345: Can you put this, or even the full case DATE block in a separate method, please_

DaanHoogland (approved):
clgtm

stephankruggg (approved):

server/src/main/java/com/cloud/api/dispatch/ParamProcessWorker.java
```diff
@@ -442,6 +422,34 @@ private void setFieldValue(final Field field, final BaseCmd cmdObj, final Object
                     " is not accessible]");
         }
     }
+
```
DaanHoogland, line 425:
```suggestion

```
trailing spaces

BryanMLima (approved):
lgtm

JoaoJandre:
I did some slight adjusting to this PR so that when opening the stats tab, we always consider the users timezone to set the default period.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

PR #7037, opened 2022-12-30
Fix volume snapshot in a VM with an ISO attached

stephankruggg:
### Description

When taking a snapshot of a volume in a VM with an ISO attached, ACS attempts to define the label of the target associated with the ISO by its ID. However, this ID may not be correctly mapped to the label, which causes the snapshot to fail.

This PR fixes this by using the defined label, instead of defining it during the snapshot process by its ID.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

In a local lab, I created 3 VMs with ISOs attached: VM1, VM2 and VM3. VM1 had 1 volume, VM2 had 2 volumes, and VM3 had 3 volumes.  The label of the target associated to the ISO was `hdc` for all of them. 

When taking a snapshot for the volume in VM1, the label was incorrectly generated as `hdb`, which caused the snapshot to fail.
When taking snapshots for the volumes in VM2, the label was correctly generated as `hdc`, and the snapshots succeeded.
When taking snapshots for the volumes in VM3, the label was incorrectly generated as `hdd`, which caused the snapshots to fail.

After applying the changes, I took snapshots for every volume of the VMs. In all of them the label was correctly passed as `hdc`, and all snapshots succeeded.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtDomainXMLParser.java
@@ -121,7 +121,7 @@ public class LibvirtDomainXMLParser {
                             }
                             def.defFileBasedDisk(diskFile, diskLabel, DiskDef.DiskBus.valueOf(bus.toUpperCase()), fmt);
                         } else if (device.equalsIgnoreCase("cdrom")) {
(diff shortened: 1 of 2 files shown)


BryanMLima (approved):

weizhouapache (approved):
code lgtm

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

PR #6139, opened 2022-03-18
agent: Detect existing hosts with UEFI support

davidjumani:
### Description

Fixes https://github.com/apache/cloudstack/issues/6087
Adds support to detect existing hosts with UEFI capability

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [x] Minor

### How Has This Been Tested?

- Bring up a 4.13 env with VMware 6.5 (anything that supports UEFI)
- Upgrade to latest main 
- Try to deploy a VM with UEFI
- Deployment now passes Previously failed to find a suitable host as none were UEFI compatible

#### Before :
```
MariaDB [cloud]> select * from host_details where name = 'host.uefi.enable' \G
Empty set (0.00 sec)

```

#### After :
```
MariaDB [cloud]> select * from host_details where name = 'host.uefi.enable' \G
*************************** 1. row ***************************
     id: 34
host_id: 1
   name: host.uefi.enable
  value: true
*************************** 2. row ***************************
     id: 40
host_id: 2
   name: host.uefi.enable
  value: true
2 rows in set (0.00 sec)

```

PR #7084, opened 2023-01-12
kvm: make UEFI host check to support both Ubuntu and EL

rohityadavcloud:
Small fix to also support UEFI check on agent connect. Refer: https://github.com/apache/cloudstack/pull/6139/files

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
@@ -42,15 +42,18 @@ public final class LibvirtReadyCommandWrapper extends CommandWrapper<ReadyComman
     public Answer execute(final ReadyCommand command, final LibvirtComputingResource libvirtComputingResource) {
         Map<String, String> hostDetails = new HashMap<String, String>();
 
-        if (hostSupportsUefi() && libvirtComputingResource.isUefiPropertiesFileLoaded()) {
+        if (hostSupportsUefi(libvirtComputingResource.isBaseOsUbuntu()) && libvirtComputingResource.isUefiPropertiesFileLoaded()) {
             hostDetails.put(Host.HOST_UEFI_ENABLE, Boolean.TRUE.toString());
         }
 
         return new ReadyAnswer(command, hostDetails);
     }
 
-    private boolean hostSupportsUefi() {
+    private boolean hostSupportsUefi(boolean isUbuntuHost) {
         String cmd = "rpm -qa | grep -i ovmf";
+        if (isUbuntuHost) {
+            cmd = "dpkg -l ovmf";
+        }
         s_logger.debug("Running command : " + cmd);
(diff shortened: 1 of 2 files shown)


rohityadavcloud:
Didn't test it myself, based on some code review.
@blueorangutan package

rohityadavcloud:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@blueorangutan help

weizhouapache (approved):
code lgtm

Need minor change on trillian so we can test the UEFI feature on ubuntu oses.

rohityadavcloud:

rohityadavcloud:

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
```diff
@@ -42,18 +42,21 @@ public final class LibvirtReadyCommandWrapper extends CommandWrapper<ReadyComman
...
 
         return new ReadyAnswer(command, hostDetails);
     }
 
-    private boolean hostSupportsUefi() {
+    private boolean hostSupportsUefi(boolean isUbuntuHost) {
         String cmd = "rpm -qa | grep -i ovmf";
+        if (isUbuntuHost) {
+            cmd = "dpkg -l ovmf";
+        }
```
DaanHoogland, line 56:
```suggestion
        String cmd = null;
        if (isUbuntuHost) {
            cmd = "dpkg -l ovmf";
        } else {
            cmd = "rpm -qa | grep -i ovmf";
        }
```

DaanHoogland (approved):
clgtm

DaanHoogland:
@rohityadavcloud, will this require extra testing?
Will you apply my suggestion?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
rohityadavcloud: I prefer to not have the cmd set to null when we know that's the default case.

rohityadavcloud:
@DaanHoogland only if you insist, I prefer to set cmd to the default case (and not null; just the style to having less branching read better to me). I think it may not need additional testing; what we're doing is to enable the case when Ubuntu based KVM hosts. I've done testing of the pkg name and dpkg -l output manually to verify that it returns non-zero exit code when the pkg is not installed.

DaanHoogland:
> @DaanHoogland only if you insist, I prefer to set cmd to the default case (and not null; just the style to having less branching read better to me). I think it may not need additional testing; what we're doing is to enable the case when Ubuntu based KVM hosts. I've done testing of the pkg name and dpkg -l output manually to verify that it returns non-zero exit code when the pkg is not installed.

Not insisting, Ilike @weizhouapache 's suggestion better anyway ;)

shwstppr (approved):

PR #7094, opened 2023-01-13
Handle console session in multiple management servers

GutoVeronezi:
### Description

PR #6577 introduced a new mechanism of validation of console sessions, which makes the access tokens one time usable in order to avoid reply attacks and prevent access to the console in case of data leak. When generating the console token, the management server (MS) that processed the request saves it in memory for further validation. When using the token, a request is sent to the CPVM that will validate it (the token) against the MS. 

In an environment that has more than one MS, the CPVM will establish communication with the first available MS that it founds. In this case, if the CPVM communicates with `MS A` and the console token is generated by `MS A`, the token will work; however, if the CPVM communicates with `MS A` and the console token is generated by `MS B`, the token will not work because the token is being validated against `MS A` and the information is in the memory of `MS B`.

This PR intends to handle this situation by temporally saving the console token in the database instead of saving it in the memory of the MS that generated it. This way, independently of which MS generated the token or which MS the CPVM communicates with, the token will be validated. After using the token, it is removed from the database and also from the CPVM.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

image

### How Has This Been Tested?
I created an environment with 2 MS (A and B) and started only A. I enabled the zone in order to create the CPVM. After that, guaranting that the CPVM was communicating with MS A, I started MS B. 
Previous to the changes, I executed two test cases:
- generated a new console access URL (with the API `createConsoleEndpoint`) through MS A and used it. As the CPVM was communicating with MS A, it worked.
- generated a new console access URL (with the API `createConsoleEndpoint`) through MS B and used it. As the CPVM was communicating with MS A, it did not worked, presenting the message `Failed to connect to server / access token has expired`

With the changes, I executed again the two test cases:
- generated a new console access URL (with the API `createConsoleEndpoint`) through MS A and used it. As the CPVM was communicating with MS A, it worked.
- generated a new console access URL (with the API `createConsoleEndpoint`) through MS B and used it. As the CPVM was communicating with MS A, it worked too, because the session was being stored in the database.

After using the URLs, I checked the database and verified that there were no entries in `cloud.allowed_console_session`, as expected.

### DIFF
--- a/engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java
+++ b/engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java
@@ -1,0 +1,64 @@
+//
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+//
+
+
+package com.cloud.vm;
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "allowed_console_session")
+public class AllowedConsoleSessionVo {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    @Column(name = "id")
+    private long id;
+
+    @Column(name = "session_uuid")
+    private String sessionUuid;
+
+    public AllowedConsoleSessionVo() {
+    }
+
+    public AllowedConsoleSessionVo(String sessionUuid) {
+        this.sessionUuid = sessionUuid;
+    }
+
+    public long getId() {
+        return id;
+    }
+
+    public void setId(long id) {
+        this.id = id;
+    }
+
+    public String getSessionUuid() {
+        return sessionUuid;
+    }
+
+    public void setSessionUuid(String sessionUuid) {
+        this.sessionUuid = sessionUuid;
+    }
+}
--- a/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
@@ -1036,15 +1036,20 @@ INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`)
 SELECT UUID(), `roles`.`id`, 'isAccountAllowedToCreateOfferingsWithTags', 'ALLOW'
 FROM `cloud`.`roles` WHERE `role_type` = 'DomainAdmin';
 
---- Create table for handling allowed console session temporally #7094
+--- Create table for handling console sessions #7094
 
 CREATE TABLE IF NOT EXISTS `cloud`.`console_session` (
     `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
     `uuid` varchar(40) NOT NULL COMMENT 'UUID generated for the session',
     `created` datetime NOT NULL COMMENT 'When the session was created',
-    `account_id` bigint(20) unsigned NOT NULL COMMENT 'Account who generated the session' references cloud.account(id),
-    `user_id` bigint(20) unsigned NOT NULL COMMENT 'User who generated the session' references `cloud`.`user`(`id`),
-    `instance_id` bigint(20) unsigned NOT NULL COMMENT 'VM for which the session was generated' references `cloud`.`vm_instance`(`id`) ,
-    `host_id` bigint(20) unsigned NOT NULL COMMENT 'Host where the VM was when the session was generated' references `cloud`.`host`(`id`) ,
-    `removed` datetime COMMENT 'When the session was removed/used'
+    `account_id` bigint(20) unsigned NOT NULL COMMENT 'Account who generated the session',
+    `user_id` bigint(20) unsigned NOT NULL COMMENT 'User who generated the session',
+    `instance_id` bigint(20) unsigned NOT NULL COMMENT 'VM for which the session was generated',
+    `host_id` bigint(20) unsigned NOT NULL COMMENT 'Host where the VM was when the session was generated',
+    `removed` datetime COMMENT 'When the session was removed/used',
+    CONSTRAINT `fk_consolesession__account_id` FOREIGN KEY(`account_id`) REFERENCES `cloud`.`account` (`id`),
+    CONSTRAINT `fk_consolesession__user_id` FOREIGN KEY(`user_id`) REFERENCES `cloud`.`user`(`id`),
+    CONSTRAINT `fk_consolesession__instance_id` FOREIGN KEY(`instance_id`) REFERENCES `cloud`.`vm_instance`(`id`),
+    CONSTRAINT `fk_consolesession__host_id` FOREIGN KEY(`host_id`) REFERENCES `cloud`.`host`(`id`),
+    CONSTRAINT `uc_consolesession__uuid` UNIQUE (`uuid`)
 );
--- a/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
@@ -1040,5 +1040,11 @@ FROM `cloud`.`roles` WHERE `role_type` = 'DomainAdmin';
 
 CREATE TABLE IF NOT EXISTS `cloud`.`console_session` (
     `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
-    `session_uuid` VARCHAR(40) NOT NULL
+    `uuid` varchar(40) NOT NULL COMMENT 'UUID generated for the session',
+    `created` datetime NOT NULL COMMENT 'When the session was created',
+    `account_id` bigint(20) unsigned NOT NULL COMMENT 'Account who generated the session' references cloud.account(id),
+    `user_id` bigint(20) unsigned NOT NULL COMMENT 'User who generated the session' references `cloud`.`user`(`id`),
+    `instance_id` bigint(20) unsigned NOT NULL COMMENT 'VM for which the session was generated' references `cloud`.`vm_instance`(`id`) ,
+    `host_id` bigint(20) unsigned NOT NULL COMMENT 'Host where the VM was when the session was generated' references `cloud`.`host`(`id`) ,
+    `removed` datetime COMMENT 'When the session was removed/used'
 );
--- a/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
@@ -1035,3 +1035,10 @@ WHERE   role_id = (SELECT id FROM `cloud`.`roles` WHERE name = 'Read-Only User -
 INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`)
 SELECT UUID(), `roles`.`id`, 'isAccountAllowedToCreateOfferingsWithTags', 'ALLOW'
 FROM `cloud`.`roles` WHERE `role_type` = 'DomainAdmin';
+
+--- Create table for handling allowed console session temporally
+
+CREATE TABLE `cloud`.`allowed_console_session` (
+    `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    `session_uuid` VARCHAR(40) NOT NULL
+);
--- a/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
@@ -1036,9 +1036,9 @@ INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`)
 SELECT UUID(), `roles`.`id`, 'isAccountAllowedToCreateOfferingsWithTags', 'ALLOW'
 FROM `cloud`.`roles` WHERE `role_type` = 'DomainAdmin';
 
---- Create table for handling allowed console session temporally
+--- Create table for handling allowed console session temporally #7094
 
-CREATE TABLE `cloud`.`allowed_console_session` (
+CREATE TABLE IF NOT EXISTS `cloud`.`console_session` (
     `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
     `session_uuid` VARCHAR(40) NOT NULL
 );
--- a/server/src/main/java/org/apache/cloudstack/consoleproxy/ConsoleAccessManagerImpl.java
+++ b/server/src/main/java/org/apache/cloudstack/consoleproxy/ConsoleAccessManagerImpl.java
@@ -63,10 +63,8 @@ import javax.naming.ConfigurationException;
 
 import java.util.Arrays;
 import java.util.Date;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.UUID;
 
 public class ConsoleAccessManagerImpl extends ManagerBase implements ConsoleAccessManager {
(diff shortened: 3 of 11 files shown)


engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java
```diff
@@ -0,0 +1,64 @@
...
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "allowed_console_session")
+public class AllowedConsoleSessionVo {
```
DaanHoogland, line 32: Can this be just `console_session` and `ConsoleSessionVO`?

weizhouapache, line 32: normally we use "VO" instead of "Vo"

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -1035,3 +1035,10 @@ WHERE   role_id = (SELECT id FROM `cloud`.`roles` WHERE name = 'Read-Only User -
 INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`)
 SELECT UUID(), `roles`.`id`, 'isAccountAllowedToCreateOfferingsWithTags', 'ALLOW'
 FROM `cloud`.`roles` WHERE `role_type` = 'DomainAdmin';
+
+--- Create table for handling allowed console session temporally
+
+CREATE TABLE `cloud`.`allowed_console_session` (
```
weizhouapache, line 1041: may be better to use `allowed_vm_console_sessions` ?

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -1035,3 +1035,10 @@ WHERE   role_id = (SELECT id FROM `cloud`.`roles` WHERE name = 'Read-Only User -
...
+    `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
weizhouapache, line 1042: foreign key ?

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
weizhouapache, line 1041: create table if not exists

DaanHoogland (commented):
except for the name part 'allowed', which seems to be immplied in any case, looks good.
Also I am still not comfortable approving logs that are
1. debug
2. expanding parameters
3. not having a guard with isDebugEnabled()
You have an PR in the making that will deal with that, can you share something about that?

DaanHoogland:
@blueorangutan package

engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java
nvazquez: @GutoVeronezi what do you think of reusing the console_proxy table for this (there is a column already for this: `session_details`) instead of a new table?

DaanHoogland:
@blueorangutan test matrix

engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java
rohityadavcloud: Agree, if this is just one column we need that we think all VMs should have, reusing existing table makes sense.

DaanHoogland:
@blueorangutan test matrix

GutoVeronezi:
> except for the name part 'allowed', which seems to be immplied in any case, looks good. Also I am still not comfortable approving logs that are
> 
> 1. debug
> 2. expanding parameters
> 3. not having a guard with isDebugEnabled()
>    You have an PR in the making that will deal with that, can you share something about that?

@DaanHoogland, regarding the PR, we are working to open it today. As soon as it is opened, we ping you.

engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java
nvazquez: LGTM

nvazquez:
@GutoVeronezi I've tested it on a single mgmt server and I'm not able to access any console, could you please verify?

````
2023-01-16 16:44:48,918 INFO  [cloud.consoleproxy.ConsoleProxyResourceHandler] (Thread-161:null) Get resource request for /resource/noVNC/vnc.html
2023-01-16 16:44:48,928 INFO  [cloud.consoleproxy.ConsoleProxyResourceHandler] (Thread-161:null) Sent file /resource/noVNC/vnc.html with content type text/html
2023-01-16 16:44:49,900 INFO  [cloud.consoleproxy.ConsoleProxyNoVNCHandler] (qtp465655017-41:null) Get websocket connection request from remote IP : 10.0.3.251
2023-01-16 16:44:49,955 WARN  [cloud.consoleproxy.ConsoleProxy] (qtp465655017-41:null) External authenticator failed authentication request for vm 60461a74-61a0-4951-b496-39f88e43b151 with sid zGvr4WyQaxJhz_lV7jCYfw
2023-01-16 16:44:49,955 WARN  [cloud.consoleproxy.ConsoleProxyNoVNCHandler] (qtp465655017-41:null) Failed to create viewer due to External authenticator failed request for vm 60461a74-61a0-4951-b496-39f88e43b151 with sid zGvr4WyQaxJhz_lV7jCYfw
com.cloud.consoleproxy.AuthenticationException: External authenticator failed request for vm 60461a74-61a0-4951-b496-39f88e43b151 with sid zGvr4WyQaxJhz_lV7jCYfw
	at com.cloud.consoleproxy.ConsoleProxy.authenticationExternally(ConsoleProxy.java:561)
	at com.cloud.consoleproxy.ConsoleProxy.getNoVncViewer(ConsoleProxy.java:595)
	at com.cloud.consoleproxy.ConsoleProxyNoVNCHandler.onConnect(ConsoleProxyNoVNCHandler.java:143)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at org.eclipse.jetty.websocket.common.events.annotated.CallableMethod.call(CallableMethod.java:70)
	at org.eclipse.jetty.websocket.common.events.JettyAnnotatedEventDriver.onConnect(JettyAnnotatedEventDriver.java:178)
	at org.eclipse.jetty.websocket.common.events.AbstractEventDriver.openSession(AbstractEventDriver.java:217)
	at org.eclipse.jetty.websocket.common.WebSocketSession.open(WebSocketSession.java:493)
	at org.eclipse.jetty.websocket.common.WebSocketSession.onOpened(WebSocketSession.java:459)
	at org.eclipse.jetty.io.AbstractConnection.onOpened(AbstractConnection.java:213)
	at org.eclipse.jetty.io.AbstractConnection.onOpen(AbstractConnection.java:205)
	at org.eclipse.jetty.io.AbstractEndPoint.upgrade(AbstractEndPoint.java:444)
	at org.eclipse.jetty.server.HttpConnection.onCompleted(HttpConnection.java:412)
	at org.eclipse.jetty.server.HttpChannel.onCompleted(HttpChannel.java:919)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:467)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:277)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
	at org.eclipse.jetty.io.ChannelEndPoint$1.run(ChannelEndPoint.java:104)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:883)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1034)
	at java.base/java.lang.Thread.run(Thread.java:829)
2023-01-16 16:44:50,231 INFO  [cloud.consoleproxy.ConsoleProxyResourceHandler] (Thread-162:null) Get resource request for /resource/noVNC/app/images/error.svg
2023-01-16 16:44:50,236 INFO  [cloud.consoleproxy.ConsoleProxyResourceHandler] (Thread-162:null) file does not exist/resource/noVNC/app/images/error.svg
````

File: engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java (hunk 1 above)
Comment (weizhouapache) [reply engine/schema/src/main/java/com/cloud/vm/AllowedConsoleSessionVo.java:32]: > @DaanHoogland @nvazquez @rohityadavcloud
> 
> I forgot to mention it in the description. My idea while implementing this way was to extend it in the future to store more data (when/which user generated the console session, when/which IP consumed it, and so on - indeed, `console_session` would be a better name), to allow operators to track the console sessions; And the entries would be marked as removed/used instead of being deleted.
> 
> As gathering those data would require a more accurate analysis of the console session workflow, I first implemented this way to handle the situation; then I would improve the feature.
> 
> What do you think?

@GutoVeronezi 
make sense. can you please add created/removed columns and VOs ? it seems simple.
please also use `remove` instead of `expunge` when you remove a session.

File: server/src/main/java/org/apache/cloudstack/consoleproxy/ConsoleAccessManagerImpl.java [hunk 4]
```diff
@@ -151,11 +156,16 @@ public boolean isSessionAllowed(String sessionUuid) {
```
nvazquez, line 156: @GutoVeronezi the problem listed previously for one mgmt server comes from the method `isSessionAllowed` that does not query the database for the session UUID and always returns false, therefore the `ConsoleAccessAuthenticationAnswer` is false

nvazquez (changes requested):

server/src/main/java/org/apache/cloudstack/consoleproxy/ConsoleAccessManagerImpl.java
GutoVeronezi: @nvazquez, thanks. I missed a line while porting the changes to the PR. I fixed it with commit dec6e3fb461763605ca6be34013d7e2c2438827a and was doing some more testing before pinging you.

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -1036,9 +1036,9 @@ INSERT INTO `cloud`.`role_permissions` (`uuid`, `role_id`, `rule`, `permission`)
 SELECT UUID(), `roles`.`id`, 'isAccountAllowedToCreateOfferingsWithTags', 'ALLOW'
 FROM `cloud`.`roles` WHERE `role_type` = 'DomainAdmin';
 
---- Create table for handling allowed console session temporally
+--- Create table for handling allowed console session temporally #7094
 
-CREATE TABLE `cloud`.`allowed_console_session` (
+CREATE TABLE IF NOT EXISTS `cloud`.`console_session` (
     `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
```
rohityadavcloud, line 1042: Is this VM ID? If yes, this should be linked to vm_instance table and to be deleted (on cascade...). @GutoVeronezi

GutoVeronezi:
@DaanHoogland @nvazquez @rohityadavcloud @weizhouapache 

I redesigned the table `console_session` to store more data, based on the idea in https://github.com/apache/cloudstack/pull/7094#discussion_r1071235407. When generated via UI (ids 1, 2, and 3), the session is set as removed right away, as the UI redirects to the URL and consumes the session. When generated via API directly (id 4) , the session will not be set as removed until be consumed:

image

During the tests I did not notice a workflow to expire not used sessions. Will they be valid "forever" if not used? Or am I missing something? cc: @nvazquez 

---

Regarding the `on cascade` clause, as the referenced entities are never deleted by ACS (they are only marked as removed), I do not see a reason to use it.

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
GutoVeronezi: @weizhouapache, the table was redesigned, vide https://github.com/apache/cloudstack/pull/7094#issuecomment-1385309721.

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
GutoVeronezi: @weizhouapache, the table was redesigned, vide https://github.com/apache/cloudstack/pull/7094#issuecomment-1385309721.

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
GutoVeronezi: @rohityadavcloud, the table was redesigned, vide https://github.com/apache/cloudstack/pull/7094#issuecomment-1385309721.

nvazquez:
@GutoVeronezi thanks for the refactor, you are right I have now checked that the lastUsedTime on the ConsoleProxyLoadReportCommand keeps being updated even when the session is not used, that makes the session not to timeout

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -1035,3 +1035,16 @@ WHERE   role_id = (SELECT id FROM `cloud`.`roles` WHERE name = 'Read-Only User -
...
 SELECT UUID(), `roles`.`id`, 'isAccountAllowedToCreateOfferingsWithTags', 'ALLOW'
 FROM `cloud`.`roles` WHERE `role_type` = 'DomainAdmin';
+
+--- Create table for handling allowed console session temporally #7094
+
+CREATE TABLE IF NOT EXISTS `cloud`.`console_session` (
+    `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT PRIMARY KEY,
+    `uuid` varchar(40) NOT NULL COMMENT 'UUID generated for the session',
+    `created` datetime NOT NULL COMMENT 'When the session was created',
+    `account_id` bigint(20) unsigned NOT NULL COMMENT 'Account who generated the session' references cloud.account(id),
```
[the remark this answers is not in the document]
GutoVeronezi: @weizhouapache, I found out the reason:

image

https://mariadb.com/kb/en/foreign-keys/

The syntax I firstly used would do nothing (neither warn the user :man_shrugging:) for environments with MariaDB version previous to 10.5.  I changed it to work with the previous syntax (https://github.com/apache/cloudstack/pull/7094/commits/6173aa0896e42df81fcae5e4c806c9f56ad54b08), thanks.

server/src/main/java/org/apache/cloudstack/consoleproxy/ConsoleAccessManagerImpl.java
```diff
@@ -303,6 +310,16 @@ private ConsoleEndpoint composeConsoleAccessEndpoint(String rootUrl, VirtualMach
         return consoleEndpoint;
     }
 
+    protected void persistConsoleSession(String sessionUuid, long instanceId, long hostId) {
+        ConsoleSessionVO consoleSessionVo = new ConsoleSessionVO();
+        consoleSessionVo.setUuid(sessionUuid);
```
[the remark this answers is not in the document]
GutoVeronezi: On principle, I would rather not create a constructor for informing the object properties, as the more properties it has, the longer the constructor would get, becoming harder to read/interpret the code. For Python, that supports keyword arguments, this approach works fine, though.

rohityadavcloud (approved):
Lgtm, the license header has extra newlines compared to existing files. Didn’t test it.

rohityadavcloud:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

DaanHoogland (approved):
clgtm

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan test rocky8 vmware-67u3

nvazquez (approved):
LGTM manually tested on KVM Rocky 8 env with 2 management servers

mlsorensen:
Nice job @GutoVeronezi - I saw this issue as well earlier.   Regarding the `console_session` table, do you see any concern with never cleaning these up and only marking them removed? There could be a lot of rows. Perhaps there should be a cleanup introduced, similar to the events and alerts tables?

GutoVeronezi:
@mlsorensen, as this data can be used for auditing access to the consoles, I do not see a problem with keeping them; however, the data may become obsolete or operators may not want to store them. Your proposal for a cleanup task is a good idea. We can implement it with customized intervals and retention, just like other tasks ACS has.

@DaanHoogland, do we still have time to implement this feature in 4.18 or it is already in freeze?

nvazquez:
@DaanHoogland @GutoVeronezi I've created the PR: https://github.com/apache/cloudstack/pull/7132, can you please review?

mlsorensen:
> @mlsorensen, as this data can be used for auditing access to the consoles, I do not see a problem with keeping them; however, the data may become obsolete or operators may not want to store them. Your proposal for a cleanup task is a good idea. We can implement it with customized intervals and retention, just like other tasks ACS has.
> 
> @DaanHoogland, do we still have time to implement this feature in 4.18 or it is already in freeze?

Yes, I agree it could be useful for audit. I am mostly thinking about size, a decent cloudstack install could have thousands of console requests a day.  I think mysql could handle it, but it is like a log that never rolls.  Probably not an immediate issue.

PR #7095, opened 2023-01-13
fix guestOsMapper and move mapping code to latest upgrade

DaanHoogland:
### Description

This PR fixes a buh in the guestos mapping facility and moves mappings, so current installations get the right mappings.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

upgraded from 4.16.0 and checked if the mappings where alright.
upgraded from 4.17 checked that mapping where alright and for the ubuntu on xenserver that an existing VM could still be stopped/started and migrated.


weizhouapache:
it seems some codes in Upgrade41400to41500 needs to be moved as well. 

```
$ grep -rl addGuestOsAndHypervisorMappings
engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41400to41500.java
engine/schema/src/main/java/com/cloud/upgrade/GuestOsMapper.java

$ grep -rl updateGuestOsNameFromMapping
engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
engine/schema/src/main/java/com/cloud/upgrade/GuestOsMapper.java
```

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
```diff
@@ -250,130 +249,6 @@ private void correctGuestOsNames(final Connection conn) {
         guestOsMapper.updateGuestOsNameFromMapping("OpenSUSE Linux (32 bit)", mapping);
     }
```
DaanHoogland, line 250: @weizhouapache , your comment in https://github.com/apache/cloudstack/pull/7095#issuecomment-1382247322 means that also this method must be moved? It contains a call to `updateGuestOsNameFromMapping()` as well.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
weizhouapache: double-checked just now.

both `addGuestOsAndHypervisorMappings` and `updateGuestOsIdInHypervisorMapping` call the method `getGuestOsId`, they need to be moved.

`updateGuestOsNameFromMapping` is correct. sorry for my mistake

DaanHoogland:
@blueorangutan test

nvazquez (dismissed):
LGTM, haven't tested it

engine/schema/src/main/java/com/cloud/upgrade/dao/Upgrade41500to41510.java
```diff
@@ -69,8 +66,6 @@ public InputStream[] getPrepareScripts() {
 
     @Override
     public void performDataMigration(Connection conn) {
-        correctGuestOsNames(conn);
-        updateGuestOsMappings(conn);
```
nvazquez, line 73: I'm not sure about these schema changes, what would happen to templates/VMs created from existing guest OS on previous versions?

weizhouapache (approved):
code lgtm

we need to verify the guest os mapping in environments below
(1) fresh 4.18
(2) upgrade from 4.17 to 4.18

I will verify when packages are done.

weizhouapache:
@blueorangutan package

harikrishna-patnala (approved):
Code LGTM.

weizhouapache:
> LGTM @DaanHoogland would we need to do that with every release?

if we will do another 4.17 minor release (4.17.3.0), we need to fix in 4.17 as well.

DaanHoogland:
@weizhouapache should I rebase on 4.17?

DaanHoogland:
> LGTM @DaanHoogland would we need to do that with every release?

Basically, no. Only because the mapping code was broken between 4.16.1 and 4.18 we need to redo the mapping in 4.18

weizhouapache:
> @weizhouapache should I rebase on 4.17?

it will be complicated ...

you need to 
(1) add upgrade path 4.17.2.0->4.17.3.0
(2) move everything to Upgrade41720to41730.java

and
what if there is not 4.17.3.0 release ? 
what if 4.17.3.0 is released after 4.18.0.0, can upgrade from 4.17.3.0 to 4.18.0.0 ?
we may need to keep the codes in both files: Upgrade41720to41730.java and Upgrade41720to41800.java

DaanHoogland:
> > @weizhouapache should I rebase on 4.17?
> 
> @DaanHoogland 
> 
> it will be complicated ...
> 
> you need to 
> (1) add upgrade path 4.17.2.0->4.17.3.0
> (2) move everything to Upgrade41720to41730.java
> 
> and
> what if there is not 4.17.3.0 release ? 
> what if 4.17.3.0 is released after 4.18.0.0, can upgrade from 4.17.3.0 to 4.18.0.0 ?
> we may need to keep the codes in both files: Upgrade41720to41730.java and Upgrade41720to41800.java

And merge forward guaranteed conflicts. Let's leave it as is.

vladimirpetrov:
@blueorangutan package

vladimirpetrov (approved):
LGTM based on manual testing.

BryanMLima:
Hey @DaanHoogland, I know it has been a while, but did you check if the Guest OS types were duplicate after this change? I have some environments that were recently updated to the `4.18` from the `4.16` and there are 43 duplicate Guest OS types.

weizhouapache:
@BryanMLima 
yes, I have noticed it. 
The method `addGuestOsAndHypervisorMappings` used in this PR will add a new guest OS (even it already exists) and then add guest os mappings.
I have created an issue to track it: #7783

PR #7111, opened 2023-01-18
marvin: install mysql-connector-python version 8.0.31

rohityadavcloud:
On EL7 and EL8, when Marvin is installed and if newer mysql-connector-python is installed leads to errors which aren't compatible with python 2.7 and 3.6.

### DIFF
--- a/tools/marvin/setup.py
+++ b/tools/marvin/setup.py
@@ -46,7 +46,7 @@ setup(name="Marvin",
                 "marvin.sandbox.basic"],
       license="LICENSE.txt",
       install_requires=[
-          "mysql-connector-python >= 1.1.6",
+          "mysql-connector-python == 8.0.20",
           "requests >= 2.2.1",
           "paramiko >= 1.13.0",
           "nose >= 1.3.3",
--- a/tools/marvin/setup.py
+++ b/tools/marvin/setup.py
@@ -46,7 +46,7 @@ setup(name="Marvin",
                 "marvin.sandbox.basic"],
       license="LICENSE.txt",
       install_requires=[
-          "mysql-connector-python == 8.0.20",
+          "mysql-connector-python == 8.0.31",
           "requests >= 2.2.1",
           "paramiko >= 1.13.0",
           "nose >= 1.3.3",
--- a/tools/marvin/setup.py
+++ b/tools/marvin/setup.py
@@ -46,7 +46,7 @@ setup(name="Marvin",
                 "marvin.sandbox.basic"],
       license="LICENSE.txt",
       install_requires=[
-          "mysql-connector-python == 8.0.30",
+          "mysql-connector-python <= 8.0.30",
           "requests >= 2.2.1",
           "paramiko >= 1.13.0",
           "nose >= 1.3.3",
--- a/tools/marvin/setup.py
+++ b/tools/marvin/setup.py
@@ -46,7 +46,7 @@ setup(name="Marvin",
                 "marvin.sandbox.basic"],
       license="LICENSE.txt",
       install_requires=[
-          "mysql-connector-python == 8.0.31",
+          "mysql-connector-python == 8.0.30",
           "requests >= 2.2.1",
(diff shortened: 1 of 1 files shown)


rohityadavcloud:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@weizhouapache I can bump to .29 as .30 which is latest breaks Marvin install and cloudstack-setup-management  on el7, el8. .20 what was tried and tested and included in the cloudstack-management pkg as a wheel

weizhouapache (approved):
code lgtm

weizhouapache:
looks good.
```
[root@pr7111-t5901-kvm-centos7-marvin ~]# pip3 list |grep mysql
mysql-connector-python 8.0.20
[root@pr7111-t5901-kvm-centos7-marvin ~]# pip list |grep mysql
mysql-connector-python 8.0.20
```

tools/marvin/setup.py
```diff
@@ -46,7 +46,7 @@
                 "marvin.sandbox.basic"],
       license="LICENSE.txt",
       install_requires=[
-          "mysql-connector-python >= 1.1.6",
+          "mysql-connector-python == 8.0.20",
```
DaanHoogland, line 49: dare we use >= here?

DaanHoogland: quick stackoverflow search: `"mysql-connector-python>=8.0.20,<8.0.30"` should work

rohityadavcloud: Yup that may work as well; but we might be specific like we are with the pom.xml. As with EL8, cpython version is 3.6 we want to use a dependency that is certified for cpython3.6 ; see https://pypi.org/project/mysql-connector-python/8.0.20/#files supports both cp3.6 and cp2.7. Happy to include your workaround as well. I think 8.0.30 and 8.0.31 might work, it's the new 8.0.32 that breaks. I see 8.0.31 used here https://github.com/apache/cloudstack/blob/main/packaging/centos8/cloud.spec#L309 (we bundle these for cloudstack-management)

weizhouapache: the version might be not important. we just need a version which works.

by the way, is there EOL date of the versions ?

DaanHoogland: let's pick the newest that works

rohityadavcloud: done

rohityadavcloud:
@blueorangutan package

DaanHoogland:
@blueorangutan test

rohityadavcloud:
@blueorangutan test

tools/marvin/setup.py
```diff
@@ -46,7 +46,7 @@
                 "marvin.sandbox.basic"],
       license="LICENSE.txt",
       install_requires=[
-          "mysql-connector-python >= 1.1.6",
+          "mysql-connector-python == 8.0.31",
```
rohityadavcloud, line 49:
```suggestion
          "mysql-connector-python == 8.0.30",
```

tools/marvin/setup.py
```diff
@@ -46,7 +46,7 @@
                 "marvin.sandbox.basic"],
       license="LICENSE.txt",
       install_requires=[
-          "mysql-connector-python >= 1.1.6",
+          "mysql-connector-python == 8.0.30",
```
rohityadavcloud, line 49:
```suggestion
          "mysql-connector-python <= 8.0.30",
```

rohityadavcloud:
@blueorangutan package

weizhouapache:
@blueorangutan test

DaanHoogland (approved):
code looks cood regression tests on this will certainly suffice, though I think gha already proof it...?

rohityadavcloud:
Yes I hope so @DaanHoogland 
Let’s wait for smoketests result to be sure.

weizhouapache (approved):
code lgtm

smoke tests look ok

PR #7120, opened 2023-01-23
marvin: newer python setuptools doesn't like -SNAPSHOT in marvin version

rohityadavcloud:
### Description

Newer python setuptools doesn't like -SNAPSHOT in marvin version. This has been reported such as https://github.com/pypa/setuptools/issues/3772

The fix is to either downgrade or fix the version string to use something that setuptools likes.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested on mbx. Found that when setuptools is upgraded to v66+, this breaks Marvin installation. Refer: https://github.com/pypa/setuptools/issues/3772

### DIFF
--- a/tools/marvin/mvn-setup.py
+++ b/tools/marvin/mvn-setup.py
@@ -48,7 +48,7 @@ def runSetupScript(args):
 
 
 if __name__ == "__main__":
-    version = sys.argv[1]
+    version = sys.argv[1].replace("-SNAPSHOT", "")
     remainingArgs = sys.argv[2:]
     replaceVersion(setupScript, version)
     runSetupScript(remainingArgs)


rohityadavcloud:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

DaanHoogland (approved):
clgtm
is there any postfix that would be accepted that we could replace it with (instead of an empty string)?

rohityadavcloud:
I think the semver will cause issue so we can’t have snapshot in the version name. @DaanHoogland

DaanHoogland:
I realize that, @rohityadavcloud . So can we replace it with ".a" or ".9"?

astrojuanlu:
Check out PEP 440 https://peps.python.org/pep-0440/ I think you could use `.snapshot0` (please verify).

weizhouapache (approved):
code lgtm

rohityadavcloud:
Thanks @astrojuanlu - given the marvin library's use is primarily for testing only. This isn't used in prod and all we're doing is to rename the python/pip pkg. The rpm/deb file will continue to have -SNAPSHOT so won't be a problem from pkg management perspective.

rohityadavcloud:
I think this is good to merge @DaanHoogland - pl review and do the honours.

PR #7153, opened 2023-01-31
Introducing concept of domain VPCs

GaOrtiga:
In ACS the creation of tiers of a VPC is restricted to the same account that owns the VPC; therefore, each account needs to have its own VPC and it is not possible to group tiers owned by different accounts in the same VPC. These tiers cannot share the same VR that is used to implement the VPC, for instance.

However, in private cloud scenarios to reduce the number of VRs, it might be interesting to have VPCs, where its tiers are owned by different accounts; thus, they (the accounts) share the same VR/VPC, but each one has their own broadcast domain and features implemented by the VPC, such as DHCP, NAT, and so on.

To address this situation, the concept of Domain VPCs has been created (only available on the API so far), where a VPC can be managed by a domain and its tiers can be created to accounts inside the domain. In the `createNetwork` API it will be possible to create networks (tiers) to a VPC from a different account; however, the target account must be accessible for the account that owns the VPC. The tiers will be isolated from the broadcast domain and will consume the same VR, in accordance with the current behavior. Also, if a VPN is setup in the VPC, the user will have access to all networks, in accordance with the current behavior.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
I created VPCs with Root Admin, Domain Admin, and user accounts and tried creating tiers in these VPCs for the other accounts (should only be able to create if the VPC account has access to the account that owns the network). 
| # | VPC owner account | Network Owner account | Same Domain | Could Create | Expected Result
| ------ | ------ | ------ | ------ |------ | ------ |
| 1 | Root Admin | Any | Any | Y | Y |
| 2 | Domain Admin |Any | N | N| Y |
| 3 | Domain Admin | Any | Y  | Y | Y |
| 4 | User | Same User | Y | Y | Y |
| 5 | User | Any Other | Any | N |Y |

I also ran some basic tests like deploying VMs in the created networks and checking that the networks were working properly.


server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -2490,7 +2490,11 @@ public NetworkResponse createNetworkResponse(ResponseView view, Network network)
         if (network.getVpcId() != null) {
             Vpc vpc = ApiDBUtils.findVpcById(network.getVpcId());
             if (vpc != null) {
-                response.setVpcId(vpc.getUuid());
+                try {
+                    _accountMgr.checkAccess(CallContext.current().getCallingAccount(), null, false, vpc);
+                    response.setVpcId(vpc.getUuid());
+                } catch (PermissionDeniedException e){
+                }
```
DaanHoogland, line 2497: this is strange, we check on access but there is no consequence?
at least we should log that something was denied/not setting or disclosing vpcid.

DaanHoogland (commented):
looks good, I recommend extensive manual testing as un foreseen consequences may come from this concept. The change is simple enough and is great if it turns out it works out of the box.

server/src/main/java/com/cloud/api/ApiResponseHelper.java
GaOrtiga: @DaanHoogland , the reason to do this is to abstract the ID of the VPC from the account that does not have access to it; this way, the account would know the network belongs to a VPC, but would not be able to redirect to the VPC tab in the UI. Here follows an example, with VPCUser being accessible by the account and VPCRoot not being accessible:

image


You are right, there should be a log there, I will add it.

server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -1476,7 +1476,8 @@ public IPAddressVO associateIPToGuestNetwork(long ipId, long networkId, boolean
         //     - if shared network in Advanced zone
         //     - and it belongs to the system
         if (network.getAccountId() != owner.getId()) {
-            if (zone.getNetworkType() != NetworkType.Basic && !(zone.getNetworkType() == NetworkType.Advanced && network.getGuestType() == Network.GuestType.Shared)) {
+            if (zone.getNetworkType() != NetworkType.Basic &&
+                    !(zone.getNetworkType() == NetworkType.Advanced && network.getGuestType() == Network.GuestType.Shared || network.getVpcId() == ipToAssoc.getVpcId()))  {
```
DaanHoogland, line 1480: please look at the bug sonar reported, I think this one is serious.

server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -2534,7 +2534,12 @@ public NetworkResponse createNetworkResponse(ResponseView view, Network network)
         if (network.getVpcId() != null) {
             Vpc vpc = ApiDBUtils.findVpcById(network.getVpcId());
             if (vpc != null) {
-                response.setVpcId(vpc.getUuid());
+                try {
+                    _accountMgr.checkAccess(CallContext.current().getCallingAccount(), null, false, vpc);
+                    response.setVpcId(vpc.getUuid());
+                } catch (PermissionDeniedException e){
+                    s_logger.debug("Not setting the vpcId to the response because the caller does not have access to the VPC");
+                }
```
DaanHoogland, line 2542: Can you extract this as a separate method?

server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
```diff
@@ -1476,7 +1476,8 @@ public IPAddressVO associateIPToGuestNetwork(long ipId, long networkId, boolean
         //     - if shared network in Advanced zone
         //     - and it belongs to the system
         if (network.getAccountId() != owner.getId()) {
-            if (zone.getNetworkType() != NetworkType.Basic && !(zone.getNetworkType() == NetworkType.Advanced && network.getGuestType() == Network.GuestType.Shared)) {
+            if (zone.getNetworkType() != NetworkType.Basic &&
+                    !(zone.getNetworkType() == NetworkType.Advanced && network.getGuestType() == Network.GuestType.Shared || network.getVpcId().equals(ipToAssoc.getVpcId())))  {
```
DaanHoogland, line 1480: So if `network.getVpcId().equals(ipToAssoc.getVpcId())` we should throw an exception! Can you explain?

server/src/main/java/com/cloud/network/vpc/VpcManagerImpl.java
```diff
@@ -1764,9 +1764,14 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
...
-                    if (vpc.getAccountId() != networkOwner.getId()) {
-                        throw new InvalidParameterValueException("Vpc " + vpc + " owner is different from the network owner " + networkOwner);
+                    // 4) Vpc's account should be able to access network owner's account
+                    Account vpcaccount = _accountMgr.getAccount(vpc.getAccountId());
+                    try {
+                        _accountMgr.checkAccess(vpcaccount, null, false, networkOwner);
+                    }
+                    catch (PermissionDeniedException e) {
+                        s_logger.error(e.getMessage());
+                        throw new InvalidParameterValueException(String.format("VPC owner does not have access to account [%s].", networkOwner.getAccountName()));
```
DaanHoogland, line 1774: can you extract this as a method?

server/src/main/java/com/cloud/network/IpAddressManagerImpl.java
GaOrtiga: Theres a subtle "!" in the beginning of the expression
`(zone.getNetworkType() == NetworkType.Advanced && network.getGuestType() == Network.GuestType.Shared || network.getVpcId().equals(ipToAssoc.getVpcId()))`
which makes it equivalent to the following expression
`(!zone.getNetworkType() == NetworkType.Advanced && !network.getVpcId().equals(ipToAssoc.getVpcId()) || !network.getGuestType() == Network.GuestType.Shared && !network.getVpcId().equals(ipToAssoc.getVpcId()))`

So we are actually checking if both VpcIds are different and throwing an exception if that is the case.

DaanHoogland: right, missed that. Maybe you can extract the expression to a method for clarity?
thanks for the explanation.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

kiranchavala (commented):
@GaOrtiga 

what about the scenario where the network account owner is a user account in the same domain?

Ideally the network should get created since the user is present in the same domain 

|   | VPC owner account | Network Owner account | Same Domain | Could Create | Expected Result | Comments                                                                                                                                                                          |
|---|-------------------|-----------------------|-------------|--------------|-----------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|   | Domain Admin      | User (role)           | Y           | N            | the network should get created      | 🙈 Error:   (HTTP 531, error code 4365) Account   [{"accountName":"<>r","id":6,"uuid":"9d9daf47-04be-4488-af71-281e255fedfe"}]   does not have permission to operate with resource |
|   |                   |                       |             |              |                 |                                                                                                                                                                                   |
|   |                   |                       |             |              |                 |                                                                                                                                                                                   |

kiranchavala (commented):
@GaOrtiga 

Another scenario where sub domains are present 

Also, what will happen if an account is present under a sub domains of a domain 



Ideally, the network should get created since the subdomain is also falls under the  same domain 

|   | VPC owner account | Network Owner account | Same Domain | Could Create | Expected Result | Comments                                                                                                                                                                          |
|---|-------------------|-----------------------|-------------|--------------|-----------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
|   | Domain Admin      | Account under a subdomain          | Y           | N            | Network tier should get created        | 🙈 Error: (HTTP 531, error code 4365) Account <>does not have permission to operate within domain id=51427f97-6c38-4a7d-bf1c-8cdc86232cb8 |
|   |                   |                       |             |              |                 |                                                                                                                                                                                   |
|   |                   |                       |             |              |                 |                                                                                                                                                                                   |

DaanHoogland:
@blueorangutan test keepEnv

GaOrtiga:
> @GaOrtiga
> 
> Another scenario where sub domains are present
> 
> Also, what will happen if an account is present under a sub domains of a domain
> 
> Ideally, the network should get created since the subdomain is also falls under the same domain
> 	VPC owner account 	Network Owner account 	Same Domain 	Could Create 	Expected Result 	Comments
> 	Domain Admin 	Account under a subdomain 	Y 	N 	Network tier should get created 	see_no_evil Error: (HTTP 531, error code 4365) Account <>does not have permission to operate within domain id=51427f97-6c38-4a7d-bf1c-8cdc86232cb8

hi @kiranchavala thanks for testing.

Which account did you use to run the API commands?

I ran both tests twice, once using the VPC owner and the other using the network owner. While running it with the VPC owner, both creations were successful. When running it with network owner I had the same results as you, however I do believe this is the correct outcome, since these accounts should not be able to access the the domain admin's resources such as the VPC.

kiranchavala:
Hi @GaOrtiga 

Thanks for the update

I used the network owner account to execute the api calls


Also Let me do a quick check with the VPC owner account

DaanHoogland:
@blueorangutan test keepEnv

kiranchavala (commented):
@GaOrtiga 

I have tested manually the combinations specified with the table with the  VPC owner account credentials and they work fine 

| # | VPC owner account | Network Owner account | Same Domain | Could Create | Expected Result |
|---|---|---|---|---|---|
| 1 | Root Admin | Any | Any | Y | Y |
| 2 | Domain Admin | Any | N | N | Y |
| 3 | Domain Admin | Any | Y | Y | Y |
| 4 | User | Same User | Y | Y | Y |
| 5 | User | Any Other | Any | N | Y |

@DaanHoogland 

Also when i was site to site vpn connectivity i was wondering if the feature can be extended a bit more 

Currently in order to create a site to site vpn connection, It's only possible between 2 vpc  within the same account in a domain 

https://cloudstack.apache.org/api/apidocs-4.18/apis/createVpnConnection.html



Using the concept of domain vpc,  can the feature be extended to 

1. To Create a site to site vpn connection between 2 vpc in  different account but in the same domain 

2. To Create a site to site vpn connection beween 2 vpc's present in 2 separate domains




Currently the following exception is thrown when we try to establish site to site vpn between two different accounts in the same domain


`2023-07-27 07:55:55,257 INFO  [a.c.c.a.ApiServlet] (qtp1239807799-19:ctx-a1619bcc ctx-71562f34) (logid:2fdc4dd6) (userId=2 accountId=2 sessionId=node01098v0azatkeyry62st73mrt54) 10.0.3.251 -- GET s2svpngatewayid=03df5137-17f7-4b73-adfa-02710e811f96&s2scustomergatewayid=f823f7ea-e998-4a34-b05d-fe1cbef768c9&passive=false&command=createVpnConnection&response=json 431 VPN connection can only be esitablished between same account's VPN gateway and customer gateway!`

kiranchavala (approved):
LGTM,  tested manually the combinations specified with the table with the  VPC owner account credentials and they work fine 

| # | VPC owner account | Network Owner account | Same Domain | Could Create | Expected Result |
|---|---|---|---|---|---|
| 1 | Root Admin | Any | Any | Y | Y |
| 2 | Domain Admin | Any | N | N | Y |
| 3 | Domain Admin | Any | Y | Y | Y |
| 4 | User | Same User | Y | Y | Y |
| 5 | User | Any Other | Any | N | Y |


Also tested by launching vm in different tiers belonging to different account and the communication between them was working fine

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland (commented):
clgtm

GaOrtiga:
doc PR:
https://github.com/apache/cloudstack-documentation/pull/410

PR #7157, opened 2023-02-01
ui: fix build issue on arm/Mac

rohityadavcloud:
- Fixes UI build issue on M1/Mac

This is only useful if we're sure that node-sass isn't used by UI during development. The node-sass isn't a build dependency but only dev dependency.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

- Run npm install and npm run serve which got no error
- Based on https://www.priestch.com/replace-node-sass-with-dart-sass-in-vue-cli3-based-project/ looks like we're not using sass at all in our vue.config or other places
- Not a sass/css expert but seems this doesn't cause any regression and most likely an unnecessary dependency


rohityadavcloud:
@blueorangutan ui

rohityadavcloud:
Hi @svenvogel we may need a UI/css/sass expert to advise, based on my testing this doesn't cause any regression. Could you or any of your colleagues help advise as well? Thanks.

ui/package.json
```diff
@@ -89,7 +89,6 @@
     "eslint-plugin-vue": "^7.0.0",
     "less": "^3.0.4",
     "less-loader": "^5.0.0",
-    "node-sass": "^4.0.0",
     "sass": "^1.49.9",
```
rohityadavcloud, line 92: We're using the dart-sass already (see https://github.com/sass/node-sass#node-sass)

shwstppr (approved):
LGTM. node-sass has been deprecated and `sass` which is already used seems more popular

DaanHoogland (approved):

DaanHoogland:
the qa version of this does not allow login :(

nvazquez (approved):
LGTM - I use M1 as well but was able to run the UI without this change. Have tested this PR and also works

harikrishna-patnala (approved):
LGTM, Tested.

Got the same issue before the fix and it is working fine after applying the fix. Thanks @rohityadavcloud

weizhouapache:
code lgtm

tested on ubuntu 22, it is not broken :smile:

rohityadavcloud:
@DaanHoogland works for me, could be caching issue - can you try in incognito mode?

DaanHoogland:
> @DaanHoogland works for me, could be caching issue - can you try in incognito mode?

good call @rohityadavcloud , login works

PR #7210, opened 2023-02-13
Set number of hours per month with actual number in Quota

stephankruggg:
### Description

The Quota usage value for monthly usage types uses the cost per hour to calculate the total tariff value in a particular period. In order to define the cost per hour, it takes into account the total number of hours in each month. However, this value is currently fixed at 720 hours for every month. 

This causes incorrect values when calculating the cost per hour, resulting in costs being higher than expected. Therefore, this PR aims to calculate the number of hours in each month based on the actual number of days of each month.

For example, if the tariff value is 100 per month, the current cost per hour is 0.138888889 for every month. For 1 hour per month the total cost in a year would be 1.666666668.
With the adjustments to the cost per hour of this PR, the costs per hour would be: 0.134408602 (January, March, May, July, August, October, December), 0.148809524 (February of a non-leap year), 0.143678161 (February of a leap year), and 0.138888889 (April, June, September, November). This results in 1.645225294 for a non-leap year and 1.640093931 for a leap year at 1 hour per month, which is a difference of about 1.3% in costs to the current method.

### Types of changes

- [ ] Major
- [X] Minor

### How Has This Been Tested?

In a local lab I applied the changes, and manually changed the current month of the MS to every month in a year (including February of a leap year) and verified if the cost per hour was calculated correctly using the number of hours of each month when the `quotaUpdate` API was called.

### DIFF
--- a/framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
+++ b/framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
@@ -18,6 +18,9 @@ package org.apache.cloudstack.quota;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
+import java.time.LocalDate;
+import java.time.Month;
+import java.time.YearMonth;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Date;
@@ -89,10 +92,10 @@ public class QuotaManagerImpl extends ManagerBase implements QuotaManager {
 
     private TimeZone _usageTimezone;
     private int _aggregationDuration = 0;
-
-    static final BigDecimal s_hoursInMonth = BigDecimal.valueOf(DateUtil.HOURS_IN_A_MONTH);
     static final BigDecimal GiB_DECIMAL = BigDecimal.valueOf(ByteScaleUtils.GiB);
     List<Account.Type> lockablesAccountTypes = Arrays.asList(Account.Type.NORMAL, Account.Type.DOMAIN_ADMIN);
+
+    static BigDecimal hoursInCurrentMonth;
 
     public QuotaManagerImpl() {
         super();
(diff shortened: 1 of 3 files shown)


framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
```diff
@@ -558,6 +563,14 @@ protected BigDecimal getUsageValueAccordingToUsageUnitType(UsageVO usageRecord,
         }
     }
 
+    protected void setHoursInCurrentMonth() {
+        LocalDate currentDate = LocalDate.now();
```
DaanHoogland, line 567: @stephankruggg is it possible that this calculation happens for just a minute ago/last month on the 1st of the next month. How will this be handled?

JoaoJandre (approved):
CLTM, didn't test it

shwstppr:
@stephankruggg cc @GaOrtiga can you please check the outstanding review comments?

DaanHoogland:
@stephankruggg can you answer my question https://github.com/apache/cloudstack/pull/7210#discussion_r1105876041 ?
Any daylight saving or leap year/ - second issues possible there?

rohityadavcloud (approved):

framework/quota/src/main/java/org/apache/cloudstack/quota/QuotaManagerImpl.java
GaOrtiga: @DaanHoogland, sorry for the delay. 
I have implemented some changes to address the issues you pointed out by getting the date directly from each record.

DaanHoogland:
@blueorangutan package

DaanHoogland:
ui build error is in codecov push, not in code.

DaanHoogland:
@blueorangutan test

DaanHoogland (approved):
clgtm

PR #7229, opened 2023-02-14
[Usage] Fix wrong usage_type

SadiJr:
### Description

IO write events have been saved in the database using the `usage_type` of the `VM_DISK_BYTES_WRITE` type, which causes an inconsistency in the database. This behavior has been fixed by this PR. Plus, the old data saved in the database has been normalized.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?
It was tested in a local lab:

1. I waited for the job of the Usage plugin to run;
2. Before, ACS saved the IO write event with the wrong usage_type in the database;
3. Now, ACS saves these events with the right usage_type, and the old events are normalized.


engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -1569,4 +1569,12 @@ CREATE VIEW `cloud`.`user_view` AS
...
-            and async_job.job_status = 0;
\ No newline at end of file
+            and async_job.job_status = 0;
+
+-- Change usage of VM_DISK_IO_WRITE to use right usage_type
+UPDATE
+  `cloud_usage`.`cloud_usage`
+SET
+  usage_type = 22
+WHERE
```
weizhouapache, line 1579: This might take very long time if usage database is huge. (It is quite normal if usage server has heen running for some years).

Might it be helpful if use the following ?

WHERE usage_type = <type of UsageTypes.VM_DISK_BYTES_WRITE> AND usage_display like '% io write';

SadiJr: @weizhouapache done, thanks for your suggestion.

stephankruggg (approved):
CLGTM, not manually tested

DaanHoogland (approved):
clgtm

weizhouapache (approved):
code lgtm

BryanMLima (approved):
LGTM, tested this in a local lab. I upgraded from a version with this error and the old events were fixed; furthermore, the new usage events were with the correct usage type in the DB.

engine/schema/src/main/resources/META-INF/db/schema-41720to41800.sql
```diff
@@ -1574,3 +1574,12 @@ CREATE VIEW `cloud`.`user_view` AS
...
 WHERE store_role = "Primary" AND store_id IN (SELECT id FROM storage_pool WHERE removed IS NOT NULL);
+
+
+-- Change usage of VM_DISK_IO_WRITE to use right usage_type
+UPDATE
+  `cloud_usage`.`cloud_usage`
+SET
+  usage_type = 22
+WHERE
+  usage_type = 24 AND usage_display like '% io write';
```
BryanMLima, line 1585: As the milestone changed to version `4.18.1.0` this needs to be changed to the correct schema file.

weizhouapache: @BryanMLima 
I think this can be merged into 4.18.0.0

PR #7242, opened 2023-02-16
Don't allow users to remove their account or user.

soreana:
### Description

Cloudstack users can delete their account accidentally and lock themselves from accessing the CloudStack panel. A link to an account placed next to ISO, template, VM and other entities in CloudStack's UI. If customers doesn't paying attention, they would click on the account link instead of the link to the entity they want (vm, template, ISO etc.) Then, if they had wanted to delete that entity, they would press delete button without realising they were on the account page, and press Confirm.

Mailing list [[link]](https://lists.apache.org/thread/0787mmvfo32tqgx63jqc58ko1zfzbs90)

This PR don't allow users to remove their account or their user.

### Types of changes

- [ ] Major
- [x] Minor

https://user-images.githubusercontent.com/9499410/219402523-e878deb1-23f9-400b-98ef-113c447dec88.mp4

### How Has This Been Tested?

1. Create two account, `Acc1` and `Acc2`.
2. Login as the `Acc1`
3. Go to the `Acc1` page, delete button is `not` there.
4. Go to the `Acc2` page, delete button is there.
5. Logout and login as the `Acc2`
6. Again, you can see the delete button for the `Acc1` but not for yourself.

### DIFF
--- a/ui/src/config/section/account.js
+++ b/ui/src/config/section/account.js
@@ -196,6 +196,11 @@ export default {
       message: 'message.delete.account',
       dataView: true,
       show: (record, store) => {
+        // Don't allow users to delete their account
+        if (record.id !== 'undefined' && store.userInfo.accountid === record.id) {
+          return false
+        }
+
         return ['Admin', 'DomainAdmin'].includes(store.userInfo.roletype) && !record.isdefault &&
           !(record.domain === 'ROOT' && record.name === 'admin' && record.accounttype === 1)
       },
--- a/ui/src/config/section/user.js
+++ b/ui/src/config/section/user.js
@@ -145,6 +145,11 @@ export default {
       message: 'message.delete.user',
       dataView: true,
       show: (record, store) => {
+        // Don't allow users to delete their account
+        if (record.id !== 'undefined' && store.userInfo.id === record.id) {
+          return false
+        }
+
         return ['Admin', 'DomainAdmin'].includes(store.userInfo.roletype) && !record.isdefault &&
           !(record.domain === 'ROOT' && record.account === 'admin' && record.accounttype === 1)
       }


weizhouapache:
@soreana 
is there a way to disable it in backend ?

soreana:
@weizhouapache I guess we can disable that in the backend as well, but I think anyone who uses the API, or the CloudMonkey knows what he does. So, I didn't disable that.

Another reason was that in the UI we are not able to delete other admin but we can do that using the CloudMonkey.

ui/src/config/section/account.js
```diff
@@ -196,6 +196,11 @@ export default {
       message: 'message.delete.account',
       dataView: true,
       show: (record, store) => {
+        // Don't allow users to delete their account
+        if (record.id !== 'undefined' && store.userInfo.accountid === record.id) {
+          return false
+        }
+
         return ['Admin', 'DomainAdmin'].includes(store.userInfo.roletype) && !record.isdefault &&
```
harikrishna-patnala, line 204: May be disabling the action button makes it better, something like 

disabled: (record, store) => { return (record.id !== 'undefined' && store.userInfo.accountid === record.id) },

JoaoJandre: I agree that disabling the button would be a better option, what do you think @soreana ?

soreana: @harikrishna-patnala @JoaoJandre Yeah I totally agree, I didn't have time to fix it :D
Now I've applied your suggestion.

JoaoJandre:
@soreana I tested these changes manually, I do not know if my environment is too slow, but if I am fast enough, I can click the delete button before it becomes disabled. 

I'm not sure how to fix this though...

soreana:
> @soreana I tested these changes manually, I do not know if my environment is too slow, but if I am fast enough, I can click the delete button before it becomes disabled.
> 
> I'm not sure how to fix this though...

@JoaoJandre I guess the issue is related to the way UI is loaded. I'm not sure if I can control that, that is beyond my knowledge 😃

kiranchavala:
@blueorangutan ui

kiranchavala:
@soreana  Tested the ui  LGTM

The delete account icon is graded out

domainadmin

JoaoJandre (approved):
Apart from the one caveat I mentioned, it worked fine in my tests and the code looks good to me.

soreana:
@harikrishna-patnala Can you please check this pr when you have time? Thanks.

kiranchavala (approved):
LGTM, tested the changes

shwstppr (approved):

PR #6979, opened 2022-12-12
Guest os mappings improvements

DaanHoogland:
### Description

This PR is a followup of #6561

This includes the following Guest OS hypervisor mapping improvements.

- Support for the Guest OS / hypervisor mapping addition from upgrade path (using GuestOsMapper)
- Support to copy hypervisor mappings from older version from upgrade path (using GuestOsMapper)
- Provision to validate the OS mapping name in hypervisor when enabled using the new parameter 'osmappingcheckenabled' in addGuestOsMapping API
- Provision to get/check the hypervisor guest OS names (that can be used to add proper mapping) for VMware and XenServer using new API: getHypervisorGuestOsNames.
- Display the Guest OS hypervisor mappings in UI.

Document PR : https://github.com/apache/cloudstack-documentation/pull/300

Fixes: https://github.com/apache/cloudstack/issues/3609

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

PR #7272, opened 2023-02-21
guestos: prevent upgrade failures on existing mappings

shwstppr:
### Description
4.18 upgrades may fail when a certain guest OS is not present in the DB or if the guest OS hypervisor mapping is already present.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/engine/schema/src/main/java/com/cloud/upgrade/GuestOsMapper.java
+++ b/engine/schema/src/main/java/com/cloud/upgrade/GuestOsMapper.java
@@ -103,8 +103,24 @@ public class GuestOsMapper {
         return (guestOS != null);
     }
 
-    public void addGuestOsHypervisorMapping(GuestOSHypervisorMapping mapping, long guestOsId) {
+    private boolean shouldAddingGuestOsHypervisorMappingBeSkipped(GuestOSHypervisorMapping mapping, long guestOsId) {
         if (!isValidGuestOSHypervisorMapping(mapping)) {
+            return true;
+        }
+        if (guestOSDao.findById(guestOsId) == null) {
+            LOG.debug(String.format("Skipping adding guest OS hypervisor mapping - %s as guest OS ID: %d not present", mapping, guestOsId));
+            return true;
+        }
+        GuestOSHypervisorVO existingMapping = guestOSHypervisorDao.findByOsIdAndHypervisor(guestOsId, mapping.getHypervisorType(), mapping.getHypervisorVersion());
+        if (existingMapping != null) {
+            LOG.debug(String.format("Skipping adding guest OS hypervisor mapping - %s for guest OS ID: %d as a mapping already exist", mapping, guestOsId));
+            return true;
+        }
+        return false;
+    }
+
+    public void addGuestOsHypervisorMapping(GuestOSHypervisorMapping mapping, long guestOsId) {
+        if (shouldAddingGuestOsHypervisorMappingBeSkipped(mapping, guestOsId)) {
             return;
         }
         if (guestOSDao.findById(guestOsId) == null) {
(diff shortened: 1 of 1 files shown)


shwstppr:
@blueorangutan package

weizhouapache (commented):
will some guest os mappings be missing ?

I think the database needs to be fixed before upgrade. but I do not have a good idea how to do it.

engine/schema/src/main/java/com/cloud/upgrade/GuestOsMapper.java
```diff
@@ -104,7 +104,16 @@ private boolean addGuestOs(long categoryId, String displayName) {
     }
 
     public void addGuestOsHypervisorMapping(GuestOSHypervisorMapping mapping, long guestOsId) {
-        if(!isValidGuestOSHypervisorMapping(mapping)) {
+        if (!isValidGuestOSHypervisorMapping(mapping)) {
+            return;
+        }
+        if (guestOSDao.findById(guestOsId) == null) {
+            LOG.debug(String.format("Skipping adding guest OS hypervisor mapping - %s as guest OS ID: %d not preset", mapping, guestOsId));
```
shwstppr, line 111:
```suggestion
            LOG.debug(String.format("Skipping adding guest OS hypervisor mapping - %s as guest OS ID: %d not present", mapping, guestOsId));
```

shwstppr:
@weizhouapache In my opinion, this method `addGuestOsHypervisorMapping` could create issues as it takes the guest OS DB ID. If checks are not done this could cause failures when guest OS entry for that DB ID is not present. This PR just tries to avoid those failures

DaanHoogland (approved):
clgtm

weizhouapache:
> @weizhouapache In my opinion, this method `addGuestOsHypervisorMapping` could create issues as it takes the guest OS DB ID. If checks are not done this could cause failures when guest OS entry for that DB ID is not present. This PR just tries to avoid those failures

I agree that the failures in upgrade will be gone and cloudstack will succeed. but the guest os mappings will be even worse (more mapping are missing which could take much longer time to fix and cause more issues)

borisstoyanov:
@blueorangutan package

borisstoyanov (approved):
Upgrade was successful

DaanHoogland:
@weizhouapache @borisstoyanov I think that if the user did anything in terms of mappings, they assume responsibility to maintain their guestos-hypervisor mapping. We can make a repair procedure or provide default lists (which is easily done by instelling fresh and then exporting the appropriate tables.

I also think all this is out of scope. This could be a standard procedure in future releases or a separate tool.

DaanHoogland:
@weizhouapache , do you approve?

weizhouapache (approved):

weizhouapache:
> @weizhouapache In my opinion, this method `addGuestOsHypervisorMapping` could create issues as it takes the guest OS DB ID. If checks are not done this could cause failures when guest OS entry for that DB ID is not present. This PR just tries to avoid those failures

@DaanHoogland thanks. this PR looks good to me.

However, do not upgrade cloudstack if guest os and guest os mappings mess up.  @borisstoyanov

DaanHoogland:
@weizhouapache , you convinced me that we shouldn't merge this. We do not want to give a false impression everything is alright. Users that have blocking mappings should first remove those.

DaanHoogland:
as we have merged 7279 and Guest os mappings improvements in #6979 is underway I think we can close this one @shwstppr ???

PR #7304, opened 2023-03-02
Fix VNC TLS for non-root linux users on the QEMU configuration

nvazquez:
### Description

This PR fixes an issue on KVM environments where the user and group set on the /etc/libvirt/qemu.conf file is not root:root.

- A common user and group to be used is qemu:qemu. 
- This PR allows different user within the qemu group to access the certificates for the VNC TLS encryption on the VM console access

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

Screenshot 2023-03-02 at 08 24 24

### How Has This Been Tested?
On KVM hosts, edit the /etc/libvirt/qemu.conf file and set user="qemu" and group="qemu", then restart the libvritd service. Try to deploy a new VM.
Before the fix -> ERROR: cannot access the certificates

### DIFF
--- a/scripts/util/keystore-cert-import
+++ b/scripts/util/keystore-cert-import
@@ -121,6 +121,9 @@ if [ -f "$LIBVIRTD_FILE" ]; then
     ln -sf /etc/pki/libvirt/servercert.pem /etc/pki/libvirt-vnc/server-cert.pem
     ln -sf /etc/pki/libvirt/private/serverkey.pem /etc/pki/libvirt-vnc/server-key.pem
     cloudstack-setup-agent -s > /dev/null
+
+    chgrp qemu /etc/pki/libvirt /etc/pki/libvirt-vnc /etc/pki/CA /etc/pki/libvirt/private /etc/pki/libvirt/servercert.pem /etc/pki/libvirt/private/serverkey.pem /etc/pki/CA/cacert.pem /etc/pki/libvirt-vnc/ca-cert.pem /etc/pki/libvirt-vnc/server-cert.pem /etc/pki/libvirt-vnc/server-key.pem
+    chmod 750 /etc/pki/libvirt /etc/pki/libvirt-vnc /etc/pki/CA /etc/pki/libvirt/private /etc/pki/libvirt/servercert.pem /etc/pki/libvirt/private/serverkey.pem /etc/pki/CA/cacert.pem /etc/pki/libvirt-vnc/ca-cert.pem /etc/pki/libvirt-vnc/server-cert.pem /etc/pki/libvirt-vnc/server-key.pem
 fi
 
 # Update ca-certs if we're in systemvm
@@ -138,6 +141,6 @@ if [ -f "$SYSTEM_FILE" ]; then
 fi
 
 # Fix file permission
-chmod 600 $CACERT_FILE
-chmod 600 $CERT_FILE
-chmod 600 $PRIVKEY_FILE
+chmod 750 $CACERT_FILE
+chmod 750 $CERT_FILE
+chmod 750 $PRIVKEY_FILE
--- a/scripts/util/keystore-cert-import
+++ b/scripts/util/keystore-cert-import
@@ -122,7 +122,7 @@ if [ -f "$LIBVIRTD_FILE" ]; then
     ln -sf /etc/pki/libvirt/private/serverkey.pem /etc/pki/libvirt-vnc/server-key.pem
     cloudstack-setup-agent -s > /dev/null
 
-    QEMU_GROUP=$(cat /etc/group | awk -F: '{ print $1}' | grep qemu)
+    QEMU_GROUP=$(sed -n 's/^group=//p' /etc/libvirt/qemu.conf | awk -F'"' '{print $2}')
     chgrp $QEMU_GROUP /etc/pki/libvirt /etc/pki/libvirt-vnc /etc/pki/CA /etc/pki/libvirt/private /etc/pki/libvirt/servercert.pem /etc/pki/libvirt/private/serverkey.pem /etc/pki/CA/cacert.pem /etc/pki/libvirt-vnc/ca-cert.pem /etc/pki/libvirt-vnc/server-cert.pem /etc/pki/libvirt-vnc/server-key.pem
     chmod 750 /etc/pki/libvirt /etc/pki/libvirt-vnc /etc/pki/CA /etc/pki/libvirt/private /etc/pki/libvirt/servercert.pem /etc/pki/libvirt/private/serverkey.pem /etc/pki/CA/cacert.pem /etc/pki/libvirt-vnc/ca-cert.pem /etc/pki/libvirt-vnc/server-cert.pem /etc/pki/libvirt-vnc/server-key.pem
 fi
--- a/scripts/util/keystore-cert-import
+++ b/scripts/util/keystore-cert-import
@@ -122,7 +122,8 @@ if [ -f "$LIBVIRTD_FILE" ]; then
     ln -sf /etc/pki/libvirt/private/serverkey.pem /etc/pki/libvirt-vnc/server-key.pem
     cloudstack-setup-agent -s > /dev/null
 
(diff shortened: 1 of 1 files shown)


nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

scripts/util/keystore-cert-import
```diff
@@ -138,6 +141,6 @@ if [ -f "$SYSTEM_FILE" ]; then
 fi
 
 # Fix file permission
-chmod 600 $CACERT_FILE
-chmod 600 $CERT_FILE
-chmod 600 $PRIVKEY_FILE
+chmod 750 $CACERT_FILE
+chmod 750 $CERT_FILE
+chmod 750 $PRIVKEY_FILE
```
DaanHoogland, line 146: are execute bits needed for these files? seems 640 would be good.

scripts/util/keystore-cert-import
```diff
@@ -121,6 +121,9 @@ if [ -f "$LIBVIRTD_FILE" ]; then
     ln -sf /etc/pki/libvirt/servercert.pem /etc/pki/libvirt-vnc/server-cert.pem
     ln -sf /etc/pki/libvirt/private/serverkey.pem /etc/pki/libvirt-vnc/server-key.pem
     cloudstack-setup-agent -s > /dev/null
+
+    chgrp qemu /etc/pki/libvirt /etc/pki/libvirt-vnc /etc/pki/CA /etc/pki/libvirt/private /etc/pki/libvirt/servercert.pem /etc/pki/libvirt/private/serverkey.pem /etc/pki/CA/cacert.pem /etc/pki/libvirt-vnc/ca-cert.pem /etc/pki/libvirt-vnc/server-cert.pem /etc/pki/libvirt-vnc/server-key.pem
+    chmod 750 /etc/pki/libvirt /etc/pki/libvirt-vnc /etc/pki/CA /etc/pki/libvirt/private /etc/pki/libvirt/servercert.pem /etc/pki/libvirt/private/serverkey.pem /etc/pki/CA/cacert.pem /etc/pki/libvirt-vnc/ca-cert.pem /etc/pki/libvirt-vnc/server-cert.pem /etc/pki/libvirt-vnc/server-key.pem
```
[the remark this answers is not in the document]
nvazquez: Unfortunately it didn't work, got the permission issue for certificates while trying to deploy

scripts/util/keystore-cert-import
nvazquez: These 3 are included on the command above so at this stage the permissions are overriden

borisstoyanov (commented):
@nvazquez will be looking for a dynamic way to find the libvirt user in different distros because in centos it's qemu:qemu, but in ubuntu it's libvirt-qemu:libvirt-qemu. 

I've suggested to check the /etc/passwd file and grep for qemu, 1st column should be the user

borisstoyanov:
@blueorangutan package

borisstoyanov (approved):
manually patched and checked the lates changes, they are working both on ubuntu22 and centos7 and 8. I'm a bit doubtful if there would be a custom group named qemu that could mess up the setup

scripts/util/keystore-cert-import
```diff
@@ -121,6 +121,10 @@ if [ -f "$LIBVIRTD_FILE" ]; then
     ln -sf /etc/pki/libvirt/servercert.pem /etc/pki/libvirt-vnc/server-cert.pem
     ln -sf /etc/pki/libvirt/private/serverkey.pem /etc/pki/libvirt-vnc/server-key.pem
     cloudstack-setup-agent -s > /dev/null
+
+    QEMU_GROUP=$(cat /etc/group | awk -F: '{ print $1}' | grep qemu)
+    chgrp $QEMU_GROUP /etc/pki/libvirt /etc/pki/libvirt-vnc /etc/pki/CA /etc/pki/libvirt/private /etc/pki/libvirt/servercert.pem /etc/pki/libvirt/private/serverkey.pem /etc/pki/CA/cacert.pem /etc/pki/libvirt-vnc/ca-cert.pem /etc/pki/libvirt-vnc/server-cert.pem /etc/pki/libvirt-vnc/server-key.pem
```
weizhouapache, line 126: @nvazquez 
what if there are multiple groups in the output ?

weizhouapache: @nvazquez 
since the user/group is specified in `/etc/libvirt/qemu.conf`, should the group be fetched from qemu.conf ?

nvazquez: That's a good idea @weizhouapache, thanks. I've created a documentation PR with a note for the limitation you spotted: https://github.com/apache/cloudstack-documentation/pull/306 but I think using the qemu.conf is a better approach cc @borisstoyanov

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

scripts/util/keystore-cert-import
weizhouapache: @nvazquez 
1. You can add 'head -n1' to get the first result.
2. You may need to check if result is empty or not

scripts/util/keystore-cert-import
```diff
@@ -121,6 +121,10 @@ if [ -f "$LIBVIRTD_FILE" ]; then
     ln -sf /etc/pki/libvirt/servercert.pem /etc/pki/libvirt-vnc/server-cert.pem
     ln -sf /etc/pki/libvirt/private/serverkey.pem /etc/pki/libvirt-vnc/server-key.pem
     cloudstack-setup-agent -s > /dev/null
+
+    QEMU_GROUP=$(sed -n 's/^group=//p' /etc/libvirt/qemu.conf | awk -F'"' '{print $2}')
```
rohityadavcloud, line 125: This isn't fullproof and covers the case generally.  Perhaps the logic can be (a) assuming the libvirt group/user can access this file, then this would work; othewise on error detect the distro and (b) use the group "libvirt-qemu" if it's ubuntu, otherwise use the group "qemu" if it's RHEL or one of the other EL clones (I haven't tested for suse15).

As normal non-root user I couldn't run this, but can run as sudoer:

1. On RHEL9:
```
[rohit@plex ~]$ sudo sed -n 's/^group=//p' /etc/libvirt/qemu.conf | awk -F'"' '{print $2}'
root
[rohit@plex ~]$ sed -n 's/^group=//p' /etc/libvirt/qemu.conf | awk -F'"' '{print $2}'
sed: can't read /etc/libvirt/qemu.conf: Permission denied
```

2. On Ubuntu 22.04:
```
> sed -n 's/^group=//p' /etc/libvirt/qemu.conf | awk -F'"' '{print $2}'
sed: can't read /etc/libvirt/qemu.conf: Permission denied
> sudo sed -n 's/^group=//p' /etc/libvirt/qemu.conf | awk -F'"' '{print $2}'
root
```

scripts/util/keystore-cert-import
rohityadavcloud: same as Daan - I don't think we need executable flag on certificate/keys file; read should be good?

rohityadavcloud (changes requested):
Left some remarks to see if can further improve the logic around qemu groups.
Also see https://libvirt.org/drvqemu.html#posix-users-groups

This will work on the root user (default/case), but for non-root linux users we need to check/fix or document if we're making any assumptions here (for ex. non-root user already is part of the qemu group and can read the /etc/libvirt/qemu.conf file and call the sed command without failing).

scripts/util/keystore-cert-import
nvazquez: Thanks @weizhouapache, I've used tail actually to get the last value

scripts/util/keystore-cert-import
nvazquez: Hi @rohityadavcloud I think this won't be a problem since we just want to access the qemu.conf file from the CloudStack script to obtain the group that the administrator has set, that won't be running as the non-root user. In case qemu doesn't start, then the admin may need to fix the qemu.conf file accordingly

nvazquez:
@blueorangutan test

borisstoyanov (approved):
LGTM, tested both on ubuntu22 and centos7

DaanHoogland (approved):
approving in spite of the remaining  comment. It seems 640 does not work.
clgtm otherwise and this is a qemu feature, not a cloudstack issue.

Issue #7357, opened 2023-03-22
VM creation with data volume fails when storage tags are used and 1st cluster has insufficient storage capacity


NuxRo:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME
~~~
Storage, scheduler
~~~

##### CLOUDSTACK VERSION

~~~
Verified in 4.11 as well as 4.18, it's safe to assume it's present in all versions.
~~~

##### SUMMARY

In certain circumstances - when storage tags are used and there is also a cluster with insufficient capacity that gets picked first - we end up with a situation where avoid sets for the root volume exclude the pools needed for the data volume, thus leading to a failed deployment.

##### STEPS TO REPRODUCE

~~~
1. Create a compute offering with storage tag1 and a disk offering with storage tag2
2. Create two clusters with primary storages having two sperate tags for root disk and data disk 
offering ( avoid adding data disk tag to the first cluster primary storage).
3. First cluster (first one to be selected by deployment planner) should have host capacity, 
but should not have storage capacity to deploy vm with root and data disk, 
or do not add data disk tag so that it won't return suitable storage pools.
4. The second cluster will be selected next. Although it has host capacity,
matching root and data disk tags the primary storages having the middle disk
will be ignored logging that it is in avoid set( incorrectly).

~~~

PR #7372, opened 2023-03-27
Fixed avoid set variables which is causing deployment failures

harikrishna-patnala:
### Description

This PR fixes the issue #7357 where VM deployment is failing when storage tags are used in certain way with multiple clusters, though a proper cluster to deploy VM is available.

Root cause is the usage of certain variables to figure out the final avoid pools set in the deployment planner.

image

because of this, "originalAvoidPoolSet" is getting updated whenever avoid.getPoolsToAvoid() is updated and this should not be the case.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
+++ b/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
@@ -1576,11 +1576,11 @@ StateListener<State, VirtualMachine.Event, VirtualMachine>, Configurable {
 
         // for each volume find list of suitable storage pools by calling the
         // allocators
-        Set<Long> originalAvoidPoolSet = avoid.getPoolsToAvoid();
-        if (originalAvoidPoolSet == null) {
-            originalAvoidPoolSet = new HashSet<Long>();
+        Set<Long> originalAvoidPoolSet = new HashSet<>();
+        if (avoid.getPoolsToAvoid() != null) {
+            originalAvoidPoolSet.addAll(avoid.getPoolsToAvoid());
         }
-        Set<Long> poolsToAvoidOutput = new HashSet<Long>(originalAvoidPoolSet);
+        Set<Long> poolsToAvoidOutput = new HashSet<>(originalAvoidPoolSet);
 
         for (VolumeVO toBeCreated : volumesTobeCreated) {
             s_logger.debug("Checking suitable pools for volume (Id, Type): (" + toBeCreated.getId() + "," + toBeCreated.getVolumeType().name() + ")");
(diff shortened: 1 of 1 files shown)


harikrishna-patnala:
This issue is in 4.17 as well, so rebased the branch to 4.17 as this is a critical issue

harikrishna-patnala:
@blueorangutan package

server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
```diff
@@ -1576,11 +1576,11 @@ protected Pair<Map<Volume, List<StoragePool>>, List<Volume>> findSuitablePoolsFo
...
         // allocators
-        Set<Long> originalAvoidPoolSet = avoid.getPoolsToAvoid();
-        if (originalAvoidPoolSet == null) {
-            originalAvoidPoolSet = new HashSet<Long>();
+        Set<Long> originalAvoidPoolSet = new HashSet<>();
+        if (avoid.getPoolsToAvoid() != null) {
+            originalAvoidPoolSet.addAll(avoid.getPoolsToAvoid());
         }
-        Set<Long> poolsToAvoidOutput = new HashSet<Long>(originalAvoidPoolSet);
+        Set<Long> poolsToAvoidOutput = new HashSet<>(originalAvoidPoolSet);
```
DaanHoogland, line 1583: @harikrishna-patnala , what is the logic here? As I see it it just adds to the avoid set to remember later and doesn´t clean the pools from trying the prior cluster from the avoid set trying this cluster. 

If it works it works but It doesn't read like logical to me.

harikrishna-patnala: Logic is already down there @DaanHoogland.
https://github.com/apache/cloudstack/blob/4c60621458a3f9680d14bfb5eda001e99c2d7221/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java#L1685-L1715

It is just that originalAvoidPoolSet was not maintained properly because of "Set<Long> originalAvoidPoolSet = avoid.getPoolsToAvoid();".  Later in the planner whenever avoid.getPoolsToAvoid() is getting updated, originalAvoidPoolSet is also getting updated.

For example with 2 volumes, during the loop for each volume, for 1st volume avoid.getPoolsToAvoid() adds p3, p4 and also in originalAvoidPoolSet. For the 2nd volume p3, p4 should not be in avoid set, but since originalAvoidPoolSet also got updated planner is avoiding p3, p4. 

Yes, this is working during my testing. I've verified with the debugger too.

DaanHoogland (approved):
clgtm

harikrishna-patnala:
@blueorangutan test

rajujith (approved):
Followed the issue reproduction steps and the issue is not observed. Its LGTM.

PR #7384, opened 2023-03-31
Roles search by keyword

### LINKED ISSUE #7362: api call listRoles name filter not working as expected
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API
~~~

##### CLOUDSTACK VERSION

~~~
4.18.0.0
~~~

##### SUMMARY

When using parameter name with api call listRoles it lists all roles which have the value of name in the actual name instead of the roles with the exact value of name. Parameter keyword should be used to filter when wanting to find all roles where the value is part of the name.

##### STEPS TO REPRODUCE

~~~
Create a new role with name "New Domain Admin".
Do the following api call: /client/api/?command=listRoles&name=Domain%20Admin&response=json
~~~

##### EXPECTED RESULTS

~~~
You should only get 1 result back. The result should be the default "Domain Admin" role.
~~~

##### ACTUAL RESULTS

~~~
You get 2 results back. The default "Domain Admin" and the new created "New Domain Admin".
~~~

harikrishna-patnala:
### Description

This PR fixes #7362 and also other search criteria to use the name as an exact search where keyword is also there.

Made UI changes for roles search to make use of keyword instead of name.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/plugins/metrics/src/main/java/org/apache/cloudstack/metrics/MetricsServiceImpl.java
+++ b/plugins/metrics/src/main/java/org/apache/cloudstack/metrics/MetricsServiceImpl.java
@@ -287,7 +287,7 @@ public class MetricsServiceImpl extends MutualExclusiveIdsManagerBase implements
 
         SearchBuilder<VMInstanceVO> sb =  vmInstanceDao.createSearchBuilder();
         sb.and("idIN", sb.entity().getId(), SearchCriteria.Op.IN);
-        sb.and("name", sb.entity().getName(), SearchCriteria.Op.LIKE);
+        sb.and("name", sb.entity().getName(), SearchCriteria.Op.EQ);
         sb.and("state", sb.entity().getState(), SearchCriteria.Op.EQ);
         sb.and("type", sb.entity().getType(), SearchCriteria.Op.NEQ);
 
@@ -340,7 +340,7 @@ public class MetricsServiceImpl extends MutualExclusiveIdsManagerBase implements
 
         SearchBuilder<VolumeVO> sb =  volumeDao.createSearchBuilder();
         sb.and("idIN", sb.entity().getId(), SearchCriteria.Op.IN);
-        sb.and("name", sb.entity().getName(), SearchCriteria.Op.LIKE);
+        sb.and("name", sb.entity().getName(), SearchCriteria.Op.EQ);
         sb.and("state", sb.entity().getState(), SearchCriteria.Op.EQ);
 
         SearchCriteria<VolumeVO> sc = sb.create();
--- a/server/src/main/java/com/cloud/network/vpc/VpcManagerImpl.java
+++ b/server/src/main/java/com/cloud/network/vpc/VpcManagerImpl.java
@@ -1391,7 +1391,7 @@ public class VpcManagerImpl extends ManagerBase implements VpcManager, VpcProvis
         final SearchBuilder<VpcVO> sb = vpcDao.createSearchBuilder();
         _accountMgr.buildACLSearchBuilder(sb, domainId, isRecursive, permittedAccounts, listProjectResourcesCriteria);
 
-        sb.and("name", sb.entity().getName(), SearchCriteria.Op.LIKE);
+        sb.and("name", sb.entity().getName(), SearchCriteria.Op.EQ);
         sb.and("id", sb.entity().getId(), SearchCriteria.Op.EQ);
         sb.and("displayText", sb.entity().getDisplayText(), SearchCriteria.Op.LIKE);
         sb.and("vpcOfferingId", sb.entity().getVpcOfferingId(), SearchCriteria.Op.EQ);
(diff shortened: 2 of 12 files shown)


harikrishna-patnala:
@blueorangutan package

harikrishna-patnala:
@blueorangutan package

harikrishna-patnala:
@blueorangutan package

harikrishna-patnala:
@blueorangutan test

plugins/metrics/src/main/java/org/apache/cloudstack/metrics/MetricsServiceImpl.java
```diff
@@ -340,7 +340,7 @@ protected Pair<List<VolumeVO>, Integer> searchForVolumesInternal(ListVolumesUsag
 
         SearchBuilder<VolumeVO> sb =  volumeDao.createSearchBuilder();
         sb.and("idIN", sb.entity().getId(), SearchCriteria.Op.IN);
-        sb.and("name", sb.entity().getName(), SearchCriteria.Op.LIKE);
```
DaanHoogland, line 343: incompatible

server/src/main/java/com/cloud/network/vpc/VpcManagerImpl.java
```diff
@@ -1391,7 +1391,7 @@ public Pair<List<? extends Vpc>, Integer> listVpcs(final Long id, final String v
         final SearchBuilder<VpcVO> sb = vpcDao.createSearchBuilder();
         _accountMgr.buildACLSearchBuilder(sb, domainId, isRecursive, permittedAccounts, listProjectResourcesCriteria);
 
-        sb.and("name", sb.entity().getName(), SearchCriteria.Op.LIKE);
```
DaanHoogland, line 1394: incompatible

plugins/metrics/src/main/java/org/apache/cloudstack/metrics/MetricsServiceImpl.java
```diff
@@ -287,7 +287,7 @@ protected Pair<List<VMInstanceVO>, Integer> searchForSystemVmsInternal(ListSyste
 
         SearchBuilder<VMInstanceVO> sb =  vmInstanceDao.createSearchBuilder();
         sb.and("idIN", sb.entity().getId(), SearchCriteria.Op.IN);
-        sb.and("name", sb.entity().getName(), SearchCriteria.Op.LIKE);
```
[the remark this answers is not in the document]
DaanHoogland: also, but not in scope maybe, searchbuilders do not belong in the service layer. These should be moved to the DAOs

server/src/main/java/org/apache/cloudstack/backup/BackupManagerImpl.java
```diff
@@ -223,7 +223,7 @@ public Pair<List<BackupOffering>, Integer> listBackupOfferings(final ListBackupO
         final Filter searchFilter = new Filter(BackupOfferingVO.class, "id", true, cmd.getStartIndex(), cmd.getPageSizeVal());
         SearchBuilder<BackupOfferingVO> sb = backupOfferingDao.createSearchBuilder();
         sb.and("zone_id", sb.entity().getZoneId(), SearchCriteria.Op.EQ);
-        sb.and("name", sb.entity().getName(), SearchCriteria.Op.LIKE);
```
DaanHoogland, line 226: incompatible

harikrishna-patnala:
> one problem @harikrishna-patnala , people using a search on name with the API won´t be able to find by partial names anymore

They can use the 'keyword' param @DaanHoogland. Yes I agree already existing scripts/configurations in environments may have issues if they are using APIs with name parameter. But I feel this has to be fixed, for example

1. There are two VMs with names "ABCD" and "ABCDxyz"
2. When I search with name "ABCD" using API both VMs will appear in the list (if name search considers partial names)

If this is the way it has used in some scripts or configuration, it will be difficult to find the VM.

harikrishna-patnala:
> > one problem @harikrishna-patnala , people using a search on name with the API won´t be able to find by partial names anymore
> 
> They can use the 'keyword' param @DaanHoogland. Yes I agree already existing scripts/configurations in environments may have issues if they are using APIs with name parameter. But I feel this has to be fixed, for example
> 
> 1. There are two VMs with names "ABCD" and "ABCDxyz"
> 2. When I search with name "ABCD" using API both VMs will appear in the list (if name search considers partial names)
> 
> If this is the way it has used in some scripts or configuration, it will be difficult to find the VM.

@DaanHoogland any thoughts here

DaanHoogland:
> > > one problem @harikrishna-patnala , people using a search on name with the API won´t be able to find by partial names anymore
> > 
> > 
> > They can use the 'keyword' param @DaanHoogland. Yes I agree already existing scripts/configurations in environments may have issues if they are using APIs with name parameter. But I feel this has to be fixed, for example
> > 
> > 1. There are two VMs with names "ABCD" and "ABCDxyz"
> > 2. When I search with name "ABCD" using API both VMs will appear in the list (if name search considers partial names)
> > 
> > If this is the way it has used in some scripts or configuration, it will be difficult to find the VM.
> 
> @DaanHoogland any thoughts here

yes, too many and contradicting. I was mostly biased against backwards incompatibilities when I first responded. I still tend to, but we might get away with a notice and considder it a fix.

shwstppr:
@blueorangutan package

shwstppr (approved):
Code LGTM. Makes behaviour consistent with other APIs

shwstppr:
@blueorangutan test

rohityadavcloud (approved):
LGTM didn't test it, but code looks alright

PR #6918, opened 2022-11-23
Support network-scope settings and add setting to override vr service offering

soreana:
### Description

As of now, it is possible to override the virtual router service offering with global, domain, and account level configurations. What if a customer wants to change the service offering for only one network? There isn't such an option in Cloudstack. As an admin, you can only override the network offering for a domain, account, or global, affecting all the networks in that domain, account, or zone. This pull request aims to add a network-level setting to override the virtual router service offering per network.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [x] Major
- [ ] Minor

### How Has This Been Tested?

To test this feature, please follow the below steps:

1. Create a network
2. Create VM in that network (to create a VR)
3. List the network and get the UUID of the network
4. Check the service offering of the network
5. Create a new router service offering
6. Set the `network.router.service.offering` to the UUID of the service offering for the network using the cmk
7. Restart the network with clean up
8. Check the service offering of the network

Useful `cmk` commands:

```
(localcloud) 🐱 > list networks filter=id,name
{
  "count": 2,
  "network": [
    {
      "id": "c9d70872-b877-4f02-92b5-b4b08e1095d7",
      "name": "isolated-network"
    },
    {
      "id": "1ebf1ca9-7930-4f2f-8263-8da1840f02fc",
      "name": "shared"
    }
  ]
}
(localcloud) 🐱 > list configurations networkid=c9d70872-b877-4f02-92b5-b4b08e1095d7
{
  "configuration": [
    {
      "category": "Advanced",
      "description": "Uuid of the service offering used by virtual routers of this network; if NULL - account/system offering will be used",
      "isdynamic": true,
      "name": "network.router.service.offering",
      "scope": "network",
      "value": ""
    }
  ],
  "count": 1
}
(localcloud) 🐱 > list routers filter=serviceofferingid,serviceofferingname
{
  "count": 1,
  "router": [
    {
      "serviceofferingid": "f37ed5fd-6f26-4316-8b27-dbebb0b56a5a"
      "serviceofferingname": "System Offering For Software Router"
    }
  ]
}
[...]

PR #7405, opened 2023-04-05
Enable defining a network as redundant during restart through the UI

GaOrtiga:
### Description

By utilizing the `restartNetwork` API and setting the `makeredundant` parameter to true, ACS allows the user to make a guest network redundant during its restart process; however, this feature was only available through the API.
An adjustment was made to add this functionality to the UI.
### Types of changes

- [ ] Major
- [X] Minor

image

### DIFF
--- a/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
+++ b/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
@@ -2992,7 +2992,6 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService, C
         User callerUser = _accountMgr.getActiveUser(CallContext.current().getCallingUserId());
         if (makeRedundant && !_accountMgr.isRootAdmin(callerUser.getAccountId()) && !AllowUsersToMakeNetworksRedundant.value() ) {
             throw new InvalidParameterValueException("Could not make the network redundant. Please contact administrator.");
-                    AllowUsersToMakeNetworksRedundant.key()));
         }
 
         boolean livePatch = cmd.getLivePatch();
--- a/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
+++ b/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
@@ -280,6 +280,9 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService, C
     private static final ConfigKey<Boolean> AllowEmptyStartEndIpAddress = new ConfigKey<>("Advanced", Boolean.class,
             "allow.empty.start.end.ipaddress", "true", "Allow creating network without mentioning start and end IP address",
             true, ConfigKey.Scope.Account);
+    public static final ConfigKey<Boolean> AllowUsersToMakeNetworksRedundant = new ConfigKey<>("Advanced", Boolean.class,
+            "allow.users.to.make.networks.redundant", "true", "Allow Users to make Networks Redundant",
+            true, ConfigKey.Scope.Global);
     private static final long MIN_VLAN_ID = 0L;
     private static final long MAX_VLAN_ID = 4095L; // 2^12 - 1
     private static final long MIN_GRE_KEY = 0L;
@@ -2986,8 +2989,13 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService, C
             throwInvalidIdException("Cannot restart a VPC tier with cleanup, please restart the whole VPC.", network.getUuid(), "network tier");
         }
         boolean makeRedundant = cmd.getMakeRedundant();
-        boolean livePatch = cmd.getLivePatch();
         User callerUser = _accountMgr.getActiveUser(CallContext.current().getCallingUserId());
+        if (makeRedundant && !_accountMgr.isRootAdmin(callerUser.getAccountId()) && !AllowUsersToMakeNetworksRedundant.value() ) {
+            throw new InvalidParameterValueException(String.format("Could not make network redundant as calling user is not Root Admin and [%s] is set to false.",
+                    AllowUsersToMakeNetworksRedundant.key()));
+        }
+
+        boolean livePatch = cmd.getLivePatch();
         return restartNetwork(network, cleanup, makeRedundant, livePatch, callerUser);
     }
 
@@ -6266,7 +6274,7 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService, C
 
     @Override
     public ConfigKey<?>[] getConfigKeys() {
-        return new ConfigKey<?>[] {AllowDuplicateNetworkName, AllowEmptyStartEndIpAddress, VRPrivateInterfaceMtu, VRPublicInterfaceMtu, AllowUsersToSpecifyVRMtu};
+        return new ConfigKey<?>[] {AllowDuplicateNetworkName, AllowEmptyStartEndIpAddress, AllowUsersToMakeNetworksRedundant, VRPrivateInterfaceMtu, VRPublicInterfaceMtu, AllowUsersToSpecifyVRMtu};
     }
 
     public boolean isDefaultAcl(Long aclId) {
--- a/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
+++ b/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
@@ -2991,7 +2991,7 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService, C
         boolean makeRedundant = cmd.getMakeRedundant();
         User callerUser = _accountMgr.getActiveUser(CallContext.current().getCallingUserId());
         if (makeRedundant && !_accountMgr.isRootAdmin(callerUser.getAccountId()) && !AllowUsersToMakeNetworksRedundant.value() ) {
-            throw new InvalidParameterValueException(String.format("Could not make network redundant as calling user is not Root Admin and [%s] is set to false.",
+            throw new InvalidParameterValueException("Could not make the network redundant. Please contact administrator.");
                     AllowUsersToMakeNetworksRedundant.key()));
         }
 
--- a/ui/public/config.json
+++ b/ui/public/config.json
@@ -95,6 +95,7 @@
     ]
   },
   "plugins": [],
+  "allowMakingRouterRedundant": false,
   "apidocs": true,
   "basicZoneEnabled": true,
   "multipleServer": false,
--- a/ui/src/config/section/network.js
+++ b/ui/src/config/section/network.js
@@ -171,13 +171,16 @@ export default {
             if (isGroupAction || record.vpcid == null) {
               fields.push('cleanup')
             }
+            if (!record.redundantrouter) {
+              fields.push('makeredundant')
+            }
             fields.push('livepatch')
             return fields
           },
           show: (record) => record.type !== 'L2',
           groupAction: true,
           popup: true,
-          groupMap: (selection, values) => { return selection.map(x => { return { id: x, cleanup: values.cleanup } }) }
+          groupMap: (selection, values) => { return selection.map(x => { return { id: x, cleanup: values.cleanup, makeredundant: values.makeredundant } }) }
         },
         {
           api: 'replaceNetworkACLList',
--- a/ui/src/config/section/network.js
+++ b/ui/src/config/section/network.js
@@ -20,6 +20,7 @@ import store from '@/store'
 import tungsten from '@/assets/icons/tungsten.svg?inline'
 import { isAdmin } from '@/role'
 import { isZoneCreated } from '@/utils/zone'
+import { vueProps } from '@/vue-app'
 
 export default {
   name: 'network',
@@ -171,7 +172,7 @@ export default {
             if (isGroupAction || record.vpcid == null) {
               fields.push('cleanup')
             }
-            if (!record.redundantrouter) {
+            if (!record.redundantrouter && vueProps.$config.allowMakingRouterRedundant) {
               fields.push('makeredundant')
             }
             fields.push('livepatch')


DaanHoogland:
@GaOrtiga i think this was removed from the UI on purpose. I don´t mind either way but it seems to me operators want to have control on whether users are allowed to make this mistake (as i understood it).

DaanHoogland:
code lgtm btw

borisstoyanov:
exactly, tbh I'm against this api option to exist at all since whether redundant or not should be defined in the offering not as a  single action

GaOrtiga:
> @GaOrtiga i think this was removed from the UI on purpose. I don´t mind either way but it seems to me operators want to have control on whether users are allowed to make this mistake (as i understood it).

I understand, however, users already can do this through the API; therefore, operators don't fully control it. Also, for VPCs this option exists in the UI while restarting them. 
I see @borisstoyanov's point on redundant being a feature defined on the offering, however, the discussion would be outside of the PR's scope; could we create an issue or a thread on the mail-list to discuss the `make redundant` feature (for both VPC and guest networks)?

weizhouapache (approved):
As far as I know, some users wanted to disable this on UI, but some others want this. 
As a compromise, can we enable it only for root admin ? if users want it to be visible for domain admins or regular users, create a new network offering and do it through network update with the new network offering.

GaOrtiga:
> As far as I know, some users wanted to disable this on UI, but some others want this. As a compromise, can we enable it only for root admin ? if users want it to be visible for domain admins or regular users, create a new network offering and do it through network update with the new network offering.

We could also apply these changes to the UI to maintain consistence with the VPCs section and provide an option in the backend that lets admins determine whether or not to allow users to have access to this. This would standardize the code and address both issues. What are your thoughts on this? If you guys agree on that, we can move with this PR as is, and then create a new one to introduce this other featureset into ACS.

DaanHoogland:
> > As far as I know, some users wanted to disable this on UI, but some others want this. As a compromise, can we enable it only for root admin ? if users want it to be visible for domain admins or regular users, create a new network offering and do it through network update with the new network offering.
> 
> We could also apply these changes to the UI to maintain consistence with the VPCs section and provide an option in the backend that lets admins determine whether or not to allow users to have access to this. This would standardize the code and address both issues. What are your thoughts on this? If you guys agree on that, we can move with this PR as is, and then create a new one to introduce this other featureset into ACS.

I agree with the addition that disabling it in the backend should also remove the option on the UI.

DaanHoogland (approved):

weizhouapache:
what's your opinions ?
cc @rohityadavcloud @alexandremattioli @andrijapanicsb @NuxRo

andrijapanicsb:
There were a lot of cases where people have "accidentally" switched this on and didn't know the consequences, so... not sure what is the best way forward.

rohityadavcloud:
Sorry -1, we had a conversation in the past. In fact this feature never had existed, until I had put it in. After discussions and feedback from users esp service providers, we decided to remove the option from the UI though the API param may still be available via UI. This is because service providers may have an explicit network offering with redundant routers, users without using/paying for one could this form to circumvent to get redundant routers for a network not intended for them. 

However, there is a middle ground but probably not necessary - it could be via adding a global setting for operators to decide if they want to enable such a feature for end users. (listCapabilities or other means can export this option to the UI to show/hide such a setting).

weizhouapache:
@GaOrtiga 
it seems difficult to get a consensus.

will you consider a new global setting to manage it ? (requires UI, API and service layer changes)

GaOrtiga:
> There were a lot of cases where people have "accidentally" switched this on and didn't know the consequences, so... not sure what is the best way forward.

We could add a warning message explaining the consequences when that option is selected

> Sorry -1, we had a conversation in the past. In fact this feature never had existed, until I had put it in. After discussions and feedback from users esp service providers, we decided to remove the option from the UI though the API param may still be available via UI. This is because service providers may have an explicit network offering with redundant routers, users without using/paying for one could this form to circumvent to get redundant routers for a network not intended for them.
> 
> However, there is a middle ground but probably not necessary - it could be via adding a global setting for operators to decide if they want to enable such a feature for end users. (listCapabilities or other means can export this option to the UI to show/hide such a setting).

They can still do the same process using the API, so adding this to the UI would not make a difference in that case.

As for the global setting, I do agree that is a good option, however it would fall out of the scope of this PR. And I still believe we should make these changes in order to keep consistence between the UI and the API, regardless of the addition of a global setting.

DaanHoogland:
@GaOrtiga how do you want to move forward with this? (moving it to 4.20 for now)

yadvr (dismissed):
We can't accept this because (a) this was reverted after several service providers and users complaint that they don't want users to have rvr when the offering doesn't allow that by default, (b) isn't aligned with the original use-case. So, while the API may still allow for it, we don't necessarily want to expose that via UI.

rohityadavcloud:
@GaOrtiga the only way could be to introduce a feature flag in config.json (in UI) and have it disabled by default, when enabled you can pass the options in UI to show the elements. Could you review if this approach works for your use-case or close the PR. Thanks.

andrijapanicsb:
Agree with what Rohit said, this was enabled some releases ago, and it created issues for the Operators. It should be "impossible" by default, with the option to make it configured/enabled if the operator really needs it.

alexandremattioli:
-1 on adding this to the UI. Until the fundamental issues with VRRP are solved this feature should be avoided at all costs.

NuxRo:
> Agree with what Rohit said, this was enabled some releases ago, and it created issues for the Operators. It should be "impossible" by default, with the option to make it configured/enabled if the operator really needs it.

+1 what Andrija said. Not against having the feature, but let's make it depend on a global option or something, or a network offering like @borisstoyanov suggested.
We should the same feature that exists in a VPC as well to be honest.

weizhouapache:
> > Agree with what Rohit said, this was enabled some releases ago, and it created issues for the Operators. It should be "impossible" by default, with the option to make it configured/enabled if the operator really needs it.
> 
> +1 what Andrija said. Not against having the feature, but let's make it depend on a global option or something, or a network offering like @borisstoyanov suggested.

> We should the same feature that exists in a VPC as well to be honest.

"remove" after "should"? 

In my opinion,  redundant VR support is a very good feature,  it works pretty well,  if the environment has configured correctly and VRRP works without any issue.  

However,  considering the various network devices,  multiple upstream routers,  VMware switch or openvswitch,  etc.  I have to say network configuration is too complicated ..... 

I have the experience in the past that redundant VRs work well for some years but suddenly few (of many) networks have two MASTER VRs which lead to some downtime of user VMs.  The issue is not reproducible. After troubleshooting for some months,  the root cause was found, it was caused by a small configuration on one of the two core switches. 

Be careful to enable redundant VRs in production...

GaOrtiga:
Instead of a global setting or a flag in the config.json I propose that we wait for https://github.com/apache/cloudstack/pull/6918 to be merged and create a network scope setting for this feature, since it will give operators more flexibility to enable this to each network individually.

@DaanHoogland @borisstoyanov @weizhouapache @andrijapanicsb @rohityadavcloud @alexandremattioli @NuxRo

Do you have any concerns about this strategy, or can I advance with this implementation when the PR gets merged?

DaanHoogland:
> Instead of a global setting or a flag in the config.json I propose that we wait for #6918 to be merged and create a network scope setting for this feature, since it will give operators more flexibility to enable this to each network individually.
> 
> @DaanHoogland @borisstoyanov @weizhouapache @andrijapanicsb @rohityadavcloud @alexandremattioli @NuxRo
> 
> Do you have any concerns about this strategy, or can I advance with this implementation when the PR gets merged?

I think it is kind of strange to allow to change a network to redundant based on a setting of the network. How do you see that @GaOrtiga ?

DaanHoogland:
> > > Instead of a global setting or a flag in the config.json I propose that we wait for #6918 to be merged and create a network scope setting for this feature, since it will give operators more flexibility to enable this to each network individually.
> > > @DaanHoogland @borisstoyanov @weizhouapache @andrijapanicsb @rohityadavcloud @alexandremattioli @NuxRo
> > > Do you have any concerns about this strategy, or can I advance with this implementation when the PR gets merged?
> > 
> > 
> > I think it is kind of strange to allow to change a network to redundant based on a setting of the network. How do you see that @GaOrtiga ?
> 
> I believe it gives operators the flexibility to turn this feature on/off individually for each user, while keeping the current behaviour for the operators that do not want to use it, so I see it as a positive implementation. However, I am open to feedback.

The strange thing, to me is that if it is a setting of the network, the operator would have to inable it for each network that a user creates. instead of at a higher level (zone or domain or some other enitity). I would think if the user requests it you just make it redundant, no further action.

rohityadavcloud:
A global setting or UI flag based (via config.json) option is okay but then we should make this feature opt-in, and disabled by default. Admins/operators who want this feature can enable it for the UI.

JoaoJandre:
> @DaanHoogland @borisstoyanov @weizhouapache @andrijapanicsb @rohityadavcloud @alexandremattioli @NuxRo
> 
> I see that the comunity has mixed opinions regarding this functionality, however, I still believe we should maintain conformity on the API and the UI. In order to achieve this without raising further discussions, I propose that we create a global configuration with default value `false` that dictates whether this parameter can be used (both on the UI and the API). If set to false, it will not show up on the UI and trying to use it on the API will generate an exception.
> 
> This would give operators the flexibility to remove this feature from the backend if they see it as harmful, or leave it both on the backend and the frontend if they see it as helpful.
> 
> Would this implementation be enough to offset your concerns regarding this feature?

@GaOrtiga The problem with the proposed solution is that this is **breaking backwards compatibility**. When users update to 4.20, they'll start to get an exception when informing this parameter on the API. This kind of change should only be made when changing major versions. Last month I made a proposal for a release schedule that would let us do this kind of change (see https://lists.apache.org/thread/o6o9h3qp8gqrpq4v7o81tl6vp51tkjhg) but sadly it got almost no traction; there is a new discussion thread about it now (https://github.com/apache/cloudstack/discussions/8970), I hope we can advance it this time. As the project stands right now, this proposal should not be accepted.

sureshanaparti:
@blueorangutan package

DaanHoogland:
@JoaoJandre do you still want this (behind a feature flag)?

JoaoJandre:
> @JoaoJandre do you still want this (behind a feature flag)?

As long as the default value for the config is **true**, to maintain the same behavior as the older versions, I'm fine with it.

DaanHoogland:
> > @JoaoJandre do you still want this (behind a feature flag)?
> 
> As long as the default value for the config is **true**, to maintain the same behavior as the older versions, I'm fine with it.

@GaOrtiga will you implement such a flag?

GaOrtiga:
@DaanHoogland I added the config

DaanHoogland:
> @DaanHoogland I added the config

thanks @GaOrtiga . @andrijapanicsb @yadvr , can you concur with this solution?

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -2986,8 +2989,13 @@ public boolean restartNetwork(RestartNetworkCmd cmd) throws ConcurrentOperationE
             throwInvalidIdException("Cannot restart a VPC tier with cleanup, please restart the whole VPC.", network.getUuid(), "network tier");
         }
         boolean makeRedundant = cmd.getMakeRedundant();
-        boolean livePatch = cmd.getLivePatch();
         User callerUser = _accountMgr.getActiveUser(CallContext.current().getCallingUserId());
+        if (makeRedundant && !_accountMgr.isRootAdmin(callerUser.getAccountId()) && !AllowUsersToMakeNetworksRedundant.value() ) {
+            throw new InvalidParameterValueException(String.format("Could not make network redundant as calling user is not Root Admin and [%s] is set to false.",
```
sureshanaparti:
```suggestion
            throw new InvalidParameterValueException("Could not make the network redundant. Please contact administrator.");
```

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -2986,8 +2989,13 @@ public boolean restartNetwork(RestartNetworkCmd cmd) throws ConcurrentOperationE
...
+                    AllowUsersToMakeNetworksRedundant.key()));
```
sureshanaparti:
```suggestion
```

DaanHoogland: @sureshanaparti , this is a syntax error. I think I know what you mean, but let’s not apply this.

rajujith:
@GaOrtiga Since this is for the 4.22.1 release, could you retarget the PR to the 4.22 branch?

DaanHoogland:
@rajujith , this is kind of a hot topic. There is still no consensus on what the default value of the configuration will be and we are at a stalemate. I am not sure if this should be part of 22.1.

GaOrtiga:
> > @DaanHoogland I added the config
> 
> thanks @GaOrtiga . @andrijapanicsb @yadvr , can you concur with this solution?

@andrijapanicsb @yadvr

yadvr:
If we must really have it - add some kind of UI config or global setting with the default to not show the option. This is because many MSPs don't want their tenants to be able to spin up an additional VR (i.e. redundant VR).

yadvr:
Looks like I'm tagged to review something which is hidden in several other messages I cannot see - take my last comment with a pinch of salt, if such a UI config is already there +1 to the PR.

NuxRo:
> If we must really have it - add some kind of UI config or global setting with the default to not show the option. This is because many MSPs don't want their tenants to be able to spin up an additional VR (i.e. redundant VR).

+1 that.

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -280,6 +280,9 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService, C
     private static final ConfigKey<Boolean> AllowEmptyStartEndIpAddress = new ConfigKey<>("Advanced", Boolean.class,
             "allow.empty.start.end.ipaddress", "true", "Allow creating network without mentioning start and end IP address",
             true, ConfigKey.Scope.Account);
+    public static final ConfigKey<Boolean> AllowUsersToMakeNetworksRedundant = new ConfigKey<>("Advanced", Boolean.class,
+            "allow.users.to.make.networks.redundant", "true", "Allow Users to make Networks Redundant",
+            true, ConfigKey.Scope.Global);
```
DaanHoogland, line 285:
```suggestion
    public static final ConfigKey<Boolean> AllowUsersToMakeNetworksRedundant = new ConfigKey<>("Advanced", Boolean.class,
            "allow.users.to.make.networks.redundant", "false", "Allow Users to make Networks Redundant",
            true, ConfigKey.Scope.Global);
```

as per @yadvr ’s comment: https://github.com/apache/cloudstack/pull/7405#issuecomment-3971991901 that requests the default to be false.

DaanHoogland:
@GaOrtiga , are you alright with changing the default?

andrijapanicsb:
+1 to make it false by default, in the global settings - this is just messing people up - end those who have unstable physical networking = it happens 2 Vrs can not talk via VRRP amd then both promote themself to MASTER or BACKUP - either way, not good

GaOrtiga:
> @GaOrtiga , are you alright with changing the default?

@DaanHoogland 
I personally would, I don't have a personal opinion on which should be the default. However, it would change the current behaviour of the API, which is not ideal.

> Looks like I'm tagged to review something which is hidden in several other messages I cannot see - take my last comment with a pinch of salt, if such a UI config is already there +1 to the PR.

I have done as @yadvr mentioned in this comment, added a UI flag, with default value false, making the field hidden by default. I will leave the backend configuration, but will not change the default value in this PR (if anyone wishes to, they can open a PR to change it in the future).

This approach maintains current behaviour, both on the GUI and the API and adds the possibility for the operator to add the field if they want, or leave it hidden if they don't.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

PR #7441, opened 2023-04-17
Use max secondary storage defined on the account during upload 

JoaoJandre:
### Description

Currently ACS uses the `max.account.secondary.storage` setting to set the initial secondary storage limit for accounts. However, even after changing the limit for an account, the configuration value is still used to validate the consumption of secondary storage during the upload of volumes, templates and ISOs.
This PR makes the limit defined in the account be used in the validation during the upload, instead of the configuration value.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
A user account was created and configured with a max secondary storage limit of 2 GiB, while the `max.account.secondary.storage` configuration was set to 200GiB. 
I uploaded two ISOs of 1GiB, and when I tried to upload a third ISO an error of max limit reached was received (which was expected).
I set the `max.account.secondary.storage` to 2GiB and set the accounts max secondary storage to 200GiB, then I retried to upload the ISO and everything went ok.


rohityadavcloud (commented):
Didn't test or reviewed it, but agree we should definitely do such checks. (pre/post)

DaanHoogland (commented):
code looks good, but we have a functional change here.
Before the global setting was a cap on the usage. Now it is just a default. I don´t mind either way but backwards compatibilty dictates that the cap should be lowest of the local and the global configuration.
What do you think @JoaoJandre ?

JoaoJandre:
> code looks good, but we have a functional change here. Before the global setting was a cap on the usage. Now it is just a default. I don´t mind either way but backwards compatibilty dictates that the cap should be lowest of the local and the global configuration. What do you think @JoaoJandre ?

For the other limits, that is how it currently works, the most specific setting is the one that prevails. For user VMs for example, if you have the global setting at 1 and the account limit at 2, that account will be able to create two accounts. In the VM creation process, the `getCheckedUserVmResource` method will be called, which eventually will call the `findCorrectResourceLimitForAccount` method to get the account's limit, which is the same as I propose to use with the secondary storage.

stephankruggg (approved):
LGTM, manually tested

weizhouapache (approved):
code lgtm


@JoaoJandre 
can you rebase with 4.18 and change the target branch to 4.18 ?

weizhouapache:
> @weizhouapache sorry for the delay, but now it's done :)

great, thanks

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -477,13 +477,12 @@ public GetUploadParamsResponse doInTransaction(TransactionStatus status) throws
...
-                if (account.getType().equals(Account.Type.PROJECT)) {
-                    command.setDefaultMaxSecondaryStorageInGB(ResourceLimitService.MaxProjectSecondaryStorage.value());
-                } else {
-                    command.setDefaultMaxSecondaryStorageInGB(ResourceLimitService.MaxAccountSecondaryStorage.value());
-                }
+
+                Long accountId = vol.getAccountId();
+                Account account = _accountDao.findById(accountId);
+
+                command.setDefaultMaxSecondaryStorageInGB(_resourceLimitMgr.findCorrectResourceLimitForAccount(account, ResourceType.secondary_storage));
```
weizhouapache, line 484: @JoaoJandre 
I think the domain resource limitation should be considered as well, as the account resource limitation might exceed the domain resource limitation.

JoaoJandre: @weizhouapache I did some testing and confirmed this, also tested to see if this kind of check is done for other things, like number of vm instances, and the domain limit is considered in that case. So I added the domain resource limitation check to the secondary storage as well.

kiranchavala:
@blueorangutan package

kiranchavala (commented):
@JoaoJandre @weizhouapache @DaanHoogland 



Under the following scenario, the account limits are not honored

Steps

1. Create a user role account
2. Set the Max. secondary storage (GiB) to 2 gb at the account level 

Screenshot 2023-08-01 at 2 40 09 PM

3. Login to the user role account
4.  Register a iso of size (1.8 gb)  from a remote webserver
5.  Upload a iso of size (1.8 gb) from Local Desktop
6.  The upload of iso is successful
7. The account limit after the upload is 
Screenshot 2023-08-01 at 2 39 35 PM


**Ideally the upload of the iso should fail as the account limit reached** 

------
Normal upload of iso is failing if the account limit is reached and its the expected behavior 

1. Uploaded a ISOs of 1GiB,
2. Upload a second iso of another 1gb
3. Upload of third iso  and it fails 


Screenshot 2023-08-01 at 2 18 41 PM

JoaoJandre:
> @JoaoJandre @weizhouapache @DaanHoogland
> 
> Under the following scenario, the account limits are not honored
> 
> Steps
> 
>     1. Create a user role account
> 
>     2. Set the Max. secondary storage (GiB) to 2 gb at the account level
> 
> 
> Screenshot 2023-08-01 at 2 40 09 PM
> 
>     3. Login to the user role account
> 
>     4. Register a iso of size (1.8 gb)  from a remote webserver
> 
>     5. Upload a iso of size (1.8 gb) from Local Desktop
> 
>     6. The upload of iso is successful
> 
>     7. The account limit after the upload is
> 
> 
> Screenshot 2023-08-01 at 2 39 35 PM
> 
> **Ideally the upload of the iso should fail as the account limit reached**
> 
> Normal upload of iso is failing if the account limit is reached and its the expected behavior
> 
>     1. Uploaded a ISOs of 1GiB,
> 
>     2. Upload a second iso of another 1gb
> 
>     3. Upload of third iso  and it fails
> 
> 
> Screenshot 2023-08-01 at 2 18 41 PM


---
This PR deals only with upload from local. Therefore problems with registering ISOs are out of scope for this PR.

kiranchavala:
> > @JoaoJandre @weizhouapache @DaanHoogland
> > Under the following scenario, the account limits are not honored
> > Steps
> > ```
> > 1. Create a user role account
> > 
> > 2. Set the Max. secondary storage (GiB) to 2 gb at the account level
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Screenshot 2023-08-01 at 2 40 09 PM
> > ```
> > 3. Login to the user role account
> > 
> > 4. Register a iso of size (1.8 gb)  from a remote webserver
> > 
> > 5. Upload a iso of size (1.8 gb) from Local Desktop
> > 
> > 6. The upload of iso is successful
> > 
> > 7. The account limit after the upload is
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Screenshot 2023-08-01 at 2 39 35 PM
> > **Ideally the upload of the iso should fail as the account limit reached**
> > Normal upload of iso is failing if the account limit is reached and its the expected behavior
> > ```
> > 1. Uploaded a ISOs of 1GiB,
> > 
> > 2. Upload a second iso of another 1gb
> > 
> > 3. Upload of third iso  and it fails
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Screenshot 2023-08-01 at 2 18 41 PM
> 
> This PR deals only with upload from local. Therefore problems with registering ISOs are out of scope for this PR.

@JoaoJandre  thanks for the update, I have created a separate issue #7839 to track it

weizhouapache:
thanks for the testing @kiranchavala 

merging

PR #7446, opened 2023-04-18
Move subdomains between domains

JoaoJandre:
### Description

Sometimes users have the need to move resources between domains, for example, in a big company, a department may be moved from one part of the company to another, changing the company's department hierarchy, the easiest way of reflecting this change on the company's cloud environment would be to move subdomains between domains, but currently ACS offers no option to do that. 
This PR adds the `moveDomain` API,  which will move domains between subdomains. Furthermore, if the domain that is being moved has any subdomains, those will also be moved, maintaining the current subdomain tree.
### Types of changes

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

## Description of tests

| Test | Result |
| ------ | ------ |
| Move an Inactive Domain | Error, as an inactive domain should not be able to be moved |
| Move a Domain to a new inactive parent domain | Error, as a domain that  is inactive should not receive any new children|
| Move a domain to a new parent domain that has a child with the same name as the domain to be moved | Error, as two subdomains of the same domain should not have the same name |
| Move Domain ROOT | Error, root domain by definition must always be the father of all other domains|
| Move a domain that is already on the destination | Error, as it is already is were it should be |
| Move a domain to itself (moved domain = new parent domain) | Error, as this makes no sense |
| Move a domain that uses a shared network that is inaccessible to the new parent domain | Error, as the new parent domain must have access to all of its children's resources |
| Move a domain that has a child subdomain that uses a shared network that is inaccessible to the new parent domain | Error, as the new parent domain must have access to all of its children's resources |
| Move a domain that uses a affinity group that is inaccessible to the new parent domain | Error, as the new parent domain must have access to all of its children's resources |
| Move a domain that has a child subdomain that uses a affinity group that is inaccessible to the new parent domain | Error,  as the new parent domain must have access to all of its children's resources|
| Move a domain that owns a shared network | Success |
| Move a domain that has a subdomain that owns a shared network | Success |
| Move a domain that uses a shared network that is accessible to the new parent domain | Success |
| Move a domain that has a subdomain that uses a shared network that is accessible to the new parent domain | Success |
| Move a domain that owns an affinity group | Success |
| Move a domain that has a subdomain that owns an affinity group | Success |
| Move a domain that uses a affinity group that is accessible to the new parent domain | Success |
| Move a domain that has a subdomain that uses a affinity group that is accessible to the new parent domain | Success |
| Move a domain that exceeds the resource limits of the new parent domain | Success |

All tests performed as expected. In the case of shared networks, all had subdomain access.

For successful cases, it was checked whether the following attributes of the moved domain and its children were correctly changed:
- Path
- Level
- Use of resources
- Limits
- Parent domain

In addition, it was verified that the number of children of the current parent domain and the target domain were correctly accounted for.


rohityadavcloud:
Thanks for the PR @JoaoJandre interesting use-case. Do you think there should also be some kind of domain-limits check (for ex. moving sub-domain1 to a new parent could mean different limits on vms, volume, cpu, templates, storage etc. need to be recalculated/checked)? Also side-effects in (already produced) usage data and networks?

JoaoJandre:
> Thanks for the PR @JoaoJandre interesting use-case. Do you think there should also be some kind of domain-limits check (for ex. moving sub-domain1 to a new parent could mean different limits on vms, volume, cpu, templates, storage etc. need to be recalculated/checked)? Also side-effects in (already produced) usage data and networks?

Hey @rohityadavcloud, thanks for the suggestion, I actually agree with you, there should be some kind of domain-limits check. I will put this in draft until I have the time to add this check.

About the side-effects in usage data, there shouldn't be any, as the usage data only uses the domain ID, and that is not changed when the domain is moved, only the domain path is changed. About networks, we already validate shared networks before moving the domain, what other side-effects could be an issue with networks?

JoaoJandre:
@blueorangutan package

DaanHoogland:
@rohityadavcloud @JoaoJandre was this extra consideration of quota limits addressed here?

> > Thanks for the PR @JoaoJandre interesting use-case. Do you think there should also be some kind of domain-limits check (for ex. moving sub-domain1 to a new parent could mean different limits on vms, volume, cpu, templates, storage etc. need to be recalculated/checked)? Also side-effects in (already produced) usage data and networks?
> 
> Hey @rohityadavcloud, thanks for the suggestion, I actually agree with you, there should be some kind of domain-limits check. I will put this in draft until I have the time to add this check.
> 
> About the side-effects in usage data, there shouldn't be any, as the usage data only uses the domain ID, and that is not changed when the domain is moved, only the domain path is changed. About networks, we already validate shared networks before moving the domain, what other side-effects could be an issue with networks?

JoaoJandre:
> @rohityadavcloud @JoaoJandre was this extra consideration of quota limits addressed here?
> 
> > > Thanks for the PR @JoaoJandre interesting use-case. Do you think there should also be some kind of domain-limits check (for ex. moving sub-domain1 to a new parent could mean different limits on vms, volume, cpu, templates, storage etc. need to be recalculated/checked)? Also side-effects in (already produced) usage data and networks?
> > 
> > 
> > Hey @rohityadavcloud, thanks for the suggestion, I actually agree with you, there should be some kind of domain-limits check. I will put this in draft until I have the time to add this check.
> > About the side-effects in usage data, there shouldn't be any, as the usage data only uses the domain ID, and that is not changed when the domain is moved, only the domain path is changed. About networks, we already validate shared networks before moving the domain, what other side-effects could be an issue with networks?

@DaanHoogland yes I addressed it on commit 88d77658cb9e66e339495d0bd9af9d90041e9674, I just forgot to make a comment about it.

DaanHoogland:
@blueorangutan test

DaanHoogland (dismissed):
code looks good

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

DaanHoogland:
@blueorangutan LLtest

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

api/src/main/java/com/cloud/user/DomainService.java
```diff
@@ -18,12 +18,14 @@
 
 import java.util.List;
 
+import com.cloud.exception.ResourceAllocationException;
```
harikrishna-patnala, line 21: reorder these imports please

server/src/main/java/com/cloud/user/DomainManagerImpl.java
```diff
@@ -900,4 +912,171 @@ private void updateDomainChildren(DomainVO domain, String updatedDomainPrefix) {
         }
     }
 
+    @Override
+    @ActionEvent(eventType = EventTypes.EVENT_DOMAIN_MOVE, eventDescription = "moving Domain")
+    @DB
+    public Domain moveDomainAndChildrenToNewParentDomain(MoveDomainCmd cmd) throws ResourceAllocationException {
+        Long idOfDomainToBeMoved = cmd.getDomainId();
+        Long idOfNewParentDomain = cmd.getParentDomainId();
+
```
harikrishna-patnala, line 921: I think we also need to check the access permissions something like

        Account caller = getCaller();
        _accountMgr.checkAccess(caller, domain);

and

       _accountMgr.checkAccess(caller, parentDomain);

harikrishna-patnala (commented):
Thanks for the PR @JoaoJandre nice work.

I've given some comments, can you please check.

api/src/main/java/org/apache/cloudstack/api/command/admin/domain/MoveDomainCmd.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.cloudstack.api.ApiConstants;
+import org.apache.cloudstack.api.ApiErrorCode;
+import org.apache.cloudstack.api.BaseCmd;
+import org.apache.cloudstack.api.Parameter;
+import org.apache.cloudstack.api.ServerApiException;
+import org.apache.cloudstack.api.response.DomainResponse;
+
+import java.util.Objects;
+
+@APICommand(name = "moveDomain", description = "Moves a domain and its children to a new parent domain.", responseObject = DomainResponse.class,
```
harikrishna-patnala, line 33: Can you please add "Since" parameter here

DaanHoogland:
@JoaoJandre will you address @harikrishna-patnala 's comments?

JoaoJandre:
@DaanHoogland sure, I've been a bit busy lately, but I will address the comments as soon as I can  :)

api/src/main/java/org/apache/cloudstack/api/command/admin/domain/MoveDomainCmd.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Parameter(name = ApiConstants.PARENT_DOMAIN_ID, type = CommandType.UUID, entityType = DomainResponse.class,
+            description = "The ID of the new parent domain of the domain to be moved. If no parent domain is specified, the ROOT domain is assumed.")
+    private Long parentDomainId;
+
+    public Long getDomainId() {
+        return domainId;
+    }
+
+    public Long getParentDomainId() {
```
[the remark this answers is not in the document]
JoaoJandre: I agree, I'll change this

rajujith (commented):
@JoaoJandre 
I tested the PR, the feature works in general, and it's good. However, I encountered the below scenarios that need to be reviewed and discussed. 

1. Moving a domain to a parent domain with no resource limit i.e unlimited doesn't work. There appears to be something wrong with the implementation. 

```
(MoveDomainPR) 🐱 > move domain domainid=010957cd-4388-4ba0-a18d-d5df0814ad06 parentdomainid=41928b19-adf8-4524-8ca6-16d395a1136c
🙈 Error: (HTTP 535, error code 4370) Cannot move domain [010957cd-4388-4ba0-a18d-d5df0814ad06] to parent domain [41928b19-adf8-4524-8ca6-16d395a1136c] as maximum domain resource limit of type [user_vm] would be exceeded. The current resource count for domain [010957cd-4388-4ba0-a18d-d5df0814ad06] is [0], the resource count for the new parent domain [41928b19-adf8-4524-8ca6-16d395a1136c] is [0], and the limit is [-1].

```

`https://github.com/scclouds/cloudstack/blob/029164cb1d33b92ece6a9557fb514b490988d1a0/server/src/main/java/com/cloud/user/DomainManagerImpl.java#L981C12-L981C108`


2. I am able to move a domain while there are resources deployed on a cluster dedicated to the previous parent domain. Further start operation and new deployment fails as it is the only cluster in the zone. This doesn't seem ideal. 
3. Move with instance created from computer offering dedicated to parent domain : This is currently allowed but not sure whether it should be allowed. On Clicking the Compute offering (from instance details) after moving the domain we get 404. 
4. Move domain with instances deployed to network created from network offering dedicated to parent domain: This is currently allowed but not sure whether it should be allowed. On Clicking the Network offering (from network details) after moving the domain we get 404. 

cc @harikrishna-patnala

JoaoJandre:
> @JoaoJandre I tested the PR, the feature works in general, and it's good. However, I encountered the below scenarios that need to be reviewed and discussed.
> 
>     1. Moving a domain to a parent domain with no resource limit i.e unlimited doesn't work. There appears to be something wrong with the implementation.
> 
> 
> ```
> (MoveDomainPR) 🐱 > move domain domainid=010957cd-4388-4ba0-a18d-d5df0814ad06 parentdomainid=41928b19-adf8-4524-8ca6-16d395a1136c
> 🙈 Error: (HTTP 535, error code 4370) Cannot move domain [010957cd-4388-4ba0-a18d-d5df0814ad06] to parent domain [41928b19-adf8-4524-8ca6-16d395a1136c] as maximum domain resource limit of type [user_vm] would be exceeded. The current resource count for domain [010957cd-4388-4ba0-a18d-d5df0814ad06] is [0], the resource count for the new parent domain [41928b19-adf8-4524-8ca6-16d395a1136c] is [0], and the limit is [-1].
> ```
> 
> `https://github.com/scclouds/cloudstack/blob/029164cb1d33b92ece6a9557fb514b490988d1a0/server/src/main/java/com/cloud/user/DomainManagerImpl.java#L981C12-L981C108`
> 
>     2. I am able to move a domain while there are resources deployed on a cluster dedicated to the previous parent domain. Further start operation and new deployment fails as it is the only cluster in the zone. This doesn't seem ideal.
> 
>     3. Move with instance created from computer offering dedicated to parent domain : This is currently allowed but not sure whether it should be allowed. On Clicking the Compute offering (from instance details) after moving the domain we get 404.
> 
>     4. Move domain with instances deployed to network created from network offering dedicated to parent domain: This is currently allowed but not sure whether it should be allowed. On Clicking the Network offering (from network details) after moving the domain we get 404.
> 
> 
> cc @harikrishna-patnala

@rajujith thanks for testing, I'll try to address your comments next week.

shwstppr:
@JoaoJandre any update on this?

JoaoJandre:
> @JoaoJandre any update on this?

I've addressed some issues already, but some are a bit more complex. I'm working on it.

JoaoJandre:
@DaanHoogland @rajujith @shwstppr @harikrishna-patnala I've addressed the comments/test failures.

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test

engine/schema/src/main/java/com/cloud/network/dao/NetworkDomainDaoImpl.java
```diff
@@ -71,4 +92,34 @@ public List<Long> listNetworkIdsByDomain(long domainId) {
...
+                    List<String> vmUuidsAndNetworkUuids = Arrays.asList(rs.getString(2).split(","));
+
+                    domainsOfSharedNetworksUsedByDomainPath.put(domainId, vmUuidsAndNetworkUuids);
+                }
+            }
+
+            return domainsOfSharedNetworksUsedByDomainPath;
+        } catch (SQLException e) {
+            logger.error(String.format("Failed to retrieve the domains of the shared networks with subdomain access used by domain with path [%s] due to [%s]. Returning an empty "
+                    + "list of domains.", domainPath, e.getMessage()), e);
```
DaanHoogland, line 120:
```suggestion
            logger.error(String.format("Failed to retrieve the domains of the shared networks with subdomain access used by domain with path [%s] due to [%s]. Returning an empty "
                    + "list of domains.", domainPath, e.getMessage()));
            if (logger.isDebugEnabled()) {
                logger.debug(String.format("Failed to retrieve the domains of the shared networks with subdomain access used by domain with path [%s]. Returning an empty "
                        + "list of domains.", domainPath), e);
            }
```

engine/schema/src/main/java/org/apache/cloudstack/affinity/dao/AffinityGroupDomainMapDaoImpl.java
```diff
@@ -62,4 +85,35 @@ public List<AffinityGroupDomainMapVO> listByDomain(Object... domainId) {
...
+                    List<String> vmUuidsAndAffinityGroupUuids = Arrays.asList(rs.getString(2).split(","));
+
+                    domainsOfAffinityGroupsUsedByDomainPath.put(domainId, vmUuidsAndAffinityGroupUuids);
+                }
+            }
+
+            return domainsOfAffinityGroupsUsedByDomainPath;
+        } catch (SQLException e) {
+            logger.error(String.format("Failed to retrieve the domains of the affinity groups with subdomain access used by domain with path [%s] due to [%s]. Returning an " +
+                    "empty list of domains.", domainPath, e.getMessage()), e);
```
DaanHoogland, line 113:
```suggestion
            logger.error(String.format("Failed to retrieve the domains of the affinity groups with subdomain access used by domain with path [%s] due to [%s]. Returning an " +
                    "empty list of domains.", domainPath, e.getMessage()));
            if (logger.isDebugEnabled()) {
                logger.debug(String.format("Failed to retrieve the domains of the affinity groups with subdomain access used by domain with path [%s]. Returning an " +
                        "empty list of domains.", domainPath), e);
            }
```

server/src/main/java/com/cloud/api/query/dao/NetworkOfferingJoinDaoImpl.java
```diff
@@ -120,4 +138,36 @@ public NetworkOfferingJoinVO newNetworkOfferingView(NetworkOffering offering) {
...
+                    List<String> networkUuids = Arrays.asList(rs.getString(2).split(","));
+
+                    domainsOfNetworkOfferingsUsedByDomainPath.put(domainId, networkUuids);
+                }
+            }
+
+            return domainsOfNetworkOfferingsUsedByDomainPath;
+        } catch (SQLException e) {
+            s_logger.error(String.format("Failed to retrieve the domains of the network offerings used by domain with path [%s] due to [%s]. Returning an empty "
+                    + "list of domains.", domainPath, e.getMessage()), e);
```
DaanHoogland, line 168:
```suggestion
            s_logger.error(String.format("Failed to retrieve the domains of the network offerings used by domain with path [%s] due to [%s]. Returning an empty "
                    + "list of domains.", domainPath, e.getMessage()));
            if (s_logger.isDebugEnabled()) {
                 s_logger.debug(String.format("Failed to retrieve the domains of the network offerings used by domain with path [%s]. Returning an empty "
                         + "list of domains.", domainPath), e);
            }
```

server/src/main/java/com/cloud/api/query/dao/ServiceOfferingJoinDaoImpl.java
```diff
@@ -165,4 +183,35 @@ public ServiceOfferingJoinVO newServiceOfferingView(ServiceOffering offering) {
...
+                    List<String> vmUuids = Arrays.asList(rs.getString(2).split(","));
+
+                    domainsOfServiceOfferingsUsedByDomainPath.put(domainId, vmUuids);
+                }
+            }
+
+            return domainsOfServiceOfferingsUsedByDomainPath;
+        } catch (SQLException e) {
+            s_logger.error(String.format("Failed to retrieve the domains of the service offerings used by domain with path [%s] due to [%s]. Returning an empty "
+                    + "list of domains.", domainPath, e.getMessage()), e);
```
DaanHoogland, line 212:
```suggestion
            s_logger.error(String.format("Failed to retrieve the domains of the service offerings used by domain with path [%s] due to [%s]. Returning an empty "
                    + "list of domains.", domainPath, e.getMessage()));
            if (s_logger.isDebugEnabled()) {
                s_logger.debug(String.format("Failed to retrieve the domains of the service offerings used by domain with path [%s]. Returning an empty "
                        + "list of domains.", domainPath), e);
            }
```

server/src/main/java/com/cloud/dc/dao/DedicatedResourceDaoImpl.java
```diff
@@ -428,4 +447,35 @@ public List<Long> findHostsByZone(Long zoneId) {
...
+                    List<String> vmUuids = Arrays.asList(rs.getString(2).split(","));
+
+                    domainsOfDedicatedResourcesUsedByDomainPath.put(domainId, vmUuids);
+                }
+            }
+
+            return domainsOfDedicatedResourcesUsedByDomainPath;
+        } catch (SQLException e) {
+            LOGGER.error(String.format("Failed to retrieve the domains of the dedicated resources used by domain with path [%s] due to [%s]. Returning an empty "
+                    + "list of domains.", domainPath, e.getMessage()), e);
```
DaanHoogland, line 476:
```suggestion
            LOGGER.error(String.format("Failed to retrieve the domains of the dedicated resources used by domain with path [%s] due to [%s]. Returning an empty "
                    + "list of domains.", domainPath, e.getMessage()));
            if (LOGGER.isDebugEnabled()) {
                LOGGER.debug(String.format("Failed to retrieve the domains of the dedicated resources used by domain with path [%s]. Returning an empty "
                        + "list of domains.", domainPath), e);
            }
```

DaanHoogland:
I recently encoutered a similar issue and I am actually -1 on this without my sugestion because of that. I think you even agreed with me on that ticket,  @JoaoJandre. we need to allow users to minimise logging by using a lesser level.
cc @shwstppr

shwstppr:
@blueorangutan package

JoaoJandre:
> @JoaoJandre @DaanHoogland any consensus here? Are all issues/comments raised by @rajujith already addressed? [#7446 (review)](https://github.com/apache/cloudstack/pull/7446#pullrequestreview-1724439749)

@shwstppr yes, all the comments raised by @rajujith were addressed.



> I recently encoutered a similar issue and I am actually -1 on this without my sugestion because of that. I think you even agreed with me on that ticket, @JoaoJandre. we need to allow users to minimise logging by using a lesser level. cc @shwstppr
> 
> ad. #8278

@DaanHoogland I thought a little more about it and in this instance I don't see a problem with this approach. But I don't think that this should always be the case (only have stacktraces on debug or lower). I'll make a post on the discussion thread later.

DaanHoogland:
@harikrishna-patnala @rajujith can you approve on the current state of this PR?

shwstppr:
ping @harikrishna-patnala @rajujith

shwstppr:
Thanks for the testing @rajujith 
@JoaoJandre is improving the error message something that can be done in a separate PR?

shwstppr:
@blueorangutan test

PR #7453, opened 2023-04-20
Support local storage live migration for direct download templates

mlsorensen:
### Description

This PR fixes live local storage migration for VMs created with "Direct Download" images. 

Currently, the local storage driver has no special handling for direct download templates during live storage migration. The workflow is something like 1) check to see if root volume's backing template exists on destination primary storage 2) if not, then send a CopyCommand to copy template to destination storage 3) create new destination volume on destination primary storage against the backing template. 4) send destination volume info during MigrateCommand so we can rewrite the volume path in the VM XML during migration 5) migrate VM.  In step 2 - CopyCommand is not the right thing to do for putting a "Direct Download" template onto primary storage, it should instead be a type of DirectDownloadCommand.  

Additionally, Direct Downloads don't use the template ID as the name of the template file on primary, they use a dynamically generated name, so the template paths are different between source and destination pools. Thus we need to pass the new template path along in MigrateCommand so we can replace this backing store path in the VM XML during live migration.

This patch updates the KvmNonManagedStorageDataMotionStrategy used during local storage migration to handle staging of DirectDownload templates via a DirectDownloadCommand instead of a CopyCommand, and also provides the backing store path and relevant XML substitution during MigrateCommand handling.

Additional testing would be appreciated for workflows involving the KVMNonManagedStorageDataMotionStrategy, though I suspect the scope is fairly limited to live storage migration of Filesystem and NFS storage types. Since this class extends the StorageSystemDataMotionStrategy, some minor changes were made to it as well but nothing that should change its behavior.

I haven't tested migration of Direct Download VMs while stopped yet, that may need a separate/similar PR.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested in local virtual environment with local storage KVM hypervisors, first reproduced the problem and then fixed it.

### DIFF
--- a/core/src/main/java/com/cloud/agent/api/MigrateCommand.java
+++ b/core/src/main/java/com/cloud/agent/api/MigrateCommand.java
@@ -171,6 +171,7 @@ public class MigrateCommand extends Command {
         private final DriverType driverType;
         private final Source source;
         private final String sourceText;
+        private final String backingStoreText;
         private boolean isSourceDiskOnStorageFileSystem;
 
         public MigrateDiskInfo(final String serialNumber, final DiskType diskType, final DriverType driverType, final Source source, final String sourceText) {
@@ -179,6 +180,16 @@ public class MigrateCommand extends Command {
             this.driverType = driverType;
             this.source = source;
             this.sourceText = sourceText;
+            this.backingStoreText = null;
+        }
+
+        public MigrateDiskInfo(final String serialNumber, final DiskType diskType, final DriverType driverType, final Source source, final String sourceText, final String backingStoreText) {
+            this.serialNumber = serialNumber;
+            this.diskType = diskType;
+            this.driverType = driverType;
+            this.source = source;
+            this.sourceText = sourceText;
+            this.backingStoreText = backingStoreText;
         }
 
         public String getSerialNumber() {
@@ -200,6 +211,8 @@ public class MigrateCommand extends Command {
         public String getSourceText() {
             return sourceText;
         }
+
+        public String getBackingStoreText() { return backingStoreText; }
 
         public boolean isSourceDiskOnStorageFileSystem() {
             return isSourceDiskOnStorageFileSystem;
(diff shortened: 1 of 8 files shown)


rohityadavcloud:
@blueorangutan package

davidjumani:
@blueorangutan test

davidjumani:
Thanks @mlsorensen for raising this. I'm unable to reproduce the issue but have manually tested live migration with local storage on direct download templates, as well as with secondary storage and it works as expected. Smoke tests are also passing

davidjumani (approved):

mlsorensen:
> Thanks @mlsorensen for raising this. I'm unable to reproduce the issue but have manually tested live migration with local storage on direct download templates, as well as with secondary storage and it works as expected. Smoke tests are also passing

Thanks @davidjumani This concerns me, I don't want to make the change if it is working for you - however I don't see how it is possible that it is working just based on reading the code.  One way it may work is if the direct download template has already been used on destination storage before an attempt for VM live migration. I'd like to get to the bottom of this before we merge.

mlsorensen:
@davidjumani I was able to get the live migration to work with local storage, however there are a few steps that silently error, and the result is that the VM after moving is not linked to a template, it is its own full copy. These steps that silently fail revolve around staging the template to destination storage which are addressed in this patch.

I suspect if you look at your test setup, you may also see errors and also see that post migration there is no longer a backing file on your VM's root disk, even though the migration succeeded.

shwstppr (approved):

rohityadavcloud:
@nvazquez can you review/test this? If this is ready for merge, pl go ahead. Thanks.

nvazquez (approved):
LGTM. Tested OK on 3x KVM on Rocky8 env with local storage.

Verified that before the fix the following was broken, and after its fixed:
- Deploy VM on host A with local storage offering
- Live migrate VM to host B
- Verify the local pool on host B contains a copy of the template
- Deploy a new VM on host B

rohityadavcloud:
This is ready for merging but @mlsorensen can you rebase (or change) base of your PR to 4.18. It would benefit to have the fix in 4.18 branch and fwd-merged into main. Thanks.

nvazquez:
Hi @mlsorensen can you please rebase to branch 4.18?

nvazquez:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

PR #7477, opened 2023-04-26
Send console text slower to avoid overloading remote keyboard buffer

mlsorensen:
### Description

This PR slows down the console paste text function in the console ever so slightly, adding 25ms between each character. It was found that when large text is pasted, say an SSH key or something over 100 characters, it would stop somewhere in the 100-200 character range, and if the last character was capitalized it would be stuck with left shift held down. 

In debugging, I ran noVNC locally with websockify against a Qemu VNC and took the whole console proxy out of the equation and was still able to reproduce it. I traced the websocket packets and we were clearly firing off all of the keypresses just fine, but the ones on the end were getting dropped VNC server side. It seems we are overloading a keyboard buffer, or something along those lines when we send all of the keypress messages at once. It was also observed that we were able to send off all of the messages via websocket in just a few ms but the typing on the guest side took a few seconds, so there is some buffering and processing going on and I think we are just hitting some limit in this.

This `sendText` function doesn't seem to exist in upstream noVNC, so just patching it here seems reasonable. I suspect we added this for feature parity with whatever noVNC replaced.

In testing it is now much more reliable in sending a few paragraphs of text, and not visibly slower than before, but obviously still too slow to do anything really big with it. It isn't really the goal to be able to paste full documents.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### Screenshots

Before fix, paste is cut off:

Screenshot 2023-04-26 at 10 56 11 AM

After fix:

Screenshot 2023-04-26 at 10 51 44 AM

### How Has This Been Tested?
Hot patched several console proxies in testing environments, and their handling of large pastes improved.

### DIFF
--- a/systemvm/agent/noVNC/core/rfb.js
+++ b/systemvm/agent/noVNC/core/rfb.js
@@ -433,21 +433,30 @@ export default class RFB extends EventTargetMixin {
         this._resumeAuthentication();
     }
 
-    sendText(text) {
-        for (var i = 0; i < text.length; i++) {
-            const character = text.charAt(i);
-            var charCode = USKeyTable[character] || false;
-            if (charCode) {
-                this.sendKey(charCode, character, true);
-                this.sendKey(charCode, character, false);
-            } else {
-                charCode = text.charCodeAt(i)
-                this.sendKey(KeyTable.XK_Shift_L, "ShiftLeft", true);
-                this.sendKey(charCode, character, true);
-                this.sendKey(charCode, character, false);
-                this.sendKey(KeyTable.XK_Shift_L, "ShiftLeft", false);
+     sendText(text) {
+        const sleep = (time) => {
+            return new Promise(resolve => setTimeout(resolve, time))
+        }
+
+        const keyboardTypeText = async () => {
+            for (var i = 0; i < text.length; i++) {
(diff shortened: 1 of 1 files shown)


rohityadavcloud:
@blueorangutan package

nvazquez (approved):
LGTM. Tested OK trying to paste the PR description into a text file through the VM console.

- Before the fix:
Screenshot 2023-05-05 at 15 59 43

- After the fix:
Screenshot 2023-05-05 at 16 00 06

Thanks to @weizhouapache I have tested reducing the sleep time to 10 ms (https://gitlab.com/qemu-project/qemu/-/commit/d3b0db6dfea6b3a9ee0d96aceb796bdcafa84314) but observed at some point the pasted text turned into uppercase, so I'm fine with the current sleep time:
Screenshot 2023-05-05 at 16 21 20

rohityadavcloud (approved):

rohityadavcloud:
Since changes were in novnc js code, only manual testing is necessary. Merging based on review and @nvazquez 's manual testing. Thanks.

PR #7498, opened 2023-05-05
Fix null pointer on PowerFlex attach volume edge case

mlsorensen:
### Description

This PR fixes a null pointer edge case where a PowerFlex volume is attached to a VM.

In this edge case, a VM has been created, started, stopped, and then reimaged. VM has a last host ID but the newly attached volume does not yet have a storage pool assigned, it will be assigned on the VM start.  However, we assume that if the VM's host is not null, we need to try to revoke access to the volume for the host. Since there is no storage pool yet for the volume, we hit a null pointer when checking to see if the volume's pool is PowerFlex.

This was affecting all storage types, I could reproduce it with local storage, since the null pointer is at the check for pool's type.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

Verified but was reproducible with a local storage dev environment. Applied patch and confirmed it fixed the issue.

### DIFF
--- a/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
+++ b/server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
@@ -4260,7 +4260,7 @@ public class VolumeApiServiceImpl extends ManagerBase implements VolumeApiServic
                         _volsDao.update(volumeToAttach.getId(), volumeToAttach);
                     }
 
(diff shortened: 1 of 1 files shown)


rohityadavcloud (approved):
NPE check LGTM

rohityadavcloud:
This is only NPE check, simply packaging and github actions tests are enough.
@blueorangutan package

shwstppr (approved):
LGTM

PR #7509, opened 2023-05-10
cks: enable by default

shwstppr:
### Description

This change will allow CKS to be enabled by default on new installs. It should not affect the server or performance but would help to highlight k8s support in CloudStack.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/plugins/integrations/kubernetes-service/src/main/java/com/cloud/kubernetes/cluster/KubernetesClusterService.java
+++ b/plugins/integrations/kubernetes-service/src/main/java/com/cloud/kubernetes/cluster/KubernetesClusterService.java
@@ -38,7 +38,7 @@ public interface KubernetesClusterService extends PluggableService, Configurable
 
     static final ConfigKey<Boolean> KubernetesServiceEnabled = new ConfigKey<Boolean>("Advanced", Boolean.class,
             "cloud.kubernetes.service.enabled",
-            "false",
+            "true",
             "Indicates whether Kubernetes Service plugin is enabled or not. Management server restart needed on change",
             false);
     static final ConfigKey<String> KubernetesClusterNetworkOffering = new ConfigKey<String>("Advanced", String.class,


weizhouapache (commented):
LGTM, may need some discussion with the community

btw, can we make it as a dynamic setting so it does not require a management server restart ?

andrijapanicsb:
@blueorangutan package

andrijapanicsb:
@blueorangutan test

andrijapanicsb:
@blueorangutan SFpackage

andrijapanicsb:
@blueorangutan LLpackage

rohityadavcloud (approved):

DaanHoogland:
> LGTM, may need some discussion with the community

@weizhouapache , is there any ramification on enabling it, like performance degradation? I don t see why we couldn't enable it if not. 

> btw, can we make it as a dynamic setting so it does not require a management server restart ?

@shwstppr , I checked the code and it seems to me we can make it dynamic. Any concerns with that?

plugins/integrations/kubernetes-service/src/main/java/com/cloud/kubernetes/cluster/KubernetesClusterService.java
```diff
@@ -38,7 +38,7 @@ public interface KubernetesClusterService extends PluggableService, Configurable
 
     static final ConfigKey<Boolean> KubernetesServiceEnabled = new ConfigKey<Boolean>("Advanced", Boolean.class,
             "cloud.kubernetes.service.enabled",
-            "false",
+            "true",
             "Indicates whether Kubernetes Service plugin is enabled or not. Management server restart needed on change",
             false);
```
DaanHoogland, line 43:
```suggestion
            true);
```

weizhouapache:
as far as remember, smoke test test_kubernetes_clusters.py updates the globals setting and restarts the mangement server.
If the setting is dynamic, restarting mgmt server is not required (actually it is not required even if the setting is not dynamic)
cc @shwstppr

shwstppr:
@weizhouapache test updates the config and restarts MS only when the config is false.
If the config is made dynamic then restart won't be needed.
Config was not made dynamic initially because APIs are accessible only when the config is true. If it is made dynamic then APIs would always be available. Though we can add checks that they throw proper error  when the config is false.
In terms of overhead, only two GC and state scanner threads would be running additionally. We may also have to add a config to control interval for state scanner thread.

plugins/integrations/kubernetes-service/src/main/java/com/cloud/kubernetes/cluster/KubernetesClusterService.java
rohityadavcloud: It seems it’s not dynamic config

rohityadavcloud (approved):
Lgtm this is only for fresh env and won’t affect existing env, fresh env can still try to disable cks. I don’t see any harm or issue with the change.

DaanHoogland:
> ....
> Config was not made dynamic initially because APIs are accessible only when the config is true. If it is made dynamic then APIs would always be available. Though we can add checks that they throw proper error when the config is false. In terms of overhead, only two GC and state scanner threads would be running additionally. We may also have to add a config to control interval for state scanner thread.

@shwstppr , I did not find any check to disable API publishing based on this config value. Can you link me to that mechanism?

plugins/integrations/kubernetes-service/src/main/java/com/cloud/kubernetes/cluster/KubernetesClusterService.java
DaanHoogland: I have not found a reason why it can't be (yet)

DaanHoogland (approved):

shwstppr:
@DaanHoogland https://github.com/apache/cloudstack/blob/main/plugins/integrations/kubernetes-service/src/main/java/com/cloud/kubernetes/cluster/KubernetesClusterManagerImpl.java#L1333-L1347

DaanHoogland:
thanks @shwstppr , yes, that means that we can not just turn it into a dynamic setting. More needs to be done and I wonder if the setting is still needed at all. Let's visit that later.

shwstppr:
@blueorangutan LLpackage

shwstppr:
@blueorangutan test

shwstppr:
@blueorangutan LLtest

shwstppr:
@blueorangutan package

rohityadavcloud:
@blueorangutan LLtest

shwstppr:
@blueorangutan test

weizhouapache (approved):
@shwstppr 
looks good to me

shwstppr:
@weizhouapache @DaanHoogland @rohityadavcloud @andrijapanicsb are we okay to merge this?

weizhouapache:
> @weizhouapache @DaanHoogland @rohityadavcloud @andrijapanicsb are we okay to merge this?

@shwstppr yes, lgtm

PR #7554, opened 2023-05-24
server,api,ui: host auto-select for migrateVirtualMachineWithVolume

### LINKED ISSUE #6773: AutoSelect live migration fails with local storage
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
UI, API
~~~

##### CLOUDSTACK VERSION
~~~
4.16.1
~~~

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT
N/A

##### SUMMARY
When live migrating an instance and choosing "AutoSelect" it is not possible to live migrate an instance with local storage, while when selecting the destination host, it is possible to live migrate an instance with local storage.

##### STEPS TO REPRODUCE
~~~
- Spawn vm on local storage
- Live migrate instance in UI using "AutoSelect"
- Migration fails

CS calls migrateVirtualMachine instead of migrateVirtualMachineWithVolume
~~~

##### EXPECTED RESULTS
~~~
CloudStack to be smart enough to also live migrate instances with local storage when using "AutoSelect"
CS calls migrateVirtualMachineWithVolume instead of migrateVirtualMachine
~~~

##### ACTUAL RESULTS

- Error is returned:
```
Migration failed
Unsupported operation, VM uses Local storage, cannot migrate
```

shwstppr:
### Description

Fixes #6773 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested with migration of a VM with 2 KVM hosts with local storage,

https://github.com/apache/cloudstack/assets/153340/990484ef-2007-4c0a-b77d-c90b834b1a66


shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test matrix

DaanHoogland:
@shwstppr do you need eyes on this?

api/src/main/java/org/apache/cloudstack/api/command/admin/vm/MigrateVirtualMachineWithVolumeCmd.java
```diff
@@ -146,28 +152,31 @@ public ApiCommandResourceType getApiResourceType() {
...
-        if (hostId == null && MapUtils.isEmpty(migrateVolumeTo)) {
-            throw new InvalidParameterValueException(String.format("Either %s or %s must be passed for migrating the VM.", ApiConstants.HOST_ID, ApiConstants.MIGRATE_TO));
+        if (hostId == null && MapUtils.isEmpty(migrateVolumeTo) && !Boolean.TRUE.equals(autoSelect)) {
+            throw new InvalidParameterValueException(String.format("Either %s or %s must be passed or %s must be true for migrating the VM.", ApiConstants.HOST_ID, ApiConstants.MIGRATE_TO, ApiConstants.AUTO_SELECT));
         }
 
         VirtualMachine virtualMachine = _userVmService.getVm(getVirtualMachineId());
-        if (!VirtualMachine.State.Running.equals(virtualMachine.getState()) && hostId != null) {
+        if (!VirtualMachine.State.Running.equals(virtualMachine.getState()) && (hostId != null || Boolean.TRUE.equals(autoSelect))) {
             throw new InvalidParameterValueException(String.format("%s is not in the Running state to migrate it to the new host.", virtualMachine));
```
DaanHoogland, line 161: not in scope, but I think this meesage is misleading a bit:
```suggestion
            throw new InvalidParameterValueException(String.format("%s is not in the Running state. It must be running to migrate to a new host.", virtualMachine));
```

api/src/main/java/org/apache/cloudstack/api/command/admin/vm/MigrateVirtualMachineWithVolumeCmd.java
```diff
@@ -146,28 +152,31 @@ public ApiCommandResourceType getApiResourceType() {
...
         }
 
-        if (!VirtualMachine.State.Stopped.equals(virtualMachine.getState()) && hostId == null) {
-            throw new InvalidParameterValueException(String.format("%s is not in the Stopped state to migrate, use the %s parameter to migrate it to a new host.",
-                    virtualMachine, ApiConstants.HOST_ID));
+        if (!VirtualMachine.State.Stopped.equals(virtualMachine.getState()) && hostId == null && Boolean.FALSE.equals(autoSelect)) {
+            throw new InvalidParameterValueException(String.format("%s is not in the Stopped state to migrate, use the %s or %s parameter to migrate it to a new host.",
```
DaanHoogland, line 165: again, not really in scope:
```suggestion
            throw new InvalidParameterValueException(String.format("%s is not in the Stopped. To migrate it, use the %s or %s parameter to select a new host.",
```

api/src/main/java/org/apache/cloudstack/api/command/admin/vm/MigrateVirtualMachineWithVolumeCmd.java
```diff
@@ -146,28 +152,31 @@ public ApiCommandResourceType getApiResourceType() {
...
             VirtualMachine migratedVm = null;
-            if (hostId != null) {
-                Host destinationHost = _resourceService.getHost(getHostId());
-                // OfflineVmwareMigration: destination host would have to not be a required parameter for stopped VMs
-                if (destinationHost == null) {
-                    s_logger.error(String.format("Unable to find the host with ID [%s].", getHostId()));
-                    throw new InvalidParameterValueException("Unable to find the specified host to migrate the VM.");
+            if (getHostId() != null || Boolean.TRUE.equals(autoSelect)) {
+                Host destinationHost = null;
+                if (!Boolean.TRUE.equals(autoSelect)) {
```
DaanHoogland, line 173: this is actually an implicit null check on gethostId(), maybe we can make it more explicit for readability:
```suggestion
                if (getHostId() != null) {
```
or 
```suggestion
                if (Boolean.FALSE.equals(autoSelect)) {
```
in the end it seems we only care about the host id not being null in this block. so maybe remove both the checks on `autoSelect` completely from this block?

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -6942,8 +6957,9 @@ public VirtualMachine migrateVirtualMachineWithVolume(Long vmId, Host destinatio
         Pair<Host, Host> sourceDestinationHosts = getHostsForMigrateVmWithStorage(vm, destinationHost);
         Host srcHost = sourceDestinationHosts.first();
 
-        if (!isVMUsingLocalStorage(vm) && MapUtils.isEmpty(volumeToPool)
-                && (destinationHost.getClusterId().equals(srcHost.getClusterId()) || isVmVolumesOnZoneWideStore(vm))){
+        final List<VolumeVO> volumes = _volsDao.findCreatedByInstance(vm.getId());
+        if (!isAnyVmVolumeUsingLocalStorage(volumes) && MapUtils.isEmpty(volumeToPool) && destinationHost != null
+                && (destinationHost.getClusterId().equals(srcHost.getClusterId()) || isAllVmVolumesOnZoneWideStore(volumes))) {
```
DaanHoogland, line 6962: this is a complicated condition. maybe you can extract it to a method for readability?
A good name would also clarify the use of the condition and making any comments unnecessary. Something like :
```
        if (isDestinationClearAndUnambiguousFor(volumes, volumeToPool, destinationHost)) {
```

DaanHoogland (approved):
code looks good

api/src/main/java/org/apache/cloudstack/api/command/admin/vm/MigrateVirtualMachineWithVolumeCmd.java
shwstppr: Leaving this as it can be addressed separately

api/src/main/java/org/apache/cloudstack/api/command/admin/vm/MigrateVirtualMachineWithVolumeCmd.java
shwstppr: Leaving this as it can be addressed separately

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
shwstppr: Done

api/src/main/java/org/apache/cloudstack/api/command/admin/vm/MigrateVirtualMachineWithVolumeCmd.java
shwstppr: Done

shwstppr:
@blueorangutan package

api/src/main/java/org/apache/cloudstack/api/command/admin/vm/MigrateVirtualMachineWithVolumeCmd.java
```diff
@@ -146,28 +152,31 @@ public ApiCommandResourceType getApiResourceType() {
...
-                    throw new InvalidParameterValueException("Unable to find the specified host to migrate the VM.");
+            if (getHostId() != null || Boolean.TRUE.equals(autoSelect)) {
+                Host destinationHost = null;
+                if (getHostId() != null) {
+                    destinationHost = _resourceService.getHost(getHostId());
+                    // OfflineVmwareMigration: destination host would have to not be a required parameter for stopped VMs
+                    if (destinationHost == null) {
+                        s_logger.error(String.format("Unable to find the host with ID [%s].", getHostId()));
+                        throw new InvalidParameterValueException("Unable to find the specified host to migrate the VM.");
+                    }
```
DaanHoogland, line 179: @shwstppr , isn´t this equivilent?  This is a genuine question. I do not see how `autoSelect is used in this bit
```suggestion
            if (getHostId() != null) {
                Host destinationHost = _resourceService.getHost(getHostId());
                // OfflineVmwareMigration: destination host would have to not be a required parameter for stopped VMs
                if (destinationHost == null && Boolean.FALSE.equals(autoSelect)) {
                    s_logger.error(String.format("Unable to find the host with ID [%s].", getHostId()));
                    throw new InvalidParameterValueException("Unable to find the specified host to migrate the VM.");
```

shwstppr: @DaanHoogland no, not the same. If either hostid is provided or autoselect is true then the code tries to do migration with UserVmService.migrateVirtualMachineWithVolume else it uses UserVmService.vmStorageMigration

DaanHoogland: ok, i see now what I missed. If we enter this block because `Boolean.TRUE.equals(autoSelect)` we do nothing and just pass a `null` to `migrateVirtualMachineWithVolume`. I think this would read easier if the block was extracted into a new method.

shwstppr:
@blueorangutan package

rohityadavcloud (approved):

rohityadavcloud:
@blueorangutan test matrix

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test matrix

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan test matrix

DaanHoogland:
tested on KVM and checked if the ROOT volume would end up on the new host's local storage.

DaanHoogland:
@weizhouapache do we need this on 4.18? (cc @shwstppr )

shwstppr:
@weizhouapache changes in this PR adds a new API request param which is why I based it on main.
If you think that would be okay for 4.18.1, I can update it to Target 4.18 branch

DaanHoogland:
> @weizhouapache changes in this PR adds a new API request param which is why I based it on main. If you think that would be okay for 4.18.1, I can update it to Target 4.18 branch

A corner case @shwstppr , it is not incompatible and seems like a small enhancement. Whatever you say.

PR #7575, opened 2023-05-31
Removing use of SmartConnectNoSSL and adding custom SSL context

msinhore:
### Description

This PR removes the dependency on the deprecated method SmartConnectNoSSL from the pyVim.connect module. The pyVmomi library removed this method in newer versions, leading to errors when trying to import it.

Instead of using SmartConnectNoSSL, I have modified the code to use SmartConnect with a custom SSL context when SSL verification is disabled. Python has an inbuilt method to create an SSL context that does not verify the authenticity of the SSL certificate, ssl._create_unverified_context(). This method is used to create a context which is passed to SmartConnect.

The following changes were made to the code:

Import of ssl added at the top of the file.
SmartConnectNoSSL was removed from the imports from the pyVim.connect module. Modified the block of code where SmartConnect or SmartConnectNoSSL was used to create a service instance. It now creates an unverified SSL context if the disable_ssl_verification option is set and Python supports creating unverified SSL contexts. This context is then passed to SmartConnect. This PR helps to maintain compatibility with newer versions of pyVmomi and avoids using deprecated methods.

Finally, it should be noted that disabling SSL verification can lead to potential security issues. This option should be used with care and preferably in a testing or development environment.
...

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### How Has This Been Tested?

pyVmomi installed using pip
Management Server: Alma Linux 8

Command used to test: 
/usr/share/cloudstack-common/scripts/vm/hypervisor/vmware/discover_networks.py  -s vcenter-host-ip -u "Administrator" -p "vcenter_password" -S

### DIFF
--- a/scripts/vm/hypervisor/vmware/discover_networks.py
+++ b/scripts/vm/hypervisor/vmware/discover_networks.py
@@ -22,6 +22,7 @@ import atexit
 import sys
 import argparse
 import json
+import ssl
 
 isDebugLogs = False
 hostClusterNameDict = {}
--- a/scripts/vm/hypervisor/vmware/discover_networks.py
+++ b/scripts/vm/hypervisor/vmware/discover_networks.py
@@ -16,7 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from pyVim.connect import SmartConnect, SmartConnectNoSSL, Disconnect
+from pyVim.connect import SmartConnect, Disconnect
 from pyVmomi import vim
 import atexit
 import sys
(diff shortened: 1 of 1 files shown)


DaanHoogland (approved):
clgtm, needs regression testing

DaanHoogland:
@blueorangutan package

rohityadavcloud:
Do we use this file, how (if it's documented somewhere)?

msinhore:
@rohityadavcloud yes. You can found it here: https://docs.cloudstack.apache.org/en/latest/adminguide/virtual_machines.html#importing-virtual-machines

PR #7577, opened 2023-06-01
Upgrade mockito

vishesh92:
### Description

This PR upgrades the mockito version and removes powermockito from `utils` module.

Moving to Mockito to mock static methods (available since 3.4.0) instead of powermockito since it hasn't had any release since 2020 and is not compatible with newer versions of mockito.

I will create separate PRs for other modules overtime and eventually remove powermockito.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -113,7 +113,7 @@
         <cs.junit.dataprovider.version>1.13.1</cs.junit.dataprovider.version>
         <cs.junit.jupiter.version>5.9.1</cs.junit.jupiter.version>
         <cs.guava-testlib.version>18.0</cs.guava-testlib.version>
-        <cs.mockito.version>3.2.4</cs.mockito.version>
+        <cs.mockito.version>3.12.4</cs.mockito.version>
         <cs.powermock.version>2.0.5</cs.powermock.version>
         <cs.selenium.server.version>1.0-20081010.060147</cs.selenium.server.version>
         <cs.selenium-java-client-driver.version>1.0.1</cs.selenium-java-client-driver.version>
--- a/pom.xml
+++ b/pom.xml
@@ -1062,7 +1062,7 @@
                             <exclude>ui/public/**</exclude>
                             <exclude>ui/legacy/**</exclude>
                             <exclude>utils/testsmallfileinactive</exclude>
-                            <exclude>utils/src/test/resources/mockito-extension/org.mockito.plugins.MockMaker</exclude>
+                            <exclude>utils/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker</exclude>
                         </excludes>
                     </configuration>
                 </plugin>
--- a/pom.xml
+++ b/pom.xml
@@ -1062,6 +1062,7 @@
                             <exclude>ui/public/**</exclude>
                             <exclude>ui/legacy/**</exclude>
                             <exclude>utils/testsmallfileinactive</exclude>
+                            <exclude>utils/src/test/resources/mockito-extension/org.mockito.plugins.MockMaker</exclude>
                         </excludes>
                     </configuration>
                 </plugin>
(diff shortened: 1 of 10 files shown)


harikrishna-patnala:
@blueorangutan package

pom.xml
```diff
@@ -113,7 +113,7 @@
         <cs.junit.dataprovider.version>1.13.1</cs.junit.dataprovider.version>
         <cs.junit.jupiter.version>5.9.1</cs.junit.jupiter.version>
         <cs.guava-testlib.version>18.0</cs.guava-testlib.version>
-        <cs.mockito.version>3.2.4</cs.mockito.version>
```
harikrishna-patnala, line 116: we may also need the following change, so mockstatic() can be used with Mockito itself.

-      <artifactId>mockito-core</artifactId>
+      <artifactId>mockito-inline</artifactId>

Let's first run tests and see nothing is broken

DaanHoogland: @harikrishna-patnala tests have been run in the github acrions. Do you need to see anything else?

vishesh92: @harikrishna-patnala Let me make this change. Then we can incrementally remove PowerMockito.

utils/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
DaanHoogland, line 1: license?

vishesh92, line 1: Need to make this change to fix this error.
ref: https://javadoc.io/doc/org.mockito/mockito-core/3.12.4/org/mockito/Mockito.html#Mocking_Final
```
org.mockito.exceptions.base.MockitoException: 
The used MockMaker PowerMockMaker does not support the creation of static mocks

Mockito's inline mock maker supports static mocks based on the Instrumentation API.
You can simply enable this mock mode, by placing the 'mockito-inline' artifact where you are currently using 'mockito-core'.
Note that Mockito's inline mock maker is not supported on Android.
```

DaanHoogland: We can add it to the excluded files in pom.xml as well @vishesh92 (if that is the wiser thing to do?)

vishesh92: I don't have much experience with `pom.xml`. Are you suggesting that we add exclusion criteria to `powermock` dependencies in `pom.xml`

DaanHoogland: no , just an exclusion to the license requitrement for this config file that only contains code that we did not come up with ;)
in pom.xml at project level around line 1064 add
```
                            <exclude>utils/src/test/resources/mockito-extension/org.mockito.plugins.MockMaker</exclude>
```

vishesh92: Got it. I have updated the PR.

DaanHoogland (approved):
clgtm

rohityadavcloud (approved):

utils/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
rohityadavcloud, line 1: @vishesh92 is this boilerplate or does it serve any purpose? If you want an empty file you can touch the file and git add.

rohityadavcloud:
@blueorangutan package

utils/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
vishesh92: We will need to add this file to module when we remove powermock. It specifies the MockMaker to use in that module for running tests.

more info here: https://github.com/apache/cloudstack/pull/7577/files#r1217685958

Issue #7574, opened 2023-05-31
kvm: Agent does not honor control.cidr anymore


wido:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
KVM Agent
~~~

##### CLOUDSTACK VERSION
~~~
4.18
~~~

##### CONFIGURATION
- Advanced Networking
- FRR BGP daemon with BGP Unnumbered and VXLAN

##### OS / ENVIRONMENT
Ubuntu 22.04

##### SUMMARY
The 'control.cidr' configuration value from the agent.properties is no longer honered and thus the subnet 169.254.0.0/16 is always used.

This conflicts with BGP Unnumbered and thus renders the network configuration broken.

##### STEPS TO REPRODUCE
- Install KVM Agent 4.18
- Set control.cidr to 169.254.240.0/20
- See the cloud0 still gets the address 169.254.0.1/255.255.0.0

##### EXPECTED RESULTS
I would expect the CIDR to be honered.

##### LOGS

```
May 31 12:18:09 hv-pod2-r1-21 java[5559]: DEBUG [utils.script.Script] (main:) (logid:) Executing: /bin/bash -c ip address add 169.254.0.1/255.255.0.0 dev cloud0
May 31 12:18:09 hv-pod2-r1-21 java[5559]: DEBUG [utils.script.Script] (main:) (logid:) Cannot find device "cloud0"
May 31 12:18:09 hv-pod2-r1-21 java[5559]: DEBUG [utils.script.Script] (main:) (logid:) Executing: /bin/bash -c ip route add 169.254.0.0/16 dev cloud0 src 169.254.0.1
May 31 12:18:09 hv-pod2-r1-21 java[5559]: DEBUG [utils.script.Script] (main:) (logid:) Cannot find device "cloud0"
May 31 12:18:09 hv-pod2-r1-21 java[5559]: DEBUG [utils.script.Script] (main:) (logid:) Executing: /bin/bash -c ip link add name cloud0 type bridge
May 31 12:18:09 hv-pod2-r1-21 java[5559]: DEBUG [utils.script.Script] (main:) (logid:) Executing: /bin/bash -c ip link set cloud0 up
May 31 12:18:09 hv-pod2-r1-21 systemd-networkd[3152]: cloud0: Link UP
[...]

Issue #6715, opened 2022-09-08
ControlNetworkGuru does not take into account custom CIDR when assigning System VM link local IP address


Rubueno:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
Management, Network, SystemVM
~~~

##### CLOUDSTACK VERSION
~~~
4.16.1
~~~

##### CONFIGURATION
In database:
```
MariaDB [cloud]> select name,value from configuration where name = 'control.cidr';
+--------------+------------------+
| name         | value            |
+--------------+------------------+
| control.cidr | 169.254.240.0/20 |
+--------------+------------------+
MariaDB [cloud]> select name,value from configuration where name = 'control.gateway';
+-----------------+---------------+
| name            | value         |
+-----------------+---------------+
| control.gateway | 169.254.240.1 |
+-----------------+---------------+
MariaDB [cloud]> select count(*) from op_dc_link_local_ip_address_alloc;
+----------+
| count(*) |
+----------+
|    65533 |
+----------+
```
On agent:
```
root@hv-04-02-21:~# grep control.cidr /etc/cloudstack/agent/agent.properties
control.cidr=169.254.240.0/20
```

##### SUMMARY
When having set a `control.cidr` different from the norm (`168.254.0.0/16`) and the database table `op_dc_link_local_ip_address_alloc` is already populated with all the addresses in the default CIDR, then allocating an IP to a System VM does not take into account the requested CIDR, thus allocating an IP address in the wrong range.

https://github.com/apache/cloudstack/blob/main/engine/schema/src/main/java/com/cloud/dc/dao/DataCenterLinkLocalIpAddressDaoImpl.java#L52 should take this into account. This function is called in https://github.com/apache/cloudstack/blob/main/server/src/main/java/com/cloud/network/guru/ControlNetworkGuru.java#L134

##### STEPS TO REPRODUCE
- Install new management server
- After kickstart, disable the zone to prevent any system vm deploys
- In mysql database: `select count(*) from op_dc_link_local_ip_address_alloc;`, it would return `65533`
- Connect agents
- Configure `control.cidr` on the manager and agents, and `control.gateway` on the manager, ie:
```
control.cidr=169.254.240.0/20
control.gateway=169.254.240.1
```
- Enable the zone, check the system vm link local address

##### EXPECTED RESULTS
- The System VMs should receive an IP address in the configured CIDR.
[...]

PR #7580, opened 2023-06-01
kvm agent: get the right controll cidr

DaanHoogland:
### Description

This PR ...

Fixes: #7574

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
@@ -65,10 +65,7 @@ public class BridgeVifDriver extends VifDriverBase {
 
         String networkScriptsDir = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.NETWORK_SCRIPTS_DIR);
 
-        String controlCidr = (String)params.get("control.cidr");
-        if (StringUtils.isNotBlank(controlCidr)) {
-            _controlCidr = controlCidr;
-        }
+        _controlCidr = getControllCidr(_controlCidr, params);
 
         String value = (String)params.get("scripts.timeout");
         _timeout = NumbersUtil.parseInt(value, 30 * 60) * 1000;
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
@@ -65,7 +65,7 @@ public class BridgeVifDriver extends VifDriverBase {
 
         String networkScriptsDir = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.NETWORK_SCRIPTS_DIR);
 
-        _controlCidr = getControlCidr(_controlCidr, params);
+        _controlCidr = getControlCidr(_controlCidr);
 
         String value = (String)params.get("scripts.timeout");
         _timeout = NumbersUtil.parseInt(value, 30 * 60) * 1000;
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
@@ -65,7 +65,7 @@ public class BridgeVifDriver extends VifDriverBase {
 
         String networkScriptsDir = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.NETWORK_SCRIPTS_DIR);
 
-        _controlCidr = getControllCidr(_controlCidr, params);
+        _controlCidr = getControlCidr(_controlCidr, params);
 
         String value = (String)params.get("scripts.timeout");
         _timeout = NumbersUtil.parseInt(value, 30 * 60) * 1000;
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
@@ -44,15 +44,15 @@ public abstract class VifDriverBase implements VifDriver {
         _pifs = (Map<String, String>)params.get("libvirt.host.pifs");
     }
 
-    protected String getControllCidr(String defaultValue, Map<String, Object> params) {
-        String controllCidr = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.CONTROLL_CIDR);
-        if (StringUtils.isBlank(controllCidr)) {
-            controllCidr = (String) params.get("control.cidr");
+    protected String getControlCidr(String defaultValue, Map<String, Object> params) {
+        String controlCidr = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.CONTROL_CIDR);
+        if (StringUtils.isBlank(controlCidr)) {
+            controlCidr = (String) params.get("control.cidr");
         }
-        if (StringUtils.isBlank(controllCidr)) {
-            controllCidr = defaultValue;
+        if (StringUtils.isBlank(controlCidr)) {
+            controlCidr = defaultValue;
         }
-        return controllCidr;
+        return controlCidr;
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
```diff
@@ -65,10 +65,7 @@ public void configure(Map<String, Object> params) throws ConfigurationException
 
         String networkScriptsDir = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.NETWORK_SCRIPTS_DIR);
 
-        String controlCidr = (String)params.get("control.cidr");
-        if (StringUtils.isNotBlank(controlCidr)) {
-            _controlCidr = controlCidr;
-        }
+        _controlCidr = getControllCidr(_controlCidr, params);
```
wido, line 68: Single 'l' instead of 'll' for control?

wido (commented):
So far it looks good. I'm going on holiday for two weeks now, but this is a fix in the right direction

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
DaanHoogland: I think the double l is correct English (you dutchy;)

DaanHoogland:
> So far it looks good. I'm going on holiday for two weeks now, but this is a fix in the right direction

any colleague that could test this, @wido ?

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
ping @wido can you test?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/BridgeVifDriver.java
wido: I'd still say one l because that's how it's usually written and in the code right now everywhere

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
```diff
@@ -44,15 +44,15 @@ public void configure(Map<String, Object> params) throws ConfigurationException
...
     }
 
-    protected String getControllCidr(String defaultValue, Map<String, Object> params) {
-        String controllCidr = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.CONTROLL_CIDR);
-        if (StringUtils.isBlank(controllCidr)) {
-            controllCidr = (String) params.get("control.cidr");
+    protected String getControlCidr(String defaultValue, Map<String, Object> params) {
+        String controlCidr = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.CONTROL_CIDR);
+        if (StringUtils.isBlank(controlCidr)) {
+            controlCidr = (String) params.get("control.cidr");
```
wido, line 50: I think this is not needed. The AgentPropertiesFileHandler class should provide us what's in the agent.properties file or return the default.

Do we maybe want to add some debug logging here as well?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
```diff
@@ -44,15 +44,15 @@ public void configure(Map<String, Object> params) throws ConfigurationException
...
         }
-        if (StringUtils.isBlank(controllCidr)) {
-            controllCidr = defaultValue;
+        if (StringUtils.isBlank(controlCidr)) {
+            controlCidr = defaultValue;
```
wido, line 53: I think this defaultValue can go as you already define this in [AgentProperties.java](https://github.com/apache/cloudstack/pull/7580/files/44752f1d208d03c6ee3e9ac8bbef0dad2cef8803..d4da6fa9e067d3c7a900ff459f1332a8a021f0c6#diff-967488af285b7f13f46833e85a5bee82912ec7af150055f14a0b7e70a8efc3fe)

wido (changes requested):
I think we are almost there!

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
DaanHoogland: I think the value may also come from the MS, so if this is not configured (for a certain agent) the global configuration is used.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
DaanHoogland: You may want to only define this for a small number (a pod or one of the zones) (??)

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
wido: params.get() does not fetch anything from the MS, right?

Besides that, the cloud0 bridge is configured before the MS is contacted, so any value from from the MS is useless at that point.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
wido: This value configures how the agent configures the cloud0 Linux bridge locally on the hypervisor.

But the default value is already set in the AgentProperties, so we don't need this 'defaultValue' variable anymore.

Rubueno:
@wido can you verify if this also fixes the problem mentioned in #6715? To verify if it is fixed now: bootstrap the management server with /16, then change to /20 and start a system vm.

wido:
> @wido can you verify if this also fixes the problem mentioned in #6715? To verify if it is fixed now: bootstrap the management server with /16, then change to /20 and start a system vm.

No, it doesn't. This PR is to fix the subnet assigned to the cloud0 bridge on the KVM agent.

What you mention is something I observed as well and didn't put much effort into that.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
DaanHoogland: I'll go through the flow once more, but as I recall params is not only filled from local (agent.properties/env) configuration but also from the values supplied in the commands from the MS. So `params.get()` does not retrieve from the server but `params` already contains values sent by the server.

wido: You might be right, however, once the cloud0 bridge is configured by the agent it's never touched again. This happens before it connects to the mgmt server.

DaanHoogland:
@blueorangutan test

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/VifDriverBase.java
```diff
@@ -41,6 +43,11 @@ public void configure(Map<String, Object> params) throws ConfigurationException
         _pifs = (Map<String, String>)params.get("libvirt.host.pifs");
     }
 
+    protected String getControlCidr(String defaultValue, Map<String, Object> params) {
```
wido, line 46: Almost looks good. You aren't using the params variable anywhere, right?

DaanHoogland: sorry, I do not give this pr proper tlc (bit detached fromit)

wido (approved):
This looks good to me!

DaanHoogland:
@weizhouapache can you have a look?

rohityadavcloud (approved):

DaanHoogland:
> > @wido can you verify if this also fixes the problem mentioned in #6715? To verify if it is fixed now: bootstrap the management server with /16, then change to /20 and start a system vm.
> 
> No, it doesn't. This PR is to fix the subnet assigned to the cloud0 bridge on the KVM agent.
> 
> What you mention is something I observed as well and didn't put much effort into that.

@wido, @Rubueno , can you have a look at #7746 ? I intend to solve that issue there.

PR #7594, opened 2023-06-06
Fix: Volumes on lost local storage cannot be removed

nvazquez:
### Description

This PR fixes the following:
- When destroying a local storage pool during host force deletion: then also destroy the volumes on the pool from DB
- When expunging a destroyed volume: check if the volume's storage pool exists (not removed). In case its removed, skip the volume removal from storage

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?
- Create a local storage KVM zone
- Deploy a VM to local storage
- Force remove the host
- Delete the VM -> This succeeds but its volumes remain forever in "Destroy" (before the fix) -> After the fix: volumes are destroyed

### DIFF
--- a/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
+++ b/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
@@ -38,6 +38,9 @@ import javax.naming.ConfigurationException;
 
 import com.cloud.exception.StorageConflictException;
 import com.cloud.exception.StorageUnavailableException;
+import com.cloud.storage.Volume;
+import com.cloud.storage.VolumeVO;
+import com.cloud.storage.dao.VolumeDao;
 import org.apache.cloudstack.annotation.AnnotationService;
 import org.apache.cloudstack.annotation.dao.AnnotationDao;
 import org.apache.cloudstack.api.ApiConstants;
@@ -294,6 +297,8 @@ public class ResourceManagerImpl extends ManagerBase implements ResourceManager,
     private UserVmDetailsDao userVmDetailsDao;
     @Inject
     private AnnotationDao annotationDao;
+    @Inject
+    private VolumeDao volumeDao;
 
     private final long _nodeId = ManagementServerNode.getManagementServerId();
 
@@ -979,6 +984,7 @@ public class ResourceManagerImpl extends ManagerBase implements ResourceManager,
                     final Long poolId = pool.getPoolId();
                     final StoragePoolVO storagePool = _storagePoolDao.findById(poolId);
                     if (storagePool.isLocal() && isForceDeleteStorage) {
+                        destroyLocalStoragePoolVolumes(poolId);
                         storagePool.setUuid(null);
                         storagePool.setClusterId(null);
                         _storagePoolDao.update(poolId, storagePool);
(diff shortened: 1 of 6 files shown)


nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1635,6 +1645,17 @@ private void expungeVolumesInPrimaryOrSecondary(VolumeVO volume, DataStoreRole r
             }
         }
     }
+
+    private boolean canAccessVolumeStore(VolumeVO volume, DataStoreRole role) {
+        if (volume == null) {
+            throw new CloudRuntimeException("Cannot find the volume");
```
DaanHoogland, line 1651: this is a strange message in this place. We are not trying to find it here, but we claim that is what failed. maybe this check should be somewhere else or the message should be adjusted to "No Volume given" or so.

DaanHoogland (commented):
mostly sensible but I found one strangity

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
nvazquez: Thanks @DaanHoogland, fixed

nvazquez:
@blueorangutan package

DaanHoogland (approved):
clgtm

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -979,6 +983,7 @@ public void doInTransactionWithoutResult(final TransactionStatus status) {
                     final Long poolId = pool.getPoolId();
                     final StoragePoolVO storagePool = _storagePoolDao.findById(poolId);
                     if (storagePool.isLocal() && isForceDeleteStorage) {
+                        destroyLocalStoragePoolVolumes(poolId);
```
weizhouapache, line 986: this should be good :+1:

server/src/main/java/com/cloud/storage/VolumeApiServiceImpl.java
```diff
@@ -1635,6 +1645,17 @@ private void expungeVolumesInPrimaryOrSecondary(VolumeVO volume, DataStoreRole r
...
         }
     }
+
+    private boolean canAccessVolumeStore(VolumeVO volume, DataStoreRole role) {
+        if (volume == null) {
+            throw new CloudRuntimeException("No volume given, cannot check access to volume store");
+        }
+        InternalIdentity pool = role == DataStoreRole.Primary ?
+                _storagePoolDao.findById(volume.getPoolId()) :
+                imageStoreDao.findById(volume.getPoolId());
```
weizhouapache, line 1655: use switch/case ?

shwstppr (approved):
Code LGTM

kiranchavala (commented):
Hi @nvazquez 

Observing the following issue 

1. Create a disk offering  with storage type as local 
2. Create a compute offering with storage type as local 
3. Launch a vm with data disk , wait till its in running state
4. Disable the host
5. Force delete the host
6. Found the vm gets into destroyed state and is not expunged
7. The data disk is still in ready state 

Ideally the vm and data disk on the local storage should also get remvoed automatically if the local storage is removed

Currently the end user has to manually expunge the vm and the destory data disk

nvazquez:
Thanks @kiranchavala, have pushed a fix for that scenario

nvazquez:
@blueorangutan package

kiranchavala (approved):
LGTM, Tested the data disk fix and it's working fine

rohityadavcloud (approved):
LGTM didn't test it

Issue #7595, opened 2023-06-06
Start VM shows the prefer last host option for user accounts


yadvr:
As a normal user account, I'm being asked if CloudStack should consider last host. This options shouldn't be shown to end users.

<img width="568" alt="Screenshot 2023-06-05 at 4 22 19 AM" src="https://github.com/apache/cloudstack/assets/95203/6e09362b-2ff2-4a19-bd30-77509ddfe4f9">

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.18
~~~

PR #7597, opened 2023-06-06
consider last host only for root-admin

DaanHoogland:
### Description

This PR removes the option to (not) consider last host when re-starting VMs for regular users.

Fixes: #7595 

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vm/StartVMCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vm/StartVMCmd.java
@@ -86,7 +86,8 @@ public class StartVMCmd extends BaseAsyncCmd implements UserCmd {
             type = CommandType.BOOLEAN,
             description = "True by default, CloudStack will firstly try to start the VM on the last host where it run on before stopping, if destination host is not specified. " +
                     "If false, CloudStack will not consider the last host and start the VM by normal process.",
-            since = "4.18.0")
+            since = "4.18.0",
+            authorized = {RoleType.Admin})
     private Boolean considerLastHost;
 
     @Parameter(name = ApiConstants.DEPLOYMENT_PLANNER, type = CommandType.STRING, description = "Deployment planner to use for vm allocation. Available to ROOT admin only", since = "4.4", authorized = { RoleType.Admin })
--- a/ui/src/views/compute/StartVirtualMachine.vue
+++ b/ui/src/views/compute/StartVirtualMachine.vue
@@ -88,6 +88,13 @@
               </a-select-option>
             </a-select>
           </a-form-item>
+
+          <a-form-item name="considerlasthost" ref="considerlasthost">
+            <template #label>
+              <tooltip-label :title="$t('label.considerlasthost')" :tooltip="apiParams.considerlasthost.description"/>
+            </template>
+            <a-switch v-model:checked="form.considerlasthost" />
+          </a-form-item>
         </div>
 
         <a-form-item name="bootintosetup" ref="bootintosetup" v-if="resource.hypervisor === 'VMware'">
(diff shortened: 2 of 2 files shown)


DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan ui

DaanHoogland:
> Do we need to make java changes as well ? cc @DaanHoogland @rohityadavcloud

this is just a UI change @weizhouapache . Do we need/want to block this on the API as well? I see no need for that but if we think so I will look into it.

DaanHoogland:
> > Do we need to make java changes as well ? cc @DaanHoogland @rohityadavcloud
> 
> this is just a UI change @weizhouapache . Do we need/want to block this on the API as well? I see no need for that but if we think so I will look into it.

I checked the code and the `clusterid` and `hostid` are not guarded against use by users in the API, so I think it makes no sense to guard against the use of `considerlasthost`. it is true by default whic is the old bahaviour. Having it as false will only make the process resemble a start of a fresh installed VM. 
Can you agree @weizhouapache ?

DaanHoogland:
> > > > Do we need to make java changes as well ? cc @DaanHoogland @rohityadavcloud
> > > 
> > > 
> > > this is just a UI change @weizhouapache . Do we need/want to block this on the API as well? I see no need for that but if we think so I will look into it.
> > 
> > 
> > I checked the code and the `clusterid` and `hostid` are not guarded against use by users in the API, so I think it makes no sense to guard against the use of `considerlasthost`. it is true by default whic is the old bahaviour. Having it as false will only make the process resemble a start of a fresh installed VM. Can you agree @weizhouapache ?
> 
> @DaanHoogland if normal user passes `clusterid` or `hostid`, it will might get an exception
> 
> https://github.com/apache/cloudstack/blob/4.18.0.0/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java#L5421-L5451
> 
> ```
>     private Cluster getDestinationCluster(Long clusterId, boolean isRootAdmin) {
>         Cluster destinationCluster = null;
>         if (clusterId != null) {
>             if (!isRootAdmin) {
>                 throw new PermissionDeniedException(
>                         "Parameter " + ApiConstants.CLUSTER_ID + " can only be specified by a Root Admin, permission denied");
>             }
>             destinationCluster = _clusterDao.findById(clusterId);
>             if (destinationCluster == null) {
>                 throw new InvalidParameterValueException("Unable to find the cluster to deploy the VM, cluster id=" + clusterId);
>             }
>         }
>         return destinationCluster;
>     }
> 
>     private Host getDestinationHost(Long hostId, boolean isRootAdmin, boolean isExplicitHost) {
>         Host destinationHost = null;
>         if (hostId != null) {
>             if (isExplicitHost && !isRootAdmin) {
>                 throw new PermissionDeniedException(
>                         "Parameter " + ApiConstants.HOST_ID + " can only be specified by a Root Admin, permission denied");
>             }
>             destinationHost = _hostDao.findById(hostId);
>             if (destinationHost == null) {
>                 throw new InvalidParameterValueException("Unable to find the host to deploy the VM, host id=" + hostId);
>             } else if (destinationHost.getResourceState() != ResourceState.Enabled || destinationHost.getStatus() != Status.Up ) {
>                 throw new InvalidParameterValueException("Unable to deploy the VM as the host: " + destinationHost.getName() + " is not in the right state");
>             }
>         }
>         return destinationHost;
>     }
> ```

ok I didn´t dig that deep, but still unsetting considerlasthost is not doing much. I think not bothering normal users with it is enough, we don't need to forbid the use.

weizhouapache:
> 
> ok I didn´t dig that deep, but still unsetting considerlasthost is not doing much. I think not bothering normal users with it is enough, we don't need to forbid the use.

@DaanHoogland ok, fine with me

weizhouapache (approved):
lgtm

tested on qacloud as normal user

shwstppr (commented):
UI change LGTM but in my opinion it would be better to have consistency between UI and API.
Adding `authorized` property for the parameter in the startVirtualMachine should do the trick

DaanHoogland:
@blueorangutan package

rohityadavcloud (approved):

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

Issue #5759, opened 2021-12-08
migrateVolume does not copy the template


GutoVeronezi:
When a volume is copied to another storage pool, if the template does not exist in the destination storage pool yet, it should be copied. The API `migrateVirtualMachineWithVolume` already does it; however, when migrating the volume with `migrateVolume` (VM stopped), the template is not copied, which causes inconsistencies in the database and primary storage systems.

DaanHoogland:
is this what you are fixing in #5758, @GutoVeronezi ? or do you mean that more improvement is needed after that?

GutoVeronezi:
@DaanHoogland it will need more improvement after that.

nvazquez:
Hi @GutoVeronezi are you working on this issue or planning to?

GutoVeronezi:
Hi, @nvazquez, I have other priorities at the moment, this issue is not in my plannings for now.

nvazquez:
Thanks for clarifying @GutoVeronezi

shwstppr:
Moving this to next milestone

rohityadavcloud:
I've hit this @GutoVeronezi for local storage (but not shared storage). Did you reproduce this only for KVM + local storage (or any other hypervisor/storage type)?

GutoVeronezi:
@rohityadavcloud, at the time I created this issue, I was running ACS with 4.15 or 4.16 (I do not remember quite well). I will test it again and verify if the situation still happens with the current version.

GutoVeronezi:
> @rohityadavcloud @GutoVeronezi I have tested with local storage as well shared storage. both work.
> 
> The migrateVolume will firstly copy the image to secondary storage (full clone) and then copy to another primary storage. It should not be a bug. therefore I have removed the label 'type:bug'
> 
> The whole process can be improved of course.

Thanks for the tests, @weizhouapache.

weizhouapache:
> > @rohityadavcloud @GutoVeronezi I have tested with local storage as well shared storage. both work.
> > The migrateVolume will firstly copy the image to secondary storage (full clone) and then copy to another primary storage. It should not be a bug. therefore I have removed the label 'type:bug'
> > The whole process can be improved of course.
> 
> Thanks for the tests, @weizhouapache.

@GutoVeronezi 
please feel free to improve it.

rohityadavcloud:
@weizhouapache there are some issues; cc @DaanHoogland @GutoVeronezi 
[...]

PR #7615, opened 2023-06-09
server: Fix NPE when migrating volume without template

gpordeus:
### Description

If you try to migrate a VM with volume and its template was removed, it results in a NullPointerException and the migration doesn't occur. 

If the template is already in both pools, the NPE happens in `getVolumeBackingFile`, when `_vmTemplateDao.findById()` returns null and it then tries to call from it. The fix was to check if it is null (and since it can only be null if it was removed, since `srcVolumeInfo` has `templateId`) and return null so it does a FullClone Migration.

If the template is not in both pools, the NPE happens in `copyTemplateToTargetFilesystemStorageIfNeeded`, when `TemplateObjectTO` tries to `getInstallPath()` on a null object. The fix was the same, check if `findById()` returns null and, if so, don't copy it.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?
- Uploaded two templates (A and B) to ACS, created one VM for each. Deleted B.
- Called migrateVirtualMachineWithVolume to migrate VM A on both host and storage pool.
- Called migrateVirtualMachineWithVolume to migrate VM B on both host and storage pool.
- Deleted A.
- Called migrateVirtualMachineWithVolume to migrate VM A, again.
All migrations were successful.


weizhouapache:
is it applicable for 4.18 ?

gpordeus:
> thanks @gpordeus for the fix (or enhancement even) . Do you think you can codify your testing by adding an integration test for these fixes to make sure we have no regression on those in the future?

@DaanHoogland I have added the integration tests.

> is it applicable for 4.18 ?

@weizhouapache Yes

test/integration/component/test_storage_motion_kvm.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def setUp(self):
+        self.apiclient = self.testClient.getApiClient()
+        self.dbclient = self.testClient.getDbConnection()
+        self.cleanup = []
+
+    def tearDown(self):
+        #Clean up, terminate the created ISOs
+        cleanup_resources(self.apiclient, self.cleanup)
+        return
```
DaanHoogland, line 153:
```suggestion
    def tearDown(self):
        super(TestStorageMotionAndTemplate, self).tearDown()
```

test/integration/component/test_storage_motion_kvm.py
```diff
@@ -0,0 +1,454 @@
...
+
+        self.debug("Creating vm with: %s" % template)
+        new_virtual_machine = VirtualMachine.create(
+            self.api_client,
+            self.services["small"],
+            accountid=self.account.name,
+            domainid=self.account.domainid,
+            templateid=template.id,
+            serviceofferingid=self.small_offering.id
+        )
```
DaanHoogland, line 330:
```suggestion
        new_virtual_machine = VirtualMachine.create(
            self.api_client,
            self.services["small"],
            accountid=self.account.name,
            domainid=self.account.domainid,
            templateid=template.id,
            serviceofferingid=self.small_offering.id
        )
        self.cleanup.append(new_virtual_machine)
```

test/integration/component/test_storage_motion_kvm.py
```diff
@@ -0,0 +1,454 @@
...
+        self.debug("Created vm %s" % new_virtual_machine)
+
+        self.debug("Deleting template: %s" % template)
+        self.cleanup.pop()
+        template.delete(self.apiclient)
```
DaanHoogland, line 335:
```suggestion
        template.delete(self.apiclient)
        self.cleanup.remove(template)
```

test/integration/component/test_storage_motion_kvm.py
```diff
@@ -0,0 +1,454 @@
...
+
+        cmd = migrateVirtualMachineWithVolume.migrateVirtualMachineWithVolumeCmd()
+        cmd.hostid = host.id
+        cmd.virtualmachineid = new_virtual_machine.id
+        cmd.migrateto = []
+        for volume, pool in migrateto.items():
+            cmd.migrateto.append({'volume':volume, 'pool':pool})
+        self.debug("hostid: %s, vmid: %s, migrateto: %s", cmd.hostid, cmd.virtualmachineid, cmd.migrateto)
+
+        self.apiclient.migrateVirtualMachineWithVolume(cmd)
```
DaanHoogland, line 402: in base.py you can find a class `VirtualMachine`, which has a method `migrate_vm_with_volume`, it does basically the same.

DaanHoogland (commented):
test code looks good, but uses some out of date constructs. suggestions added.

DaanHoogland:
> > thanks @gpordeus for the fix (or enhancement even) . Do you think you can codify your testing by adding an integration test for these fixes to make sure we have no regression on those in the future?
> 
> @DaanHoogland I have added the integration test.

good work, i left some comments, but nothing serious.

> 
> > is it applicable for 4.18 ?
> 
> @weizhouapache Yes

can you rebase the change on the 4.18 branch than?

.asf.yaml
```diff
@@ -49,11 +44,6 @@ github:
     rebase: false
 
   collaborators:
-    - kiranchavala
-    - rajujith
-    - alexandremattioli
-    - vishesh92
-    - soreana
```
DaanHoogland: Why do you want these contributors from this list?

DaanHoogland:
@gpordeus all of a sudden all kinds of unrelated files are changed in your PR. Can you have a look?

gpordeus:
> @gpordeus all of a sudden all kinds of unrelated files are changed in your PR. Can you have a look?

Fixed, but the labels added by the bot are still there.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland (approved):
code looks good to me

vladimirpetrov:
Hi @gpordeus, I tested and indeed there's no NPE anymore but the VMs are not able to start after a stop/start operation (on a KVM environment), did you notice the same behavior?
image

gpordeus:
> Hi @gpordeus, I tested and indeed there's no NPE anymore but the VMs are not able to start after a stop/start operation (on a KVM environment), did you notice the same behavior? image

Hi, @vladimirpetrov ! I did not, could you specify in which circumstances this happens? If it's before or after deleting the template and/or migrating, for instance.

vladimirpetrov:
This happens after deleting the template and successfully migrating the VM. Then when I do stop/start of the VM it doesn't start because of this error @gpordeus .

DaanHoogland:
@blueorangutan package

weizhouapache:
@blueorangutan test matrix

rohityadavcloud:
@blueorangutan package

DaanHoogland:
@blueorangutan test

rohityadavcloud (approved):
LGTM but I haven't tested it

DaanHoogland:
> > Hi @gpordeus, I tested and indeed there's no NPE anymore but the VMs are not able to start after a stop/start operation (on a KVM environment), did you notice the same behavior? image
> 
> Hi, @vladimirpetrov ! I did not, could you specify in which circumstances this happens? If it's before or after deleting the template and/or migrating, for instance.

@gpordeus (cc @vladimirpetrov ) any explanation/resolution on this?

weizhouapache:
@gpordeus 
do you have any update ?
in my opinion, if template does not exist on secondary storage and primary storage, an exception should be thrown somewhere.

gpordeus:
> @gpordeus (cc @vladimirpetrov ) any explanation/resolution on this?

@DaanHoogland , I've managed to reproduce the error (It happened when I migrated from an NFS primary storage to another), but I'm still looking into what's causing it. 


> @gpordeus do you have any update ? in my opinion, if template does not exist on secondary storage and primary storage, an exception should be thrown somewhere.

@weizhouapache I don't think an Exception should be thrown since the user is already allowed to force delete the template and the migration copies the template into the volume, so there is no absolute need for the backing file.

weizhouapache:
> the migration copies the template into the volume

@gpordeus 
I hope so, can you confirm it ?

rohityadavcloud:
@blueorangutan package

DaanHoogland:
@blueorangutan test

weizhouapache:
@blueorangutan package

weizhouapache:

weizhouapache:
@gpordeus 
I can reproduce the issue @vladimirpetrov faced in his testing

- create a env with 2 primary pools, and disable pool primary2
- register a template
- create two vms (A and B)
- force-delete the template
- enable pool primary2
- migrate running vm A to primary2, it succeeds
- stop vm B, and migrate it to primary2, it succeeds
- start vm B, all looks good.
- stop vm A, and start it again, vm A is stuck at booting
image

the qcow2 image does not have a backing file
```
# qemu-img info -U /mnt/c2876c2e-5312-3cf9-b332-64d597086dc9/1a6593ad-8d02-4b10-a520-3eeefb336b13
image: /mnt/c2876c2e-5312-3cf9-b332-64d597086dc9/1a6593ad-8d02-4b10-a520-3eeefb336b13
file format: qcow2
virtual size: 50 MiB (52428800 bytes)
disk size: 1.06 MiB
cluster_size: 65536
Format specific information:
    compat: 1.1
    compression type: zlib
    lazy refcounts: false
    refcount bits: 16
    corrupt: false
    extended l2: false
```

weizhouapache:
@DaanHoogland @rohityadavcloud @gpordeus 

my summary: if template is removed,

Before this PR,
- when migrate volume of stopped vm, it succeeds. vm also works after starting
- when migrate a running vm with volumes, got an NPE
Failed to migrate VM [VM instance {"id":3,"instanceName":"i-2-3-VM","type":"User","uuid":"1d3541d7-187e-417f-9cd4-cd18d0fc1e3d"}] along with its volumes due to [java.lang.NullPointerException].

With this PR:
- the NPE issue in live storage migration is fixed
- However, vm is corrupted after live storage migration

I would rather get an NPE exception than vm corruption , as data loss is a more critical issue for users.

weizhouapache (changes requested):
@gpordeus 
can you please fix the vm corruption issue in our testing ?

gpordeus:
> @gpordeus I can reproduce the issue @vladimirpetrov faced in his testing
> 
> image of vm B does not have backing file either, but it works (the disk size is much bigger than vm A)

@weizhouapache Great, thank you for testing and for pointing out the disk size. I hadn't been able to replicate it before [last week](https://github.com/apache/cloudstack/pull/7615#issuecomment-1684389067) because I was testing from and to iSCSI and NFS storages, while this error only happens in NFS to NFS.

> @DaanHoogland @rohityadavcloud @gpordeus
> 
> my summary: if template is removed,
> 
> Before this PR,
> 
>     * when migrate volume of stopped vm, it succeeds. vm also works after starting
> 
>     * when migrate a running vm with volumes, got an NPE
>       Failed to migrate VM [VM instance {"id":3,"instanceName":"i-2-3-VM","type":"User","uuid":"1d3541d7-187e-417f-9cd4-cd18d0fc1e3d"}] along with its volumes due to [java.lang.NullPointerException].
> 
> 
> With this PR:
> 
>     * the NPE issue in live storage migration is fixed
> 
>     * However, vm is corrupted after live storage migration
> 
> 
> I would rather get an NPE exception than vm corruption , as data loss is a more critical issue for users.

> @gpordeus can you please fix the vm corruption issue in our testing ?

Yes, I agree and I'm already working on it. The problem is a flag that is passed to libvirt migration only when it's NFS to NFS, for some reason, which makes it so that only the last layer of the vm is copied and, since the template was deleted and wasn't copied first, it corrupts them. Getting rid of the flag makes it work, but I'm testing if there are any other side effects or if there is a better way of going about it.

DaanHoogland:
@gpordeus any progress? can we add this to 4.19?

gpordeus:
> @gpordeus any progress? can we add this to 4.19?
@DaanHoogland Sorry for the delay, it's being a little trickier than I anticipated. It won't take much longer, but not enough to make it to 4.19.

DaanHoogland:
@gpordeus when you get to this, will it include a fix for #5759 ?

DaanHoogland:
@gpordeus should we still try to merge this? (as 4.18 is coming to end of maintemnance)

gpordeus:
> @gpordeus should we still try to merge this? (as 4.18 is coming to end of maintemnance)

@DaanHoogland I suppose not, especially with #8911. I'm closing this, but I'll still be working on #5759. I was sidetracked, but I'll give it priority now.

PR #7617, opened 2023-06-10
Retrieve only resource count on user dashboard

GutoVeronezi:
### Description

The user's dashboard presents the user's amount of running VMs, stopped VMs, total VMs, volumes, networks, and public IP addresses. The APIs `listVirtualMachines`, `listVolumes`, `listNetworks`, and `listPublicIpAddresses` process and return the whole data, which can be slow when dealing with a great number of resources. However, for the user's dashboard context, we just need the count of resources.

This PR intends to create a new parameter (`retrieveonlyresourcecount`) in those APIs to skip the data processing and return only the resource count.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?

I created several VMs in my environment, logged in with a domain admin, and observed how much time the dashboard was taking to completely load. Before the changes, it was taking more than 30 seconds to load and the time would increase according to the amount of resources:

before

image

With the changes it is taking around 2 seconds to completely load:

after

image

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
+++ b/api/src/main/java/org/apache/cloudstack/api/ApiConstants.java
@@ -368,6 +368,7 @@ public class ApiConstants {
     public static final String RESOURCE_TYPE = "resourcetype";
     public static final String RESOURCE_TYPE_NAME = "resourcetypename";
     public static final String RESPONSE = "response";
+    public static final String RETRIEVE_ONLY_RESOURCE_COUNT = "retrieveonlyresourcecount";
     public static final String REVERTABLE = "revertable";
     public static final String REVOKED = "revoked";
     public static final String REGISTERED = "registered";
--- a/api/src/main/java/org/apache/cloudstack/api/BaseListRetrieveOnlyResourceCountCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/BaseListRetrieveOnlyResourceCountCmd.java
@@ -1,0 +1,28 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+package org.apache.cloudstack.api;
+
+import org.apache.commons.lang3.BooleanUtils;
+
+public abstract class BaseListRetrieveOnlyResourceCountCmd extends BaseListTaggedResourcesCmd {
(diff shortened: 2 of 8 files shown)


GutoVeronezi:
@blueorangutan package

shwstppr:
@blueorangutan test

weizhouapache:
@GutoVeronezi 
Is it better to create a separate API to list all resource counts ?

GutoVeronezi:
> @GutoVeronezi Is it better to create a separate API to list all resource counts ?

@weizhouapache, as the APIs we have today already return the data we need; we would just need to improve the processes [with the new parameter]. Therefore, I believe there is no need for a new API [to maintain] that would do the same function as the others.

weizhouapache:
> > @GutoVeronezi Is it better to create a separate API to list all resource counts ?
> 
> @weizhouapache, as the APIs we have today already return the data we need; we would just need to improve the processes [with the new parameter]. Therefore, I believe there is no need for a new API [to maintain] that would do the same function as the others.

@GutoVeronezi 
I think it is better to get all resource counts/limitations by a single API call, not multiple api calls as you can see on the UI source code. You can refer to
https://cloudstack.apache.org/api/apidocs-4.18/apis/listInfrastructure.html
https://cloudstack.apache.org/api/apidocs-4.18/apis/listZonesMetrics.html

GutoVeronezi:
@weizhouapache, creating an API to return a handful of resource counts will limit the users to the data we (the developers) define the API to return; for new use cases, we would have to extend the API to return a new resource count, which makes it a limited approach. 

On the other hand, creating the proposed parameter in those APIs will give users the flexibility to combine the API parameters the way they want. The user's dashboard is just an example of where we can use the parameter; however, we have to remember that users can and will consume the APIs in different ways. Therefore, with that in addition to my initial point, I believe that the parameter creation in this scenario is a better approach rather than creating an API, as it brings flexibility for the users.

weizhouapache:
@GutoVeronezi 
I think an API is better. 

back to your pr, I think it would be better to create an abstract class .

weizhouapache:
@blueorangutan package

DaanHoogland:
@blueorangutan test keepEnv

JoaoJandre (approved):
LGTM, tested manually.

BryanMLima (approved):
CLGTM

DaanHoogland (approved):
clgtm+ small monkeytest

DaanHoogland:
@weizhouapache are your concerns addressed?

GutoVeronezi:
ping @weizhouapache

GutoVeronezi:
Errors in https://github.com/apache/cloudstack/actions/runs/5249960611/jobs/9861114168?pr=7617 does not seem related to this PR; merging this one based on the package results, test results and approval.

rohityadavcloud:
Sorry I'm late to the discussion, this PR wasn't really needed. The information we need is largely in the listAccounts API, which is then used to display the account usage/quota/limits. For ex, see: https://qa.cloudstack.cloud/simulator/#/account/36b4c0a4-b339-40f0-9991-b17282a686a9?tab=resources

A single API call would also save from calling too many API calls for the user/project dashboard. I think we can optimise things in future with this approach.

GutoVeronezi:
@rohityadavcloud, `listAccounts` returns only the resources belonging to the account, which produces a different result than the expected: resources that the account has access to.

Anyways, the PR is not limited to the dashboard case, as I pointed in https://github.com/apache/cloudstack/pull/7617#issuecomment-1586403372, and still is a valid use case that adds to ACS. Also, the PR complied with the community requirements (approval and test passing); therefore, the opinion of a specific person is not mandatory.

Issue #7616, opened 2023-06-09
[4.18] Restarting cloudstack-agent on Rocky Linux or RHEL does not work


weizhouapache:
After updating version.sh (see #7570) on rocky8/kvm nodes, restarting of cloudstack-agent failed with error

```
2023-06-09 14:49:01,657 DEBUG [c.c.a.m.AgentManagerImpl] (AgentConnectTaskPool-9:ctx-f3c5da4e) (logid:7f27c10b) Failed to handle host connection:
java.lang.IllegalArgumentException: Can't add host: 10.0.32.193 with hostOS: Rocky Linux into a cluster,in which there are Rocky hosts added
        at com.cloud.hypervisor.kvm.discoverer.LibvirtServerDiscoverer.createHostVOForConnectedAgent(LibvirtServerDiscoverer.java:474)
        at com.cloud.resource.ResourceManagerImpl.dispatchToStateAdapters(ResourceManagerImpl.java:1997)
        at com.cloud.resource.ResourceManagerImpl.createHostVO(ResourceManagerImpl.java:2224)
        at com.cloud.resource.ResourceManagerImpl.createHostVOForConnectedAgent(ResourceManagerImpl.java:2556)
```

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~

~~~

##### CLOUDSTACK VERSION

~~~

~~~

##### OS / ENVIRONMENT

4.18

##### STEPS TO REPRODUCE

~~~

~~~

##### EXPECTED RESULTS

~~~

~~~

##### ACTUAL RESULTS

~~~

~~~

weizhouapache:
this has been fixed via #7620
thanks @DaanHoogland

PR #7620, opened 2023-06-12
[AGENT] host os accepted if only the first word is defined for the cluster

DaanHoogland:
### Description

This PR...

Fixes: #7616

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
+++ b/server/src/main/java/com/cloud/hypervisor/kvm/discoverer/LibvirtServerDiscoverer.java
@@ -471,8 +471,12 @@ public abstract class LibvirtServerDiscoverer extends DiscovererBase implements 
             String hostOsInCluster = oneHost.getDetail("Host.OS");
             String hostOs = ssCmd.getHostDetails().get("Host.OS");
             if (!hostOsInCluster.equalsIgnoreCase(hostOs)) {
-                throw new IllegalArgumentException("Can't add host: " + firstCmd.getPrivateIpAddress() + " with hostOS: " + hostOs + " into a cluster," +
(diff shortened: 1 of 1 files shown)


weizhouapache:
> @weizhouapache , architecturally I don't whink this is the best solution but it will allow running environments to keep running. I don't know of an example where two different OSses use the same first word in their name, but if that exists, this won't work (i think for our Rocky and Red instances it will work).

@DaanHoogland 
this might work, at least restarting cloudstack-agent should.
when all cloudstack-agent are started, all agent should have the correct Host.OS in host_details

I will test it

another option might be updating existing Host.OS in host_details (mostly rocky/rhel). is alma impacted?

DaanHoogland:
> I will test it

thanks

> another option might be updating existing Host.OS in host_details (mostly rocky/rhel). is alma impacted?

I was thinking of that as well, but we cannot be sure to be complete and correct with this strategy. Rocky and Red make us make assumptions that might not be correct and as you mentioned, I forgot about Alma. This is not a problem btw as it is called "AlmaLinux".

To be sure we would need to check every possible host type that is supported and query "Host.OS" from the host_details tables. Can we be sure about the limits of this?

DaanHoogland:
@blueorangutan package

borisstoyanov:
@blueorangutan test keepEnv

weizhouapache:

weizhouapache (approved):
code lgtm

harikrishna-patnala (approved):
LGTM.

Tried looking for other possible solutions, but this seems to be fair with the existing list of OSes I can think of.

borisstoyanov (approved):
LGTM

PR #7670, opened 2023-06-21
If ResourceCountCheckTask throws an exception the scheduled task is not going to run again until the management servers are restarted.

mprokopchuk:
Added try/catch to log the issue, continue processing of the next resources and running again according to the schedule.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested manually

### DIFF
--- a/server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
+++ b/server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
@@ -1177,7 +1177,7 @@ public class ResourceLimitManagerImpl extends ManagerBase implements ResourceLim
 
             for (ResourceType type : ResourceType.values()) {
                 if (type.supportsOwner(ResourceOwnerType.Domain)) {
-                    recalculateDomainResourceCount(Domain.ROOT_DOMAIN, type);
+                    recalculateDomainResourceCountInContext(Domain.ROOT_DOMAIN, type);
                     for (Domain domain : domains) {
                         recalculateDomainResourceCount(domain.getId(), type);
                     }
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
```diff
@@ -1156,8 +1156,24 @@ public ResourceCountCheckTask() {
...
+                domains = new ArrayList<>();
+            }
+            // try/catch task, otherwise it won't be rescheduled in case of exception
+            try {
+                accounts = _accountDao.findActiveAccountsForDomain(Domain.ROOT_DOMAIN);
+            } catch (Exception e) {
+                s_logger.warn("Resource counters recalculation periodic task failed, unable to fetch active accounts for domain " + Domain.ROOT_DOMAIN, e);
+                // initialize accounts as empty list to do best effort recalculation
+                accounts = new ArrayList<>();
+            }
```
DaanHoogland, line 1176: does it make sense to keep running if one of these retrievals fail (for this pass I mean)?
I like the logging but i think a little further restructuring makes sense; two separate methods, one for domains and one for accounts for instance.
the recursion at lines 1197 and 1204 looks quite strange, if at all it works I would suggest a renaming for readability's sake.

shwstppr:
@blueorangutan package

shwstppr (approved):
LGTM

shwstppr:
@blueorangutan package

server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
mprokopchuk: It is Shadowing in Java nested classes: https://docs.oracle.com/javase/tutorial/java/javaOO/nested.html#shadowing

DaanHoogland: yeah, well this doesn't make the code very readable. If there isn't a very strong design argument to do it this way, please get rid of it.

DaanHoogland (changes requested):

mprokopchuk:
@DaanHoogland renamed methods to recalculateAccountResourceCountInContext and recalculateDomainResourceCountInContext

server/src/main/java/com/cloud/resourcelimit/ResourceLimitManagerImpl.java
mlsorensen:
```
does it make sense to keep running if one of these retrievals fail (for this pass I mean)?
```
Probably, I guess? At least we get the error logged now and if there is a problem with that specific account or domain's data it won't effect processing of the others.

mlsorensen: > does it make sense to keep running if one of these retrievals fail (for this pass I mean)?

Probably, I guess? At least we get the error logged now and if there is a problem with that specific account or domain's data it won't affect processing of the others.

> I like the logging but i think a little further restructuring makes sense; two separate methods, one for domains and one for accounts for instance.
> the recursion at lines 1197 and 1204 looks quite strange, if at all it works I would suggest a renaming for readability's sake.

I think changing the method names to be distinct has been addressed now, can we resolve the conversation?

mlsorensen:
@blueorangutan package

mlsorensen:
@blueorangutan test

mlsorensen:
Let's try again...

@blueorangutan test

DaanHoogland (approved):
clgtm

mlsorensen:
Errors look unrelated to this code change, they're at the KVM hypervisor layer (maybe due to an earlier commit, seems the live volume migration feature is unsupported in whatever host):

```
org.libvirt.LibvirtException: internal error: unable to execute QEMU command \\\'migrate\\\': this feature or command is not currently supported
```

lgtm

mlsorensen (approved):

PR #7671, opened 2023-06-22
Add unit test cases using a secret with slashes to UriUtils

vdombrovski:
### Description

This PR adds URL parsing test cases against real credentials produced by Ceph. One specific target is the infamous slash character should break the existing parser.

I am expecting this PR to fail unit tests, as it is meant to engage a discussion about cleaning up the URI parser.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

N/A

### How Has This Been Tested?

N/A

### DIFF
--- a/utils/src/test/java/com/cloud/utils/UriUtilsTest.java
+++ b/utils/src/test/java/com/cloud/utils/UriUtilsTest.java
@@ -122,6 +122,8 @@ public class UriUtilsTest {
         String url6 = String.format("rbd://%s:3300", host);
         String url7 = String.format("rbd://%s", host);
         String url8 = String.format("rbd://user@%s", host);
+        String url9 = String.format("rbd://user:AQC18pNko3r2GRAAh/3PxIr0lTKMzLL1CAaDww==@%s:3300/pool", host);
+        String url10 = String.format("rbd://user:AQC18pNko3r2GRAAh/3PxIr0lTKMzLL1CAaDww==@%s:3300/pool/volume2", host);
 
         testGetUriInfoInternal(url0, host);
         testGetUriInfoInternal(url1, host);
@@ -132,6 +134,8 @@ public class UriUtilsTest {
         testGetUriInfoInternal(url6, host);
         testGetUriInfoInternal(url7, host);
         testGetUriInfoInternal(url8, host);
+        testGetUriInfoInternal(url9, host);
+        testGetUriInfoInternal(url10, host);
     }
 
     @Test


DaanHoogland:
@blueorangutan package

vdombrovski:
Okay, now the unit test fails exactly where it was supposed to:

https://github.com/apache/cloudstack/actions/runs/5342685611/jobs/9685337907?pr=7671#step:7:3864

@DaanHoogland how do you think we should proceed? Unfortunately I'm not comfortable writing code, not sure I'll be able to provide much more help than this.

DaanHoogland:
@vdombrovski create a new issue, label it for 4.18.1 (provisionally) and add any other label that seem appropriate. Unfortunately you will depend on good will or a fellow victim of the issue. No guarantees :(

If they appear they will have to copy your code or get access to your branch btw.

vdombrovski:
Alright, issue created, let's wait and see. Thank you for the help.

vdombrovski:
In case this helps, here is the regexp we're using to test for the secret in our automation (can also be used for properly parsing the URI I believe):

```rbd:\/\/([\w\-\.]+):([A-Za-z0-9+/=]+)@([^:]+):(\d+)/([\w_.]+)/?([\w_.]+)?```

**Comment by vdombrovski** [comment]
@weizhouapache I'm not working on a fix; as stated previously, I'm not really comfortable in making such bigger changes to the Java code. I've added the regexp as is, simply because we are already using it as a check in our automation (outside Cloudstack).

In case someone would be willing to make the required changes to this parser function (pretty sure it's this one) to make the new testcase pass, it would be quite nice and we'd be able to close this PR/Issue:

https://github.com/apache/cloudstack/blob/0466bf9d4713ab9221a4f4ae8a257996dd6cd8c6/utils/src/main/java/com/cloud/utils/UriUtils.java#L706

**Comment by weizhouapache** [comment]
> In case this helps, here is the regexp we're using to test for the secret in our automation (can also be used for properly parsing the URI I believe):
> 
> `rbd:\/\/([\w\-\.]+):([A-Za-z0-9+/=]+)@([^:]+):(\d+)/([\w_.]+)/?([\w_.]+)?`

@vdombrovski 
regarding the password, does `([A-Za-z0-9+/=]+)` cover all possible values ?

**Comment by vdombrovski** [comment]
> > In case this helps, here is the regexp we're using to test for the secret in our automation (can also be used for properly parsing the URI I believe):
> > `rbd:\/\/([\w\-\.]+):([A-Za-z0-9+/=]+)@([^:]+):(\d+)/([\w_.]+)/?([\w_.]+)?`
> 
> @vdombrovski regarding the password, does `([A-Za-z0-9+/=]+)` cover all possible values ?

Yes, because Ceph passwords are base64 encoded.

However judging by other test cases (which may or may not be realistic), I believe that no single regexp is able to match all of them. So I would guess either maching repeatedly against different regular expressions or keeping a more "traditional" failover mechanism might be the answer here.

**Comment by vdombrovski** [comment]
Okay, now I don't know if this project is AI-friendly, but here's a regexp generated with the help of GPT that will consistenly match all testcases except the one with a user but no password (e.g. `rbd://user@host` which to be honest might be a non-realistic test case). Tested against IPv4, RFC1035, IPv6 hosts, and base64 passwords.

```
String regex = "rbd:\\/\\/(?:(\\w+):([A-Za-z0-9+/=]+)@)?([\\[\\]:\\w.-]+)(?::(\\d+))?(?:\\/([\\w]+)(?:\\/([\\w]+))?)?";
```

The capture groups (if they exist) will be as follows:
```
Capture Group 1: user
Capture Group 2: password
Capture Group 3: host
Capture Group 4: 3300
Capture Group 5: pool
Capture Group 6: volume2
```

Please tell me if any use of AI is prohibited; I will delete this comment.

weizhouapache:
> Okay, now I don't know if this project is AI-friendly, but here's a regexp generated with the help of GPT that will consistenly match all testcases except the one with a user but no password (e.g. `rbd://user@host` which to be honest might be a non-realistic test case). Tested against IPv4, RFC1035, IPv6 hosts, and base64 passwords.
> 
> ```
> String regex = "rbd:\\/\\/(?:(\\w+):([A-Za-z0-9+/=]+)@)?([\\[\\]:\\w.-]+)(?::(\\d+))?(?:\\/([\\w]+)(?:\\/([\\w]+))?)?";
> ```
> 
> The capture groups (if they exist) will be as follows:
> 
> ```
> Capture Group 1: user
> Capture Group 2: password
> Capture Group 3: host
> Capture Group 4: 3300
> Capture Group 5: pool
> Capture Group 6: volume2
> ```
> 
> Please tell me if any use of AI is prohibited; I will delete this comment.

@vdombrovski 
hmm, great. thanks

weizhouapache:
@vdombrovski 
I have tried to add a RBD pool into cloudstack, it succeeded.
CloudStack UI converted "/"to "_", and passed to cloudstack API.
so I removed the ceph pool and re-added it with the new secret (/ to _), and it succeeded as well.

The problem is, if password contains "/", `Java.net.URI` class cannot handle it correctly.
If you cannot add ceph RBD pool with "/" in the token, please replace `/` with `_` and retry

vdombrovski:
> @vdombrovski I have tried to add a RBD pool into cloudstack, it succeeded. CloudStack UI converted "/"to "_", and passed to cloudstack API. so I removed the ceph pool and re-added it with the new secret (/ to _), and it succeeded as well.
> 
> The problem is, if password contains "/", `Java.net.URI` class cannot handle it correctly. If you cannot add ceph RBD pool with "/" in the token, please replace `/` with `_` and retry

@weizhouapache I believe your test is incorrect: the URL gets added  (even if it is invalid, it's an issue by itself actually); but it fails in libvirtd (which is why I'm calling this a silent error). We have tested this with multiple Ceph installs, we have the exact same issue everytime a slash character is present, and yes we are deploying using API only (again, nowhere is it stated that the API is actually broken and that we can't use the tools that use the ACS API).

Furthermore, asking the users to replace characters in URL manually is awful and frustrating in terms of user experience (especially as it is not properly documented). Why not asking them to provision the secret directly in libvirtd while we are at it? This example might go too far, but isn't one of the goals of this projectl to provide a easy to use and simple experience to cloud administrators and users?

Finally, I'm pretty sure that Java.net.URI is not the culprit here, as your function signature returns a custom struct (UriInfo) which has nothing to do with this lib. Also, in no way you're constrained into using Java.net.URI in your code, thus I don't see a valid reasoning here.

The way I see it, there are two main ways to fix this:

1. Improving the parser (e.g. via the regexp provided above) to handle all cases
2. Sanitizing the password input inside the API call (instead of the UI), to replace the slash character with underscore, which, and updating the unit test with passwords containing underscores.

I'll leave it up to the community to decide which implementation is best in this case.

weizhouapache:
> > @vdombrovski I have tried to add a RBD pool into cloudstack, it succeeded. CloudStack UI converted "/"to "_", and passed to cloudstack API. so I removed the ceph pool and re-added it with the new secret (/ to _), and it succeeded as well.
> > The problem is, if password contains "/", `Java.net.URI` class cannot handle it correctly. If you cannot add ceph RBD pool with "/" in the token, please replace `/` with `_` and retry
> 
> @weizhouapache I believe your test is incorrect: the URL gets added (even if it is invalid, it's an issue by itself actually); but it fails in libvirtd (which is why I'm calling this a silent error). We have tested this with multiple Ceph installs, we have the exact same issue everytime a slash character is present, and yes we are deploying using API only (again, nowhere is it stated that the API is actually broken and that we can't use the tools that use the ACS API).
> 
> Furthermore, asking the users to replace characters in URL manually is awful and frustrating in terms of user experience (especially as it is not properly documented). Why not asking them to provision the secret directly in libvirtd while we are at it? This example might go too far, but isn't one of the goals of this projectl to provide a easy to use and simple experience to cloud administrators and users?
> 
> Finally, I'm pretty sure that Java.net.URI is not the culprit here, as your function signature returns a custom struct (UriInfo) which has nothing to do with this lib. Also, in no way you're constrained into using Java.net.URI in your code, thus I don't see a valid reasoning here.
> 
> The way I see it, there are two main ways to fix this:
> 
> 1. Improving the parser (e.g. via the regexp provided above) to handle all cases
> 2. Sanitizing the password input inside the API call (instead of the UI), to replace the slash character with underscore, which, and updating the unit test with passwords containing underscores.
> 
> I'll leave it up to the community to decide which implementation is best in this case.

@vdombrovski 
what I provided in my previous comment is a workaround for your issue, not the final solution. 
I have tested it and it worked as expected. could you please test it (on UI or API ) ?

vdombrovski:
Another workaround is to use a Ceph password without slashes.

I'll test this on the next occasion, for now I don't have a means to do it.

weizhouapache:
> Another workaround is to use a Ceph password without slashes.
> 
> I'll test this on the next occasion, for now I don't have a means to do it.

@vdombrovski 
I have created a PR https://github.com/apache/cloudstack/pull/7708
Please feel free to review and test it

vdombrovski:
Sure, the fix looks good. Thank you for your help, closing this one.

Issue #7560, opened 2023-05-25
All templates from different zones and hypervisors are availible for reinstall VM


top-secrett:
##### ISSUE TYPE

 * Bug Report

##### CLOUDSTACK VERSION

~~~
4.18.0.0
~~~

##### CONFIGURATION

Two anvanced zones:
- xcp-ng
- VMware

##### SUMMARY

We have two zones with different hypervisors, but templates from all zones are available for each other when reinstall VM
![image](https://github.com/apache/cloudstack/assets/92435079/3d85c132-3299-47af-af35-390a108b34aa)
![image](https://github.com/apache/cloudstack/assets/92435079/29ef1658-8f81-42b8-b018-4298f68cfe6c)
![image](https://github.com/apache/cloudstack/assets/92435079/f85df94d-ba71-46a5-8d7a-710c277a2096)
![image](https://github.com/apache/cloudstack/assets/92435079/5569f606-a121-4f09-b95f-cd281b8411f3)

kiranchavala:
@DaanHoogland  this looks like  a bug as the  listTemplates api call is not passing the zone id when a user tries to restore a vm (https://cloudstack.apache.org/api/apidocs-4.18/apis/restoreVirtualMachine.html)

I think we should list only templates that are active in the zone and also based on the vm zone id 

I have tried this with a normal user account and it lists all the templates

https://cloudstack.apache.org/api/apidocs-4.18/apis/listTemplates.html

http://<>:8080/client/api/?listall=true&templatefilter=executable&showicon=true&command=listTemplates&response=json

DaanHoogland:
@top-secrett if you can please test #7739?

top-secrett:
@DaanHoogland sorry, but I don't know how to test it

DaanHoogland:
> @DaanHoogland sorry, but I don't know how to test it

ok, we'll have to wait till someone else does then

zap51:
@DaanHoogland I tested, LGTM. Template present in different zones are no longer shown during reinstall.

DaanHoogland:
fixed in #7739

PR #7739, opened 2023-07-11
Filter templates in vue

DaanHoogland:
### Description

This PR...

Fixes: #7560 

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested manually in the UI on and environment with two zones with different hypervisors.

An alternative way of testing could be testing in a single zone with two different hypervisor types or in two zones with the same hypervisor type. In both cases the list of available templates to choose from should be restricted to the current zone and hypervisor type. I.E. and template that is of a different hypervisortype or not available in the zone the VM is in should not be offered as a choice in the drop-down.

### DIFF
--- a/ui/src/config/section/compute.js
+++ b/ui/src/config/section/compute.js
@@ -165,8 +165,11 @@ export default {
           dataView: true,
           args: ['virtualmachineid', 'templateid'],
           filters: (record) => {
-            var filters = []
-            filters.push({ key: 'templateid', value: { hypervisortype: record.hypervisor } })
+            var filters = {}
+            var filterParams = {}
+            filterParams.hypervisortype = record.hypervisor
+            filterParams.zoneid = record.zoneid
+            filters.templateid = filterParams
             return filters
           },
           show: (record) => { return ['Running', 'Stopped'].includes(record.state) },
--- a/ui/src/config/section/compute.js
+++ b/ui/src/config/section/compute.js
@@ -164,6 +164,11 @@ export default {
           message: 'message.reinstall.vm',
           dataView: true,
           args: ['virtualmachineid', 'templateid'],
+          filters: (record) => {
+            var filters = []
+            filters.push({ key: 'templateid', value: { hypervisortype: record.hypervisor } })
+            return filters
+          },
           show: (record) => { return ['Running', 'Stopped'].includes(record.state) },
           mapping: {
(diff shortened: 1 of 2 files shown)


DaanHoogland:
@blueorangutan ui

DaanHoogland:
> @DaanHoogland Shouldn't we also fix the api call as well? In my view, those templates shouldn't be visible in the front end at all.

No, i don´t see why @soreana. The user has rights to see those templates, it is just that these are not applicable for restoring the VM.

soreana:
So 

> > @DaanHoogland Shouldn't we also fix the api call as well? In my view, those templates shouldn't be visible in the front end at all.
> 
> No, i don´t see why @soreana. The user has rights to see those templates, it is just that these are not applicable for restoring the VM.

Yeah I agree that this is a default behaviour and user can see those. My question is what would be the point of showing those templates to user while they can't use them :D

DaanHoogland:
> So
> 
> > > @DaanHoogland Shouldn't we also fix the api call as well? In my view, those templates shouldn't be visible in the front end at all.
> > 
> > 
> > No, i don´t see why @soreana. The user has rights to see those templates, it is just that these are not applicable for restoring the VM.
> 
> Yeah I agree that this is a default behaviour and user can see those. My question is what would be the point of showing those templates to user while they can't use them :D

none, that is why this PR (and why #7560 was reported) . Not sure I get your comment @soreana .

kiranchavala:
@blueorangutan package

kiranchavala (approved):
LGTM

Tested manually and only the zone specific templates are visible when you try to reinstall a vm

DaanHoogland:
second test verification reported at https://github.com/apache/cloudstack/issues/7560#issuecomment-1656584114

weizhouapache:
code lgtm

weizhouapache:
@DaanHoogland 
it looks UI test failed after merging this
https://github.com/apache/cloudstack/actions/runs/5710998969/job/15472424976

DaanHoogland:
see #7792 

> @DaanHoogland it looks UI test failed after merging this https://github.com/apache/cloudstack/actions/runs/5710998969/job/15472424976

PR #7746, opened 2023-07-18
make control.cidr dynamic

DaanHoogland:
### Description

This PR attempts to make choosing local ip choices adhere to a more dynamics control.cidr so people can actually choose their control network....

Fixes: #6715

### Types of changes

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial


DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

engine/components-api/src/main/java/com/cloud/configuration/ConfigurationManager.java
```diff
@@ -54,10 +55,30 @@
...
+            Boolean.class,
+            "system.vm.use.local.storage",
+            "Advanced",
+            "false",
+            "Indicates whether to use local storage pools or shared storage pools for system VMs.",
+            false,
+            ConfigKey.Scope.Zone,
+            null);
+
+    ConfigKey<String> ControlCidr = new ConfigKey<>(
```
harikrishna-patnala, line 73: we need to add this config to getConfigKeys() method @DaanHoogland 
https://github.com/shapeblue/cloudstack/blob/1275db4081aa8b09148909d6e6b65bc377d4992f/server/src/main/java/com/cloud/configuration/ConfigurationManagerImpl.java#L7630

server/src/main/java/com/cloud/network/guru/ControlNetworkGuru.java
```diff
@@ -21,10 +21,12 @@
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 
-import org.apache.cloudstack.framework.config.dao.ConfigurationDao;
+import com.cloud.network.NetworkModel;
```
harikrishna-patnala, line 24: reorder this please

DaanHoogland:
@blueorangutan package

GutoVeronezi:
@DaanHoogland, there are some refactorings being done that are not related to the PR proposal; would not be better to create a new PR for the refactoring and separate the contexts?

GutoVeronezi:
> > @DaanHoogland, there are some refactorings being done that are not related to the PR proposal; would not be better to create a new PR for the refactoring and separate the contexts?
> 
> I could do @GutoVeronezi , but I find that I do these refactoring for undestanding the software and the problem. Thus the changes are usually intertwined a bit. I gather you ask this because it makes reviewing harder?
> 
> (btw, i still have to think of a way to verify this change)

In this scale of changes (~250 lines), the reviewing still is quite simple. Personally, I like to separate the contexts when possible, to better organize and track the changes. It is just a suggestion, though. :grin:

harikrishna-patnala (approved):
Code LGTM

Rubueno:
Thanks for spending time to work on this Daan! LGTM

DaanHoogland:
> Thanks for spending time to work on this Daan! LGTM

Did you test it @Rubueno ?

Rubueno:
> > Thanks for spending time to work on this Daan! LGTM
> 
> Did you test it @Rubueno ?

No, not yet. I've not had the time to do so. Pretty swamped!

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test keepEnv

server/src/main/java/com/cloud/configuration/Config.java
```diff
@@ -885,14 +885,6 @@ public enum Config {
...
             null),
-    ControlCidr(
-            "Advanced",
-            ManagementServer.class,
-            String.class,
-            "control.cidr",
-            "169.254.0.0/16",
-            "Changes the cidr for the control network traffic.  Defaults to using link local.  Must be unique within pods",
-            null),
     ControlGateway("Advanced", ManagementServer.class, String.class, "control.gateway", "169.254.0.1", "gateway for the control network traffic", null),
```
weizhouapache, line 888: @DaanHoogland 
should ControlGateway be dynamic as well ?

server/src/main/java/com/cloud/network/guru/ControlNetworkGuru.java
```diff
@@ -215,18 +215,6 @@ public boolean configure(String name, Map<String, Object> params) throws Configu
...
 
-        _cidr = dbParams.get(Config.ControlCidr.toString());
-        if (_cidr == null) {
-            _cidr = NetUtils.getLinkLocalCIDR();
-        }
-
-        _gateway = dbParams.get(Config.ControlGateway.toString());
-        if (_gateway == null) {
-            _gateway = NetUtils.getLinkLocalGateway();
-        }
```
weizhouapache, line 226: @DaanHoogland 
should _gateway be set ?

server/src/main/java/com/cloud/configuration/Config.java
DaanHoogland: I'll add it (and move it as well)

engine/components-api/src/main/java/com/cloud/configuration/ConfigurationManager.java
```diff
@@ -20,6 +20,7 @@
 import java.util.Map;
 import java.util.Set;
 
+import net.sf.ehcache.management.ManagementServer;
```
weizhouapache, line 23: this is not required

server/src/main/java/com/cloud/network/guru/ControlNetworkGuru.java
weizhouapache: ok, looks good

engine/components-api/src/main/java/com/cloud/configuration/ConfigurationManager.java
DaanHoogland: hm, how did that get in?

DaanHoogland:
@blueorangutan test

DaanHoogland:
@Rubueno @wido I gave this another look and it is not going to work. The link local addresses are created on createPod and changing the configurations will break any SVM created after changing the settings. We'll need a more extensive disign. Can you give some input on what a valid scenario for you would be?

thanks

wido:
> @Rubueno @wido I gave this another look and it is not going to work. The link local addresses are created on createPod and changing the configurations will break any SVM created after changing the settings. We'll need a more extensive disign. Can you give some input on what a valid scenario for you would be?
> 
> thanks

We should move to IPv6 Link-Local at all for the SSVM :-) These addresses with IPv4 aren't needed at all ;-) Makes life much easier, I promise!

Anyway, this isn't the biggest problem now.

In RFC5549 the IPv4 address 169.254.0.1 is used for routing IPv4 traffic via IPv6, something which you will use when using EVPN+BGP+VXLAN.

Therefor you want to change the control CIDR to something else. It's 'stupid' that we allocate a /16 to this as that's not needed at all, a /20 by default should be more then enough.

Any existing installations work just fine. Maybe we should just change the default from this /16 to a /20 (169.254.240.0/20) and be done with it for now.

weizhouapache:
> > @Rubueno @wido I gave this another look and it is not going to work. The link local addresses are created on createPod and changing the configurations will break any SVM created after changing the settings. We'll need a more extensive disign. Can you give some input on what a valid scenario for you would be?
> > thanks
> 
> We should move to IPv6 Link-Local at all for the SSVM :-) These addresses with IPv4 aren't needed at all ;-) Makes life much easier, I promise!
> 
> Anyway, this isn't the biggest problem now.
> 
> In RFC5549 the IPv4 address 169.254.0.1 is used for routing IPv4 traffic via IPv6, something which you will use when using EVPN+BGP+VXLAN.
> 
> Therefor you want to change the control CIDR to something else. It's 'stupid' that we allocate a /16 to this as that's not needed at all, a /20 by default should be more then enough.
> 
> Any existing installations work just fine. Maybe we should just change the default from this /16 to a /20 (169.254.240.0/20) and be done with it for now.

thanks @wido 

@DaanHoogland I think the main problem for this PR is, all link local IPs are saved in database table `op_dc_link_local_ip_address_alloc` which is not needed at all. 
We need to save only the allocated IPs in the table and remove all unused IPs.
Some classes need to be updated to support the IP allocation from cidr/gateway, instead of picking up IPs from `op_dc_link_local_ip_address_alloc` table.

wido:
> > > @Rubueno @wido I gave this another look and it is not going to work. The link local addresses are created on createPod and changing the configurations will break any SVM created after changing the settings. We'll need a more extensive disign. Can you give some input on what a valid scenario for you would be?
> > > thanks
> > 
> > 
> > We should move to IPv6 Link-Local at all for the SSVM :-) These addresses with IPv4 aren't needed at all ;-) Makes life much easier, I promise!
> > Anyway, this isn't the biggest problem now.
> > In RFC5549 the IPv4 address 169.254.0.1 is used for routing IPv4 traffic via IPv6, something which you will use when using EVPN+BGP+VXLAN.
> > Therefor you want to change the control CIDR to something else. It's 'stupid' that we allocate a /16 to this as that's not needed at all, a /20 by default should be more then enough.
> > Any existing installations work just fine. Maybe we should just change the default from this /16 to a /20 (169.254.240.0/20) and be done with it for now.
> 
> thanks @wido
> 
> @DaanHoogland I think the main problem for this PR is, all link local IPs are saved in database table `op_dc_link_local_ip_address_alloc` which is not needed at all. We need to save only the allocated IPs in the table and remove all unused IPs. Some classes need to be updated to support the IP allocation from cidr/gateway, instead of picking up IPs from `op_dc_link_local_ip_address_alloc` table.

And in addition I think we should opt for a /20 by default for all new installations. If we do that, this whole problem is kind of solved. For new deployments the /20 would be used and never get into the way of BGP with v4 via v6 routing.

engine/components-api/src/main/java/com/cloud/configuration/ConfigurationManager.java
```diff
@@ -54,10 +55,40 @@
...
+            "Indicates whether to use local storage pools or shared storage pools for system VMs.",
+            false,
+            ConfigKey.Scope.Zone,
+            null);
+
+    ConfigKey<String> ControlCidr = new ConfigKey<>(
+            String.class,
+            "control.cidr",
+            "Advanced",
+            "169.254.0.0/16",
```
DaanHoogland, line 77:
```suggestion
            "169.254.240.0/20",
```
like this and abandon all other changes in this PR, right @wido ?

wido: Almost @DaanHoogland , but also in:

- NetUtils.java
- AgentProperties.java
- configuration/Config.java

If we change it to a /20 in those files we should be good.

DaanHoogland: This will only work for new environments. I don't think this will work when upgrading environment. These hardcoded values of an already otherwise deplayed config item will break systems and needs a bit more TLC .

weizhouapache: @DaanHoogland 
as I understand, Wido wants /20 to be the default in new environments.
It would be good to update the config description to emphasize that the new value only applies on new pods.

DaanHoogland: well, this means that the hardcoded values in NetUtils.java and AgentProperties.java have to be intelligent... we can not assume returning the one or the other in these methods. That complicates things a bit.

DaanHoogland:
i disagree with this PR myself

Issue #7700, opened 2023-06-28
VM Settings validation blocks custom key value pairs


rajujith:
##### ISSUE TYPE

  * Improvement Request
 
##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.17,4.18 mostly related to the new UI. 
~~~

##### SUMMARY

Currently, the VM settings in the UI validate (not in API) the parameters. It allows only a known/predefined list of parameters in the drop-down available. However, this is restrictive for advanced users and more use cases. Custom parameter/key-value pairs could be allowed if required to display a message that it is not in the predefined list and the user is expected to make sure it is supported etc. 

##### STEPS TO REPRODUCE

~~~
1. Stop a VM
2. Try adding a new parameter via the VM settings tab.
~~~

##### EXPECTED RESULTS

~~~
VM settings UI could allow custom key-value pairs. 
~~~

##### ACTUAL RESULTS

~~~
VM settings UI doesn't allow custom key-value pairs. 

~~~

![Screenshot 2023-06-28 at 11 38 11 AM](https://github.com/apache/cloudstack/assets/5229147/ebda81dc-8238-4faa-82cd-e8bd8c931574)

weizhouapache:
fixed by #7793

weizhouapache:
thanks @DaanHoogland for fixing it

Issue #6937, opened 2022-12-01
Extraconfig values cannot be added for VMs that are shut down


Arceoavs:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VM Management
~~~

##### CLOUDSTACK VERSION

~~~
4.17.1.0
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

When a user (with admin rights or without) creates an extraconfig under Settings for a VM that is switched off, one cannot enter a value in the Value input field.
The UI refuses any keyboard input and the browser displays error messages. 

![image](https://user-images.githubusercontent.com/16841883/205068736-2b3a5ae3-fbd4-4735-903c-71b86b267eea.png)

##### STEPS TO REPRODUCE

~~~
Create a VM.
Switch it off. 
Go to Settings and try to add a configuration. 
Look at the Browser console. 
~~~

##### EXPECTED RESULTS

~~~
Values can be set. 
~~~

##### ACTUAL RESULTS

~~~
Values cannot be set. 
~~~

Output in the Browser console:
```
config and theme applied [App.vue:40:12](webpack:///src/App.vue)
Object { data: {…}, status: 401, statusText: "Unauthorized", headers: {…}, config: {…}, request: XMLHttpRequest }
[request.js:37:12](webpack:///src/utils/request.js)
Some cookies are misusing the recommended “SameSite“ attribute 22
Cookie “rl_anonymous_id” has been rejected because it is already expired. [api](https://sparci-cloudstack.uni-koblenz.de/client/api/?command=logout&response=json)
Cookie “rl_page_init_referrer” has been rejected because it is already expired. [api](https://sparci-cloudstack.uni-koblenz.de/client/api/?command=logout&response=json)
Cookie “rl_page_init_referring_domain” has been rejected because it is already expired. [api](https://sparci-cloudstack.uni-koblenz.de/client/api/?command=logout&response=json)
Cookie “rl_user_id” has been rejected because it is already expired. [api](https://sparci-cloudstack.uni-koblenz.de/client/api/?command=logout&response=json)
Cookie “rl_trait” has been rejected because it is already expired. [api](https://sparci-cloudstack.uni-koblenz.de/client/api/?command=logout&response=json)
Object { data: {…}, status: 432, statusText: "432", headers: {…}, config: {…}, request: XMLHttpRequest }
[request.js:37:12](webpack:///src/utils/request.js)
[...]

PR #7793, opened 2023-07-31
UI: allow new keys for VM details

DaanHoogland:
### Description

This PR...

~~Fixes: #6937~~
Fixes: #7700 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/ui/src/components/view/DetailSettings.vue
+++ b/ui/src/components/view/DetailSettings.vue
@@ -170,7 +170,11 @@ export default {
         return []
       }
       if (!Array.isArray(this.detailOptions[this.newKey])) {
-        return { value: this.detailOptions[this.newKey] }
+        if (this.detailOptions[this.newKey]) {
+          return { value: this.detailOptions[this.newKey] }
+        } else {
+          return ''
+        }
       }
       return this.detailOptions[this.newKey].map(value => {
         return { value: value }


harikrishna-patnala:
@DaanHoogland does this mean, we now allow empty values for the settings?

weizhouapache:
@DaanHoogland 
after some thinking, I think it is not a UI bug, but a missing feature.

actually when I add a vm setting "extraconfig", it failed
```
Failed to update VM, due to: 'extraconfig' should not be included in details as key
```

in the EditVM dialog, we need to add a new field "extraconfig"

DaanHoogland:
> @DaanHoogland does this mean, we now allow empty values for the settings?

good question. If the API accepts it we do, i guess.

harikrishna-patnala (commented):
what I observed with this fix is

1. We are not allowing empty values even after the fix. The observed fix is, we are now handling this properly in UI without any errors in the debugger console. In my opinion, this is good.
image

2. We could now add extra custom details which are not predefined. Not sure if this is the intention of this fix, but this is also good, there is another issue raised to allow custom keys https://github.com/apache/cloudstack/issues/7700
image

3. Specific to the key value "extraconfig" mentioned in the original issue, intentionally are not allowing it in the backend, not sure about the reason.
https://github.com/apache/cloudstack/blob/981d74825a87727d7f8e397cad75877986c96d79/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java#L2772-L2774

@DaanHoogland am I correct with these points ? specifically point 1, it is contradicting the subject

DaanHoogland:
> what I observed with this fix is
> 
>     1. We are not allowing empty values even after the fix. The observed fix is, we are now handling this properly in UI without any errors in the debugger console. In my opinion, this is good.

yes

>     2. We could now add extra custom details which are not predefined. Not sure if this is the intention of this fix, but this is also good, there is another issue raised to allow custom keys [VM Settings validation blocks custom key value pairs #7700](https://github.com/apache/cloudstack/issues/7700)

yes, it is

>     3. Specific to the key value "extraconfig" mentioned in the original issue, intentionally we are not allowing it in the backend, not sure about the reason.

functionally i am not sure, but the intention was to allow any key. I did not test with this particular string :face_with_head_bandage: .

> 
> @DaanHoogland am I correct with these points ? specifically point 1, it is contradicting the subject

I will change the subject.

@weizhouapache @rohityadavcloud @vishesh92 @nvazquez , what do and do we not  want to allow?

nvazquez:
Hi @DaanHoogland @harikrishna-patnala, I think the reason the 'extraconfig' key is not allowed is because it is a parameter on the deployVirtualMachine API. In case the user passes extra config, those are persisted as VM details with the key 'extraconfig' which can cause issues

weizhouapache:
> 
> So I think what we can do here is (if at all we decide to allow extraconfig from UI) to check if the detail key is "extraconfig" and then send the value in API parameter "extraconfig" (not the "details").

this looks like a good solution. @harikrishna-patnala @DaanHoogland

nvazquez:
The 'extraconfig' parameter was introduced by this feature: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Enable+additional+configuration+metadata+to+virtual+machines. It lacks UI support, so was only possible to set it in the API. I think that it can be a good partial solution as the way the extra config is stored on the vm_details table is using either named keys such as: extraconfig-NAME1, extraconfig-NAME2 or numerical: extraconfig-1, extraconfig-2. So the updateVirtualMachine API may allow editing the first time but not the second time as the details will be split into multiple details with the extraconfig prefix (all this needs to be tested, I had tested it long ago :) )

weizhouapache:
> The 'extraconfig' parameter was introduced by this feature: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Enable+additional+configuration+metadata+to+virtual+machines. It lacks UI support, so was only possible to set it in the API. I think that it can be a good partial solution as the way the extra config is stored on the vm_details table is using either named keys such as: extraconfig-NAME1, extraconfig-NAME2 or numerical: extraconfig-1, extraconfig-2. So the updateVirtualMachine API may allow editing the first time but not the second time as the details will be split into multiple details with the extraconfig prefix (all this needs to be tested, I had tested it long ago :) )

thanks @nvazquez for the info

it looks we need to investigate more on #6937 before. sorry @DaanHoogland , I thought it is a simple bug fix

harikrishna-patnala:
> > The 'extraconfig' parameter was introduced by this feature: https://cwiki.apache.org/confluence/display/CLOUDSTACK/Enable+additional+configuration+metadata+to+virtual+machines. It lacks UI support, so was only possible to set it in the API. I think that it can be a good partial solution as the way the extra config is stored on the vm_details table is using either named keys such as: extraconfig-NAME1, extraconfig-NAME2 or numerical: extraconfig-1, extraconfig-2. So the updateVirtualMachine API may allow editing the first time but not the second time as the details will be split into multiple details with the extraconfig prefix (all this needs to be tested, I had tested it long ago :) )
> 
> thanks @nvazquez for the info
> 
> it looks we need to investigate more on #6937 . sorry @DaanHoogland , I thought it is a simple bug fix

@weizhouapache this PR can still be merged as this is fixing the other issue https://github.com/apache/cloudstack/issues/7700 to allow custom keys, also this PR handles the errors that are coming in the debugger console. We can handle #6937 separately.

weizhouapache:
> 

ok @harikrishna-patnala ,thanks for pointing it out
this lgtm

weizhouapache (approved):
tested ok

harikrishna-patnala (approved):
LGTM. I've verified the fix for the issue #7700

weizhouapache:
> LGTM. I've verified the fix for the issue #7700

thanks @harikrishna-patnala . merging

PR #6420, opened 2022-05-30
Enable CPU cap for SSVMs and CPVMs

JoaoJandre:
### Description

Currently it is not possible to create CPU-capped SSVMs or CPVMs, even if the system offering used by these system VMs has this enabled. This is due to the fact that, when creating and persisting SSVMs and CPVMs in the database, the CPU cap configuration of the system offering is not being respected and the value false is always used for the CPU usage limitation option. This behavior has been corrected, so that it is now possible to deploy a CPU-limited SSVM or CPVM, according to the settings of the system offering.

In addition, the standard system offerings for System VMs has been updated so that the CPU cap is enabled, as keeping system VMs without CPU usage limits opens up space for different types of infrastructure attacks/misuses.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
To test this patch, I ran the SQL script, then checked the DB to see if everything is as expected; then, I recreated the SSVM and CPVM and checked if they were configured with CPU limits.

PR #6970, opened 2022-12-10
cpvm: fix cpu usage for console vm when using vnc over websockets

alexandru-bagu:
### Description
On a VMware ESXi 7 environment the Console VM would always use 100% cpu after one connection and it would not stop until a VM restart (or well cloud service restart inside of the VM).

This PR fixes this issue by adding a Thread.sleep(1). 

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

PR #7808, opened 2023-08-03
engine/schema: fix CPU cap limitation for upgraded environment

rohityadavcloud:
CPU cap limitation was enabled as part of
https://github.com/apache/cloudstack/pull/6420 that changes behaviour for existing environments.

The CPU cap limitation on KVM causes systemvms to not start or be really slow in nested and virtualised environments. We shouldn't change the defaults, but this could be documented or an operator can create new system offerings and use them.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

In KVM env, post upgrade SSVM operations were slow and a couple of times SSVM didn't start. By setting the value to false, no cpu share/limitation was put on new SSVMs and they could start and work. On git history check, found this to be a recent change introduced between 4.17->4.18 via #6420. This should be reverted.

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-41800to41810.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41800to41810.sql
@@ -31,3 +31,8 @@ CALL ADD_GUEST_OS_AND_HYPERVISOR_MAPPING (6, 'Windows Server 2022 (64-bit)', 'VM
 CALL ADD_GUEST_OS_AND_HYPERVISOR_MAPPING (6, 'Windows Server 2022 (64-bit)', 'VMware', '8.0', 'windows2019srvNext_64Guest');
 CALL ADD_GUEST_OS_AND_HYPERVISOR_MAPPING (6, 'Windows Server 2022 (64-bit)', 'VMware', '8.0.0.1', 'windows2019srvNext_64Guest');
 CALL ADD_GUEST_OS_AND_HYPERVISOR_MAPPING (6, 'Windows Server 2022 (64-bit)', 'Xenserver', '8.2.0', 'Windows Server 2022 (64-bit)');
+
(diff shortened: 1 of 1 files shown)


rohityadavcloud:
@blueorangutan package

rohityadavcloud:
@blueorangutan package

harikrishna-patnala (approved):
Code LGTM.

agree with the point, one can create new system offerings and use them.

rohityadavcloud:
@blueorangutan test

DaanHoogland (commented):
code looks good, @JoaoJandre please comment on this.

rohityadavcloud:
@blueorangutan test alma8 kvm-alma8

rohityadavcloud:
@blueorangutan test alma8 kvm-alma8

JoaoJandre:
Sorry for the delayed response @dahn,  I somehow missed the notification. 

I am against reverting #6420, as was explained there, keeping system VMs without CPU usage limits opens up space for different types of infrastructure attacks/misuses. Also, bugs like the ones fixed on #7826 and #6970 will affect much more the Cloud without a CPU cap on CPVMs, for example. 

The problem reported by @rohityadavcloud, where system VMs fail to start, is caused by another situation. When a system VM is provisioned, ACS sends commands to the system VM to generate the CSR and subsequently generate the certificate; these commands have a timeout of 60 seconds (hardcoded). When the CPU cap is enabled for the system VMs, the request and certificate generation procedures can take more than 60 seconds each to complete, causing a timeout in the execution of the commands and the non-provisioning of certificates for the system VMs. Thus, generating a handshake failure when the `ca.plugin.root.auth.strictness setting is true`. This config should be externalized.

Also, I agree with @harikrishna-patnala, if the system VMs are too slow, you could just create new system offerings and use them. But the default offerings should protect users from attacks and possible bugs or misuses like the ones mentioned.

Issue #7794, opened 2023-07-31
System VMs not starting with CentOS 7 hosts


harikrishna-patnala:
##### ISSUE TYPE

 * Bug Report
 * Improvement Request
 * Enhancement Request
 * Feature Idea
 * Documentation Report
 * Other

##### COMPONENT NAME

~~~

~~~

##### CLOUDSTACK VERSION

~~~

~~~

##### SUMMARY

Tried creating a new environment with CentOS 7 and Systems VMs failed to start with the below error 

2023-07-31 11:15:56,455 INFO [c.c.v.VirtualMachineManagerImpl] (Work-Job-Executor-55:ctx-2c1e5436 job-2/job-582 ctx-5d943e87) (logid:8d9e353e) Unable to start VM on Host {"id":1,"name":"pr7769-t7215-kvm-centos7-kvm1","type":"Routing","uuid":"d12436b0-8fb1-46ad-85c0-582f786fe180"} due to unsupported configuration: deflate-on-oom is not supported by this QEMU binary

# rpm -qa |grep qemu
qemu-img-1.5.3-175.el7_9.6.x86_64
qemu-kvm-1.5.3-175.el7_9.6.x86_64
qemu-kvm-common-1.5.3-175.el7_9.6.x86_64
libvirt-daemon-driver-qemu-4.5.0-36.el7_9.5.x86_64

This PR adds new hot memory ballooning capabilities, may be we need version checks there.

##### STEPS TO REPRODUCE

~~~

~~~

##### EXPECTED RESULTS

~~~
System VMs should come up and running
~~~

##### ACTUAL RESULTS

~~~
System VMs start attempt failed
~~~

kiranchavala:
Observing the same issue, Able to reproduce the issue 

@DaanHoogland @rajujith

DaanHoogland:
@kiranchavala @harikrishna-patnala , I am investigating the issue. We don't set `deflate-on-oom` explicitely on the xml defenition. We do set a memballoon attribute `autodeflate`, so I am trying to see if I can manipulate that in our favour. Will update here.

weizhouapache:
> @kiranchavala @harikrishna-patnala , I am investigating the issue. We don't set `deflate-on-oom` explicitely on the xml defenition. We do set a memballoon attribute `autodeflate`, so I am trying to see if I can manipulate that in our favour. Will update here.

@DaanHoogland 
any update ?

DaanHoogland:
[...]

Issue #7443, opened 2023-04-18
Smoketests are failing if run on Ubuntu 22.04 and python version (3.10.6)


kiranchavala:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Testing
~~~

##### CLOUDSTACK VERSION
N/A

##### CONFIGURATION
N/A

##### OS / ENVIRONMENT

Ubuntu 22.0.4 

##### SUMMARY

Smoketests are failing if run on Ubuntu 22.04 and python version (3.10.6)

In the #6978 , Dockerfile was updated to use Ubuntu 22.0.4

In Ubuntu 22.04 the python version is 

root@6c66e082ef0c:~# python --version 
Python 3.10.6

In Ubuntu 20.0.4 the python version is 

root@6c6890001b5b:~# python --version 
Python 3.8.10

##### STEPS TO REPRODUCE

Deploy latest docker image for the cloudstack-simulator which based on Ubuntu 22.0.4 

1. docker run --name simulator -p 8080:5050 -d apache/cloudstack-simulator:4.19.0.0-SNAPSHOT-20230411-063947

Deploy a datacenter

2. docker exec -it simulator python /root/tools/marvin/marvin/deployDataCenter.py -i /root/setup/dev/advanced.cfg

Run a sample smoke test 

3. docker exec -it simulator  nosetests -v --with-marvin   --marvin-config=setup/dev/advanced.cfg   --with-xunit   --xunit-file=xunit.xml   -a tags=advanced,required_hardware=false   --zone=Sandbox-simulator   --hypervisor=simulator    test/integration/smoke/test_vm_life_cycle.py

~~~

==== Marvin Init Started ====

=== Marvin Parse Config Successful ===

=== Marvin Setting TestData Successful===

==== Log Folder Path: /tmp/MarvinLogs/Apr_17_2023_06_56_34_IHX5TS All logs will be available here ====

=== Marvin Init Logging Successful===

==== Marvin Init Successful ====
Traceback (most recent call last):
  File "/usr/local/bin/nosetests", line 8, in <module>
    sys.exit(run_exit())
  File "/usr/local/lib/python3.10/dist-packages/nose/core.py", line 118, in __init__
    unittest.TestProgram.__init__(
  File "/usr/lib/python3.10/unittest/main.py", line 100, in __init__
    self.parseArgs(argv)
  File "/usr/local/lib/python3.10/dist-packages/nose/core.py", line 179, in parseArgs
[...]

PR #7663, opened 2023-06-21
KVM: fix SSVM starting when overprovisioning memory

DaanHoogland:
### Description

This PR bypasses overprovisioning settings on memory for SSVMs to make sure these can always start.

In addition this PR adds autodeflate to memballoon on KVM VMs and sets the memory to `memory` instead of `currentMemory`. It implement sugestions by @rajujith at https://github.com/apache/cloudstack/issues/7430#issuecomment-1600245780. This solution seems not to fixs overprovisioning issues in VMs and a new PR will have to be created for a definite solution on that.

Fixes: #7430

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

PR #7810, opened 2023-08-03
Allow KVM overcommit to work without reducing minimum VM memory when vm ballooning is disabled

### LINKED ISSUE #7430: Virtual Router VM fails to boot with kernel panic when overprovisioinig is > 1
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Virtual Router VM / service offering / shared storage
~~~

##### CLOUDSTACK VERSION

~~~
4.18
~~~

##### CONFIGURATION

Advanced Zone with NFS primary Storage.

##### OS / ENVIRONMENT

Ubuntu Focal Fossa

##### SUMMARY

When overprovisioning factors are > 1, the virtual router vm for a new zone is not able to boot and ends up in a kernel panic while trying to mount the root volume.
By default a local storage service offering is picked for the router.
If an id of a service offering that uses shared storage is specified in the global settings under router.service.offering it is ignored.
Somehow the Volumes attached to the working system and router vms (when overprovisioning == 1) are not accessible via the UI and lead to 404.
Taking a look at the db entry they look like:

                        id: 26
                account_id: 1
                 domain_id: 1
                   pool_id: 4
              last_pool_id: NULL
               instance_id: 21
                 device_id: 0
                      name: ROOT-21
                      uuid: 29537b97-e6d1-484b-9e06-92cd905ead8e
                      size: 5242880000
                    folder: NULL
                      path: 29537b97-e6d1-484b-9e06-92cd905ead8e
                    pod_id: NULL
            data_center_id: 4
                iscsi_name: NULL
                   host_ip: NULL
               volume_type: ROOT
                 pool_type: NULL
          disk_offering_id: 12
               template_id: 3
               first_snapshot_backup_uuid: NULL
               recreatable: 1
                   created: 2023-04-10 11:51:38
                  attached: NULL
                   updated: 2023-04-10 11:51:40
                   removed: NULL
                     state: Ready
                chain_info: NULL
              update_count: 2
                 disk_type: NULL
    vm_snapshot_chain_size: NULL
                    iso_id: NULL
            display_volume: 1
                    format: QCOW2
                  min_iops: NULL
                  max_iops: NULL
             hv_ss_reserve: NULL
         provisioning_type: thin
             external_uuid: NULL
             passphrase_id: NULL
            encrypt_format: NULL

The disk offering for the volume is type "shared".

In addition there is a second primary storage that was setup by accident, always comes up again (after disabling), but has no visible resources attached to it.

##### STEPS TO REPRODUCE

Set overprovisioning to > 1 and create a new VPC.

~~~

~~~

##### EXPECTED RESULTS

~~~
The virtual router vm should use the shared storage offering and boot.
~~~

##### ACTUAL RESULTS

~~~
It ends up ina kernel panic mounting the root volume.
~~~

rohityadavcloud:
This makes the KVM guru pass the max memory to KVM agent by picking the max. memory from the VM transfer object. This lets the cluster/agent decide if VM should start with the min or max memory depending on whether vm.memballoon.disable is false (default) or true.

When vm.memballoon.disable is true, VMs start with max. memory. Right now the max memory isn't sent to the KVM agent causing the regression. This feature is defined at https://github.com/apache/cloudstack/blob/main/agent/conf/agent.properties#L220

Fixes: #7794

Also
Fixes #7430 if user sets `vm.memballoon.disable` to true in agent.properties.

Doc PR to clarify this further at https://github.com/apache/cloudstack-documentation/pull/337

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2780,15 +2780,10 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
 
         grd.setMemBalloning(!_noMemBalloon);
 
-        long maxRam = ByteScaleUtils.bytesToKibibytes(vmTO.getMaxRam());
-        long currRam = vmTO.getType() == VirtualMachine.Type.User ? getCurrentMemAccordingToMemBallooning(vmTO, maxRam) : maxRam;
-
-        if (s_logger.isTraceEnabled()) {
-            s_logger.trace(String.format("memory values for VM %s are %d/%d",vmTO.getName(),maxRam, currRam));
-        }
+        Long maxRam = ByteScaleUtils.bytesToKibibytes(vmTO.getMaxRam());
 
         grd.setMemorySize(maxRam);
-        grd.setCurrentMem(currRam);
+        grd.setCurrentMem(getCurrentMemAccordingToMemBallooning(vmTO, maxRam));
 
         int vcpus = vmTO.getCpus();
         Integer maxVcpus = vmTO.getVcpuMaxLimit();
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2795,11 +2795,17 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     }
 
     protected long getCurrentMemAccordingToMemBallooning(VirtualMachineTO vmTO, long maxRam) {
-        if (_noMemBalloon) {
-            s_logger.warn(String.format("Setting VM's [%s] current memory as max memory [%s] due to memory ballooning is disabled. If you are using a custom service offering, verify if memory ballooning really should be disabled.", vmTO.toString(), maxRam));
+        if (_noMemBalloon || (vmTO != null && vmTO.getType() != VirtualMachine.Type.User)) {
+            if (_noMemBalloon) {
+                s_logger.warn(String.format("Setting VM's [%s] current memory as max memory [%s] due to memory ballooning is disabled. If you are using a custom service offering, verify if memory ballooning really should be disabled.", vmTO.toString(), maxRam));
+            } else {
+                s_logger.warn(String.format("Setting System VM's [%s] current memory as max memory [%s].", vmTO.toString(), maxRam));
+            }
             return maxRam;
         } else {
-            return ByteScaleUtils.bytesToKibibytes(vmTO.getMinRam());
+            long minRam = ByteScaleUtils.bytesToKibibytes(vmTO.getMinRam());
+            s_logger.debug(String.format("Setting VM's [%s] current memory as min memory [%s] due to memory ballooning is enabled.", vmTO.toString(), minRam));
+            return minRam;
         }
     }
 
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -2795,18 +2795,17 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
     }
 
     protected long getCurrentMemAccordingToMemBallooning(VirtualMachineTO vmTO, long maxRam) {
-        if (_noMemBalloon || (vmTO != null && vmTO.getType() != VirtualMachine.Type.User)) {
-            if (_noMemBalloon) {
-                s_logger.warn(String.format("Setting VM's [%s] current memory as max memory [%s] due to memory ballooning is disabled. If you are using a custom service offering, verify if memory ballooning really should be disabled.", vmTO.toString(), maxRam));
-            } else {
-                s_logger.warn(String.format("Setting System VM's [%s] current memory as max memory [%s].", vmTO.toString(), maxRam));
-            }
-            return maxRam;
+        long retVal = maxRam;
+        if (_noMemBalloon) {
+            s_logger.warn(String.format("Setting VM's [%s] current memory as max memory [%s] due to memory ballooning is disabled. If you are using a custom service offering, verify if memory ballooning really should be disabled.", vmTO.toString(), maxRam));
+        } else if (vmTO != null && vmTO.getType() != VirtualMachine.Type.User) {
+            s_logger.warn(String.format("Setting System VM's [%s] current memory as max memory [%s].", vmTO.toString(), maxRam));
         } else {
             long minRam = ByteScaleUtils.bytesToKibibytes(vmTO.getMinRam());
             s_logger.debug(String.format("Setting VM's [%s] current memory as min memory [%s] due to memory ballooning is enabled.", vmTO.toString(), minRam));
-            return minRam;
+            retVal = minRam;
         }
+        return retVal;
     }
 
     /**
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
@@ -282,7 +282,7 @@ public class LibvirtVMDef {
         @Override
         public String toString() {
             StringBuilder response = new StringBuilder();
-            response.append(String.format("<memory>%s</memory>\n", this.memory));
+            response.append(String.format("<memory>%s</memory>\n", this.currentMemory));
             response.append(String.format("<currentMemory>%s</currentMemory>\n", this.currentMemory));
 
             if (this.memory > this.currentMemory) {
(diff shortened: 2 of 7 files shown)


rohityadavcloud:
@weizhouapache in that case let me grab Daan's workaround to not apply min memory for systemvms

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2780,15 +2780,10 @@ protected GuestResourceDef createGuestResourceDef(VirtualMachineTO vmTO){
...
-        long currRam = vmTO.getType() == VirtualMachine.Type.User ? getCurrentMemAccordingToMemBallooning(vmTO, maxRam) : maxRam;
-
-        if (s_logger.isTraceEnabled()) {
-            s_logger.trace(String.format("memory values for VM %s are %d/%d",vmTO.getName(),maxRam, currRam));
-        }
+        Long maxRam = ByteScaleUtils.bytesToKibibytes(vmTO.getMaxRam());
 
         grd.setMemorySize(maxRam);
-        grd.setCurrentMem(currRam);
+        grd.setCurrentMem(getCurrentMemAccordingToMemBallooning(vmTO, maxRam));
```
rohityadavcloud, line 2786: I just realised the maxRam seetting is fixed as in the KVMGuru the correct maxRam is passed now.
cc @DaanHoogland @weizhouapache let's test

rohityadavcloud:
I did whole lot of testing; the VM ballooning feature isn't dynamic or automatic like I thought earlier and the only feasible feature is to have dynamic scaling of VMs (that's manual); so as a best compromise this PR tries to do the following;

1. Pass the maxMem to KVMGuru that sends the right max. memory to KVM agent
2. When memory ballooning is disable or when the VM isn't a User VM, then use the maxMemory (new/tweaked behaviour)
3. When memory balloning is enabled and VM is a User VM type, then use min memory (old behaviour)

Ideally I think the ballooning feature should be disabled esp if overcommit is used; and use should log or document such as case for operators/users to be aware of (i.e. what to expect from the memory ballooning feature in case of KVM, and its behaviour with memory overcommit enabled).

@blueorangutan package

rohityadavcloud:
@blueorangutan package

rohityadavcloud:
@blueorangutan test matrix

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
```diff
@@ -1238,7 +1238,7 @@ public String getMemBalloonStatsPeriod() {
         @Override
         public String toString() {
             StringBuilder memBalloonBuilder = new StringBuilder();
-            memBalloonBuilder.append("<memballoon model='" + memBalloonModel + "' autodeflate='on'>\n");
+            memBalloonBuilder.append("<memballoon model='" + memBalloonModel + "'>\n");
```
harikrishna-patnala, line 1241: @rohityadavcloud does this mean, the blocker issue https://github.com/apache/cloudstack/issues/7794 is fixed here?

rohityadavcloud: Possibly - we need to check/test cc @DaanHoogland

rohityadavcloud:
@DaanHoogland @weizhouapache I've raised a doc PR pl help review if along with the doc PR we cover all bases - https://github.com/apache/cloudstack-documentation/pull/337

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
DaanHoogland: @harikrishna-patnala , yes as shown by 
https://github.com/apache/cloudstack/pull/7810#issuecomment-1664869520

DaanHoogland:
I think we should merge this and re-open #7430 / re-implement #7663 if needed.

DaanHoogland:
@blueorangutan test keepEnv

harikrishna-patnala (approved):
Code LGTM

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2800,11 +2795,17 @@ protected GuestResourceDef createGuestResourceDef(VirtualMachineTO vmTO){
     }
 
     protected long getCurrentMemAccordingToMemBallooning(VirtualMachineTO vmTO, long maxRam) {
-        if (_noMemBalloon) {
-            s_logger.warn(String.format("Setting VM's [%s] current memory as max memory [%s] due to memory ballooning is disabled. If you are using a custom service offering, verify if memory ballooning really should be disabled.", vmTO.toString(), maxRam));
+        if (_noMemBalloon || (vmTO != null && vmTO.getType() != VirtualMachine.Type.User)) {
```
rohityadavcloud, line 2798: @DaanHoogland https://github.com/apache/cloudstack/issues/7430 is addressed here :)

rohityadavcloud:
Thanks @DaanHoogland if you see my comment, we need not re-open the previous issue. Ask @weizhouapache to review and lgtm/advise;

This PR handles the following cases:
1. Doc PR addresses what to expect, what/how to configure KVM host env when overcommiting memory - https://github.com/apache/cloudstack-documentation/pull/337
2. For non-User VMs, we use max memory and in KVM guru the correct max memory is passed (fixes #7443)
3. Fixes #7794 as centos7/el7 smoketests passing now https://github.com/apache/cloudstack/pull/7810#issuecomment-1664869520
4. Reverts/ensures we don't disable memory balloon feature for KVM

I think this covers most of the cases and doc PR handles advising users the best practice they deem fit for what they're doing.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
```diff
@@ -2800,11 +2795,17 @@ protected GuestResourceDef createGuestResourceDef(VirtualMachineTO vmTO){
...
+                s_logger.warn(String.format("Setting System VM's [%s] current memory as max memory [%s].", vmTO.toString(), maxRam));
+            }
             return maxRam;
         } else {
-            return ByteScaleUtils.bytesToKibibytes(vmTO.getMinRam());
+            long minRam = ByteScaleUtils.bytesToKibibytes(vmTO.getMinRam());
+            s_logger.debug(String.format("Setting VM's [%s] current memory as min memory [%s] due to memory ballooning is enabled.", vmTO.toString(), minRam));
+            return minRam;
         }
     }
```
DaanHoogland, line 2810: for simplification and readability i'd make this
```suggestion
    protected long getCurrentMemAccordingToMemBallooning(VirtualMachineTO vmTO, long maxRam) {
        long retVal = maxRam;
        if (_noMemBalloon) {
            s_logger.warn(String.format("Setting VM's [%s] current memory as max memory [%s] due to memory ballooning is disabled. If you are using a custom service offering, verify if memory ballooning really should be disabled.", vmTO.toString(), maxRam));
        } else if (vmTO != null && vmTO.getType() != VirtualMachine.Type.User) {
            s_logger.warn(String.format("Setting System VM's [%s] current memory as max memory [%s].", vmTO.toString(), maxRam));
        } else {
            long minRam = ByteScaleUtils.bytesToKibibytes(vmTO.getMinRam());
            s_logger.debug(String.format("Setting VM's [%s] current memory as min memory [%s] due to memory ballooning is enabled.", vmTO.toString(), minRam));
            retVal = minRam;
        }
        return retVal;
    }
```

rohityadavcloud: this might break unit tests as-is; feel free to push changes :)

DaanHoogland: ok, i'll be a dare devil (we can always revert)

DaanHoogland: not accepted by github, will leave for prosterity.

DaanHoogland:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

DaanHoogland:
@rohityadavcloud, #7430 is fixed in this PR. When memory overprovisioning is set to 3 on the cluster:

```
  <maxMemory slots='16' unit='KiB'>524288</maxMemory>
  <memory unit='KiB'>174080</memory>
  <currentMemory unit='KiB'>174080</currentMemory>
```

when set to 5

```
  <maxMemory slots='16' unit='KiB'>524288</maxMemory>
  <memory unit='KiB'>104448</memory>
  <currentMemory unit='KiB'>104448</currentMemory>
```

DaanHoogland:
@blueorangutan test matrix

rohityadavcloud:
Thanks for testing @DaanHoogland I suppose that's true if the vm divide setting is enabled for user-vms; for systemvms the memory would be consistently the same (max memory).

DaanHoogland:
> Thanks for testing @DaanHoogland I suppose that's true if the vm divide setting is enabled for user-vms; for systemvms the memory would be consistently the same (max memory).

Yes @rohityadavcloud , I can test that as well, but we already showed that SVMs start which is what is what matters. Errors in the kvm version seem unrelated, merging.

PR #7846, opened 2023-08-10
ui: Fix refresh and re-route behaviour

### LINKED ISSUE #7840: UI bug: internal "refresh" of the UI can cause a wrong VM to be deleted
##### ISSUE TYPE

 * Bug Report
 
##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
4.17.2
~~~

##### CONFIGURATION

No special configuration needed.

##### OS / ENVIRONMENT

NA

##### SUMMARY

Start VM deployment process and leave the VM being deployed (it takes 5+ seconds)
Go immediately and list e.g. single VPC details (chose VPC and click on it), wait for a bit until the previous API call (to deploy VM) is completed, and the current view (a specific VPC view) will be replaced with the VM view (the VM which you have deployed).
This is not critical.

**However, a user searching for VM (many results, so all API calls in background take time to complete), then quickly clicking the VM, clicking the delete button - can result in user deleting the wrong VM - see the video demonstraion here: https://screenpal.com/watch/c0j6j7VkPJC**

**DUE TO THIS, I MARK IT AS CRITICAL.**

##### STEPS TO REPRODUCE

See the Summary above, see the video.
Beside fixing the "refresh of UI after previous API call completion" general problem - it's critical to reproduce the deletion bug issue, to be able to confirm it is fixed later and not rely on general fix and hope it fixes the deletion issue.
I assume that introducing additional latency (Chrome can do this in Developer Tools) will help make sure that API calls takes long time to complete, so it should be doable to reproduce the issue (I'm still working on reproducing the deletion bug)

##### EXPECTED RESULTS

~~~

~~~

##### ACTUAL RESULTS

~~~

~~~

rohityadavcloud:
This introduces change in AutogenView to do the following:

- Cancel old list API requests when the route is changed (between different list view or list and resource views)
- In the handler of an API call made in the fetchData(), to check if the json received is for the current route/id (API)
- Add console log for debugging which may be removed if this passes code review and testing

Fixes #7840

How was this tested?
- In my env the listVirtualMachines API takes time to complete, so I go to the instances page and go quick back/forth other views; for example, go to instances page that doesn't load and click on 2nd VM (or go to instance page and search something that returns faster than the whole list API and go the VM). When the original list API call completes, the current VM is replaced by the first VM in the list.
- Check console log with the fix and network to see that old API (xhr) requests that haven't completed are cancelled when the route/view is changed between list and resource views of different resources

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -622,7 +622,7 @@ export default {
     next()
   },
   beforeRouteLeave (to, from, next) {
-    console.log('DEBUG - Due to route change, cancelling any on-going API request', this.apiName)
+    console.log('DEBUG - Due to route change, ignoring results for any on-going API request', this.apiName)
     sourceToken.cancel()
     sourceToken.init()
     this.currentPath = this.$route.fullPath
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -442,6 +442,7 @@ import { ref, reactive, toRaw } from 'vue'
 import { api } from '@/api'
 import { mixinDevice } from '@/utils/mixin.js'
 import { genericCompare } from '@/utils/sort.js'
+import { sourceToken } from '@/utils/request'
 import store from '@/store'
 import eventBus from '@/config/eventBus'
 
@@ -621,6 +622,9 @@ export default {
     next()
   },
   beforeRouteLeave (to, from, next) {
+    console.log('DEBUG - Due to route change, cancelling any on-going API request', this.apiName)
+    sourceToken.cancel()
+    sourceToken.init()
     this.currentPath = this.$route.fullPath
     next()
   },
@@ -918,18 +922,36 @@ export default {
             break
           }
         }
-        this.itemCount = 0
+        var apiItemCount = 0
         for (const key in json[responseName]) {
           if (key === 'count') {
-            this.itemCount = json[responseName].count
+            apiItemCount = json[responseName].count
             continue
           }
           objectName = key
           break
         }
+
+        if ((this.apiName.toLowerCase() + 'response') !== responseName.toLowerCase()) {
+          console.log('DEBUG - Discarding API response as expected API response key does not match response name:', responseName)
+          return
+        }
+        if ('id' in this.$route.params && this.$route.params.id !== params.id) {
+          console.log('DEBUG - Discarding API response as its `id` does not match $route.param.id')
+          return
+        }
+        if (this.dataView && apiItemCount > 1) {
+          console.log('DEBUG - Discarding API response as got more than one item in data view', this.$route.params, this.items)
+          return
+        }
+
         this.items = json[responseName][objectName]
         if (!this.items || this.items.length === 0) {
           this.items = []
+        }
+        this.itemCount = apiItemCount
+        if (this.itemCount !== this.items.length) {
+          console.log('WARN: API items length does not match the API return count, something is wrong')
         }
 
         if (['listTemplates', 'listIsos'].includes(this.apiName) && this.items.length > 1) {
@@ -979,6 +1001,10 @@ export default {
           }
         }
       }).catch(error => {
+        if (!error || !error.message) {
+          console.log('API request likely got cancelled due to route change:', this.apiName)
+          return
+        }
         if ([401].includes(error.response.status)) {
           return
         }
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -933,8 +933,7 @@ export default {
         }
 
         if ((this.apiName.toLowerCase() + 'response') !== responseName.toLowerCase()) {
-          console.log('DEBUG - Discarding API response as expected API response key does not match response name:', responseName)
-          return
+          console.log('DEBUG - Found that API', this.apiName, ' does not match its expected response key name (apiname+response): ', responseName)
         }
         if ('id' in this.$route.params && this.$route.params.id !== params.id) {
           console.log('DEBUG - Discarding API response as its `id` does not match $route.param.id')
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -932,11 +932,8 @@ export default {
           break
         }
 
-        if ((this.apiName.toLowerCase() + 'response') !== responseName.toLowerCase()) {
-          console.log('DEBUG - Found that API', this.apiName, ' does not match its expected response key name (apiname+response): ', responseName)
-        }
         if ('id' in this.$route.params && this.$route.params.id !== params.id) {
-          console.log('DEBUG - Discarding API response as its `id` does not match $route.param.id')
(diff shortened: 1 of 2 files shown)


ui/src/views/AutogenView.vue
```diff
@@ -918,19 +922,37 @@ export default {
...
+          return
+        }
+
         this.items = json[responseName][objectName]
         if (!this.items || this.items.length === 0) {
           this.items = []
         }
+        this.itemCount = apiItemCount
+        if (this.itemCount !== this.items.length) {
+          console.log('WARN: API items length does not match the API return count, something is wrong')
```
rohityadavcloud, line 954: this is one part that needs testing, I observed it a few times

rohityadavcloud:
@blueorangutan ui
cc @borisstoyanov @vladimirpetrov @kiranchavala @NuxRo

DaanHoogland:
@rohityadavcloud , I'll test but I have my doubt because of the following; If you reroute it will not help to cancel any outstanding APIs. These are still being executed and I am unsure of the value of the concept of "cancel". Can you explain that?
There is only a problem if the user will start an action (fire a new call) fr wich it can not know the changed parameters.

ui/src/views/AutogenView.vue
```diff
@@ -621,6 +622,9 @@ export default {
     next()
   },
   beforeRouteLeave (to, from, next) {
+    console.log('DEBUG - Due to route change, cancelling any on-going API request', this.apiName)
```
shwstppr, line 625: @rohityadavcloud when it says cancelling on-going API request, we are just canceling the response handling of the API call?

DaanHoogland: thanks @shwstppr , that was confusing me.

rohityadavcloud:
@blueorangutan ui

ui/src/views/AutogenView.vue
DaanHoogland:
```suggestion
    console.log('DEBUG - Due to route change, ignoring results for any on-going API request', this.apiName)
```

rohityadavcloud:
@blueorangutan ui

ui/src/views/AutogenView.vue
rohityadavcloud: The log doesn't matter, I'll remove it before it gets merged; see PR description `Add console log for debugging which may be removed if this passes code review and testing
`

rohityadavcloud:
cc @shwstppr @Pearl1594 I'll need some help to fix the failing UI unit tests - any tips or assistance you can offer?

rohityadavcloud:
cc @nvazquez - if you can advise/assist on the failing unit tests, thanks.

DaanHoogland:
@blueorangutan ui

DaanHoogland (approved):
code looks good, testing will be a bit of a hack unless you test against a busy production env. I'll give that a show

DaanHoogland:
> code looks good, testing will be a bit of a hack unless you test against a busy production env. I'll give that a show

I have not reliably reproduced this. Only once in about ten attempts did I get to have a detail load in the version without the patch that didn't seem the right one, and never with this patch. I am ok with merging, no side effects were happening.

weizhouapache (approved):
code lgtm

not tested

DaanHoogland:
@weizhouapache , I did test it and did not encounter the error, but as said I could not **reliably** reproduce without the patch either.
I think we can merge though.

rohityadavcloud:
All, I need assistance on manually testing/validating this cc @NuxRo @DaanHoogland @borisstoyanov @vladimirpetrov @kiranchavala @weizhouapache (any one?)
And, assistance on fixing the UI unit tests cc @shwstppr @nvazquez @Pearl1594

shwstppr:
@rohityadavcloud I tried fixing the unit test but it didn't work as of now. I tried adding a mock for the newly added sourceToken. Will add an update if it works.

rohityadavcloud:
Thanks @shwstppr much appreciated

rohityadavcloud:
@blueorangutan ui

rohityadavcloud:
@blueorangutan package
@blueorangutan ui

DaanHoogland:
@blueorangutan test keepEnv

GutoVeronezi:
I was able to reproduce the situation with the current version; I will try to test the patch in the next few days.

Though, I agree with @DaanHoogland about the "cancel the old API requests" message; we should make it explicity to the operators and users that the backend will continue to process the request, and that the UI will just ignore the results.

rohityadavcloud:
There is no ` "cancel the old API requests" message` I'm not sure where this got picked up from this PR? There are a few console.log msgs for debugging/testing purposes, before merging the PR if it passes - those should be removed.

@blueorangutan package
@blueorangutan ui

GutoVeronezi:
> There is no ` "cancel the old API requests" message` I'm not sure where this got picked up from this PR? There are a few console.log msgs for debugging/testing purposes, before merging the PR if it passes - those should be removed.

@rohityadavcloud the PR's description says:

> This introduces change in AutogenView to do the following:
> 
> * Cancel old API requests when the route is changed (between different list view or list and resource views)

This leads one who is reading to think that the requests will be canceled.

rohityadavcloud:
@GutoVeronezi ah, I see. I've updated the description to make more sense - the proposed code only cancels the list API related to the list or resource view, not that of any async/action or non-async actions that maybe on-going, this happens when a user quickly changes view (for ex. the listVirtualMachinesMetrics may be slow and user don't wait for the instances list view to render completely and decides to navigate to templates or networks tab). The list API response is what renders the list view (table) or the resource view. And, there is no message prompted to the user.

rohityadavcloud:
@blueorangutan package
@blueorangutan ui

DaanHoogland:
@blueorangutan package

DaanHoogland:
all seems fine but one strange thing I noticed is that an "Add VPC" notification remained "in progress"
image
 while the VPC was allready added
image

DaanHoogland:
I think this is no issue in this code. can we merge @weizhouapache @rohityadavcloud ?

rohityadavcloud:
@DaanHoogland that could a notification regression or somehow things got opened in a new tab / refreshed, I've noticed that too (but not related to changes in autogenview)

rohityadavcloud:
@DaanHoogland I've asked @andrijapanicsb to confirm testing, perhaps @GutoVeronezi was testing as well - appreciate all the QA, feedback we can get on this. Thanks.

weizhouapache:
> I think this is no issue in this code. can we merge @weizhouapache @rohityadavcloud ?

@DaanHoogland 
thanks for testing
I am ok with merging .

andrijapanicsb:
Wait for it... give me 1h

andrijapanicsb (approved):
LGTM

DaanHoogland:
ok, merging. @GutoVeronezi if you find anything please open a new issue on that? thanks

PR #7853, opened 2023-08-10
Fix role escalation prevention

hsato03:
### Description

While checking an account access to prevent the role escalation, the `checkAccess` method is using both `DynamicRoleBasedAPIAccessChecker` and `StaticRoleBasedAPIAccessChecker` (legacy) classes and that is causing some inconsistences in the validations, consequently allowing the role escalation.

For this reason, the `StaticRoleBasedAPIAccessChecker` was disabled if `DynamicRoleBasedAPIAccessChecker` is enabled.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

- I have created a domain admin custom role (DA2) and allowed all API's to it using the `*` regex;
- Removed an API that it had in common with default domain admin role;
- Tried to create a default domain admin account using a DA2 account;
- Received an exception from the role eslacation prevention.

### DIFF
--- a/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
+++ b/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
@@ -107,6 +107,10 @@ public class StaticRoleBasedAPIAccessChecker extends AdapterBase implements APIA
 
     @Override
     public boolean checkAccess(Account account, String commandName) {
+        if (isEnabled()) {
+            return true;
+        }
+
         RoleType roleType = accountService.getRoleType(account);
         if (isApiAllowed(commandName, roleType)) {
             return true;
--- a/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
+++ b/plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
@@ -76,12 +76,12 @@ public class StaticRoleBasedAPIAccessChecker extends AdapterBase implements APIA
         if (roleService.isEnabled()) {
             LOGGER.debug("RoleService is enabled. We will use it instead of StaticRoleBasedAPIAccessChecker.");
         }
-        return roleService.isEnabled();
+        return !roleService.isEnabled();
     }
 
     @Override
     public List<String> getApisAllowedToUser(Role role, User user, List<String> apiNames) throws PermissionDeniedException {
-        if (isEnabled()) {
+        if (!isEnabled()) {
             return apiNames;
         }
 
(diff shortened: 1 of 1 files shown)


rohityadavcloud:
Hi @hsato03 thanks for the PR. The dynamic checker is used by default and we don't ship commands.properties with cloudstack for several years now, how can we test/verify your claims?

hsato03:
Hello @rohityadavcloud. You can test it by doing the same tests I did. Before the changes, the account will be created and no exception will be thrown.
- Create a domain admin custom role (DA2) and allow all APIs to it using the `*` regex;
- Still in DA2, remove an API that it has in common with the default domain admin role (`listProjects` for an example);
- Login into a DA2 account and try to create an account that has the default domain admin role;
- The account should not be created as the caller role (DA2) does not have a permission (`listProjects`) that the role's account being created has (default domain admin).

kiranchavala:
@blueorangutan  package

plugins/acl/static-role-based/src/main/java/org/apache/cloudstack/acl/StaticRoleBasedAPIAccessChecker.java
```diff
@@ -107,6 +107,10 @@ public boolean checkAccess(User user, String commandName) throws PermissionDenie
 
     @Override
     public boolean checkAccess(Account account, String commandName) {
+        if (isEnabled()) {
+            return true;
+        }
```
DaanHoogland, line 112: if this prevents the escalation talked about fine, but it seems illogical to return true to prevent access. would return false not make more sense?

hsato03: @DaanHoogland I don't think so. Actually this `isEnabled` method checks if `DynamicRoleBasedAPIAccessChecker` is enabled. Returning true here is just passing the check responsibility to another class.

DaanHoogland: yes @hsato03 I think you are right. I consider it a bug that the StaticRoleBasedAPIAccessChecker.isEnabled() would return true if it is disabled (but the dynamic one is enabled). Maybe this is a discussion out of scope for your change. As said, "if this prevents the escalation talked about (it is) fine"

weizhouapache: yes, it looks like a bug @DaanHoogland 
we should change the return value of `isEnabled` in this class

DaanHoogland: cc @rohityadavcloud ^

weizhouapache (approved):
code lgtm

It should be better to change the return value of method `isEnabled`, but it might have large impact.

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

weizhouapache:
@hsato03 has changed the return value of `isEnabled` method in the class, the code lgtm and trillian test results look good. do you have any other concerns ?

PR #7859, opened 2023-08-11
server: set Default NIC when VM has no default NIC

hsato03:
### Description

ACS allows you to add NICs to a VM without them, although it is not possible to set them as default since the `updateDefaultNicForVirtualMachine` API only updates the default NIC if a default NIC already exists, otherwise an exception is thrown.

For this reason, when a NIC is added and the VM that the NIC is being created for has no default NIC, the NIC will be marked as default.

### Types of changes

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Imported a VM without NICs, added some NICs and the first one was set as default.

### DIFF
--- a/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
+++ b/engine/orchestration/src/main/java/org/apache/cloudstack/engine/orchestration/NetworkOrchestrator.java
@@ -4329,7 +4329,9 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
         if (nic == null || vmProfile.getType() == VirtualMachine.Type.User) {
             final int deviceId = _nicDao.getFreeDeviceId(vm.getId());
 
-            nic = allocateNic(requested, network, false, deviceId, vmProfile).first();
+            boolean isDefaultNic = getNicProfileDefaultNic(requested);
+
+            nic = allocateNic(requested, network, isDefaultNic, deviceId, vmProfile).first();
 
             if (nic == null) {
                 throw new CloudRuntimeException("Failed to allocate nic for vm " + vm + " in network " + network);
@@ -4355,6 +4357,16 @@ public class NetworkOrchestrator extends ManagerBase implements NetworkOrchestra
         }
 
         return nic;
+    }
+
+    private boolean getNicProfileDefaultNic(NicProfile nicProfile) {
+        if (nicProfile != null) {
+            s_logger.debug(String.format("Using requested nic profile isDefaultNic value [%s].", nicProfile.isDefaultNic()));
+            return nicProfile.isDefaultNic();
+        }
+
+        s_logger.debug("Using isDefaultNic default value [false] as requested nic profile is null.");
+        return false;
     }
 
     @Override
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -1425,6 +1425,8 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
             }
         }
 
+        setNicAsDefaultIfNeeded(vmInstance, profile);
+
         NicProfile guestNic = null;
         boolean cleanUp = true;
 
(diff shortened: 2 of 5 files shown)


kiranchavala:
@blueorangutan package

weizhouapache:
@hsato03 
if you want this to be in 4.18.1.0, can you rebase with 4.18 ?

DaanHoogland (commented):
clgtm, but can you extract the new code into separate methods for readability please

hsato03:
> @hsato03 if you want this to be in 4.18.1.0, can you rebase with 4.18 ?

@weizhouapache Yes, done.

DaanHoogland (approved):
clgtm

DaanHoogland:
@blueorangutan package

weizhouapache:
@hsato03 
code lgtm

can you describe how to reproduce the issue the PR aim to fix ?

DaanHoogland:
@blueorangutan test rocky8 vmware-70u3 keepEnv

hsato03:
> @hsato03 code lgtm
> 
> can you describe how to reproduce the issue the PR aim to fix ?

@weizhouapache If an error occurs with the network adaptor while importing a VM, it's necessary to remove the NIC via VCenter and create a new one in ACS. You can reproduce this by doing the following:
 - Using VMWare, create a VM and unmanage it;
 - On VCenter, go to the VM settings and remove all NICs from it;
 - Import the VM without NICs, create a new NIC for it and try to set the NIC as default.

DaanHoogland:
@hsato03 , I am not allowed to import without NIC: 
image
But I cannot select a network either as there is none to select.

I am trying your scenario from https://github.com/apache/cloudstack/pull/7859#issuecomment-1679387117

hsato03:
> @hsato03 , I am not allowed to import without NIC. But I cannot select a network either as there is none to select.
> 
> I am trying your scenario from [#7859 (comment)](https://github.com/apache/cloudstack/pull/7859#issuecomment-1679387117)

@DaanHoogland 
Looks like you still have a NIC on your VM in VMware; this way, trying to import it without selecting the NIC makes ACS throw the exception.

The problem this PR intends to solve is that, when we have a VM without NIC and we add a NIC to it, we cannot set it as the default NIC. A way to reproduce this situation is importing a VM that does not have a NIC in VMware. 
To reproduce the situation I did the following:
1. created a VM in ACS;
2. unmanaged the VM from ACS;
3. via vCenter, I removed the VM's NIC. This is what the VM looks like in vCenter:
image

Then, I imported the VM to ACS. This is the import form when a VM does not have NIC on vCenter (the `Network adapter` field is not presented):

image

With that, the VM is successfully imported to ACS, however, without NIC:

image

Without this patch, we can attach a NIC to the VM, however, it would not be the default NIC and we would not be able to set it as the default NIC; ACS returns the message `Failed to find a nic profile for the existing default network. This is bad and probably means some sort of configuration corruption`. With the patch, the first NIC of the VM is set as the default, and we can manage it afterwards.

@DaanHoogland, could you try following these steps?

---

Despite this PR solves the situation where we cannot set the default NIC of the VM (given the situation I mentioned), should ACS allow importing VMs without NICs, as ACS does not allow creating VMs without NIC neither remove the default NIC?. @DaanHoogland, what do you think? I can open the discussion in an issue or in the mailing list if valid.

weizhouapache (approved):
code lgtm

hsato03:
Hello, @DaanHoogland. Could you reproduce the tests? Let me know your opinion about the discussion I proposed.

DaanHoogland:
> Did you manage to reproduce the situation?

I have not. I must admit I let other issues take precedence, however. 

> Regarding importing the VM without NICs, do you think it is valid to open a discussion or we should block it?

I think your question warrants a [DISCUSS] thread on dev@ . I am not against allowing it, but there may be some technical, procedural or plain historical objections. In that light it is always good to hide the behaviour of allowing VMs without NIC behind a global setting. A discuss thread should also resolve that question.

DaanHoogland:
@hsato03 I have seen no significant discussions on your thread. Let's proceed. I would say a good implementation displays a warning asking to stop or create a default NIC. What do you think?

hsato03:
@DaanHoogland Good idea. I have added a warning message.

DaanHoogland:
@blueorangutan package

shwstppr:
@blueorangutan package

shwstppr (approved):
Code LGTM

DaanHoogland:
@blueorangutan test alma8 vmware-70u3

PR #7868, opened 2023-08-14
Removed state for removed accounts

hsato03:
### Description

When deleting an account, its state is maintained as `Enabled` even if it is removed.

Accordingly the `Removed` state for accounts was created, which is always set when an account is deleted.

### Types of changes

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

I created and deleted an account. Then, I created a SELECT SQL query for the `account` table and the account state was `Removed`.

### DIFF
--- a/api/src/main/java/com/cloud/user/Account.java
+++ b/api/src/main/java/com/cloud/user/Account.java
@@ -30,7 +30,7 @@ public interface Account extends ControlledEntity, InternalIdentity, Identity {
      *  Account states.
      * */
     enum State {
-        DISABLED, ENABLED, LOCKED;
+        DISABLED, ENABLED, LOCKED, REMOVED;
 
         /**
          * The toString method was overridden to maintain consistency in the DB, as the GenericDaoBase uses toString in the enum value to make the sql statements
--- a/engine/schema/src/main/resources/META-INF/db/schema-41810to41900.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-41810to41900.sql
@@ -180,3 +180,6 @@ CREATE TABLE `cloud`.`vm_scheduled_job` (
 -- Add support for different cluster types for kubernetes
 ALTER TABLE `cloud`.`kubernetes_cluster` ADD COLUMN `cluster_type` varchar(64) DEFAULT 'CloudManaged' COMMENT 'type of cluster';
 ALTER TABLE `cloud`.`kubernetes_cluster` MODIFY COLUMN `kubernetes_version_id` bigint unsigned NULL COMMENT 'the ID of the Kubernetes version of this Kubernetes cluster';
+
+-- Set removed state for all removed accounts
+UPDATE `cloud`.`account` SET state='removed' WHERE `removed` IS NOT NULL;
--- a/server/src/main/java/com/cloud/user/AccountManagerImpl.java
+++ b/server/src/main/java/com/cloud/user/AccountManagerImpl.java
@@ -814,6 +814,9 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
             return false;
         }
 
+        account.setState(State.REMOVED);
+        _accountDao.update(accountId, account);
+
         if (s_logger.isDebugEnabled()) {
             s_logger.debug("Removed account " + accountId);
         }


DaanHoogland (approved):
Good change! one remark, other entities have a state DESTROYED. maybe it would make sense to be in line with that?

weizhouapache:
@blueorangutan package

DaanHoogland:
@hsato03 do you want this in 4.18.1?

hsato03:
> 



> Good change! one remark, other entities have a state DESTROYED. maybe it would make sense to be in line with that?

@DaanHoogland Yes, that's a good idea. But some entities have the `Inactive` removed state like `domain` and `vm_template`.

soreana (approved):

weizhouapache:
> Good change! one remark, other entities have a state DESTROYED. maybe it would make sense to be in line with that?

@DaanHoogland 
to be honest, I think `removed` makes more sense.

DaanHoogland:
@weizhouapache REMOVED does make sense but having different terms with the same meaning is something we should prevent, and adding a new one should replace all occurences of the old ones. Not going to block this for that reason, but it will be confusing for new operators and developers alike.

worth a discuss thread?

cc @soreana @hsato03

weizhouapache:
@hsato03 
since this is not a bug, but an improvement. let's move to next major release 4.19.0.0
cc @DaanHoogland

DaanHoogland:
I think this can be included in 4.18.1 as well, but do not mind either way

api/src/main/java/com/cloud/user/Account.java
```diff
@@ -30,7 +30,7 @@ public interface Account extends ControlledEntity, InternalIdentity, Identity {
      *  Account states.
      * */
     enum State {
-        DISABLED, ENABLED, LOCKED;
+        DISABLED, ENABLED, LOCKED, DESTROYED;
```
rohityadavcloud, line 33: Can we call it REMOVED ?

rohityadavcloud (approved):
LGTM - I'm not sure if there could be only possible side-effects; or it should also emit some kind of event for usage-server?

weizhouapache:
> LGTM - I'm not sure if there could be only possible side-effects; or it should also emit some kind of event for usage-server?

it is one of my concerns as well.
let's consider it for 4.19.0.0

soreana:
@blueorangutan package

hsato03:
I agree with @rohityadavcloud and @weizhouapache, REMOVED sounds better for accounts. @DaanHoogland Although your idea of ​​standardizing the state of entities is good, there are already other entities that use the REMOVED state.

> LGTM - I'm not sure if there could be only possible side-effects; or it should also emit some kind of event for usage-server?

@rohityadavcloud ACS already emits an event when deleting an account.

soreana:
@blueorangutan package

soreana:
@DaanHoogland @rohityadavcloud @weizhouapache I did some research to understand the definitions of `removed` and `destroyed` in CloudStack terminology.

Once an object is Destroyed in CloudStack, it remains visible to the root admin, who has the exclusive ability to either restore or remove it. Even when an object is destroyed, it can still be restored. But once an object is removed, it cannot be restored and will be lost permanently.

As of now, it is not possible to restore the account that was removed. Therefore, using the term `Remove` state would be more appropriate than `Destroyed`.

DaanHoogland:
have you tested @soreana ?

hsato03:
Hi @soreana,

Any updates on this one?

soreana:
Hi @DaanHoogland @hsato03,

I've tested the upgrade from 4.18.1 to 4.19 the process went smooth and the account have removed state now.

Before upgrade:
```
MariaDB [cloud]> select account_name,state,removed from account ;
+--------------------------+---------+---------------------+
| account_name             | state   | removed             |
+--------------------------+---------+---------------------+
| system                   | enabled | NULL                |
| admin                    | enabled | NULL                |
| baremetal-system-account | enabled | NULL                |
| Removed-01               | enabled | 2023-09-28 08:41:35 |
| Removed-02               | enabled | 2023-09-28 08:45:18 |
| Removed-03               | enabled | 2023-09-28 08:46:16 |
+--------------------------+---------+---------------------+
6 rows in set (0.001 sec)

MariaDB [cloud]>
```

After upgrade:
```
MariaDB [cloud]> select account_name,state,removed from account ;
+--------------------------+---------+---------------------+
| account_name             | state   | removed             |
+--------------------------+---------+---------------------+
| system                   | enabled | NULL                |
| admin                    | enabled | NULL                |
| baremetal-system-account | enabled | NULL                |
| Removed-01               | removed | 2023-09-28 08:41:35 |
| Removed-02               | removed | 2023-09-28 08:45:18 |
| Removed-03               | removed | 2023-09-28 08:46:16 |
+--------------------------+---------+---------------------+
6 rows in set (0.001 sec)

MariaDB [cloud]>
```

hsato03:
@soreana Thank's for testing.

PR #7877, opened 2023-08-17
server: try with all supported hypervisors (ordered, no duplications) when deploy a virtual router

### LINKED ISSUE #7433: Virtual Router fails to deploy with multi-hypervisor zone
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
VR, Deployment Planner
~~~

##### CLOUDSTACK VERSION

~~~
4.18, All before 4.18
~~~

##### CONFIGURATION

Any zone with more than one cluster type: XenServer, KVM, VMware
##### STEPS TO REPRODUCE

~~~
1. Create a zone with two clusters of different hypervisor types XenServer/VMware ESXi/KVM
2. Disable One Cluster example: VMware
3. Deploy a virtual machine and create a new guest network, now both VM and VR are created on the XenServer/KVM cluster
4. Disable the XenServer/KVM cluster from Step 3
5. Retart the guest network with clean-up, the new virtual router fails to get created.
(This issue appears randomly(based on the logic mentioned below), retry a couple of times if this is not observed. ) 
~~~

##### EXPECTED RESULTS

~~~
The virtual router should get created
~~~

##### ACTUAL RESULTS

~~~
The virtual router fails to get created. 
~~~

The same could affect SystemVMs as well, the issue is appearing due to the current deployment logic where the new virtual router hypervisor type is decided even before checking if there is any available/suitable host of the same type in the zone and this won't be retried. 

Example from logs: 

 Allocating the VR with id=6 in datacenter Zone .......with the hypervisor type KVM
Deploy avoids pods: [], clusters: [2, 3], hosts: [] , <---- Disabled clusters KVM & VMware
Cluster: 1 has HyperVisorType that does not match the VM, skipping this cluster <---- XenServer Cluser
No clusters found after removing disabled clusters and clusters in avoid list, returning.
state transited from [Starting] to [Stopped] with event [OperationFailed]

weizhouapache:
### Description

This PR fixes #7433 

It was a bit strange that cloudstack only tries to deploy VRs with at most two hypervisors. If there are more than 2 hypervisors (e.g. kvm/vmware/xen) on the environment, the VR deployment might fail.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

- create an env with 3 hosts (1 kvm,1 vmware,1 xen)
- create a vm on isolated network on each hypervisor (3 networks and 3 vms in total)
- restart networks with cleanup, all work
- disable 2 of 3 hypervisors. for example, kvm is enabled, vmware/xen are disabled.
- restart networks with cleanup, all work
- disable kvm/xen host, and enable vmware
- restart networks with cleanup, all work
- disable kvm/vmware host, and enable xen
- restart networks with cleanup, all work

### DIFF
--- a/server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java
+++ b/server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java
@@ -36,6 +36,7 @@ import org.apache.cloudstack.framework.config.ConfigKey;
 import org.apache.cloudstack.framework.config.dao.ConfigurationDao;
 import org.apache.cloudstack.network.router.deployment.RouterDeploymentDefinition;
 import org.apache.cloudstack.utils.CloudStackVersion;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.log4j.Logger;
 
 import com.cloud.agent.AgentManager;
@@ -43,6 +44,7 @@ import com.cloud.agent.api.Answer;
 import com.cloud.agent.api.to.NicTO;
 import com.cloud.agent.manager.Commands;
 import com.cloud.alert.AlertManager;
+import com.cloud.capacity.CapacityManager;
 import com.cloud.configuration.Config;
 import com.cloud.dc.ClusterVO;
 import com.cloud.dc.DataCenter;
@@ -168,6 +170,8 @@ public class NetworkHelperImpl implements NetworkHelper {
     RouterHealthCheckResultDao _routerHealthCheckResultDao;
     @Inject
     Ipv6Service ipv6Service;
+    @Inject
+    CapacityManager capacityMgr;
 
     protected final Map<HypervisorType, ConfigKey<String>> hypervisorsMap = new HashMap<>();
 
@@ -507,6 +511,8 @@ public class NetworkHelperImpl implements NetworkHelper {
         for (final Iterator<HypervisorType> iter = hypervisors.iterator(); iter.hasNext();) {
             final HypervisorType hType = iter.next();
             try {
+                checkIfZoneHasCapacity(routerDeploymentDefinition.getDest().getDataCenter(), hType, routerOffering);
+
                 final long id = _routerDao.getNextInSequence(Long.class, "id");
                 if (s_logger.isDebugEnabled()) {
                     s_logger.debug(String.format("Allocating the VR with id=%s in datacenter %s with the hypervisor type %s", id, routerDeploymentDefinition.getDest()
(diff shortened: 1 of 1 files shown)


weizhouapache:
@blueorangutan test

server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java
```diff
@@ -618,7 +616,7 @@ protected List<HypervisorType> getHypervisors(final RouterDeploymentDefinition r
             throw new InsufficientServerCapacityException("Unable to create virtual router, there are no clusters in the zone." + getNoHypervisorsErrMsgDetails(),
                     DataCenter.class, dest.getDataCenter().getId());
         }
-        return hypervisors;
+        return new ArrayList(new LinkedHashSet<>(hypervisors));
```
DaanHoogland, line 619: it is strange that `hypervisors` is not a `Set` in the first place, but this will work.

DaanHoogland (approved):
so we only try once but for all possible hypervisors. this makes sense.

server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java
weizhouapache: @DaanHoogland 
right.
I was thinking about it, but if we define it as a Set, we need to convert the following from List to Set
```
            hypervisors = _resourceMgr.getSupportedHypervisorTypes(dest.getDataCenter().getId(), true, routerDeploymentDefinition.getPlan().getPodId());
```

DaanHoogland: yes, I also saw another call being done in the method that might require changes: `filterSupportedHypervisors(final List<HypervisorType> hypervisors)`.

I don't care if we leave that for later.

weizhouapache: > yes, I also saw another call being done in the method that might require changes: `filterSupportedHypervisors(final List<HypervisorType> hypervisors)`.
> 
> I don't care if we leave that for later.

yep, the method did nothing

server/src/main/java/com/cloud/network/router/NetworkHelperImpl.java
```diff
@@ -547,14 +547,12 @@ public DomainRouterVO deployRouter(final RouterDeploymentDefinition routerDeploy
                 reallocateRouterNetworks(routerDeploymentDefinition, router, template, null);
                 router = _routerDao.findById(router.getId());
             } catch (final InsufficientCapacityException ex) {
-                if (allocateRetry < 2 && iter.hasNext()) {
+                if (iter.hasNext()) {
```
rohityadavcloud, line 550: could it cause regression, where earlier for one hypervisor it was trying at least twice?

weizhouapache: @rohityadavcloud 
`allocateRetry < 2` means that it will try with 2 hypervisors, not try twice for each hypervisor.

rohityadavcloud: ah :) my bad - LGTM thnx @weizhouapache

rohityadavcloud (approved):

rohityadavcloud:
Should we merge this @weizhouapache @DaanHoogland or do we need manual QA ?

weizhouapache:
> Should we merge this @weizhouapache @DaanHoogland or do we need manual QA ?

@rohityadavcloud 
would be better to test it before merge. let me create a env with 3 hypervisors and ask you guys to test it.

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test

weizhouapache:
@DaanHoogland 
this is ready for review/testing

what I have tested are
- create an env with 3 hosts (1 kvm,1 vmware,1 xen)
- create a vm on isolated network on each hypervisor (3 networks and 3 vms in total)
- restart networks with cleanup, all work
- disable 2 of 3 hypervisors. for example, kvm is enabled, vmware/xen are disabled.
- restart networks with cleanup, all work
- disable kvm/xen host, and enable vmware
- restart networks with cleanup, all work
- disable kvm/vmware host, and enable xen
- restart networks with cleanup, all work

DaanHoogland:
@weizhouapache,
I tested ok as described, but I also tested maintenance. In the above scenario I put the vmware host to disabled and the xenhost to maintenance. Now the ssvm came up on the kvm host and the cpvm kept retrying. I am not sure if this is the same issue or should be fixed in a new PR though. As said, the described scenario works as it says on the tin.

DaanHoogland:
I verified with a VR as well, and in the end all start on the available host irrespective of hypervisor type. merging

weizhouapache:
> I verified with a VR as well, and in the end all start on the available host irrespective of hypervisor type. merging

thanks @DaanHoogland !

PR #7878, opened 2023-08-17
Publish event for VM.STOP when out of band stop is detected

mlsorensen:
### Description

This PR publishes an action event for VM.STOP when the power state processor detects a VM is gone from hypervisor. Currently only a power state event is published on the message bus. This allows events to be seen and processed when VM is detected to be stopped out of band.

Additionally, it was discovered that the existing missing VM code is triggered when a VM is taking awhile to start. For example if we are waiting on the router VM to come up, the report can possibly see no VM when one is expected. The VM is assigned to the host, but doesn't exist yet, and triggers the missing VM code.  A check was added to ignore the VM if it is still in "Starting" state.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

Screenshot 2023-08-17 at 10 37 38 AM

### How Has This Been Tested?

Tested out of band stop by shutting down guest within the VM, confirmed new event triggered.

Tested startup of VM where power state is processed while waiting on router to come up, confirmed events no longer triggered detecting a "missing VM" when VM is in starting state.

Tested live migration, confirmed we processed power state reports for both source and destination hypervisor hosts and did not issue the new VM.STOP event.

### DIFF
--- a/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -47,6 +47,7 @@ import javax.inject.Inject;
 import javax.naming.ConfigurationException;
 import javax.persistence.EntityExistsException;
 
+import com.cloud.event.ActionEventUtils;
 import org.apache.cloudstack.affinity.dao.AffinityGroupVMMapDao;
 import org.apache.cloudstack.annotation.AnnotationService;
 import org.apache.cloudstack.annotation.dao.AnnotationDao;
@@ -4804,6 +4805,8 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
                     VM_SYNC_ALERT_SUBJECT, "VM " + vm.getHostName() + "(" + vm.getInstanceName() + ") state is sync-ed (" + vm.getState()
                     + " -> Running) from out-of-context transition. VM network environment may need to be reset");
 
+            ActionEventUtils.onActionEvent(User.UID_SYSTEM, Account.ACCOUNT_ID_SYSTEM, vm.getDomainId(),
+                    EventTypes.EVENT_VM_START, "Out of band VM power on", vm.getId(), ApiCommandResourceType.VirtualMachine.toString());
             s_logger.info("VM " + vm.getInstanceName() + " is sync-ed to at Running state according to power-on report from hypervisor");
             break;
 
@@ -4837,6 +4840,8 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         case Stopping:
         case Running:
         case Stopped:
+            ActionEventUtils.onActionEvent(User.UID_SYSTEM, Account.ACCOUNT_ID_SYSTEM,vm.getDomainId(),
+                    EventTypes.EVENT_VM_STOP, "Out of band VM power off", vm.getId(), ApiCommandResourceType.VirtualMachine.toString());
         case Migrating:
             if (s_logger.isInfoEnabled()) {
                 s_logger.info(


weizhouapache (approved):
code lgtm

this PR has only impact on the events

DaanHoogland:
@mlsorensen have you considered an out of bounds migration scenario? this might give an undesirable effect if the VM was started on another host, would it?.

mlsorensen:
> @mlsorensen have you considered an out of bounds migration scenario? this might give an undesirable effect if the VM was started on another host, would it?.

Good point. Whatever is happening in this code during out of band migration, this doesn't change that. I don't know if anyone is looking at the message bus that is currently already publishing, or if there are existing issues with how it handles the VM state, but maybe publishing to the event system will make it more obvious if there's an existing bug here.

I don't know that anyone using KVM at least is doing out of band migration because the tools aren't great for that, and in many cases simply not possible due to storage and network plugins needing to run and make storage/network accessible on the host. Perhaps useful for someone using VMware, where there is a separate system to manage VMs independently. I don't know that a test via KVM is going to really exercise it in the same way the VMware integration does.

Worst case, we see an event for VMware users indicating the VM was removed from its host out of band, but we aren't making a change to how the VM state is handled and synced up for these kinds of migrations.

With the existing code what I see is during live migration, both source and destination hypervisors send a report that the VM is powered on, and both systems are updating the VM power state when they send pings. Thus the VM's `power_host` flip flops back and forth, but the `host_id` is already set to the new host. When the VM finally migrates, it doesn't complain that the VM is powered off because the host complaining the VM is gone is not matching the `host_id`. With out of band, we need to make sure we understand how the `host_id` will get updated and that this is compatible.

mlsorensen:
Ok, it took some manual intervention to test with KVM (copying config drive iso, editing XML) but here is what I'm seeing:

1) With OOB live migration, instead of just the `power_host` flip-flopping, the `host_id` itself flip flops. This could be an issue right now as looking at the VM during this time just shows the VM "Running" but we don't know which host it's supposed to land on. Something like a volume resize or snapshot issued during this time would probably fail. The fix to this is probably to properly detect the migrating scenario and change the VM state to `Migrating`, but outside of the scope of this PR. Really it seems out of band moves are not supported (at least on KVM), Cloudstack just does what it can to reconcile.

2) This change doesn't seem to have any effect on OOB handling. What I observe is that given two hosts issuing power state reports: `H1 Report ... t1 ... H2 Report ... t2 ... H1 Report`, and VM OOB migrating from H1 to H2, if the migration completes during t1, then H2 Report claims the VM properly and when the last H1 report comes there is no change.  If the migration completes during t2, H2 was the last one to claim the VM, so again there is no effect.  The only case where there could be an issue is if the H2 agent is offline, or some other situation where we receive two H1 reports in a row before H2 can claim the VM.  In this case it is probably fair to notify that a VM has stopped if someone has out of band moved it to a host CloudStack has no active control over.

DaanHoogland:
> Perhaps useful for someone using VMware, where there is a separate system to manage VMs independently. I don't know that a test via KVM is going to really exercise it in the same way the VMware integration does.
> 
> Worst case, we see an event for VMware users indicating the VM was removed from its host out of band, but we aren't making a change to how the VM state is handled and synced up for these kinds of migrations.

A scenario we have seen (which is purely a billing issue but still) Is that a VM was at a certain moment powered off and hence absent in reports from both hosts, while migration was still going on. This led to billing issues as those are a bit fragile. it relies a bit too much on a proper order of events and matching on/off create/delete etc combinations of events. What I am missing in this PR is the start event (also for OOB).

We will have to test this and I think you are right this will probably only be an issue on vmware (not sure about xen). 

@borisstoyanov (@vladimirpetrov) I think we need to give an extra eye to testing this.

> 1. The fix to this is probably to properly detect the migrating scenario and change the VM state to `Migrating`, but outside of the scope of this PR.

agreed

weizhouapache:
smoke tests passed with message
```
22:26:44 msg: {"body": "<b>[SF] Trillian test result (tid-7476)</b>\nEnvironment: kvm-centos7 (x2), Advanced Networking with Mgmt server 7\nTotal time taken: 41024 seconds\nMarvin logs: [https://github.com/blueorangutan/acs-prs/releases/download/trillian/pr7878-t7476-kvm-centos7.zip\nSmoke](https://github.com/blueorangutan/acs-prs/releases/download/trillian/pr7878-t7476-kvm-centos7.zip/nSmoke) tests completed. 108 look OK, 0 have errors, 0 did not run\nOnly failed and skipped tests results shown below:\n\n\nTest | Result | Time (s) | Test File\n--- | --- | --- | ---\n"}
```

DaanHoogland:
> smoke tests passed with message
> 
> ```
> 22:26:44 msg: {"body": "<b>[SF] Trillian test result (tid-7476)</b>\nEnvironment: kvm-centos7 (x2), Advanced Networking with Mgmt server 7\nTotal time taken: 41024 seconds\nMarvin logs: [https://github.com/blueorangutan/acs-prs/releases/download/trillian/pr7878-t7476-kvm-centos7.zip\nSmoke](https://github.com/blueorangutan/acs-prs/releases/download/trillian/pr7878-t7476-kvm-centos7.zip/nSmoke) tests completed. 108 look OK, 0 have errors, 0 did not run\nOnly failed and skipped tests results shown below:\n\n\nTest | Result | Time (s) | Test File\n--- | --- | --- | ---\n"}
> ```

api rate limit again?

weizhouapache:
> api rate limit again?
yes @DaanHoogland

rohityadavcloud:
@blueorangutan package

rohityadavcloud (approved):
LGTM, haven't tested it, also don't know if this can cause side-effects for all hypervisors and cases, as we're skipping

DaanHoogland:
@blueorangutan test matrix

weizhouapache:
@mlsorensen @DaanHoogland 
I had a discussion with Daan, it looks the whole process can be improved in `handlePowerOnReportWithNoPendingJobsOnVM` and `handlePowerOffReportWithNoPendingJobsOnVM` in `VirtualMachineManagerImpl.java`
we could publish not only the STOP event, but also START/MIGRATE events, so that users have a better information of out-of-band actions.

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachinePowerStateSyncImpl.java
```diff
@@ -162,6 +177,8 @@ private void processReport(long hostId, Map<Long, VirtualMachine.PowerState> tra
                         }
 
                         _messageBus.publish(null, VirtualMachineManager.Topics.VM_POWER_STATE, PublishScope.GLOBAL, instance.getId());
+                        ActionEventUtils.onActionEvent(User.UID_SYSTEM, Account.ACCOUNT_ID_SYSTEM,instance.getDomainId(),
+                                EventTypes.EVENT_VM_STOP, "Out of band VM power off", instance.getId(), ApiCommandResourceType.VirtualMachine.toString());
```
DaanHoogland, line 181: I think this event should be spewed at `handlePowerOffReportWithNoPendingJobsOnVM` and countered by a `EVENT_VM_START` in `handlePowerOnReportWithNoPendingJobsOnVM` both in `VirtualMachineManagerImpl`. Or alternatively a proper place to handle the OOB start event in this class should be found. I think this is not a big deal for you @mlsorensen , but for people using event/usgae server based billing it would be. If an OOB migration happens their billing for this VM would/could stop or continue based on the order of events.

DaanHoogland (changes requested):

weizhouapache:
moved to 4.18.2.0

engine/orchestration/src/main/java/com/cloud/vm/VirtualMachinePowerStateSyncImpl.java
mlsorensen: Understood, I'll look into that.

mlsorensen: This may get is what you're looking for @DaanHoogland - in testing this didn't trigger any event at all during out of band migration on KVM. However it still responded to OOB stop, and even OOB start, publishing an event in both situations.

Screenshot 2023-08-24 at 10 32 46 AM

DaanHoogland: @mlsorensen , this is very hard to reproduce as it depends on the order of power sync reports coming in from hosts. I'll give this a swing after 4.18.1, and think about simulating race conditions. code looks good.

DaanHoogland (approved):
clgtm

rohityadavcloud:
LGTM
@blueorangutan package

rohityadavcloud:
@blueorangutan test

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@weizhouapache @mlsorensen @rohityadavcloud Do we merge or do we need more testing on this?

mlsorensen:
@DaanHoogland from my end it's good to go.

PR #7911, opened 2023-08-25
ui: speed up compute instance listing

### LINKED ISSUE #7910: GUI -  search instance always give the first VM name and GUI is very slow
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

Categorize the issue - UI

##### CLOUDSTACK VERSION
CloudStack 4.18.0.0

##### OS / ENVIRONMENT
KVM with OEL9

##### SUMMARY
in GUI - when  search instance with particular instance name always give the first VM name in the dashboard 
 GUI is very slow

##### STEPS TO REPRODUCE
GUI- Compute-Instances - search tab - give instance name 

##### EXPECTED RESULTS
search should give the exact VM(instance) name 

##### ACTUAL RESULTS
searching one VM gives the  other VM name ( the first VM in list ) 
~~~

rohityadavcloud:
The default listVirtualMachinesMetrics APIs assumings details=all which isn't really used in the list view. By changing this to a subset of details, we can see gains upto 10x in listing speed in the UI. When moving to the resource/detail view of the UI, don't pass state or details so `all` the details of the instance are returned.

Possibly fixes #7910

In the screenshot below, we can see passing details=min, and other combinations show significant reduction in network time (from 300-400ms to 20-40ms)

Screenshot 2023-08-25 at 12 25 22 PM
Screenshot 2023-08-25 at 12 28 25 PM

### Types of changes

### DIFF
--- a/ui/src/config/section/compute.js
+++ b/ui/src/config/section/compute.js
@@ -32,9 +32,9 @@ export default {
       permission: ['listVirtualMachinesMetrics'],
       resourceType: 'UserVm',
       params: () => {
-        var params = {}
+        var params = { details: 'servoff,tmpl,nics' }
         if (store.getters.metrics) {
-          params = { state: 'running' }
+          params = { details: 'servoff,tmpl,nics,stats' }
         }
         return params
       },
--- a/ui/src/store/mutation-types.js
+++ b/ui/src/store/mutation-types.js
@@ -37,6 +37,7 @@ export const DOMAIN_STORE = 'DOMAIN_STORE'
 export const DARK_MODE = 'DARK_MODE'
 export const VUE_VERSION = 'VUE_VERSION'
 export const CUSTOM_COLUMNS = 'CUSTOM_COLUMNS'
+export const RELOAD_ALL_PROJECTS = 'RELOAD_ALL_PROJECTS'
 
 export const CONTENT_WIDTH_TYPE = {
   Fluid: 'Fluid',
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -858,6 +858,11 @@ export default {
         delete params.showunique
       }
 
+      if (['listVirtualMachinesMetrics'].includes(this.apiName) && this.dataView) {
+        delete params.state
+        delete params.details
+      }
+
       this.loading = true
       if (this.$route.params && this.$route.params.id) {
         params.id = this.$route.params.id
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -859,7 +859,6 @@ export default {
       }
 
       if (['listVirtualMachinesMetrics'].includes(this.apiName) && this.dataView) {
-        delete params.state
         delete params.details
       }
 


rohityadavcloud:
@blueorangutan ui

ui/src/config/section/compute.js
```diff
@@ -32,9 +32,9 @@ export default {
       permission: ['listVirtualMachinesMetrics'],
       resourceType: 'UserVm',
       params: () => {
-        var params = {}
+        var params = { details: 'servoff,tmpl,nics' }
         if (store.getters.metrics) {
-          params = { state: 'running' }
```
rohityadavcloud, line 37: I've removed running as otherwise metrics doesn't return all VMs in the list. One can select VM `state` explicitly in the compute/instance page if they want.

ui/src/store/mutation-types.js
```diff
@@ -37,6 +37,7 @@ export const DOMAIN_STORE = 'DOMAIN_STORE'
 export const DARK_MODE = 'DARK_MODE'
 export const VUE_VERSION = 'VUE_VERSION'
 export const CUSTOM_COLUMNS = 'CUSTOM_COLUMNS'
+export const RELOAD_ALL_PROJECTS = 'RELOAD_ALL_PROJECTS'
```
rohityadavcloud, line 40: This fixes a warning

ui/src/config/section/compute.js
```diff
@@ -32,9 +32,9 @@ export default {
...
+          params = { details: 'servoff,tmpl,nics,stats' }
```
rohityadavcloud, line 37: VM stats in the API response can cause 5-10x delay compared to normal list API response

ui/src/views/AutogenView.vue
```diff
@@ -858,6 +858,11 @@ export default {
         delete params.showunique
       }
 
+      if (['listVirtualMachinesMetrics'].includes(this.apiName) && this.dataView) {
+        delete params.state
+        delete params.details
+      }
```
weizhouapache, line 864: @rohityadavcloud 
can you explain why state and details are removed ?

all others look good

ui/src/config/section/compute.js
DaanHoogland: is `tmpl` needed in the details? or `stats`?

DaanHoogland (approved):
code looks good. I can't judge the performance gain in qa, but am relying on the author in this case.

Pearl1594 (approved):

ui/src/config/section/compute.js
weizhouapache: my concern is opposite to @DaanHoogland  :-D
is `'servoff,tmpl,nics`' or `servoff,tmpl,nics,stats` enough for UI ?

ui/src/views/AutogenView.vue
rohityadavcloud: These are added by the router config, I'll ping you on that code; so we must remove if the keys exists for dataView (or resource view, when only one item is being listed such as the VM)

ui/src/config/section/compute.js
```diff
@@ -32,9 +32,9 @@ export default {
       permission: ['listVirtualMachinesMetrics'],
       resourceType: 'UserVm',
       params: () => {
-        var params = {}
+        var params = { details: 'servoff,tmpl,nics' }
```
rohityadavcloud, line 35: @weizhouapache because of these params, details and state...

ui/src/config/section/compute.js
rohityadavcloud: just realised I can remove state from the params

ui/src/config/section/compute.js
rohityadavcloud: Yes, these are enough just for list view (table view); the stats are only needed when the metrics button is clicked. For the resource view, we delete the params.details, so by default details=all is used when listing a single VM.

ui/src/views/AutogenView.vue
```diff
@@ -858,6 +858,10 @@ export default {
         delete params.showunique
       }
 
+      if (['listVirtualMachinesMetrics'].includes(this.apiName) && this.dataView) {
+        delete params.details
```
rohityadavcloud, line 862: we remove the details are it may not be `all` as set for the list view (tabular view)

ui/src/config/section/compute.js
rohityadavcloud: @weizhouapache @DaanHoogland 
I tested this again both as root admin and normal user account; for VMs and templates with and without resource icons; these are enough details to render the list view (table); when the user clicks any VM/instance, then we call the API remove the details and all details are listed/used in the resource view. While not ideal, this speeds up the listing.

rohityadavcloud:
@blueorangutan ui

ui/src/config/section/compute.js
weizhouapache: great, thanks @rohityadavcloud 
I will test it

weizhouapache (approved):
code lgtm

the display is almost identical to the original page

without this PR
image


with this PR
image

dseralathan:
the VM search give exact VM and is not changing to the first VM in list after we exclude state,IP of VM at filter in field

PR #7924, opened 2023-08-30
PowerFlex: Handle missing volumes gracefully during delete volume

mlsorensen:
### Description

This PR allows the PowerFlex plugin to gracefully handle the case where volume has already been removed from the backend PowerFlex storage and a user attempts to expunge volume.  

Without this, attempting to also clean up in cloudstack results in volumes that are stuck in `Destroy` state, unless an admin manually edits the database.  If the volume is missing on the back end and admins want to recover, they can still attempt to fix this in some way and the record would be preserved in CloudStack as long as a volume expunge is not attempted.

```
Failed to expunge the volume Vol[37950|name=testvol|vm=null|DATADISK] in Primary data store : Unable to delete PowerFlex volume: xxxxx:xxxxx due to API failed due to: {"message":"Could not find the volume","httpStatusCode":500,"errorCode":79}
```

We catch this error case and match the error message to determine if the volume is already gone. There may be some document for handling the error codes more directly, but currently the PowerFlex client doesn't handle errors in this way.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/client/ScaleIOGatewayClientImpl.java
+++ b/plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/client/ScaleIOGatewayClientImpl.java
@@ -738,11 +738,20 @@ public class ScaleIOGatewayClientImpl implements ScaleIOGatewayClient {
         try {
             unmapVolumeFromAllSdcs(volumeId);
         } catch (Exception ignored) {}
-        Boolean removeVolumeStatus = post(
-                "/instances/Volume::" + volumeId + "/action/removeVolume",
-                "{\"removeMode\":\"ONLY_ME\"}", Boolean.class);
-        if (removeVolumeStatus != null) {
-            return removeVolumeStatus;
+
+        try {
+            Boolean removeVolumeStatus = post(
+                    "/instances/Volume::" + volumeId + "/action/removeVolume",
+                    "{\"removeMode\":\"ONLY_ME\"}", Boolean.class);
+            if (removeVolumeStatus != null) {
+                return removeVolumeStatus;
+            }
+        } catch (Exception ex) {
+            if (ex instanceof ServerApiException && ex.getMessage().contains("Could not find the volume")) {
+                LOG.warn(String.format("API says deleting volume %s does not exist, handling gracefully", volumeId));
+                return true;
+            }
+            throw ex;
         }
         return false;
     }


rohityadavcloud (approved):
Code LGTM didn't test it though

rohityadavcloud:
@blueorangutan package

harikrishna-patnala (approved):
LGTM

rohityadavcloud:
@blueorangutan test

harikrishna-patnala (approved):
Quickly verified the issue and the fix. Volume delete with expunge completed gracefully.

2023-08-31 06:19:26,717 WARN  [o.a.c.s.d.c.ScaleIOGatewayClientImpl] (API-Job-Executor-1:ctx-c77978d8 job-154 ctx-2c92ed87) (logid:2628581a) API says deleting volume 4ace104b00000005 does not exist, handling gracefully

plugins/storage/volume/scaleio/src/main/java/org/apache/cloudstack/storage/datastore/client/ScaleIOGatewayClientImpl.java
```diff
@@ -738,11 +738,20 @@ public boolean deleteVolume(final String volumeId) {
...
+
+        try {
+            Boolean removeVolumeStatus = post(
+                    "/instances/Volume::" + volumeId + "/action/removeVolume",
+                    "{\"removeMode\":\"ONLY_ME\"}", Boolean.class);
+            if (removeVolumeStatus != null) {
+                return removeVolumeStatus;
+            }
+        } catch (Exception ex) {
+            if (ex instanceof ServerApiException && ex.getMessage().contains("Could not find the volume")) {
```
DaanHoogland, line 750: depending on the textual contents of the exception message seems a bit unrelyable and maintenance sensitive, is there no return code that can be queried? (excuse my lack of scaleIO/PowerFlex knowledge) (and clgtm otherwise)

harikrishna-patnala: @DaanHoogland Marcus mentioned this in the PR description
"We catch this error case and match the error message to determine if the volume is already gone. There may be some document for handling the error codes more directly, but currently the PowerFlex client doesn't handle errors in this way."

DaanHoogland: sorry, only reviewed the code :(

weizhouapache (approved):
code lgtm

This will be merged when trillian tests finish.

PR #7692, opened 2023-06-26
Add Custom hypervisor minimal changes

nvazquez:
### Description

This PR introduces the minimal changes to add a new hypervisor type (internally named Custom in the codebase, and configurable display name), allowing to write an external hypervisor plugin as a Custom Hypervisor to CloudStack

The custom hypervisor name is set by the setting: 'hypervisor.custom.display.name'

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

PR #7939, opened 2023-09-04
UI: Fix user role login due to missing API access on custom hypervisor name

nvazquez:
### Description

This PR fixes the user role login in the UI on the main branch, bug introduced by the PR: #7692 

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
- As administrator, add 'Custom' to the 'hypervisor.list'
- Restart management server
- Create a Custom hypervisor zone
- Attempt login as a user -> verify it works
- As a user check the custom hypervisor is listed as a hypervisor on the new zone when trying to register a template

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/response/CapabilitiesResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/CapabilitiesResponse.java
@@ -101,7 +101,7 @@ public class CapabilitiesResponse extends BaseResponse {
     private boolean kubernetesClusterExperimentalFeaturesEnabled;
 
     @SerializedName("customhypervisordisplayname")
-    @Param(description = "Display name for custom hypervisor")
+    @Param(description = "Display name for custom hypervisor", since = "4.19.0")
     private String customHypervisorDisplayName;
 
     @SerializedName("defaultuipagesize")
--- a/api/src/main/java/org/apache/cloudstack/api/response/CapabilitiesResponse.java
+++ b/api/src/main/java/org/apache/cloudstack/api/response/CapabilitiesResponse.java
@@ -100,6 +100,10 @@ public class CapabilitiesResponse extends BaseResponse {
     @Param(description = "true if experimental features for Kubernetes cluster such as Docker private registry are enabled, false otherwise")
     private boolean kubernetesClusterExperimentalFeaturesEnabled;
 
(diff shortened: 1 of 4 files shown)


shwstppr (commented):
@nvazquez normal user may not have `listConfigurations` API access. Is this to bypass that?

nvazquez:
@shwstppr yes, in case they don't have access then we set the default name for the Custom hypervisor

shwstppr:
> @shwstppr yes, in case they don't have access then we set the default name for the Custom hypervisor

@nvazquez in that case a normal user will always see custom hypervisor as `Custom` even though the admin has set some display name. Not sure if the value is used in UI/API. Would it be better to return the config value as a response param of listCapabilities API?

nvazquez:
@shwstppr good idea, let me fix that

nvazquez:
@blueorangutan package

nvazquez:
@blueorangutan test

api/src/main/java/org/apache/cloudstack/api/response/CapabilitiesResponse.java
```diff
@@ -100,6 +100,10 @@ public class CapabilitiesResponse extends BaseResponse {
     @Param(description = "true if experimental features for Kubernetes cluster such as Docker private registry are enabled, false otherwise")
     private boolean kubernetesClusterExperimentalFeaturesEnabled;
 
+    @SerializedName("customhypervisordisplayname")
+    @Param(description = "Display name for custom hypervisor")
```
harikrishna-patnala, line 104: do we have the tradition of adding since parameter in response ? If yes, can you please add that. Remaining code LGTM.

harikrishna-patnala (approved):

shwstppr (approved):
Code LGTM

rohityadavcloud (approved):
LGTM - didn't test it

api/src/main/java/org/apache/cloudstack/api/response/CapabilitiesResponse.java
nvazquez: Done, thanks

nvazquez:
@blueorangutan package

PR #7947, opened 2023-09-06
ui: don't skip if dataview has multiple items in response

rohityadavcloud:
This would fix the case of multiple items return in API response for a resource such as a template or ISO in case of multi-zone env.

### Types of changes

### DIFF
--- a/ui/src/views/AutogenView.vue
+++ b/ui/src/views/AutogenView.vue
@@ -942,10 +942,6 @@ export default {
           console.log('DEBUG - Discarding API response as its `id` does not match the uuid on the browser path')
           return
         }
-        if (this.dataView && apiItemCount > 1) {
-          console.log('DEBUG - Discarding API response as got more than one item in data view', this.$route.params, this.items)
-          return
-        }
 
         this.items = json[responseName][objectName]
         if (!this.items || this.items.length === 0) {


rohityadavcloud:
@blueorangutan ui

weizhouapache (approved):
Verified OK on an environment with 2 zones

nvazquez:
@rohityadavcloud could this fix cause regressions on PR https://github.com/apache/cloudstack/pull/7846 which introduced the check?

rohityadavcloud:
@nvazquez yes we need to regression test, but I think this code wasn't that useful in the first place; was put in as an additional check. We should check for #7846

shwstppr (approved):
code lgtm

rohityadavcloud:
I've tested for https://github.com/apache/cloudstack/pull/7846 and can't reproduce the issue with this fix, I think this PR should address @rajujith 's issue and not cause further regression. Thank @rajujith @weizhouapache for reporting.
/cc @weizhouapache @nvazquez

rohityadavcloud:
@blueorangutan package

nvazquez (approved):
Thanks @rohityadavcloud - LGTM

weizhouapache:
Reproduced the issue and verified it is fixed by this PR.

Without this PR
image


With this PR
image

rajujith (commented):
The issue is no longer observed in environments with 2 zones.

PR #8033, opened 2023-10-04
Add plugin dir to developer jetty:run classpath for plugin dev

mlsorensen:
### Description

This PR is a simple change to allow adding plugins to CloudStack when running in developer mode.  Simply drop plugin jar into `client/target/plugins` before launching `mvn -pl :cloud-client-ui jetty:run`.

Also bumping the version of jetty-maven-plugin to the same as main jetty server.

### Types of changes

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -154,7 +154,7 @@
         <cs.jaxws.version>2.3.2-1</cs.jaxws.version>
         <cs.jersey-client.version>2.26</cs.jersey-client.version>
         <cs.jetty.version>9.4.51.v20230217</cs.jetty.version>
-        <cs.jetty-maven-plugin.version>9.4.27.v20200227</cs.jetty-maven-plugin.version>
+        <cs.jetty-maven-plugin.version>9.4.51.v20230217</cs.jetty-maven-plugin.version>
         <cs.jna.version>5.5.0</cs.jna.version>
         <cs.joda-time.version>2.12.5</cs.joda-time.version>
         <cs.jpa.version>2.2.1</cs.jpa.version>
--- a/client/pom.xml
+++ b/client/pom.xml
@@ -642,7 +642,7 @@
                     <webAppSourceDirectory>${project.build.directory}/classes/META-INF/webapp/</webAppSourceDirectory>
                     <webApp>
                         <contextPath>/client</contextPath>
-                        <extraClasspath>${project.build.directory}/conf/;${project.build.directory}/common;${project.build.directory}/utilities/scripts/db/;${project.build.directory}/utilities/scripts/db/db/;${project.build.directory}/plugins/*;${project.build.directory}/cloud-client-ui-${project.version}.jar</extraClasspath>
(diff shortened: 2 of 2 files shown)


pom.xml
```diff
@@ -154,7 +154,7 @@
         <cs.jaxws.version>2.3.2-1</cs.jaxws.version>
         <cs.jersey-client.version>2.26</cs.jersey-client.version>
         <cs.jetty.version>9.4.51.v20230217</cs.jetty.version>
-        <cs.jetty-maven-plugin.version>9.4.27.v20200227</cs.jetty-maven-plugin.version>
+        <cs.jetty-maven-plugin.version>9.4.51.v20230217</cs.jetty-maven-plugin.version>
```
rohityadavcloud, line 157: The 9.4.27 was a known/working version, I'm not sure if with more recent working the jetty-maven-plugin works @mlsorensen have you tested it locally (see most Github actions jobs are failing)

mlsorensen: It worked locally, and it's the version of jetty that is bundled with the production cloudstack, but perhaps there is something else that doesn't work?

mlsorensen: Seems like the new versions of jetty may be more aggressive at scanning? I see more duplicate classes found, seems to be searching current directory as well as finding them in `client/target/cloud-client-ui` jar. Will try a few things...

mlsorensen: Ok I think maybe I found the issue @rohityadavcloud. Newer jetty-maven-plugin versions add the pom dependencies to class path automatically, with older jetty we have defined `extraClassPath` and add a built cloud-client-ui.jar location. This results in a lot of duplicate classes discovered.  Also notice that without specifying `-Psimulator` the jetty:run doesn't have simulator available which breaks the simulator CI build.

We can add explicit simulator dependency for the jetty-maven-plugin, or we can add `-Psimulator` to the CI (but I don't know how to do the latter).

mlsorensen: Eh, I'm not confident that I know exactly what's going on here, but I think the same [PR](https://github.com/eclipse/jetty.project/pull/5131) in jetty that I need for this feature is probably the same one that cloudstack's use of jetty:run is incompatible with. Or certainly the very next release after this.

May need to resolve the underlying reason why we can't upgrade to newer jetty-maven-plugin before we consider adding a plugin dir for devs.

rohityadavcloud:
@mlsorensen smoketests (github actions) are still failing

rohityadavcloud:
@mlsorensen I think this is incompatible version and may require some additional changes to make them work. Does it work for you locally?

DaanHoogland:
@mlsorensen @rohityadavcloud , closing this one as it has conflicts and is old. please update and reopen if you think it is still relevant.

PR #8048, opened 2023-10-06
systemvm: remove config in /etc/pam.d/systemd-user to fix user@0.service

weizhouapache:
### Description

the service `user@0.service` fails in system vms and virtual routers

This PR removes a change to fix memory leak of SSH connections in the systemvm templates with old linux kernel.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/systemvm/debian/opt/cloud/bin/setup/init.sh
+++ b/systemvm/debian/opt/cloud/bin/setup/init.sh
@@ -209,8 +209,7 @@ setup_interface_sshd() {
   fi
 
   systemctl restart systemd-journald
-  # Patch known systemd/sshd memory leak - https://github.com/systemd/systemd/issues/8015#issuecomment-476160981
-  echo '@include null' >> /etc/pam.d/systemd-user
+
   # Enable and Start SSH
   systemctl enable --now --no-block ssh
 }


weizhouapache:
image

weizhouapache:
@blueorangutan package

DaanHoogland (approved):
looks good, the original bug for which this was introduced is fixed. I see no reference to any problems, so it should be good. Not sure if we can test anything other than the SVMs running fine.

shwstppr:
@weizhouapache this will require new systemvm template?

weizhouapache:
> @weizhouapache this will require new systemvm template?

@shwstppr 
I think so
the changes with `init.sh` and `cloud-early-config` require new systemvm template

rohityadavcloud (approved):
LGTM - didn't test it but changes make sense, esp that the upstream systemd is fixed

rohityadavcloud:
@weizhouapache @DaanHoogland @harikrishna-patnala  I'm assuming this doesn't require a systemvmtemplate build (since changes are not in the cloud-early-config ?)

shwstppr:
@rohityadavcloud This may require new systemvm template build. https://github.com/apache/cloudstack/pull/8048#issuecomment-1754819284

rohityadavcloud:
Thanks @shwstppr @weizhouapache - that seems to be correct, but the other question is - is this fixing any issue? Is the failure of the service causing issues with systemvms and vrouters?

weizhouapache:
@rohityadavcloud 
this fixes the service user@0.service in VRs and system vms. 
but It seems other services in the vms are not impacted by the failure with `user@0.service`. this is a minor/trivial issue

rohityadavcloud:
Thanks for checking @weizhouapache cc @shwstppr - if necessary let's build systemvmtemplate for 4.19.0 ?

PR #8089, opened 2023-10-12
Fix failure on agent reconnection

vishesh92:
### Description

Depending on the agents' configuration, restarting a management server (preferred MS for the agent) will make the agent connect to another management server (non preferred MS). When the preferred MS comes back up, agent will try to disconnect with non-preferred MS and connect with the preferred MS. A race condition can happen during this process in which disconnection from non-preferred MS completes after the connection with preferred MS. This leads to agent to go into an `Alert` state. During this time, agent is still sending Ping to the preferred MS.

This PR solves this issue by:
* Taking a lock in database which ensures that for a host, only connection or disconnection can happen across different MS.
* While processing the `Ping` command if the Host is not in `Up` state, we request the agent to send a startup command again to the connection. If the startup is successful, the agent will come back in Up state.

To reproduce the issue,
1. apply the patch below.
```
diff --git a/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java b/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
index b74c11cf..aec8a8dd 100644
--- a/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
+++ b/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
@@ -836,6 +836,9 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
 
         removeAgent(attache, nextStatus);
         // update the DB
+        try {
+            Thread.sleep(5000L);
+        } catch (Exception e){}
         if (host != null && transitState) {
             disconnectAgent(host, event, _nodeId);
         }
```
2. Setup an environment with two management servers and below global configuration
```
agent.lb.enabled = true
indirect.agent.lb.algorithm = roundrobin
indirect.agent.lb.check.interval = 30
```
3. Restart the preferred management server.
4. Check status of host in database or in the UI

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?
1. Tried by applying the above patch.
2. Updated the state of a host to `Alert` state in database. After getting a ping, it gets a startup command after which it turns back to `Up` state.

### DIFF
--- a/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
+++ b/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
@@ -40,6 +40,7 @@ import javax.naming.ConfigurationException;
 
 import com.cloud.configuration.Config;
 import com.cloud.utils.NumbersUtil;
+import com.cloud.utils.db.GlobalLock;
 import org.apache.cloudstack.agent.lb.IndirectAgentLB;
 import org.apache.cloudstack.ca.CAManager;
 import org.apache.cloudstack.engine.orchestration.service.NetworkOrchestrationService;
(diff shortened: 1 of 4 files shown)


vishesh92:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@blueorangutan test alma8 kvm-alma8

test/integration/smoke/test_host_ping.py
```diff
@@ -0,0 +1,112 @@
...
+        )
+        for host in listHost:
+            self.logger.debug('Hypervisor = {}'.format(host.id))
+
+
+        hostToTest = listHost[0]
+        sql_query = "UPDATE host SET status = 'Alert' WHERE uuid = '" + hostToTest.id + "'"
+        self.dbConnection.execute(sql_query)
+
+        hostUpInCloudstack = wait_until(40, 10, self.checkHostStateInCloudstack, "Up", hostToTest.id)
```
DaanHoogland, line 108: any logic behind the wait period  and retry count here?
default ping interval time two would be my logic.

weizhouapache (approved):
code lgtm

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -1120,18 +1131,26 @@ private AgentAttache handleConnectedAgent(final Link link, final StartupCommand[
...
-                ready = new ReadyCommand(host.getDataCenterId(), host.getId(), NumbersUtil.enableHumanReadableSizes);
-
-                if (!indirectAgentLB.compareManagementServerList(host.getId(), host.getDataCenterId(), agentMSHostList, lbAlgorithm)) {
-                    final List<String> newMSList = indirectAgentLB.getManagementServerList(host.getId(), host.getDataCenterId(), null);
-                    ready.setMsHostList(newMSList);
-                    ready.setLbAlgorithm(indirectAgentLB.getLBAlgorithmName());
-                    ready.setLbCheckInterval(indirectAgentLB.getLBPreferredHostCheckInterval(host.getClusterId()));
-                    s_logger.debug("Agent's management server host list is not up to date, sending list update:" + newMSList);
-                }
+                GlobalLock joinLock = getHostJoinLock(host.getId());
```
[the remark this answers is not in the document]
vishesh92: I am not sure what you mean by life-lock.

If you mean a deadlock, as per my understanding of Agent's implementation, it will first send a disconnect (not waiting for disconnect to complete on server) and then send a connect to a different server. So, it's unlikely. Even if that happens, it will wait for lock for only 60 seconds.

test/integration/smoke/test_host_ping.py
vishesh92: Default is 60 seconds. Maximum time it can take is almost 2 minutes (159.9seconds). 

Updating the change to check it every 30 seconds and max wait time as 4 minutes.

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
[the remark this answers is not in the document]
DaanHoogland: no @vishesh92, I did mean a life-lock. a dead-lock means it waits for a lock that won´t be released, A life lock means two parties keep trying but keep hitting each others lock. In this case a disconnect hold the lock while the connect on the other side tries for it and fails as the disconnect is busy (for just to long) and than the other way around.

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -798,49 +799,65 @@ public boolean stop() {
...
-        if (s_logger.isDebugEnabled()) {
-            s_logger.debug("Deregistering link for " + hostId + " with state " + nextStatus);
-        }
+                removeAgent(attache, nextStatus);
 
-        removeAgent(attache, nextStatus);
-        // update the DB
-        if (host != null && transitState) {
-            disconnectAgent(host, event, _nodeId);
+                // update the DB
```
DaanHoogland, line 850:
```suggestion
```

vishesh92: It was part of the initial code and we are updating the state on disconnecting the agent.

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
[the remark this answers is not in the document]
vishesh92: @DaanHoogland No. It won't happen. We disconnect only when the connection was successful with the management server.

shwstppr (approved):
Code LGTM

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
DaanHoogland: do you mean this comment should stay there @vishesh92 ?

vishesh92: IMO, yes. The comment should be more detailed instead.

vishesh92:
@blueorangutan package

vishesh92:
@blueorangutan test

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

Issue #7970, opened 2023-09-18
After force deleting a deploy-as-is template, the VMs using it cannot start 


vladimirpetrov:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
API, UI
~~~

##### CLOUDSTACK VERSION

~~~
4.16, 4.17, 4.18.1
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

If a deploy-as-is template (VMWare vApp, for example) is removed using 'Force delete', the VMs using it won't be able to start properly because the details in the 'template_deploy_as_is_details' are removed.

##### STEPS TO REPRODUCE

~~~
Register a deploy-as-is template (enabling the Read VM settings from OVA ), start a VM with this template, then remove the template using 'Force delete' and stop/start the VM.

Note: There is another regression bug in 4.17 - 4.18.1 that makes reproducing this one much harder:
https://github.com/apache/cloudstack/issues/7998

So to reproduce it you'll need a 4.16 environment or 4.18 branch that includes the fix.
~~~

##### EXPECTED RESULTS

~~~
The VM starts normally.
~~~

##### ACTUAL RESULTS

~~~
The VM cannot start.
~~~

DaanHoogland:
Vladi, I think if you use force delete you have the intention and thought about it. What would you expect to happen? I think for the VM to just keep working is not what the force delete functionality is for.

weizhouapache:
> Vladi, I think if you use force delete you have the intention and thought about it. What would you expect to happen? I think for the VM to just keep working is not what the force delete functionality is for.

I think the deploy-as-it details of template should be kept.

harikrishna-patnala:
@weizhouapache the problem is with this line
String value = property.isPassword() ? DBEncryptionUtil.decrypt(detail.getValue()) : detail.getValue();

where we are checking if the property is a password.

I'm thinking to add a new column something like "is_password" to the table "user_vm_deploy_as_is_details", this will remove the dependency on the template.
As an alternative as you said we can keep the template deploy as is details but thinking if it will keep those entries always in the database.

What you say ?

harikrishna-patnala:
had a discussion with @weizhouapache, going with the solution to keep the template details after it is deleted.

JoaoJandre:
Closing as this was fixed by #8115

PR #8115, opened 2023-10-18
Fix deploy as is VM start after template deletion

harikrishna-patnala:
### Description

This PR fixes the issue #7970 

Fix is to keep the template deploy as is details even after the template is deleted. VM start need to know whether the vapp property is password or not which are saved in template_deploy_as_is_details table.

### Types of changes

- [ ] build/CI

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

1. Registered a VMware template with "Read VM settings from OVA" enabled
2. Created a VM with it
3. Force delete the template
4. Stop and start the VM => succeeded

### DIFF
--- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
+++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
@@ -29,6 +29,8 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 
 import com.cloud.domain.Domain;
+import com.cloud.vm.VMInstanceVO;
+import com.cloud.vm.dao.VMInstanceDao;
 import org.apache.cloudstack.agent.directdownload.CheckUrlAnswer;
 import org.apache.cloudstack.agent.directdownload.CheckUrlCommand;
 import org.apache.cloudstack.annotation.AnnotationService;
@@ -142,6 +144,8 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
     private TemplateDeployAsIsDetailsDao templateDeployAsIsDetailsDao;
     @Inject
     private AnnotationDao annotationDao;
+    @Inject
+    VMInstanceDao _vmInstanceDao;
 
     @Override
     public String getName() {
@@ -664,6 +668,14 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
 
             // Remove template details
             templateDetailsDao.removeDetails(template.getId());
+
+            // Remove deploy-as-is details (if any and if there are no VMs using it)
+            if (template.isDeployAsIs()) {
+                List<VMInstanceVO> vmInstanceVOList = _vmInstanceDao.listNonExpungedByTemplate(template.getId());
+                if (CollectionUtils.isEmpty(vmInstanceVOList)) {
+                    templateDeployAsIsDetailsDao.removeDetails(template.getId());
+                }
+            }
 
             // Remove comments (if any)
             AnnotationService.EntityType entityType = template.getFormat().equals(ImageFormat.ISO) ?
--- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
+++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
@@ -666,16 +666,7 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
             Pair<Class<?>, Long> tmplt = new Pair<Class<?>, Long>(VirtualMachineTemplate.class, template.getId());
             _messageBus.publish(_name, EntityManager.MESSAGE_REMOVE_ENTITY_EVENT, PublishScope.LOCAL, tmplt);
 
-            // Remove template details
-            templateDetailsDao.removeDetails(template.getId());
-
-            // Remove deploy-as-is details (if any and if there are no VMs using it)
-            if (template.isDeployAsIs()) {
-                List<VMInstanceVO> vmInstanceVOList = _vmInstanceDao.listNonExpungedByTemplate(template.getId());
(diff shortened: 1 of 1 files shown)


harikrishna-patnala:
@blueorangutan package

weizhouapache (approved):
code lgtm

I am fine with this change, although this will lead to some records in the database.
I think this is safest way to do, unless we can 100% figure out what details needs to be kept after template is removed.

We probably need to look into the template details as well
```
            // Remove template details
            templateDetailsDao.removeDetails(template.getId());
```

weizhouapache:
> I think the best way is to copy template details to the ivm_instance_details on deploy

@DaanHoogland 
it may require some code changes (e.g. read property in vm deploy-as-is details instead of template deploy-as-is details )

DaanHoogland:
> > I think the best way is to copy template details to the ivm_instance_details on deploy
> 
> @DaanHoogland it may require some code changes (e.g. read property in vm deploy-as-is details instead of template deploy-as-is details )

yes, but that would solve the DB remnance. I am not demanding we change this, just suggesting what could be a better design choice.

harikrishna-patnala:
template_deploy_as_is_details looks like this
image

The corresponding values in user_vm_deploy_as_is_details looks like this
image

So I dont think we can directly copy the template_deploy_as_is_details to user_vm_deploy_as_is_details as it is.
If we have to do that, we need a logic to differentiate the properties entries of templates and their values in the same table "user_vm_deploy_as_is_details", which I'm not sure if it is possible. May be @nvazquez can help here.

Currently we are parsing everything from user_vm_deploy_as_is_details table and passing them to the resource.
https://github.com/shapeblue/cloudstack/blob/4a0636374931e4e6b562def76acb9f8673a7f22a/engine/storage/src/main/java/org/apache/cloudstack/storage/image/deployasis/DeployAsIsHelperImpl.java#L317-L324

To simplify all these, I thought it will be better to keep the template_deploy_as_is_details even after deletion.
Or else like I mentioned in the issue ticket, we need another column in user_vm_deploy_as_is details table which is "is_password", because thats the only field we are looking in template details table.

cc @DaanHoogland @weizhouapache

server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
```diff
@@ -665,9 +665,6 @@ public boolean delete(TemplateProfile profile) {
             // Remove template details
             templateDetailsDao.removeDetails(template.getId());
 
-            // Remove deploy-as-is details (if any)
```
shwstppr, line 668: I think we atleast need to add a check if the template is used by existing VMs or not. If it is not used then I don't see a problem removing deployasis details.
And maybe preserve specific details if needed.

harikrishna-patnala: added the check @shwstppr, please review it again.

harikrishna-patnala:
@blueorangutan package

server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
```diff
@@ -665,8 +669,13 @@ public boolean delete(TemplateProfile profile) {
             // Remove template details
```
DaanHoogland, line 669: letś delete this comment

server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
```diff
@@ -665,8 +669,13 @@ public boolean delete(TemplateProfile profile) {
...
             templateDetailsDao.removeDetails(template.getId());
 
-            // Remove deploy-as-is details (if any)
-            templateDeployAsIsDetailsDao.removeDetails(template.getId());
+            // Remove deploy-as-is details (if any and if there are no VMs using it)
```
DaanHoogland, line 672: letś make this a javadoc of a new method

DaanHoogland (approved):
some style critique but clgtm

DaanHoogland:
@blueorangutan test

shwstppr (approved):
Code LGTM

shwstppr:
@harikrishna-patnala are you making any more changes for review comments?

server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
harikrishna-patnala: Added a new method and java doc

harikrishna-patnala:
@blueorangutan package

rohityadavcloud:
@blueorangutan package

DaanHoogland:
@blueorangutan test matrix

DaanHoogland:
@blueorangutan test rocky8 xenserver-71

DaanHoogland:
@blueorangutan test alma9 vmware-70u3

rajujith (approved):
LGTM. Instance starts fine with this fix although the template (vAPP) is deleted, it was failing earlier. Also observed is that the data is retained in the table `template_deploy_as_is_details` even after the template deletion as required by this fix. As mentioned already here this has side effects.

shwstppr:
@blueorangutan package

shwstppr:
@blueorangutan package

DaanHoogland:
@blueorangutan test alma9 vmware-70u3

weizhouapache (approved):
code lgtm

harikrishna-patnala:
can this PR be merged, I see the above failures on other PRs also

DaanHoogland:
> can this PR be merged, I see the above failures on other PRs also

yes, we shouldn't ignore them, but I am going through them and will add any cleanup or added robustness in #8196

PR #8132, opened 2023-10-23
[NSX] Allow NSX isolated networks

nvazquez:
Allow creating NSX isolated networks (not as VPC tiers)


Pearl1594 (approved):
LGTM - NIce stuff @nvazquez !!

PR #8202, opened 2023-11-08
Clear pool id if volume allocation fails

anniejili:
### Description

When there is outage in storage API coincides with a VM start, the volume could be stuck in "Allocated" state but with a pool Id assigned in DB. It blocked the ability to start the VM  during planDeployment process,  in the method storagePoolCompatibleWithVolumePool() shown below when validating a volume with a storage pool, if the volume's pool Id is not null and state is Allocated, it would always return false for any storage pools. As a result, the allocator cannot find a suitable pool and it blocks the VM start process.

```
@Override
    public boolean storagePoolCompatibleWithVolumePool(StoragePool pool, Volume volume) {
        if (pool == null || volume == null) {
            return false;
        }

        if (volume.getPoolId() == null) {
            // Volume is not allocated to any pool. Not possible to check compatibility with other pool, let it try
            return true;
        }

        StoragePool volumePool = _storagePoolDao.findById(volume.getPoolId());
        if (volumePool == null) {
            // Volume pool doesn't exist. Not possible to check compatibility with other pool, let it try
            return true;
        }

        if (volume.getState() == Volume.State.Ready) {
            if (volumePool.getPoolType() == Storage.StoragePoolType.PowerFlex && pool.getPoolType() != Storage.StoragePoolType.PowerFlex) {
                return false;
            } else if (volumePool.getPoolType() != Storage.StoragePoolType.PowerFlex && pool.getPoolType() == Storage.StoragePoolType.PowerFlex) {
                return false;
            }
        } else {
            return false;
        }

        return true;
    }
```

This PR fixed the issue above by validating volumes's state and pool id before the deployment in DeploymentPlanningManagerImpl.java's findSuitablePoolsForVolumes() method. Before DeploymentPlanningManager is trying to find suitable storage pools for volumes, if any volume is in Allocated state with a pool ID, the pool id will be cleared.
**Testing**
For a stopped VM, manually updated its root volume state to "Allocated" with a pool ID. Then restart the VM. 

Before the change

2023-08-22 23:40:00,948 INFO  [o.a.c.a.c.u.v.StartVMCmd] (API-Job-Executor-70:ctx-xxxx job-xxxx ctx-xxxx) (logid:xxxx) No destination found for a deployment for VM instance {id: "xxxx", name: "xxxx, uuid: "xxxx", type="User"}

After the change

The VM is started successfully.

### Types of changes

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

The change is tested via added junit test cases. The change is also tested and verified in Cloud Stack development environment.

### DIFF
--- a/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
+++ b/server/src/main/java/com/cloud/deploy/DeploymentPlanningManagerImpl.java
@@ -1651,6 +1651,16 @@ StateListener<State, VirtualMachine.Event, VirtualMachine>, Configurable {
         for (VolumeVO toBeCreated : volumesTobeCreated) {
             s_logger.debug("Checking suitable pools for volume (Id, Type): (" + toBeCreated.getId() + "," + toBeCreated.getVolumeType().name() + ")");
 
+            if (toBeCreated.getState() == Volume.State.Allocated && toBeCreated.getPoolId() != null) {
+                toBeCreated.setPoolId(null);
+                if (!_volsDao.update(toBeCreated.getId(), toBeCreated)) {
+                    throw new CloudRuntimeException(String.format("Error updating volume [%s] to clear pool Id.", toBeCreated.getId()));
+                }
+                if (s_logger.isDebugEnabled()) {
+                    String msg = String.format("Setting pool_id to NULL for volume id=%s as it is in Allocated state", toBeCreated.getId());
+                    s_logger.debug(msg);
+                }
+            }
             // If the plan specifies a poolId, it means that this VM's ROOT
             // volume is ready and the pool should be reused.
             // In this case, also check if rest of the volumes are ready and can
--- a/server/src/test/java/com/cloud/deploy/DeploymentPlanningManagerImplTest.java
+++ b/server/src/test/java/com/cloud/deploy/DeploymentPlanningManagerImplTest.java
@@ -413,12 +413,12 @@ public class DeploymentPlanningManagerImplTest {
 
         _dpm.avoidDisabledResources(vmProfile, dc, avoids);
 
-        verify(_dpm, times(timesRouter)).isRouterDeployableInDisabledResources();
-        verify(_dpm, times(timesAdminVm)).isAdminVmDeployableInDisabledResources();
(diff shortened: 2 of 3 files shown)


rohityadavcloud:
@blueorangutan package

harikrishna-patnala:
@blueorangutan test

harikrishna-patnala (approved):
LGTM, Tested both root and data disks, VM started fine.

DaanHoogland (commented):
clgtm, on worry though; @anniejili Do we need to check if the volume is actually present on the pool? I.E. if the volume has been allocated before it might actually exist (even though it has the state of Allocated)

harikrishna-patnala:
> clgtm, one worry though; @anniejili Do we need to check if the volume is actually present on the pool? I.E. if the volume has been allocated before it might actually exist (even though it has the state of Allocated)

@DaanHoogland according to the volume state machine, volume will be in allocated state either as the starting state or when the operation is failed. So I dont think volume will actually present. I feel this code check is good enough.

DaanHoogland:
> > clgtm, one worry though; @anniejili Do we need to check if the volume is actually present on the pool? I.E. if the volume has been allocated before it might actually exist (even though it has the state of Allocated)
> 
> @DaanHoogland according to the volume state machine, volume will be in allocated state either as the starting state or when the operation is failed. So I dont think volume will actually present. I feel this code check is good enough.

ok @harikrishna-patnala.

DaanHoogland (approved):
CLGTM, simpel principle.

However I accept @harikrishna-patnala 's argument, it won't hurt to try if they can break it, as per the discussion here: https://github.com/apache/cloudstack/pull/8202#issuecomment-1805265855

DaanHoogland:
@blueorangutan test

anniejili:
Rebased with the latest 4.18 branch.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

anniejili:
Thanks everyone!

Issue #8086, opened 2023-10-12
add domain path to all entities


DaanHoogland:
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~

~~~

##### CLOUDSTACK VERSION

~~~

~~~

##### SUMMARY

as agreed in https://github.com/apache/cloudstack/pull/7386#issuecomment-1759616648, we need to make the enhancement by @pierrelefevre more generic.

##### STEPS TO REPRODUCE

~~~

~~~

##### EXPECTED RESULTS

~~~
localcloud) 🐱 > list networks filter=name,type,domainpath
{
  "count": 2,
  "network": [
    {
      "domainpath": "ROOT",
      "name": "bla",
      "type": "Isolated"
    },
    {
      "domainpath": "ROOT",
      "name": "sla",
      "type": "Shared"
    }
  ]
}
~~~

##### ACTUAL RESULTS

~~~
localcloud) 🐱 > list networks filter=name,type,domainpath
{
  "count": 2,
  "network": [
    {
      "name": "bla",
      "type": "Isolated"
    },
    {
      "domainpath": "ROOT",
      "name": "sla",
      "type": "Shared"
    }
  ]
}
~~~

rohityadavcloud:
@DaanHoogland is the fix required that we always send the domain path for all networks? Concern - could it slow down listing networks at scale?

DaanHoogland:
> @DaanHoogland is the fix required that we always send the domain path for all networks? Concern - could it slow down listing networks at scale?

@rohityadavcloud , we could think of making this an option on the call, if it really is a performance hog. I have not considdered that, just the automation consequence as described by the refenced [PR (comment)](https://github.com/apache/cloudstack/pull/7386#issuecomment-1759616648) and the original issue #7363. I am not sure but I think the issue was with automation using terraform.

rohityadavcloud:
Thanks @DaanHoogland let me figure out and raise a PR. (related - the list networks API right now, works by listing all networks and then return a page out of it - pagination isn't implemented in it properly; likely @abh1sar 's PR may fix it)

PR #8207, opened 2023-11-09
Adding a default setter for domainPath in ControlledEntityResponse and populated the same in createNetworkResponse method of ApiResponseHelper  #8086

isupersky:
### Description

- Adding a default method setDomainPath in ControlledEntityResponse to avoid creating a variable and setter definitions in all implementations of ControlledEntityResponse  class. 
- and populated the same in createNetworkResponse method of ApiResponseHelper

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

**Before-**
Screenshot from 2023-11-09 16-14-55

**After-**
Screenshot from 2023-11-09 16-15-17

### How Has This Been Tested?

Created a guest network and called the `list networks filter=name,type,domainpath`  before and after the code change

### DIFF
--- a/server/src/main/java/com/cloud/api/ApiResponseHelper.java
+++ b/server/src/main/java/com/cloud/api/ApiResponseHelper.java
@@ -2583,9 +2583,10 @@ public class ApiResponseHelper implements ResponseGenerator {
                 Domain domain = ApiDBUtils.findDomainById(domainNetworkDetails.first());
                 if (domain != null) {
                     response.setDomainId(domain.getUuid());
-
-                    StringBuilder domainPath = new StringBuilder("ROOT");
-                    (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
+                    StringBuilder domainPath = new StringBuilder();
+                    if(!ObjectUtils.isEmpty(domain.getPath())){
+                        (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
+                    }
                     response.setDomainPath(domainPath.toString());
                 }
             }
@@ -2598,8 +2599,10 @@ public class ApiResponseHelper implements ResponseGenerator {
             if (domain != null) {
                 response.setDomainId(domain.getUuid());
                 response.setDomainName(domain.getName());
-                StringBuilder domainPath = new StringBuilder("ROOT");
-                (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
+                StringBuilder domainPath = new StringBuilder();
+                if(!ObjectUtils.isEmpty(domain.getPath())){
+                    (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
+                }
                 response.setDomainPath(domainPath.toString());
             }
 
@@ -2849,8 +2852,10 @@ public class ApiResponseHelper implements ResponseGenerator {
         Domain domain = ApiDBUtils.findDomainById(object.getDomainId());
         response.setDomainId(domain.getUuid());
         response.setDomainName(domain.getName());
-        StringBuilder domainPath = new StringBuilder("ROOT");
-        (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
+        StringBuilder domainPath = new StringBuilder();
+        if(!ObjectUtils.isEmpty(domain.getPath())){
+            (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
+        }
         response.setDomainPath(domainPath.toString());
     }
 
--- a/server/src/main/java/com/cloud/api/ApiResponseHelper.java
+++ b/server/src/main/java/com/cloud/api/ApiResponseHelper.java
@@ -2583,14 +2583,7 @@ public class ApiResponseHelper implements ResponseGenerator {
                 Domain domain = ApiDBUtils.findDomainById(domainNetworkDetails.first());
                 if (domain != null) {
                     response.setDomainId(domain.getUuid());
-                    StringBuilder domainPath = new StringBuilder();
-                    if(ObjectUtils.isEmpty(domain.getPath()) || domain.getPath().equals("/")){
-                        domainPath.append("/");
-                    }
-                    else{
-                        (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
-                    }
-                    response.setDomainPath(domainPath.toString());
+                    response.setDomainPath(getDomainPath(domain.getPath()));
                 }
             }
             response.setSubdomainAccess(domainNetworkDetails.second());
@@ -2602,14 +2595,7 @@ public class ApiResponseHelper implements ResponseGenerator {
             if (domain != null) {
                 response.setDomainId(domain.getUuid());
                 response.setDomainName(domain.getName());
-                StringBuilder domainPath = new StringBuilder();
-                if(ObjectUtils.isEmpty(domain.getPath()) || domain.getPath().equals("/")){
-                    domainPath.append("/");
-                }
-                else{
-                    (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
-                }
-                response.setDomainPath(domainPath.toString());
+                response.setDomainPath(getDomainPath(domain.getPath()));
             }
 
         }
@@ -2858,14 +2844,7 @@ public class ApiResponseHelper implements ResponseGenerator {
         Domain domain = ApiDBUtils.findDomainById(object.getDomainId());
         response.setDomainId(domain.getUuid());
         response.setDomainName(domain.getName());
-        StringBuilder domainPath = new StringBuilder();
-        if(ObjectUtils.isEmpty(domain.getPath()) || domain.getPath().equals("/")){
-            domainPath.append("/");
-        }
-        else{
-            (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
-        }
-        response.setDomainPath(domainPath.toString());
+        response.setDomainPath(getDomainPath(domain.getPath()));
     }
 
     private void populateOwner(ControlledViewEntityResponse response, ControlledEntity object) {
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -2598,6 +2598,9 @@ public NetworkResponse createNetworkResponse(ResponseView view, Network network)
             if (domain != null) {
                 response.setDomainId(domain.getUuid());
                 response.setDomainName(domain.getName());
+                StringBuilder domainPath = new StringBuilder("ROOT");
```
DaanHoogland, line 2601: @isupersky , I think we should not prepend "ROOT" in case there is any other path following (like "ROOT/mydomain").
We could replace if it is "/", I have no preference on that.

isupersky: @DaanHoogland  Please correct me if I'm wrong

what I understand is -
- "ROOT" should not be prepended  in any domainPath added in this PR.
- and if somewhere this has been done previously also then it should be replaced with "/"

DaanHoogland: no, that is not what I meant
if the root domain is the it we return "/" or instead "ROOT", or even "".
in all other cases we return something like "/domain/subdomain/..."

I would not bother about other places withing the scope of this PR (unless you feel like it)

DaanHoogland: looks like you went for the empty string right?

isupersky: yes, or else we can go with ->

`StringBuilder domainPath = new StringBuilder(); `
        `if(ObjectUtils.isEmpty(domain.getPath())){`
           ` domainPath.append("/");`
       ` }`
        `else{`
           ` (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);`
       ` }`
        `response.setDomainPath(domainPath.toString());`

DaanHoogland: i like that, to creep the scope a bit, we could even make the "/" a configuration item ;)
but lgtm!!!

isupersky: made the changes.  

Not sure, what you mean by configurable here.  is it that rather than using a string "/" multiple times, i should extract it in single variable? 

something like -> `private static final String defaultDomainPath = "/";`

DaanHoogland: You can have a look at the uses of `ConfigKey<>` , you'd be using something like `defaultRootDomainName.value()` and the operator can than change that. But let's not do that. It is fine like so.

shwstppr:
@blueorangutan package

server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -2583,9 +2583,13 @@ public NetworkResponse createNetworkResponse(ResponseView view, Network network)
                 Domain domain = ApiDBUtils.findDomainById(domainNetworkDetails.first());
                 if (domain != null) {
                     response.setDomainId(domain.getUuid());
-
-                    StringBuilder domainPath = new StringBuilder("ROOT");
-                    (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
+                    StringBuilder domainPath = new StringBuilder();
+                    if(ObjectUtils.isEmpty(domain.getPath()) || domain.getPath().equals("/")){
```
DaanHoogland, line 2587: not sure what you intent to do here @isupersky . it seems now that you append / to / resulting in //???
also you keep changing this in three locations. Would be good to have a method so we can change the code as needed in only one location.

isupersky: @DaanHoogland there are two variables here, domain.getPath() and domainPath.
domainPath is initially empty `StringBuilder domainPath = new StringBuilder();`  , hence it will only be initialised with single "/".

Thanks for the second suggestion,  I'll move the code repetition into a method.

rohityadavcloud:
@blueorangutan package

DaanHoogland:
@isupersky can you resolve the conflict?

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test keepEnv

server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -2585,10 +2585,7 @@ public NetworkResponse createNetworkResponse(ResponseView view, Network network)
                 Domain domain = ApiDBUtils.findDomainById(domainNetworkDetails.first());
                 if (domain != null) {
                     response.setDomainId(domain.getUuid());
-
-                    StringBuilder domainPath = new StringBuilder("ROOT");
-                    (domainPath.append(domain.getPath())).deleteCharAt(domainPath.length() - 1);
-                    response.setDomainPath(domainPath.toString());
+                    response.setDomainPath(getDomainPath(domain.getPath()));
```
shwstppr, line 2588: Didn't test but will this cause a change in response with `ROOT` not getting prepended

shwstppr (changes requested):
Adding setDomainPath() method in ControlledEntityResponse makes sense but it is an interface and the method is not overridden in any of the implemented classes. Only NetworkResponse will have domainpath param which was having it even before this PR and the method there as well is not overridden.

server/src/main/java/com/cloud/api/ApiResponseHelper.java
DaanHoogland: is there any functional reason to prepend "ROOT"? It is already equivilent with "/", functionally and the current code returns an empty string.

shwstppr: @DaanHoogland cc @isupersky I'm okay either. My only concern was that users/API consumers would start seeing a difference. Will that be an issue?

DaanHoogland: I think "ROOT" is a representational depiction of "/", and should be only shown in the UI. It makes no sense in the API and leads to less convenient processing. I would not want it in an API. Right now the response parameter `domainpath` is empty. I tested the difference:
```
(localcloud) 🐱 > list networks filter=domainpath
```
before:
```
{
  "count": 1,
  "network": [
    {}
  ]
}
```
after:
```
{
  "count": 1,
  "network": [
    {
      "domainpath": "/"
    }
  ]
}
```

isupersky:
> Adding setDomainPath() method in ControlledEntityResponse makes sense but it is an interface and the method is not overridden in any of the implemented classes. Only NetworkResponse will have domainpath param which was having it even before this PR and the method there as well is not overridden.

DaanHoogland:
@shwstppr you "requested changes" but it not entirely clear what changes. Do you require "ROOT" to be returned? As said in https://github.com/apache/cloudstack/pull/8207#discussion_r1395338161, I think this should only be so in the UI.

isupersky:
> Adding setDomainPath() method in ControlledEntityResponse makes sense but it is an interface and the method is not overridden in any of the implemented classes. Only NetworkResponse will have domainpath param which was having it even before this PR and the method there as well is not overridden.

@shwstppr  I added it at two new flows of createNetworkResponse in ApiResponseHelper. I'm not sure which other implementation class would require domainpath for ListNetwork API ?

shwstppr:
> > Adding setDomainPath() method in ControlledEntityResponse makes sense but it is an interface and the method is not overridden in any of the implemented classes. Only NetworkResponse will have domainpath param which was having it even before this PR and the method there as well is not overridden.
> 
> @shwstppr I added it at two new flows of createNetworkResponse in ApiResponseHelper. I'm not sure which other implementation class would require domainpath for ListNetwork API ?

In that case please update PR title/description that the change is limited to listNetworks API and not all implementations of ControlledEntityResponse class

server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -5109,4 +5108,15 @@ public IpQuarantineResponse createQuarantinedIpsResponse(PublicIpQuarantine quar
 
         return quarantinedIpsResponse;
     }
+
+    private String getDomainPath(String path){
+        StringBuilder domainPath = new StringBuilder();
+        if(ObjectUtils.isEmpty(path) || path.equals("/")){
```
vishesh92, line 5114: IMO. Since we know that path is a string, better to use `org.apache.commons.lang3.StringUtils.isEmpty()`..

shwstppr:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland (approved):
clgtm

DaanHoogland:
@shwstppr @vishesh92 Are you alright with the code like this?

server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -5159,4 +5158,15 @@ public BucketResponse createBucketResponse(Bucket bucket) {
         populateAccount(bucketResponse, bucket.getAccountId());
         return bucketResponse;
     }
+
+    private String getDomainPath(String path){
+        StringBuilder domainPath = new StringBuilder();
+        if (StringUtils.isEmpty(path) || path.equals("/")){
+            domainPath.append("/");
+        }
+        else{
```
DaanHoogland, line 5167: @isupersky , @vishesh92 meant also to change 
```suggestion
        else {
```

server/src/main/java/com/cloud/api/ApiResponseHelper.java
```diff
@@ -5159,4 +5158,15 @@ public BucketResponse createBucketResponse(Bucket bucket) {
...
+            (domainPath.append(path)).deleteCharAt(domainPath.length() - 1);
```
vishesh92, line 5168:
```suggestion
        if (StringUtils.isEmpty(path) || path.equals("/")) {
            domainPath.append("/");
        } else {
            (domainPath.append(path)).deleteCharAt(domainPath.length() - 1);
```
@isupersky please note the code style changes.

server/src/main/java/com/cloud/api/ApiResponseHelper.java
isupersky: sorry, missed that. pushing the changes right away

shwstppr:
@blueorangutan package

sureshanaparti:
hi @isupersky please fix the conflicts in the branch.

DaanHoogland:
@isupersky , closing this one as it has conflicts and is old. please update and reopen if you think it is still relevant.

PR #8234, opened 2023-11-14
Allow altering only either CPU or memory during VM live scale

GaOrtiga:
### Description

The API  `scaleVirtualMachine`  requires the specification of both vCPU quantity and memory amount, resulting in an error if either was omitted. This means that if an operator wants to increase the quantity of vCPUs while maintaining the existing memory, they were obligate to provide the current memory value as well.

An alteration has been introduced to the API to allow operators to input only the parameters they intend to modify while retaining the current values for the rest.

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

### How Has This Been Tested?

I applied the changes in a local lab and verified that if I used the API while inputting only one of the parameters, the scale was successful, and the  others kept the current value.

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -1237,9 +1237,9 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
     protected void updateInstanceDetails (Map<String, String> details, VirtualMachine vmInstance, Long newServiceOfferingId) {
         ServiceOfferingVO currentServiceOffering = serviceOfferingDao.findByIdIncludingRemoved(vmInstance.getId(), vmInstance.getServiceOfferingId());
         ServiceOfferingVO newServiceOffering = serviceOfferingDao.findById(newServiceOfferingId);
-        updateInstanceDetailsWithCurrentValue(newServiceOffering.getSpeed(), details, VmDetailConstants.CPU_SPEED, currentServiceOffering.getSpeed());
-        updateInstanceDetailsWithCurrentValue(newServiceOffering.getRamSize(), details, VmDetailConstants.MEMORY, currentServiceOffering.getRamSize());
-        updateInstanceDetailsWithCurrentValue(newServiceOffering.getCpu(), details, VmDetailConstants.CPU_NUMBER, currentServiceOffering.getCpu());
+        updateInstanceDetailsKeepCurrentValueIfNull(newServiceOffering.getSpeed(), details, VmDetailConstants.CPU_SPEED, currentServiceOffering.getSpeed());
+        updateInstanceDetailsKeepCurrentValueIfNull(newServiceOffering.getRamSize(), details, VmDetailConstants.MEMORY, currentServiceOffering.getRamSize());
+        updateInstanceDetailsKeepCurrentValueIfNull(newServiceOffering.getCpu(), details, VmDetailConstants.CPU_NUMBER, currentServiceOffering.getCpu());
     }
 
     /**
@@ -1251,7 +1251,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
      * @param currentValue     the current value of the detail constant
      */
 
-    protected void updateInstanceDetailsWithCurrentValue(Integer newValue, Map<String, String> details, String detailsConstant, Integer currentValue) {
+    protected void updateInstanceDetailsKeepCurrentValueIfNull(Integer newValue, Map<String, String> details, String detailsConstant, Integer currentValue) {
         if (newValue == null && details.get(detailsConstant) == null) {
             String currentValueString = String.valueOf(currentValue);
             s_logger.debug(String.format("[%s] was not specified, keeping the current value: [%s].", detailsConstant, currentValueString));
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -1925,7 +1925,6 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
         }
         CallContext.current().setEventDetails("Vm Id: " + vm.getUuid());
 
-//        boolean result = upgradeVirtualMachine(vmId, newServiceOfferingId, cmd.getDetails());
         Map<String, String> cmdDetails = cmd.getDetails();
 
         updateInstanceDetails(cmdDetails, vm, newServiceOfferingId);
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -1254,7 +1254,7 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
     protected void updateInstanceDetailsKeepCurrentValueIfNull(Integer newValue, Map<String, String> details, String detailsConstant, Integer currentValue) {
         if (newValue == null && details.get(detailsConstant) == null) {
             String currentValueString = String.valueOf(currentValue);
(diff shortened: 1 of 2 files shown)


server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -1891,7 +1924,12 @@ public UserVm upgradeVirtualMachine(ScaleVMCmd cmd) throws ResourceUnavailableEx
         }
         CallContext.current().setEventDetails("Vm Id: " + vm.getUuid());
 
-        boolean result = upgradeVirtualMachine(vmId, newServiceOfferingId, cmd.getDetails());
+//        boolean result = upgradeVirtualMachine(vmId, newServiceOfferingId, cmd.getDetails());
```
hsato03, line 1927:
```suggestion
```

DaanHoogland (approved):
agree with @hsato03 , code lgtm otherwise

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -1226,6 +1226,39 @@ public UserVm upgradeVirtualMachine(UpgradeVMCmd cmd) throws ResourceAllocationE
...
+    /**
+     * Updates a specific instance detail with the current instance value if the new value is null.
+     *
+     * @param newValue         the new value to be set
+     * @param details          a map of instance details
+     * @param detailsConstant  the name of the detail constant to be updated
+     * @param currentValue     the current value of the detail constant
+     */
+
+    protected void updateInstanceDetailsWithCurrentValue(Integer newValue, Map<String, String> details, String detailsConstant, Integer currentValue) {
```
JoaoJandre, line 1253:
```suggestion
    protected void updateInstanceDetailsKeepCurrentValueIfNull(Integer newValue, Map<String, String> details, String detailsConstant, Integer currentValue) {
```
The current name gives the impression that you're always updating using the current value.

JoaoJandre (approved):
CLGTM,  didn't test it

GutoVeronezi:
@hsato03, are all your concerns met?

Just for the sake: @blueorangutan package

hsato03 (approved):
> @hsato03, are all your concerns met?
> 
> Just for the sake: @blueorangutan package

Yes, didn't test but CLGTM.

shwstppr:
@borisstoyanov are you doing tests on this?

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan package

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -1227,6 +1227,39 @@ public UserVm upgradeVirtualMachine(UpgradeVMCmd cmd) throws ResourceAllocationE
...
+     * @param newValue         the new value to be set
+     * @param details          a map of instance details
+     * @param detailsConstant  the name of the detail constant to be updated
+     * @param currentValue     the current value of the detail constant
+     */
+
+    protected void updateInstanceDetailsKeepCurrentValueIfNull(Integer newValue, Map<String, String> details, String detailsConstant, Integer currentValue) {
+        if (newValue == null && details.get(detailsConstant) == null) {
+            String currentValueString = String.valueOf(currentValue);
+            s_logger.debug(String.format("[%s] was not specified, keeping the current value: [%s].", detailsConstant, currentValueString));
```
JoaoJandre, line 1257:
```suggestion
            logger.debug("{} was not specified, keeping the current value: {}.", detailsConstant, currentValueString);
```

DaanHoogland:
@GaOrtiga this had already passed but needs new attention since the log upgrade

DaanHoogland:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

GutoVeronezi:
Merging based on approvals, packaging and test results.

DaanHoogland:
As this is merged into main, I am setting the milestone to 20

weizhouapache:
@DaanHoogland @GutoVeronezi 
has anyone tested this PR ?

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -1892,7 +1925,11 @@ public UserVm upgradeVirtualMachine(ScaleVMCmd cmd) throws ResourceUnavailableEx
         }
         CallContext.current().setEventDetails("Vm Id: " + vm.getUuid());
 
-        boolean result = upgradeVirtualMachine(vmId, newServiceOfferingId, cmd.getDetails());
+        Map<String, String> cmdDetails = cmd.getDetails();
+
+        updateInstanceDetails(cmdDetails, vm, newServiceOfferingId);
+
+        boolean result = upgradeVirtualMachine(vmId, newServiceOfferingId, cmdDetails);
```
weizhouapache, line 1932: Sorry for late review.

@GaOrtiga 
if the upgrade fail, should the details be restored to the original values ?

ping @DaanHoogland @GutoVeronezi , anyone has tested it ?

GutoVeronezi: > ping @DaanHoogland @GutoVeronezi , anyone has tested it ?

Yes, it was tested.

> if the upgrade fail, should the details be restored to the original values ?

The PR does not address changes to this behavior. it remains the same; we can discuss it in an issue or another PR.

weizhouapache: > > ping @DaanHoogland @GutoVeronezi , anyone has tested it ?
> 
> Yes, it was tested.
> 
Ok, I asked it because I did not see a comment mentioning it has been manually tested ok.  If not, it would be good to be manually tested.
Why has tested it, except the author?

> > if the upgrade fail, should the details be restored to the original values ?
> 
> The PR does not address changes to this behavior. it remains the same; we can discuss it in an issue or another PR.
> 
> 
Do you think it is a regression of this PR we should address?

GutoVeronezi: > Why has tested it, except the author?

I don't think I understood your question.

> Do you think it is a regression of this PR we should address?

The PR does not address changes to method `upgradeVirtualMachine`; therefore, the topic you are questioning would not be a regression of this PR.

weizhouapache: > > Why has tested it, except the author?
> 
> I don't think I understood your question.
> 
Sorry typo. 
Who has tested it, except the author?

> > Do you think it is a regression of this PR we should address?
> 
> The PR does not address changes to method `upgradeVirtualMachine`; therefore, the topic you are questioning would not be a regression of this PR.
> 

This PR updates the vm instance details, but not restore the vm details if upgrade fails. It should be restored if catch an exception or result is false.
In the past, vm details are not updated so not need to restore in case of failures.

GutoVeronezi: > Who has tested it, except the author?

I tested it; I do not know if anybody else has tested it too.

> please ignore my question regarding the vm details, They are not saved in database before upgrade, so no need to restore.

Yes. The only thing this PR does is facilitate calling the API. For instance, without this PR, if one has a VM with 1 vCPU and 1 GB RAM and wants to scale it to 2 GB RAM, one has to call the API with both the current vCPU count and the new RAM size. With this PR, one will only have to pass the new RAM size in the parameters; ACS will retrieve the current VM's vCPU count and will add it to the `details` map to be updated; thus, passing the current vCPU count and the new RAM size will have the same behavior as just passing the new RAM size. It just updates the `details` map (if necessary) before scaling the VM. This applies to RAM, frequency and vCPU.

I created PR #8673 to improve the methods names and docs.

Issue #6744, opened 2022-09-15
Problem with operating systems that use cgroup v2 related to cpu speed. 


correajl:
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME
~~~
Cloudstack management and cloudstack agent. 
~~~

##### CLOUDSTACK VERSION
~~~
4.17.0.1
~~~

##### CONFIGURATION
2 management servers, 2 databases, advanced network, everything working fine. 

##### OS / ENVIRONMENT
Ubuntu Server 22.04 LTS, KVM, libvirt 8.0.0-1ubuntu7.1, cgroup2.

##### SUMMARY
The value of "CPU (in MHz)" used in some compute offering definitions is mapped to 'shares' element in 'cputune' section of domain definition file (xml). According to [libvirt documentation](libvirt.org/formatdomain.html) the value should be in range [2, 262144]. But, for operating systems using cgroup v2 the maximum value is 10000. I know that Ubuntu 22.04 I'm using here is not supported yet. But this will be an issue as other OSs adopt cgroup v2 too. So, I think this parameter deserves attention. If the value of (N. CPU) * CPU (in MHz) is greater than 10000 I get "**Value specified in CPUWeight is out of range**" in hypervisor.

As a workaround I configured service offering with 1 Mhz. This implies that VMs with more CPUs have much more chance to get the CPU than VMs with lower number of CPUs, because besides having more CPUs, they have more chance to get the host' CPUs.

VM1 10 CPU * 1 MHz -> shares = 10
VM2 80 CPU * 1 MHz -> shares = 80

If we look at 1st CPU of VM2, it will have 8 times more chance to get the host's CPU than 1st CPU of VM1. 

Shouldn't all virtual CPUs have the same chance of using one host CPU?

##### STEPS TO REPRODUCE
Using Ubuntu 22.04 or other OS that uses cgroup v2, create a compute offering with a CPU (in MHz) value of 1000. Try to launch an instance with more than 10 CPUs. It will be generated a configuration file for the new instance with shares > 10000 inside cputune section. Hypervisor can't launch the instance with _Value specified in CPUWeight is out of range_ error.

##### EXPECTED RESULTS
Create a way for all virtual CPUs to have an equal chance of using a host CPU. 
Check cgroup version and try to not generate values bigger than 10000. 

##### ACTUAL RESULTS
[...]

PR #8252, opened 2023-11-20
kvm: Add support for cgroupv2

BryanMLima:
## <section id="problem-description">1. Problem description</section>

In Apache CloudStack (ACS), when a VM is deployed in a host with the KVM hypervisor, an XML file is created in the assigned host, which has a property <code>shares</code> that defines the weight of the VM to access the host CPU. The value of this property has no unit, and it is a relative measure to calculate how much CPU a given VM will have in the host. However, this value has a limit, which depends on the version of cgroup utilized by the host's kernel. The problem lies at the range value of shares that varies between both versions: [2, 264144] for cgroups version 1; and [1, 10000] for cgroups version 2. Currently, ACS calculates the value of <code>shares</code> using Equation 1, presented below, where <code>CPU</code> is the number of cores and <code>speed</code> is the CPU frequency; both specified in the VM's compute offering. Therefore, if a compute offering has, for example, 6 cores at 2 GHz, the <code>shares</code> value will be 12000 and an exception will be thrown by libvirt if the host utilizes cgroup v2. The second version is becoming the default one in current Linux distributions; thus, it is necessary to address this limitation.

- **Equation 1**
  <code>shares = CPU * speed</code>

---
Fixes: #6744 

## <section id="proposed-changes">2. Proposed changes</section>

To address the problem described, we propose to apply a scale conversion considering the max <code>shares</code> of the host. Using the same formula currently utilized by ACS, it is possible to calculate the maximum <code>shares</code> of a VM for a given host. In other words, using the number of cores and the nominal speed of the host's CPU as the upper limit of <code>shares</code> allowed to a VM. Then, this value will be scaled to the allowed interval of [1, 10000] of cgroup v2 by using a linear scale conversion.

The VM <code>shares</code> would be calculated as Equation 2, presented below, where <code>VM requested shares</code> is the requested <code>shares</code> value calculated using Equation 1, <code>cgroup upper limit</code> is fixed with a value of 10000 (cgroups v2 upper limit), and <code>host max shares</code> is the maximum <code>shares</code> value of the host, calculated using Equation 1. Using Equation 2, the only case where a VM passes the cgroup v2 limit is when the user requests more resources than the host has, which is not possible with the current implementation of ACS.

- **Equation 2**
  <code>shares = (VM requested shares * cgroup upper limit)/host max shares</code>

To implement the proposal, the following APIs will be updated: <code>deployVirtualMachine</code>, <code>migrateVirtualMachine</code> and <code>scaleVirtualMachine</code>. When a VM is being deployed, a new verification will be added to find a suitable host. The max <code>shares</code> of each host will be calculated, and the VM calculated <code>shares</code> will be verified if it does not surpass the host's value. Likewise, the migration of VMs will have a similar new verification. Lastly, the scale of VMs will also have the same verification for the VM's host.

To determine the max <code>shares</code> of a given host, we will use the same equation currently used in ACS for calculating the <code>shares</code> of VMs, presented in <a href="#problem-description" class="internal-link">Section 1</a>. When Equation 1 is used to determine the maximum <code>shares</code> of a host, <code>CPU</code> is the number of cores of the host, and <code>speed</code> is the nominal CPU speed, i.e., considering the CPU's base frequency.

It is important to note that these changes are only for hosts with the KVM hypervisor using cgroup v2 for now.

## <section id="example">Example</section>

To exemplify the proposed changes, consider a host with the following specification: 32 CPU cores with nominal speed of 2 GHz; and a VM with a compute offering with 8 CPU cores and with speed of 2 GHz. With the current ACS implementation, the <code>shares</code> of the VM would be calculated as Equation 1. Thus, the VM <code>shares</code> would be 16000, over the cgroup v2 limit of 10000.

With the proposed changes, the VM <code>shares</code> would be calculated as Equation 2. In this example, <code>VM requested shares</code> is 16000, <code>cgroup upper limit</code> is fixed with a value of 10000, and <code>host max shares</code> is 64000. Therefore, the VM <code>shares</code> results in 2500, well below the cgroup v2 limit.

## <section id="real-case-scenarios">Real case scenarios</section>

To demonstrate real case scenarios, consider the following hosts:

- **Host A**
  - **\# of Cores:** 32
  - **CPU nominal frequency:** 2 GHz
  - **Max Shares:** 64000
- **Host B**
  - **\# of Cores:** 16
  - **CPU nominal frequency:** 2 GHz
  - **Max Shares:** 32000

Table 1 below presents a set of VMs with their requested resources, alongside the <code>shares</code> values considering the current implementation, and the new <code>shares</code> value, for each host, considering the proposed change using Equation 2.

- Table 1

<table>

<thead>

<tr>

<th rowspan="2">VM</th>

<th rowspan="2">CPU cores</th>

<th rowspan="2">CPU frequency (GHz)</th>

<th rowspan="2">Current shares</th>

<th colspan="2">New shares</th>

</tr>

<tr>

<th>For Host A</th>

<th>For Host B</th>

</tr>

</thead>

<tbody>

<tr>

<td>VM 1</td>

<td>2</td>

<td>2</td>

<td>4000</td>

<td>625</td>

<td>1250</td>

</tr>

<tr>

<td>VM 2</td>

<td>4</td>

<td>2</td>

<td>8000</td>

<td>1250</td>

<td>2500</td>

</tr>

<tr>

<td>VM 3</td>

<td>6</td>

<td>2</td>

<td>12000</td>

<td>1875</td>

<td>3750</td>

</tr>

<tr>

<td>VM 4</td>

<td>8</td>

<td>2</td>

<td>16000</td>

<td>2500</td>

<td>5000</td>

</tr>

<tr>

<td>VM 5</td>

<td>16</td>

<td>2</td>

<td>32000</td>

<td>5000</td>

<td>10000</td>

</tr>

<tr>

<td>VM 6</td>

<td>32</td>

<td>2</td>

<td>64000</td>

<td>10000</td>

<td>20000</td>

</tr>

</tbody>

</table>

Table 2 below presents if the same VMs in Table 1 would be allowed to be allocated to a given host, or if an exception would be thrown, considering current and proposed implementations. As we can see, with the current ACS implementation, VMs 3 through 6 would throw an exception when deploying in host A; even though the host has enough resources. VM 6 should throw an exception when trying to deploy it in host B in both implementations, as the host does not have enough resources to allocate it.

- Table 2

<table>

<thead>

<tr>

<th rowspan="2">VM</th>

<th colspan="2">Host A<br></th>

<th colspan="2">Host B</th>

</tr>

<tr>

<th>Current Implementation</th>

<th>Proposed Implementation</th>

<th>Current Implementation</th>

<th>Proposed Implementation</th>

</tr>

</thead>

<tbody>

<tr>

<td>VM 1</td>

<td>Allowed</td>

<td>Allowed</td>

<td>Allowed</td>

<td>Allowed</td>

</tr>

<tr>

<td>VM 2</td>

<td>Allowed</td>

<td>Allowed</td>

<td>Allowed</td>

<td>Allowed</td>

</tr>

<tr>

<td>VM 3</td>

<td>Exception</td>

<td>Allowed</td>

<td>Exception</td>

<td>Allowed</td>

</tr>

<tr>

<td>VM 4</td>

<td>Exception</td>

<td>Allowed</td>

<td>Exception</td>

<td>Allowed</td>

</tr>

<tr>

<td>VM 5</td>

<td>Exception</td>

<td>Allowed</td>

<td>Exception</td>

<td>Allowed</td>

</tr>

<tr>

<td>VM 6</td>

<td>Exception</td>

<td>Allowed</td>

<td>Exception</td>

<td>Exception</td>

</tr>

</tbody>

</table>

It is important to note that Equation 2 rounds up the <code>shares</code> value; thus, there is a precision loss with the conversion. Nevertheless, this precision loss should not be noticeable to the end user, as the <code>shares</code> value would need to be in a very close interval, e.g. <code>shares</code> values of <code>3997</code>, <code>3998</code> and <code>3999</code> would be considered as <code>1249</code> in host B with the new implementation. However, the precision loss is a small drawback for enabling support of cgroup v2 to ACS.

## <section id="future-works">3. Future works</section>

With the current proposal, only cgroups version 2 is addressed, as it has impactful limitations. Thus, as future work, cgroups version 1 will also be addressed using the same strategy of linear scale conversion.

### Types of changes

- [ ] build/CI

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Consider the host `host-1` with cgroup v2, `host-2` with cgroup v1 and the following custom constrained compute offering below:

- **Frequency:** 2 GHz
- **Cores:** 2 - 6
- **RAM**: 1 GB

## Deploy of VMs

I created the VM `vm-tmpfs` with 5 cores and allocated it to host `host-2`.
```bash
ubuntu@host-2:~$ virsh dumpxml --domain i-2-13-VM | grep shares
    <shares>10000</shares>
```

I created the VM `vm-cgroupv2` with 5 cores and allocated it to host `host-1`.
```bash
ubuntu@host-1:~$ virsh dumpxml --domain i-2-15-VM | grep shares
    <shares>6667</shares>
```

As expected, ACS considered the host resources to set the `shares` values. When the host utilizes the cgroup v1, the default behavior is not changed.

## VM live scale

I lived scale the VM `vm-tmpfs`, changing its number of cores from 5 to 6.

```bash
ubuntu@host-2:~$ virsh dumpxml --domain i-2-13-VM | grep shares
    <shares>10000</shares>
ubuntu@host-2:~$ virsh dumpxml --domain i-2-13-VM | grep shares
    <shares>12000</shares>
```

I lived scale the VM `vm-cgroupv2`, changing its number of cores from 5 to 6.
```bash
ubuntu@host-1:~$ virsh dumpxml --domain i-2-15-VM | grep shares
    <shares>6667</shares>
ubuntu@host-1:~$ virsh dumpxml --domain i-2-15-VM | grep shares
    <shares>8000</shares>
```

As expected, ACS considered the host resources to set the `shares` values. When the host utilizes the cgroup v1, the default behavior is not changed.

## VM migration

I migrated VM `vm-tmpfs` from host `host-2` to host `host-1` (from cgroupv1 to cgroupv2). After the migration, the `shares` values was changed to `8000`, as expected.
```bash
ubuntu@host-1:~$ virsh dumpxml --domain i-2-13-VM | grep shares
    <shares>8000</shares>
```

I migrated VM `vm-cgroupv2` from host `host-1` to host `host-2` (from cgroupv2 to cgroupv1). After the migration, the `shares` values was changed to `12000`, as expected.
```bash
ubuntu@host-2:~$ virsh dumpxml --domain i-2-15-VM | grep shares
    <shares>12000</shares>
```

I migrated VMs between hosts with different cgroup versions, the VM migration section above describes this.

### DIFF
--- a/core/src/main/java/com/cloud/agent/api/MigrateCommand.java
+++ b/core/src/main/java/com/cloud/agent/api/MigrateCommand.java
@@ -40,6 +40,9 @@ public class MigrateCommand extends Command {
     private boolean executeInSequence = false;
     private List<MigrateDiskInfo> migrateDiskInfoList = new ArrayList<>();
     private Map<String, DpdkTO> dpdkInterfaceMapping = new HashMap<>();
+
+    private int newVmCpuShares;
+
     Map<String, Boolean> vlanToPersistenceMap = new HashMap<>();
 
     public Map<String, DpdkTO> getDpdkInterfaceMapping() {
@@ -136,6 +139,14 @@ public class MigrateCommand extends Command {
 
     public void setMigrateDiskInfoList(List<MigrateDiskInfo> migrateDiskInfoList) {
         this.migrateDiskInfoList = migrateDiskInfoList;
+    }
+
+    public int getNewVmCpuShares() {
+        return newVmCpuShares;
+    }
+
+    public void setNewVmCpuShares(int newVmCpuShares) {
+        this.newVmCpuShares = newVmCpuShares;
     }
 
     public static class MigrateDiskInfo {
--- a/core/src/main/java/com/cloud/agent/api/PrepareForMigrationAnswer.java
+++ b/core/src/main/java/com/cloud/agent/api/PrepareForMigrationAnswer.java
@@ -28,6 +28,8 @@ public class PrepareForMigrationAnswer extends Answer {
 
     private Map<String, DpdkTO> dpdkInterfaceMapping = new HashMap<>();
 
+    private Integer newVmCpuShares = null;
+
     protected PrepareForMigrationAnswer() {
     }
 
@@ -49,5 +51,13 @@ public class PrepareForMigrationAnswer extends Answer {
 
     public Map<String, DpdkTO> getDpdkInterfaceMapping() {
         return this.dpdkInterfaceMapping;
+    }
+
+    public Integer getNewVmCpuShares() {
+        return newVmCpuShares;
+    }
+
+    public void setNewVmCpuShares(Integer newVmCpuShares) {
+        this.newVmCpuShares = newVmCpuShares;
     }
 }
--- a/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
+++ b/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java
@@ -48,6 +48,7 @@ import javax.naming.ConfigurationException;
 import javax.persistence.EntityExistsException;
 
 import com.cloud.event.ActionEventUtils;
+import com.google.gson.Gson;
 import org.apache.cloudstack.affinity.dao.AffinityGroupVMMapDao;
 import org.apache.cloudstack.annotation.AnnotationService;
 import org.apache.cloudstack.annotation.dao.AnnotationDao;
@@ -2790,23 +2791,9 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
         }
 
         boolean migrated = false;
-        Map<String, DpdkTO> dpdkInterfaceMapping = null;
+        Map<String, DpdkTO> dpdkInterfaceMapping = new HashMap<>();
         try {
-            final boolean isWindows = _guestOsCategoryDao.findById(_guestOsDao.findById(vm.getGuestOSId()).getCategoryId()).getName().equalsIgnoreCase("Windows");
-            Map<String, Boolean> vlanToPersistenceMap = getVlanToPersistenceMapForVM(vm.getId());
-            final MigrateCommand mc = new MigrateCommand(vm.getInstanceName(), dest.getHost().getPrivateIpAddress(), isWindows, to, getExecuteInSequence(vm.getHypervisorType()));
-            if (MapUtils.isNotEmpty(vlanToPersistenceMap)) {
-                mc.setVlanToPersistenceMap(vlanToPersistenceMap);
-            }
-
-            boolean kvmAutoConvergence = StorageManager.KvmAutoConvergence.value();
-            mc.setAutoConvergence(kvmAutoConvergence);
-            mc.setHostGuid(dest.getHost().getGuid());
-
-            dpdkInterfaceMapping = ((PrepareForMigrationAnswer) pfma).getDpdkInterfaceMapping();
-            if (MapUtils.isNotEmpty(dpdkInterfaceMapping)) {
-                mc.setDpdkInterfaceMapping(dpdkInterfaceMapping);
-            }
+            final MigrateCommand mc = buildMigrateCommand(vm, to, dest, pfma, dpdkInterfaceMapping);
 
             try {
                 final Answer ma = _agentMgr.send(vm.getLastHostId(), mc);
@@ -2876,6 +2863,43 @@ public class VirtualMachineManagerImpl extends ManagerBase implements VirtualMac
             work.setStep(Step.Done);
             _workDao.update(work.getId(), work);
         }
+    }
+
+    /**
+     * Create and set parameters for the {@link MigrateCommand} used in the migration and scaling of VMs.
+     */
+    protected MigrateCommand buildMigrateCommand(VMInstanceVO vmInstance, VirtualMachineTO virtualMachineTO, DeployDestination destination, Answer answer,
+                                                 Map<String, DpdkTO> dpdkInterfaceMapping) {
+        final boolean isWindows = _guestOsCategoryDao.findById(_guestOsDao.findById(vmInstance.getGuestOSId()).getCategoryId()).getName().equalsIgnoreCase("Windows");
+        final MigrateCommand migrateCommand = new MigrateCommand(vmInstance.getInstanceName(), destination.getHost().getPrivateIpAddress(), isWindows, virtualMachineTO,
+                getExecuteInSequence(vmInstance.getHypervisorType()));
+
+        Map<String, Boolean> vlanToPersistenceMap = getVlanToPersistenceMapForVM(vmInstance.getId());
+        if (MapUtils.isNotEmpty(vlanToPersistenceMap)) {
+            s_logger.debug(String.format("Setting VLAN persistence to [%s] as part of migrate command for VM [%s].", new Gson().toJson(vlanToPersistenceMap), virtualMachineTO));
+            migrateCommand.setVlanToPersistenceMap(vlanToPersistenceMap);
+        }
+
+        migrateCommand.setAutoConvergence(StorageManager.KvmAutoConvergence.value());
+        migrateCommand.setHostGuid(destination.getHost().getGuid());
+
+        PrepareForMigrationAnswer prepareForMigrationAnswer = (PrepareForMigrationAnswer) answer;
+
+        Map<String, DpdkTO> answerDpdkInterfaceMapping = prepareForMigrationAnswer.getDpdkInterfaceMapping();
+        if (MapUtils.isNotEmpty(answerDpdkInterfaceMapping) && dpdkInterfaceMapping != null) {
+            s_logger.debug(String.format("Setting DPDK interface mapping to [%s] as part of migrate command for VM [%s].", new Gson().toJson(vlanToPersistenceMap),
+                    virtualMachineTO));
+            dpdkInterfaceMapping.putAll(answerDpdkInterfaceMapping);
+            migrateCommand.setDpdkInterfaceMapping(dpdkInterfaceMapping);
+        }
+
+        Integer newVmCpuShares = prepareForMigrationAnswer.getNewVmCpuShares();
+        if (newVmCpuShares != null) {
+            s_logger.debug(String.format("Setting CPU shares to [%d] as part of migrate command for VM [%s].", newVmCpuShares, virtualMachineTO));
+            migrateCommand.setNewVmCpuShares(newVmCpuShares);
+        }
+
+        return migrateCommand;
     }
 
     private void updateVmPod(VMInstanceVO vm, long dstHostId) {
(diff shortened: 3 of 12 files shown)


DaanHoogland:
Thanks @BryanMLima , for picking this up and for the extensive explanation. I have two questions:

- where you say `6 cores at 2 GHz, the shares value will be 12000 ` I would say `huh` as it would either be 12 or 12.000.000.000. What is the factorial used there? (I probably just need a link to a definition)
- I see no considerations on live systems, i.e. upgrades. Please, expand on that. Will it have any consequence, or will it be seamless?

regards,

DaanHoogland:
Oh, and number 3

- How will this work in mixed systems, with old hosts using cgroups v1 and newer hosts using cgroups v2

BryanMLima:
> Thanks @BryanMLima , for picking this up and for the extensive explanation. I have two questions:
> 
>     * where you say `6 cores at 2 GHz, the shares value will be 12000 ` I would say `huh` as it would either be 12 or 12.000.000.000. What is the factorial used there? (I probably just need a link to a definition)
> 
>     * I see no considerations on live systems, i.e. upgrades. Please, expand on that. Will it have any consequence, or will it be seamless?
> 
> 
> regards,

@DaanHoogland, regarding the first question, ACS calculates the shares by multiplying the frequency by the number of cores, both specified in the compute offering; this is done in method `LibvirtComputingResource#createCpuTuneDef`. Therefore, 6 cores * 2000 MHz (2 Ghz) results in 12,000 shares. This is the current behaviour of ACS, and the PR does not change it for cgroups v1; this PR only change the way ACS calculates the shares for hosts that use cgroups v2. OpenStack[^openstack] solved this same problem by not setting the shares value of all VMs, allowing advanced users to set willingly. TBH, I agree with this approach, setting the shares value like ACS does is misleading, as a more experienced user may question how ACS limits the CPU frequency of VMs. AFAIK, no hypervisor does this, the frequency of the VM will display the host's CPU frequency; hypervisors will only limit the CPU access time (and burst limits) of a VM to “simulate” the specified frequency.

About the second question, I think I did not understand it fully; could you add more details? By live systems, do you mean hosts or VMs?
I am assuming you mean upgrading a host from cgroupv1 to cgroupv2 (or even downgrading from cgroupv1 to cgroupv2). The core of this strategy happens in the `LibvirtComputingResource#initialize()` which is called when the `cloudstack-agent` service is (re)started. It is required to reboot the system when changing the version of cgroup, thus, when the `cloudstack-agent` service starts, ACS will check the version of the cgroup (`stat -fc %T /sys/fs/cgroup/`) and it will set its maximum CPU shares capacity. With this, ACS will always have the updated version of the cgroup utilized by the host.

> Oh, and number 3
> 
>     * How will this work in mixed systems, with old hosts using cgroups v1 and newer hosts using cgroups v2

This PR already address the migration of VMs between hosts with different versions, as the shares value is calculated in this process considering the VM's host destination. Thus, two VMs with the exact same compute offering will have different shares values for cgroups v1 and v2. The shares value is only a proportional weighted; as long as all VMs in the same hosts are in the same scale, the CPU time will be distributed accordingly. If the shares value is not set in the domain XML for libvirt (this never happens in ACS, it is always set), it will use the OS default value, which, for cgroupv2, is 100[^cgroup]; thus, the default behaviour for processes in the same cgroup is to have proportional CPU access time.

This PR, however, does not address updating the shares of VMs on hosts with cgroupv2 that are already running, requiring restarting, migrating or scaling the VM.

[^cgroup]: https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#weights
[^openstack]: https://review.opendev.org/c/openstack/nova/+/824048?tab=comments

DaanHoogland:
thanks @BryanMLima good work.

weizhouapache (approved):
code lgtm

not tested yet

shwstppr:
@blueorangutan package

DaanHoogland:
@blueorangutan test

rohityadavcloud:
@blueorangutan package

mlsorensen:
This makes sense from an implementation perspective. Have not dug into the code in great detail.

mlsorensen (approved):
The shares are all relative, so as long as we calculate some proportional scale that is compatible with the KVM host, it should work.

shwstppr (approved):
Code LGTM, will need testing.

@BryanMLima @DaanHoogland @rohityadavcloud @weizhouapache should this be considered for 4.19? I think we should

DaanHoogland:
> Code LGTM, will need testing.
> 
> @BryanMLima @DaanHoogland @rohityadavcloud @weizhouapache should this be considered for 4.19? I think we should

agreed, or even for the 4.18 branch

DaanHoogland:
> Rebasing on 4.18 depends on @BryanMLima or we can backport later

preferably not, @BryanMLima can you comment?

BryanMLima:
> > Rebasing on 4.18 depends on @BryanMLima or we can backport later
> 
> preferably not, @BryanMLima can you comment?

I will rebase onto 4.18, better to release this patch as soon as possible.

I think it is important to be very explicit in the release notes that this change requires attention to hosts with the cgroupv2 with running VMs. The `shares` value for running VMs will not be updated to the new equation until being restarted, migrated or scaled. Otherwise, VMs that were running in the hosts will be in another scale.

vladimirpetrov:
@blueorangutan package

DaanHoogland:
@BryanMLima canyou check https://github.com/apache/cloudstack/actions/runs/7124755734/job/19399504808?pr=8252#step:7:15647 ?

DaanHoogland:

BryanMLima:
@blueorangutan package

shwstppr:
@blueorangutan test

vladimirpetrov (approved):
LGTM based on manual testing against 4.19 and 4.18.

Testing with Ubuntu 22 environment, host with 6 CPUs x 2000 MHz, the caluclated shares is 9524 (overprovisioning factor 1). Without the fix it's 12000 and deployment fails.

Just one clarification - the actual formula for calculating shares is

shares = NUMBER_CPUS * MIN_SPEED_OR_SPEED (SPEED is a legacy parameter for compatibility with ACS 4.0/4.1)

when MIN_SPEED is calculated like this:

int minspeed = (int)(offering.getSpeed() / (divideCpuByOverprovisioning ? vmProfile.getCpuOvercommitRatio() : 1));

So the overprovisioning factor is also used in the equation.

shwstppr:
Let's merge this once integration test results are in

shwstppr:
Test results from backend,
```
SF] Trillian test result (tid-8550)
Environment: kvm-centos7 (x2), Advanced Networking with Mgmt server 7
Total time taken: 44744 seconds
Marvin logs: [https://github.com/blueorangutan/acs-prs/releases/download/trillian/pr8252-t8550-kvm-centos7.zip
Smoke](https://github.com/blueorangutan/acs-prs/releases/download/trillian/pr8252-t8550-kvm-centos7.zip
Smoke) tests completed. 108 look OK, 1 have errors, 0 did not run
Only failed and skipped tests results shown below:


Test | Result | Time (s) | Test File
--- | --- | --- | ---
test_07_deploy_kubernetes_ha_cluster | `Failure` | 3618.04 | test_kubernetes_clusters.py
test_08_upgrade_kubernetes_ha_cluster | `Failure` | 0.04 | test_kubernetes_clusters.py
test_09_delete_kubernetes_ha_cluster | `Failure` | 0.04 | test_kubernetes_clusters.py
test_10_vpc_tier_kubernetes_cluster | `Failure` | 50.89 | test_kubernetes_clusters.py
ContextSuite context=TestKubernetesCluster>:teardown | `Error` | 117.53 | test_kubernetes_clusters.py

```

PR #8253, opened 2023-11-20
Removed redundant input fields from ACL List rules modal

OlegChuev:
### Description

This PR removes redundant input fields from the front end (FE) on the 'Add ACL Rules' modal window.

According to `server/src/main/java/com/cloud/network/vpc/NetworkACLServiceImpl.java`, line 588:
> ICMP code and ICMP type can't be passed in for any protocol other than ICMP.

In the current state, the input field exists but isn't functional.

### Types of changes

- [ ] build/CI

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

Current state:
Screenshot 2023-11-20 at 21 16 50

After fix:
image

### DIFF
--- a/ui/src/views/network/AclListRulesTab.vue
+++ b/ui/src/views/network/AclListRulesTab.vue
@@ -212,7 +212,7 @@
           <a-input v-model:value="form.protocolnumber" />
         </a-form-item>
 
-        <div v-if="['icmp', 'protocolnumber'].includes(form.protocol)">
+        <div v-if="form.protocol === 'icmp'">
           <a-form-item :label="$t('label.icmptype')" ref="icmptype" name="icmptype">
             <a-input v-model:value="form.icmptype" :placeholder="$t('icmp.type.desc')" />
           </a-form-item>


OlegChuev:
Maybe it's also worth mentioning somewhere in the ACL Rules API documentation?

`api/src/main/java/org/apache/cloudstack/api/command/user/network/UpdateNetworkACLItemCmd.java`

DaanHoogland:
@blueorangutan ui

DaanHoogland:
@OlegChuev , thanks for the fix. I am not sure if it covers all combinations this way, but I'll have a test/look.

DaanHoogland:
@OlegChuev , it works but...
image
shouldn't in the above case the icmp type and code be shown instead of the start and end port?
I think the showing of the icmp fields in case of `Protocol number` was on purpose.

OlegChuev:
@DaanHoogland yep, I think it makes sense to make this field visible if the user enters Protocol Number equal to 1 (ICMP) 

however, personally, I find it a bit weird to enter the protocol number directly instead of choosing the already predefined ICMP, don't you think so? :p

DaanHoogland:
> @DaanHoogland yep, I think it makes sense to make this field visible if the user enters Protocol Number equal to 1 (ICMP)
> 
> however, personally, I find it a bit weird to enter the protocol number directly instead of choosing the already predefined ICMP, don't you think so? :p

maybe, but there are 255 protocol numbers possible and I have no idea if all of those use port ranges or maybe some of them use service/qos or some other combination. I don't think we can cater for everything. Also I do not know if your change is affending someone's intended use. Given those remarks, it is fine by me!

DaanHoogland (approved):

DaanHoogland:
@alexandremattioli you want to have a look?

shwstppr (commented):
UI code change looks okay but I'm not certain about the use-case.
@alexandremattioli @andrijapanicsb @NuxRo @rajujith can comment better

andrijapanicsb:
Since we have "ICMP" as a dedicated menu item (next to Protocol Number) - I think this PR makes sense - hide other ICMP fields if "ICMP" is not specifically the chosen protocol.

weizhouapache:
> Since we have "ICMP" as a dedicated menu item (next to Protocol Number) - I think this PR makes sense - hide other ICMP fields if "ICMP" is not specifically the chosen protocol.

+1
ideally these 2 fields should be visible when protocol number is 1 (ICMP) or 58 (IPv6-ICMP),  it requires some other changes.

OlegChuev:
> ideally these 2 fields should be visible when protocol number is 1 (ICMP) or 58 (IPv6-ICMP), it requires some other changes.

Just a quick question here: since we are making these two fields visible, does it make sense to change the backend API logic as well? Currently, even if we set the Protocol Number to `1` or `58`, `ICMP type` and `ICMP code` will raise an error because of the following code:

```
boolean isIcmpProtocol = protocol.equalsIgnoreCase(NetUtils.ICMP_PROTO);
if (!isIcmpProtocol && (icmpCode != null || icmpType != null)) {
    throw new InvalidParameterValueException("Can specify icmpCode and icmpType for ICMP protocol only");
}
```

`server/src/main/java/com/cloud/network/vpc/NetworkACLServiceImpl.java`

andrijapanicsb:
> > ideally these 2 fields should be visible when protocol number is 1 (ICMP) or 58 (IPv6-ICMP), it requires some other changes.
> 
> Just a quick question here: since we are making these two fields visible, does it make sense to change the backend API logic as well? Currently, even if we set the Protocol Number to `1` or `58`, `ICMP type` and `ICMP code` will raise an error because of the following code:
> 
> ```
> boolean isIcmpProtocol = protocol.equalsIgnoreCase(NetUtils.ICMP_PROTO);
> if (!isIcmpProtocol && (icmpCode != null || icmpType != null)) {
>     throw new InvalidParameterValueException("Can specify icmpCode and icmpType for ICMP protocol only");
> }
> ```
> 
> `server/src/main/java/com/cloud/network/vpc/NetworkACLServiceImpl.java`

If you can do that easily - sure, and with minimal effort/risk - otherwise the API error message is very descriptive and will help the operator learn his IP skills :)

weizhouapache (approved):
let's merge this. cc @DaanHoogland @OlegChuev @andrijapanicsb @shwstppr 

I will take the rest (API, service, etc)
- list protocol numbers in the dropdown
- display icmp type/code if protocol number is 1 (ICMP) or protocol is ICMP, but start/end ports will be hidden then
- list icmp types/codes in the dropdown
- check icmp type/code if protocol number is 1 or protocol is ICMP

PR #8287, opened 2023-11-30
UI: Project column in Default View

hsato03:
### Description

The Default View has a `projects` toggle button that allows users to see which resources belong to projects, but there is nothing to indicate which project they belong to.

For this reason, a `project` column was added if the `projects` button is enabled, indicating the name of the project to which the resource belongs.

### Types of changes

- [ ] build/CI

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

I created the resources that can be listed outside of a project (instance, guest network, VPC, etc.). Then, I enabled the `projects` button in the Default View and verified that the `project` column was filled with the name of the project that the resources belong to.

### DIFF
--- a/ui/src/config/section/compute.js
+++ b/ui/src/config/section/compute.js
@@ -61,17 +61,17 @@ export default {
         if (store.getters.metrics) {
           fields.push(...metricsFields)
         }
-        if (store.getters.listAllProjects) {
-          fields.push('project')
-        }
         if (store.getters.userInfo.roletype === 'Admin') {
           fields.splice(2, 0, 'instancename')
-          fields.push('account')
           fields.push('hostname')
+          fields.push('account')
         } else if (store.getters.userInfo.roletype === 'DomainAdmin') {
           fields.push('account')
         } else {
           fields.push('serviceofferingname')
+        }
+        if (store.getters.listAllProjects) {
+          fields.push('project')
         }
         fields.push('zonename')
         return fields
@@ -465,12 +465,14 @@ export default {
       resourceType: 'VMSnapshot',
       columns: () => {
         const fields = ['displayname', 'state', 'name', 'type', 'current', 'parentName', 'created']
-        if (store.getters.listAllProjects) {
-          fields.push('project')
-        }
         if (['Admin', 'DomainAdmin'].includes(store.getters.userInfo.roletype)) {
           fields.push('account')
+          if (store.getters.listAllProjects) {
+            fields.push('project')
+          }
           fields.push('domain')
+        } else if (store.getters.listAllProjects) {
+          fields.push('project')
         }
         return fields
       },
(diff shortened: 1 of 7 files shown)


ui/src/config/section/compute.js
```diff
@@ -65,9 +65,15 @@ export default {
         if (store.getters.userInfo.roletype === 'Admin') {
           fields.splice(2, 0, 'instancename')
           fields.push('account')
+          if (store.getters.listAllProjects) {
```
weizhouapache, line 68: @hsato03 
project is available for all types of users (root, domain admin  and users). 
the code block can be moved to line 64
```
          if (store.getters.listAllProjects) {
            fields.push('project')
          }
```

weizhouapache (commented):
in my opinion, if project is displayed, account field can be hidden (or not set)

ui/src/config/section/compute.js
hsato03: Account are also available for all types of users, but the `account` column is only displayed for root and domain admins. I guess it would be better to keep the current behavior since the project is an account type.

hsato03:
> in my opinion, if project is displayed, account field can be hidden (or not set)

When the project is displayed, both account and project resources are listed (except in `listNetworks` API). I don't think hiding the account column makes sense in this case.

shwstppr:
@blueorangutan ui

ui/src/config/section/compute.js
weizhouapache: @hsato03 
I was not talking about the `account`, but `project`
`project` can be displayed  not only for root and domain admins, but also regular users

hsato03: @weizhouapache I think you are right, as regular users can create a `project` but not an `account` by default. I will work to change this behavior.

DaanHoogland:
@blueorangutan ui

shwstppr (commented):
@hsato03 I check the UI after changes. Currently, the `Project` column is added to the list of columns but it is not displayed by default and it needs to be checked. Would it make sense to make sure it is visible in the view on toggle?

weizhouapache:
checked on QA server: https://qa.cloudstack.cloud/client/pr/8287/#/vm

it works in some pages: instance snapshots, CKS, ssh key pairs, affinity groups

it does not work on: Instances, AutoScale instance group, Userdata
on "[Instance groups](https://qa.cloudstack.cloud/client/pr/8287/#/vmgroup)", "Project" column is right after "Account", which is different as other pages.

hsato03:
> @hsato03 I check the UI after changes. Currently, the `Project` column is added to the list of columns but it is not displayed by default and it needs to be checked. Would it make sense to make sure it is visible in the view on toggle?

Yes, thank's for your suggestion. I made the changes to address that.

> checked on QA server: https://qa.cloudstack.cloud/client/pr/8287/#/vm
> 
> it works in some pages: instance snapshots, CKS, ssh key pairs, affinity groups
> 
> it does not work on: Instances, AutoScale instance group, Userdata on "[Instance groups](https://qa.cloudstack.cloud/client/pr/8287/#/vmgroup)", "Project" column is right after "Account", which is different as other pages.

That is weird, I also checked at https://qa.cloudstack.cloud/client/pr/8287/#/vm and it is working on Instances and on AutoScale Instance Group. The `listUserdata` API does not have a `project` attribute, it returns both account and project in the Account field.

I normalized the order that Project and Account are displayed.

shwstppr:
@blueorangutan ui

shwstppr (approved):
Tested on QA. LGTM

shwstppr:
@weizhouapache if you approve this, we may consider it for 4.19

weizhouapache (approved):
code lgtm

I created an issue for userdata: https://github.com/apache/cloudstack/issues/8363

PR #8337, opened 2023-12-08
Allow locking and unlocking users via UI

winterhazel:
### Description

Admins are able to lock and unlock users through the APIs `lockUser` and `enableUser`; however, this feature is not currently available in the UI. To address this issue, this PR adds a button to the page of `Enabled` users which allows admins to lock them through the UI. Furthermore, when a user is `Locked`, the "Enable User" button is now visible for admins.

This PR also makes some changes in `AutogenView.vue` to allow the replacement of placeholders in `message`s and `successMessage`s of confirmation dialogs.

### Types of changes

- [ ] Major
- [X] Minor

image
image

### How Has This Been Tested?

1. In the root admin account, I accessed the page of an `Enabled` user and verified that the "Lock User" button was available;
2. I clicked the "Lock User" button and verified that the user's state changed to `Locked`;
3. I verified that the "Enable User" button was now visible.
4. I clicked the "Enable User" button and verified that the user's state changed to `Enabled`;
5. In an account that had the role `User`, I accessed the user's own page and verified that the "Lock User" button was not visible.

Regarding the replacement of placeholders in `message`s and `successMessage`s of confirmation dialogs:
- I verified that the warning (`message`) and success messages (`successMessage`) showed the user's name when locking a user;
- I verified that existing dialogs did not change when the `message` field consisted of only a label;
- I verified that existing dialogs did not change when the `message` field consisted of a function that returns a label;
- I verified that existing dialogs did not change when the `successMessage` field consisted of only a label.


shwstppr:
@blueorangutan ui

DaanHoogland (approved):
clgtm, tested in qa

ui/src/config/section/user.js
```diff
@@ -96,6 +96,19 @@ export default {
...
+      icon: 'LockOutlined',
+      label: 'label.action.lock.user',
+      dataView: true,
+      popup: true,
+      show: (record, store) => {
+        return ['Admin', 'DomainAdmin'].includes(store.userInfo.roletype) && !record.isdefault &&
+          !(record.domain === 'ROOT' && record.account === 'admin' && record.accounttype === 1) &&
+          record.state === 'enabled'
+      },
+      component: shallowRef(defineAsyncComponent(() => import('@/views/iam/LockUser.vue')))
```
shwstppr, line 110: @winterhazel I'm not sure but isn't there a mechanism already to show a confirm dialog with custom text? I guess that's all this new vue component is doing?

winterhazel: There is. I created this component because I thought it would be nice to include the user's name in the warning and success messages so that operators do not accidentally lock the wrong user and so that they get a confirmation showing which user was locked, which was not possible with the existing mechanism since it does not provide the option to replace placeholders.

JoaoJandre (approved):
LGTM, tested in a local lab

ui/src/config/section/user.js
DaanHoogland: That makes sense @winterhazel , but on second though, should we maybe extend the existing component instead? just a though, no :-1:

winterhazel: @DaanHoogland this seems like the optimal solution, as it would allow easily doing the same thing in other parts of the UI. Since this feature would prove useful to this PR, I will try implementing it here.

winterhazel: @DaanHoogland @shwstppr I have made some changes in this PR to allow the replacement of placeholders in `message`s and `successMessage`s of confirmation dialogs, and updated the description. Could you guys review again? Thanks.

DaanHoogland:
code looks good @winterhazel .
@JoaoJandre , did you test the new functionality exclusively or poke around for regressions as well?

JoaoJandre:
> code looks good @winterhazel . @JoaoJandre , did you test the new functionality exclusively or poke around for regressions as well?

@DaanHoogland sorry for not explaining my tests. 
1. Create a new account through the UI and then lock it through the UI;
2. Try to log in with it, the UI displays a message saying that the account was locked;
3. Unlock the account through the UI;
4. Log in.

DaanHoogland:
@blueorangutan ui

DaanHoogland:
I created and deleted a user in the ui and the notification message appeared:
image

I am actually not sure if this is enough. Could you tell @shwstppr ?

shwstppr (approved):
Code LGTM

rohityadavcloud:
@blueorangutan ui

DaanHoogland:
works in qa, no backend code so the GHA failure can't be related.

Issue #8266, opened 2023-11-26
Vmstats records are not deleted as expected


havengit:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI
~~~

##### CLOUDSTACK VERSION

~~~
ACS 4.18.0.0
kvm

2 management servers with galera mariadb cluster
~~~

##### SUMMARY

When accessing the instance list, it takes about 3 minutes for a return, but other pages request return soon. The database has a lot of read IO. Check that the database is executing Creating sort index.

I find the vm_stats table has 1270000 records,From 4.17 the vm_stat is retained persistently. Setting in globe  "Vm stats max retention time " is 720 min default ,but there some record was not deleted 。
https://github.com/apache/cloudstack/pull/5984
![image](https://github.com/apache/cloudstack/assets/20832243/4759d7a5-cb61-4e6b-8822-54cbc29bc689)

A temporary solution is to truncate the vm_stats table.

##### STEPS TO REPRODUCE

~~~
No steps required to reproduce
~~~

##### EXPECTED RESULTS

~~~
Records are cleared normally, or collection of vmstat is allowed to be turned off
~~~

##### ACTUAL RESULTS

~~~
Some records was not deleted
~~~

havengit:
After some testing, this problem has been solved. This shouldn't be a bug. The global parameters need some optimization.
default allocation:
Vm stats interval 60s
Vm stats max retention time 720min
Modified configuration
Vm stats interval 300s
Vm stats max retention time 60min

There are about 200+ running VMs in our environment. If calculated according to the default parameters, it will be retained
200*720=144000, after optimization: 200*60=12000

It may be that our database performance is insufficient and the amount of records is too large to complete the deletion operation as expected. There will be more and more records.

It is recommended to optimize the setting value.

DaanHoogland:
@havengit Do you mean to close this ticket or is still work to be done?

weizhouapache:
> It may be that our database performance is insufficient and the amount of records is too large to complete the deletion operation as expected. There will be more and more records.

if so, this needs to be fixed

havengit:
> @havengit Do you mean to close this ticket or is still work to be done?

This problem will occur when there are a large number of running VMs. I've confirmed that adjusting the global settings resolves the issue. No more work needed. It is recommended to modify the default value。

weizhouapache:
persistent vm stats were introduced by PR #5984 in CloudStack 4.17.0.0
[...]

PR #6803, opened 2022-10-04
ui,server,api: resource metrics improvements

shwstppr:
### Description
Doc PR: https://github.com/apache/cloudstack-documentation/pull/294

This PR improves resource metrics/stats visualization functionality.
It allows stats for:
**System VMs**
Persisting and showing stats for system VMs. New API - `listSystemVmsUsageHistory` has been added to support this. SSVM, CPVM and VRs now show `Metrics` tab similar to user VMs. System VM stats will be stored in the existing DB table `cloud.vm_stats` and will use the same cleanup mechanism as the user VMs. To control storing system VM stats a new global setting has been added,
`vm.stats.user.vm.only` - When set to 'false' stats for system VMs will be collected otherwise stats collection will be done only for user VMs. The default value is 'false'

**VM Disks**
Persisting and showing stats for VM disks. New API - `listVolumesUsageHistory` has been added to support this. Volumes attached to a running user VM show the `Metrics` tab. VM disks stats will be stored in a new DB table `cloud.volume_stats`. To control storing VM disk stats two new global setting have been added,
`vm.disk.stats.retention.enabled` - When set to 'true' stats for VM disks will be stored in the database otherwise disk stats will not be stored. The default value is 'false'.
`vm.disk.stats.max.retention.time` - The maximum time (in minutes) for keeping Volume stats records in the database. The VM stats cleanup process will be disabled if this is set to 0 or less than 0. The default value is 720.

Improvements have been made to the Metrics tab wrt to time filtering and display. Order for statistics for an instance is now - CPU, memory, disks, network.

Also, the default value for the global setting - `vm.stats.max.retention.time` has been increased to `720` or 12 hours for more meaningful representation in the UI.
Based on tests VM stats DB storage will be in the order of,
Screenshot from 2022-10-04 14-18-27
Therefore, with a default VM stat interval of 1min, 12 hours stats may make sense.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### Screenshots (if appropriate):
[...]

PR #8350, opened 2023-12-13
Reduce the number of vmstats data saved and improve the response speed of the instance list page

havengit:
### Description

This PR...

### Types of changes

- [ ] build/CI

### Feature/Enhancement Scale or Bug Severity

improve the following issues
https://github.com/apache/cloudstack/issues/8266

For vmstats data, the default global settings will cause a large amount of data to be stored in the database. When accessing the instance page, vmstats data will be requested, causing the page to open extremely slowly. Especially when multiple users access the instance page at the same time, the problem will be exacerbated.
My environment has about 200 running instances.

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

Because it is a modification of global settings, the test method is to directly modify the global parameters. The corresponding code modifications are for new deployments. There are no tests for newly deployed scenarios.
test environment:
ACS 4.18.0.0 + Galera mariadb server CLuster 10.6
Ubuntu 20.04+KVM

### DIFF
--- a/engine/schema/src/main/resources/META-INF/db/schema-222to224.sql
+++ b/engine/schema/src/main/resources/META-INF/db/schema-222to224.sql
@@ -133,7 +133,7 @@ INSERT IGNORE INTO `cloud`.`configuration` VALUES
 ('Advanced','DEFAULT','management-server','vm.op.cleanup.wait','3600','Time (in seconds) to wait before cleanuping up any vm work items'),
 ('Advanced','DEFAULT','management-server','vm.op.lock.state.retry','5','Times to retry locking the state of a VM for operations'),
 ('Advanced','DEFAULT','management-server','vm.op.wait.interval','120','Time (in seconds) to wait before checking if a previous operation has succeeded'),
-('Advanced','DEFAULT','management-server','vm.stats.interval','60000','The interval (in milliseconds) when vm stats are retrieved from agents.'),
+('Advanced','DEFAULT','management-server','vm.stats.interval','300000','The interval (in milliseconds) when vm stats are retrieved from agents.'),
 ('Advanced','DEFAULT','management-server','vm.tranisition.wait.interval','3600','Time (in seconds) to wait before taking over a VM in transition state'),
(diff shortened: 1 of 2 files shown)


DaanHoogland:
Is this still a draft PR, @havengit ?

havengit:
> Is this still a draft PR, @havengit ?

This is a complete PR with very few changes.

BryanMLima:
@DaanHoogland @havengit, the default global configuration values should only be used as a guideline, and it will not work in every environment.

This feature was introduced in PR #5984 with the configuration `vm.stats.max.retention.time` with value of `1` (to maintain backward compatibility), which would not cause any problems, performance-wise. However, the default value was changed to `720` with PR #6803, which seems to be too high for larger environments. I do not think that changing the default value again is going to help anybody. Documentation regarding the combination of the parameters `vm.stats.max.retention.time` and `vm.stats.interval` would benefit everyone. IMHO, the default value `vm.stats.max.retention.time` should not have been changed since the introduction of the feature, as the operator must know its underlying systems to set its value.

weizhouapache:
> @DaanHoogland @havengit, the default global configuration values should only be used as a guideline, and it will not work in every environment.
> 
> This feature was introduced in PR #5984 with the configuration `vm.stats.max.retention.time` with value of `1` (to maintain backward compatibility), which would not cause any problems, performance-wise. However, the default value was changed to `720` with PR #6803, which seems to be too high for larger environments. I do not think that changing the default value again is going to help anybody. Documentation regarding the combination of the parameters `vm.stats.max.retention.time` and `vm.stats.interval` would benefit everyone. IMHO, the default value `vm.stats.max.retention.time` should not have been changed since the introduction of the feature, as the operator must know its underlying systems to set its value.

+1 with documentation changes. 

We should not change the default value of existing configurations,  unless the value is extremely unreasonable.

havengit:
That makes sense, I'll close this PR

PR #6567, opened 2022-07-25
Password policies

JoaoJandre:
## Description

Currently, ACS does not have a feature to manage user password policies. New domain-scoped settings were created to allow operators to define password policies. Configured password policies will be applied whenever a user changes their password or a new user is created. The following settings were created:

| Setting | Purpose |
| ------ | ------ |
|`password.policy.minimum.length`| Minimum password length |
|`password.policy.minimum.lowercase.letters`| Minimum number of lower case letters|
|`password.policy.minimum.uppercase.letters`| Minimum number of upper case letters|
|`password.policy.minimum.special.characters`| Minimum number of special characters|
|`password.policy.minimum.digits`| Minimum number of digits|
|`password.policy.allowPasswordToContainUsername`| If the password may contain the user's username|
|`password.policy.regex`| A regular expression that the password must match|

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [X] Minor

### How Has This Been Tested?

Unit tests were created for the new methods. Furthermore each configuration was individually tested in the ROOT domain. The creation of a user was tested, to ensure that verification of password policies was being called. All other tests were done by changing the user's password via the UI.

| No. | Test | Result | Expected result? |
| ------ | ------ | ------ | ------ |
| 1 | `password.policy.minimum.length` = 0 and password "abc" | Password successfully changed | Yes |
| 2 | `password.policy.minimum.length` = 5 and password "abc" | Error thrown warning that the length does not reach the minimum expected | Yes |
| 3 | `password.policy.minimum.length` = 5 and password "abcde" | Password successfully changed | Yes |
| 4 | `password.policy.minimum.lowercase.letters` = 0 and password "ABC" | Password successfully changed | Yes |
| 5 | `password.policy.minimum.lowercase.letters` = 1 and password "ABC" | Error thrown warning that the number of lowercase letters does not reach the expected minimum | Yes |
| 6 | `password.policy.minimum.lowercase.letters` = 1 and password "aBC" | Password successfully changed | Yes |
| 7 | `password.policy.minimum.uppercase.letters` = 0 and password "abc" | Password successfully changed | Yes |
[...]

PR #8370, opened 2023-12-18
server: skip password policies check on empty password

JoaoJandre:
### Description

This PR changes the `password.policy.regex` default value to empty. With an empty value for the configuration, it is skipped during the password policy check, only when the configuration is set to something different than a blank string, the regex will get checked.  
This way, when creating a user on `org.apache.cloudstack.ldap.LdapAuthenticator#authenticate()`  we won't get an error by default, as a empty value for the password is passed.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/user/PasswordPolicy.java
+++ b/server/src/main/java/com/cloud/user/PasswordPolicy.java
@@ -78,8 +78,8 @@ public interface PasswordPolicy {
             "Advanced",
             String.class,
             "password.policy.regex",
-            ".+",
-            "A regular expression that the user's password must match. The default expression '.+' will match with any password.",
+            "",
+            "A regular expression that the user's password must match. By default no expression is used.",
             true,
             ConfigKey.Scope.Domain);
 
--- a/server/src/main/java/com/cloud/user/PasswordPolicy.java
+++ b/server/src/main/java/com/cloud/user/PasswordPolicy.java
@@ -78,8 +78,8 @@ public interface PasswordPolicy {
             "Advanced",
             String.class,
             "password.policy.regex",
-            "",
-            "A regular expression that the user's password must match. By default no expression is used.",
+            ".+",
+            "A regular expression that the user's password must match. The default expression '.+' will match with any password.",
             true,
             ConfigKey.Scope.Domain);
 
--- a/server/src/main/java/com/cloud/user/PasswordPolicyImpl.java
+++ b/server/src/main/java/com/cloud/user/PasswordPolicyImpl.java
@@ -188,12 +188,12 @@ public class PasswordPolicyImpl implements PasswordPolicy, Configurable {
         logger.trace(String.format("Validating if the new password for user [%s] matches regex [%s] defined in the configuration [%s].",
                 username, passwordPolicyRegex, PasswordPolicyRegex.key()));
 
-        if (passwordPolicyRegex == null){
+        if (StringUtils.isBlank(passwordPolicyRegex)) {
             logger.trace(String.format("Regex is null; therefore, we will not validate if the new password matches with regex for user [%s].", username));
             return;
         }
 
-        if (!password.matches(passwordPolicyRegex)){
+        if (!password.matches(passwordPolicyRegex)) {
             logger.error(String.format("User [%s] informed a new password that does not match with regex [%s]. Refusing the user's new password.", username, passwordPolicyRegex));
             throw new InvalidParameterValueException("User password does not match with password policy regex.");
         }
--- a/server/src/main/java/com/cloud/user/PasswordPolicyImpl.java
+++ b/server/src/main/java/com/cloud/user/PasswordPolicyImpl.java
@@ -27,6 +27,12 @@ public class PasswordPolicyImpl implements PasswordPolicy, Configurable {
     private Logger logger = Logger.getLogger(PasswordPolicyImpl.class);
 
     public void verifyIfPasswordCompliesWithPasswordPolicies(String password, String username, Long domainId) {
+        if (StringUtils.isEmpty(password)) {
+            logger.warn(String.format("User [%s] has an empty password, skipping password policy checks. " +
+                    "If this is not a LDAP user, there is something wrong.", username));
+            return;
+        }
+
         int numberOfSpecialCharactersInPassword = 0;
         int numberOfUppercaseLettersInPassword = 0;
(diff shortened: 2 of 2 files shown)


JoaoJandre:
I'm not sure if this is significant enough for 4.19, if not, I can target 4.18.2.

shwstppr:
@JoaoJandre would be good to fix in 4.18 branch but it seems it needs a DB change in the upgrade path. Not sure if it is desired at this point to have a 4181to4182 upgrade path

DaanHoogland:
is it wise to allow for empty passwords? I may be misinterpreting this code, but I would not want this in a default install. Or is this only applicable for LDAP logins?

BTW, the update is not necessary if the workaround is mentioned in the release notes. To my knowledge this fix is directed only at installations using LDAP, as mentioned in https://lists.apache.org/thread/tqtcwsfb0knmvqct1hlow4ty2nc2w2j2, is it?

these are questions/discussion points, no comments on the code.

JoaoJandre:
> is it wise to allow for empty passwords? I may be misinterpreting this code, but I would not want this in a default install. Or is this only applicable for LDAP logins?

As the password is a required field when creating users using createAccount and createUser APIs, blank passwords are not allowed. Furthermore, on the updateUser API the password we check if the password is blank as well. The password policies were created on #6567 with the intention of being optional to keep backwards compatibility; therefore, I'd say that it makes sense to not block the operation using the default regex.

> BTW, the update is not necessary if the workaround is mentioned in the release notes. To my knowledge this fix is directed only at installations using LDAP, as mentioned in https://lists.apache.org/thread/tqtcwsfb0knmvqct1hlow4ty2nc2w2j2, is it?

Yes this fix is directed only on installations using LDAP, from the code inspection that I did, the only instance of an empty password is on that method I mentioned on the description, the rest of the code will not allow a password that does not have at least one character.

Anyway, do you think that a note on the release notes is all that is needed on this case @DaanHoogland @weizhouapache @shwstppr ?

weizhouapache:
@JoaoJandre 
can checks be skipped if source is User.Source.LDAP ?

JoaoJandre:
> @JoaoJandre can checks be skipped if source is User.Source.LDAP ?

I'd rather keep the feature agnostic and avoid adding different workflows. If the consensus is that changing the configuration value is not the best solution, I'd rather either document the behavior, or change the call on `org.apache.cloudstack.ldap.LdapAuthenticator#authenticate()` to something that is not a blank string.

DaanHoogland:
> > @JoaoJandre can checks be skipped if source is User.Source.LDAP ?
> 
> I'd rather keep the feature agnostic and avoid adding different workflows. If the consensus is that changing the configuration value is not the best solution, I'd rather either document the behavior, or change the call on `org.apache.cloudstack.ldap.LdapAuthenticator#authenticate()` to something that is not a blank string.

I agree, certainly if the LDAP autentication is the only one that suffers. I do not remember exactly how this works, but LDAP should put its own requirements on passwords.
see my dumb comment there ;)
```
        // TODO not allowing an empty password is a policy we shouldn't decide on. A private cloud may well want to allow this.
        if (StringUtils.isNoneEmpty(username, password)) {
```

server/src/main/java/com/cloud/user/PasswordPolicyImpl.java
```diff
@@ -188,12 +188,12 @@ protected void validateIfPasswordMatchesRegex(String password, String username,
...
                 username, passwordPolicyRegex, PasswordPolicyRegex.key()));
 
-        if (passwordPolicyRegex == null){
+        if (StringUtils.isBlank(passwordPolicyRegex)) {
             logger.trace(String.format("Regex is null; therefore, we will not validate if the new password matches with regex for user [%s].", username));
             return;
         }
 
-        if (!password.matches(passwordPolicyRegex)){
+        if (!password.matches(passwordPolicyRegex)) {
```
sureshanaparti, line 196: Hi @JoaoJandre Does the _password.policy.regex_ setting update checks the existing users password match (at sometime, may be login), and enforces user with non matching password to change the password immediately or after certain grace period?

JoaoJandre: @sureshanaparti All the password policies are applied only on user/account creation or password change. They are not validated on log in.

sureshanaparti: > @sureshanaparti All the password policies are applied only on user/account creation or password change. They are not validated on log in.

ok, thanks for the update @JoaoJandre

server/src/main/java/com/cloud/user/PasswordPolicy.java
```diff
@@ -79,8 +79,8 @@ public interface PasswordPolicy {
             "Advanced",
             String.class,
             "password.policy.regex",
-            ".+",
-            "A regular expression that the user's password must match. The default expression '.+' will match with any password.",
+            "",
+            "A regular expression that the user's password must match. By default no expression is used.",
```
DaanHoogland, line 83: @JoaoJandre , this means that empty passwords are allowed, right? If so, do we think this is ok?

weizhouapache: @DaanHoogland 
password is a required field when create a user, it is not empty when create user via API or UI.
but it is empty when import a ldap user. we need to support both scenarios.

@JoaoJandre 
can we just skip the check if password is empty ?
```
     public void verifyIfPasswordCompliesWithPasswordPolicies(String password, String username, Long domainId) {
+        if (StringUtils.isEmpty(password)) {
+            return; // password is empty when import LDAP user
+        }
         int numberOfSpecialCharactersInPassword = 0;
         int numberOfUppercaseLettersInPassword = 0;
         int numberOfLowercaseLettersInPassword = 0;
```

weizhouapache (approved):
code lgtm

@JoaoJandre 
can you rebase with 4.18 ?

shwstppr:
@blueorangutan package

weizhouapache:
@blueorangutan test

weizhouapache:
testing result look ok

cc  @shwstppr @DaanHoogland @JoaoJandre

shwstppr (approved):
Code LGTM

PR #8377, opened 2023-12-19
remove StorPool tags from detached volumes

slavkap:
### Description

This PR removes the VolumeCare policy and the VM's UUID tags when a VM has attached data disks and during the destruction of the VM those volumes remain active

### Types of changes

- [ ] build/CI

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

manual and smoke tests

### DIFF
--- a/plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/driver/StorPoolPrimaryDataStoreDriver.java
+++ b/plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/driver/StorPoolPrimaryDataStoreDriver.java
@@ -194,9 +194,8 @@ public class StorPoolPrimaryDataStoreDriver implements PrimaryDataStoreDriver {
             final VolumeVO volume = volumeDao.findById(data.getId());
             if (volume.getInstanceId() == null) {
                 StorPoolUtil.spLog("Removing tags from detached volume=%s", volume.toString());
-                Map<String, String> tags = StorPoolHelper.addStorPoolTags(null, "", null, "");
                 SpConnectionDesc conn = StorPoolUtil.getSpConnection(dataStore.getUuid(), dataStore.getId(), storagePoolDetailsDao, primaryStoreDao);
-                StorPoolUtil.volumeRemoveTags(StorPoolStorageAdaptor.getVolumeNameFromPath(volume.getPath(), true), tags, conn);
+                StorPoolUtil.volumeRemoveTags(StorPoolStorageAdaptor.getVolumeNameFromPath(volume.getPath(), true), conn);
             }
         }
 
--- a/plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/driver/StorPoolPrimaryDataStoreDriver.java
+++ b/plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/driver/StorPoolPrimaryDataStoreDriver.java
@@ -190,6 +190,16 @@ public class StorPoolPrimaryDataStoreDriver implements PrimaryDataStoreDriver {
 
     @Override
     public void revokeAccess(DataObject data, Host host, DataStore dataStore) {
+        if (DataObjectType.VOLUME == data.getType()) {
+            final VolumeVO volume = volumeDao.findById(data.getId());
+            if (volume.getInstanceId() == null) {
+                StorPoolUtil.spLog("Removing tags from detached volume=%s", volume.toString());
+                Map<String, String> tags = StorPoolHelper.addStorPoolTags(null, "", null, "");
+                SpConnectionDesc conn = StorPoolUtil.getSpConnection(dataStore.getUuid(), dataStore.getId(), storagePoolDetailsDao, primaryStoreDao);
+                StorPoolUtil.volumeUpadateTags(StorPoolStorageAdaptor.getVolumeNameFromPath(volume.getPath(), true), tags, conn);
+            }
+        }
+
     }
 
     private void updateStoragePool(final long poolId, final long deltaUsedBytes) {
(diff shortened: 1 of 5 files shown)


plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/util/StorPoolUtil.java
```diff
@@ -517,6 +517,12 @@ public static SpApiResponse volumeUpdate(final String name, final Long newSize,
         return POST("MultiCluster/VolumeUpdate/" + name, json, conn);
     }
 
+    public static SpApiResponse volumeUpadateTags(String name, Map<String, String> tags, SpConnectionDesc conn) {
```
sureshanaparti, line 520:
```suggestion
    public static SpApiResponse volumeRemoveTags(String name, SpConnectionDesc conn) {
            Map<String, String> tags = StorPoolHelper.addStorPoolTags(null, "", null, "");
```

seems this method is used to remove tags

plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/driver/StorPoolPrimaryDataStoreDriver.java
```diff
@@ -190,6 +190,16 @@ public boolean grantAccess(DataObject data, Host host, DataStore dataStore) {
 
     @Override
     public void revokeAccess(DataObject data, Host host, DataStore dataStore) {
+        if (DataObjectType.VOLUME == data.getType()) {
+            final VolumeVO volume = volumeDao.findById(data.getId());
+            if (volume.getInstanceId() == null) {
+                StorPoolUtil.spLog("Removing tags from detached volume=%s", volume.toString());
+                Map<String, String> tags = StorPoolHelper.addStorPoolTags(null, "", null, "");
+                SpConnectionDesc conn = StorPoolUtil.getSpConnection(dataStore.getUuid(), dataStore.getId(), storagePoolDetailsDao, primaryStoreDao);
+                StorPoolUtil.volumeUpadateTags(StorPoolStorageAdaptor.getVolumeNameFromPath(volume.getPath(), true), tags, conn);
```
sureshanaparti, line 199: update this if method is updated to _volumeRemoveTags_ as suggested

sureshanaparti, line 199: StorPoolUtil.volumeRemoveTags(StorPoolStorageAdaptor.getVolumeNameFromPath(volume.getPath(), true), conn);


update this if method is updated to _volumeRemoveTags_ as suggested

plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/util/StorPoolUtil.java
```diff
@@ -517,6 +517,12 @@ public static SpApiResponse volumeUpdate(final String name, final Long newSize,
...
+        Map<String, Object> json = new HashMap<>();
+        json.put("tags", tags);
+        return POST("MultiCluster/VolumeUpdate/" + name, json, conn);
+    }
+
     public static SpApiResponse volumeUpdateTags(final String name, final String uuid, Long iops,
```
sureshanaparti, line 526:
```suggestion
    public static SpApiResponse volumeUpdateIopsAndTags(final String name, final String uuid, Long iops,
```

plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/util/StorPoolUtil.java
slavkap: Thanks @sureshanaparti! I prefer it to stay with this name. In future, this method could be used to set tags also

DaanHoogland: in the future `this method could be used to set tags also` but the name is `volumeRemoveTags`, @slavkap ? Remove and set seem in contradiction to me. Can we add/rename whatever we need in the future later?

DaanHoogland: @slavkap `volumeUpadateTags` has a spello.
also
When we need a thing in the future we can implement it then, right?

slavkap: @DaanHoogland, I will rename it as you both suggested. Still, I prefer to leave it with a common name like the used StorPool API call even though now it is used only to remove

DaanHoogland: that seems ok, but the spello is still strange: volumeUpadateTags instead of volumeUpdateTags?

slavkap: I renamed the method with the latest commit as Suresh suggested to volumeRemoveTags

DaanHoogland: :shrug:

plugins/storage/volume/storpool/src/main/java/org/apache/cloudstack/storage/datastore/util/StorPoolUtil.java
```diff
@@ -517,7 +517,13 @@ public static SpApiResponse volumeUpdate(final String name, final Long newSize,
         return POST("MultiCluster/VolumeUpdate/" + name, json, conn);
     }
 
-    public static SpApiResponse volumeUpdateTags(final String name, final String uuid, Long iops,
+    public static SpApiResponse volumeRemoveTags(String name, Map<String, String> tags, SpConnectionDesc conn) {
+        Map<String, Object> json = new HashMap<>();
```
sureshanaparti, line 521: @slavkap Better to create empty tags map here, instead of passing it to indicate removal of tags in the method, otherwise it's like an update tags.

_Map<String, String> tags = StorPoolHelper.addStorPoolTags(null, "", null, "");_

sureshanaparti (approved):
code lgtm

sureshanaparti:
@blueorangutan package

rohityadavcloud (commented):
LGTM

PR #7945, opened 2023-09-06
kvm: fix live vm migration between local storage pools

weizhouapache:
### Description

This PR fixes #7942

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

PR #7408, opened 2023-04-06
Allow volume migrations in ScaleIO within and across ScaleIO storage clusters

harikrishna-patnala:
### Description
This feature allows the storage migration for ScaleIO volumes between the ScaleIO storage pools of the same and different ScaleIO storage clusters. Previously this operation was blocked and now we are allowing it from API/UI.

The volume migration within the same storage cluster is performed using ScaleIO’s native V-Tree migration operation where the volume will be live migrated to another storage pool. The volume migration across the different storage cluster is performed using libvirt’s live blockcopy operation which will first copy the volume to the destination storage pool and after a successful copy, the source volume will be deleted. Upon any failures during the block copy, source volume will remain to continue and other blockcopy operations will be reverted.

A document PR is added here https://github.com/apache/cloudstack-documentation/pull/321

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

Issue #7942, opened 2023-09-05
Local Storage migration fails due to storage pool not found on destination KVM host


rohityadavcloud:
On a 4.18.1.0 RC2 adv zone KVM env using local only storage and local storage for systemvms, migration fails from kvm1 to kvm2 due to the following seen in the target kvm host

```
2023-09-05 21:12:06,645 INFO  [kvm.storage.LibvirtStorageAdaptor] (agentRequest-Handler-2:null) (logid:f38dafe4) Attempting to create volume 23ef733c-c26e-405a-829e-6e0010b3cbeb (Filesystem) in pool 3a11d1ea-eff4-4f5b-9b7c-e0ff89a8c9dd with size (2.20 GB) 2361393152
2023-09-05 21:12:06,691 INFO  [kvm.resource.LibvirtConnection] (agentRequest-Handler-3:null) (logid:f38dafe4) No existing libvirtd connection found. Opening a new one
2023-09-05 21:12:06,692 WARN  [kvm.resource.LibvirtConnection] (agentRequest-Handler-3:null) (logid:f38dafe4) Can not find a connection for Instance i-2-453-VM. Assuming the default connection.
2023-09-05 21:12:08,499 INFO  [kvm.storage.LibvirtStorageAdaptor] (agentRequest-Handler-4:null) (logid:f38dafe4) Trying to fetch storage pool 8aa9768c-cbcf-4e8e-8875-f94a7f9445b6 from libvirt
2023-09-05 21:12:08,500 WARN  [cloud.agent.Agent] (agentRequest-Handler-4:null) (logid:f38dafe4) Caught:
com.cloud.utils.exception.CloudRuntimeException: Could not fetch storage pool 8aa9768c-cbcf-4e8e-8875-f94a7f9445b6 from libvirt due to org.libvirt.LibvirtException: Storage pool not found: no storage pool with matching uuid '8aa9768c-cbcf-4e8e-8875-f94a7f9445b6'
	at com.cloud.hypervisor.kvm.storage.KVMStoragePoolManager.getStoragePool(KVMStoragePoolManager.java:278)
	at com.cloud.hypervisor.kvm.storage.KVMStoragePoolManager.getStoragePool(KVMStoragePoolManager.java:264)
```

The above can sometimes cause the VM to crash as it fails right at the last step on the dest. kvm host end, and CloudStack may further report that domain is not running.

However, systemvm migration with storage work. I'm not sure if this is an env issue or a bug.

[...]

PR #8437, opened 2024-01-03
Fix libvirt domain event listener by properly processing events

mlsorensen:
### Description

This PR fixes the libvirt event processing to handle changes in Libvirt domains (and other events if registered).

There is some history around this. Initially it seems there was an event loop initialization added in an earlier [PR](https://github.com/apache/cloudstack/pull/7408) that supports live block copy. This introduced `Library.initEventLoop();` to `LibvirtConnection`, however it never polled for events, and this caused the keepalive issue warned about by libvirt doc [here](https://libvirt.org/html/libvirt-libvirt-event.html#virEventRegisterDefaultImpl):

```
For proper event handling, it is important that the event implementation is registered before a connection to the Hypervisor is opened.

Once registered, the application has to invoke [virEventRunDefaultImpl](https://libvirt.org/html/libvirt-libvirt-event.html#virEventRunDefaultImpl)() in a loop to process events. Failure to do so may result in connections being closed unexpectedly as a result of keepalive timeout.
```

Thus it was found that long running VM migrations would fail due to keep alive issues, which resulted in https://github.com/apache/cloudstack/pull/7945

In the meantime, [support for detecting out of band VM shutdowns](https://github.com/apache/cloudstack/pull/7963) was being developed to also utilize events, and in parallel was relying on the above `Library.initEventLoop();` in `LibvirtConnection`.  This new feature added the missing polling, but ultimately doesn't work (fails gracefully to provide new event handling) because `Library.initEventLoop()` was lost in parallel.

This PR moves the thread responsible for running the Libvirt event loop out of LibvirtComputingResource and into LibvirtConnection, and reintroduces `Library.initEventLoop()`, which shouldn't result in keep alive errors now that we are properly polling for events. It also sets up the event loop before any Libvirt connection as documentation requires.

Finally, when a connection is tested using `conn.getVersion()`, if an exception is thrown we simply create a new connection (but no attempt is made to clean up the old connection). This was resulting in file handle leaks in cases such as the keep alive timeout that was causing connections to go stale, so now we try to close the connection before creating a new one.

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested with main by deleting VMs out of band, also checking for file handle leaks that were apparent in the previous observed introduction of `Library.initEventLoop()` using `lsof +E -aUc java` on the hypervisor hosts.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
@@ -114,8 +114,12 @@ public class LibvirtConnection {
         return "qemu:///system";
     }
 
-    // stand up libvirt event handling and polling. This is not specific to a connection object instance, but needs to
-    // exist prior to creating connections.
+    /**
+     * Set up Libvirt event handling and polling. This is not specific to a connection object instance, but needs
+     * to be done prior to creating connections. See the Libvirt documentation for virEventRegisterDefaultImpl and
+     * virEventRunDefaultImpl or the libvirt-java Library Javadoc for more information.
+     * @throws LibvirtException
+     */
     private static synchronized void setupEventListener() throws LibvirtException {
         if (libvirtEventThread == null || !libvirtEventThread.isAlive()) {
             // Registers a default event loop, must be called before connecting to hypervisor
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
@@ -23,6 +23,7 @@ import com.cloud.agent.properties.AgentProperties;
 import com.cloud.agent.properties.AgentPropertiesFileHandler;
 import org.apache.log4j.Logger;
 import org.libvirt.Connect;
+import org.libvirt.Library;
 import org.libvirt.LibvirtException;
 
 import com.cloud.hypervisor.Hypervisor;
@@ -34,6 +35,7 @@ public class LibvirtConnection {
 
     static private Connect s_connection;
     static private String s_hypervisorURI;
+    static private Thread libvirtEventThread;
 
     static public Connect getConnection() throws LibvirtException {
         return getConnection(s_hypervisorURI);
@@ -45,6 +47,8 @@ public class LibvirtConnection {
 
         if (conn == null) {
             s_logger.info("No existing libvirtd connection found. Opening a new one");
+
+            setupEventListener();
             conn = new Connect(hypervisorURI, false);
             s_logger.debug("Successfully connected to libvirt at: " + hypervisorURI);
             s_connections.put(hypervisorURI, conn);
@@ -53,7 +57,15 @@ public class LibvirtConnection {
                 conn.getVersion();
             } catch (LibvirtException e) {
                 s_logger.error("Connection with libvirtd is broken: " + e.getMessage());
+
+                try {
+                    conn.close();
+                } catch (LibvirtException closeEx) {
+                    s_logger.debug("Ignoring error while trying to close broken connection:" + closeEx.getMessage());
+                }
+
                 s_logger.debug("Opening a new libvirtd connection to: " + hypervisorURI);
+                setupEventListener();
                 conn = new Connect(hypervisorURI, false);
                 s_connections.put(hypervisorURI, conn);
             }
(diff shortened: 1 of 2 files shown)


plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
```diff
@@ -101,4 +113,29 @@ static String getHypervisorURI(String hypervisorType) {
 
         return "qemu:///system";
     }
+
+    // stand up libvirt event handling and polling. This is not specific to a connection object instance, but needs to
+    // exist prior to creating connections.
```
GutoVeronezi, line 118: Comments would be better as Javadocs

harikrishna-patnala:
@blueorangutan package

DaanHoogland (approved):
clgtm, no event handling changes (other than FAILED) should happen. needs testing though.

DaanHoogland:
@blueorangutan package

weizhouapache (approved):
code lgtm

DaanHoogland:
@blueorangutan test ol9 kvm-ol9 keepEnv

DaanHoogland:
@blueorangutan test alm9 kvm-alma9 keepEnv

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
```diff
@@ -101,4 +113,33 @@ static String getHypervisorURI(String hypervisorType) {
...
+                while (true) {
+                    try {
+                        // This blocking call contains a loop of its own that will process events until the event loop is stopped or exception is thrown.
+                        Library.runEventLoop();
+                    } catch (LibvirtException e) {
+                        s_logger.error("LibvirtException was thrown in event loop: ", e);
+                    } catch (InterruptedException e) {
+                        s_logger.error("Libvirt event loop was interrupted: ", e);
+                    }
+                }
```
GutoVeronezi, line 137: Perhaps we could extract this block to a method (and turn the comment into a javadoc), and pass the method to the thread. What do you think, @mlsorensen?

DaanHoogland:
@blueorangutan test alma9 kvm-alma9 keepEnv

DaanHoogland:
teste out-of-bounds shutdown of a vm. Is there anything else that should be tested here, @mlsorensen ?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
mlsorensen: The comment doesn't apply to the while loop, it's just a reminder of what `Library.runEventLoop()` does. `Library.runEventLoop()` already has javadoc though so I'd be inclined to just remove the comment - the reason I added it is because it is not immediately obvious whether `Library.runEventLoop()` is just running event loop once, or calling `virEventRunDefaultImpl` repeatedly as required by the libvirt event framework.

mlsorensen:
I tested libvirt restart. This triggers an agent restart (as it did before) so it doesn't really test well what happens if connection to libvirt is lost, but the idea is that any LibvirtConnection should set up event processing again, if needed.

DaanHoogland:
> I tested libvirt restart. This triggers an agent restart (as it did before) so it doesn't really test well what happens if connection to libvirt is lost, but the idea is that any LibvirtConnection should set up event processing again, if needed.

ok, now focussing on 4.19, but I'll try to device a (more intrusive) test for this later.

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtConnection.java
GutoVeronezi: Makes sense. In this case, I would be inclined to remove the comment as well.

harikrishna-patnala (approved):
LGTM, also tested the VM migrations with local storage.

sureshanaparti:
> I tested libvirt restart. This triggers an agent restart (as it did before) so it doesn't really test well what happens if connection to libvirt is lost, but the idea is that any LibvirtConnection should set up event processing again, if needed.

tested it, the agent restarts after libvirtd restart (`service libvirtd restart`). Can the agent retry / wait for libvirtd sometime, and setup connection & event listeners?

sureshanaparti (approved):
Tested some VM / Volumes operations from CloudStack, including the live migration between local pools (validates the issue https://github.com/apache/cloudstack/issues/7942, caused by Library.initEventLoop() earlier), and out of band shutdown. No issues observed. Changes LGTM.

rohityadavcloud:
@blueorangutan package

rohityadavcloud (approved):

rohityadavcloud:
cc @JoaoJandre if you want to consider this fix for 4.18.2.0, I'm not sure if it would be risky to have it in 4.18 branch so I've shifted this to 4.19 branch.

PR #8481, opened 2024-01-09
Enable over provisioning for SharedMountPoint primary storages

GutoVeronezi:
### Description

ACS does not allow operators to set the configuration `storage.overprovisioning.factor` for `SharedMountPoint` primary storages, presenting the following message when one tries to change it:

image

Also, the global setting does not impact `SharedMoundPoint` primary storages, which means that the over-provisioning is not applied properly for this storage type. 

Furthermore, the metric presents that the `SharedMountPoint` has the global over-provisioning applied; however, in a scenario where the over-provision factor is 2, the allocation threshold is 80%, and the metrics show that storage has 40% allocated, one cannot allocate to it anymore.

This PR intends to enable over-provisioning for `SharedMountPoint` primary storages.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

In an environment where the global over-provision factor was 1, the allocation threshold was 80%, and the `SharedMountPoint` primary storage had 80% allocated, I was able to change the over-provisioning factor of the `SharedMountPoint` primary storage and allocate a new volume on it, as, with the over-provisioning, ACS considered it as having 40% allocated.

### DIFF
--- a/api/src/main/java/com/cloud/storage/Storage.java
+++ b/api/src/main/java/com/cloud/storage/Storage.java
@@ -152,12 +152,12 @@ public class Storage {
         StorPool(true, true, true);
 
         private final boolean shared;
-        private final boolean overprovisioning;
+        private final boolean overProvisioning;
         private final boolean encryption;
 
-        StoragePoolType(boolean shared, boolean overprovisioning, boolean encryption) {
+        StoragePoolType(boolean shared, boolean overProvisioning, boolean encryption) {
             this.shared = shared;
-            this.overprovisioning = overprovisioning;
+            this.overProvisioning = overProvisioning;
             this.encryption = encryption;
         }
 
@@ -166,14 +166,16 @@ public class Storage {
         }
 
         public boolean supportsOverProvisioning() {
-            return overprovisioning;
+            return overProvisioning;
         }
 
-        public boolean supportsEncryption() { return encryption; }
+        public boolean supportsEncryption() {
+            return encryption;
+        }
     }
 
     public static List<StoragePoolType> getNonSharedStoragePoolTypes() {
-        List<StoragePoolType> nonSharedStoragePoolTypes = new ArrayList<StoragePoolType>();
+        List<StoragePoolType> nonSharedStoragePoolTypes = new ArrayList<>();
         for (StoragePoolType storagePoolType : StoragePoolType.values()) {
             if (!storagePoolType.isShared()) {
                 nonSharedStoragePoolTypes.add(storagePoolType);
--- a/api/src/main/java/com/cloud/storage/Storage.java
+++ b/api/src/main/java/com/cloud/storage/Storage.java
@@ -138,7 +138,7 @@ public class Storage {
         LVM(false, false, false), // XenServer local LVM SR
         CLVM(true, false, false),
         RBD(true, true, false), // http://libvirt.org/storage.html#StorageBackendRBD
-        SharedMountPoint(true, false, true),
+        SharedMountPoint(true, true, true),
         VMFS(true, true, false), // VMware VMFS storage
         PreSetup(true, true, false), // for XenServer, Storage Pool is set up by customers.
         EXT(false, true, false), // XenServer local EXT SR
--- a/api/src/test/java/com/cloud/storage/StorageTest.java
+++ b/api/src/test/java/com/cloud/storage/StorageTest.java
@@ -63,7 +63,7 @@ public class StorageTest {
         Assert.assertFalse(StoragePoolType.CLVM.supportsOverProvisioning());
         Assert.assertTrue(StoragePoolType.RBD.supportsOverProvisioning());
         Assert.assertTrue(StoragePoolType.PowerFlex.supportsOverProvisioning());
-        Assert.assertFalse(StoragePoolType.SharedMountPoint.supportsOverProvisioning());
+        Assert.assertTrue(StoragePoolType.SharedMountPoint.supportsOverProvisioning());
         Assert.assertTrue(StoragePoolType.VMFS.supportsOverProvisioning());
         Assert.assertTrue(StoragePoolType.PreSetup.supportsOverProvisioning());
         Assert.assertTrue(StoragePoolType.EXT.supportsOverProvisioning());
--- a/api/src/test/java/com/cloud/storage/StorageTest.java
+++ b/api/src/test/java/com/cloud/storage/StorageTest.java
@@ -52,7 +52,7 @@ public class StorageTest {
     }
 
     @Test
-    public void supportsOverprovisioningStoragePool() {
+    public void supportsOverProvisioningTestAllStoragePoolTypes() {
         Assert.assertTrue(StoragePoolType.Filesystem.supportsOverProvisioning());
         Assert.assertTrue(StoragePoolType.NetworkFilesystem.supportsOverProvisioning());
         Assert.assertFalse(StoragePoolType.IscsiLUN.supportsOverProvisioning());


GutoVeronezi:
@blueorangutan package

sureshanaparti (approved):
code changes lgtm, needs testing with SharedMountPoint pools.

DaanHoogland:
code looks good. I have a worry though. overprovisioning on storage in general is a dangerous thing. On cpu and memory any errors due to overprovisioning are likely to not be fatal, but on storage these can easily lead to data loss. when storage is allocated but not used we can apply overprovisioning but when parts of it get used these should not be considered in overprovisioning. I don't think our model supports such behaviour. any ideas @GutoVeronezi @sureshanaparti  (sorry, i don't know what wisdom is in respect to this)

JoaoJandre:
@GutoVeronezi could you target 4.18.2?

JoaoJandre (commented):
CLGTM, didn't test it

JoaoJandre (approved):

weizhouapache:
> code looks good. I have a worry though. overprovisioning on storage in general is a dangerous thing. On cpu and memory any errors due to overprovisioning are likely to not be fatal, but on storage these can easily lead to data loss. when storage is allocated but not used we can apply overprovisioning but when parts of it get used these should not be considered in overprovisioning. I don't think our model supports such behaviour. any ideas @GutoVeronezi @sureshanaparti  (sorry, i don't know what wisdom is in respect to this)

@DaanHoogland 
I believe most cloudstack users use thin provisioning and storage overprovisioning. It was never a critical issue, there was no data loss either. If the storage is almost occupied, users should consider adding more storage before reaching full disk space.

GutoVeronezi:
> @GutoVeronezi could you target 4.18.2?

@JoaoJandre, done.


> code looks good. I have a worry though. overprovisioning on storage in general is a dangerous thing. On cpu and memory any errors due to overprovisioning are likely to not be fatal, but on storage these can easily lead to data loss. when storage is allocated but not used we can apply overprovisioning but when parts of it get used these should not be considered in overprovisioning. I don't think our model supports such behaviour. any ideas @GutoVeronezi @sureshanaparti (sorry, i don't know what wisdom is in respect to this)

Sure, dealing with over-provisioning without caution can be harmful. Before changing the overprovisioning configurations, operators should understand how the environment is being consumed. Furthermore, it is important to have active monitoring over the environment; this way, it is possible to take actions, like the one @weizhouapache mentioned in https://github.com/apache/cloudstack/pull/8481#issuecomment-1884729728, to avoid problems.

The configuration `cluster.storage.capacity.notificationthreshold` makes ACS notify the operators when the storage reaches the capacity threshold (for memory and CPU we have similar configurations). I am not sure if we have much more to do from the ACS perspective, but we can think about ways to improve this feature.

GutoVeronezi:
@blueorangutan package

weizhouapache:
> > @GutoVeronezi could you target 4.18.2?
> 
> @JoaoJandre, done.
> 
> 
> > code looks good. I have a worry though. overprovisioning on storage in general is a dangerous thing. On cpu and memory any errors due to overprovisioning are likely to not be fatal, but on storage these can easily lead to data loss. when storage is allocated but not used we can apply overprovisioning but when parts of it get used these should not be considered in overprovisioning. I don't think our model supports such behaviour. any ideas @GutoVeronezi @sureshanaparti (sorry, i don't know what wisdom is in respect to this)
> 
> @DaanHoogland
> Sure, dealing with over-provisioning without caution can be harmful. Before changing the overprovisioning configurations, operators should understand how the environment is being consumed. Furthermore, it is important to have active monitoring over the environment; this way, it is possible to take actions, like the one @weizhouapache mentioned in https://github.com/apache/cloudstack/pull/8481#issuecomment-1884729728, to avoid problems.
> 
> The configuration `cluster.storage.capacity.notificationthreshold` makes ACS notify the operators when the storage reaches the capacity threshold (for memory and CPU we have similar configurations). I am not sure if we have much more to do from the ACS perspective, but we can think about ways to improve this feature.
> 
> 
> 

Yes @GutoVeronezi, there are global settings to (1) notify users, or (2) disable new volume allocation, if allocated capacity or actual used capacity reaches a threshold.

The thresholds are around 75% or 85%, so users have enough time to get more storage or clean the storage.

DaanHoogland:
> > > @GutoVeronezi could you target 4.18.2?
> > 
> > 
> > @JoaoJandre, done.
> > > code looks good. I have a worry though. overprovisioning on storage in general is a dangerous thing. On cpu and memory any errors due to overprovisioning are likely to not be fatal, but on storage these can easily lead to data loss. when storage is allocated but not used we can apply overprovisioning but when parts of it get used these should not be considered in overprovisioning. I don't think our model supports such behaviour. any ideas @GutoVeronezi @sureshanaparti (sorry, i don't know what wisdom is in respect to this)
> > 
> > 
> > @DaanHoogland
> > Sure, dealing with over-provisioning without caution can be harmful. Before changing the overprovisioning configurations, operators should understand how the environment is being consumed. Furthermore, it is important to have active monitoring over the environment; this way, it is possible to take actions, like the one @weizhouapache mentioned in [#8481 (comment)](https://github.com/apache/cloudstack/pull/8481#issuecomment-1884729728), to avoid problems.
> > The configuration `cluster.storage.capacity.notificationthreshold` makes ACS notify the operators when the storage reaches the capacity threshold (for memory and CPU we have similar configurations). I am not sure if we have much more to do from the ACS perspective, but we can think about ways to improve this feature.
> 
> Yes @GutoVeronezi, there are global settings to (1) notify users, or (2) disable new volume allocation, if allocated capacity or actual used capacity reaches a threshold.
> 
> The thresholds are around 75% or 85%, so users have enough time to get more storage or clean the storage.

ok, I thought only overprovisioned threshold would be signalled. I don't know of an extra guard on the actual threshold , but then again we would expect operators to do their due diligence. Sorry for the over-paranoia.

weizhouapache:
> > Yes @GutoVeronezi, there are global settings to (1) notify users, or (2) disable new volume allocation, if allocated capacity or actual used capacity reaches a threshold.
> > The thresholds are around 75% or 85%, so users have enough time to get more storage or clean the storage.
> 
> ok, I thought only overprovisioned threshold would be signalled. I don't know of an extra guard on the actual threshold , but then again we would expect operators to do their due diligence. Sorry for the over-paranoia.

no worries @DaanHoogland 
I think this is a frequently asked question for many users.

```
mysql> select name,default_value,description from configuration where name like "cluster.storage%" or name like "pool.storage%";
+----------------------------------------------------------+---------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| name                                                     | default_value | description                                                                                                                                                      |
+----------------------------------------------------------+---------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| cluster.storage.allocated.capacity.notificationthreshold | 0.75          | Percentage (as a value between 0 and 1) of allocated storage utilization above which alerts will be sent about low storage available.                            |
| cluster.storage.capacity.notificationthreshold           | 0.75          | Percentage (as a value between 0 and 1) of storage utilization above which alerts will be sent about low storage available.                                      |
| cluster.storage.operations.exclude                       | false         | Exclude cluster from storage operations                                                                                                                          |
| pool.storage.allocated.capacity.disablethreshold         | 0.85          | Percentage (as a value between 0 and 1) of allocated storage utilization above which allocators will disable using the pool for low allocated storage available. |
| pool.storage.capacity.disablethreshold                   | 0.85          | Percentage (as a value between 0 and 1) of storage utilization above which allocators will disable using the pool for low storage available.                     |
+----------------------------------------------------------+---------------+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
5 rows in set (0.00 sec)

```

GutoVeronezi:
Aside from the third-party testing, do we still need anything else on this one?

cc: @DaanHoogland @weizhouapache

DaanHoogland:
I don´t know, I didn't test it @GutoVeronezi

weizhouapache:
> Aside from the third-party testing, do we still need anything else on this one?
> 
> cc: @DaanHoogland @weizhouapache

@GutoVeronezi 
code looks ok to me.

@slavkap 
can you please review and test it ?

DaanHoogland:
> I don´t know, I didn't test it @GutoVeronezi

sorry, I didn't read very well and answered hastely. Yes @GutoVeronezi, clgtm

rohityadavcloud (approved):

GutoVeronezi:
@blueorangutan package

slavkap (approved):
@weizhouapache, sorry for the late response!
Unfortunately, I don't have an environment with shared mount point but the code LGTM

weizhouapache:
> @weizhouapache, sorry for the late response! Unfortunately, I don't have an environment with shared mount point but the code LGTM

thanks @slavkap

PR #8438, opened 2024-01-04
JRE17 Migration Investigation

rohityadavcloud:
This PR is to investigate if CloudStack can be used against JRE17 (LTS) as JRE11 reached EOL.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)
- [x] build/CI

#### Feature/Enhancement Scale

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

PR #8433, opened 2024-01-03
Fix KVM import unmanaged instance

nvazquez:
### Description

This PR fixes KVM manage/unmanage functionality on 4.19.0 RC1 - was introduced on https://github.com/apache/cloudstack/pull/7976 but the latest merge commits on the PR removed the execution for KVM

image

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)
- [ ] build/CI

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
- Create KVM VM
- Unmanage VM
- Manage VM selecting the same template, network and service offering

PR #8502, opened 2024-01-12
Fix host stuck in connecting state

vishesh92:
### Description

There are a lot of test failures due to `test_vm_life_cycle.py` in multiple PRs due to host not available for migration of VMs.
https://github.com/apache/cloudstack/pull/8438#issuecomment-1886990285
https://github.com/apache/cloudstack/pull/8433#issuecomment-1878406524
https://github.com/apache/cloudstack/pull/7344#issuecomment-1886811369

While debugging I noticed that the hosts get stuck in `Connecting` state because MS is waiting for a response of the `ReadyCommand` from the agent. Since we take a lock on connection and disconnection, restarting the agent doesn't work. To fix this, we have to restart the MS or wait for ~1 hour (default timeout).

On the agent side, it gets stuck waiting for a response from the Script execution.

To reproduce, run `smoke/test_vm_life_cycle.py` (`TestSecuredVmMigration` test class to be specific). Once the tests are complete, you will notice that some hosts are stuck in `Connecting` state. And restarting the agent fails due to the named lock. Locks on DB can be checked using the below query.
```sql
SELECT *
FROM performance_schema.metadata_locks
INNER JOIN performance_schema.threads ON THREAD_ID = OWNER_THREAD_ID
WHERE PROCESSLIST_ID <> CONNECTION_ID() \G;
```

This PR adds a wait for the ready command and a timeout to the Script execution to ensure that the thread doesn't get stuck and the named lock from database is released.

### Types of changes

- [ ] build/CI

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

Ran `smoke/test_vm_life_cycle.py` multiple times. The host does get stuck again but only for 1 minute instead of 1 hour.

### DIFF
--- a/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
+++ b/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
@@ -596,6 +596,7 @@ public class AgentManagerImpl extends ManagerBase implements AgentManager, Handl
 
         final Long dcId = host.getDataCenterId();
         final ReadyCommand ready = new ReadyCommand(dcId, host.getId(), NumbersUtil.enableHumanReadableSizes);
+        ready.setWait(60);
         final Answer answer = easySend(hostId, ready);
         if (answer == null || !answer.getResult()) {
             // this is tricky part for secondary storage
--- a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
+++ b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
@@ -121,6 +121,7 @@ public class DefaultHostListener implements HypervisorHostListener {
     public boolean hostConnect(long hostId, long poolId) throws StorageConflictException {
         StoragePool pool = (StoragePool) this.dataStoreMgr.getDataStore(poolId, DataStoreRole.Primary);
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
+        cmd.setWait(60);
         final Answer answer = agentMgr.easySend(hostId, cmd);
 
         if (answer == null) {
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
@@ -55,7 +55,7 @@ public final class LibvirtReadyCommandWrapper extends CommandWrapper<ReadyComman
             cmd = "dpkg -l ovmf";
         }
         s_logger.debug("Running command : " + cmd);
-        int result = Script.runSimpleBashScriptForExitValue(cmd);
+        int result = Script.runSimpleBashScriptForExitValue(cmd, 60, false);
         s_logger.debug("Got result : " + result);
         return result == 0;
     }


vishesh92:
@blueorangutan package

GutoVeronezi:
@vishesh92, please, describe the situation you are facing and the tests you have done.

vishesh92:
@blueorangutan package

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
```diff
@@ -596,6 +596,7 @@ protected AgentAttache notifyMonitorsOfConnection(final AgentAttache attache, fi
 
         final Long dcId = host.getDataCenterId();
         final ReadyCommand ready = new ReadyCommand(dcId, host.getId(), NumbersUtil.enableHumanReadableSizes);
+        ready.setWait(60);
```
GutoVeronezi, line 599: These wait/timeouts should be externalized.

weizhouapache: I remember somebody complained there are too much global settings.
what's the benefit of adding another setting ? @GutoVeronezi 

normally ReadyCommand is executed in less than 1 second. 
60 seconds is very safe value.

weizhouapache:
@vishesh92 
I also suggest to add the following change
```
diff --git a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
index e344a87831d..90e8742c84d 100644
--- a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
+++ b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
@@ -121,6 +121,7 @@ public class DefaultHostListener implements HypervisorHostListener {
     public boolean hostConnect(long hostId, long poolId) throws StorageConflictException {
         StoragePool pool = (StoragePool) this.dataStoreMgr.getDataStore(poolId, DataStoreRole.Primary);
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
+        cmd.setWait(60);
         final Answer answer = agentMgr.easySend(hostId, cmd);
 
         if (answer == null) {
```

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
GutoVeronezi: In general, hard coded "magic" numbers are not interesting; we bind the system to what we think is proper and do not allow the operators to shape the system according to their use case.

For the timeout/wait settings, we could think on a flexible mecanism to allow operators overriding the timeout/wait values in a granular way, e.g. defining a timeout/wait for each command, and fallbacking to the global setting `wait` in the absence of the specifc setting.

weizhouapache: @GutoVeronezi 
the problem is actually caused by the falling back to the default `wait` which is 30 mins by default.

weizhouapache:
@blueorangutan package

shwstppr:
@blueorangutan test

weizhouapache:

weizhouapache:
test ubuntu22 kvm-rocky8

weizhouapache:
@blueorangutan test ubuntu22 kvm-rocky8

shwstppr:
@blueorangutan ubuntu22 kvm-rocky8

shwstppr:
@blueorangutan test ubuntu22 kvm-rocky8

shwstppr:
@blueorangutan test

shwstppr (approved):
LGTM. No host in connecting state issue now. Some failures with U22 mgmt server looks intermittent and unrelated

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
DaanHoogland: I agree with @GutoVeronezi , but in this case if it fixes we can add a new ticket for externalizing it and merge this.

DaanHoogland (approved):
clgtm

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
vishesh92: I am with @weizhouapache on this. We already have a lot of global settings which makes it confusing for the end user as well.

shwstppr: @vishesh92 @DaanHoogland @GutoVeronezi @weizhouapache I'm inclined to include this fix in 4.19.0. I'm creating an improvement ticket for follow-up on externalizing the wait. Maybe we can look into some generic solution while addressing it.

shwstppr: https://github.com/apache/cloudstack/issues/8506

weizhouapache (approved):
code lgtm

engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java
DaanHoogland: the number of global setting we already have should not be an argument. Nobody know them by head and anybody will have to use search facilities to manage those anyway. We have not the value 60 in three separate extra places for a value that has a default of 1800 (?). Both tuning and code maintenance are served by creating an externalisation for those.
@weizhouapache I understand your argument against the need for tuning, however we are now setting 60 for a value that you say should be one. That sounds like we will want to tune it to 10 or 5 or maybe to 7 afterwards.

Anyway, this is merged. Let's discuss on #8506.

PR #8543, opened 2024-01-21
ui: Updated units names

OlegChuev:
### Description

This PR...

### Types of changes

- [ ] build/CI

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

Before the fix
image

After the fix
image


ui/public/locales/en.json
```diff
@@ -1197,7 +1197,7 @@
 "label.maxinstance": "Max instances",
 "label.maxiops": "Max IOPS",
 "label.maxmembers": "Max members",
-"label.maxmemory": "Max. memory (MiB)",
+"label.maxmemory": "Max. memory (MB)",
```
DaanHoogland, line 1200: are we sure it is not in multitudes of 1024*1024 instead of 1.000.000?

rohityadavcloud: MiB is appropriate here; our memory calculations are not dividing bytes by 10^6. Is there a specific place you're seeing something not expected?

DaanHoogland:
@OlegChuev any opinion on our comments? we are planning for 4.18.2 and it would be nice to resolve this before then.

ui/public/locales/en.json
rohityadavcloud:
```suggestion
"label.maxmemory": "Max. memory (MiB)",
```

ui/public/locales/en.json
```diff
@@ -1253,7 +1253,7 @@
 "label.miniops": "Min IOPS",
 "label.minmaxiops": "Min IOPS / Max IOPS",
 "label.minmembers": "Min members",
-"label.minmemory": "Min memory (in MB)",
+"label.minmemory": "Min. memory (MB)",
```
rohityadavcloud, line 1256:
```suggestion
"label.minmemory": "Min. memory (MiB)",
```

rohityadavcloud (approved):

rohityadavcloud:
@blueorangutan ui

shwstppr (approved):

Issue #8506, opened 2024-01-15
improvement: externalize timeout for hypervisor/agent commands


shwstppr:
As discussed in https://github.com/apache/cloudstack/pull/8502#discussion_r1450410628 it could be useful to externalize the wait timeouts for hypervisor/agent commands like ModifyStoragePoolCommand.
The downside of introducing new global configurations for each of these commands is having too many configurations so maybe a generic solution can be implemented.

shwstppr:
Moving to 4.19.1.0 milestone for possible work on a generic solution for externalizing command timeouts

GutoVeronezi:
Hey @DaanHoogland @vishesh92 

I did not have time to elaborate on the proposal until now, but here is the draft.

For some edge cases, we have "wait"/"timeout" hardcoded values or values that come from global settings to set the commands timeout, like for `GetAutoScaleMetricsCommand` and for `MigrateCommand`. Currently, we have 385 classes that extend `com.cloud.agent.api.Command` (I skipped classes that extend `com.cloud.agent.api.Answer`); therefore, externalizing global settings for each command  to provide flexibility is not feasible; however, we can reach flexibility by other means. 

Currently, if we do not set a `wait` to the command, `AgentManagerImpl` will fallback to the `wait` global setting:

https://github.com/apache/cloudstack/blob/af8a582055c4194d1e42c8c1abd96969692046ed/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentManagerImpl.java#L441-L443
 
I propose to create a table of `command_timeout` in which we would store the classpath of the command and its wait/timeout. Then, instead of fall backing directly to the `wait` global setting, we would check if there is an entry of the command in this table. If it exists, then we use the entry's value; otherwise, we fall back to the `wait` global setting. Thus, we can change the commands wait/timeout on demand and in a flexible way.

The first step of the proposal would be creating the table and changing `AgentManagerImpl` to check the table. We could already create some entries for hardcoded values, like for the `GetAutoScaleMetricsCommand`.

Then, as a second step, we could create APIs to allow operators to list, persist, and remove the wait/timeouts. We could also create an annotation to describe the commands and create an API to allow operators to list the annotated commands; thus, making it clear to them which processes they can manage the timeout.

As a third and final step, we could extend this feature to be managed via GUI.

I will create a POC of this proposal; meanwhile, let me know your thoughts about the proposal.

Note: this proposal cover the common cases; eventually we will discuss the edge cases.

---

[...]

Issue #8011, opened 2023-09-27
Add timeout for handling of Command on agent's implementation


vishesh92:
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
Agent, ServerResource
~~~

##### CLOUDSTACK VERSION

~~~
All versions
~~~

##### CONFIGURATION

N/A

##### OS / ENVIRONMENT

N/A

##### SUMMARY

The value for [wait](https://github.com/apache/cloudstack/blob/main/api/src/main/java/com/cloud/agent/api/Command.java#L39) set in Command is not honored by agents (SSVM, host agent, etc.) while processing the request. These requests can also pile up if the same request is sent multiple times because of retries by user or the management server itself. This can potentially cause issues on SSVM and agents while executing long tasks leading to high wait times for commands to get executed on host agents, ssvm, etc.

##### STEPS TO REPRODUCE

1. Add a sleep in a ssvm command for 60 seconds and a log line before and after the sleep to check number of executions for that command.
2. Send 100 requests in parallel which will execute the above ssvm command with cmd.setWait(5).
3. Check logs on ssvm to check the actual number of requests received. 

Note: The above example is for a Command for which executeInSequence is false.

##### EXPECTED RESULTS

~~~
If the request on management server times out in 5 seconds, the processing of command on SSVM should stop after 5 seconds as well.
~~~

##### ACTUAL RESULTS

The number of requests received on the SSVM would be executing these requests in batches of 10 requests or so. The requests would still get processed even if the server itself has timed out and no longer waiting for the answer. You will the log for both before sleep and after sleep.

vishesh92:
cc: @DaanHoogland

DaanHoogland:
@vishesh92 I think you are right in your expectations. There are some more cases to define/consider;
- will responses be sent of timed out but completed commands?
- will the time-out be from receiving of the command? I think you want the commands to not be executed if these have timeout before being on top of the queue, right?
- are time-out responses being sent?
- ?

DaanHoogland:
@vishesh92 is this still relevant? it seems the configurable timeouts include this.

sureshanaparti:
@vishesh92 is this case addressed with https://github.com/apache/cloudstack/pull/9659?

DaanHoogland:
@sureshanaparti @vishesh92 , can we close this?

sureshanaparti:
[...]

PR #8547, opened 2024-01-23
Externalise a few timeouts & fix timeout for hostSupportsUefi in libvirt ready command wrapper

vishesh92:
### Description

This PR fixes bug introduced in #8502. Timeout for script execution was set to 60 ms instead of 60s which resulted in host not getting UEFI enabled. This is a blocker for 4.19 release.

We do this by introducing a new agent parameter `agent.script.timeout` (default - 60 seconds) to use as a timeout for the script checking host's UEFI status.

We also externalize the timeout for the ReadyCommand by introducing a new global setting `ready.command.wait` (default - 60 seconds).

For ModifyStoragePoolCommand, we don't externalize the timeout to avoid confusion for the user. Since, the required timeout can vary depending on the provider in use and we are only setting the wait for default host listener for now. Instead, we reuse the global `wait` setting by dividing it by `5` making the default value of 6 minutes (1800/5 = 360s) for ModifyStoragePoolCommand.

Note: the actual time, the MS waits is twice the wait set for a Command. Check reference code below.
https://github.com/apache/cloudstack/blob/19250403e645c76f60b17aa4aeb4dc915f5ca206/engine/orchestration/src/main/java/com/cloud/agent/manager/AgentAttache.java#L406-L442

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

#### Bug Severity

- [x] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/agent/conf/agent.properties
+++ b/agent/conf/agent.properties
@@ -379,7 +379,7 @@ iscsi.session.cleanup.enabled=false
 #stop.script.timeout=120
 
 # Time (in seconds) to wait for scripts to complete.
-# This is not used for every script execution.
+# This is currently used only while checking if the host supports UEFI.
 #agent.script.timeout=60
 
 # Definition of VMs video model type.
--- a/agent/conf/agent.properties
+++ b/agent/conf/agent.properties
@@ -5,9 +5,9 @@
 # to you under the Apache License, Version 2.0 (the
 # "License"); you may not use this file except in compliance
 # with the License.  You may obtain a copy of the License at
-# 
+#
 #   http://www.apache.org/licenses/LICENSE-2.0
-# 
+#
 # Unless required by applicable law or agreed to in writing,
 # software distributed under the License is distributed on an
 # "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
@@ -377,6 +377,10 @@ iscsi.session.cleanup.enabled=false
 # Time (in seconds) to wait for the VM to shutdown gracefully.
 # If the time is exceeded shutdown will be forced.
 #stop.script.timeout=120
+
+# Time (in seconds) to wait for scripts to complete.
+# This is not used for every script execution.
+#agent.script.timeout=60
 
 # Definition of VMs video model type.
 #vm.video.hardware=
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -660,12 +660,12 @@ public class AgentProperties{
     public static final Property<Integer> STOP_SCRIPT_TIMEOUT = new Property<>("stop.script.timeout", 120);
 
     /**
-     * Time (in seconds) to wait for libvirt script to check host uefi status.<br>
-     * If the time is exceeded, uefi will be disabled on the host.<br>
+     * Time (in seconds) to wait for scripts to complete.<br>
+     * This is not used for every script execution.<br>
      * Data type: Integer.<br>
      * Default value: <code>60</code>
      */
-    public static final Property<Integer> LIBVIRT_HOST_UEFI_SCRIPT_TIMEOUT = new Property<>("libvirt.host.uefi.script.timeout", 60);
+    public static final Property<Integer> AGENT_SCRIPT_TIMEOUT = new Property<>("agent.script.timeout", 60);
 
     /**
      * Definition of VMs video model type.<br>
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -660,6 +660,14 @@ public class AgentProperties{
     public static final Property<Integer> STOP_SCRIPT_TIMEOUT = new Property<>("stop.script.timeout", 120);
 
     /**
+     * Time (in seconds) to wait for libvirt script to check host uefi status.<br>
+     * If the time is exceeded, uefi will be disabled on the host.<br>
+     * Data type: Integer.<br>
+     * Default value: <code>60</code>
+     */
+    public static final Property<Integer> LIBVIRT_HOST_UEFI_SCRIPT_TIMEOUT = new Property<>("libvirt.host.uefi.script.timeout", 60);
+
+    /**
      * Definition of VMs video model type.<br>
      * Data type: String.<br>
      * Default value: <code>null</code>
--- a/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
+++ b/agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
@@ -661,7 +661,7 @@ public class AgentProperties{
 
     /**
      * Time (in seconds) to wait for scripts to complete.<br>
-     * This is not used for every script execution.<br>
+     * This is currently used only while checking if the host supports UEFI.<br>
      * Data type: Integer.<br>
      * Default value: <code>60</code>
      */
--- a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
+++ b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
@@ -124,7 +124,7 @@ public class DefaultHostListener implements HypervisorHostListener {
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
         cmd.setWait(Wait.value() / 5);
         s_logger.debug(String.format("Sending modify storage pool command to agent: %d for storage pool: %d with timeout %d seconds",
-                hostId, poolId, cmd.getWait() / 5));
+                hostId, poolId, cmd.getWait()));
         final Answer answer = agentMgr.easySend(hostId, cmd);
 
         if (answer == null) {
--- a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
+++ b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
@@ -45,8 +45,6 @@ import com.cloud.storage.dao.StoragePoolHostDao;
 import com.cloud.utils.exception.CloudRuntimeException;
 import org.apache.cloudstack.engine.subsystem.api.storage.DataStoreManager;
 import org.apache.cloudstack.engine.subsystem.api.storage.HypervisorHostListener;
-import org.apache.cloudstack.framework.config.ConfigKey;
-import org.apache.cloudstack.framework.config.Configurable;
 import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolDetailVO;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolDetailsDao;
@@ -57,12 +55,10 @@ import org.apache.log4j.Logger;
 import javax.inject.Inject;
 import java.util.List;
 
-public class DefaultHostListener implements HypervisorHostListener, Configurable {
-    private static final Logger s_logger = Logger.getLogger(DefaultHostListener.class);
-    ConfigKey<Integer> ModifyStoragePoolCommandWait = new ConfigKey<Integer>("Advanced", Integer.class,
-            "modify.storage.pool.command.wait", "60",
-            "Time in seconds to wait for ModifyStoragePoolCommand command to return", true);
+import static com.cloud.agent.AgentManager.Wait;
 
+public class DefaultHostListener implements HypervisorHostListener {
+    private static final Logger s_logger = Logger.getLogger(DefaultHostListener.class);
     @Inject
     AgentManager agentMgr;
     @Inject
@@ -89,15 +85,6 @@ public class DefaultHostListener implements HypervisorHostListener, Configurable
     ConfigurationManager configManager;
     @Inject
     NetworkDao networkDao;
-
-    @Override
-    public String getConfigComponentName() {
-        return DefaultHostListener.class.getSimpleName();
-    }
-    @Override
-    public ConfigKey<?>[] getConfigKeys() {
-        return new ConfigKey<?>[] {ModifyStoragePoolCommandWait};
-    }
 
     @Override
     public boolean hostAdded(long hostId) {
@@ -135,7 +122,7 @@ public class DefaultHostListener implements HypervisorHostListener, Configurable
     public boolean hostConnect(long hostId, long poolId) throws StorageConflictException {
         StoragePool pool = (StoragePool) this.dataStoreMgr.getDataStore(poolId, DataStoreRole.Primary);
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
-        cmd.setWait(ModifyStoragePoolCommandWait.value());
+        cmd.setWait(Wait.value()/5);
         final Answer answer = agentMgr.easySend(hostId, cmd);
 
         if (answer == null) {
--- a/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
+++ b/engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
@@ -45,6 +45,8 @@ import com.cloud.storage.dao.StoragePoolHostDao;
 import com.cloud.utils.exception.CloudRuntimeException;
 import org.apache.cloudstack.engine.subsystem.api.storage.DataStoreManager;
 import org.apache.cloudstack.engine.subsystem.api.storage.HypervisorHostListener;
+import org.apache.cloudstack.framework.config.ConfigKey;
+import org.apache.cloudstack.framework.config.Configurable;
 import org.apache.cloudstack.storage.datastore.db.PrimaryDataStoreDao;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolDetailVO;
 import org.apache.cloudstack.storage.datastore.db.StoragePoolDetailsDao;
@@ -55,8 +57,12 @@ import org.apache.log4j.Logger;
 import javax.inject.Inject;
 import java.util.List;
 
-public class DefaultHostListener implements HypervisorHostListener {
+public class DefaultHostListener implements HypervisorHostListener, Configurable {
     private static final Logger s_logger = Logger.getLogger(DefaultHostListener.class);
+    ConfigKey<Integer> ModifyStoragePoolCommandWait = new ConfigKey<Integer>("Advanced", Integer.class,
+            "modify.storage.pool.command.wait", "60",
+            "Time in seconds to wait for ModifyStoragePoolCommand command to return", true);
+
     @Inject
     AgentManager agentMgr;
     @Inject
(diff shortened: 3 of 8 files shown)


vishesh92:
@blueorangutan package

shwstppr:
@blueorangutan package

sureshanaparti (approved):
code lgtm

DaanHoogland (approved):
clgtm

DaanHoogland:
is this maybe suitable for 4.18.2 even?

vishesh92:
> is this maybe suitable for 4.18.2 even?

We will need to add changes in this PR https://github.com/apache/cloudstack/pull/8502 to 4.18 as well.

weizhouapache (approved):
code lgtm

rohityadavcloud (approved):

vishesh92:
@blueorangutan package

vishesh92:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
```diff
@@ -51,11 +53,12 @@ public Answer execute(final ReadyCommand command, final LibvirtComputingResource
 
     private boolean hostSupportsUefi(boolean isUbuntuHost) {
         String cmd = "rpm -qa | grep -i ovmf";
+        int timeout = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.LIBVIRT_HOST_UEFI_SCRIPT_TIMEOUT) * 1000; // Get property value & convert to milliseconds
         if (isUbuntuHost) {
             cmd = "dpkg -l ovmf";
         }
         s_logger.debug("Running command : " + cmd);
```
weizhouapache, line 60: add timeout value to debug message ?

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
```diff
@@ -55,8 +57,12 @@
 import javax.inject.Inject;
 import java.util.List;
 
-public class DefaultHostListener implements HypervisorHostListener {
+public class DefaultHostListener implements HypervisorHostListener, Configurable {
     private static final Logger s_logger = Logger.getLogger(DefaultHostListener.class);
+    ConfigKey<Integer> ModifyStoragePoolCommandWait = new ConfigKey<Integer>("Advanced", Integer.class,
+            "modify.storage.pool.command.wait", "60",
+            "Time in seconds to wait for ModifyStoragePoolCommand command to return", true);
```
sureshanaparti, line 64: CS have lot of such cmds, maybe not good idea to add a config at cmd level. (later, might end up one config for each cmd). generic configs, eg. storage.command.wait / host.command.wait, based on the category with appropriate descriptions might be good.

rohityadavcloud (approved):

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
vishesh92: I agree with you. But generic config might not work. e.g. `storage.command.wait` as 60 seconds will work for `ModifyStoragePoolCommand` but not might be appropriate for other storage related commands. It needs to be specific to a `Command`.

IMO, it would be better to have something which allows the user to set configuration if required like `command.<command name>.wait` and wait is set only if the value is set. If it's not set, then some default value is used.

vishesh92:
@blueorangutan package

vishesh92:
@blueorangutan test

vishesh92:
@blueorangutan test matrix

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
weizhouapache: What about
agent.startup.command.timeout
agent.startup.script.timeout 
?

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/wrapper/LibvirtReadyCommandWrapper.java
```diff
@@ -51,11 +53,12 @@ public Answer execute(final ReadyCommand command, final LibvirtComputingResource
 
     private boolean hostSupportsUefi(boolean isUbuntuHost) {
         String cmd = "rpm -qa | grep -i ovmf";
+        int timeout = AgentPropertiesFileHandler.getPropertyValue(AgentProperties.LIBVIRT_HOST_UEFI_SCRIPT_TIMEOUT) * 1000; // Get property value & convert to milliseconds
```
vishesh92, line 56: Add a common timeout for all scripts here.

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
```diff
@@ -121,7 +135,7 @@ private NicTO createNicTOFromNetworkAndOffering(NetworkVO networkVO, NetworkOffe
     public boolean hostConnect(long hostId, long poolId) throws StorageConflictException {
         StoragePool pool = (StoragePool) this.dataStoreMgr.getDataStore(poolId, DataStoreRole.Primary);
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
-        cmd.setWait(60);
+        cmd.setWait(ModifyStoragePoolCommandWait.value());
```
vishesh92, line 138: Use (wait.timeout / 6) here.

vishesh92:
@blueorangutan package

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
```diff
@@ -121,7 +122,7 @@ private NicTO createNicTOFromNetworkAndOffering(NetworkVO networkVO, NetworkOffe
     public boolean hostConnect(long hostId, long poolId) throws StorageConflictException {
         StoragePool pool = (StoragePool) this.dataStoreMgr.getDataStore(poolId, DataStoreRole.Primary);
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
-        cmd.setWait(60);
+        cmd.setWait(Wait.value()/5);
```
shwstppr, line 125: @vishesh92 do we need to add some logging here?

sureshanaparti, line 125:
```suggestion
        cmd.setWait(Wait.value() / 5);
```

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
```diff
@@ -121,7 +122,9 @@ private NicTO createNicTOFromNetworkAndOffering(NetworkVO networkVO, NetworkOffe
     public boolean hostConnect(long hostId, long poolId) throws StorageConflictException {
         StoragePool pool = (StoragePool) this.dataStoreMgr.getDataStore(poolId, DataStoreRole.Primary);
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
-        cmd.setWait(60);
+        cmd.setWait(Wait.value() / 5);
+        s_logger.debug(String.format("Sending modify storage pool command to agent: %d for storage pool: %d with timeout %d seconds",
+                hostId, poolId, cmd.getWait() / 5));
```
sureshanaparti, line 127:
```suggestion
                hostId, poolId, cmd.getWait()));
```

agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
```diff
@@ -659,6 +659,14 @@ public class AgentProperties{
      */
     public static final Property<Integer> STOP_SCRIPT_TIMEOUT = new Property<>("stop.script.timeout", 120);
 
+    /**
+     * Time (in seconds) to wait for scripts to complete.<br>
+     * This is not used for every script execution.<br>
+     * Data type: Integer.<br>
+     * Default value: <code>60</code>
+     */
+    public static final Property<Integer> AGENT_SCRIPT_TIMEOUT = new Property<>("agent.script.timeout", 60);
```
harikrishna-patnala, line 668: @vishesh92 we need to add this in agent.properties to make it configurable

vishesh92: We don't need to. If the user adds it in the file, it will get picked up.

harikrishna-patnala: yeah correct but that has to be known to the operator right, we normally put them in the properties file and comment them 

eg.

# Instance Conversion from Vmware to KVM through virt-v2v. Enable verbose mode
# virtv2v.verbose.enabled=false

vishesh92:
@blueorangutan package

vishesh92:
@blueorangutan test matrix

agent/src/main/java/com/cloud/agent/properties/AgentProperties.java
```diff
@@ -659,6 +659,14 @@ public class AgentProperties{
      */
     public static final Property<Integer> STOP_SCRIPT_TIMEOUT = new Property<>("stop.script.timeout", 120);
 
+    /**
+     * Time (in seconds) to wait for scripts to complete.<br>
+     * This is not used for every script execution.<br>
```
vishesh92, line 664:
```suggestion
     * This is currently used only while checking if the host supports UEFI.<br>
```

agent/conf/agent.properties
```diff
@@ -378,6 +378,10 @@ iscsi.session.cleanup.enabled=false
 # If the time is exceeded shutdown will be forced.
 #stop.script.timeout=120
 
+# Time (in seconds) to wait for scripts to complete.
+# This is not used for every script execution.
```
vishesh92, line 382:
```suggestion
# This is currently used only while checking if the host supports UEFI.
```

harikrishna-patnala (approved):
code LGTM

weizhouapache (approved):
code lgtm

weizhouapache:
@blueorangutan test ubuntu22 kvm-ubuntu22

weizhouapache:

rohityadavcloud:
@blueorangutan package

GutoVeronezi:
-1 (binding)

@vishesh92 this PR does not fix #8506. The issue and the thread it mentions is related to a generic solution, which was not even discussed.

shwstppr:
> -1 (binding)
> 
> @vishesh92 this PR does not fix #8506. The issue and the thread it mentions is related to a generic solution, which was not even discussed.

@GutoVeronezi would it make sense to keep the issue open for further work and continue with this PR or are you -1 on the PR changes as well?

DaanHoogland:
> -1 (binding)

binding does not apply to PRs. That said, if you have objections and a way forward we will not merge (on any solidly argued -1)

> @vishesh92 this PR does not fix #8506. The issue and the thread it mentions is related to a generic solution, which was not even discussed.

@GutoVeronezi , I suppose you are referring to the solution mentioned in this comment: https://github.com/apache/cloudstack/pull/8502#discussion_r1450506118 ?

If so, this is a good idea and would obsolete this PR, but not without implementation. Do you have code ready to replace the code here?

vishesh92:
> -1 (binding)
> 
> @vishesh92 this PR does not fix #8506. The issue and the thread it mentions is related to a generic solution, which was not even discussed.

@GutoVeronezi I agree that this PR does not fix #8506. We need to revisit the timeouts for Commands on both the MS (#8506) & agents (#8011 - Agents don't even consider timeout which is set in the Commands).

The main purpose of #8502 was to fix the failures we were seeing in e2e tests, but I introduced another bug (because of using 60ms instead of 60s for the script's timeout).

In this PR, I wanted to fix the bug I introduced and address your concerns at least for the changes I did in my previous PR (#8502).

Regarding externalizing timeouts for all commands, IMO we shouldn't try to fix it for now. I would like to do that but it would be a very big change and it will require more discussion and planning because the same `Command` can processed by different providers depending on an environment and the required timeout can be different depending on the scenario and the setup.

I would like to get this PR merged for now since it's a blocker for unblock 4.19 release. Let me know what you think?

GutoVeronezi:
> @GutoVeronezi would it make sense to keep the issue open for further work...

Sure do, because the issue is not being fixed by this PR.

> ... are you -1 on the PR changes as well?

No. Indeed, I proposed the externalization of the timeout settings, vide https://github.com/apache/cloudstack/pull/8502#discussion_r1450410628 (and the author of the PR was against it at first: https://github.com/apache/cloudstack/pull/8502#discussion_r1452036208). My point is that this PR does not even touch the essence of the issue; therefore, saying it solves the issue does not make sense.

> binding does not apply to PRs.

Thanks, @DaanHoogland. I think I missed this detail from the docs.

> @GutoVeronezi , I suppose you are referring to the solution mentioned in this comment: [#8502 (comment)](https://github.com/apache/cloudstack/pull/8502#discussion_r1450506118) ?
> 
> If so, this is a good idea and would obsolete this PR, but not without implementation. Do you have code ready to replace the code here?

No, there is no code ready yet; before implementing something every each way, we should discuss the pros and cons. That is the purpose of the issue.

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
```diff
@@ -121,7 +122,9 @@ private NicTO createNicTOFromNetworkAndOffering(NetworkVO networkVO, NetworkOffe
     public boolean hostConnect(long hostId, long poolId) throws StorageConflictException {
         StoragePool pool = (StoragePool) this.dataStoreMgr.getDataStore(poolId, DataStoreRole.Primary);
         ModifyStoragePoolCommand cmd = new ModifyStoragePoolCommand(true, pool);
-        cmd.setWait(60);
+        cmd.setWait(Wait.value() / 5);
```
GutoVeronezi, line 125: Why divide by 5?

GutoVeronezi:
> I would like to get this PR merged for now since it's a blocker for unblock 4.19 release. Let me know what you think?

I see #8502 was a blocker, but why this one is a blocker?

GutoVeronezi:
@vishesh92, please, improve the description of the PR to reflect the real proposal (which is different from issue #8506), mentioning PR #8502 and explaining why it is a blocker.

vishesh92:
> I see #8502 was a blocker, but why this one is a blocker?
I used 60 ms instead of 60000 ms as timeout for the script. This resulted in all hosts having UEFI status as false.

> @vishesh92, please, improve the description of the PR to reflect the real proposal (which is different from issue #8506), mentioning PR #8502 and explaining why it is a blocker.

I have updated the PR description.

engine/storage/volume/src/main/java/org/apache/cloudstack/storage/datastore/provider/DefaultHostListener.java
vishesh92: For ModifyStoragePoolCommand, we don't externalize the timeout to avoid confusion for the user. Since, the required timeout can vary depending on the provider in use and we are only setting the wait for default host listener for now. Instead, we reuse the global wait setting by dividing it by 5 making the default value of 6 minutes (1800/5 = 360s) for ModifyStoragePoolCommand.

GutoVeronezi: It still does not make sense. Why 5? Just to match 6 minutes when the default `wait` is 1800?

vishesh92: Yes. Reasoning behind this is that user might decrease the wait in some scenario. It's quite unlikely but if the user sets it to 5 min, we will wait at least 1 minute here.

P.S. - the actual wait time is twice the value set for wait. Check the note in the description.

GutoVeronezi: User also might increase `wait` to 18000 and we would have 1 hour here. It does not make sense to use an unfounded math function to derive a value; please, either use the same configuration or externalize a new one.

vishesh92: It's more about the worst case scenario. It will wait for the specified time only if the agent is stuck because of some reason. Before #8502, if the wait was set to 18000, this command would have waited for 10 hours (5 hours *2). This solution doesn't completely solve the the problem but reduces the issues to some extent until issues #8506 or #8011 is resolved.

GutoVeronezi: In this case, externalizing a new configuration is the best choice.

GutoVeronezi: In contrast with taking a value and dividing it by a random number, externalizing a new configuration would be the best option. We will not externalize every command timeout right now, so it would not be a problem. I agree a more robust solution, like what I mentioned in https://github.com/apache/cloudstack/pull/8502#discussion_r1450506118, is better; however, we have to discuss it first.

As last resort, if you do not want to externalize a new configuration, creating a constant with javadoc for this case is better than deriving the value.

vishesh92:
@blueorangutan package

vishesh92:
@blueorangutan test matrix

shwstppr (approved):
LGTM, uefi cmd running fine,

```
[root@pr8547-t6913-kvm-rocky8-kvm1 ~]# systemctl restart cloudstack-agent && tail -f /var/log/cloudstack/agent/agent.log 
2024-01-26 11:04:17,735 DEBUG [resource.wrapper.LibvirtReadyCommandWrapper] (agentRequest-Handler-4:null) (logid:c84198da) Running command : [rpm -qa | grep -i ovmf] with timeout : 60000 ms
2024-01-26 11:04:17,735 DEBUG [utils.script.Script] (agentRequest-Handler-4:null) (logid:c84198da) Executing command [/bin/bash -c rpm -qa | grep -i ovmf ].
2024-01-26 11:04:18,028 DEBUG [utils.script.Script] (agentRequest-Handler-4:null) (logid:c84198da) Successfully executed process [22872] for command [/bin/bash -c rpm -qa | grep -i ovmf ].
```

Without fix,
```
[root@ref-trl-6184-k-Mr8-nicolas-vazquez-kvm2 ~]# tail -f /var/log/cloudstack/agent/agent.log
2024-01-23 04:53:27,983 INFO  [cloud.agent.Agent] (agentRequest-Handler-5:null) (logid:b473a821) Set agent id 7
2024-01-23 04:53:27,984 INFO  [cloud.agent.Agent] (agentRequest-Handler-5:null) (logid:b473a821) Ready command is processed for agent id = 7
2024-01-23 04:53:28,064 WARN  [utils.script.Script] (Script-4:null) (logid:) Interrupting script.
2024-01-23 04:53:28,065 WARN  [utils.script.Script] (agentRequest-Handler-5:null) (logid:b473a821) Process [320455] for command [/bin/bash -c rpm -qa | grep -i ovmf ] timed out. Output is [].
```

shwstppr:
@GutoVeronezi are your concerns with this PR addressed? I'll proceed with 4.19RC4 if no objections

PR #8556, opened 2024-01-23
Allow deletion of system VM templates

GaOrtiga:
### Description

Currently, ACS does not allow the deletion of system VM templates. However, it might be of interest to the operator to delete templates from older versions that are no longer in use.

A change was applied to remove the restriction on the deletion of system VM templates, allowing operators to delete such templates.

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

### How Has This Been Tested?

I applied the changes and successfully deleted old system VM templates

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
@@ -55,6 +55,9 @@ public class DeleteTemplateCmd extends BaseAsyncCmd {
     @Parameter(name = ApiConstants.FORCED, type = CommandType.BOOLEAN, required = false, description = "Force delete a template.", since = "4.9+")
     private Boolean forced;
 
+    @Parameter(name = ApiConstants.IS_SYSTEM, type = CommandType.BOOLEAN, required = false, description = "Necessary if the template's type is system.", since = "4.20.0")
+    private Boolean isSystem;
+
     /////////////////////////////////////////////////////
     /////////////////// Accessors ///////////////////////
     /////////////////////////////////////////////////////
@@ -69,6 +72,10 @@ public class DeleteTemplateCmd extends BaseAsyncCmd {
 
     public boolean isForced() {
         return BooleanUtils.toBooleanDefaultIfNull(forced, false);
+    }
+
+    public boolean getIsSystem() {
+        return BooleanUtils.toBooleanDefaultIfNull(isSystem, false);
     }
 
     /////////////////////////////////////////////////////
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
@@ -19,6 +19,7 @@ package org.apache.cloudstack.api.command.user.template;
 import org.apache.commons.lang3.BooleanUtils;
 import org.apache.log4j.Logger;
 
+import org.apache.cloudstack.acl.RoleType;
 import org.apache.cloudstack.api.APICommand;
 import org.apache.cloudstack.api.ApiCommandResourceType;
 import org.apache.cloudstack.api.ApiConstants;
@@ -55,7 +56,7 @@ public class DeleteTemplateCmd extends BaseAsyncCmd {
     @Parameter(name = ApiConstants.FORCED, type = CommandType.BOOLEAN, required = false, description = "Force delete a template.", since = "4.9+")
     private Boolean forced;
 
-    @Parameter(name = ApiConstants.IS_SYSTEM, type = CommandType.BOOLEAN, required = false, description = "Necessary if the template's type is system.", since = "4.20.0")
+    @Parameter(name = ApiConstants.IS_SYSTEM, type = CommandType.BOOLEAN, required = false, description = "Necessary if the template's type is system.", since = "4.20.0", authorized = {RoleType.Admin})
     private Boolean isSystem;
 
     /////////////////////////////////////////////////////
--- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
+++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
@@ -748,9 +748,6 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
     public TemplateProfile prepareDelete(DeleteTemplateCmd cmd) {
         TemplateProfile profile = super.prepareDelete(cmd);
         VMTemplateVO template = profile.getTemplate();
-        if (template.getTemplateType() == TemplateType.SYSTEM) {
-            throw new InvalidParameterValueException("The DomR template cannot be deleted.");
-        }
         checkZoneImageStores(profile.getTemplate(), profile.getZoneIdList());
         return profile;
     }
--- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
+++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
@@ -748,6 +748,9 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
     public TemplateProfile prepareDelete(DeleteTemplateCmd cmd) {
         TemplateProfile profile = super.prepareDelete(cmd);
         VMTemplateVO template = profile.getTemplate();
+        if (template.getTemplateType() == TemplateType.SYSTEM && !cmd.getIsSystem()) {
+            throw new InvalidParameterValueException("Could not delete template as it is a SYSTEM template and isSystem is set to false.");
+        }
         checkZoneImageStores(profile.getTemplate(), profile.getZoneIdList());
         return profile;
     }


weizhouapache:
@GaOrtiga 
If you update the template to USER, then you will be able to remove it.
this will prevent some human mistakes

weizhouapache:
> @GaOrtiga If you update the template to USER, then you will be able to remove it. this will prevent some human mistakes

@GaOrtiga @DaanHoogland 
do you agree and close this PR ?

GaOrtiga:
> @GaOrtiga If you update the template to USER, then you will be able to remove it. this will prevent some human mistakes

@weizhouapache I understand that there is a way to circumvent this restriction, however, since this would only be possible to Root Admins, I believe deleting it directly should still be an option. To address your concerns about this change, I could make the `force` flag necessary when the template's type is `SYSTEM`.

Would this measure be enough to offset the concerns regarding this implementation?

server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
```diff
@@ -748,9 +748,6 @@ void checkAndRemoveTemplateDetails(VMTemplateVO template) {
     public TemplateProfile prepareDelete(DeleteTemplateCmd cmd) {
         TemplateProfile profile = super.prepareDelete(cmd);
         VMTemplateVO template = profile.getTemplate();
-        if (template.getTemplateType() == TemplateType.SYSTEM) {
-            throw new InvalidParameterValueException("The DomR template cannot be deleted.");
-        }
```
DaanHoogland, line 753:
```sugestion
        if (template.getTemplateType() == TemplateType.SYSTEM && ! cmd.isForced()) {
            throw new InvalidParameterValueException("The DomR template cannot be deleted.");
        }
```
is what you mean @GaOrtiga ?

GaOrtiga: Yes, and also update the message in the exception.

GaOrtiga: > ```
>         if (template.getTemplateType() == TemplateType.SYSTEM && ! cmd.isForced()) {
>             throw new InvalidParameterValueException("The DomR template cannot be deleted.");
>         }
> ```
> 
> is what you mean @GaOrtiga ?

@DaanHoogland @weizhouapache given that this operation can only be carried out by Root Admins, could we proceed with this proposed implementation?

weizhouapache:
@GaOrtiga @DaanHoogland 
sorry this PR is not needed in my opinion

DaanHoogland:
@GaOrtiga as a user wants this convenience. Even when I agree that it is not vital, that would be enough for it to be needed.
- Is there any more pressing objection @weizhouapache?
- Alternatively, is there a modus operandus that would make for a more acceptable change?

weizhouapache:
> @GaOrtiga as a user wants this convenience. Even when I agree that it is not vital, that would be enough for it to be needed.
> 
> * Is there any more pressing objection @weizhouapache?
> * Alternatively, is there a modus operandus that would make for a more acceptable change?

@DaanHoogland 
SYSTEM template is very important for cloudstack environments. We need to be very careful to remove it
a force flag is not enough to me.

since we already have a way to delete a SYSTEM template (see https://github.com/apache/cloudstack/pull/8556#issuecomment-1906168608), this change is unnecessary.

my question to @GaOrtiga @DaanHoogland 
is there special use case that we must be able to delete a SYSTEM template in one step ?

weizhouapache:
@DaanHoogland @GaOrtiga 
let me explain a bit more
users might delete the unused SYSTEM template one or twice every year by average. It takes just few minutes.

However, with this PR, they will be able to delete SYSTEM template by a single API/command.
If the SYSTEM template is removed by mistake, it will be a critical issue .
can you compare the pros and cons ?

GaOrtiga:
> @GaOrtiga @DaanHoogland sorry this PR is not needed in my opinion

Having a workaround should not be a reason to restrict an operation. If the operator does not know that you can circumvent this restriction they may feel obliged to perform the operation directly in the DB, especially since the exception thrown is very vague and editing the template's type to make it deletable is very non-intuitive.

GaOrtiga:
> > @GaOrtiga as a user wants this convenience. Even when I agree that it is not vital, that would be enough for it to be needed.
> > 
> > * Is there any more pressing objection @weizhouapache?
> > * Alternatively, is there a modus operandus that would make for a more acceptable change?
> 
> @DaanHoogland SYSTEM template is very important for cloudstack environments. We need to be very careful to remove it a force flag is not enough to me.
> 
> since we already have a way to delete a SYSTEM template (see [#8556 (comment)](https://github.com/apache/cloudstack/pull/8556#issuecomment-1906168608)), this change is unnecessary.
> 
> my question to @GaOrtiga @DaanHoogland is there special use case that we must be able to delete a SYSTEM template in one step ?

The use case would be any operator that wants to delete `SYSTEM` templates and does not know about the workaround, since it's not really something self-explanatory.

weizhouapache:
> The use case would be any operator that wants to delete `SYSTEM` templates and does not know about the workaround, since it's not really something self-explanatory.

@GaOrtiga 
this can be fixed by a PR with changes in the error message.

weizhouapache:
> > @GaOrtiga @DaanHoogland sorry this PR is not needed in my opinion
> 
> Having a workaround should not be a reason to restrict an operation. If the operator does not know that you can circumvent this restriction they may feel obliged to perform the operation directly in the DB, especially since the exception thrown is very vague and editing the template's type to make it deletable is very non-intuitive.

If the solution is not clear, make it clear then. for example
```
diff --git a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
index b886f0868f6..6456ca452ed 100644
--- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
+++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
@@ -749,7 +749,7 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
         TemplateProfile profile = super.prepareDelete(cmd);
         VMTemplateVO template = profile.getTemplate();
         if (template.getTemplateType() == TemplateType.SYSTEM) {
-            throw new InvalidParameterValueException("The DomR template cannot be deleted.");
+            throw new InvalidParameterValueException("The SYSTEM template cannot be deleted. Please update template type to USER and retry.");
         }
         checkZoneImageStores(profile.getTemplate(), profile.getZoneIdList());
         return profile;
```

DaanHoogland:
What do you think of @weizhouapache 's last sugestion @GaOrtiga ? would this satisfy both your requirements?

GaOrtiga:
> > > @GaOrtiga @DaanHoogland sorry this PR is not needed in my opinion
> > 
> > 
> > Having a workaround should not be a reason to restrict an operation. If the operator does not know that you can circumvent this restriction they may feel obliged to perform the operation directly in the DB, especially since the exception thrown is very vague and editing the template's type to make it deletable is very non-intuitive.
> 
> If the solution is not clear, make it clear then. for example
> 
> ```
> diff --git a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
> index b886f0868f6..6456ca452ed 100644
> --- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
> +++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
> @@ -749,7 +749,7 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
>          TemplateProfile profile = super.prepareDelete(cmd);
>          VMTemplateVO template = profile.getTemplate();
>          if (template.getTemplateType() == TemplateType.SYSTEM) {
> -            throw new InvalidParameterValueException("The DomR template cannot be deleted.");
> +            throw new InvalidParameterValueException("The SYSTEM template cannot be deleted. Please update template type to USER and retry.");
>          }
>          checkZoneImageStores(profile.getTemplate(), profile.getZoneIdList());
>          return profile;
> ```
@weizhouapache 
While the proposed implementation provides clarity, it's not intuitive. Since changes are already necessary, instead of merely guiding users through an arbitrary process, the aim should be to enhance overall code understanding and user experience.

The `force` flag should be more than enough to address any remaining concerns. This flag serves as a clear indicator that the operation requires careful execution and is not intended for routine use.

Besides, deleting the current `SYSTEM` template does not pose a significant risk to the environment, as it can be reuploaded.

weizhouapache:
> > > > @GaOrtiga @DaanHoogland sorry this PR is not needed in my opinion
> > > 
> > > 
> > > Having a workaround should not be a reason to restrict an operation. If the operator does not know that you can circumvent this restriction they may feel obliged to perform the operation directly in the DB, especially since the exception thrown is very vague and editing the template's type to make it deletable is very non-intuitive.
> > 
> > If the solution is not clear, make it clear then. for example
> > 
> > ```
> > diff --git a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
> > index b886f0868f6..6456ca452ed 100644
> > --- a/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
> > +++ b/server/src/main/java/com/cloud/template/HypervisorTemplateAdapter.java
> > @@ -749,7 +749,7 @@ public class HypervisorTemplateAdapter extends TemplateAdapterBase {
> >          TemplateProfile profile = super.prepareDelete(cmd);
> >          VMTemplateVO template = profile.getTemplate();
> >          if (template.getTemplateType() == TemplateType.SYSTEM) {
> > -            throw new InvalidParameterValueException("The DomR template cannot be deleted.");
> > +            throw new InvalidParameterValueException("The SYSTEM template cannot be deleted. Please update template type to USER and retry.");
> >          }
> >          checkZoneImageStores(profile.getTemplate(), profile.getZoneIdList());
> >          return profile;
> > ```
> @weizhouapache 
> While the proposed implementation provides clarity, it's not intuitive. Since changes are already necessary, instead of merely guiding users through an arbitrary process, the aim should be to enhance overall code understanding and user experience.
> 
Can you tell me how frequent you deleted the SYSTEM template?

> The `force` flag should be more than enough to address any remaining concerns. This flag serves as a clear indicator that the operation requires careful execution and is not intended for routine use.
> 
force flag is being used by other purpose. 

> Besides, deleting the current `SYSTEM` template does not pose a significant risk to the environment, as it can be reuploaded.

That is not true. Without SYSTEM template, users cannot create new networks and recreate VRs/system vms. Do you think if it is a critical issue if it happens on your production?

GaOrtiga:
@weizhouapache 
> Can you tell me how frequent you deleted the SYSTEM template?

Even if it's not something that is done frequently, it being purposely less intuitive impacts user experience.

> That is not true. Without SYSTEM template, users cannot create new networks and recreate VRs/system vms. Do you think if it is a critical issue if it happens on your production?

While it is true that the environment will not function properly without the `SYSTEM` template, it can be easily reuploaded as soon as the operator realizes their mistake.

> force flag is being used by other purpose.

I could add a new parameter, maybe `system` to make it very clear on what's the intent of the flag. This would make it so that these templates can be deleted without the need for workarounds, while adding extra security. This would also make it so that unused templates can be deleted only using the `system` flag, with the exception of the current template being used by the systemvms, which would require the `force` flag aswell as the `system`.

weizhouapache:
> @weizhouapache
> 
> > Can you tell me how frequent you deleted the SYSTEM template?
> 
> Even if it's not something that is done frequently, it being purposely less intuitive impacts user experience.
> 
We need to compare the pros and cons ...
Comparing with possible unexpected removal of SYSTEM template, the 2 steps removal is a very minor issue.

> > That is not true. Without SYSTEM template, users cannot create new networks and recreate VRs/system vms. Do you think if it is a critical issue if it happens on your production?
> 
> While it is true that the environment will not function properly without the `SYSTEM` template, it can be easily reuploaded as soon as the operator realizes their mistake.
> 
Are you operating a production ? If yes, you would not say this ...

> > force flag is being used by other purpose.
> 
> I could add a new parameter, maybe `system` to make it very clear on what's the intent of the flag. This would make it so that these templates can be deleted without the need for workarounds, while adding extra security. This would also make it so that unused templates can be deleted only using the `system` flag, with the exception of the current template being used by the systemvms, which would require the `force` flag aswell as the `system`.

IMHO do not make things too complicated...

weizhouapache:
> 
> I could add a new parameter, maybe `system` to make it very clear on what's the intent of the flag. This would make it so that these templates can be deleted without the need for workarounds, while adding extra security. This would also make it so that unused templates can be deleted only using the `system` flag, with the exception of the current template being used by the systemvms, which would require the `force` flag aswell as the `system`.

Although I think it is unnecessary to add an API parameter which reduces the steps from 2 to 1, for an operation which run once or twice a year, it looks ok to me as it does not lead to any risk. 
If you want to implement it, go ahead. we have spent too much time on this topic.

weizhouapache:
@GaOrtiga 
will you support it on UI ?


API is tested ok
```
(localcloud) 🐱 > delete template id=d0b0a4db-1125-4a99-8cca-cf7934bbc19b
{
  "account": "admin",
  "accountid": "75837c93-aff0-11ee-8ad9-1e00940002db",
  "cmd": "org.apache.cloudstack.api.command.user.template.DeleteTemplateCmd",
  "completed": "2024-01-26T20:25:33+0000",
  "created": "2024-01-26T20:25:33+0000",
  "domainid": "476c9122-aff0-11ee-8ad9-1e00940002db",
  "domainpath": "ROOT",
  "jobid": "b9ba7dd4-58a4-4647-936a-44ca0278532f",
  "jobinstanceid": "d0b0a4db-1125-4a99-8cca-cf7934bbc19b",
  "jobinstancetype": "Template",
  "jobprocstatus": 0,
  "jobresult": {
    "errorcode": 431,
    "errortext": "Could not delete template as it is a SYSTEM template and isSystem is set to false."
  },
  "jobresultcode": 431,
  "jobresulttype": "object",
  "jobstatus": 2,
  "userid": "7584f1d6-aff0-11ee-8ad9-1e00940002db"
}
🙈 Error: async API failed for job b9ba7dd4-58a4-4647-936a-44ca0278532f
(localcloud) 🐱 > delete template id=d0b0a4db-1125-4a99-8cca-cf7934bbc19b issystem=true
{
  "success": true
}
```

api/src/main/java/org/apache/cloudstack/api/command/user/template/DeleteTemplateCmd.java
```diff
@@ -55,6 +55,9 @@ public class DeleteTemplateCmd extends BaseAsyncCmd {
     @Parameter(name = ApiConstants.FORCED, type = CommandType.BOOLEAN, required = false, description = "Force delete a template.", since = "4.9+")
     private Boolean forced;
 
+    @Parameter(name = ApiConstants.IS_SYSTEM, type = CommandType.BOOLEAN, required = false, description = "Necessary if the template's type is system.", since = "4.20.0")
```
weizhouapache, line 58: @GaOrtiga 
can you add `authorized = {RoleType.Admin}` ?

weizhouapache:
> > @GaOrtiga
> > will you support it on UI ?
> 
> @weizhouapache No I will not. In my opinion, adding it to the UI would risk some of the security concerns that you pointed out earlier in the discussion.
> 
> Also, I can not think of a use case that would require it as of now, and if one eventually comes up, we can implement it to the UI when needed.

ok @GaOrtiga 
so admins have to use api or cmk, right ? this does not help admins a lot.

GaOrtiga:
> ok @GaOrtiga so admins have to use api or cmk, right ?

Yes.

> this does not help admins a lot.

Since some security concerns have been brought up, I felt that limiting it to the API would be a good compromise in order to implement this functionality while avoiding further security discussions, since, as you mentioned, we have spent too much time on this. However, I would be happy to implement this in the UI if the community sees it as beneficial.

@weizhouapache @DaanHoogland What are your opinions on this?

weizhouapache:
@blueorangutan package

DaanHoogland:
I think both sides of the discussion have valid points and this will have to prove itself in production. I am ok with the API only implementation.

weizhouapache:
> I think both sides of the discussion have valid points and this will have to prove itself in production. I am ok with the API only implementation.

thanks @DaanHoogland 
I am ok with both. that's say, I am ok with this PR.

rohityadavcloud (commented):
- Would this also deletion of systemvmtemplate current in use?
- If that's the case, can we add a check so it won't allow deleting systemvmtemplate in use
- Does it need some UI changes, so admins can delete systemvmtemplates (not in use) from the UI?

GaOrtiga:
@rohityadavcloud 
> * Would this also deletion of systemvmtemplate current in use?
> * If that's the case, can we add a check so it won't allow deleting systemvmtemplate in use

It will not allow the deletion of the current template unless both the `isSystem` and the `forced` flag are passed. 

> * Does it need some UI changes, so admins can delete systemvmtemplates (not in use) from the UI?

Given some of the previous discussions regarding this implementation, mainly about security, I believe this PR should focus solely on the backend. If, in the future, we decide that UI changes are beneficial, we can make a new PR to implement it.

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

JoaoJandre:
@blueorangutan package

JoaoJandre:
@DaanHoogland @sureshanaparti @rohityadavcloud @shwstppr could we run the CI here?

DaanHoogland:
@blueorangutan test

JoaoJandre:
@blueorangutan package

JoaoJandre (approved):
CLGTM

JoaoJandre:
Merging based on approvals, CI result and manual tests (https://github.com/apache/cloudstack/pull/8556#issuecomment-1912653892)

PR #8639, opened 2024-02-09
Limit `listRoles` API visibility

hsato03:
### Description

When calling the `listRoles` API, users can see roles with more permissions than theirs.

Therefore, the behavior of the `listRoles` API was changed so that users can only see roles that their role has permission to access (roles with same and less permissions).

### Types of changes

- [ ] build/CI

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

1. I created a custom role based on User role and added the `listRoles` API to it.
2. I created an account using the role from step 1 and logged into it.
3. I called the `listRoles` API via CloudMonkey and verfied that the roles with more permissions than mine were not listed, such as default admin roles.

### DIFF
--- a/server/src/main/java/org/apache/cloudstack/acl/RoleManagerImpl.java
+++ b/server/src/main/java/org/apache/cloudstack/acl/RoleManagerImpl.java
@@ -420,15 +420,15 @@ public class RoleManagerImpl extends ManagerBase implements RoleService, Configu
 
     /**
      * Checks if the role of the caller account has compatible permissions of the specified role.
-     * For each permission of the role of the caller, the roleToAccess needs to contain the same permission.
+     * For each permission of the role of the caller, the target role needs to contain the same permission.
      *
-     * @param rolePermissions the permissions of the caller role.
-     * @param roleToAccess the role that the caller role wants to access.
+     * @param sourceRolePermissions the permissions of the caller role.
+     * @param targetRole the role that the caller role wants to access.
      * @return True if the role can be accessed with the given permissions; false otherwise.
      */
-    protected boolean roleHasPermission(Map<String, Permission> rolePermissions, Role roleToAccess) {
+    protected boolean roleHasPermission(Map<String, Permission> sourceRolePermissions, Role targetRole) {
         Set<String> rulesAlreadyCompared = new HashSet<>();
-        for (RolePermission rolePermission : findAllPermissionsBy(roleToAccess.getId())) {
+        for (RolePermission rolePermission : findAllPermissionsBy(targetRole.getId())) {
             boolean permissionIsRegex = rolePermission.getRule().getRuleString().contains("*");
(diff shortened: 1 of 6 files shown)


DaanHoogland (approved):
clgtm and unit-test included, but i think some monkey testing is still in order

server/src/main/java/org/apache/cloudstack/acl/RoleManagerImpl.java
```diff
@@ -382,42 +385,100 @@ public List<Role> findRolesByName(String name) {
...
         return count;
     }
 
+    /**
+     * Checks if the role of the caller account has compatible permissions of the specified role.
+     * For each permission of the role of the caller, the roleToAccess needs to contain the same permission.
+     *
+     * @param rolePermissions the permissions of the caller role.
+     * @param roleToAccess the role that the caller role wants to access.
+     * @return True if the role can be accessed with the given permissions; false otherwise.
```
GutoVeronezi, line 427: @hsato03, what do you think about naming these parameters as `sourceRolePermissions` and `targetRole`? IMO, it seems more intuitive. What do you think?

hsato03: Yes, I agree.

GutoVeronezi (approved):
CLGTM

JoaoJandre:
@blueorangutan package

JoaoJandre:
@DaanHoogland @sureshanaparti @rohityadavcloud @shwstppr could we run the CI here?

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan test

bernardodemarco (approved):
LGTM, I manually tested the PR in a local environment.

1. I created a role called `list-roles-test` based on the `User` role.
2. I added the `listRoles` API to the created role.
3. I created an account called `list-roles-test-account` with the custom role.
4. Through CloudMonkey, I created a profile for the account and I executed the `sync` command.
5. I called the `listRoles` API and I got the following output:
```bash
(test-roles-api) 🐱 > listRoles
{
  "count": 2,
  "role": [
    {
      "description": "Default user role",
      "id": "8cead084-771e-11ee-8e59-5254003754dc",
      "isdefault": true,
      "ispublic": true,
      "name": "User",
      "type": "User"
    },
    {
      "id": "ca565c1b-8589-4a78-b7eb-531d5505cadd",
      "isdefault": false,
      "ispublic": true,
      "name": "list-roles-test",
      "type": "User"
    }
  ]
}
```
6. As it can be noticed, default admin roles and the ones with more permission than the `list-roles-test` role do not appear in the API's response.

PR #8670, opened 2024-02-16
Linstor 4.19 fix script alllines

rp-:
### Description

This PR enabled draining on the `AllLinesParser` so `Script.execute()` doesn't timeout if it has to read larger data.
Right now commands with larger output and `AllLinesParser` won't finish, because noone is emptying the stdout buffer
and the process will never exit until killed by the timeout "interrupt".

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested with local `Script` executions and an added unittest.

### DIFF
--- a/utils/src/test/java/com/cloud/utils/ScriptTest.java
+++ b/utils/src/test/java/com/cloud/utils/ScriptTest.java
@@ -112,6 +112,20 @@ public class ScriptTest {
     }
 
     @Test
+    public void executeWithOutputInterpreterAllLinesParserLargeOutput() {
+        Assume.assumeTrue(SystemUtils.IS_OS_LINUX);
+        OutputInterpreter.AllLinesParser parser = new OutputInterpreter.AllLinesParser();
+        Script script = new Script("seq");
+        script.add("-f");
+        script.add("my text to test cloudstack %g");
+        script.add("4096"); // AllLinesParser doesn't work with that amount of data
+        String value = script.execute(parser);
+        // it is a stack trace in this case as string
+        Assert.assertNull(value);
+        Assert.assertEquals(129965, parser.getLines().length());
+    }
+
+    @Test
     public void runSimpleBashScriptNotExisting() {
         Assume.assumeTrue(SystemUtils.IS_OS_LINUX);
         String output = Script.runSimpleBashScript("/not/existing/scripts/"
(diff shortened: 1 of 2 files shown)


sureshanaparti (approved):
clgtm

utils/src/test/java/com/cloud/utils/ScriptTest.java
```diff
@@ -111,6 +111,20 @@ public String interpret(BufferedReader reader) throws IOException {
...
+        Assume.assumeTrue(SystemUtils.IS_OS_LINUX);
+        OutputInterpreter.AllLinesParser parser = new OutputInterpreter.AllLinesParser();
+        Script script = new Script("seq");
+        script.add("-f");
+        script.add("my text to test cloudstack %g");
+        script.add("4096"); // AllLinesParser doesn't work with that amount of data
+        String value = script.execute(parser);
+        // it is a stack trace in this case as string
+        Assert.assertNull(value);
+        Assert.assertEquals(129965, parser.getLines().length());
```
DaanHoogland, line 125: @rp- how can you be sure of this value?

rp-: well, it will always create the same text? see:
```
seq -f "my text to test cloudstack %g" 4096 | wc -c
129965
```

DaanHoogland: ah, missed the meaning of the command :+1: 
(must say beauty is not this test's purpose ;)

DaanHoogland (approved):

DaanHoogland:
So this must be included to allow for disaster recovery on large scale installations I read.
Why is that?

rp-:
> So this must be included to allow for disaster recovery on large scale installations I read. Why is that?

I use `Script` to read `drbdsetup` command output  to check resource states and if the stdout gets a bit larger (depends on buffer settings of the system), the `script.execute() + AllLinesParser` simply timeout, because nobody is reading the output buffer to EOF.

rp-:
Also wanted to mention this fix is more a must have for 4.19.1, otherwise any slightly larger Linstor setup will break.
Because you can't fully disabled HA: https://github.com/apache/cloudstack/issues/8682

DaanHoogland:
@sureshanaparti I think we can merge this after smoke tests

DaanHoogland:
@blueorangutan package

sureshanaparti:
@blueorangutan test

DaanHoogland:

Issue #8692, opened 2024-02-21
Console Proxy VM: High CPU when using secure VNC / NIOConnection


kohrar:
##### ISSUE TYPE
 * Bug Report

##### COMPONENT NAME
~~~
Console Proxy
~~~

##### CLOUDSTACK VERSION
~~~
4.18, 4.19
~~~

##### CONFIGURATION
Clean install of Apache CloudStack 4.18 and 4.19

##### OS / ENVIRONMENT
Rocky Linux 8.9

##### SUMMARY
When using the secure KVM VNC console proxy in 4.18 or 4.19, the Console Proxy SystemVM uses 100% of its single CPU when handling the connection.

A quick analysis of the java process suggests the high CPU usage is from the [NIOConnection](https://github.com/apache/cloudstack/blob/main/utils/src/main/java/com/cloud/utils/nio/NioConnection.java) components:
![Screenshot from 2024-02-21 11-48-20](https://github.com/apache/cloudstack/assets/4450716/f1f2867a-fe88-4ada-ae8c-a69797896f86)
[NoVncClient getNextBytes](https://github.com/apache/cloudstack/blob/main/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyNoVncClient.java#L125):
![Screenshot from 2024-02-21 11-48-35](https://github.com/apache/cloudstack/assets/4450716/b1d09320-c105-4458-a836-7c2b5076cf05)

The ~100% CPU usage is happening even when the VNC screen is blank (eg. when the screen is off).

In the mean time, I've changed the console proxy's SystemVM compute offering with additional CPUs to reduce CPU contention as we've had some users complain VNC connections were not responding.

If this is expected behavior of NIOConnection, it may be reasonable to increase the CPU allocations in the default console proxy compute offering.

##### STEPS TO REPRODUCE
1. Connect to a KVM VNC session on ACS >4.18.
2. Connect to the console proxy and watch the java process's CPU usage.

##### EXPECTED RESULTS
Reasonable CPU usage when proxying a connection when little data (or screen changes) are happening.

##### ACTUAL RESULTS
I am seeing ~100% CPU usage per active VNC connection even when nothing is happening on the session (eg. a black screen).

kohrar:
[...]

PR #8694, opened 2024-02-23
console proxy client sleep more generic

DaanHoogland:
### Description

This PR...

Fixes: #8692

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyNoVncClient.java
+++ b/services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyNoVncClient.java
@@ -114,11 +114,6 @@ public class ConsoleProxyNoVncClient implements ConsoleProxyClient {
                                 updateFrontEndActivityTime();
                             }
                             connectionAlive = session.isOpen();
-                            try {
-                                Thread.sleep(1);
-                            } catch (InterruptedException e) {
-                                s_logger.error("Error on sleep for vnc over websocket", e);
-                            }
                         } else if (client.isVncOverNioSocket()) {
                             byte[] bytesArr;
                             int nextBytes = client.getNextBytes();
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@weizhouapache @nvazquez , this is purely based on the description in #8692. Any advice on verification?

DaanHoogland:
@blueorangutan package

services/console-proxy/server/src/main/java/com/cloud/consoleproxy/ConsoleProxyNoVncClient.java
```diff
@@ -139,6 +134,11 @@ public void run() {
                                 connectionAlive = false;
                             }
                         }
+                        try {
+                            Thread.sleep(1);
+                        } catch (InterruptedException e) {
+                            s_logger.error("Error on sleep for vnc over websocket", e);
```
weizhouapache, line 140:
```suggestion
                            s_logger.error("Error on sleep for vnc sessions", e);
```

DaanHoogland:

weizhouapache (approved):
code lgtm

kohrar:
A couple questions, if you don't mind:

1. Would it make sense from a latency-perspective to only wait 1ms if there's actually no data coming in? If there's a big chunk of data coming in, you might not want to wait 1ms to read ~200 bytes (which appears to be the maximum the NIO readNextBytes method would return)
2. Why 1ms? Could we raise this to reduce wasted CPU cycles even more? My testing showed the CPU went from 100% to 25%. This might still seem a bit excessive.

DaanHoogland:
> A couple questions, if you don't mind:
> 
>     1. Would it make sense from a latency-perspective to only wait 1ms if there's actually no data coming in? If there's a big chunk of data coming in, you might not want to wait 1ms to read ~200 bytes (which appears to be the maximum the NIO readNextBytes method would return)
> 
>     2. Why 1ms? Could we raise this to reduce wasted CPU cycles even more? My testing showed the CPU went from 100% to 25%. This might still seem a bit excessive.

valid questions @kohrar , sounds like you are arguing for configurability. This is the simplest solution to the issue in your initial ticket. What you are asking now is a bigger (functional) change however. I would prefer to discuss and implement that separately.

rohityadavcloud (approved):

PR #8724, opened 2024-02-29
Support KVM storage implementations controlling logical/physical block io size

mlsorensen:
### Description

This PR allows a given KVM storage implementation to decide if it wants to override the Libvirt disk XML's physical/logical IO size.  For now all in-tree implementations will default to not using this.

In the future it may also be possible to extend the disk offerings to provide more granular control.

Note that using a *logical* sector size other than 512 would require the image to be built to support it, so this may be better suited as a property of the image at some point. Physical sector size can be changed as desired. This at least puts the XML definitions in place as a start.

### Types of changes

- [ ] Major
- [x] Minor

### How Has This Been Tested?
Tested locally by modifying an in-tree plugin, but didn't commit that change to avoid changing upstream behavior.

Also added unit test to test XML generation.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtComputingResource.java
@@ -3058,6 +3058,13 @@ public class LibvirtComputingResource extends ServerResourceBase implements Serv
                     disk.setBusType(DiskDef.DiskBus.SCSI);
                 }
             } else {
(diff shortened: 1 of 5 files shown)


sureshanaparti:
@blueorangutan package

sureshanaparti (approved):
code lgtm

sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan test

sureshanaparti:
Hi @mlsorensen can this be targeted for 4.19.1?

borisstoyanov (approved):
LGTM, based on code review and test results from marvin

PR #8726, opened 2024-02-29
Usage: Fixup entity exists exception

vishesh92:
### Description

While creating helper records, it's possible that an event is processed twice which can result in failure due EntityExistsException. This PR adds a catch to ensure that job doesn't fail if the event is already present in the db.

### Types of changes

- [ ] build/CI

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
+++ b/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
@@ -32,6 +32,7 @@ import java.util.concurrent.TimeUnit;
 
 import javax.inject.Inject;
 import javax.naming.ConfigurationException;
+import javax.persistence.EntityExistsException;
 
 import org.apache.cloudstack.quota.QuotaAlertManager;
 import org.apache.cloudstack.quota.QuotaManager;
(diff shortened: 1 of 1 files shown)


usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -983,34 +984,38 @@ private boolean parseHelperTables(AccountVO account, Date currentStartDate, Date
...
+                createSecurityGroupEvent(event);
+            } else if (isVmSnapshotEvent(eventType)) {
+                handleVMSnapshotEvent(event);
+            } else if (isVmSnapshotOnPrimaryEvent(eventType)) {
+                createVmSnapshotOnPrimaryEvent(event);
+            } else if (isBackupEvent(eventType)) {
+                createBackupEvent(event);
+            }
+        } catch (EntityExistsException e) {
+            s_logger.error(String.format("Failed to create usage event: %d due to %s", event.getId(), e.getMessage()));
```
sureshanaparti, line 1018: can add _eventType_ to log, and error => warn ?

vishesh92:
@blueorangutan package

weizhouapache (approved):
code lgtm

GutoVeronezi:
I have never seem an event being processed twice without manual intervention in the database; instead, I already saw dup events, like `NETWORK.OFFERING.REMOVE`, `NETWORK.OFFERING.ASSIGN`, and `VOLUME.DELETE`, being persisted. Could the situation you are facing be similar to that? If it is the case, it would be insteresting if we could find the root cause.

@vishesh92, do you have the steps to reproduce the error?

sureshanaparti (approved):
clgtm

sureshanaparti:
@blueorangutan test

vishesh92:
> I have never seem an event being processed twice without manual intervention in the database; instead, I already saw dup events, like `NETWORK.OFFERING.REMOVE`, `NETWORK.OFFERING.ASSIGN`, and `VOLUME.DELETE`, being persisted. Could the situation you are facing be similar to that? If it is the case, it would be insteresting if we could find the root cause.
> 
> @vishesh92, do you have the steps to reproduce the error?

@GutoVeronezi I don't have any steps to reproduce the error. We saw the EntityExistsException in the log and it's probably because of duplicate events.

DaanHoogland:
> > I have never seem an event being processed twice without manual intervention in the database; instead, I already saw dup events, like `NETWORK.OFFERING.REMOVE`, `NETWORK.OFFERING.ASSIGN`, and `VOLUME.DELETE`, being persisted. Could the situation you are facing be similar to that? If it is the case, it would be insteresting if we could find the root cause.
> > @vishesh92, do you have the steps to reproduce the error?
> 
> @GutoVeronezi I don't have any steps to reproduce the error. We saw the EntityExistsException in the log and it's probably because of duplicate events.

I think you just described the method of reproducing; duplicate any of the not yet processed events for create/start/register events.

DaanHoogland (approved):
code lgtm

GutoVeronezi:
> > > I have never seem an event being processed twice without manual intervention in the database; instead, I already saw dup events, like `NETWORK.OFFERING.REMOVE`, `NETWORK.OFFERING.ASSIGN`, and `VOLUME.DELETE`, being persisted. Could the situation you are facing be similar to that? If it is the case, it would be insteresting if we could find the root cause.
> > > @vishesh92, do you have the steps to reproduce the error?
> > 
> > 
> > @GutoVeronezi I don't have any steps to reproduce the error. We saw the EntityExistsException in the log and it's probably because of duplicate events.
> 
> I think you just described the method of reproducing; duplicate any of the not yet processed events for create/start/register events.

My point here was to try to identify the workflows causing those duplicated events; until far, I only saw the exception when manually intervening in the database. However, if it an ACS' process that is causing this, it would be good if we could identify and fix it.

DaanHoogland:
> My point here was to try to identify the workflows causing those duplicated events; until far, I only saw the exception when manually intervening in the database. However, if it an ACS' process that is causing this, it would be good if we could identify and fix it.

That is a good point, but we haven't yet found way to reproduce it, It may be due to events being fired once and then handled twice, like you could expect in UDP connections, but nor reproduction scheme yet. This is symptom fighting admittedly.

weizhouapache:
> 
> My point here was to try to identify the workflows causing those duplicated events; until far, I only saw the exception when manually intervening in the database. However, if it an ACS' process that is causing this, it would be good if we could identify and fix it.

ideally yes.

IMHO, sometimes it is difficult to reproduce an issue and find the root cause. Of so, we'd better apply the workaround to avoid subsequent issues.

vishesh92:
@GutoVeronezi I spent some time but couldn't find the reason for duplication of events. Even if we find and fix the root cause, we might not be able to prevent such an occurrence again. So, it would be good to have the check on the usage server itself as well.

GutoVeronezi:
> @GutoVeronezi I spent some time but couldn't find the reason for duplication of events. Even if we find and fix the root cause, we might not be able to prevent such an occurrence again. So, it would be good to have the check on the usage server itself as well.

@vishesh92 
Thanks for the effort; indeed, the warn message is better than the exception. Hope we can find the root cause in the future.

DaanHoogland:
@blueorangutan package

weizhouapache:
@blueorangutan test

PR #8735, opened 2024-03-01
snapshot: don't schedule next snapshot job for a removed volume

rohityadavcloud:
When management server starts, it starts the snapshot scheduler. In case there is a volume snapshot policy which exists for a volume which does not exist, it can cause SQL constraint issue and cause the management server to break from starting its various components and cause HTTP 503 error.

### Types of changes

- [ ] build/CI

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/storage/snapshot/SnapshotSchedulerImpl.java
+++ b/server/src/main/java/com/cloud/storage/snapshot/SnapshotSchedulerImpl.java
@@ -381,6 +381,7 @@ public class SnapshotSchedulerImpl extends ManagerBase implements SnapshotSchedu
 
         if (_volsDao.findById(policy.getVolumeId()) == null) {
             s_logger.warn("Found snapshot policy ID: " + policyId + " for volume ID: " + policy.getVolumeId() + " that does not exist or has been removed");
+            removeSchedule(policy.getVolumeId(), policy.getId());
             return null;
         }
 
--- a/server/src/main/java/com/cloud/storage/snapshot/SnapshotSchedulerImpl.java
+++ b/server/src/main/java/com/cloud/storage/snapshot/SnapshotSchedulerImpl.java
@@ -378,6 +378,12 @@ public class SnapshotSchedulerImpl extends ManagerBase implements SnapshotSchedu
         if (policyId == Snapshot.MANUAL_POLICY_ID) {
             return null;
         }
+
+        if (_volsDao.findById(policy.getVolumeId()) == null) {
+            s_logger.warn("Found snapshot policy ID: " + policyId + " for volume ID: " + policy.getVolumeId() + " that does not exist or has been removed");
+            return null;
+        }
+
         final Date nextSnapshotTimestamp = getNextScheduledTime(policyId, _currentTimestamp);
         SnapshotScheduleVO spstSchedVO = _snapshotScheduleDao.findOneByVolumePolicy(policy.getVolumeId(), policy.getId());
         if (spstSchedVO == null) {


rohityadavcloud:
@blueorangutan package

server/src/main/java/com/cloud/storage/snapshot/SnapshotSchedulerImpl.java
```diff
@@ -378,6 +378,12 @@ public Date scheduleNextSnapshotJob(final SnapshotPolicyVO policy) {
         if (policyId == Snapshot.MANUAL_POLICY_ID) {
             return null;
         }
+
+        if (_volsDao.findById(policy.getVolumeId()) == null) {
+            s_logger.warn("Found snapshot policy ID: " + policyId + " for volume ID: " + policy.getVolumeId() + " that does not exist or has been removed");
+            return null;
+        }
```
DaanHoogland, line 385: great find, should we also delete the snapshotpolicy while we are at it?

DaanHoogland (approved):
clgtm

weizhouapache:
what if the volume is removed manually after snapshot policy is created ?

it may be better to add constraint to the table `snapshot_policy`, to prevent the volume to be removed manually, if there are snapshot policies linked to it.

server/src/main/java/com/cloud/storage/snapshot/SnapshotSchedulerImpl.java
rohityadavcloud: Makes sense

DaanHoogland:
> what if the volume is removed manually after snapshot policy is created ?
> 
> it may be better to add constraint to the table `snapshot_policy`, to prevent the volume to be removed manually, if there are snapshot policies linked to it.

I agree, but this code is still good and anybody could use it in older versions (backport it) I don't think it hurts in any way even with the constraint.

If that works a cascading delete would be usefull in this case ;)

weizhouapache:
> > what if the volume is removed manually after snapshot policy is created ?
> > it may be better to add constraint to the table `snapshot_policy`, to prevent the volume to be removed manually, if there are snapshot policies linked to it.
> 
> I agree, but this code is still good and anybody could use it in older versions (backport it) I don't think it hurts in any way even with the constraint.
> 
> If that works a cascading delete would be usefull in this case ;)

@DaanHoogland 
yes, with PR, the mgmt server will be started without any issue , even if volume is removed manually from database.

@rohityadavcloud 
code lgtm

weizhouapache (approved):
code lgtm

weizhouapache:
@blueorangutan package

rajujith:
> what if the volume is removed manually after snapshot policy is created ?
> 
> it may be better to add constraint to the table `snapshot_policy`, to prevent the volume to be removed manually, if there are snapshot policies linked to it.

Makes sense to add it.

server/src/main/java/com/cloud/storage/snapshot/SnapshotSchedulerImpl.java
rohityadavcloud: Added code to remove schedule when volume is missing

rohityadavcloud:
@weizhouapache I like your idea, could you propose that as a separate PR.

For now I've added changes to remove the schedule when volume is missing. Requesting re-review cc @weizhouapache @DaanHoogland @sureshanaparti @JoaoJandre 

@blueorangutan package

weizhouapache (approved):
code lgtm

weizhouapache:
@blueorangutan test

JoaoJandre (approved):
CLGTM

DaanHoogland (approved):
clgtm

rohityadavcloud:
@JoaoJandre since you're the RM who has called off a freeze, could you merge this?

rohityadavcloud:
This is an edge case not easy to reproduce but found in a prod customer env @JoaoJandre. To reproduce you can delete a volume from the db manually which has a snapshot policy.

JoaoJandre:
> This is an edge case not easy to reproduce but found in a prod customer env @JoaoJandre. To reproduce you can delete a volume from the db manually which has a snapshot policy.

@rohityadavcloud  I was able to test it by deleting the volume on the DB while the MGMT was down. This really is an edge case, but the PR does not introduce any regressions and is adding a layer of safety to the snapshot policy process, so I don't see any issues in merging.

PR #8740, opened 2024-03-04
Add configuration to limit the number of rows deleted from vm_stats

JoaoJandre:
### Description

ACS has a job that periodically removes records from the `vm_stats` table that are older than the retention time (setting `vm.stats.max.retention.time`). In environments that have a large amount of statistics being collected and have a long retention time, if the retention time is shortened, table cleaning may fail to finish due to the huge amount of records being deleted leading to a query timeout.

The `vm.stats.remove.batch.size` configuration has been added to define the batch size of statistics removal queries. This way, instead of trying to remove all records in a single query, ACS will remove records in batches, with the batch size being defined with the `vm.stats.remove.batch.size setting`. The default value is -1, where no limit will be applied; thus, the current behavior is maintained.

### Types of changes

- [ ] build/CI

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

In a local lab, I artificially generated 60 million rows on `vm_stats` and set the retention time to one minute. 

Before the change, the delete job would timeout and no rows would be removed.
After applying the change and setting `vm.stats.max.retention.time` to 100000, the table was successfully cleaned.


framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1208,9 +1208,14 @@ public boolean expunge(final ID id) {
...
 
-    // FIXME: Does not work for joins.
     @Override
     public int expunge(final SearchCriteria<T> sc) {
+        return expunge(sc, -1);
+    }
+
+    // FIXME: Does not work for joins.
+    @Override
+    public int expunge(final SearchCriteria<T> sc, long limit) {
```
vishesh92, line 1218:
```suggestion
    public int expunge(final SearchCriteria<T> sc, final Filter filter) {
```
I would suggest keeping the parameters for this method similar to how it is for other methods for this class.

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
```diff
@@ -1223,6 +1228,11 @@ public int expunge(final SearchCriteria<T> sc) {
             str.append(sc.getWhereClause());
         }
 
+        if (limit > 0) {
```
vishesh92, line 1231: Use `addFilter` method with filter to add limit. This will also allow to add order by in the Delete queries.

server/src/main/java/com/cloud/server/StatsCollector.java
```diff
@@ -290,6 +287,11 @@ public String toString() {
     protected static ConfigKey<Boolean> vmStatsCollectUserVMOnly = new ConfigKey<>("Advanced", Boolean.class, "vm.stats.user.vm.only", "false",
             "When set to 'false' stats for system VMs will be collected otherwise stats collection will be done only for user VMs", true);
 
+    protected static ConfigKey<Long> vmStatsRemoveBatchSize = new ConfigKey<>("Advanced", Long.class, "vm.stats.remove.batch.size", "0", "Indicates the" +
```
vishesh92, line 290: Should we have a generic global setting which we can use for other operations as well?
We can also reuse `detail.batch.query.size` for deletion as well.

JoaoJandre: We could have a generic configuration for other operations, that is a good idea, however, this would incur in a lot more changes that escape this PR's scope. Furthermore, I don't think that we should reuse `detail.batch.query.size` for deletion, it is better to have more granularity on these configurations.

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
JoaoJandre: I'd rather stick with the simpler long limit, as I don't see any need to use the Filter's features on an expunge.

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
JoaoJandre: Could you give an example where we would want to order the deletes? This would only make the deletes even slower.

vishesh92: @JoaoJandre If you add order by and limit on a query, it can be faster if there is an index on the order by column. This depends on the query plan and the data distribution.

Also, while deleting stats, it would be better to remove the older entries first. Otherwise the user might see some missing timestamps if they fetch the data while it's being deleted.

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
vishesh92: Added explanation in the comment below.
https://github.com/apache/cloudstack/pull/8740#discussion_r1513377957

framework/db/src/main/java/com/cloud/utils/db/GenericDaoBase.java
JoaoJandre: I see how the index might be beneficial to SELECTs; however, we're only talking about DELETEs here, do you have any test cases using DELETEs?

FelipeM525 (approved):
CLGTM!

BryanMLima (approved):
CLGTM

JoaoJandre:
@blueorangutan package

JoaoJandre:
@blueorangutan package

JoaoJandre:
@sureshanaparti @DaanHoogland could we run the CI here?

weizhouapache:

JoaoJandre:
@sureshanaparti I think we are ready to merge this one

engine/schema/src/main/java/com/cloud/vm/dao/VmStatsDao.java
```diff
@@ -75,8 +75,9 @@ public interface VmStatsDao extends GenericDao<VmStatsVO, Long> {
     /**
      * Removes (expunges) all VM stats with {@code timestamp} less than
      * a given Date.
-     * @param limit the maximum date to keep stored. Records that exceed this limit will be removed.
+     * @param limitDate the maximum date to keep stored. Records that exceed this limit will be removed.
+     * @param limitPerQuery the maximum amount of rows to be removed in a single query. We loop if there are still rows to be removed after a given query.
      */
-    void removeAllByTimestampLessThan(Date limit);
+    void removeAllByTimestampLessThan(Date limitDate, Long limitPerQuery);
```
weizhouapache, line 81: Long means it could be null
Is _long_ better?

weizhouapache (commented):
@JoaoJandre 
It seems Vishesh had some concerns, not sure if they are addressed

engine/schema/src/main/java/com/cloud/vm/dao/VmStatsDao.java
JoaoJandre: Yes, it should be long

JoaoJandre:
> @JoaoJandre It seems Vishesh had some concerns, not sure if they are addressed

@weizhouapache I've answered @vishesh92's concerns a few months ago, as he did not respond anymore, I thought that he was ok with my reasoning.

vishesh92:
> @JoaoJandre It seems Vishesh had some concerns, not sure if they are addressed

My only concern is increasing number of global settings. I would personally prefer a generic global setting which can be reused for multiple tables instead of a table specific global setting. I don't see any benefit of having multiple global settings per table for deletion.

DaanHoogland:
@vishesh92 , is this something that would block this PR, or can it be added as a future improvement. cc @JoaoJandre .
In any case do you agree on a way forward?

JoaoJandre:
@vishesh92 @DaanHoogland I agree that if we are to have a configuration for each table it would be too much. However, the configuration that @vishesh92 proposed to use (`detail.batch.query.size`) is meant for listing, not removing; thus, I would not use it for the purpose of this PR. 

As this is a broader discussion, I think that we should move this discussion somewhere else to agree on which configurations should exist, then when/if something is decided, we can create a PR to normalize all the required tables. I feel like this is out of scope for this PR, as we would have to change a lot of code that is not related to the `vm_stats` table and metrics deletion, which is the focus for this PR.

vishesh92:
> @vishesh92 @DaanHoogland I agree that if we are to have a configuration for each table it would be too much. However, the configuration that @vishesh92 proposed to use (`detail.batch.query.size`) is meant for listing, not removing; thus, I would not use it for the purpose of this PR.
> 
> As this is a broader discussion, I think that we should move this discussion somewhere else to agree on which configurations should exist, then when/if something is decided, we can create a PR to normalize all the required tables. I feel like this is out of scope for this PR, as we would have to change a lot of code that is not related to the `vm_stats` table and metrics deletion, which is the focus for this PR.

@JoaoJandre Can we rename the global setting to something more generic like `delete.batch.query.size` for deletions across table? If not, what is our way forward when we have a lot of these?

JoaoJandre:
> > @vishesh92 @DaanHoogland I agree that if we are to have a configuration for each table it would be too much. However, the configuration that @vishesh92 proposed to use (`detail.batch.query.size`) is meant for listing, not removing; thus, I would not use it for the purpose of this PR.
> > As this is a broader discussion, I think that we should move this discussion somewhere else to agree on which configurations should exist, then when/if something is decided, we can create a PR to normalize all the required tables. I feel like this is out of scope for this PR, as we would have to change a lot of code that is not related to the `vm_stats` table and metrics deletion, which is the focus for this PR.
> 
> @JoaoJandre The PR looks good. Can we rename the global setting to something more generic like `delete.batch.query.size` for deletions across table? If not, what is our way forward when we have a lot of these?
> 
> Also the Build Github Action is failing due to test failures. Please check.

As I said on my previous message, this PR is focusing only on the `vm_stats` table, as it is the one that caused issues. To create a generic configuration is out of scope for this PR. If we face a situation where we need to create lots of these configurations, then we can discuss a generic configuration, but for now, this is the only one that exists.

Also, the tests should be fixed now :)

JoaoJandre:
@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

DaanHoogland:
@vishesh92 @JoaoJandre can you guys agree? Personally I have a slight preference to start out with as generic settings as we can and speciallise as we go. But I cannot be bothered much.

JoaoJandre:
> @vishesh92 @JoaoJandre can you guys agree? Personally I have a slight preference to start out with as generic settings as we can and speciallise as we go. But I cannot be bothered much.

Creating a generic config would significantly broaden the scope of this PR. 

I could just rename the config that I created; **however**, as I won't touch other tables in this PR, it would be misleading to have a configuration with a generic name that only affects a single table. Thus, I would rather we keep this name, as it is more intuitive and descriptive.

A future PR that wants to do something similar for other tables could change this configuration's name/description to be more generic and use it as well if the author wills it.

PR #8691, opened 2024-02-21
Fix `deleteUser` API to prevent deletion of the caller

lucas-a-martins:
### Description

Currently, ACS allows users to delete themselves by calling the `deleteUser` API via CLI. Fixing this behavior has been previously suggested in this [comment](https://github.com/apache/cloudstack/pull/7242#issuecomment-1433260909), as it leaves room for errors if users don't pay enough attention (e.g., when using the console's autocomplete feature), since the parameter passed is the user's `UUID` and it's not easily identifiable to whom this `UUID` belongs. Furthermore, there's already a feature to prevent a user from deleting themselves via the UI; however, slower environments create a small window where it's possible for the user to click the button before it's disabled, allowing deletion even through the UI (see this [comment](https://github.com/apache/cloudstack/pull/7242#issuecomment-1444300925)).

This pull request addresses this issue by adding a validation that checks the caller's `UUID` and compares it with the `UUID` of the user they are trying to delete. If they are the same, an exception is thrown.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)
- [ ] build/CI

#### Feature/Enhancement Scale

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### Screenshots (if appropriate):

delete-user

### How Has This Been Tested?

I tested by creating a new user (Temp) and attempted to delete it using the same user via CloudMonkey.

An exception was thrown, and the user was not deleted.

Next, I attempted to delete the user via the UI by rapidly clicking the delete icon before being disabled. Once again, an exception was thrown without deleting the user as shown in the screenshot above.

```(Temp) 🐱 > delete user id=02392623-cd84-4695-ba49-ec187742b2cd 
🙈 Error: (HTTP 431, error code 4350) The caller is requesting to delete itself. As a security measure, ACS will not allow this operation. To delete user Temp (ID: 02392623-cd84-4695-ba49-ec187742b2cd, Domain: d57f818f-efe2-44e4-810c-820f048c3129), request to another user with permission to execute the operation. ```

PR #8743, opened 2024-03-04
Fix `deleteAccount` API to prevent deletion of the caller

lucas-a-martins:
### Description

ACS allows users to delete their own account by calling the `deleteAccount` API via CLI. Fixing this behavior has been previously suggested in this [comment](https://github.com/apache/cloudstack/pull/7242#issuecomment-1433260909). Furthermore, there's already a feature to prevent a user from deleting their own account via UI; however, slower environments create a small window where it's possible for the user to click the button before it's disabled, allowing deletion even through the UI (see this [comment](https://github.com/apache/cloudstack/pull/7242#issuecomment-1444300925)).

This pull request addresses this issue by adding a validation that checks the `UUID` of the caller's account and compares it with the `UUID` of the account they are trying to delete. If they are the same, an exception is thrown.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

delete-account

### How Has This Been Tested?

I tested by creating a new account (Temp) and attempted to delete it using a user in this same account via CloudMonkey.
As expected, an exception was thrown.

Next, I attempted to delete the account via the UI by rapidly clicking the delete icon. Once again, an exception was thrown without deleting the account.

```
"errortext": "The caller is requesting to delete their own account. As a security measure, ACS will not allow this operation.To delete account Temp (ID: 510c8c3a-f2ab-4e42-8e52-55ac968b7291, Domain: ab95574f-7857-4a44-9510-2c72ccc770fe), request to another user with permission to execute the operation."
```

### DIFF
--- a/server/src/main/java/com/cloud/user/AccountManagerImpl.java
+++ b/server/src/main/java/com/cloud/user/AccountManagerImpl.java
@@ -1838,7 +1838,6 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
             throw new InvalidParameterValueException("The account id=" + accountId + " manages project(s) with ids " + projectIds + "and can't be removed");
         }
 
-
         CallContext.current().putContextParameter(Account.class, account.getUuid());
 
         return deleteAccount(account, callerUserId, caller);
--- a/server/src/main/java/com/cloud/user/AccountManagerImpl.java
+++ b/server/src/main/java/com/cloud/user/AccountManagerImpl.java
@@ -1815,7 +1815,15 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
         // If the user is a System user, return an error. We do not allow this
         AccountVO account = _accountDao.findById(accountId);
 
-        if (! isDeleteNeeded(account, accountId, caller)) {
+        if (caller.getId() == accountId) {
+            Domain domain = _domainDao.findById(account.getDomainId());
+            throw new InvalidParameterValueException(String.format("The caller is requesting to delete their own account. As a security measure, ACS will not allow this " +
+                            "operation." +
+                            "To delete account %s (ID: %s, Domain: %s), request to another user with permission to execute the operation.",
+                    account.getAccountName(), account.getUuid(), domain.getUuid()));
+        }
+
+        if (!isDeleteNeeded(account, accountId, caller)) {
             return true;
         }
 
@@ -1829,6 +1837,7 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
 
             throw new InvalidParameterValueException("The account id=" + accountId + " manages project(s) with ids " + projectIds + "and can't be removed");
         }
+
 
         CallContext.current().putContextParameter(Account.class, account.getUuid());
 
--- a/server/src/main/java/com/cloud/user/AccountManagerImpl.java
+++ b/server/src/main/java/com/cloud/user/AccountManagerImpl.java
@@ -1817,9 +1817,8 @@ public class AccountManagerImpl extends ManagerBase implements AccountManager, M
 
         if (caller.getId() == accountId) {
             Domain domain = _domainDao.findById(account.getDomainId());
-            throw new InvalidParameterValueException(String.format("The caller is requesting to delete their own account. As a security measure, ACS will not allow this " +
-                            "operation." +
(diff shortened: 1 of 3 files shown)


KlausDornsbach:
clgtm

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -1815,7 +1815,15 @@ public boolean deleteUserAccount(long accountId) {
         // If the user is a System user, return an error. We do not allow this
         AccountVO account = _accountDao.findById(accountId);
 
-        if (! isDeleteNeeded(account, accountId, caller)) {
+        if (caller.getId() == accountId) {
+            Domain domain = _domainDao.findById(account.getDomainId());
+            throw new InvalidParameterValueException(String.format("The caller is requesting to delete their own account. As a security measure, ACS will not allow this " +
```
sureshanaparti, line 1820: @lucas-a-martins May need to rephrase the message, something like
`Deletion of your own account %s (ID: %s, Domain: %s) is not allowed, request another user with permissions to perform the operation.`

lucas-a-martins: Hey @sureshanaparti, thanks for the suggestion. What do you think about this message instead?

`Deletion of your own account is not allowed. To delete account %s (ID: %s, Domain: %s), request to another user with permissions to perform the operation.`

sureshanaparti (approved):
CLGTM

DaanHoogland:
@blueorangutan test

DaanHoogland (approved):
clgtm

weizhouapache:
I think, on many platforms, users can delete their accounts, right?
of course there are some extra confirmation.

lucas-a-martins:
> I think, on many platforms, users can delete their accounts, right? of course there are some extra confirmation.

Hey @weizhouapache,

It's true that a user can delete their accounts on many plataforms, but on these plataforms the user is responsible for creating their account, which usually isn't the case here. In order to create an account, it is necessary to access another account with permission to perform the operation. This should be the same when trying to delete. If the user has the access, they don't need the API to allow deletion of the caller.

Additionally, a single account can be used by multiple users, so I don't think it's right to give every user the power to delete a whole account and, as a consequence, every user within that account. Futhermore, not applying this change could provide a way for the user to circumvent the changes made in the `deleteUser` API that prevent deletion of the caller (PR [#8691](https://github.com/apache/cloudstack/pull/8691)), but with the possibility of deleting multiple users at the same time.

I agree that this kind of flexibility is interesting in scenarios where accounts are independent, being created, used and managed by a single user, however, usually this is not the case in ACS.

lucas-a-martins:
Hey @weizhouapache,

Are there any further concerns regarding this implementation or can we move on this PR?

weizhouapache:
> Hey @weizhouapache,
> 
> Are there any further concerns regarding this implementation or can we move on this PR?

@lucas-a-martins 
sorry I missed your reply.
looks good to me

BryanMLima:
@blueorangutan package

DaanHoogland:
@blueorangutan alma9 kvm-alma9 keepEnv

server/src/test/java/com/cloud/user/AccountManagerImplTest.java
```diff
@@ -198,6 +198,38 @@ public void deleteUserAccountCleanup() {
...
+
+            Mockito.doReturn(accountVoMock).when(callContextMock).getCallingAccount();
+            Mockito.doReturn(accountVoMock).when(_accountDao).findById(Mockito.anyLong());
+            Mockito.doReturn(domainVoMock).when(_domainDao).findById(Mockito.anyLong());
+            Mockito.doReturn(1L).when(accountVoMock).getId();
+
+            accountManagerImpl.deleteUserAccount(accountId);
+        }
+    }
+    @Test
```
bernardodemarco, line 216:
```suggestion
    }

    @Test
```

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan package

JoaoJandre:
@blueorangutan package

DaanHoogland:
@blueorangutan test

PR #8751, opened 2024-03-06
Configuration to disable URL validation when registering templates/ISOs

winterhazel:
### Description

When registering a template/ISO through an URL, the Management Servers always validate if the URL is resolvable before inserting an entry in the database, so that the SSVM downloads it afterwards. However, if the Management Servers do not have access to the Internet, and no internal DNS is used in the environment, it is not possible to register a template/ISO through an URL because this validation fails.

To address this issue, this PR creates the global setting `validate.url.is.resolvable.before.registering.template`, which allows operators to disable this validation. By default, this setting is set to `true` so that the current behavior remains the same.

### Types of changes

- [ ] Major
- [X] Minor

### How Has This Been Tested?

With the setting set to `true`:

1. I tried to register a template using the URL http://dl.openvm.eu/cloudstack/macchinina/x86_64/macchinina-kvm.qcow2.bz2. I verified that the template was successfully registered.
2. I tried to register a template using the URL http://dl.naszsajhdashdjakshd.eu/cloudstack/macchinina/x86_64/macchinina-kvm.qcow2.bz2. I verified that the error `Unable to resolve dl.naszsajhdashdjakshd.eu.` was shown, and that the template was not registered.
3. I tried to register an ISO using the URL http://archlinux.c3sl.ufpr.br/iso/2024.03.01/archlinux-2024.03.01-x86_64.iso. I verified that the ISO was successfully registered.
4. I tried to register an ISO using the URL http://asdazxcxzc.br/iso/2024.03.01/archlinux-2024.03.01-x86_64.iso. I verified that the error `Unable to resolve asdazxcxzc.br.` was shown, and that the template was not registered.

With the setting set to `false`:
1. I tried to register a template using the URL http://dl.openvm.eu/cloudstack/macchinina/x86_64/macchinina-kvm.qcow2.bz2. I verified that the template was successfully registered.
2. I tried to register a template using the URL http://dl.naszsajhdashdjakshd.eu/cloudstack/macchinina/x86_64/macchinina-kvm.qcow2.bz2. I verified that the template was successfully registered.
3. I tried to register an ISO using the URL http://archlinux.c3sl.ufpr.br/iso/2024.03.01/archlinux-2024.03.01-x86_64.iso. I verified that the ISO was successfully registered.
4. I tried to register an ISO using the URL http://asdazxcxzc.br/iso/2024.03.01/archlinux-2024.03.01-x86_64.iso. I verified that the ISO was successfully registered.


winterhazel:
@blueorangutan package

engine/components-api/src/main/java/com/cloud/template/TemplateManager.java
```diff
@@ -51,6 +51,11 @@ public interface TemplateManager {
     static final ConfigKey<Integer> TemplatePreloaderPoolSize = new ConfigKey<Integer>("Advanced", Integer.class, TemplatePreloaderPoolSizeCK, "8",
             "Size of the TemplateManager threadpool", false, ConfigKey.Scope.Global);
 
+    ConfigKey<Boolean> ValidateUrlIsResolvableBeforeRegisteringTemplate = new ConfigKey<>("Advanced", Boolean.class,
+            "validate.url.is.resolvable.before.registering.template", "true", "Indicates whether CloudStack "
+            + "will validate if the provided URL is resolvable during the register of templates/ISOs before persisting them in the database.",
+            true);
```
DaanHoogland, line 57: @winterhazel , this is now a global setting. Does it make sense to norrow the scope? 

i.e. allow it only in specific zones/clusters, like dedicated ones?

DaanHoogland (approved):
clgtm

engine/components-api/src/main/java/com/cloud/template/TemplateManager.java
winterhazel: @DaanHoogland I think it would make sense to have it management server scoped, as its use-case comes from a management server specific configuration; making it zone/cluster scoped does not make sense as the same management server can manage multiple zones/clusters. 

However, even though the `ManagementServer` scope type exists, its implementation has not been made yet. I think we could implement it in a separate PR and change this setting in the future, though.

DaanHoogland:

JoaoJandre (approved):
LGTM. Using a MGMT server without access to the Internet, I added a template for http://dl.openvm.eu/cloudstack/macchinina/x86_64/macchinina-kvm.qcow2.bz2. Before the changes, an error would be thrown as the MGMT did not have a way to reach the server. Now, with the config as true, the template is successfully added and downloaded. If the config is false, the old behavior is maintained.

PR #8782, opened 2024-03-13
Do not retrieve VM's stats on normal VM listing

JoaoJandre:
### Description

For both the `listVirtualMachines` and `listVirtualMachineMetrics` APIs, by default, ACS queries all VM details (the `details` parameter is default `all`) and consequently queries the `cloud.vm_stats` table for each VM, creating a summary of statistics. Therefore, every process carried out in the UI that uses the `listVirtualMachines` API also loads the VM statistics, which can cause slowdowns depending on the number of records.

A new configuration `return.vm.stats.on.vm.list` (with default true), was added, when it is false, the `listVirtualMachines` API will not list the VM's metrics by default.

### Types of changes

- [ ] build/CI

- [X] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Called listVirtualMachines API, without any parameters:
- before this PR, the all the VM's details were returned;
- with this PR and `return.vm.stats.on.vm.list` as true (default), all the details were returned as well;
- with this PR and `return.vm.stats.on.vm.list` as false, all the details except the metrics were returned.

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
@@ -16,9 +16,10 @@
 // under the License.
 package org.apache.cloudstack.api.command.user.vm;
 
-import java.util.ArrayList;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 import org.apache.cloudstack.acl.RoleType;
 import org.apache.cloudstack.affinity.AffinityGroupResponse;
@@ -45,6 +46,7 @@ import org.apache.cloudstack.api.response.UserVmResponse;
 import org.apache.cloudstack.api.response.VpcResponse;
 import org.apache.cloudstack.api.response.ZoneResponse;
 import org.apache.commons.lang3.BooleanUtils;
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.log4j.Logger;
 
 import com.cloud.exception.InvalidParameterValueException;
@@ -58,7 +60,6 @@ import com.cloud.vm.VirtualMachine;
 public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements UserCmd {
     public static final Logger s_logger = Logger.getLogger(ListVMsCmd.class.getName());
 
-    private static final String s_name = "listvirtualmachinesresponse";
 
     /////////////////////////////////////////////////////
     //////////////// API parameters /////////////////////
@@ -98,7 +99,8 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements 
                collectionType = CommandType.STRING,
                description = "comma separated list of vm details requested, "
                    + "value can be a list of [all, group, nics, stats, secgrp, tmpl, servoff, diskoff, backoff, iso, volume, min, affgrp]."
-                   + " If no parameter is passed in, the details will be defaulted to all")
+                   + " If no parameter is passed in, the details will be defaulted to [group, nics, secgrp, tmpl, servoff, diskoff, iso, volume, min, affgrp] for single VM "
+                   + "listing and to [all] for VM listing with metrics")
     private List<String> viewDetails;
 
     @Parameter(name = ApiConstants.TEMPLATE_ID, type = CommandType.UUID, entityType = TemplateResponse.class, description = "list vms by template")
@@ -239,22 +241,28 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements 
         return autoScaleVmGroupId;
     }
 
+    protected boolean isViewDetailsEmpty() {
+        return CollectionUtils.isEmpty(viewDetails);
+    }
+
     public EnumSet<VMDetails> getDetails() throws InvalidParameterValueException {
-        EnumSet<VMDetails> dv;
-        if (viewDetails == null || viewDetails.size() <= 0) {
-            dv = EnumSet.of(VMDetails.all);
-        } else {
-            try {
-                ArrayList<VMDetails> dc = new ArrayList<VMDetails>();
-                for (String detail : viewDetails) {
-                    dc.add(VMDetails.valueOf(detail));
-                }
-                dv = EnumSet.copyOf(dc);
-            } catch (IllegalArgumentException e) {
-                throw new InvalidParameterValueException("The details parameter contains a non permitted value. The allowed values are " + EnumSet.allOf(VMDetails.class));
-            }
+        if (isViewDetailsEmpty()) {
+            Set<VMDetails> allDetails = new HashSet<>(Set.of(VMDetails.values()));
+            allDetails.remove(VMDetails.stats);
+            allDetails.remove(VMDetails.all);
+            return EnumSet.copyOf(allDetails);
         }
-        return dv;
+
+        try {
+            Set<VMDetails> dc = new HashSet<>();
+            for (String detail : viewDetails) {
+                dc.add(VMDetails.valueOf(detail));
+            }
+
+            return EnumSet.copyOf(dc);
+        } catch (IllegalArgumentException e) {
+            throw new InvalidParameterValueException("The details parameter contains a non permitted value. The allowed values are " + EnumSet.allOf(VMDetails.class));
+        }
     }
 
     @Override
@@ -277,10 +285,6 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements 
     /////////////////////////////////////////////////////
     /////////////// API Implementation///////////////////
     /////////////////////////////////////////////////////
-    @Override
-    public String getCommandName() {
-        return s_name;
-    }
 
     @Override
     public ApiCommandResourceType getApiResourceType() {
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
@@ -99,8 +99,7 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements 
                collectionType = CommandType.STRING,
                description = "comma separated list of vm details requested, "
                    + "value can be a list of [all, group, nics, stats, secgrp, tmpl, servoff, diskoff, backoff, iso, volume, min, affgrp]."
-                   + " If no parameter is passed in, the details will be defaulted to [group, nics, secgrp, tmpl, servoff, diskoff, iso, volume, min, affgrp] for single VM "
-                   + "listing and to [all] for VM listing with metrics")
+                   + " If no parameter is passed in, the details will be defaulted to all.")
     private List<String> viewDetails;
 
     @Parameter(name = ApiConstants.TEMPLATE_ID, type = CommandType.UUID, entityType = TemplateResponse.class, description = "list vms by template")
@@ -247,6 +246,10 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements 
 
     public EnumSet<VMDetails> getDetails() throws InvalidParameterValueException {
         if (isViewDetailsEmpty()) {
+            if (_queryService.returnVmStatsOnVmList.value()) {
+                return EnumSet.of(VMDetails.all);
+            }
+
             Set<VMDetails> allDetails = new HashSet<>(Set.of(VMDetails.values()));
             allDetails.remove(VMDetails.stats);
             allDetails.remove(VMDetails.all);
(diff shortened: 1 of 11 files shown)


weizhouapache:
@JoaoJandre 
I suggest to check/fix the API calls from UI to list vms or vm metrics.

It is not good to change the default behavior if it is not a bug, IMHO.

api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
```diff
@@ -98,7 +99,8 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements
                collectionType = CommandType.STRING,
                description = "comma separated list of vm details requested, "
                    + "value can be a list of [all, group, nics, stats, secgrp, tmpl, servoff, diskoff, backoff, iso, volume, min, affgrp]."
-                   + " If no parameter is passed in, the details will be defaulted to all")
+                   + " If no parameter is passed in, the details will be defaulted to [group, nics, secgrp, tmpl, servoff, diskoff, iso, volume, min, affgrp] for single VM "
```
sureshanaparti, line 102: can use the relevant details in ui call to control the response, instead of updating the default behavior ?

JoaoJandre:
@blueorangutan package

JoaoJandre:
@rohityadavcloud , @weizhouapache  I added a configuration to control the behavior of the `listVirtualMachines` API and updated the description. Now the default behavior is maintained. I hope we can advance the discussion on #8970 so we can create a proper mechanism to change the default behavior in the future.

weizhouapache:

weizhouapache:
> @rohityadavcloud , @weizhouapache I added a configuration to control the behavior of the `listVirtualMachines` API and updated the description. Now the default behavior is maintained. I hope we can advance the discussion on #8970 so we can create a proper mechanism to change the default behavior in the future.

thanks @JoaoJandre for the update

rohityadavcloud:
I like the idea of splitting which API is called for metrics vs non-metrics list view (I might or others should steal the pattern for all metrics API usage across the UI). I'm not fully satisfied with the PR @JoaoJandre yet, and I would rather encourage you can pick some ideas from #8985 - that said, I'm also inclined to make progress in a cordial and mature manner. It would be easier to get this merge and optimise the general solution further as required (by me or others).

I'm working on a much wider PR that's in research and progress, that's around wider scalability issues of CloudStack (surprisingly I'm near the root cause, and it maybe possible to even get stats without much penalty). All that said - I wouldn't remember everything I write on Github on each and every PR, take my review with a pinch of salt and I may change my views on things as I'm dealing with a wider scalability problem. I'll leave some comments, but think let's go ahead.

api/src/main/java/org/apache/cloudstack/query/QueryService.java
```diff
@@ -125,6 +125,10 @@ public interface QueryService {
     static final ConfigKey<Boolean> SharePublicTemplatesWithOtherDomains = new ConfigKey<>("Advanced", Boolean.class, "share.public.templates.with.other.domains", "true",
             "If false, templates of this domain will not show up in the list templates of other domains.", true, ConfigKey.Scope.Domain);
 
+    ConfigKey<Boolean> returnVmStatsOnVmList = new ConfigKey<>("Advanced", Boolean.class, "return.vm.stats.on.vm.list", "true",
```
rohityadavcloud, line 128: Minor nit - the variable could be renamed:
 ```suggestion
    ConfigKey<Boolean> ReturnVmStatsOnVmList = new ConfigKey<>("Advanced", Boolean.class, "return.vm.stats.on.vm.list", "true",
```

See if there's a pattern in gobal setting name we should follow?

Comment (rohityadavcloud) [reply api/src/main/java/org/apache/cloudstack/query/QueryService.java:128]: I'm wondering if the default can be set to `false`, if we do this we'll need to add a note in release notes about page - this breaks the default behaviour but given the listVirtualMachines API is widely used by CAPC, terraform, Ansible and what have you - other integrations could be faster. See https://github.com/apache/cloudstack/issues/8975

File: plugins/metrics/src/main/java/org/apache/cloudstack/api/ListVMsMetricsCmd.java [hunk 3]
```diff
@@ -42,20 +44,24 @@
...
-        return APINAME.toLowerCase() + BaseCmd.RESPONSE_SUFFIX;
+    public EnumSet<ApiConstants.VMDetails> getDetails() throws InvalidParameterValueException {
+        if (isViewDetailsEmpty()) {
+            return EnumSet.of(ApiConstants.VMDetails.all);
+        }
+
+        return super.getDetails();
     }
 
+
```
rohityadavcloud, line 64: nit - remove extra line

api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
```diff
@@ -239,22 +240,32 @@ public Long getAutoScaleVmGroupId() {
...
-                }
-                dv = EnumSet.copyOf(dc);
-            } catch (IllegalArgumentException e) {
-                throw new InvalidParameterValueException("The details parameter contains a non permitted value. The allowed values are " + EnumSet.allOf(VMDetails.class));
+        if (isViewDetailsEmpty()) {
+            if (_queryService.returnVmStatsOnVmList.value()) {
+                return EnumSet.of(VMDetails.all);
+            }
+
+            Set<VMDetails> allDetails = new HashSet<>(Set.of(VMDetails.values()));
```
rohityadavcloud, line 253: I think it could be simpler, if the decision whether to list stats or not can be done via `server/src/main/java/com/cloud/api/query/dao/UserVmJoinDaoImpl.java` 
```
-        if (details.contains(VMDetails.all) || details.contains(VMDetails.stats)) {
+        if (details.contains(VMDetails.stats)) {
```

See - https://github.com/apache/cloudstack/pull/8985/files#diff-b180fb2c499ee386bb5b8eae08c2a004e2c33a9430d94b4253bdbd3871d36f8a

api/src/main/java/org/apache/cloudstack/api/response/UserVmResponse.java
```diff
@@ -269,6 +270,10 @@ public class UserVmResponse extends BaseResponseWithTagInformation implements Co
     @Param(description = "the hypervisor on which the template runs")
     private String hypervisor;
 
+    @SerializedName(ApiConstants.IP_ADDRESS)
+    @Param(description = "the VM's primary IP address")
+    private String ipAddress;
```
rohityadavcloud, line 275: Is there a way to avoid this, the ip address table can cause excessive scans in the DB.

plugins/metrics/src/main/java/org/apache/cloudstack/metrics/MetricsServiceImpl.java
```diff
@@ -617,7 +617,6 @@ public List<VmMetricsResponse> listVmMetrics(List<UserVmResponse> vmResponses) {
             }
 
             metricsResponse.setHasAnnotation(vmResponse.hasAnnotation());
-            metricsResponse.setIpAddress(vmResponse.getNics());
```
rohityadavcloud, line 620: Let's keep the metrics response as-is ? See remarks for listVirtualMachines API

ui/src/config/section/compute.js
```diff
@@ -28,7 +28,8 @@ export default {
       title: 'label.instances',
       icon: 'cloud-server-outlined',
       docHelp: 'adminguide/virtual_machines.html',
-      permission: ['listVirtualMachinesMetrics'],
+      permission: ['listVirtualMachines', 'listVirtualMachinesMetrics'],
+      getApiToCall: () => store.getters.metrics ? 'listVirtualMachinesMetrics' : 'listVirtualMachines',
```
rohityadavcloud, line 32: Love this, we can use this across all places where we use metrics.

rohityadavcloud (approved):
Left some remarks, otherwise LGTM

api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
```diff
@@ -98,7 +99,7 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements
                collectionType = CommandType.STRING,
                description = "comma separated list of vm details requested, "
                    + "value can be a list of [all, group, nics, stats, secgrp, tmpl, servoff, diskoff, backoff, iso, volume, min, affgrp]."
-                   + " If no parameter is passed in, the details will be defaulted to all")
+                   + " If no parameter is passed in, the details will be defaulted to all.")
```
rohityadavcloud, line 102: one more - could we mention in the API doc string here, about the global setting that determines whether stats will be sent or not by the list API?

api/src/main/java/org/apache/cloudstack/api/response/UserVmResponse.java
JoaoJandre: I don't see a reason to only return the primary IP on the metrics response, I think it should be returned when listing the user VMs as well. Furthermore, the metrics response extends the UserVmResponse, so when listing the metrics the IP will still be returned.

api/src/main/java/org/apache/cloudstack/query/QueryService.java
JoaoJandre: I would really like to change the default behavior of this and some other APIs. But we currently do not have a proper mechanism to do this. We are slowly starting to discuss such mechanism in https://github.com/apache/cloudstack/discussions/8970#discussion-6554582, and more people discussing and giving ideas would be great.

In any case, since we are already creating a configuration for this, we might as well leave the default behavior as is and let users set this configuration if they feel the need to.

plugins/metrics/src/main/java/org/apache/cloudstack/metrics/MetricsServiceImpl.java
JoaoJandre: The IP address will come from the UserVmResponse, so the metrics response will be left as-is

DaanHoogland:
@sureshanaparti ok by you like this?

DaanHoogland:
@blueorangutan package

DaanHoogland:
ping @sureshanaparti

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

rohityadavcloud (approved):

api/src/main/java/org/apache/cloudstack/query/QueryService.java
```diff
@@ -125,6 +125,10 @@ public interface QueryService {
     static final ConfigKey<Boolean> SharePublicTemplatesWithOtherDomains = new ConfigKey<>("Advanced", Boolean.class, "share.public.templates.with.other.domains", "true",
             "If false, templates of this domain will not show up in the list templates of other domains.", true, ConfigKey.Scope.Domain);
 
+    ConfigKey<Boolean> ReturnVmStatsOnVmList = new ConfigKey<>("Advanced", Boolean.class, "return.vm.stats.on.vm.list", "true",
```
rohityadavcloud, line 128:
```suggestion
    ConfigKey<Boolean> ReturnVmStatsOnVmList = new ConfigKey<>("Advanced", Boolean.class, "return.vm.stats.on.vm.list", "false",
```

After due considerations, I think this can be set to false what originally Joao wants to accomplish. As the listVirtualMachines API is used by at least three integrations I know of (CAPC, k8s-provider and terraform etc.) and several portals are built on top such APIs, it makes sense to trade off speed. BUT, if we do this we must document this change in the release notes and what users can do to get previous/old behaviour if they want it.

What do you think - @JoaoJandre @weizhouapache @DaanHoogland @sureshanaparti @shwstppr ?

weizhouapache: obviously users/other projects will benifit from the change, meanwhile it will break the backward compatibility. 
if we agree more pros than cons, let's go ahead. 
as @rohityadavcloud said, we must document in the release note and let users know (although many users might not be impacted and not be aware of the change).

DaanHoogland (approved):
I am ok with this given a short notice in the release notes. (cc @JoaoJandre @sureshanaparti )

api/src/main/java/org/apache/cloudstack/query/QueryService.java
rohityadavcloud: We add the config for users who may want the backward compatibility. But we use an appropriate default that benefits most users.

rohityadavcloud, line 128:
```suggestion
    ConfigKey<Boolean> ReturnVmStatsOnVmList = new ConfigKey<>("Advanced", Boolean.class, "return.vm.stats.on.vm.list", "false",
```

rohityadavcloud: Looks like @JoaoJandre has disabled to allow maintainers to make changes 🤦 
Let's merge this and address this in subsequent PRs.

api/src/main/java/org/apache/cloudstack/query/QueryService.java
rohityadavcloud: Generally speaking, the current mechanism of deprecation and change in behaviour that we've developed and used in recent years is through release notes documentation.

Issue #8736, opened 2024-03-04
Userdata is not registered as expected


weizhouapache:
Steps to reproduce the issue

- register a UserData with content `haha`
![image](https://github.com/apache/cloudstack/assets/57355700/7990cb9c-d993-4230-883f-ea3c4c7038bb)

- Check the content of the userdata
![image](https://github.com/apache/cloudstack/assets/57355700/03dbaaf9-401e-4483-9874-960a418c138d)

some other information
- on database, the content is not base64 encoded value of `haha`, which seems to be wrong
![image](https://github.com/apache/cloudstack/assets/57355700/5611d489-6dd7-4897-b119-0f10f747956e)

- `haha` is a valid base64 value
```
# echo -n haha |base64 -d
��Z
```

##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
mgmt server
~~~

##### CLOUDSTACK VERSION

~~~
4.19
~~~

##### STEPS TO REPRODUCE

~~~

~~~

##### EXPECTED RESULTS

~~~

~~~

##### ACTUAL RESULTS

~~~

~~~

weizhouapache:
fixed by #8791

PR #8791, opened 2024-03-14
UI: Fix Userdata registration from UI

Pearl1594:
### Description

This PR fixes: https://github.com/apache/cloudstack/issues/8736

### Types of changes

- [ ] build/CI

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

1. When "haha" is passed as userdata:

Userdata is base64 encoded:
image

Response:
image

2. When "haha" is base64 encoded and provided as userdata input:

image

Response:
image

### DIFF
--- a/ui/src/utils/plugins.js
+++ b/ui/src/utils/plugins.js
@@ -482,6 +482,15 @@ export const fileSizeUtilPlugin = {
   }
 }
 
+function isBase64 (str) {
+  try {
+    const decoded = new TextDecoder().decode(Uint8Array.from(atob(str), c => c.charCodeAt(0)))
+    return btoa(decoded) === str
+  } catch (err) {
+    return false
+  }
+}
+
 export const genericUtilPlugin = {
   install (app) {
     app.config.globalProperties.$isValidUuid = function (uuid) {
@@ -490,8 +499,7 @@ export const genericUtilPlugin = {
     }
 
     app.config.globalProperties.$toBase64AndURIEncoded = function (text) {
-      const base64regex = /^([0-9a-zA-Z+/]{4})*(([0-9a-zA-Z+/]{2}==)|([0-9a-zA-Z+/]{3}=))?$/
-      if (base64regex.test(text)) {
+      if (isBase64(text)) {
         return text
       }
       return encodeURIComponent(btoa(unescape(encodeURIComponent(text))))


Pearl1594:
@weizhouapache sorry I didn't quite understand why the gzip in between.
Because without it i.e., `echo "Apache CloudStack" | base64 -w0` if this is provided as input to the userdata param, we don't see the issue. That said,  don't mind adding the checkbox. But I see an issue with decoding it when viewing it in the details view... since its is a gzip encoded string, it attempts to decode it and it would appear as:

image

weizhouapache:
> @weizhouapache sorry I didn't quite understand why the gzip in between. 

refer to https://cloudinit.readthedocs.io/en/latest/explanation/format.html#gzip-compressed-content

Gzip compressed content
Content found to be gzip compressed will be uncompressed. The uncompressed data will then be used as if it were not compressed. This is typically useful because user data is limited to ~16384 [[1]](https://cloudinit.readthedocs.io/en/latest/explanation/format.html#id4) bytes.

> Because without it i.e., `echo "Apache CloudStack" | base64 -w0` if this is provided as input to the userdata param, we don't see the issue. That said, don't mind adding the checkbox. But I see an issue with decoding it when viewing it in the details view... since its is a gzip encoded string, it attempts to decode it and it would appear as:
> 
> image

I think it is OK as the userdata is binary.

moreover, if an user input a base64-encoded string on UI, we do not know if the user wants the string to be the userdata, or decode it as userdata, it would be good to ask the user to tell us.

Pearl1594:
Ok got it, I'll add a checkbox that will only be used by the ui to determine whether to encode the provided userdata or not..

weizhouapache:
> Ok got it, I'll add a checkbox that will only be used by the ui to determine whether to encode the provided userdata or not..

good ,looking forward to it

ui/src/utils/plugins.js
```diff
@@ -490,8 +499,7 @@ export const genericUtilPlugin = {
     }
 
     app.config.globalProperties.$toBase64AndURIEncoded = function (text) {
-      const base64regex = /^([0-9a-zA-Z+/]{4})*(([0-9a-zA-Z+/]{2}==)|([0-9a-zA-Z+/]{3}=))?$/
-      if (base64regex.test(text)) {
+      if (isBase64(text)) {
```
weizhouapache, line 502: @Pearl1594 
thanks for the update.
with your latest change, do we still need this ?
```
      if (isBase64(text)) {
        return text
      }
```

Pearl1594: Do we not want the text that's not encoded to be encoded and saved in the DB .. I thought the idea of the flag was the prevent encoding an already encoded string

weizhouapache: @Pearl1594 
I think the logic can be simple as below
- if checkbox is ticked, pass the origin input
- if checkbox is not ticked, (Base64AndURI) encode the input and pass to API

weizhouapache: oh, wait @Pearl1594 
the method `toBase64AndURIEncoded` is being used in some other files..

```
ui/src/utils/plugins.js:    app.config.globalProperties.$toBase64AndURIEncoded = function (text) {
ui/src/views/compute/CreateAutoScaleVmGroup.vue:          createVmGroupData.userdata = this.$toBase64AndURIEncoded(values.userdata)
ui/src/views/compute/DeployVM.vue:          deployVmData.userdata = this.$toBase64AndURIEncoded(values.userdata)
ui/src/views/compute/DeployVnfAppliance.vue:          createVnfAppData.userdata = this.$toBase64AndURIEncoded(values.userdata)
ui/src/views/compute/EditVM.vue:          params.userdata = this.$toBase64AndURIEncoded(values.userdata)
ui/src/views/compute/RegisterUserData.vue:        params.userdata = this.$toBase64AndURIEncoded(values.userdata)
ui/src/views/compute/ResetUserData.vue:        params.userdata = this.$toBase64AndURIEncoded(values.userdata)
```

Pearl1594: > @Pearl1594 I think the logic can be simple as below
> 
>     * if checkbox is ticked, pass the origin input
> 
>     * if checkbox is not ticked, (Base64AndURI) encode the input and pass to API

That is what is happening in RegisterUserdata component, that we see if it's unchecked and call the encode method.

shwstppr: @Pearl1594 @weizhouapache is there a consensus on changes here?

Pearl1594: I had forgotten about this PR - however, @weizhouapache 's input would be appreciated at this point.

rohityadavcloud (approved):
LGTM - didn't test/check cases

DaanHoogland:
@weizhouapache is this ok by you? or else what still needs to be done ? cc @Pearl1594 ?

weizhouapache:
> @weizhouapache is this ok by you? or else what still needs to be done ? cc @Pearl1594 ?

@DaanHoogland @Pearl1594 

verified ok

- create a user data with "haha" ("Base64 encoded" is checked)
- create a user data with "haha" ("Base64 encoded" is not checked)

it looks good in database and UI

image

weizhouapache:
thanks @Pearl1594 , merging

PR #8834, opened 2024-03-26
Fix NPE when migrating a VM that had its template removed

### LINKED ISSUE #8827: Unable to migrate instance that came from an external KVM host
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Migration to other host
~~~

##### CLOUDSTACK VERSION

~~~
4.19.0
~~~

##### CONFIGURATION

Local storage on VM hosts

##### OS / ENVIRONMENT

Ubuntu 22.04 on all hosts

##### SUMMARY

When trying to migrate an instance to a different host that was migrated from an external KVM host before, it errors out and won't migrate.

##### STEPS TO REPRODUCE

~~~
Migrate a VM from an external KVM host to Cloudstack, and then try to migrate that same instance to a different host.
~~~

##### EXPECTED RESULTS

~~~
That it migrates without issues
~~~

##### ACTUAL RESULTS

~~~
It errors out on "Failed to migrate VM [VM instance {"id":116,"instanceName":"i-8-115-VM","type":"User","uuid":"5056c82d-e436-408f-b4c7-e0be8229a1cd"}] along with its volumes due to [java.lang.NullPointerException]."

Here is the log for this action: https://gist.github.com/pcfriek1987/ba7bb24a8a373e5c2abaa22e802316ec

It seems to have something to do with the template it gets assigned when migrated is my assumption, as all migrated vm's get the: kvm-default-vm-import-dummy-template or VM default import template as listed on the side.  

When trying to migrate it, it says:
Copying template [212] of volume [122] from source storage pool [5] to target storage pool [3].
2024-03-24 00:28:17,788 DEBUG [o.a.c.s.i.TemplateDataFactoryImpl] (Work-Job-Executor-73:ctx-6a8adbf6 job-980/job-981 ctx-23d85860) (logid:61b73463) template 212 is not in store:2, type:Image

Template 212 is the dummy template according to MySQL.

This is kind of an issue since when migrating you cannot specify which server is should go too so instances you want to divide between physical servers you cannot move now. Instances created directly from CS work without issues.
~~~

Rubueno:
### Description

This PR fixes an NPE when a VM had its template removed while the template still has running VMs.

Possibly fixes #8827 but not sure as I've not worked with imported VMs. 

### Types of changes

- [ ] build/CI

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### How Has This Been Tested?

We have this currently running in our 4.16 environment and are bringing the change into a new release. 

It does not break but instead resolves an issue.

### DIFF
--- a/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
+++ b/engine/storage/datamotion/src/main/java/org/apache/cloudstack/storage/motion/StorageSystemDataMotionStrategy.java
@@ -1933,10 +1933,11 @@ public class StorageSystemDataMotionStrategy implements DataMotionStrategy {
                 }
 
                 if (srcVolumeInfo.getTemplateId() != null) {
-                    LOGGER.debug(String.format("Copying template [%s] of volume [%s] from source storage pool [%s] to target storage pool [%s].", srcVolumeInfo.getTemplateId(), srcVolumeInfo.getId(), sourceStoragePool.getId(), destStoragePool.getId()));
-                    copyTemplateToTargetFilesystemStorageIfNeeded(srcVolumeInfo, sourceStoragePool, destDataStore, destStoragePool, destHost);
+                    copyTemplateIfNeeded(destHost, srcVolumeInfo, destDataStore, destStoragePool, sourceStoragePool);
                 } else {
-                    LOGGER.debug(String.format("Skipping copy template from source storage pool [%s] to target storage pool [%s] before migration due to volume [%s] does not have a template.", sourceStoragePool.getId(), destStoragePool.getId(), srcVolumeInfo.getId()));
+                    LOGGER.debug(String.format(
+                            "Skipping copy template from source storage pool [%s] to target storage pool [%s] before migration due to volume [%s] does not have a template.",
+                            sourceStoragePool.getId(), destStoragePool.getId(), srcVolumeInfo.getId()));
                 }
 
                 VolumeVO destVolume = duplicateVolumeOnAnotherStorage(srcVolume, destStoragePool);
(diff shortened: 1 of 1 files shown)


gpordeus:
Hello! I think this hits the same roadblock as #7615, where it ends up migrating without copying the template and it corrupts the VM due to a missing backing file.

 @JoaoJandre can you take a look?

JoaoJandre:
> Hello! I think this hits the same roadblock as #7615, where it ends up migrating without copying the template and it corrupts the VM due to a missing backing file.
> 
> @JoaoJandre can you take a look?

@Rubueno @gpordeus yeah this solution has the same pitfall of #7615. I actually have another solution for this issue (and some others that we have noticed with the VM + volume migration), but I have my hands full this week and will only be able to open the PR(s) next week.

DaanHoogland:
@JoaoJandre how does your #8911 relate to this? are both sensible/needed or do they conflict?

JoaoJandre:
> @JoaoJandre how does your #8911 relate to this? are both sensible/needed or do they conflict?

They do conflict, this PR changes the template copy on live migration. #8911 removes the template copy from the live migration, since #8911 always consolidates the volume, no need to copy the template. 

Furthermore, I believe that this PR will have the same issue as #7615 where the VM can be corrupted.

sureshanaparti:
@blueorangutan package

rohityadavcloud:
@Rubueno can you fix the conflict?

Rubueno:
﻿Currently traveling, won't be back for another two weeks. Should this still be merged even when it results in broken VMs due to missing backing disks according to #7615?On 25 Jun 2024, at 19:55, Rohit Yadav ***@***.***> wrote:﻿
@Rubueno can you fix the conflict?

DaanHoogland:
@Rubueno #7615 is reported to be fixed in #9259. Should this still be merged/is it still relevant?

Rubueno:
I don't think so then. I'll close this.

PR #8887, opened 2024-04-05
libvirtstorage: Make sure netfs storage was really mounted

rp-:
### Description

I encountered a situation, where libvirt/CloudStack believed a nfs/secondary storage was correctly mounted
and the agent wrote images without any problem into it.
It turned out, that it just wrote everything into a a local /mnt/<uuid> directory and later was only found, because
we couldn't create instances(on a other agent) of the just created template.

I don't know yet how the agent/node got into this state, thinking everything was mounted ok, but instead
writing into a local directory.

This should check after creating the mount that it is really mounted and fail if not.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
@@ -272,6 +272,16 @@ public class LibvirtStorageAdaptor implements StorageAdaptor {
         }
     }
 
+    private void checkNetfsStoragePoolMounted(String uuid) {
+        String targetPath = _mountPoint + File.separator + uuid;
+        int mountpointResult = Script.runSimpleBashScriptForExitValue("mountpoint -q " + targetPath);
+        if (mountpointResult != 0) {
+            String errMsg = String.format("libvirt failed to mount storage pool %s at %s", uuid, targetPath);
+            s_logger.error(errMsg);
+            throw new CloudRuntimeException(errMsg);
+        }
+    }
+
     private StoragePool createNetfsStoragePool(PoolType fsType, Connect conn, String uuid, String host, String path) throws LibvirtException {
         String targetPath = _mountPoint + File.separator + uuid;
         LibvirtStoragePoolDef spd = new LibvirtStoragePoolDef(fsType, uuid, uuid, host, path, targetPath);
@@ -692,6 +702,10 @@ public class LibvirtStorageAdaptor implements StorageAdaptor {
                 sp.create(0);
             }
 
+            if (type == StoragePoolType.NetworkFilesystem) {
+                checkNetfsStoragePoolMounted(name);
+            }
+
             return getStoragePool(name);
         } catch (LibvirtException e) {
             String error = e.toString();
@@ -756,10 +770,10 @@ public class LibvirtStorageAdaptor implements StorageAdaptor {
             if (e.toString().contains("exit status 16")) {
                 String targetPath = _mountPoint + File.separator + uuid;
                 s_logger.error("deleteStoragePool removed pool from libvirt, but libvirt had trouble unmounting the pool. Trying umount location " + targetPath +
-                        "again in a few seconds");
+                        " again in a few seconds");
                 String result = Script.runSimpleBashScript("sleep 5 && umount " + targetPath);
                 if (result == null) {
-                    s_logger.error("Succeeded in unmounting " + targetPath);
+                    s_logger.info("Succeeded in unmounting " + targetPath);
                     return true;
                 }
                 s_logger.error("Failed to unmount " + targetPath);


DaanHoogland (approved):
clgtm

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
```diff
@@ -692,6 +702,10 @@ public KVMStoragePool createStoragePool(String name, String host, int port, Stri
                 sp.create(0);
             }
 
+            if (type == StoragePoolType.NetworkFilesystem) {
+                checkNetfsStoragePoolMounted(name);
```
sureshanaparti, line 706: @rp- can check mounted or not after _createNetfsStoragePool()_ call above?

sureshanaparti:
@blueorangutan package

plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
rp-: as far as I understood, the `createNetfsStoragePool()` only defines the libvirt XML storage pool
and it will activated/created/mounted in the `sp.create(0)` call

sureshanaparti: yes, got it.

sureshanaparti (approved):
code lgtm

rp-:
So I got new logs and found the underlying issue.
This PR is probably obsolete, the real problem is that  actions (I have to further analyze) created 2 same storage pool NFS
objects and the 1. one deleted the mount point of the other.

```
rg "(create|remove) storage pool 59eb8ccb-e7bb-3f5e-a0a7-9f7cf7dfadc0" agent.log
305216:2024-04-08 19:46:18,231 INFO  [kvm.storage.LibvirtStorageAdaptor] (agentRequest-Handler-5:null) (logid:0a2b21c6) Attempting to create storage pool 59eb8ccb-e7bb-3f5e-a0a7-9f7cf7dfadc0 (NetworkFilesystem) in libvirt
305645:2024-04-08 19:48:05,184 INFO  [kvm.storage.LibvirtStorageAdaptor] (agentRequest-Handler-2:null) (logid:57cbaf10) Attempting to create storage pool 59eb8ccb-e7bb-3f5e-a0a7-9f7cf7dfadc0 (NetworkFilesystem) in libvirt
305919:2024-04-08 19:49:07,622 INFO  [kvm.storage.LibvirtStorageAdaptor] (agentRequest-Handler-5:null) (logid:0a2b21c6) Attempting to remove storage pool 59eb8ccb-e7bb-3f5e-a0a7-9f7cf7dfadc0 from libvirt
306306:2024-04-08 19:50:46,508 INFO  [kvm.storage.LibvirtStorageAdaptor] (agentRequest-Handler-2:null) (logid:57cbaf10) Attempting to remove storage pool 59eb8ccb-e7bb-3f5e-a0a7-9f7cf7dfadc0 from libvirt
```

@DaanHoogland
Do you know  if there is a policy regarding requesting the same multiple times?
Should this never happen and a safe guard exception in place?

rp-:
Further details for this specific case:

1. Linstor is asked to createAsync Volume from Snapshot
2. Linstor is creating that from the management server call
3. CloudStack sends a CopyCommand to some host (src the just created Linstor resource, target the new template in secondary storage)

If 3. takes some time and you do the same action it happens that the 3. step happens on the same host, but the previous CopyCommand finishes and removes the same NFS secondary storage mount that the 2. CopyCommand would use.

DaanHoogland:
@rp- , I am not sure. I think that the same command running twice should be prevented by the `executeInSequence` flag. As the host is handled by a single agent, this should prevent what I understand from your description.
createAsync is in contradiction with this.
If the same mount on the same mountpoint happening is an issue, it is new to me and it sounds like it must be guarded against in the agent.

any opinion @harikrishna-patnala , @sureshanaparti , @weizhouapache ?

rp-:
I could reproduce this today in our lab and created a real issue #8899 out of it, with DEBUG logs.

It still might be that the Linstor driver is the only  one, who triggers this kind of storagemotionstrategy, but I'm not experienced enough with CloudStack to know for sure.

rp-:
I also think we can close this PR, as it would probably never really trigger anyway.
As the real problem is somewhere else, as I see now.

harikrishna-patnala:
@rp- I see your point about the actual issue with copycommand, though we need to verify with other storages also. Regarding this PR, I feel this check is still valid, may be you can combine this and the fix for the issue #8899

lge:
Would it be acceptable to "mark" a mount point as such, and make it immutable?
The goal is to make it impossible to "accidentally" create stuff inside the mount point that should have been created inside the supposedly mounted file system.
The idea is to place an indicator file into the mount point directory (not the mounted file system), remove all permissions from the mount point (that will be overridden by the actual mount, once that happens), and make it immutable just in case (because root tends to ignore permissions).
If you like it, it should probably be done more generically for all `mkdir`d mount points, and not only for libvirt nfs storage pools, but I don't know my way around cloud stack code good enough yet to find the "proper" location for this.
Suggestion below. I'm not Java native, so consider this "pseudo code" if it is not sufficiently java-ish...

```java
diff --git a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
index 3002fea41d..0afc232828 100644
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/LibvirtStorageAdaptor.java
@@ -32,6 +32,7 @@ import org.apache.cloudstack.utils.qemu.QemuImgException;
 import org.apache.cloudstack.utils.qemu.QemuImgFile;
 import org.apache.cloudstack.utils.qemu.QemuObject;
 import org.apache.commons.codec.binary.Base64;
+import org.apache.commons.lang.SystemUtils;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
 import org.libvirt.Connect;
@@ -276,7 +277,29 @@ public class LibvirtStorageAdaptor implements StorageAdaptor {
     private StoragePool createNetfsStoragePool(PoolType fsType, Connect conn, String uuid, String host, String path) throws LibvirtException {
         String targetPath = _mountPoint + File.separator + uuid;
         LibvirtStoragePoolDef spd = new LibvirtStoragePoolDef(fsType, uuid, uuid, host, path, targetPath);
-        _storageLayer.mkdir(targetPath);
+        if (SystemUtils.IS_OS_LINUX) {
+		// targetPath better not contain funny characters!
+		// maybe s/'/'\''/g first?
+                int createMntPointResult = Script.runSimpleBashScriptForExitValue(
+			// That is a directory already? Good enough.
+                        "! test -L '" + targetPath + "' && test -d '"  + targetPath + "' && exit 0;" +
+                        // we really want that mount point
+                        "mkdir -p '"  + targetPath + "' || exit;" +
+                        // optionally /try/ to mark it as mount point and make it immutable,
+                        // so we won't "accidentally" create stuff inside the mount point
+                        // that should have been created in the mounted file system
+                        "touch '"     + targetPath + "/__MOUNT_POINT_ONLY__'; " +
+                        "chmod 0 '"   + targetPath + "'; " +
+                        "chattr +i '" + targetPath + "'; " +
+                        // mkdir was successful: propagate exit code 0
+                        "exit 0");
+		if (createMntPointResult != 0) {
+                    logger.error("Unable to create mount point " + targetPath);
+		    // What now? Just let Libvirt try itself and fail below.
+		}
+        } else {
+                _storageLayer.mkdir(targetPath);
+        }
         StoragePool sp = null;
         try {
             logger.debug(spd.toString());

```

Is libvirt on non IS_OS_LINUX even a thing?

Possibly check for existence of the directory in a java native way first, or use JNA, if calling the shell looks too heavy.

Bonus points to avoid the race between the first one to "mkdir", someone else winning the race to mount some file system on that mount point, and then the `chattr +i` happening on the top level directory of the mounted file system instead of the mount point ;-)

DaanHoogland:
> Is this good to merge @rp- @sureshanaparti @DaanHoogland @shwstppr @weizhouapache ?

this is purely linstor ; if @rp- says so and smoke tests pass it should be good to go. however, this does not address @rp- 's point about multiple jobs requiring the same mount and umounting it while the other is still busy.

rp-:
As @DaanHoogland said, it doesn't fix the real issue, but somewhat helps to identify it and also fail instead of silently write into a local folder and sometimes "creating templates" will fail...

sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan test

sureshanaparti:
Merging this, based on reviews & test results.

PR #8894, opened 2024-04-08
Fix VR inheriting boot settings from user's VM

hsato03:
### Description

In a situation where there is a stopped user VM and this VM's network VR is also stopped, starting the user VM will also cause the VR to start. However, in this situation, the VR is inheriting the user VM's boot parameters, causing it to not start when the user VM has the UEFI boot type.

Because of this, the inheritance of user VM boot parameters has been removed from VR and other boot parameters have been defined in their place.

Fixes: #8734

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [X] Minor
- [ ] Trivial

### How Has This Been Tested?

I followed the same steps from `How did you try to break this feature and the system with this change?` and verified that the user's VM and VR started correctly.

1. Using VMware, I created an isolated network (`isolated`);
2. I deployed a VM (`test`) using the `isolated` network and started it;
3. After implementing the `isolated` network, I stopped the VM `test` and the VR of the `isolated` network;
4. I added the `UEFI=SECURE` setting to the `test` VM and started it again, causing the `isolated` network VR to start as well;
5. The VR could not start because it had the UEFI boot type.

### DIFF
--- a/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
+++ b/server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
@@ -272,12 +272,15 @@ NetworkMigrationResponder, AggregatedCommandExecutor, RedundantResource, DnsServ
             return false;
         }
 
+        Map<VirtualMachineProfile.Param, Object> params = new HashMap<>();
+        params.put(VirtualMachineProfile.Param.ReProgramGuestNetworks, Boolean.TRUE);
+
         final RouterDeploymentDefinition routerDeploymentDefinition =
                 routerDeploymentDefinitionBuilder.create()
                 .setGuestNetwork(network)
                 .setDeployDestination(dest)
                 .setAccountOwner(_accountMgr.getAccount(network.getAccountId()))
-                .setParams(vm.getParameters())
+                .setParams(params)
                 .build();
 
         final List<DomainRouterVO> routers = routerDeploymentDefinition.deployVirtualRouter();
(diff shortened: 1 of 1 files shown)


hsato03:
I will keep it in draft because I'm still testing it, but it's ready to be reviewed.

DaanHoogland (commented):
code looks good but I wonder if ReProgramGuestNetworks is really the only parameter that needs to be set.

server/src/main/java/com/cloud/network/element/VirtualRouterElement.java
```diff
@@ -272,12 +272,15 @@ public boolean prepare(final Network network, final NicProfile nic, final Virtua
             return false;
         }
 
+        Map<VirtualMachineProfile.Param, Object> params = new HashMap<>();
```
weizhouapache, line 275:
```suggestion
        implement(network, offering, dest, context);
```

weizhouapache: @DaanHoogland 
right. all lines before `return true`

github did not work well when I selected many lines. .

DaanHoogland: i know, to bad, it would help with suggestions like these.

weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan test ubuntu22 kvm-ubuntu22

DaanHoogland:

sureshanaparti (approved):
clgtm

weizhouapache:

weizhouapache (approved):
code lgtm

Issue #8590, opened 2024-02-01
Usage leaves a job record after stopping


DaanHoogland:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Usage server
~~~

##### CLOUDSTACK VERSION

~~~
4.16.1
~~~

##### SUMMARY

After the usage server stops it (sometimes) does not cleanup it's job, which leads to the next instance starting not processing usage events.

##### STEPS TO REPRODUCE

~~~
systemctl stop cloudstack-usage
mysql <auth options> cloud_usage select from usage_jobs where success is NULL;
~~~

##### EXPECTED RESULTS

~~~
no unprocessed jobs remaining
~~~

##### ACTUAL RESULTS

~~~
the job for the last instance of the process remains
~~~

DaanHoogland:
fixed in #8896

PR #8896, opened 2024-04-10
Add a shutdownhook to remove jobs owned by the process

DaanHoogland:
### Description

This PR...

Fixes: #8590
by adding a shutdown hook that removes the last planned usage job on exit, if owned by the current process.

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

- start the usage server
- see the job record being created for the running process
- stop the server
- check that the record is no longer in the DB

### DIFF
--- a/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
+++ b/usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
@@ -316,6 +316,7 @@ public class UsageManagerImpl extends ManagerBase implements UsageManager, Runna
             _sanity = _sanityExecutor.scheduleAtFixedRate(new SanityCheck(), 1, _sanityCheckInterval, TimeUnit.DAYS);
         }
 
+        Runtime.getRuntime().addShutdownHook(new AbbandonJob());
         TransactionLegacy usageTxn = TransactionLegacy.open(TransactionLegacy.USAGE_DB);
         try {
             if (_heartbeatLock.lock(3)) { // 3 second timeout
@@ -345,8 +346,10 @@ public class UsageManagerImpl extends ManagerBase implements UsageManager, Runna
         if (_sanity != null) {
             _sanity.cancel(true);
         }
+
         return true;
     }
+
 
     @Override
     public void run() {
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@blueorangutan package

andrijapanicsb:
If you remove the last record (which, in my head is some kind of protection mechanism, but I might be wrong) - don't we run into risk if the following happens:
- you start 2 usage servers at the same/similar time (systemctl start)
- both usage servers try to run the job (while only one should be the "owner of the job" and the other one will notice this and, currently, NOT execute the duplicate job - it will never execute the duplicate job as long as there is that last record which indicates that another/first usage server is the owner of the job)

?

DaanHoogland:
> If you remove the last record (which, in my head is some kind of protection mechanism, but I might be wrong) - don't we run into risk if the following happens:
> 
>     * you start 2 usage servers at the same/similar time (systemctl start)
> 
>     * both usage servers try to run the job (while only one should be the "owner of the job" and the other one will notice this and, currently, NOT execute the duplicate job - it will never execute the duplicate job as long as there is that last record which indicates that another/first usage server is the owner of the job)
> 
> 
> ?

Yes, this is correct. The issue is when you stop the Usage server owning the job, it will no longer have an owner and no server will ever execute a job anymore. Therefore the owning process should remove their job on exit.

In the current situation, you have to delete the record manually and the risk is the same. There is just a manual action needed that you need to be aware off.

DaanHoogland:

usage/src/main/java/com/cloud/usage/UsageManagerImpl.java
```diff
@@ -2183,4 +2186,17 @@ protected void runInContext() {
             }
         }
     }
+    private class AbbandonJob extends Thread {
```
sureshanaparti, line 2189:
```suggestion
    private class AbandonJob extends Thread {
```

DaanHoogland:
@blueorangutan package

sureshanaparti (approved):
clgtm

vishesh92 (approved):
clgtm
didn't test

vladimirpetrov:
Not sure if this can be avoided but when the usage server is killed (with -9, simulating crash), this machanism no longer works - the old entry stays in the db and a new one (with the new pid) is not created when the server is started again.

DaanHoogland:
@vladimirpetrov I don't think it can but I'll have a google. does it lgty otherwise? We can create an improvement later.

vladimirpetrov (approved):
LGTM based on manual testing

PR #8911, opened 2024-04-12
Linked clone migration between file-based storages on KVM

JoaoJandre:
### Description

Currently, when a volume that was created with linked clone is live migrated from an NFS storage to another NFS storage in KVM, it continues to have the template as a backing file on the destination storage. In all other cases, such as NFS to SharedMountPoint, the volume is consolidated with its backing file during migration.

~This special case adds unnecessary complexity to ACS; In most cases, performing the migration without consolidating the volume does not significantly optimize resource usage. On the other hand, when consolidating the volume with its backing file, the hypervisor detects empty sectors and removes them from the final volume, therefore, consolidating the volume can reduce storage usage.~

Instead of only maintaining linked clone only on NFS to NFS migration, we now maintain it when migrating between NFS, shared mount-point and local storage, as all of these are file based and have no reason to always consolidate the volumes.

In order to fix the situations explained in
https://github.com/apache/cloudstack/pull/7615 and https://github.com/apache/cloudstack/pull/8834, we consolidate the volume when its template is deleted.

### Types of changes

- [ ] build/CI

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
I've tested creating a VM and migrating between NFS/SMP/Local storage, all migrations were successful, the template was copied to the storage and the VM was not corrupted.

I've also tested creating a VM, force deleting its template and then migrating it, the migration consolidated the volume, and the VM was not corrupted.


sureshanaparti:
@blueorangutan package

JoaoJandre:
@blueorangutan package

JoaoJandre:
@sureshanaparti could we run de CI here?

DaanHoogland:
@blueorangutan test

DaanHoogland:
@blueorangutan package

JoaoJandre:
@blueorangutan package

JoaoJandre:
@DaanHoogland @sureshanaparti could we run the CI here?

DaanHoogland:
@blueorangutan test

JoaoJandre:
@blueorangutan package

JoaoJandre:
@DaanHoogland @sureshanaparti could we try again?

gpordeus (commented):
@JoaoJandre The constructor and `Type` enum of `MigrationOptions`, and `StorageSystemDataMotionStrategy.createLinkedCloneMigrationOptions` are now dead code.

`StorageSystemDataMotionStrategy.createFullCloneMigrationOptions` is also redundant.

borisstoyanov (commented):
Hi @JoaoJandre, thanks for this suggestion. 

if we consolidate all link clone created VMs after migration, wouldn't that create a massive overhead in storage as they will be all merged with the template file? Sorry, maybe I'm not understanding this right, but here's my thinking. 

If our storage pool is looking something like this: 
```
{template}{VM1}
--{VM2}
--{VM3}
--{VM4}
```
the physical size of the pool would be `{template} + 4 {vms}`. If we consolidate all we'll endup with `4x template `existing on the storage? or something like this: 
```
{template+VM1}
{template+VM2}
{template+VM3}
{template+VM4}
```
Please let me know if I misunderstood.

JoaoJandre:
> @JoaoJandre The constructor and `Type` enum of `MigrationOptions`, and `StorageSystemDataMotionStrategy.createLinkedCloneMigrationOptions` are now dead code.
> 
> `StorageSystemDataMotionStrategy.createFullCloneMigrationOptions` is also redundant.

Thanks, @gpordeus, I was actually able to cleanup even more code :)

sureshanaparti:
@blueorangutan package

JoaoJandre:
> Hi @JoaoJandre, thanks for this suggestion.
> 
> if we consolidate all link clone created VMs after migration, wouldn't that create a massive overhead in storage as they will be all merged with the template file? Sorry, maybe I'm not understanding this right, but here's my thinking.
> 
> If our storage pool is looking something like this:
> 
> ```
> {template}{VM1}
> --{VM2}
> --{VM3}
> --{VM4}
> ```
> 
> the physical size of the pool would be `{template} + 4 {vms}`. If we consolidate all we'll endup with `4x template `existing on the storage? or something like this:
> 
> ```
> {template+VM1}
> {template+VM2}
> {template+VM3}
> {template+VM4}
> ```
> 
> Please let me know if I misunderstood.

Hey @borisstoyanov, 

Out of context, you are correct, if we create lots of small VMs with big templates and migrate them, there will be lots of copies of the template after migration. However, I find that this is usually not how operators run their clouds. In most cases, operators only migrate VMs after quite some time after creation, enough time that the template might have been completely overwritten, in this case, consolidating the volume would not increase storage usage. 

Furthermore, when using `qemu-img convert`, qemu will optimize storage usage while doing the convert (empty sectors are detected and suppressed from the destination image[^qemu-convert]), so this process might save us storage space. Another point to consider is that for decently sized VMs (500GB+), a template is usually around 1% or 2% of the VM's storage space, almost a rounding error. As one last point, the proposed process is much more simple and less prone to bugs.

[^qemu-convert]:https://qemu-project.gitlab.io/qemu/tools/qemu-img.html

rohityadavcloud:
My 2-cents - we shouldn't break the design; migration to any storage shouldn't break linked-clone VMs. Nor any templates should be allowed to be removed which has any VMs (or VM/volume snapshots...).

borisstoyanov:
Thanks @JoaoJandre, I got your point. I'm not sure about the space saving part, but I do believe the default behaviour should remain a proper link clone implementation, further as you identified it's already consolidating on other storage types, maybe we should look into fixing their cases to utilise link clone instead of consolidating and turning it into de facto full clone. 

to build a consensus, may I suggest we have this as alternative behaviour toggled by a global setting, with the idea that the default behaviour would refer to using link clones with backing files as a user would normally assume. Therefore it will serve both cases, since we don't want to assume how operators run their clouds. 

Regards, 
Bobby.

weizhouapache:
I agree with @borisstoyanov 
full clone will not save space, especially there are many vms having the same base image. this is probably why vms are deployed with linked clone, not full clone.
it should be good to have a global setting to let users to determine if linked clone or full clone the volume during vm migration.  

If linked clone is chosed, it is also feasible (I've done it some years ago)
- copy the base image (template) from the original storage pool to new storage pool, if it is not present on the new pool
- live-migrate the vm
- change the base image of the volume after migration , by "qemu-img rebase -u -b <backingfile>"
- just note the base image might be removed during vm migration by storage cleanup. be sure base image are not removed.

weizhouapache:
> Hi, @rohityadavcloud , @borisstoyanov , @weizhouapache 
> 
> > My 2-cents - we shouldn't break the design; migration to any storage shouldn't break linked-clone VMs. Nor any templates should be allowed to be removed which has any VMs (or VM/volume snapshots...).
> 
> The current design is to consolidate the volume. In every case, except for NFS to NFS migration, the volume is consolidated. 
> 
> > Thanks @JoaoJandre, I got your point. I'm not sure about the space saving part, but I do believe the default behaviour should remain a proper link clone implementation, further as you identified it's already consolidating on other storage types, maybe we should look into fixing their cases to utilise link clone instead of consolidating and turning it into de facto full clone.
> 
> The default behavior is to consolidate, NFS to NFS is the exception, which has no reason to exist. 
> 
> > to build a consensus, may I suggest we have this as alternative behaviour toggled by a global setting, with the idea that the default behaviour would refer to using link clones with backing files as a user would normally assume. Therefore it will serve both cases, since we don't want to assume how operators run their clouds.
> 
> Creating this configuration would be change the default behavior for most cases. Furthermore, I'm against creating a configuration for this behavior as I don't see any real gains behind giving the user this option, 99% of users will not touch this configuration as it doesn't directly impact them in 99% of cases. Moreover, adding this configuration would add even more complexity to the code, whereas this PR is removing complexity.
> 
> 
> > I agree with @borisstoyanov full clone will not save space, especially there are many vms having the same base image. this is probably why vms are deployed with linked clone, not full clone. it should be good to have a global setting to let users to determine if linked clone or full clone the volume during vm migration.
> 
> Bear in mind that this is only changing the migration, not creation of VMs. I agree that creating VMs with linked-clone saves space (and time!!) during creation. However, as I explained, it makes no sense to create VMs and then immediately move them to another storage (as a general procedure); thus, VMs that are migrated have been around for some time, enough that the original template is probably already overwritten. Furthermore, when using `qemu-img convert`, qemu will optimize storage usage while doing the convert (empty sectors are detected and suppressed from the destination image[^qemu-convert]), so this process might save us storage space. 
> 
> 
> > If linked clone is chosed, it is also feasible (I've done it some years ago)
> > 
> >     * copy the base image (template) from the original storage pool to new storage pool, if it is not present on the new pool
> > 
> >     * live-migrate the vm
> > 
> >     * change the base image of the volume after migration , by "qemu-img rebase -u -b "
> > 
> >     * just note the base image might be removed during vm migration by storage cleanup. be sure base image are not removed.
> 
> Finally, if we are to change the behavior for most cases, I'd rather not introduce a configuration and make it so that the VMs are always migrated with linked clone. But again, this is not the current behavior for most cases, and is not beneficial in a lot of situations.
> 
> [^qemu-convert]:https://qemu-project.gitlab.io/qemu/tools/qemu-img.html


If you ensure that linked-clone is used between NFS/local and NFS/local vm migration, it looks good to me that volume is consolidated in other cases. Can you confirm?

JoaoJandre:
@weizhouapache , I don't think I got your point, could you explain it?

weizhouapache:
> @weizhouapache , I don't think I got your point, could you explain it?

@JoaoJandre 
As you described in the description 
`Currently, when a volume that was created with linked clone is live migrated from an NFS storage to another NFS storage in KVM, it continues to have the template as a backing file on the destination storage.`

If it is correct, please ensure it is not changed. (migration between nfs/local are same I think)

JoaoJandre:
> > @weizhouapache , I don't think I got your point, could you explain it?
> 
> @JoaoJandre As you described in the description `Currently, when a volume that was created with linked clone is live migrated from an NFS storage to another NFS storage in KVM, it continues to have the template as a backing file on the destination storage.`
> 
> If it is correct, please ensure it is not changed. (migration between nfs/local are same I think)

@weizhouapache  And for what reasons should we keep this behavior? considering that it is the opposite in all other cases? Is there any real reason to do so?

To me this is an exception that should not exist, CloudStack should have consistent behaviors.

weizhouapache:
> > > @weizhouapache , I don't think I got your point, could you explain it?
> > 
> > @JoaoJandre As you described in the description `Currently, when a volume that was created with linked clone is live migrated from an NFS storage to another NFS storage in KVM, it continues to have the template as a backing file on the destination storage.`
> > 
> > If it is correct, please ensure it is not changed. (migration between nfs/local are same I think)
> 
> @weizhouapache  And for what reasons should we keep this behavior? considering that it is the opposite in all other cases? Is there any real reason to do so?
> 
> To me this is an exception that should not exist, CloudStack should have consistent behaviors.

@JoaoJandre 
The current behavior has been proved to be working well in most cases. Of course there are some issues, but I think it can be fixed.

Your reasons to change the behavior do not convince me. 
We know it is difficult to migrate volume from a storage to another storage type and keep the backing file, due to different image type, difficult access methods, etc. But for nfs/local to nfs/local migration, copying backing file is definitely feasible (from secondary storage or source primary storage).
You said it will save space, do you have data to prove it?
My another concern is the migration time, it would be good to test both methods.

JoaoJandre:
> > > > @weizhouapache , I don't think I got your point, could you explain it?
> > > 
> > > 
> > > @JoaoJandre As you described in the description `Currently, when a volume that was created with linked clone is live migrated from an NFS storage to another NFS storage in KVM, it continues to have the template as a backing file on the destination storage.`
> > > If it is correct, please ensure it is not changed. (migration between nfs/local are same I think)
> > 
> > 
> > @weizhouapache  And for what reasons should we keep this behavior? considering that it is the opposite in all other cases? Is there any real reason to do so?
> > To me this is an exception that should not exist, CloudStack should have consistent behaviors.
> 
> @JoaoJandre The current behavior has been proved to be working well in most cases. Of course there are some issues, but I think it can be fixed. 

The current behavior of consolidating the volume has been proved to work well in all cases seen so far, it is already working for: NFS to SMP; SMP to NFS; Local to NFS; NFS to Local, etc. I prefer a solution that is simpler and works on every case rather than a solution that needs +500 LOC and does not work on every case.

> You said it will save space, do you have data to prove it? 

The quote I posted earlier was directly taken from the qemu-img documentation, here's the quote again, but with more context: "Image conversion is also useful to get smaller image when using a growable format such as qcow: the empty sectors are detected and suppressed from the destination image." (see https://qemu-project.gitlab.io/qemu/tools/qemu-img.html).
This seems like the best source for this, as its directly from the tool documentation.

> My another concern is the migration time, it would be good to test both methods.

Has anyone reported relevant differences in migration between Shared Mount Point to NFS and NFS to NFS? We already have both methods in place; they have been tested for quite some time, and consolidating the volume on migration has never been an issue.

weizhouapache:
> > 
> > @JoaoJandre The current behavior has been proved to be working well in most cases. Of course there are some issues, but I think it can be fixed.
> 
> The current behavior of consolidating the volume has been proved to work well in all cases seen so far, it is already working for: NFS to SMP; SMP to NFS; Local to NFS; NFS to Local, etc. I prefer a solution that is simpler and works on every case rather than a solution that needs +500 LOC and does not work on every case.
> 
> > You said it will save space, do you have data to prove it?
> 
> The quote I posted earlier was directly taken from the qemu-img documentation, here's the quote again, but with more context: "Image conversion is also useful to get smaller image when using a growable format such as qcow: the empty sectors are detected and suppressed from the destination image." (see https://qemu-project.gitlab.io/qemu/tools/qemu-img.html). This seems like the best source for this, as its directly from the tool documentation.
> 
> > My another concern is the migration time, it would be good to test both methods.
> 
> Has anyone reported relevant differences in migration between Shared Mount Point to NFS and NFS to NFS? We already have both methods in place; they have been tested for quite some time, and consolidating the volume on migration has never been an issue.

thanks for the info @JoaoJandre 

I am not sure if
- when vm is migrated to a image with backing file, is the process same as image conversion ? as I know, `qemu-img convert` take very long time if the image is image and consume a lot of cpus. As you remember, ACS uses "qemu-img convert" during volume snapshot or create a template from volume.
- qcow2 is next gen of qcow. is it helpful on qcow2 images ?

we can test some scenarios
- NFS to NFS migration . the dest image has backing file
- NFS to local or SMP, the dest image does not have backing file
we could monitor
- the data to be migrated between storages. 
- the file size and disk size of dest images
- the migration time. it is not that important, due to different storage iops and network bandwidth. 
it is better the vm has some data inside

the data is more convincible than word.

mlsorensen:
I haven't followed this closely, but in principle I don't see a problem with maintaining a linked clone after storage migration. It potentially allows for less data to be copied. There is room for different storage plugins and different storage types to handle this differently depending on source and destination capabilities, and also possibly different behavior based on whether the VM is live storage migrated or offline.

As for whether or not the admin experiences storage savings - that will depend on a lot of things. Does the filesystem deduplicate all of the duplicate OS data such that flattening all of the root volumes into duplicate volumes has no impact? Or do we end up with a bunch of largely duplicate root volumes? Has the user rewritten their entire volume such that hardly anything in the linked clone points to the backing file?

I agree that if an admin is trying to work with the files manually, they would need to do some inspection and be aware that some  root volumes may have linked clones while others don't. However we are already in a somewhat mixed scenario when we have linked root volumes and distinct data volumes.

JoaoJandre:
Hey @mlsorensen, @weizhouapache, @rohityadavcloud, @borisstoyanov

After reading all your comments, doing more testing and pondering on the question a bit more, you changed my mind. I think we should work to maintain linked-clone migration whenever possible; therefore, I propose the following changes:

1. Instead of only maintaining linked clone only on NFS to NFS migration, we maintain it when migrating between NFS, shared mount-point and local storage, as all of these are file based and have no reason to always consolidate the volumes.
2. In order to fix the situations explained in #7615 and #8834, we consolidate the volume when its template is deleted. 

Using the proposed ideas, we expand the general cases where linked clone is maintained (as currently only NFS to NFS maintains it), while fixing the bug with force deleting the template. Do you agree with this approach?

JoaoJandre:
Hey @mlsorensen, @weizhouapache, @rohityadavcloud, @borisstoyanov, @DaanHoogland 

I've pushed the new change, now we keep the VM as linked clone if we are going from/to NFS/SMP/Local storage. We only consolidate the VM if the template does not exist anymore (i.e. force removed). 

I've tested creating a VM and migrating between NFS/SMP/Local storage, all migrations were successful, the template was copied to the storage and the VM was not corrupted.

I've also tested creating a VM, force deleting its template and then migrating it, the migration consolidated the volume, and the VM was not corrupted.

JoaoJandre:
> > Hey @mlsorensen, @weizhouapache, @rohityadavcloud, @borisstoyanov, @DaanHoogland
> > I've pushed the new change, now we keep the VM as linked clone if we are going from/to NFS/SMP/Local storage. We only consolidate the VM if the template does not exist anymore (i.e. force removed).
> 
> thanks @JoaoJandre for the update I have not reviewed and tested the changes yet. the logic looks gooood to me. can you please update the title of this PR ? @JoaoJandre
> 
> > I've tested creating a VM and migrating between NFS/SMP/Local storage, all migrations were successful, the template was copied to the storage and the VM was not corrupted.
> > I've also tested creating a VM, force deleting its template and then migrating it, the migration consolidated the volume, and the VM was not corrupted.

@weizhouapache I've updated the title and description, thanks for the reminder

JoaoJandre:
@blueorangutan package

JoaoJandre:
@blueorangutan package

weizhouapache:
> Could someone take a look ? @mlsorensen, @weizhouapache, @rohityadavcloud, @borisstoyanov, @gpordeus

@JoaoJandre 
code lgtm

is it possible to cover it by smoke test ?

JoaoJandre:
@weizhouapache I would have to access the storage and check if the volume was linked clone before, and if it stayed after, I'm not sure how to do this in a smoke test. I can cover it with unit tests, those are easy.

Btw, could we run the CI here?

rohityadavcloud:
cc @sureshanaparti @DaanHoogland @JoaoJandre @weizhouapache is this considered critical for the release? I think this may be useful for local storage users for example, but not sure what others are considering.

DaanHoogland (approved):
clgtm

DaanHoogland:
@JoaoJandre , I will see what I can do in terms of testing this.

DaanHoogland:
@blueorangutan package

DaanHoogland:
- tested by migrating a vm with storage to another host and its local storage,
- Then stopped the VM and migrated the volume to shared storage and started the VM again

no issues

JoaoJandre:
> cc @sureshanaparti @DaanHoogland @JoaoJandre @weizhouapache is this considered critical for the release? I think this may be useful for local storage users for example, but not sure what others are considering.

@rohityadavcloud I would like to have it merged into 4.20. But we are still missing a review/LGTM. It probably will not make it.

FelipeM525 (approved):
LGTM! I tested this PR in a local environment by live migrating a VM with storage between NFS/Local/SMP as well as force-deleting a template and then migrating it, which worked perfectly fine as the volume was consolidated and the VM did not get corrupted.

weizhouapache:
> LGTM based on manual testing on Oracle Linux 9 with NFS, local and shared mountpoint storages.

great !

weizhouapache:
> @JoaoJandre this has base branch 4.19. It is not really a fix, is it? Will we merge?

@DaanHoogland 
it is a bug fix I think

`
In order to fix the situations explained in
https://github.com/apache/cloudstack/pull/7615 and https://github.com/apache/cloudstack/pull/8834, we consolidate the volume when its template is deleted.
`

DaanHoogland:
> > @JoaoJandre this has base branch 4.19. It is not really a fix, is it? Will we merge?
> 
> @DaanHoogland it is a bug fix I think
> 
> `In order to fix the situations explained in https://github.com/apache/cloudstack/pull/7615 and https://github.com/apache/cloudstack/pull/8834, we consolidate the volume when its template is deleted.`

ok, @weizhouapache just leaving it with the author/RM

JoaoJandre:
> > > @JoaoJandre this has base branch 4.19. It is not really a fix, is it? Will we merge?
> > 
> > 
> > @DaanHoogland it is a bug fix I think
> > `In order to fix the situations explained in https://github.com/apache/cloudstack/pull/7615 and https://github.com/apache/cloudstack/pull/8834, we consolidate the volume when its template is deleted.`
> 
> ok, @weizhouapache just leaving it with the author/RM

@DaanHoogland @weizhouapache I was originally not going to merge this as it was mainly created as a normalization PR. However, **it is fixing a bug** that has been reported in #7615 and #8834. This PR has also been extensively tested (see https://github.com/apache/cloudstack/pull/8911#issuecomment-2337324662, https://github.com/apache/cloudstack/pull/8911#pullrequestreview-2338337009 and https://github.com/apache/cloudstack/pull/8911#pullrequestreview-2343094044) so we should be safe to merge it. Also, the failures in https://github.com/apache/cloudstack/pull/8911#issuecomment-2335028025 do not seem related:

```
2024-09-06 23:06:20,361 - CRITICAL - EXCEPTION: test_02_unsecure_vm_migration: ['Traceback (most recent call last):\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor\n    yield\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 622, in run\n    testMethod()\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1211, in test_02_unsecure_vm_migration\n    vm = self.deploy_vm(src_host)\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1154, in deploy_vm\n    hostid=origin_host.id)\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/lib/base.py", line 691, in create\n    virtual_machine = apiclient.deployVirtualMachine(cmd, method=method)\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackAPI/cloudstackAPIClient.py", line 3981, in deployVirtualMachine\n    response = self.connection.marvinRequest(command, response_type=response, method=method)\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 381, in marvinRequest\n    raise e\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 376, in marvinRequest\n    raise self.__lastError\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 105, in __poll\n    % async_response)\n', 'Exception: Job failed: {accountid : \'f8c22aaf-6c54-11ef-8280-1e009600027b\', account : \'admin\', domainid : \'ca87b0c6-6c54-11ef-8280-1e009600027b\', domainpath : \'ROOT\', userid : \'f8c3bd75-6c54-11ef-8280-1e009600027b\', cmd : \'org.apache.cloudstack.api.command.admin.vm.DeployVMCmdByAdmin\', jobstatus : 2, jobprocstatus : 0, jobresultcode : 530, jobresulttype : \'object\', jobresult : {errorcode : 530, errortext : \'Unable to start a VM [1826588c-f837-4e93-a5e8-696cbee3ab33] due to [Unable to create a deployment for VM instance {"id":366,"instanceName":"i-256-366-VM","type":"User","uuid":"1826588c-f837-4e93-a5e8-696cbee3ab33"}].\'}, jobinstancetype : \'VirtualMachine\', jobinstanceid : \'1826588c-f837-4e93-a5e8-696cbee3ab33\', created : \'2024-09-06T23:03:31+0000\', completed : \'2024-09-06T23:03:42+0000\', jobid : \'1d13a744-a414-43a6-b25e-384ff838c22f\'}\n']
2024-09-07 00:12:15,221 - CRITICAL - EXCEPTION: test_03_secured_to_nonsecured_vm_migration: ['Traceback (most recent call last):\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor\n    yield\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 622, in run\n    testMethod()\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1229, in test_03_secured_to_nonsecured_vm_migration\n    vm = self.deploy_vm(secure_host)\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1154, in deploy_vm\n    hostid=origin_host.id)\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/lib/base.py", line 691, in create\n    virtual_machine = apiclient.deployVirtualMachine(cmd, method=method)\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackAPI/cloudstackAPIClient.py", line 3981, in deployVirtualMachine\n    response = self.connection.marvinRequest(command, response_type=response, method=method)\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 381, in marvinRequest\n    raise e\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 376, in marvinRequest\n    raise self.__lastError\n', '  File "/usr/local/lib/python3.6/site-packages/marvin/cloudstackConnection.py", line 105, in __poll\n    % async_response)\n', 'Exception: Job failed: {accountid : \'f8c22aaf-6c54-11ef-8280-1e009600027b\', account : \'admin\', domainid : \'ca87b0c6-6c54-11ef-8280-1e009600027b\', domainpath : \'ROOT\', userid : \'f8c3bd75-6c54-11ef-8280-1e009600027b\', cmd : \'org.apache.cloudstack.api.command.admin.vm.DeployVMCmdByAdmin\', jobstatus : 2, jobprocstatus : 0, jobresultcode : 530, jobresulttype : \'object\', jobresult : {errorcode : 530, errortext : \'Unable to start a VM [03dfcf15-bd8f-4ace-bdd1-115323402a2a] due to [Unable to create a deployment for VM instance {"id":367,"instanceName":"i-256-367-VM","type":"User","uuid":"03dfcf15-bd8f-4ace-bdd1-115323402a2a"}].\'}, jobinstancetype : \'VirtualMachine\', jobinstanceid : \'03dfcf15-bd8f-4ace-bdd1-115323402a2a\', created : \'2024-09-06T23:09:50+0000\', completed : \'2024-09-07T00:09:53+0000\', jobid : \'eb2b43c4-f75f-4efb-a4d0-741217fb3576\'}\n']
2024-09-07 00:12:15,373 - CRITICAL - EXCEPTION: test_03_secured_to_nonsecured_vm_migration: ['Traceback (most recent call last):\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor\n    yield\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 625, in run\n    self.tearDown()\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1067, in tearDown\n    self.secure_all_hosts()\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1144, in secure_all_hosts\n    self.check_connection(secured=\'true\', host=host)\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1174, in check_connection\n    ", actual is: " + host.details.secured)\n', "Exception: Host detail 'secured' was expected: true, actual is: false\n"]
2024-09-07 00:14:37,744 - CRITICAL - EXCEPTION: test_04_nonsecured_to_secured_vm_migration: ['Traceback (most recent call last):\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 60, in testPartExecutor\n    yield\n', '  File "/usr/lib64/python3.6/unittest/case.py", line 618, in run\n    self.setUp()\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1063, in setUp\n    self.secure_all_hosts()\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1144, in secure_all_hosts\n    self.check_connection(secured=\'true\', host=host)\n', '  File "/marvin/tests/smoke/test_vm_life_cycle.py", line 1174, in check_connection\n    ", actual is: " + host.details.secured)\n', "Exception: Host detail 'secured' was expected: true, actual is: false\n"]
```

PR #8951, opened 2024-04-19
KVM: enable no-mac-spoofing on virtual nics

weizhouapache:
### Description

In shared, isolated networks and vpcs in advanced zone, user vms can easily perform ip/arp/mac spoofing by pretending to be another vm in the same network.

- change mac
```
ip link set dev eth0 down
ip link set dev eth0 address <MAC of another vm>
ip link set dev eth0 up
```
- change IP
```
ip addr add dev eth0 <IP/cidr of another vm in same network>
ip addr del dev eth0 <old IP/cidr>
```

- these can also be done by netplan/network-scripts configurations.

libvirt has a network traffic filtering subsystem which can be used to prevent spoofing. (https://libvirt.org/formatnwfilter.html#concepts)
- It provides some pre-existing network filters (https://libvirt.org/formatnwfilter.html#pre-existing-network-filters), no-mac-spoofing is missed in the table. 
- Ideally we should use `clean-traffic`, however, the `IP/ARP anti-spoofing does not work` in our testing, as the IP is not specified in the libvirt vm definition XML by cloudstack.
-  libvirt can auto-detect the IP but the auto-detected IP is not always correct, for example, when user configures (wrong) static IP inside the user VM.

This PR adds `no-mac-spoofing` for each nic to prevent mac spoofing.

It could be an improvement PR to support all MAC/IP/ARP spoofing
- it might lead to some overhead.
- it is better to be configurable for guest network, network offering or account
- IP/network information needs to be added to the InterfaceDef for each type of interface (bridge, direct, network, vhostuser, etc).
- it might be not needed in advanced zone with security groups, as there are already iptables/ebtables/ipset rules to prevent the spoofings
- it needs some discussion and testing. `clean-traffic` is good, but it might not be what we want. we need to evaluate the pre-existing network filters and probably consider creating customized filters. refer to https://libvirt.org/firewall.html#the-network-filter-driver

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial


weizhouapache:
@NuxRo 
this is probably you want to have. I am not sure if there are downside .

@DaanHoogland and me have tested it. mac anti-spoofing works, but ip anti-spoofing does not work.

weizhouapache:
@blueorangutan package

weizhouapache:

weizhouapache:
> Hey @weizhouapache, can you provide more context to the issue fixed by this PR? Should we add a no arp spoofing as well?

@BryanMLima 
Updated the PR description

wido:
On KVM hypervisors with Security Groups enabled (Advanced + Shared networking) this is already handled by ebtables in security_group.py

Doing this you would do double packet inspection and there might even be a conflict. Have you looked at this?

My suggestion:

- Enable this in Libvirt
- Remove functionality from security_group.py

DaanHoogland (approved):
clgtm

DaanHoogland:
> My suggestion:
> 
>     * Enable this in Libvirt
> 
>     * Remove functionality from security_group.py

meaning, @wido , this is good but we need to add a removal of some of the " -j DROP/ACCEPT" lines from the script? sounds like some precision surgery. Do you know which ones to drop? cc @weizhouapache .

wido:
> > My suggestion:
> > ```
> > * Enable this in Libvirt
> > 
> > * Remove functionality from security_group.py
> > ```
> 
> meaning, @wido , this is good but we need to add a removal of some of the " -j DROP/ACCEPT" lines from the script? sounds like some precision surgery. Do you know which ones to drop? cc @weizhouapache .

It would, I think if you take a look it starts here: https://github.com/apache/cloudstack/blob/8ff2c018cc5b3fc69bcd8756695d04b384e46ab8/scripts/vm/network/security_group.py#L280

- default_ebtables_rules()
- destroy_ebtables_rules()

Those would no longer be needed

weizhouapache:
> > > My suggestion:
> > > ```
> > > * Enable this in Libvirt
> > > 
> > > * Remove functionality from security_group.py
> > > ```
> > 
> > 
> > meaning, @wido , this is good but we need to add a removal of some of the " -j DROP/ACCEPT" lines from the script? sounds like some precision surgery. Do you know which ones to drop? cc @weizhouapache .
> 
> It would, I think if you take a look it starts here:
> 
> https://github.com/apache/cloudstack/blob/8ff2c018cc5b3fc69bcd8756695d04b384e46ab8/scripts/vm/network/security_group.py#L280
> 
> * default_ebtables_rules()
> * destroy_ebtables_rules()
> 
> Those would no longer be needed

@wido 
actually I am thinking of disabling this change for vms with security groups
the script `security_group.py` programs iptables/ebtables rules including the mac/ip/arp anti-spoofing, it has been proved to be working well with both ipv4/ipv6 addresses and one/multiple network nics.
this PR only contains `no-mac-spoofing` which is not good enough to replace the `security_group.py`. it looks like a precise surgery to remove the ebtables rules, as @DaanHoogland said.
we could drop the methods in `security_group.py` if all mac/ip/arp anti-spoofing are supported (see the PR description).

other than that, the upgrade could be an issue as the VMs started in old versions (before upgrade) do not have the configuration in their VM XML definition.

DaanHoogland:
> actually I am thinking of disabling this change for vms with security groups

I second that. It will be simpler and the will not cripple the much security groups implementation.

wido:
> > > > My suggestion:
> > > > ```
> > > > * Enable this in Libvirt
> > > > 
> > > > * Remove functionality from security_group.py
> > > > ```
> > > 
> > > 
> > > meaning, @wido , this is good but we need to add a removal of some of the " -j DROP/ACCEPT" lines from the script? sounds like some precision surgery. Do you know which ones to drop? cc @weizhouapache .
> > 
> > 
> > It would, I think if you take a look it starts here:
> > https://github.com/apache/cloudstack/blob/8ff2c018cc5b3fc69bcd8756695d04b384e46ab8/scripts/vm/network/security_group.py#L280
> > 
> > 
> > * default_ebtables_rules()
> > * destroy_ebtables_rules()
> > 
> > Those would no longer be needed
> 
> @wido actually I am thinking of disabling this change for vms with security groups the script `security_group.py` programs iptables/ebtables rules including the mac/ip/arp anti-spoofing, it has been proved to be working well with both ipv4/ipv6 addresses and one/multiple network nics. this PR only contains `no-mac-spoofing` which is not good enough to replace the `security_group.py`. it looks like a precise surgery to remove the ebtables rules, as @DaanHoogland said. we could drop the methods in `security_group.py` if all mac/ip/arp anti-spoofing are supported (see the PR description).
> 
> other than that, the upgrade could be an issue as the VMs started in old versions (before upgrade) do not have the configuration in their VM XML definition.

Sounds good. I would only add this to VMs without any SG. That would get my approval.

yadvr (approved):
Lgtm didn’t test

NuxRo:
@weizhouapache Good effort.

Like @wido says, the problems this would solve are not an issue in SG zones usually, so indeed we should not apply any of this there.

Otherwise it'd be a nice "win" for operators of regular Advanced Zones to apply anti-spoofing measures. We already have something somewhat similar for VMWare.

I'd be happy to use all reasonable libvirt nwfilter features, make them options in Network Offering:
- IP anti-spoofing (with or without auto-detect)
- ARP anti-spoofing
- MAC anti-spoofing

Would it even be reasonable to allow the operator to specify more nwfilter? Ie load whatever xml file from /usr/share/libvirt/nwfilter/ that they want?

NuxRo (approved):
LGTM, but taking into consideration the comments, particularly re security groups.

rohityadavcloud:
@weizhouapache when you've bandwidth can you check why the github actions are/were failing (I've rekicked them now)

rajujith:
@weizhouapache Since this is for the 4.22.1 release, could you retarget the PR to the 4.22 branch?

EduFrazao:
This is a awesome feature.
I use advanced networks, without Security Groups (don't want complex rule inspection if possible on hypervisors).
I have some edge cases, where users want to bring their on firewall solution and do not want to use a isolated network with a virtual router.
In this case, a shared network is very good, but without mac+ip anti-spoof its very hard to acomplish.

wido (approved):
For me this is CLGTM! Code looks good, this is tested behavior of Libvirt. Nothing special, but good addition.

EduFrazao:
> For me this is CLGTM! Code looks good, this is tested behavior of Libvirt. Nothing special, but good addition.

Yes. Its a very nice feature. For me, the ideal feature would be clean-traffic
I'm working on on a groovy hook that would implement it.
I can't allow users to change their IP's in some networks.

DaanHoogland:
@weizhouapache what work does this still need?

weizhouapache:
> @weizhouapache what work does this still need?

@DaanHoogland 
it would be better if this only applies on networks without security groups

DaanHoogland:
> > @weizhouapache what work does this still need?
> 
> @DaanHoogland it would be better if this only applies on networks without security groups

that sounds like redesign? i.e. a new command/answer set, or ate least adding a parameter?

weizhouapache:
> > > @weizhouapache what work does this still need?
> > 
> > 
> > @DaanHoogland it would be better if this only applies on networks without security groups
> 
> that sounds like redesign? i.e. a new command/answer set, or ate least adding a parameter?

yes, add more info in some commands (did not check yet)

PR #8952, opened 2024-04-19
Fix KVM live migration with NFS volumes

JoaoJandre:
### Description

Using KVM as the hypervisor, when migrating a VM that has volumes on NFS storage, if any volume of it is migrated, all volumes that are on NFS must be migrated, otherwise, the migration will fail with the error `unable to execute QEMU command 'block- export-add': Block node is read-only`. 

This situation occurs because, during the migration of VMs with volumes, the ACS does not inform which volumes should be migrated, causing Libvirt to assume that they will all be migrated, even if their path does not change, leading the hypervisor to try to overwrite the volumes of the VM during migration, resulting in the aforementioned error.

This PR changes the migration API used so that ACS informs which volumes should be migrated.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

A VM with 3 volumes on NFS storage was created and used for the tests below.

Before applying the changes, when migrating the VM and any of the volumes alone, the error `unable to execute QEMU command 'block-export-add': Block node is read-only` occurred.

After applying the changes, it was possible to migrate both the root volume alone, and migrate just one of the datadisks. After each migration, a dump of the VM XML was performed and verified that only the correct volumes were migrated.

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/LibvirtVMDef.java
@@ -972,6 +972,10 @@ public class LibvirtVMDef {
             return _diskLabel;
         }
 
+        public void setDiskLabel(String label) {
+            _diskLabel = label;
+        }
+
         public DiskType getDiskType() {
             return _diskType;
         }
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/resource/MigrateKVMAsync.java
@@ -18,13 +18,20 @@
  */
 package com.cloud.hypervisor.kvm.resource;
 
+import java.util.Iterator;
+import java.util.Set;
 import java.util.concurrent.Callable;
 
+import org.apache.log4j.Logger;
 import org.libvirt.Connect;
 import org.libvirt.Domain;
 import org.libvirt.LibvirtException;
+import org.libvirt.TypedParameter;
+import org.libvirt.TypedStringParameter;
+import org.libvirt.TypedUlongParameter;
 
 public class MigrateKVMAsync implements Callable<Domain> {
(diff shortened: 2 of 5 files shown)


weizhouapache:
@blueorangutan package

weizhouapache:
@blueorangutan package

weizhouapache:

weizhouapache (approved):
lgtm
I have reproduced the issue and verified the issue is fixed by this PR

harikrishna-patnala (approved):
code LGTM

weizhouapache:
@JoaoJandre 
will you back to 4.18 ? 
(I am not sure if there will be 4.18.3.0)

DaanHoogland:
> @JoaoJandre will you back to 4.18 ? (I am not sure if there will be 4.18.3.0)

@JoaoJandre , we can merge as is or merge to 4.18 and merge forward. For anybody doing custom maintenance that will be convenient. Just say yes or no... ;)

JoaoJandre:
> > > > @JoaoJandre will you back to 4.18 ? (I am not sure if there will be 4.18.3.0)
> > > 
> > > 
> > > @JoaoJandre , we can merge as is or merge to 4.18 and merge forward. For anybody doing custom maintenance that will be convenient. Just say yes or no... ;)
> > 
> > 
> > @DaanHoogland @weizhouapache I don't think we'll have a 4.18.3.0, but I've rebased anyway.
> 
> @JoaoJandre thanks for rebasing
> 
> although it worked in my testing, I wonder if `LinkedHashSet` is better than `HashSet`

@weizhouapache A `LinkedHashSet` is an ordered version of `HashSet`, considering that we don't care for order in this case, the `HashSet` should be better (as we won't have the ordering overhead).

weizhouapache:
> @weizhouapache A `LinkedHashSet` is an ordered version of `HashSet`, considering that we don't care for order in this case, the `HashSet` should be better (as we won't have the ordering overhead). 

ok makes sense

DaanHoogland:
@blueorangutan package

sureshanaparti (approved):
clgtm

DaanHoogland:

PR #8976, opened 2024-04-25
Fix null host tags while creating a zone

GaOrtiga:
### Description

In the UI, during zone creation, if a new host is created without specifying host tags, ACS adds the `null` tag to it automatically.

This behaviour has been fixed, making it so that no host tags are added in this situation 

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [X] Trivial

### How Has This Been Tested?

I created a zone through the UI without specifying host tags and verified that no tags were added.

I also created while specifying host tags and verified that they were added accordingly

### DIFF
--- a/ui/src/views/infra/zone/ZoneWizardLaunchZone.vue
+++ b/ui/src/views/infra/zone/ZoneWizardLaunchZone.vue
@@ -1358,7 +1358,7 @@ export default {
       hostData.clusterid = this.stepData.clusterReturned.id
       hostData.hypervisor = this.stepData.clusterReturned.hypervisortype
       hostData.clustertype = this.stepData.clusterReturned.clustertype
-      hostData.hosttags = this.prefillContent?.hostTags || null
+      hostData.hosttags = this.prefillContent?.hostTags || ''
       hostData.username = this.prefillContent?.hostUserName || null
       hostData.password = hostPassword
       const hostname = this.prefillContent?.hostName || null


DaanHoogland:
@blueorangutan ui

weizhouapache:
good catch @GaOrtiga 

I wonder if it is better to consider all the null params
```
 
+        Object.keys(args).forEach((key) => (args[key] == null) && delete args[key])
         api('addHost', {}, 'POST', args).then(json => {
```

GaOrtiga:
> good catch @GaOrtiga
> 
> I wonder if it is better to consider all the null params
> 
> ```
>  
> +        Object.keys(args).forEach((key) => (args[key] == null) && delete args[key])
>          api('addHost', {}, 'POST', args).then(json => {
> ```

I do not think this would be necessary, `host tags` is the only argument on the `addhost` section of this form that is not required, therefore it is the only one that can be null.

weizhouapache:
> > good catch @GaOrtiga
> > I wonder if it is better to consider all the null params
> > ```
> >  
> > +        Object.keys(args).forEach((key) => (args[key] == null) && delete args[key])
> >          api('addHost', {}, 'POST', args).then(json => {
> > ```
> 
> I do not think this would be necessary, `host tags` is the only argument on the `addhost` section of this form that is not required, therefore it is the only one that can be null.

that's ok, thanks @GaOrtiga

DaanHoogland (approved):
clgtm

borisstoyanov:
@blueorangutan package

borisstoyanov (approved):
LGTM, manually deployed a zone and there was no tag on the host

PR #8985, opened 2024-04-26
api: improve listVM API handling and response speed

rohityadavcloud:
- Changes behaviour of details param handling for:
  - listVirtualMachines API: when the detail param is not provided, it uses `all` details except `stats`
  - listVirtualMachinesMetrics API: when the detail param is not provided, it uses `all` details including `stats`
- Remove ConfigKey vm.stats.increment.metrics.in.memory which was renamed to `vm.stats.increment.metrics` in #5984
- Changes default value of VM stats accumulation setting `vm.stats.increment.metrics` to false until a better solution emerges. Since #5984, this is true and during the execution of listVM APIs the stats are clubbed/calculated which can immensely slow down list VM API calls.
- Fix UI that uses listVirtualMachinesMetrics to not call `stats` detail when in list view without metrics selected.

These changes see 2-4x gains in the listVirtualMachines APIs call and in the UI. For environment where code changes are not possible, disabling `vm.stats.increment.metrics` global setting saw 2-4x speed gain in the list API calls.

### Types of changes

- [ ] build/CI

- [x] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [x] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Tested a workaround in a 4.18.2.0 env KVM based, with only 12 VMs:

1. List VM API call without details and with details but global setting `vm.stats.increment.metrics` disabled:

Screenshot 2024-04-26 at 11 29 31 AM

2. List VM API call without detail and with detail and `vm.stats.increment.metrics` enabled:
Screenshot 2024-04-26 at 11 31 15 AM

- We see ~3-4x gains by simply disabling the global setting
- We see 15x gains when stats details isn't used but the global setting is enabled
- We see 4x gains when stats detail isn't used and global setting is disabled

With this PR, I propose to change the earlier behaviour for listVirtualMachines API to exclude stats when details aren't provided. This can immensely speed up several integrations, who call listVMs APIs that not necessarily need `stats` in it.
While, the listVirtualMachinesMetrics API was amended to retain old behaviour so when details aren't passed it could
return all details including stats as expected. In the release notes likely we can mention about this, so folks who don't want accumulated stats can disable it but for new env we don't want to encourage this by default (as we already have the nice vm history metrics feature/tab).


rohityadavcloud:
@blueorangutan package

weizhouapache (commented):
code looks ok

my only concern is the backward compatibility
we may need a highlighted section to list breaking changes

rohityadavcloud:
@weizhouapache this is breaking largely wrt 4.17 but not much as earlier versions; the only issue is for listVirtualMachines API which IMO shouldn't return vmstats (by looping through accumulating stats while the API is made). But I agree it could be an edge case where this break somebody's integration who aren't using the listVirtualMachinesMetrics instead (which doesn't break the compatibility).

rohityadavcloud:
@blueorangutan test

JoaoJandre:
Hello @rohityadavcloud , @weizhouapache  

Looking briefly at this PR I noticed that it has a big intersection with #8782, especially the changes on the listVirtualMachines and listVirtualMachineMetrics APIs.

When #8782 proposed to change the default behavior of them to not list the stats, both @weizhouapache  and @rohityadavcloud  were firmly against changing the default behavior (see https://github.com/apache/cloudstack/pull/8782#issuecomment-1995297443 and https://github.com/apache/cloudstack/pull/8782#issuecomment-1997341885). So now, only a month later, we get this PR: what is the big difference on these two PRs? And if there is, why not just propose an adjustment to #8782? 

In any case, this PR not only breaks compatibility on the APIs, but also changes the default value of another configuration; and all this is going into 4.18! Are we going to introduce breaking changes on a maintenance version? I would really like to understand what made you change your opinion on compatibility so fast.

sureshanaparti (approved):
clgtm

rohityadavcloud:
Thanks @weizhouapache @JoaoJandre I've addressed the feedback and comments.

Joao's remarks are valid, however, my comments at the time were largely around metrics vs non-metrics APIs - how they should be used treated. At the time I couldn't co-relate what made the API slow only to later find out through community and customer issues who are facing degraded performance.

Generally speaking, we shouldn't break API compatibility which can break potential integrations (as also on this PR Wei has objected). I feel in this case whatever we do we break what was introduced by #5984  in 4.17. The root cause that I could determine is, we accumulate VM stats for each VM in list API calls which isn't a good idea. That said, I think we need a stop-gap until a better solution emerges on how we should process and return stats are returned by the list API (I think we shouldn't reduce/accumulate stats during a list API call).

The list VM API response method is used not only for listVMs API but also for several other VM related lifecycle APIs (such as migrate, stop/start etc) which all return the same vm response class/object. So, I've tried to have a different approach with the PR, and introduce a new global setting that can allow users who want backward compatibility but largely benefit users who don't want it (as default). I've updated the description to reflect this.

rohityadavcloud:
@blueorangutan package

rohityadavcloud:
@blueorangutan test

rohityadavcloud:
@blueorangutan package

GutoVeronezi:
> At the time I couldn't co-relate what made the API slow only to later find out through community and customer issues who are facing degraded performance.

That is the main point of the PR #8782 and I explained it in https://github.com/apache/cloudstack/pull/8737#issuecomment-1995191529, which is a discussion you participated, @rohityadavcloud. 

It is not the first time that I have seen some going against a PR and sometime later the same guys or group proposing the same thing. @JoaoJandre had already proposed the same type of solution on #8782 and received several comments against it; all lacking technical reason to be against it. And now we are facing a PR from those who were against João's PR doing the same thing. This is a duplicate of PR #8782; I am -1 on this PR and PR #8782 should be prioritized.

shwstppr (approved):
Code LGTM

@blueorangutan package

rohityadavcloud:
As I don't care much for the PR anymore and #8782 has addressed most if not all my concerns, I'm closing this PR. I may raise a new PR for changes that aren't in #8782 and we probably may need a doc PR in release notes.

PR #9035, opened 2024-05-03
linstor: disconnect-disk also search for resource name in Linstor

rp-:
### Description

disconnectPhysicalDisk(String, KVMStoragePool) seems to calls the plugin with the resource name instead of the device path, so we also have to search for resource names, while cleaning up.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

6 Node CloudStack cluster, non hyperconverged.

### DIFF
--- a/plugins/storage/volume/linstor/src/main/java/com/cloud/hypervisor/kvm/storage/LinstorStorageAdaptor.java
+++ b/plugins/storage/volume/linstor/src/main/java/com/cloud/hypervisor/kvm/storage/LinstorStorageAdaptor.java
@@ -341,7 +341,7 @@ public class LinstorStorageAdaptor implements StorageAdaptor {
                     null,
                     null);
 
-            optRsc = getResourceByPath(resources, volumePath);
+            optRsc = getResourceByPathOrName(resources, volumePath);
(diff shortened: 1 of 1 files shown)


DaanHoogland:
@rp- , isn't this obfuscating a bug where the path should have been stored instead of the name?

rp-:
> @rp- , isn't this obfuscating a bug where the path should have been stored instead of the name?

Good point, but I store the disk uuid as path and reference for Linstor and it only gets translated to the real device path in the `getPhysicalDisk()` and `createPhysicalDisk()` calls in the adaptor.

So yes, the name volumePath might be misleading here, because in Linstor case it is actually part of the resource name.

sureshanaparti (approved):
CLGTM

sureshanaparti:
@blueorangutan package

DaanHoogland (approved):

DaanHoogland:
@blueorangutan test

PR #9096, opened 2024-05-21
listNetworks optimization

abh1sar:
### Description

This PR fixes the performance issue where listNetworks api goes through all the rows of the networks table and takes longer irrespective of the pagination used.

[Draft PR. Needs discussion around how the count of networks is displayed.]

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
+++ b/server/src/main/java/com/cloud/network/NetworkServiceImpl.java
@@ -2267,7 +2267,7 @@ public class NetworkServiceImpl extends ManagerBase implements NetworkService, C
             isRecursive = true;
         }
 
-        Filter searchFilter = new Filter(NetworkVO.class, "id", false, null, null);
+        Filter searchFilter = new Filter(NetworkVO.class, "id", false, cmd.getStartIndex(), cmd.getPageSizeVal());
         SearchBuilder<NetworkVO> sb = _networksDao.createSearchBuilder();
 
         if (forVpc != null) {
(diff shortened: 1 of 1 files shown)


server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -2267,7 +2267,7 @@ public Pair<List<? extends Network>, Integer> searchForNetworks(ListNetworksCmd
             isRecursive = true;
         }
 
-        Filter searchFilter = new Filter(NetworkVO.class, "id", false, null, null);
+        Filter searchFilter = new Filter(NetworkVO.class, "id", false, cmd.getStartIndex(), cmd.getPageSizeVal());
```
weizhouapache, line 2270: if networkfilter is Account, Domain or Shared, it should work

However, if the networkfilter is All or AccountDomain, it might not work if the searchFilter is the same
for example, there are 25 networks for account, 25 networks for domain. if page size is 20
- on first page, it returns 20 networks for account, it is good
- on second page, it should return the last 5 networks for account, and the first 15 networks for domain. But it will return the last 5 networks for account, and the last 5 networks for domain. right ?
I think the `searchFilter` needs to be updated after the each search.

abh1sar: That's a very good point. 
The issue here is that we are concatenating results from multiple queries ( lest's say Q1, Q2, Q3). Things **kind of** work for the first page. We can adjust pageSize of a query depending on how many results are already returned for other queries.

But if we want to display the second page, it is hard to say which startIndex should be used for Q2. As it depends on how many results were displayed for Q2 on page 1. Which in turn depends on how many Q1 results were there on Page 1. For example, if page 1 had 15 results for Q1 and 5 results for Q2, then page2 Q2 startIndex should be 6.

So, to fix this I think the filter should always have `startIndex` as `0` and `pageSize` should be `cmd.getPageSizeVal() * (cmd.getStartIndex() + 1)`

 `Filter searchFilter = new Filter(NetworkVO.class, "id", false, 0, (cmd.getStartIndex() + 1) * cmd.getPageSizeVal());`
 
 This way we have info on all previous page results and when pagination is applied at the end of the method, the correct result will be returned.
 
 We are losing some optimization if startIndex is higher but it is still better than reading all rows.

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
```diff
@@ -2322,41 +2322,52 @@ public Pair<List<? extends Network>, Integer> searchForNetworks(ListNetworksCmd
...
+                    networksToReturn.first().addAll(networks.first());
+                    networksToReturn.second(networksToReturn.second() + networks.second());
                 }
                 if (Arrays.asList(Network.NetworkFilter.Shared, Network.NetworkFilter.All).contains(networkFilter)) {
                     // get shared networks
-                    List<NetworkVO> sharedNetworks = listSharedNetworks(buildNetworkSearchCriteria(sb, keyword, id, isSystem, zoneId, guestIpType, trafficType, physicalNetworkId, networkOfferingId,
+                    Pair<List<NetworkVO>, Integer> sharedNetworks = listSharedNetworks(buildNetworkSearchCriteria(sb, keyword, id, isSystem, zoneId, guestIpType, trafficType, physicalNetworkId, networkOfferingId,
                             aclType, true, restartRequired, specifyIpRanges, vpcId, tags, display, vlanId, associatedNetworkId), searchFilter, permittedAccounts);
-                    addNetworksToReturnIfNotExist(networksToReturn, sharedNetworks);
+                    addNetworksToReturnIfNotExist(networksToReturn, sharedNetworks.first());
```
abh1sar, line 2348: The total count of result won't be super accurate now.
As before we would have gone through all the account specific and domain level networks and gotten the union with sharedNetworks. That would have returned the exact total networks count.

But now we will be working with at max 20 account and domain level networks in the list. So there is no way to apply union with all the shared networks.

For example,
Domain level netowrks = 100
Shared networks = 100
Intersection(Domain, Shared) among all of the domain and shared networks = 50

Before this change :
Total network count = Domain n/w + shared n/w - intersection = 150

After this change:
Intersection(Domain, shared) among the first 20 domain and shared networks returned by the query = 10
Total network count = Domain n/w + shared n/w - intersection = 190

shwstppr:
@blueorangutan package

server/src/main/java/com/cloud/network/NetworkServiceImpl.java
rohityadavcloud: Are there not any other API similar to this where pagination is applied where the pattern can be repeated, maybe templates, volumes, offerings?

shwstppr:
Discussed with @abh1sar. It will be worth investigating if instead of querying networks multiple times for different filters.
Worst case, we can use startIndex=0, pageSize=pageNumber*page for initial network queries
I'll also investigate if we can optimize listing for supported services, canusedeploy params.

rohityadavcloud:
any update/progress on this @shwstppr ?

abh1sar:
> any update/progress on this @shwstppr ?

@rohityadavcloud, 
Abhishek suggested to see if there is any alternative way to query the networks only once using some kind of Join instead of querying multiple times. But I didn't get time to look at it yet.

Issue #9161, opened 2024-06-02
Configure server max and min threads.


ahmedali6:
##### ISSUE TYPE

 * Improvement Request

##### CLOUDSTACK VERSION

~~~
4.19.0.1
~~~

##### OS / ENVIRONMENT

Rocky-linux 9

##### SUMMARY

Thread pool max and min are hardcoded and can't be configured https://github.com/apache/cloudstack/blob/5433e775e53ad3709db6286cb7b0a78703c823a6/client/src/main/java/org/apache/cloudstack/ServerDaemon.java#L162

I suggest adding something like ```threads.min``` and ```thread.max``` to the ```server.proprties``` file

DaanHoogland:
fixed in #11540

PR #9165, opened 2024-06-03
make server threads configurable with server.properties file

ahmedali6:
### Description

Fixes: https://github.com/apache/cloudstack/issues/9161

### Types of changes
- Enhancement (improves an existing feature and functionality)

- Minor

### How Has This Been Tested?

Tested manually by building and running the server.


lucas-a-martins (approved):
CLGTM

DaanHoogland:
@blueorangutan package

DaanHoogland:
@blueorangutan test

rohityadavcloud (approved):
LGTM, didn't test it though

rohityadavcloud:
Ideally max thread could be 4 or 8 times the number of CPUs, need to explore how jetty connection pooling works - in which case we may only want to configure the max threads (min can be set to 4 or 8 times number of CPUs).

client/src/main/java/org/apache/cloudstack/ServerDaemon.java
```diff
@@ -158,8 +164,8 @@ public void init(final DaemonContext context) {
     public void start() throws Exception {
         // Thread pool
         final QueuedThreadPool threadPool = new QueuedThreadPool();
-        threadPool.setMinThreads(10);
-        threadPool.setMaxThreads(500);
+        threadPool.setMinThreads(minThreads);
+        threadPool.setMaxThreads(maxThreads);
```
sureshanaparti, line 168: @ahmedali6 I think, it's better to some restriction on min & max threads, as it's configurable now (also, need to add as a comment in the properties file if possible, check if any doc update is required)

ahmedali6: @sureshanaparti I added the threads config to the server. properties file. Do you have a suggestion about what  restrictions to do here?

PR #9173, opened 2024-06-05
server: simplify role change validation

### LINKED ISSUE #9015: admin should be allowed to elevate user from role `User` to `Domain Admin`
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI / API
~~~

##### CLOUDSTACK VERSION

~~~

~~~

##### SUMMARY

as noticed in https://github.com/apache/cloudstack/pull/9012#issuecomment-2085425939

When an admin want to grant domain admin rights to a user with role user:
image
This is true in ROOT, where it maybe should fail but with a better message. In a subdomain this would be a perfectly acceptable action.

##### STEPS TO REPRODUCE

~~~

~~~

##### EXPECTED RESULTS

~~~

~~~

##### ACTUAL RESULTS

~~~

~~~

shwstppr:
### Description

Fixes #9015

Simplifies role change checks with the following conditions:
 - Caller should not be of Unknown role type
 - New role's type should not be Unknown
 - Caller should not be able to escalate or de-escalate an account's role which is of higher role type
 - New role should not be of type Admin with domain other than ROOT domain

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial


shwstppr:
@blueorangutan package

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -1287,16 +1287,33 @@ public Pair<Long, Account> doInTransaction(TransactionStatus status) {
...
+     Role change should follow the below conditions:
+     - Caller should not be of Unknown role type
+     - New role's type should not be Unknown
+     - Caller should not be able to escalate or de-escalate an account's role which is of same or higher role type
+     - New role should not of type Admin with domain other than ROOT domain
+     */
+    protected void validateRoleChange(Account account, Role role, Account caller) {
+        Role currentRole = roleService.findRole(account.getRoleId());
+        Role callerRole = roleService.findRole(caller.getRoleId());
+        String errorMsg = String.format("Unable to update account role to %s", role.getName());
```
DaanHoogland, line 1300: meybe add commas to the messages? i.e:
```suggestion
        String errorMsg = String.format("Unable to update account role to %s, ", role.getName());
```

DaanHoogland (approved):
clgtm, good cleanup

server/src/main/java/com/cloud/user/AccountManagerImpl.java
```diff
@@ -1287,16 +1287,33 @@ public Pair<Long, Account> doInTransaction(TransactionStatus status) {
...
-        if (role.getRoleType().ordinal() < currentRole.getRoleType().ordinal() && ((account.getType() == Account.Type.NORMAL && role.getRoleType().getAccountType().ordinal() > Account.Type.NORMAL.ordinal()) ||
-                account.getType().ordinal() > Account.Type.NORMAL.ordinal() && role.getRoleType().getAccountType().ordinal() < account.getType().ordinal() && role.getRoleType().getAccountType().ordinal() > 0)) {
-            throw new PermissionDeniedException(String.format("Unable to update account role to %s as you are " +
-                    "attempting to escalate the account %s to account type %s which has higher privileges", role.getName(), account.getAccountName(), role.getRoleType().getAccountType().name()));
+    /*
+     Role change should follow the below conditions:
+     - Caller should not be of Unknown role type
+     - New role's type should not be Unknown
+     - Caller should not be able to escalate or de-escalate an account's role which is of same or higher role type
+     - New role should not of type Admin with domain other than ROOT domain
```
shwstppr, line 1295:
```suggestion
     - New role should not be of type Admin with domain other than ROOT domain
```

DaanHoogland:
@blueorangutan test

rohityadavcloud (approved):
LGTM - didn't test it though

shwstppr:
Thanks for the review but I'm not sure if we agree on the behaviour which is why this is still in draft. I will try to summarise impact of this change
- ROOT admin can promote or demote all roles except when it tries to make a domain-admin from a subdomain to admin there will be an error
- Domain-admins (if allowed based on API permission) can promote or demote an account to roles lower than domain-admin level. _They won't be able to make a normal user account into domain-admin or vice-versa._
- Normal user accounts (if allowed based on API permission) can promote or demote an account to roles lower than normal account level (probably none).

DaanHoogland:
Thanks @shwstppr 

> * Domain-admins (if allowed based on API permission) can promote or demote an account to roles lower than domain-admin level. _They won't be able to make a normal user account into domain-admin or vice-versa._

I would expect a Domain Admin to promote a user up to their own level.

> * Normal user accounts (if allowed based on API permission) can promote or demote an account to roles lower than normal account level (probably none).

well, that mean they can make other users read only. I am not sure if this is desired.

shwstppr:
> 
> I would expect a Domain Admin to promote a user up to their own level.
> 

I feel this shouldn't be allowed. It should be done by account with higher privileges, ROOT admin here. We can change if others agree on the same.

> 
> well, that mean they can make other users read only. I am not sure if this is desired.

No, they won't be able to. Check is based on the RoleType (https://github.com/apache/cloudstack/blob/4.19/api/src/main/java/org/apache/cloudstack/acl/RoleType.java#L30-L34). So, they won't be able to do anything even if API allows unless we add a new RoleType in the code.

shwstppr:
@blueorangutan package

shwstppr:
> > I would expect a Domain Admin to promote a user up to their own level.
> 
> I feel this shouldn't be allowed. It should be done by account with higher privileges, ROOT admin here. We can change if others agree on the same.
> 
> > well, that mean they can make other users read only. I am not sure if this is desired.
> 
> No, they won't be able to. Check is based on the RoleType (https://github.com/apache/cloudstack/blob/4.19/api/src/main/java/org/apache/cloudstack/acl/RoleType.java#L30-L34). So, they won't be able to do anything even if API allows unless we add a new RoleType in the code.

Changed this to suggested check, ie, allow updating to same level as caller

@blueorangutan package

DaanHoogland:
@blueorangutan package

DaanHoogland:
@shwstppr , do we still have functional doubts on the current state of the code? (or can we postpone further discussion to a next issue/PR)

shwstppr:
@DaanHoogland I think this is okay from my side but will need some testing.

@blueorangutan package

DaanHoogland:
thanks @shwstppr , on the list ;)

DaanHoogland:
@blueorangutan test keepEnv

DaanHoogland:
tested with a user-, user+, domadmin- and domadmin+ role to see wich was possible.
I gave the domadmin roles updateRole and updateRolePermissions to try and edit roles betond their priviledges.
the - accounts had no activateProject permission 
and the + account had createAccount
the domadmin- could not assign a + role to a user
no rootadmin could be assigned in a child domain.

Issue #8975, opened 2024-04-25
When deploying node to Kubernetes it times out


PaX101:
When deploying a node on Kubernetes in Cloudstack 4.19.0.0 it times out when it tries to list the VM's:

`Error syncing load balancer: failed to ensure load balancer: error retrieving list of hosts: Get https://my.url.com/client/api?apiKey=myapikeyhere&command=listVirtualMachines&listall=true&response=json&signature=mysignature: net/http: request canceled (Client.Timeout exceeded while awaiting headers)`

![Kub](https://github.com/apache/cloudstack/assets/4069927/1e8cf594-da68-45cf-a4e7-22ed30cec9ff)

When i try and load the page myself if takes around 5-10 minutes to load the VM list similar to the bug in 4.18 when trying to list VM's. If i change the command to: listVirtualMachinesMetrics in my broswer it loads instantly.

rohityadavcloud:
@PaX101 yes because you're listing the listVirtualMachinesMetrics API which by default returns VM stats (which slows down the API request). If you don't need all the details, you can pass details=min to the API response (or specific details you'd expect).

If you're not able to control this API request, one way to address this could be to change the default behaviour that's change-able via global setting.

rohityadavcloud:
@PaX101 I've proposed a PR here which might help your case - https://github.com/apache/cloudstack/pull/8985
Can you try to disable `vm.stats.increment.metrics` global setting in your environment and see if it helps speed you listVIrtualMachines API calls?

weizhouapache:
> @PaX101 yes because you're listing the listVirtualMachinesMetrics API which by default returns VM stats (which slows down the API request). If you don't need all the details, you can pass details=min to the API response (or specific details you'd expect).
> 
> If you're not able to control this API request, one way to address this could be to change the default behaviour that's change-able via global setting.

good idea to list virtualmachine with details parameters
I created a PR to address it in ccm: https://github.com/apache/cloudstack-kubernetes-provider/pull/60

rohityadavcloud:
Additionally fixed in https://github.com/apache/cloudstack/pull/9177
To make the listing faster, the new global setting will need to be set to false.

PR #9177, opened 2024-06-05
api: listVM API improvement followup, change returning of stats detail

rohityadavcloud:
### Description

Followup from https://github.com/apache/cloudstack/pull/8782

- Changes behaviour of details param handling via global setting:
  - listVirtualMachines API: when the details param is not provided, it returns whether stats are returned controlled by a new global setting `list.vm.default.details.stats`
  - listVirtualMachinesMetrics API: when the details param is not provided, it uses `all` details including `stats`
- Users who are affected slow performance of the listVirtualMachines API response time can set `list.vm.default.details.stats` to `false`
- Remove ConfigKey vm.stats.increment.metrics.in.memory which was renamed to `vm.stats.increment.metrics` in #5984 and also remove unused/unnecessary global settings via upgrade path
- Changes default value of VM stats accumulation setting `vm.stats.increment.metrics` to false until a better solution emerges. Since #5984, this is true and during the execution of listVM APIs the stats are clubbed/calculated which can immensely slow down list VM API calls. Any costly operations such as summing of stats shouldn't be done during the course of a synchronous API, such as the list VM API.
- Fix UI that uses listVirtualMachinesMetrics to not call `stats` detail when in list view without metrics selected.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
@@ -99,8 +99,8 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements 
                collectionType = CommandType.STRING,
                description = "comma separated list of vm details requested, "
                    + "value can be a list of [all, group, nics, stats, secgrp, tmpl, servoff, diskoff, backoff, iso, volume, min, affgrp]."
-                   + " If no parameter is passed in, the details will be defaulted to all. When return.vm.stats.on.vm.list is true, the default" +
-                       "details change to [group, nics, secgrp, tmpl, servoff, diskoff, backoff, iso, volume, min, affgrp], thus the stats will not be returned. ")
+                   + " If no parameter is passed in, the details will be defaulted to all details except stats when list.vm.default.details.stats is false (default)."
+                   + " When list.vm.default.details.stats is true, all details including stats are returned")
     private List<String> viewDetails;
 
     @Parameter(name = ApiConstants.TEMPLATE_ID, type = CommandType.UUID, entityType = TemplateResponse.class, description = "list vms by template")
--- a/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/user/vm/ListVMsCmd.java
@@ -99,8 +99,8 @@ public class ListVMsCmd extends BaseListRetrieveOnlyResourceCountCmd implements 
                collectionType = CommandType.STRING,
                description = "comma separated list of vm details requested, "
                    + "value can be a list of [all, group, nics, stats, secgrp, tmpl, servoff, diskoff, backoff, iso, volume, min, affgrp]."
-                   + " If no parameter is passed in, the details will be defaulted to all details except stats when list.vm.default.details.stats is false (default)."
-                   + " When list.vm.default.details.stats is true, all details including stats are returned")
+                   + " When no parameters are passed, all the details are returned if list.vm.default.details.stats is true (default),"
+                   + " otherwise when list.vm.default.details.stats is false the API response will exclude the stats details.")
     private List<String> viewDetails;
 
     @Parameter(name = ApiConstants.TEMPLATE_ID, type = CommandType.UUID, entityType = TemplateResponse.class, description = "list vms by template")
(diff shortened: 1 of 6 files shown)


rohityadavcloud:
@blueorangutan test

DaanHoogland (approved):
clgtm

GutoVeronezi:
-1 on changing the `list.vm.default.details.stats` default value. Refer to the discussion in #8782.

rohityadavcloud:
@GutoVeronezi this isn't a vote, could you explain why you're opposing now that the default value change when we've agreed in #8782 to address this via documentation of the change in release notes?

DaanHoogland:
@GutoVeronezi , I understand your point, but we do not have a simple usable procedure yet. The implicit procedure was "we don't introduce any backwards incompatability, .. unless we all agree". And so in this case you are right. That does not remove the need to decide on individual issues.
Maybe @rohityadavcloud is a bit quick art the gun, but it makes sense to first introduce a setting and than later reverse the default. We have done it like that in the past.  Maybe we can add this to #8970 (if it is not included yet).

So how do you propose we take this forward? cc @JoaoJandre @sureshanaparti @weizhouapache

GutoVeronezi:
> @GutoVeronezi , I understand your point, but we do not have a simple usable procedure yet. The implicit procedure was "we don't introduce any backwards incompatability, .. unless we all agree". And so in this case you are right. That does not remove the need to decide on individual issues. Maybe @rohityadavcloud is a bit quick art the gun, but it makes sense to first introduce a setting and than later reverse the default. We have done it like that in the past. Maybe we can add this to #8970 (if it is not included yet).
> 
> So how do you propose we take this forward? cc @JoaoJandre @sureshanaparti @weizhouapache

As we already have a configuration to change the behavior, we should keep the current behavior and guide people through the release notes on how to change it. Then, we should continue the #8970 discussion. Once a process is defined and documented, we can look back to the configuration and change the default according to the policy we defined (or even remove the configuration if we decide so). Other systems do that way, like GitLab and OpenStack; the problem is that we do not have this defined and documented yet.

rohityadavcloud:
@GutoVeronezi by your own admission I changed my stand, closed my PR to support Joao's, and built support and consensus with others on Joao's PR and merge his PR. I also don't appreciate some of your remarks and behaviour towards me. It's unfair to me when I'm genuinely trying to help and support everybody and anybody in the community.

I don't see anything wrong with somebody changing their stand when presented with new information or convincing arguments. My reason for changing my stand wrt the default behaviour is that most users don't consume stats (outside of UI) via the list APIs, but many integrations such as Terraform provider, k8s-provider, CAPC, users ansible scripts etc (point of note that these integrations can't be changed) call the list VM APIs and can benefit from any potential speed up. @GutoVeronezi @JoaoJandre and anyone else - what are your reasons of changing your stand wrt the default behaviour now?

For example, look at the two user reported issues that also contributed to change my stand:
https://github.com/apache/cloudstack/issues/8975
https://github.com/apache/cloudstack/issues/7910

I think we introduced the behaviour, and we can discuss and agree if we want to change that if that is largely beneficial to users. I don't think it would detrimental for large number of users if they have gain speedups and still had a global setting to change the behaviour.

That said, I don't have the energy or bandwidth to fight over this, if this is the only thing blocking this I can change the global setting. Have a look, review and advise - @JoaoJandre @DaanHoogland @weizhouapache @sureshanaparti @GutoVeronezi @shwstppr and others...

shwstppr (approved):
Code LGTM

- Slow VM listing needs to be addressed as it has already been reported in various forms
- I don't see changing the value for a boolean global config as a big breaking change if we plan to document it properly. Users can revert it after upgrade if they wish to continue using incremental stats.
- If we did not agree on changing config value in the past we can discuss it now with pros and cons
- I don't think it makes sense to block this change till we decide on #8970. It deals with wider discussion and there are no clear outcomes in the past 45 days or so

GutoVeronezi:
@rohityadavcloud, I did not judge your change of opinion. If you read it carefully, you will see that what I am doing is pointing out that there was no consensus on the change of the default behavior in the whole discussion. There were no direct attacks on anyone, I am just making the timeline clear. 

We would like to introduce some breaking changes and changes to default behavior, but we do not have a documented process for this, and doing it on a case-by-case basis is prone to heated discussions and unfair decisions. Discussion #8970 is there to make these processes well-defined and documented so we can avoid this kind of discussions in the future.

> * I don't think it makes sense to block this change till we decide on [Define a release schedule for the project #8970](https://github.com/apache/cloudstack/discussions/8970). It deals with wider discussion and there are no clear outcomes in the past 45 days or so

@shwstppr 

If more people interact there, we can get some outcomes :smiley:

rohityadavcloud:
Amusing, we're going in circles now and obstinately determined about it.

I've tried my part, got non-technical and discourteous reasons to block the PR, coerced to unrelated agenda. I'm also flattered some want my attention and validation so bad.

I'm gonna sit on the fence on this one, for now; save my time and energy.

`list.vm.default.details.stats` defaults to true now.

@blueorangutan package

rohityadavcloud:
@blueorangutan ui

rohityadavcloud:
Requesting re-review as code has changed a bit

@blueorangutan test

rohityadavcloud:
@blueorangutan test

shwstppr (approved):
Code LGTM

Issue #6743, opened 2022-09-15
CPU speed in compute offering is not a MHz speed when using KVM and libvirt. 


correajl:
##### ISSUE TYPE
 * Improvement Request

##### COMPONENT NAME
~~~
Cloudstack management and cloudstack agent. 
~~~

##### CLOUDSTACK VERSION
~~~
4.17.0.1
~~~

##### CONFIGURATION
2 management servers, 2 databases, advanced network, everything working fine. 

##### OS / ENVIRONMENT
Ubuntu Server 22.04 LTS, KVM, libvirt 8.0.0-1ubuntu7.1, cgroup2.

##### SUMMARY

The value "CPU (in MHz)" used by Compute Offerings has nothing to do with the CPU speed when using KVM. The value configured as "CPU (in MHz)" is mapped to "shares" element in the cputune section of libvirt configuration of domain (KVM). This mapping is done using [this](https://github.com/apache/cloudstack/blob/731a83babfb371b9ee9bb18af9bd33ade7f2c774/engine/orchestration/src/main/java/com/cloud/vm/VirtualMachineManagerImpl.java#L4522)

```
int minSpeed = (int)(newServiceOffering.getSpeed() / (divideCpuByOverprovisioning ? cpuOvercommitRatio : 1));
```

So, if I have a 1000 Mhz value in compute offering and try to launch a VM with 80 CPUs, I will get 80000 configured as 'shares' value. 'Shares' is a relative value that defines how much CPU time a VM will have when compared to others. This description in MHz is very confusing. 

The second problem is why is this value required when creating a custom constrained compute offering? Why couldn't we just inform the number of CPUs and memory to limit the ranges that service offering permits? In UI the value is required. Looking at
createServiceOffering API definition the only two required parameters are displaytext and name. But trying to create I have:

```.shell-session
~# create serviceoffering name="Custom VM" displaytext="Custom VM" storagetype=shared provisioningtype=fat mincpunumber=2 maxcpunumber=80 minmemory=1000 maxmemory=1000000 offerha=true dynamicscalingenabled=true
Error: (HTTP 431, error code 4350) For creating a custom compute offering min/max cpu and min/max memory/cpu speed should all be null or all specified
```

My first suggestion is to permit the creation of a compute offering without inform this value.  If this is not possible for some reason, so my suggestion is to change the description of this value in some way. For KVM users/admins, this MHz notation is so confusing. 

##### STEPS TO REPRODUCE
[...]

PR #9191, opened 2024-06-10
add to cpu speed parameter a description of the cgroup2 case

DaanHoogland:
### Description

This PR adds to the `cpuSpeed` parameter of `CreateServiceOfferingCmd` a description how it will be interpreted in the case cgroups are being used on KVM.

Fixes: #6743

### Types of changes

- [ ] build/CI

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [x] Trivial

### DIFF
--- a/api/src/main/java/org/apache/cloudstack/api/command/admin/offering/CreateServiceOfferingCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/admin/offering/CreateServiceOfferingCmd.java
@@ -56,7 +56,7 @@ public class CreateServiceOfferingCmd extends BaseCmd {
     @Parameter(name = ApiConstants.CPU_NUMBER, type = CommandType.INTEGER, required = false, description = "the CPU number of the service offering")
     private Integer cpuNumber;
 
-    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz. In the case cgroups are used on KVM, this is interpreted to calculate the maximum number of shares.")
+    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz. In the case cgroups are used on KVM, this is interpreted to calculate the number of shares. see also https://libvirt.org/formatdomain.html#cpu-tuning")
     private Integer cpuSpeed;
 
     @Parameter(name = ApiConstants.DISPLAY_TEXT, type = CommandType.STRING, description = "The display text of the service offering, defaults to 'name'.")
--- a/api/src/main/java/org/apache/cloudstack/api/command/admin/offering/CreateServiceOfferingCmd.java
+++ b/api/src/main/java/org/apache/cloudstack/api/command/admin/offering/CreateServiceOfferingCmd.java
@@ -56,7 +56,11 @@ public class CreateServiceOfferingCmd extends BaseCmd {
     @Parameter(name = ApiConstants.CPU_NUMBER, type = CommandType.INTEGER, required = false, description = "the CPU number of the service offering")
     private Integer cpuNumber;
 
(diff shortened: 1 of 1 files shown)


rohityadavcloud (approved):

api/src/main/java/org/apache/cloudstack/api/command/admin/offering/CreateServiceOfferingCmd.java
```diff
@@ -56,7 +56,7 @@ public class CreateServiceOfferingCmd extends BaseCmd {
     @Parameter(name = ApiConstants.CPU_NUMBER, type = CommandType.INTEGER, required = false, description = "the CPU number of the service offering")
     private Integer cpuNumber;
 
-    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz.")
+    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz. In the case cgroups are used on KVM, this is interpreted to calculate the maximum number of shares.")
```
weizhouapache, line 59: according to https://libvirt.org/formatdomain.html#cpu-tuning

`
The optional shares element specifies the proportional weighted share for the domain. If this is omitted, it defaults to the OS provided defaults. NB, There is no unit for the value, it's a relative measure based on the setting of other VM, e.g. A VM configured with value 2048 will get twice as much CPU time as a VM configured with value 1024. The value should be in range [2, 262144] using cgroups v1, [1, 10000] using cgroups v2. Since 0.9.0
`
I am not sure if "the maximum number of shares" is correct.
maybe we could also add the link above in the description

DaanHoogland: How about this @weizhouapache : 
```suggestion
    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz. In the case cgroups are used on KVM, this is interpreted to calculate the number of shares. see also https://libvirt.org/formatdomain.html#cpu-tuning")
```

weizhouapache: that's fine !
@DaanHoogland

api/src/main/java/org/apache/cloudstack/api/command/admin/offering/CreateServiceOfferingCmd.java
```diff
@@ -56,7 +56,7 @@ public class CreateServiceOfferingCmd extends BaseCmd {
     @Parameter(name = ApiConstants.CPU_NUMBER, type = CommandType.INTEGER, required = false, description = "the CPU number of the service offering")
     private Integer cpuNumber;
 
-    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz.")
+    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz. In the case cgroups are used on KVM, this is interpreted to calculate the number of shares. see also https://libvirt.org/formatdomain.html#cpu-tuning")
```
BryanMLima, line 59: @DaanHoogland, I think it would benefit the users if we can add more detail on this parameter description. Something like below, what do you think?
```suggestion
    @Parameter(name = ApiConstants.CPU_SPEED, type = CommandType.INTEGER, required = false, description = "the CPU speed of the service offering in MHz. For the KVM hypervisor," +
            " the values of the parameters cpuSpeed and cpuNumber will be used to calculate the `shares` value. This value is used by the KVM hypervisor to calculate how much time" +
            " the VM will have access to the host's CPU. The `shares` value does not have a unit, and its purpose is be a weight value for the host to compare between its guest" +
            " VMs. For more information, see https://libvirt.org/formatdomain.html#cpu-tuning.")
```

DaanHoogland: I updated your suggestion @BryanMLima  (and I was really surprised that I was allowed to???)

I added the more explicit mention of VMware and Xen. Hope you don't mind.

rohityadavcloud (approved):

PR #7485, opened 2023-05-01
engine-storage: fix errored template becomes active

shwstppr:
### Description

Fixes #7342

When MS is restarted, failed non-public, non-featured, user templates are set to Active as their redownload is not attempted. This PR fixes the case.

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)

#### Feature/Enhancement Scale

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

PR #9206, opened 2024-06-10
storage: fix private templates are not copied to new image store

weizhouapache:
### Description

when add a new image store, some templates/ISOs are copied, but some are not.

```
2024-06-10T12:24:50,711 INFO  [o.a.c.s.i.TemplateServiceImpl] (qtp776484396-20:[ctx-eb090c22, ctx-5fa5579c]) (logid:7a783000) Template Sync did not find 211-2-d536fb03-5f89-3e77-8dea-323315bcbfab on image store 3, may request download based on available hypervisor types
...
2024-06-10T12:24:51,043 INFO  [o.a.c.s.i.TemplateServiceImpl] (qtp776484396-20:[ctx-eb090c22, ctx-5fa5579c]) (logid:7a783000) Skip sync downloading private template 211-2-d536fb03-5f89-3e77-8dea-323315bcbfab to a new image store

```

This PR fixes the issue that some template/iso are not copied to new image store

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
+++ b/engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
@@ -532,11 +532,6 @@ public class TemplateServiceImpl implements TemplateService {
                                 s_logger.info("Skip downloading template " + tmplt.getUniqueName() + " since no url is specified.");
                                 continue;
                             }
-                            // if this is private template, skip sync to a new image store
-                            if (isSkipTemplateStoreDownload(tmplt, zoneId)) {
-                                s_logger.info("Skip sync downloading private template " + tmplt.getUniqueName() + " to a new image store");
-                                continue;
-                            }
 
                             // if this is a region store, and there is already an DOWNLOADED entry there without install_path information, which
                             // means that this is a duplicate entry from migration of previous NFS to staging.


weizhouapache:
this is related to #7485 
needs some discussion . cc @shwstppr

weizhouapache:
@blueorangutan package

weizhouapache:

rohityadavcloud (approved):
LGTM - didn't test it

engine/storage/image/src/main/java/org/apache/cloudstack/storage/image/TemplateServiceImpl.java
```diff
@@ -532,11 +532,6 @@ public void handleTemplateSync(DataStore store) {
                                 s_logger.info("Skip downloading template " + tmplt.getUniqueName() + " since no url is specified.");
                                 continue;
                             }
-                            // if this is private template, skip sync to a new image store
-                            if (isSkipTemplateStoreDownload(tmplt, zoneId)) {
```
shwstppr, line 536: I feel the log here is probably wrong otherwise the last condition in the method should take care of the case where the template is private but it needs to be downloaded to a new zone? Or do we download all templates to all stores of the zones?

weizhouapache: the method `isSkipTemplateStoreDownload` was added in your pr #7485 

the original code snippet was introduced many years ago, only  public/featured/system templates are copied.
```
                            if (!tmplt.isPublicTemplate() && !tmplt.isFeatured() && tmplt.getTemplateType() != TemplateType.SYSTEM) {
                                s_logger.info("Skip sync downloading private template " + tmplt.getUniqueName() + " to a new image store");
                                continue;
                            }
```

in my opinion, all the templates/isos should be copied to new image store.
this needs some discussion
(this is why I created this pr as draft. `isSkipTemplateStoreDownload` can be removed later in this pr)

DaanHoogland: will removing this make private templates be copied cross zones? If it is intra zone yes it makes sense to have templates on all stores, though ACS should be smart enough to do with one copy IMNSHO. I think we can remove this check now and consider making template retrieval smarter later.

sureshanaparti: @weizhouapache Can this be controlled through a config (which can be default to false - keep the existing behavior as well)?

DaanHoogland: @weizhouapache , I could live with that but would consider it a gap in functional definition. It was intentionally prohibited in the original code, but I see your point why that would not make sense. Let's go forward with this fix. In this case a setting is not a blocker IMNSHO. We can do but can do without as well. cc @sureshanaparti @weizhouapache

DaanHoogland (approved):

PR #8889, opened 2024-04-06
Updates to HPE-Primera and Pure FlashArray Drivers to use Host-based VLUN Assignments

rg9975:
### Description

This PR makes a number of bug fixes to the Primera and Pure storage array discovered in working environments after the release of 4.19.0.0.  

The following items are addressed:

1.  Switched to a host-centric rather than hostgroup-centric vlun assignment model.  hostgroup-centric works better for shared filesystems but the direct volume attach method for these drivers is better suited to per-host vlun assignments.  This required changes to how vlun id's are stored as well as a number of changes in the storage framework to add missing grantAccess/revokeAccess calls or reorder those calls so data is properly available in the Disk objects sent to agents.
2. Improved error handling and logging in the storage portion of the agent when copying volumes.
3. Fix to properly handle storage tag to storage provider mapping when migrating disks to a different storage provider and selecting a new disk offering.
4. Fix an issue related to template removal on primary storage for Pure when all volumes using the template are deleted.  Pure storage 'saves' the volume in a Deleted area for some period of time after deletion so any recreation of the volume on subsequent volume creations will trigger an error.  The fix renames the volume prior to deletion by appending a timestamp.
5. Small issues with oAuth login discovered during testing when incorrect user password is entered at the login prompt.
6. Several small logging cleanups in a few modules that were producing unnecessary noise at INFO level.
7. Various exception/error handling improvements in storage classes to improve triage/debugging of storage/volume issues.
8. Fix for issue #8789 .

### Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] New feature (non-breaking change which adds functionality)
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (improves an existing feature and functionality)
- [ ] Cleanup (Code refactoring and cleanup, that may add test cases)
- [ ] build/CI

#### Feature/Enhancement Scale

- [ ] Major
- [X] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [X] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?
The standard test suite at plugins/storage/volume/adaptive/README.md was run against a local installation of Pure Flasharray and HPE Primera.

#### How did you try to break this feature and the system with this change?
Test cases include using the built-in NFS driver in combination with both Flasharray and Primera, moving volumes back and forth between all 3 file systems, and testing standard functions of NFS with these drivers in place.

Issue #8789, opened 2024-03-14
FibreChannelAdaptor disconnectPhysicalDiskByPath should not always return true


rp-:
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
KVM-Storage
~~~

##### CLOUDSTACK VERSION

~~~
4.19.0
~~~

##### SUMMARY

I'm not pointing fingers as the Linstor driver also needs a fix for that(comming), but
`disconnectPhysicalDiskByPath` should only return true if it is able to handle that path.
Currently it tries to handle every path and if it is used as the first entry in the StorageadaptorMap it will always "win",
even if it isn't the driver device.
Also the formatting looks a bit off in that method.

https://github.com/apache/cloudstack/blob/e87c6cfcb1d6ec722dc7f8f6d4579b585e5bb87e/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/MultipathSCSIAdapterBase.java#L225

##### STEPS TO REPRODUCE

~~~
Have a non FibreChannel primary storage and stop/move a VM to another host.
disconnectbyPath will be called and if FibreChannel adaptor is 1. in the Map it will take over the disconnect call.
~~~

##### EXPECTED RESULTS

~~~
FibreChannel driver only handling paths belonging to it
~~~

##### ACTUAL RESULTS

~~~

~~~

@rg9975

harikrishna-patnala:
@rp- I see the following part has be already handled in the PR #8889 
 public boolean disconnectPhysicalDiskByPath(String localPath) { 

I'm trying to fix the case of running into the loop of all adapters by somehow passing the storage pool type to that caller cleanup() method but this is touching all over the code (which I fear it creates other regressions), instead I feel we can keep it the current way only since Fibrechannel adapter has already fixed. Whats your thoughts @rp-

harikrishna-patnala:
Thanks @rp- let me add a clear java doc for that purpose and explaining the possible bug.

rohityadavcloud:
Javadoc addressed by https://github.com/apache/cloudstack/pull/9210

PR #9210, opened 2024-06-11
Update the java doc for the method disconnectPhysicalDiskByPath

harikrishna-patnala:
### Description

This PR addresses the issue #8789 

The original issue is disconnectPhysicalDiskByPath() implementation in FibreChannelAdaptor always returns true irrespective of the success of the operation. This was already fixed in the PR #8889 .

Ideally this method has to be called after choosing the right adapter based on the storage pool type of the volume path, but currently it is just called in a loop. 
https://github.com/shapeblue/cloudstack/blob/05b9b6e2e77b9b633f0fe22aff5f9f3c047b2303/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/KVMStoragePoolManager.java#L200-L212

while trying to fix the case of running into the loop of all adapters by somehow passing the storage pool type to that caller cleanup() method but this is touching all over the code (which I fear it creates other regressions), instead I feel we can keep it the current way only since Fibrechannel adapter has already fixed.

In this PR I've added the java doc explaining the method and situation.

### Types of changes

- [ ] build/CI

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [ ] Minor
- [ ] Trivial

### DIFF
--- a/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/StorageAdaptor.java
+++ b/plugins/hypervisors/kvm/src/main/java/com/cloud/hypervisor/kvm/storage/StorageAdaptor.java
@@ -50,8 +50,17 @@ public interface StorageAdaptor {
 
     public boolean disconnectPhysicalDisk(Map<String, String> volumeToDisconnect);
 
-    // given local path to file/device (per Libvirt XML), 1) check that device is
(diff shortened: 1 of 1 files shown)


rp- (approved):

DaanHoogland (approved):
lgtm, only javadoc changes

Issue #8077, opened 2023-10-11
Irritating error message when host does not have free memory


sbrueseke:
##### ISSUE TYPE

 * Improvement Request

##### COMPONENT NAME

~~~
UI, Log
~~~

##### CLOUDSTACK VERSION

~~~
4.18.1.0
~~~

##### SUMMARY

When a specified host does not have enough free memory capacity the logfile and UI shows the following error message:
Cannot deploy the VM to specified host 10; host has cpu capability? true, host has capacity? false

Which is a little bit irritating, because memory as a reason is not mentioned.

##### STEPS TO REPRODUCE

~~~
deploy instances as root and specify host in deployment wizard. When host does not have enough free available memory, you see the error in UI and logfile.
~~~

##### EXPECTED RESULTS

~~~
error message should mention the root reason for not be able to deploy vm.
~~~

##### ACTUAL RESULTS

~~~
Logfile looks like this and UI shows error message from logfile:
2023-10-11 15:52:46,928 DEBUG [c.c.v.UserVmManagerImpl] (API-Job-Executor-50:ctx-98e0cc75 job-1563 ctx-89e70556) (logid:cc26398f) Destination Host to deploy the VM is specified, specifying a deployment plan to de
ploy the VM
2023-10-11 15:52:46,939 DEBUG [c.c.c.CapacityManagerImpl] (API-Job-Executor-50:ctx-98e0cc75 job-1563 ctx-89e70556) (logid:cc26398f) Host: 10 has cpu capability (cpu:56, speed:2400) to support requested CPU: 4 and
 requested speed: 2000
2023-10-11 15:52:46,940 DEBUG [c.c.c.CapacityManagerImpl] (API-Job-Executor-50:ctx-98e0cc75 job-1563 ctx-89e70556) (logid:cc26398f) Checking if host: 10 has enough capacity for requested CPU: 8000 and requested RAM: (8.00 GB) 8589934592 , cpuOverprovisioningFactor: 4.0
2023-10-11 15:52:46,945 DEBUG [c.c.c.CapacityManagerImpl] (API-Job-Executor-50:ctx-98e0cc75 job-1563 ctx-89e70556) (logid:cc26398f) Hosts's actual total CPU: 134400 and CPU after applying overprovisioning: 537600
[...]

PR #9218, opened 2024-06-11
Fix error message if specific host does not have capacity

harikrishna-patnala:
### Description

This PR fixes the issue #8077 with some related error message

Trying to publish the exact error message is bit complicated in this scenario. Operator has anyway need to check the logs, but here tried to keep the appropriate error message to check further.

Case 1:

If the service offering has more CPU number or speed than the whole host's CPU and speed.

image

Case 2:

If the host does not have free CPU/RAM as requested from the compute offering. The proper error message depends on the over commit ratios and whether or not to consider the reserved capacity of the host, so checking of the logs is needed. 

image

### Types of changes

- [ ] build/CI

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### How Has This Been Tested?

case 1:

1. Create a service offering with more than the CPU than the host
2. Create an instance with that offering by selecting the host

case 2: 

1. Create a service offering so that the host cannot allocate the resources within its limits
2. Create an instance with that offering by selecting the host


harikrishna-patnala:
@blueorangutan package

rohityadavcloud (approved):

DaanHoogland (approved):
clgtm

rohityadavcloud:
Marvin/simulator smoketests are enough, as it's largely error/presentation change.

PR #9219, opened 2024-06-11
saml: introduce saml2.check.signature

### LINKED ISSUE #4519: SAML signature is optional
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
UI: SAML login
~~~

##### CLOUDSTACK VERSION

~~~
4.14.0.0
~~~

##### CONFIGURATION

SAML authentication activated

##### OS / ENVIRONMENT

N/A

##### SUMMARY

The SAML signature is optional and not mandatory. A user can login into a domain where he's not allowed to.

##### STEPS TO REPRODUCE

Use a tool to intercept the web requests between your browser ant the cloudstack UI (our pentester used "burp"). Login to cloudstack using a SSO account (SAML).
Now, you're logged in successfully to your cloudstack domain. 
Logout from cloudstack.

Now, log in again to cloudstack, but with the "interceptor" active. In the request, delete the whole saml-signature and change the username to another SSO user from within another tenant.
Run the modified request, and you're logged in into this foreign account.

~~~

~~~

##### EXPECTED RESULTS

User can't log in into foreign tenant
~~~

~~~

##### ACTUAL RESULTS

User can log in into foreign tenant

~~~

~~~

rohityadavcloud:
Adminstrators should ensure that IDP configuration has signing certificate for the actual signature check to be performed. In addition to this, this change introduces a new global setting `saml2.check.signature` which can deliberately fail a SAML login attempt when the SAML response has missing signature.

Fixes #4519

### Types of changes

- [ ] build/CI

### DIFF
--- a/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/api/command/SAML2LoginAPIAuthenticatorCmd.java
+++ b/plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/api/command/SAML2LoginAPIAuthenticatorCmd.java
@@ -144,6 +144,14 @@ public class SAML2LoginAPIAuthenticatorCmd extends BaseCmd implements APIAuthent
         return responseObject;
     }
 
+    protected void checkAndFailOnMissingSAMLSignature(Signature signature) {
+        if (signature == null && SAML2AuthManager.SAMLCheckSignature.value()) {
+            s_logger.error("Failing SAML login due to missing signature in the SAML response and signature check is enforced. " +
+                    "Please check and ensure the IDP configuration has signing certificate or relax the saml2.check.signature setting.");
+            throw new ServerApiException(ApiErrorCode.INTERNAL_ERROR, "Signature is missing from the SAML Response. Please contact the Administrator");
+        }
+    }
+
     @Override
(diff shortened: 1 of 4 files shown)


shwstppr (approved):
Code LGTM. Would it make sense to move the new check in a separate method and add some unit tests?

plugins/user-authenticators/saml2/src/main/java/org/apache/cloudstack/api/command/SAML2LoginAPIAuthenticatorCmd.java
```diff
@@ -238,9 +245,8 @@ public String authenticate(final String command, final Map<String, Object[]> par
                                 params, responseType));
                     }
                 }
-                if (username == null) {
```
rohityadavcloud, line 241: FYI - username is always declared as null above, so no point in checking that.

rohityadavcloud:
@blueorangutan test

DaanHoogland (approved):
clgtm

rohityadavcloud:
Tested locally, in default case the checks are totally skipped; more secured env will need testing with and without encrypted assertions - real world users can test those. Unit tests ensure/test for both cases.
@blueorangutan test

PR #9298, opened 2024-06-24
Restart agent when host comes out of maintenance

### LINKED ISSUE #9295: Storage pools are not mounted when the host is in maintenance mode  or comes out of maintenance mode
ISSUE TYPE

Bug

COMPONENT NAME

Component: storage,kvm

CLOUDSTACK VERSION

Cloudstack version 4.19.0, 4.18 
KVM

SUMMARY

Storage pools are not mounted when the host is in maintenance mode or comes out of maintenance mode 

Steps to reproduce  the issue 

Test case scenario 1

1. Have Cluster C1 with host H1 and cluster C2 with host H2
2. Have only one cluster wide storage pool P1 attached to C1
3. Put H2 on maintenance
4. Add a zone wide storage pool P2.
5. Take H2 out of maintenance
6. Deploy a VM in H2. it will fail as P2 is not mounted on H2.

Test case scenario 2

1. Add zone wide pool P2 when both hosts are not in maintenance.
2. Disable Pool P2.
3. Reboot H2
4. P2 won't be mounted on H2 as it is disabled.
5. VM deployment on H2 fails as it is not able to find any pool
 

**Actual Behaviour** 

Cloudstack is not mounting the storage pool when the host is maintenance mode or comes out of maintenance mode

**Expected behaviour** 

Cloudstack should mounting the storage pool when the host is maintenance mode or comes out of maintenance mode

abh1sar:
### Description

This PR...

Fixes #9295 

New storage pools are not added to a host which is in maintenance. This change will restart the agent when host is taken out of maintenance mode. StoragePoolMonitor.processConnect() will then get called which will add/mount the new storage pool to the host. 

One more related issue was that storage pool was not getting unmounted from a host in maintenance mode. Even when the host is taken out of maintenance mode, it will still have the pool mounted. Only reboot will unmount the pool.
So, as fix have added DeleteStoragePoolCommand to be able to run in maintenance mode similar to ModifyStoragePoolCommand.

### Types of changes

- [ ] build/CI
- [ ] test (unit or integration test code)

- [ ] Major
- [x] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [ ] Major
- [x] Minor
- [ ] Trivial

### DIFF
--- a/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
+++ b/server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
@@ -2890,7 +2890,7 @@ public class ResourceManagerImpl extends ManagerBase implements ResourceManager,
             }
         }
 
-        handleAgentIfNotConnected(host, vms_migrating);
+        handleAgentConnectAndRestart(host, vms_migrating);
 
         try {
             resourceStateTransitTo(host, ResourceState.Event.AdminCancelMaintenance, _nodeId);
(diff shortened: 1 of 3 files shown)


abh1sar:
@blueorangutan package

abh1sar:
@blueorangutan test

sureshanaparti (approved):
clgtm

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2905,16 +2905,16 @@ private boolean doCancelMaintenance(final long hostId) {
     }
 
     /**
-     * Handle agent (if available) if its not connected before cancelling maintenance.
      * Agent must be connected before cancelling maintenance.
-     * If the host status is not Up:
+     * Connect and restart agent (if available) if its not connected before cancelling maintenance.
```
sureshanaparti, line 2909:
```suggestion
     * Connect and restart agent (if available) before cancelling maintenance.
```

server/src/main/java/com/cloud/resource/ResourceManagerImpl.java
```diff
@@ -2905,16 +2905,16 @@ private boolean doCancelMaintenance(final long hostId) {
...
+     * Restart agent even if it is connected to refresh the storage pools
```
sureshanaparti, line 2910: @abh1sar check if any doc update required for this behavior change during cancel maintenance.

abh1sar: @sureshanaparti I didn't find anywhere in the docs that mentions anything regarding this

PR #9433, opened 2024-07-23
[VMware] Update data disk controller same as the root disk controller type when it is not set in the VM detail

### LINKED ISSUE #9400: Failed add volume when instance running on VMware use VMware Paravirtual SCSI controller
##### ISSUE TYPE

 * Bug Report

##### COMPONENT NAME

~~~
Instance, volume, VMware
~~~

##### CLOUDSTACK VERSION

~~~
 4.19.0.1
~~~

##### CONFIGURATION

VMware backend

##### OS / ENVIRONMENT

Ubuntu 22.04 for Management 
VMware 8.0 

##### STEPS TO REPRODUCE

~~~
Test case 1:
Set Root Disk Controller to pvscsi
Create instance from template
After create success add new volume

Test case 2:
Create Volume manual
Attach volume to instance
~~~

##### EXPECTED RESULTS

~~~
Volume attach to Instance
~~~

##### ACTUAL RESULTS

~~~
Failed to attach volume to instance

Log error:
2024-07-16 08:49:07,688 ERROR [c.c.a.ApiAsyncJobDispatcher] (API-Job-Executor-20:ctx-97d2789a job-22512) (logid:accc321a) Unexpected exception while executing org.apache.cloudstack.api.command.user.volume.AttachVolumeCmd
com.cloud.utils.exception.CloudRuntimeException: Failed to attach volume new-volume to VM test-vm
        at com.cloud.storage.VolumeApiServiceImpl.sendAttachVolumeCommand(VolumeApiServiceImpl.java:4402)
        at com.cloud.storage.VolumeApiServiceImpl.orchestrateAttachVolumeToVM(VolumeApiServiceImpl.java:2304)
        at com.cloud.storage.VolumeApiServiceImpl.orchestrateAttachVolumeToVM(VolumeApiServiceImpl.java:4803)
        at jdk.internal.reflect.GeneratedMethodAccessor4425.invoke(Unknown Source)
        at com.cloud.vm.VmWorkJobHandlerProxy.handleVmWorkJob(VmWorkJobHandlerProxy.java:107)
        at com.cloud.storage.VolumeApiServiceImpl.handleVmWorkJob(VolumeApiServiceImpl.java:4842)
        at jdk.internal.reflect.GeneratedMethodAccessor357.invoke(Unknown Source)
        at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:344)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:198)
        at org.springframework.aop.interceptor.ExposeInvocationInterceptor.invoke(ExposeInvocationInterceptor.java:97)
        at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186)
        at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:215)
        at com.sun.proxy.$Proxy235.handleVmWorkJob(Unknown Source)
        at com.cloud.vm.VmWorkJobDispatcher.runJob(VmWorkJobDispatcher.java:102)
        at org.apache.cloudstack.framework.jobs.impl.AsyncJobManagerImpl$5.runInContext(AsyncJobManagerImpl.java:654)
        at org.apache.cloudstack.managed.context.ManagedContextRunnable$1.run(ManagedContextRunnable.java:48)
        at org.apache.cloudstack.managed.context.impl.DefaultManagedContext$1.call(DefaultManagedContext.java:55)
        at org.apache.cloudstack.managed.context.impl.DefaultManagedContext.callWithContext(DefaultManagedContext.java:102)
        at org.apache.cloudstack.managed.context.impl.DefaultManagedContext.runWithContext(DefaultManagedContext.java:52)
        at org.apache.cloudstack.managed.context.ManagedContextRunnable.run(ManagedContextRunnable.java:45)
        at org.apache.cloudstack.framework.jobs.impl.AsyncJobManagerImpl$5.run(AsyncJobManagerImpl.java:602)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:829)
2024-07-16 08:49:07,690 DEBUG [o.a.c.f.j.i.AsyncJobManagerImpl] (API-Job-Executor-20:ctx-97d2789a job-22512) (logid:accc321a) Complete async job-22512, jobStatus: FAILED, resultCode: 530, result: org.apache.cloudstack.api.response.ExceptionResponse/null/{"uuidList":[],"errorcode":"530","errortext":"Failed to attach volume new-volume to VM test-vm"}
~~~

sureshanaparti:
### Description

This PR updates data disk controller (in VMware) same as the root disk controller type when it is not set in the VM detail. This is to ensure the disk controller is available for the data disks, as all the SCSI controllers are created with same controller type.

Fixes #9400 

### Types of changes

- [ ] build/CI
- [ ] test (unit or integration test code)

- [ ] Major
- [ ] Minor

#### Bug Severity

- [ ] BLOCKER
- [ ] Critical
- [x] Major
- [ ] Minor
- [ ] Trivial

### How Has This Been Tested?

Manually testing with Paravirtual controller (pvscsi) as the root disk controller.

**Before =>**

DataDiskController_Before

DiskControllers_Before

**After =>**

DataDiskController_After

DiskControllers_After

### DIFF
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -4655,11 +4655,6 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                 String vmwareRootDiskControllerTypeFromSetting = StringUtils.defaultIfEmpty(_configDao.getValue(Config.VmwareRootDiskControllerType.key()),
                         Config.VmwareRootDiskControllerType.getDefaultValue());
                 vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, vmwareRootDiskControllerTypeFromSetting);
-                if (vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER) == null) {
-                    if (vmwareRootDiskControllerTypeFromSetting.equalsIgnoreCase("scsi")) {
-                        vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "scsi");
-                    }
-                }
             }
 
             if (vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER) == null) {
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -4650,16 +4650,25 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                 vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, dataDiskControllerSetting);
             }
 
-            String controllerSetting = StringUtils.defaultIfEmpty(_configDao.getValue(Config.VmwareRootDiskControllerType.key()),
-                    Config.VmwareRootDiskControllerType.getDefaultValue());
-
             // Don't override if VM already has root/data disk controller detail
             if (vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER) == null) {
-                vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, controllerSetting);
+                String vmwareRootDiskControllerTypeFromSetting = StringUtils.defaultIfEmpty(_configDao.getValue(Config.VmwareRootDiskControllerType.key()),
+                        Config.VmwareRootDiskControllerType.getDefaultValue());
+                vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, vmwareRootDiskControllerTypeFromSetting);
+                if (vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER) == null) {
+                    if (vmwareRootDiskControllerTypeFromSetting.equalsIgnoreCase("scsi")) {
+                        vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "scsi");
+                    }
+                }
             }
+
             if (vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER) == null) {
-                if (controllerSetting.equalsIgnoreCase("scsi")) {
-                    vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "scsi");
+                String finalRootDiskController = vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER);
+                // Set the data disk controller detail same as the final scsi root disk controller if VM doesn't have data disk controller detail
+                // This is to ensure the disk controller is available for the data disks, as all the SCSI controllers are created with same controller type
+                String scsiControllerPattern = "(?i)\\b(scsi|lsilogic|lsilogicsas|lsisas1068|buslogic|pvscsi)\\b";
+                if (finalRootDiskController.matches(scsiControllerPattern)) {
+                    vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, finalRootDiskController);
                 } else {
                     vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "osdefault");
                 }
--- a/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
+++ b/server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
@@ -4663,8 +4663,11 @@ public class UserVmManagerImpl extends ManagerBase implements UserVmManager, Vir
                 // This is to ensure the disk controller is available for the data disks, as all the SCSI controllers are created with same controller type
                 String scsiControllerPattern = "(?i)\\b(scsi|lsilogic|lsilogicsas|lsisas1068|buslogic|pvscsi)\\b";
                 if (finalRootDiskController.matches(scsiControllerPattern)) {
+                    s_logger.info(String.format("Data disk controller was not defined, but root disk is using SCSI controller [%s]." +
+                            "To ensure disk controllers are available for the data disks, the data disk controller is updated to match the root disk controller.", finalRootDiskController));
                     vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, finalRootDiskController);
                 } else {
+                    s_logger.info("Data disk controller was not defined; defaulting to 'osdefault'.");
                     vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "osdefault");
                 }
             }


sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan help

sureshanaparti:
@blueorangutan test centos7 vmware-80

DaanHoogland (approved):
clgtm

Pearl1594 (approved):
cLGTM

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4650,16 +4650,25 @@ private void updateVMDiskController(UserVmVO vm, Map<String, String> customParam
...
             if (vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER) == null) {
-                vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, controllerSetting);
+                String vmwareRootDiskControllerTypeFromSetting = StringUtils.defaultIfEmpty(_configDao.getValue(Config.VmwareRootDiskControllerType.key()),
+                        Config.VmwareRootDiskControllerType.getDefaultValue());
+                vm.setDetail(VmDetailConstants.ROOT_DISK_CONTROLLER, vmwareRootDiskControllerTypeFromSetting);
+                if (vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER) == null) {
+                    if (vmwareRootDiskControllerTypeFromSetting.equalsIgnoreCase("scsi")) {
+                        vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "scsi");
+                    }
+                }
```
winterhazel, line 4662: Are those lines necessary?

In the condition below, if the `dataDiskController` detail is null, and the `rootDiskController` is a SCSI controller, we're already updating `dataDiskController` to match `rootDiskController`.

sureshanaparti: ah, forgot to remove them. thanks for the review @winterhazel, will update.

sureshanaparti:
@blueorangutan package

sureshanaparti:
@blueorangutan package

winterhazel:
I have not tested in a VMware environment to confirm, but is this really fixing the issue when attaching a volume? The volume attach workflow does not seem to use the changed code.

From what I could gather by reading the code, the problem happens because CloudStack tries to attach the data disks using default SCSI controllers (LSI Logic), but there are only PVSCSI controllers. I think that changing the data disk controllers to LSI Logic (while having PVSCSI root disk controllers) and trying to attach a data disk should still produce the same error.

If that is the case, maybe it would be good to prioritize using the root disk controllers for data disks on the volume attach workflow in case both are SCSI?

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4650,16 +4650,20 @@ private void updateVMDiskController(UserVmVO vm, Map<String, String> customParam
...
+
             if (vm.getDetail(VmDetailConstants.DATA_DISK_CONTROLLER) == null) {
-                if (controllerSetting.equalsIgnoreCase("scsi")) {
-                    vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "scsi");
+                String finalRootDiskController = vm.getDetail(VmDetailConstants.ROOT_DISK_CONTROLLER);
+                // Set the data disk controller detail same as the final scsi root disk controller if VM doesn't have data disk controller detail
+                // This is to ensure the disk controller is available for the data disks, as all the SCSI controllers are created with same controller type
+                String scsiControllerPattern = "(?i)\\b(scsi|lsilogic|lsilogicsas|lsisas1068|buslogic|pvscsi)\\b";
+                if (finalRootDiskController.matches(scsiControllerPattern)) {
+                    vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, finalRootDiskController);
```
winterhazel, line 4666:
```suggestion
                    s_logger.info(String.format("Data disk controller was not defined, but root disk is using SCSI controller [%s]. To ensure disk controllers are available for the data disks, we will update the data disk controllers to match the root disk controller.", finalRootDiskController))
                    vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, finalRootDiskController);
```

Also, I would like to add some logs here to inform how we're choosing data disk controllers

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
```diff
@@ -4650,16 +4650,20 @@ private void updateVMDiskController(UserVmVO vm, Map<String, String> customParam
...
                 } else {
                     vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "osdefault");
```
winterhazel, line 4668:
```suggestion
                    s_logger.info("Data disk controller was not defined; defaulting to 'osdefault'.");
                    vm.setDetail(VmDetailConstants.DATA_DISK_CONTROLLER, "osdefault");
```

sureshanaparti:
> I have not tested in a VMware environment to confirm, but is this really fixing the issue when attaching a volume? The volume attach workflow does not seem to use the changed code.
> 
> From what I could gather by reading the code, the problem happens because CloudStack tries to attach the data disks using default SCSI controllers (LSI Logic), but there are only PVSCSI controllers. I think that changing the data disk controllers to LSI Logic (while having PVSCSI root disk controllers) and trying to attach a data disk should still produce the same error.
> 
> If that is the case, maybe it would be good to prioritize using the root disk controllers for data disks on the volume attach workflow in case both are SCSI?

All the disk controllers in VMware VM are created using the type mentioned in root disk controller (primary controller), so the data disk controller should match with that controller type to support attach data volume. If data disk controller has a different controller type, attach fails as it is not found in the VM spec.

In this case, when data disk controller is not set, previously it was set to osdefault, which picks the default controller type (say, lsilogic) for that OS and might not be the same as root disk controller (say, pvscsi). so, it's better to sync the data disk controller (when not set) with root disk controller.

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
sureshanaparti: done

server/src/main/java/com/cloud/vm/UserVmManagerImpl.java
sureshanaparti: done

sureshanaparti:
@blueorangutan package

vladimirpetrov (approved):
LGTM based on manual testing using VMWare 8.0u1.

vladimirpetrov:
@blueorangutan test

winterhazel:
> > I have not tested in a VMware environment to confirm, but is this really fixing the issue when attaching a volume? The volume attach workflow does not seem to use the changed code.
> > From what I could gather by reading the code, the problem happens because CloudStack tries to attach the data disks using default SCSI controllers (LSI Logic), but there are only PVSCSI controllers. I think that changing the data disk controllers to LSI Logic (while having PVSCSI root disk controllers) and trying to attach a data disk should still produce the same error.
> > If that is the case, maybe it would be good to prioritize using the root disk controllers for data disks on the volume attach workflow in case both are SCSI?
> 
> All the disk controllers in VMware VM are created using the same type mentioned in root disk controller (primary controller), so the data disk controller should match with that controller type created to support attach data volume. If data disk controller has a different controller type, attach fails as it is not found in the VM spec.
> 
> In this case, when data disk controller is not set, previously it was set to osdefault, which picks the default controller type (say, lsilogic) for that OS and might not be the same as root disk controller (say, pvscsi). so, it's better to sync the data disk controller (when not set) with root disk controller.


I understand that you want to synchronize the data disk controller with the root disk controller to ensure data disks can be attached. However, I think that we are doing it in the wrong place. I think that it would be better to synchronize it in the volume attachment process itself.

As you mentioned, if both root disks and data disks use SCSI controllers, the root disk controller is prioritized when creating a VM. So, suppose we have a scenario in which a template has `rootDiskController` set to `pvscsi` and `dataDiskController` set to `osdefault` or any other SCSI subtype (which is the current configuration for some existing VMs because `dataDiskController` defaulted to `osdefault` when not set): if we deploy a VM from that template with a data disk, the deployment process will prioritize using the root disk's controller (PVSCSI) for the data disk, even though `dataDiskController` may have been set set to something else, and the VM will be created successfully with two disks that use PVSCSI controllers. However, if we try to attach another volume to that VM (or detach and reattach the data disk), an exception will be thrown because the volume attachment process tries to attach the data disk using a LSI Logic controller instead.

To avoid this exception for these scenarios as well, I think that it would be better to fix the original issue by making the volume attachment process consistent with the VM creation process: prioritizing the root disk controller in case both details are set to SCSI subtypes, instead of matching them when the VM is created only if `dataDiskController` is empty.

sureshanaparti:
> > > I have not tested in a VMware environment to confirm, but is this really fixing the issue when attaching a volume? The volume attach workflow does not seem to use the changed code.
> > > From what I could gather by reading the code, the problem happens because CloudStack tries to attach the data disks using default SCSI controllers (LSI Logic), but there are only PVSCSI controllers. I think that changing the data disk controllers to LSI Logic (while having PVSCSI root disk controllers) and trying to attach a data disk should still produce the same error.
> > > If that is the case, maybe it would be good to prioritize using the root disk controllers for data disks on the volume attach workflow in case both are SCSI?
> > 
> > 
> > All the disk controllers in VMware VM are created using the same type mentioned in root disk controller (primary controller), so the data disk controller should match with that controller type created to support attach data volume. If data disk controller has a different controller type, attach fails as it is not found in the VM spec.
> > In this case, when data disk controller is not set, previously it was set to osdefault, which picks the default controller type (say, lsilogic) for that OS and might not be the same as root disk controller (say, pvscsi). so, it's better to sync the data disk controller (when not set) with root disk controller.
> 
> I understand that you want to synchronize the data disk controller with the root disk controller to ensure data disks can be attached. However, I think that we are doing it in the wrong place. I think that it would be better to synchronize it in the volume attachment process itself.
> 
> As you mentioned, if both root disks and data disks use SCSI controllers, the root disk controller is prioritized when creating a VM. So, suppose we have a scenario in which a template has `rootDiskController` set to `pvscsi` and `dataDiskController` set to `osdefault` or any other SCSI subtype (which is the current configuration for some existing VMs because `dataDiskController` defaulted to `osdefault` when not set): if we deploy a VM from that template with a data disk, the deployment process will prioritize using the root disk's controller (PVSCSI) for the data disk, even though `dataDiskController` may have been set set to something else, and the VM will be created successfully with two disks that use PVSCSI controllers. However, if we try to attach another volume to that VM (or detach and reattach the data disk), an exception will be thrown because the volume attachment process tries to attach the data disk using a LSI Logic controller instead.
> 
> To avoid this exception for these scenarios as well, I think that it would be better to fix the original issue by making the volume attachment process consistent with the VM creation process: prioritizing the root disk controller in case both details are set to SCSI subtypes, instead of matching them when the VM is created only if `dataDiskController` is empty.

@winterhazel The data disk controller is matched with the SCSI root disk controller only when the data disk controller is not set in the VM's template and VM detail. If template has the root and data disk controllers set to _pvscsi_ and _osdefault_ respectively, then the controllers are created with _pvscsi_ controller to ensure root disk is configured to proper controller to create the VM. Later, any unsupported data disk controller (say, _lsilogic_) set would fail the attach disk, so the operator should ensure to set the root and data disk controller appropriately.

As the controllers are already created during the VM creation process, it's better to synchronize these controller details during VM creation. Synchronizing in the attach process would still set the same controller details (if you mean dynamic reconfiguration of disk controllers in VM, maybe the existing controller reconfiguration based on the data disk controller can be done as a separate improvement later). The changes here would fix the attach issue when data disk controller is not set in the VM's template and VM detail.

DaanHoogland:
@winterhazel are you content with @sureshanaparti 's reply?

weizhouapache (approved):
code lgtm

winterhazel:
> As the controllers are already created during the VM creation process, it's better to synchronize these controller details during VM creation. Synchronizing in the attach process would still set the same controller details (if you mean dynamic reconfiguration of disk controllers in VM, maybe the existing controller reconfiguration based on the data disk controller can be done as a separate improvement later). The changes here would fix the attach issue when data disk controller is not set in the VM's template and VM detail.

@sureshanaparti I recognize that your changes fix the attachment for this case. However, they are not fixing the root of the problem, just working-around one specific situation that can lead to it. This problem does not happen because the two settings have different values. It happens due to an inconsistent behavior of the code.

I think you may have misunderstood what I'm proposing.  Just to ensure that we have the same thing in mind: my idea is execute on the volume attachment process the same logic that the VM creation/start workflows use when adding the disk controllers to the VM's spec. 

Context:
- Suppose we have a template that has `rootDiskController` set to `pvscsi` and `dataDiskController` set to `lsilogic`.

Important info:
- VMware does allow us to create a VM using a PVSCSI for the root disk and LSI Logic controllers (two different SCSI subtypes) for the data disks.

What happens on VM creation/start:
- When a VM is created/started in CloudStack, we verify the controllers defined for the root and data disk ( `rootDiskController` and `dataDiskController` details).
- We see that both are using SCSI subtypes and, even though we could create one PVSCSI for the root disk and LSI Logic controllers for data disks, we intentionally ignore the `dataDiskController` setting defined by the operator and choose to use only PVSCSI controllers for both the root disk and the data disks.

What is happening on volume attachment:
- Then, when a user tries to attach a new data disk, the process fails because we only consider `dataDiskController` and try to attach with LSI Logic instead. However, there are no LSI Logic controllers because we ignored the data disk controller chosen by the operator and opted to create PVSCI controllers for both root and data disks when creating/starting the VM.

My proposal:
- Implement the same logic we did on VM creation/start (choosing PVSCSI instead of LSI Logic for the data disks) on volume attachment.

> @winterhazel The data disk controller is matched with the SCSI root disk controller only when the data disk controller is not set in the VM's template and VM detail. If template has the root and data disk controllers set to _pvscsi_ and _osdefault_ respectively, then the controllers are created with _pvscsi_ controller to ensure root disk is configured to proper controller to create the VM. Later, any unsupported data disk controller (say, _lsilogic_) set would fail the attach disk, so the operator should ensure to set the root and data disk controller appropriately.

I don't think we have to place the responsibility on operators in this case. The attachment fails because **we** ignore the `dataDiskController` defined by the operator and **we** create/start the VM using another controller instead.


> @winterhazel are you content with @sureshanaparti 's reply?

@DaanHoogland I'm fine with this being merged, because I do acknowledge that the changes here and my proposal are not exclusive, and that it did fix one situation. However, I still think that #9400 should not be considered solved with this. The changes here do not touch the root of the problem, and only solve a specific case that would be covered by modifying the volume attachment workflow. I'll address it on another PR.

sureshanaparti:
correct @winterhazel, this addresses specific case, that is when data disk controller is not set. we can merge this and proceed with the proposed improvement (in volume attach workflow) in another PR. cc @DaanHoogland

DaanHoogland:
> @DaanHoogland I'm fine with this being merged, because I do acknowledge that the changes here and my proposal are not exclusive, and that it did fix one situation. However, I still think that #9400 should not be considered solved with this. The changes here do not touch the root of the problem, and only solve a specific case that would be covered by modifying the volume attachment workflow. I'll address it on another PR.

ok, thanks @winterhazel , I'll assign #9400 to you for futher discussion/handling.

